Issue #3, opened 2019-01-01
Train to ensemble as opposed to single point


proteneer:
Show how we can train to an ensemble of various observables (eg. dij distributions) as opposed to a single final structure.

proteneer:
Implemented via #21

PR #17, opened 2019-01-08
Add API for computing integrator Jacobians and Gradients.

proteneer:
This API cleans up the code so that it's now easier to compute derivatives of the integrator with respect to energy parameters.


proteneer:
Should make #3 easier to implement. Merging.

PR #56, opened 2019-02-12
Full hessians 2

proteneer:
LennardJones and Electrostatics GPU-fied, 16666x faster than tensorflow.


proteneer:
This first phase is now complete. Builds are broken until we have a GPU CI in place

PR #63, opened 2019-02-23
Demo of timemachine training

proteneer:


proteneer:
closing via #64 instead

PR #67, opened 2019-03-03
Training merge stable

proteneer:
Train/test splits of electrostatics!


proteneer:
merging this before the big code cleanup

PR #70, opened 2019-03-19
Improve memory usage by computing velocity derivatives

proteneer:
As part of the thinking behind doing complex-step derivatives all the way - I sort of realized that we can actually store derivatives of the velocity with respect to parameters and get an easy speedup. This PR:

1. Decreases memory usage by 100x (by not having to store the buffer windows). Note that the hessians are still an issue depending on the size of the system (requiring O(N^2) to store). When we go full complex-step we should be able to get rid of this as well.
2. Allows use of arbitrary frictions for thermostats.
3. Vastly simplifies and clears the way for other types of integrators (leap-frog/verlet etc.) 

I should've realized this much, much earlier.

### DIFF
--- a/tests/test_integrator_gpu.py
+++ b/tests/test_integrator_gpu.py
@@ -13,8 +13,6 @@ class TestGPUIntegrator(unittest.TestCase):
         # (ytz): needed to clear variables
         tf.reset_default_graph()
 
-
-
     def test_gpu_electrostatic_analytic_integration(self):
         """
         Testing that lower triangular hessians are working as intended. This is because for the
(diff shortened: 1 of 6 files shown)


proteneer:
A lesson learnt: looks like we'll be stuck with double precision for a little bit. I tried changing the accumulators for dxdp_t, dvdp_t, and even the hessian matrix product into double precision but the derivatives were still relatively unstable. It also looks like the HMPs have become the dominant bottleneck. Once we move into pure complex-step land we can re-investigate this behavior again.

proteneer:
```
train ---average batch loss--- 0.025302533140443232 range(20, 24) reduce_time: 0.39158153533935547 train_time: 5.5714662075042725
train ---average EPOCH loss--- 0.017043378483878272
starting epoch... 2 global params [0.1, 0.22406062423575715, 0.16160427492210916, 0.10046394670729042, 0.12941913232140956, 0.1, 0.1, 0.1, 0.1, 0.1, -0.1, -0.018490383528728248, 0.1, 0.1, 0.1, 0.15831161141270414, -0.2, -0.17959005484708718, -0.16074646152701616, -0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1]
test ---average batch loss--- 0.009071102473312522 range(0, 4) reduce_time: 0.36956095695495605 train_time: 3.2679779529571533
test ---average batch loss--- 0.006135443241184941 range(4, 7) reduce_time: 0.27937841415405273 train_time: 2.8300108909606934
test ---average EPOCH loss--- 0.00868945469399855
```

proteneer:
Ran for 10,000,000 steps at friction of 10.0 and temperature of 100 @ 2.5fs timestep looks stable. Interestingly enough, upping the temp to 300 and decreasing friction to 1.0 (the standard MD protocol is extremely chaotic).

proteneer:
Merging this - prepping for training to experimental data next.

PR #79, opened 2019-05-06
Initial GBSA implementation.

proteneer:
After talking to @jchodera and @maxentile I've decided to port over an initial version of GBSA (from the original tensorflow code). If this roughly looks good I will add more tests. So far the tests just ensure that first and second order derivatives with respect to geometry, parameters, and mixed are stable (no NaNs).

### DIFF
--- a/tests/test_jax_gbsa.py
+++ b/tests/test_jax_gbsa.py
@@ -1,0 +1,51 @@
+import functools
+import unittest
+import numpy as np
+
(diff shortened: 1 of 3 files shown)


maxentile:
Looks great! And I think the mask-based approach you used for avoiding nans on the diagonals looks elegant / is something I should adopt, instead of adding a nonzero value to the diagonal of the distance matrix as I had been doing previously to hack around this issue...

PR #81, opened 2019-05-08
jax langevin integrator.

proteneer:
@jchodera I've implemented a simple test-case that shows the Langevin Integrator being unstable for a relatively straightforward diatomic system. You can verify that under NVE (setting temperature to 0 and removing friction) it's actually stable. This also shows how to use a jax-based MD engine as well.


proteneer:
stale, closing

PR #82, opened 2019-05-10
Fix LennardJones typo.

proteneer:
thanks @jchodera - I've typoed this at least 5 times in 3 different ways in the past 5 years


proteneer:
fixed by #83

PR #87, opened 2019-05-14
Use lax scan for much faster compilation times.

proteneer:
Old version:

real	1m42.888s
user	1m43.148s
sys	0m0.857s

New version:

real	0m4.244s
user	0m5.040s
sys	0m0.373s


proteneer:
stale, closing

PR #88, opened 2019-05-16
optimally rotated RMSD + grads

proteneer:
@mpharrigan did the initial tensorflow version. I just did some copy-pasta into jax.

This PR:
- [x] Implements the optimally rotated RMSD code
- [x] Compute derivatives of the opt_rot_rmsd w.r.t. geometries
- [x] Tests the above code

### DIFF
--- a/tests/test_rmsd.py
+++ b/tests/test_rmsd.py
@@ -1,0 +1,83 @@
+import unittest
+from timemachine.observables import rmsd
+
+from jax.config import config; config.update("jax_enable_x64", True)
+
+from scipy.stats import special_ortho_group
+
+import numpy as onp
+import jax
+
+from jax.test_util import check_grads
+from jax.experimental import optimizers
+
+class TestRMSD(unittest.TestCase):
+
+
+    def test_rmsd(self):
+
+        x0 = onp.array([
+            [1.0, 0.2, 3.3], # H
+            [-0.6,-1.1,-0.9],# C
+            [3.4, 5.5, 0.2], # H
+            [3.6, 5.6, 0.6], # H
+        ], dtype=onp.float64)
+
+        x1 = onp.array([
+            [1.0, 0.2, 3.3], # H
+            [-0.6,-1.1,-0.9],# C
+            [3.4, 5.5, 0.2], # H
+            [3.6, 5.6, 0.6], # H
+        ], dtype=onp.float64)
+
+        onp.testing.assert_almost_equal(rmsd.opt_rot_rmsd(x0, x1), 0)
+
+        # test random translation
+        for _ in range(10):
+            offset = onp.random.rand(1, 3)*10
+            onp.testing.assert_almost_equal(rmsd.opt_rot_rmsd(x0+offset, x1), 0)
+            onp.testing.assert_almost_equal(rmsd.opt_rot_rmsd(x0, x1+offset), 0)
+
+        # generate random rotation matrix
+        for _ in range(10):
+            rot_x = special_ortho_group.rvs(3)
+            rot = onp.dot(rot_x, rot_x.T)
+            onp.testing.assert_almost_equal(rmsd.opt_rot_rmsd(onp.dot(x0, rot), x1), 0)
+            onp.testing.assert_almost_equal(rmsd.opt_rot_rmsd(x0, onp.dot(x1, rot)), 0)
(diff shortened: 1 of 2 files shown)


tests/test_rmsd.py
```diff
@@ -0,0 +1,83 @@
...
+        opt_init, opt_update, get_params = optimizers.sgd(5e-2)
+
+        x0 = onp.array([
+            [1.0, 0.2, 3.3], # H
+            [-0.6,-1.1,-0.9],# C
+            [3.4, 5.5, 0.2], # H
+            [3.6, 5.6, 0.6], # H
+        ], dtype=onp.float64)
+
+        x1 = onp.array([
```
bp-kelley: Might be clearer as x0 + onp.random...

timemachine/observables/rmsd.py
```diff
@@ -0,0 +1,55 @@
...
+import jax.scipy
+import jax.numpy as np
+
+
+def optimal_rotational_quaternion(r):
+    """Just need the largest eigenvalue of this to minimize RMSD over rotations
+    
+    References
+    ----------
+    [1] http://dx.doi.org/10.1002/jcc.20110
```
bp-kelley: It’s amazing to me how this gets recycled, although the gradient is a nice touch.

This is the earliest ref I have for these methods:

http://www.ccl.net/cca/software/SOURCES/C/quaternion-mol-fit/quatfit.txt.shtml

tests/test_rmsd.py
```diff
@@ -0,0 +1,83 @@
...
+            [1.0, 0.2, 3.3], # H
+            [-0.6,-1.1,-0.9],# C
+            [3.4, 5.5, 0.2], # H
+            [3.6, 5.6, 0.6], # H
+        ], dtype=onp.float64) + onp.random.rand(x0.shape[0],3)*10
+
+        grad_fn = jax.jit(jax.grad(rmsd.opt_rot_rmsd, argnums=(0,)))
+        opt_state = opt_init(x0)
+
+        for i in range(1500):
```
bp-kelley: Is this just for convergence?  This seems to be a lot of iterations.

tests/test_rmsd.py
```diff
@@ -0,0 +1,83 @@
...
+            g = grad_fn(get_params(opt_state), x1)[0]
+            opt_state = opt_update(i, g, opt_state)
+
+        x_final = get_params(opt_state)
+
+        assert rmsd.opt_rot_rmsd(x_final, x1) < 0.1
```
bp-kelley: .1 seems high to me.

tests/test_rmsd.py
proteneer: I can fix this in a patch

tests/test_rmsd.py
proteneer: I'm not sure how much is necessary, but 1500 iterations take less than 2 seconds to run when JITted. So it's very fast. I just wanted to set a high enough number to avoid convergence issues. With quasi-newton methods we can do a much smaller number of iterations, but this is just gradient descent here.

tests/test_rmsd.py
proteneer: I thought about this for a while too. The problem is that I can't prove for sure that the optimally rotated RMSD is actually a convex function (vanilla RMSD is). I tested this for about 50 times and it varies between 0.005 to 0.02. So I don't think it's strictly convex at least in the numerical sense (else it would converge exactly to 1e-7 under double precision)

timemachine/observables/rmsd.py
proteneer: Everyday i'm standin' on the shoulders of giants

proteneer:
@bp-kelley I think I merged this too quickly.

timemachine/observables/rmsd.py
bp-kelley: This was my goto for rmsd:  https://theobald.brandeis.edu/qcp/

Note that in the later papers one of the coauthors is now head of nx at novartis...

https://theobald.brandeis.edu/qcp/

proteneer: Nice! I ported that to the GPU back in 2012 during grad school:

https://onlinelibrary.wiley.com/doi/10.1002/jcc.23110

PR #90, opened 2019-05-22
Turbocharged GPU derivatives

proteneer:
This PR implements forward-mode derivatives of up-to second order for several classical potentials. It uses a Hessian-free complex step derivative to evaluate the batched jacobian vector product as well as the hessian vector product of an energy function ```E(x(p),p): R^3N, R^P -> R```

In addition to the energy ```E```, the forward-mode derivatives it computes are:

```
1. grad of geometry: d/dx(E) = dE/dx
2. JvP: d/dp(E) = dE/dx . dx/dp + dE/dp
3. HvP: d/dp(dE/dx) = d^2E/dx^2 . dx/dp + d^2E/dxdp
```
Implemented functionals:

- [x] HarmonicBond
- [x] HarmonicAngle
- [x] PeriodicTorsion
- [x] LennardJones
- [ ] Electrostatics
- [ ] GBSA

The design goals of this class are:

1. Work efficient in the general JVP and HVP use case. That is, they're computed in an asymptotically optimal way. Note that for cases where ```dx/dp = 0``` this code is *not* work efficient for the first order ```dE/dp```.
2. Fast to run in inference mode when we don't care about parameter derivatives.
3. Arbitrary sparsity in terms of which parameters we want to differentiate w.r.t., this is similar to doing a sparse ```jax.jacfwd``` for only a subset of the parameters. For a single kernel, max P size is 65536.
4. Batching N conformations. For a single kernel, max N size is 65536.
5. Note that ```dE/dx``` is computed symbolically and simplified by hand since their expressions are readily available in many MD packages.


timemachine/kernels/wrap_kernels.cpp
```diff
@@ -0,0 +1,159 @@
...
+#include "cpu/bonded_kernels.hpp"
+
+namespace py = pybind11;
+
+// hessian matrix product
+template<typename RealType>
+py::tuple declare_harmonic_bond_hmp(
+    const py::array_t<RealType, py::array::c_style> &coords,
+    const py::array_t<RealType, py::array::c_style> &params,
+    const py::array_t<RealType, py::array::c_style> &dxdps,
```
proteneer: @bp-kelley -  I'd like this to be an array or None

bp-kelley: You can’t really send none to a c++ reference.  You could just have empty arrays with:

 const py::array_t<RealType, py::array::c_style> dxps =  py::array_t<RealType, py::array::c_style> ()

Other wise use a unique_ptr to the optional array.

If 

As a default arg.

timemachine/kernels/wrap_kernels.cpp
```diff
@@ -0,0 +1,101 @@
...
+        memset(py_d2E_dxdp.mutable_data(), 0.0, sizeof(RealType)*num_params*num_atoms*num_dims);
+
+        RealType *dxdps_ptr = nullptr;
+
+        if(dxdps.size() > 0) {
+
+            // this is dangerously safe since it then gets passed into a function
+            // that takes in a const RealType * again
+            dxdps_ptr = const_cast<RealType *>(dxdps.data());
+            std::cout << "SETTING ACTUAL DATA" << dxdps.size() << " " << dxdps.shape() << " " << dxdps.data()[0] << std::endl;
```
proteneer: @bp-kelley 
what's weird is that if I pass in a None into a np array it actually turns into a size 1 scalar of value nan :)

proteneer: I think I need to denote None in numpy land by doing something like:

np.zeros(shape=(0,)) # this has size 0

proteneer:
So I've modified this PR to be a little more magical. With a single kernel, you can now compute:

1. E
2. dE_dX
3. dE_dp
4. d2E_dxdp

This API should be flexible enough for future use. Which should make @jchodera and @maxentile happy :)

proteneer:
If this looks good I'm going to start porting over the other functional forms. But the kernels as is should make it relatively easily to plug into custom XLA ops down the road whenever they're ready.

proteneer:
Add HarmonicAngle derivatives

proteneer:
```
This looks a bit magic, what is the rationale for tpb = 32?
```
It's what used to be the standard warp size when everyone did warp-async programming, think of it as intrinsic SIMD-width (this isn't strictly true anymore with Pascal)

proteneer:
Periodic Torsions done!

proteneer:
LennardJones done!

proteneer:
@bp-kelley are 2100 line diffs considered excessive? :)

proteneer:
PS for readers: one of the *coolest* things about this new PR is that it's "Hessian-Free", i.e. you don't need to form an explicit hessian matrix right multiplied by the grad. It's taken care of for you automatically.

proteneer:
@sschoenholz currently laying the groundwork for benchmarking against jax. The dominant computation cost for most things we care about are jvp and hvps of the energy function with respect to the parameters.

proteneer:
Giving a @SimonBoothroyd that this is happening.

proteneer:
Looks like I have to do some refactor, the new timings are far worse than computing an analytic hessian + GEMM + analytic mixed partials (0.75s vs 0.004 for a 2500 atom system.)

proteneer:
Closing this in favor of #93

PR #91, opened 2019-05-24
add scale matrix to non-periodic electrostatics

guoweiqi:
added a potential change in a comment in nonbonded.py, but I'm not sure it's correct as the forces still don't match OpenMM. Something seems to be wrong with the qij matrix, but I might just not know the conventions of the code


proteneer:
Do you mind resubmitting this so that only changes in timemachine/potentials/nonbonded.py is added? This looks like you added a lot of extraneous .pyc and .pdb files you probably weren't intending to.

PR #93, opened 2019-05-28
Explicit Hessians

proteneer:
After doing a fair bit of benchmarking it looks like doing the implicit JHVPs are too slow for our needs. So instead I'm rewriting the ops to provide explicit:

1. E
2. dE/dx
3. dE/dp
4. d^2E/dx^2 (hessian)
5. d^2E/dxdp (mixed partials)

Supported functional forms:

- [x] HarmonicBond
- [x] HarmonicAngle
- [x] PeriodicTorsion
- [x] LennardJones
- [x] Electrostatics
- [ ] GBSA

It will be up to the end user to implement GEMM to right multiply with the dxdp.


proteneer:
From now all hessians computed will be in the lower triangular form (np.tril)

proteneer:
Looks like one in fact *can* compute the GBSA Hessian in O(N^2) time! The math is crazy but the asymptotic analysis is correct since the important realization is that the born radii derivative d^2Ri/(dxj_dxk) is only needed for when either i == j or i == k, i.e. becoming and O(N^2) quantity.

PR #95, opened 2019-05-31
Langevin Optimizer/Integrator

proteneer:
This is a rewrite of the langevin optimizer/integrator that's been refactored to be state-less. All the mutable states will be stored in a single Context class (coming soon).


timemachine/kernels/wrap_kernels.cpp
```diff
@@ -10,12 +13,135 @@
...
+    using Class = timemachine::Context<RealType>;
+    std::string pyclass_name = std::string("Context_") + typestr;
+    py::class_<Class>(
+        m,
+        pyclass_name.c_str(),
+        py::buffer_protocol(),
+        py::dynamic_attr()
+    )
+    .def(py::init([](
+        const std::vector<timemachine::Potential<RealType> *> system,
```
proteneer: @bp-kelley these two pointers are the two confusing ones, Potential is a pure abstract class whose concrete subclasses are instantiated and kept around on the python side, so I need a pointer here. However, I absolutely do *not* want to transfer over ownership to the Context class since I need them to be alive after the Context Destructor is called (which is how it's currently done).

proteneer:
@guoweiqi @bp-kelley alright, this is ready to be merged in. I added a test (test_optimizers.py) to show how the timemachine's analytical derivatives match exactly as that of jax, but it's about 100x faster :)

I'll do one more round of clean-up (mainly make the folder structure a little nicer) and we can move on implementing all fun the upstream/application specific code.

@bp-kelley do you have a preference for CMake vs Bazel vs SCons vs autoconf etc.? I do need something that can find CUDA installs etc. relatively painlessly. We used cmake a extensively for OpenMM in the past but was curious if the C++ landscape has changed at all.

proteneer:
Starting source code re-factoring work now

proteneer:
Added a CMakeLists file so that we can avoid having to keep on using sh commands to everything.

proteneer:
Deleted old source code that was partly used to create the current set of custom_ops

PR #107, opened 2019-06-26
Reference jax soft-core potential

proteneer:


proteneer:
this commit is messed up.

PR #108, opened 2019-06-26
Reference Softcore Potentials.

proteneer:
This implements the standard form of the soft-core potential. Custom GPU ops WIP.


proteneer:
@bp-kelley :eyes: when you get a chance

PR #109, opened 2019-06-28
Training code cleanup

guoweiqi:
Using JSON for config files


system/training.py
```diff
@@ -91,125 +98,168 @@ def filter_groups(param_groups, groups):
...
     combined_derivs[guest_dp_idxs + len(host_params)] -= G_derivs
-    # combined_derivs = 2*delta*combined_derivs # L2 derivative
-    combined_derivs = (delta/np.abs(delta))*combined_derivs # L1 derivative
+
+    if properties['loss_fn'] == 'L2':
+        combined_derivs = 2*delta*combined_derivs
+    elif properties['loss_fn'] == 'L1':
+        combined_derivs = (delta/np.abs(delta))*combined_derivs
+    elif properties['loss_fn'] == 'Huber':
+        # Modified Huber Loss
```
proteneer: Out of curiosity, did Huber loss perform better than L1/L2?

system/training.py
```diff
@@ -91,125 +98,168 @@ def filter_groups(param_groups, groups):
...
+    
+    if optimizer == 'Adam':
+        opt_init, opt_update, get_params = optimizers.adam(lr)
+    elif optimizer == 'SGD':
+        opt_init, opt_update, get_params = optimizers.sgd(lr)
+    elif optimizer == 'RMSProp':
+        opt_init, opt_update, get_params = optimizers.rmsprop(lr)
+    elif optimizer == 'Adagrad':
+        opt_init, opt_update, get_params = optimizers.adagrad(lr)
+    elif optimizer == 'SM3':
```
proteneer: Probably best to add a final catch else to throw an exception.

system/training.py
```diff
@@ -91,125 +98,168 @@ def filter_groups(param_groups, groups):
...
+    
+    config_file = glob.glob('*.json')
+    if len(config_file) == 0:
+        raise Exception('config file not found')
+    elif len(config_file) > 1:
+        raise Exception('multiple config files found')
+    with open(config_file[0], 'r') as file:
+        config = json.load(file)
+   
+    globals()['properties'] = config
```
proteneer: don't dump this into globals(). This part of the code has global visibility anyways so simply doing ```properties = config``` should work

system/training.py
guoweiqi: Depended on the system. When I was doing absolute energies with the fully minimized host-guests, Huber performed better than L1 & L2, but L2 has seemed to work fine so far for relative fitting

proteneer:
@guoweiqi I think your github commit messages don't have your correct identity. Can you set it to your github username?

https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration

I think that's interfering with the tests being ran.

proteneer:
I'm going to have to merge this by hand later. There's a little bit too much stuff going on. I'm not sure why some of the other files were changed

PR #110, opened 2019-07-02
4 dimensional potentials

proteneer:


proteneer:
Merging this one in. Seems like 4D minimization works out directly out of the box (save for the lack of a proper cross product in 4D).

proteneer:
(by pretty well I mean it minimizes correctly, not that it's substantively better than other methods)

PR #112, opened 2019-08-07
Initial Implementation of n dimensional nonbonded kernels.

proteneer:
This PR adds support for arbitrary dimensional electrostatic potentials (along with their second order derivatives). After this will be LennardJones.


timemachine/cpp/src/custom_nonbonded_gpu.cu
```diff
@@ -130,21 +132,42 @@ void Electrostatics<RealType>::derivatives_device(
...
-        d_E,
-        d_dE_dx,
-        d_d2E_dx2,
-        // parameter derivatives
-        num_dp,
-        d_param_gather_idxs,
-        d_dE_dp,
-        d_d2E_dxdp
-    );
+    if(num_dims == 3) {
```
proteneer: @bp-kelley if only we could JIT compile templates at runtime. I don't have any better ways to do this than to specialize apriori, any better ideas?

timemachine/cpp/tests/test_potentials.py
```diff
@@ -378,4 +380,20 @@ def test_aperiodic_electrostatics(self):
...
+        # test 4 dimensional hessians
+        conf4d = np.array([
+            [ 0.0637,   0.0126,   0.2203,  0.5],
+            [ 1.0573,  -0.2011,   1.2864, -0.2],
+            [ 2.3928,   1.2209,  -0.2230,  5.6],
+            [-0.6891,   1.6983,   0.0780,  2.3],
+            [-0.6312,  -1.6261,  -0.2601, -5.1]
+        ], dtype=np.float64)
+
+        self.assert_derivatives(
```
proteneer: huzzah!

proteneer:
This was less painful than I had thought. Context and nonbonded terms are now ported to 4D. GPU optimized dU/dlambda here we come!

proteneer:
Okay this is complete and ready to merge. I had to port over the bonded terms to still be in 3D but write out to a 4D Hessian/Mixed Partial buffer, not fun!

timemachine/cpp/src/kernels/k_electrostatics.cuh
```diff
@@ -1,104 +1,78 @@
...
-        x0 = 0.0;
-        y0 = 0.0;
-        z0 = 0.0;
         q0 = 0.0;
         q0_g_idx = 0;
     } else {
-        x0 = coords[conf_idx*N*3+i_idx*3+0];
-        y0 = coords[conf_idx*N*3+i_idx*3+1];
-        z0 = coords[conf_idx*N*3+i_idx*3+2];
+        #pragma unroll
```
bp-kelley: Is NDIMS really const here?  I’m unsure if the compiler can tell.

proteneer: it's templated to instances of ndims 3 and 4

PR #114, opened 2019-08-16
Force field training for RMSD and Enthalpy calculations

guoweiqi:
Force field parameter training protocol for small molecule RMSD and absolute/relative binding enthalpy. See the example files for setup details


proteneer:
Thanks for the exceptionally hardwork on this! Merging this.

PR #117, opened 2019-09-22
Cherry-Pick me

proteneer:
This PR is a temporary placeholder to be Cherry-picked from at a subsequent stage. The code base is in a state of total chaos while I was doing some science. The good news is that I now have a very good handle how to stabilize these extremely finnicky derivatives.


proteneer:
This will be redone later.

PR #118, opened 2020-01-28
Reverse Mode Derivative implementation

proteneer:
This PR completely rewrites the timemachine to use reverse automatic differentiation via a set of custom adjoint equations. The PR consists of:

1. Using fixed point precision when accumulating forces, with a decimal placement of 2^41. This means the method is safe for absolute force components of less 2^(64-41)=4194304 in default MD units.
2. The electrostatics no longer natively multiply by ONE_4PI_EPS0 (138.935456). It is now the responsibility of the forcefield to multiply all charges by sqrt(ONE_4PI_EPS0) to ensure consistency. This is required to make the MD engine unit entirely agnostic.
3. The adjoint equations of motion are natively implement on the GPU for reverse mode, allowing us to simultaneously compute derivatives with respect to P parameters in one reverse pass/
4. Analytical second order derivatives of potential are removed in favor of complex-stepped variant. This would allow us to easily port other potentials where only the first derivatives are known. We then use a trick where the vHp == Hvp for conserved forces to implement the vector Hessian products using the complex step derivative.
5. It's about 50x faster than the previous implementation due to the remove of the O(P) scaling and lack of need to compute explicit analytical hessians.


proteneer:
Hot-merging this in due to the extremely large amount of somewhat incoherent changes.

PR #121, opened 2020-01-28
Cleanup forcefield code with the new kernels.

proteneer:
This cleans up the deserialization code and the openforcefield parsers to use the updated classes from the last big refactor.


proteneer:
This is another cleanup PR. Hot-merging this.

PR #123, opened 2020-01-29
Sample code demonstrating timemachine usage

proteneer:
Just adding some example files showing how to use the new reverse mode derivatives.


system/forcefield.py
```diff
@@ -506,7 +503,7 @@ def add_param(p, p_group):
         for smirks, param in model.items():
 
             # small charges
-            param = param*np.sqrt(constants.ONE_4PI_EPS0)
+            param = param*np.sqrt(constants.ONE_4PI_EPS0)/2
```
proteneer: self reminder to remove this later

PR #132, opened 2020-02-05
Speed-ups using better math functions

proteneer:
This results in about a 12% speed up across the board.


fe/parameterize_lambda.py
```diff
@@ -0,0 +1,199 @@
...
+def recenter(conf, true_com, scale_factor=1):
+    mol_com = np.sum(conf, axis=0)/conf.shape[0]
+    centered = conf - mol_com  # centered to origin
+    return true_com + centered/scale_factor 
+
+from hilbertcurve.hilbertcurve import HilbertCurve
+
+def hilbert_sort(conf):
+    hc = HilbertCurve(16, 3)
+    int_confs = (conf*1000).astype(np.int64)+10000
```
bp-kelley: 1000 and 10000 look like magic numbers and probably should be defined as constants somewhere

fe/parameterize_lambda.py
```diff
@@ -0,0 +1,199 @@
...
+                step_sizes,
+                cas,
+                lambda_schedule,
+                perm
+            )
+
+            num_conformers = 1
+
+            guest_mol.RemoveAllConformers()
+            AllChem.EmbedMultipleConfs(guest_mol, num_conformers, randomSeed=2020)
```
bp-kelley: You should check for failure here, I think if the result is -1 no confs were generated.

fe/parameterize_lambda.py
```diff
@@ -0,0 +1,199 @@
...
+
+                combined_pdb = Chem.CombineMols(Chem.MolFromPDBFile(host_pdb_file, removeHs=False), init_mol)
+                combined_pdb_str = StringIO(Chem.MolToPDBBlock(combined_pdb))
+                out_file = os.path.join("frames", "epoch_"+str(epoch)+"_"+mode+"_"+host_name+"_conf_"+str(conf_idx)+".pdb")
+                writer = PDBWriter(combined_pdb_str, out_file)
+                all_args.append((x0, writer,  conf_idx % num_gpus))
+
+            sim.run_forward_multi(all_args[0])
+
+            sys.exit(0)
```
bp-kelley: This is def a code smell

fe/math_utils.py
```diff
@@ -0,0 +1,28 @@
...
+    Isomorphic API to numpy's trapz. Refer to np.trapz for documentation.
+    """
+    d = jnp.diff(x)
+    # reshape to correct shape
+    shape = [1]*y.ndim
+    shape[axis] = d.shape[0]
+    d = d.reshape(shape)
+
+    nd = y.ndim
+    slice1 = [slice(None)]*nd
```
bp-kelley: I’m assuming it’s ok to have nd copies of exactly the same slice object?

fe/parameterize_simple.py
```diff
@@ -0,0 +1,222 @@
...
+                mode,
+                step_sizes,
+                cas,
+                lambda_schedule,
+                perm
+            )
+
+            # sample from the rdkit DG distribution (this can be changed later to another distribution later on)
+            all_args = []
+            for conf_idx in range(num_conformers):
```
bp-kelley: This code block is use a bit, perhaps a simulation runner helper is in order?

fe/simulation.py
```diff
@@ -0,0 +1,200 @@
...
+            pdb_writer.write_header()
+            xs = ctxt.get_all_coords()
+            for frame_idx, x in enumerate(xs):
+
+                interval = max(1, xs.shape[0]//pdb_writer.n_frames)
+                if frame_idx % interval == 0:
+                    # argsort is iperm and
+                    pdb_writer.write((x*10)[self.iperm])
+        # pdb_writer.close()
+        del stepper
```
bp-kelley: These are going to be deleted anyway due to ref counts

PR #133, opened 2020-02-05
Easy speed up using reciprocals.

proteneer:
This optimizes the norm calculation for both real and complex distance functions to use rnorm variants (which have only 1/2 ULPS of precision loss in double and single precision respectively).

### DIFF
--- a/timemachine/cpp/src/kernels/k_nonbonded_deterministic.cuh
+++ b/timemachine/cpp/src/kernels/k_nonbonded_deterministic.cuh
@@ -35,6 +35,14 @@ void __global__ k_find_block_bounds(
 
 }
 
+__device__ float overloaded_sqrt(const float x) {
+    return sqrtf(x);
+}
+
+__device__ double overloaded_sqrt(const double x) {
+    return sqrt(x);
+}
+
 template<typename RealType, int D>
 __device__ RealType fast_vec_rnorm(const RealType v[D]);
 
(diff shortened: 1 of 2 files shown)


timemachine/cpp/src/kernels/k_nonbonded_deterministic.cuh
```diff
@@ -114,26 +182,28 @@ void __global__ k_nonbonded_jvp(
     Surreal<RealType> g_sigj(0.0, 0.0);
     Surreal<RealType> g_epsj(0.0, 0.0);
 
+    RealType inv_cutoff = 1/cutoff;
```
bp-kelley: I would use Static_cast<RealType> 1.0/cutoff if just to avoid compiler earnings.

bp-kelley: If you want to be really fancy see rd_cast in the rdkit sources.  In debug mode it throws an error on overflow.

timemachine/cpp/src/kernels/k_nonbonded_deterministic.cuh
```diff
@@ -275,21 +345,23 @@ void __global__ k_nonbonded_inference(
     RealType sig_j = atom_j_idx < N ? params[lj_param_idx_sig_j] : 1;
     RealType eps_j = atom_j_idx < N ? params[lj_param_idx_eps_j] : 0;
 
+    RealType inv_cutoff = 1/cutoff;
+
     // In inference mode, we don't care about gradients with respect to parameters.
     for(int round = 0; round < 32; round++) {
```
bp-kelley: 32 looks magic, perhaps use a named constant?

proteneer:
We can get a bit more of a speed up (another 12%) if we don't differentiate with respect to the vdw parameters - but I think that's a little too aggressive for now so I'm going to keep this as is.

timemachine/cpp/src/kernels/k_nonbonded_deterministic.cuh
proteneer: I'm using the templated OtherType method we spoke about so there shouldn't be any warnings.

timemachine/cpp/src/kernels/k_nonbonded_deterministic.cuh
proteneer: 32 is WARPSIZE - we will have a lot of problems on our hands if this is ever modified to be not 32

PR #139, opened 2020-02-12
Enable improper dihedral calculation using trefoil mode

chuantian:


system/forcefield.py
```diff
@@ -264,19 +265,26 @@ def add_param(p, p_group):
...
                     all_phase_idxs.append(phase_idx)
                     all_period_idxs.append(period_idx)
-
-                for m in matches:
+                
+                # remove the redundant matches and use unique pair for trefoil
+                new_matches = []
+                
+                for element in matches:
+                    tmp=[element[0], element[2], element[3]]
```
proteneer: Style nitpick

Add a space before and after the =

proteneer:
1. Do you also mind adding a bugfix for the combiner code that didn't apply the permutation for the solvent leg?
2. Are the distortions in the complex leg resolved?

proteneer:
Another question: are we guaranteed that a given atom is at most the center of exactly one set of trefoils?

system/forcefield.py
chuantian: Got it!

chuantian:
If we skip the permutation, distortion issue for both solvent leg and complex leg will be fixed.
For an improper 1-2-3-4, 2 will be the center atom for all three sets of trefoil.
The three sets of trefoil will be: 2-1-3-4, 2-3-4-1, 2-4-1-3

proteneer:
Unfortunately we can't skip the permutation when we use a cutoff. So we do need to fix the permutation code as well for this PR. If that's too difficult let me know and I can submit something later.

proteneer:
Merging this as is! Thanks for the contribution again.

PR #153, opened 2020-02-29
New schedule

proteneer:
This PR:

- [x] Allows the end user to choose an arbitrary lambda schedule define on [0, inf)
- [x] Contains a small number of optimizations to the GPU code
- [x] Adds support for parallel, multi-GPU training.
- [x] Cleans up extraneous exponents from last commit

### DIFF
--- a/fe/parameterize_simple.py
+++ b/fe/parameterize_simple.py
@@ -61,7 +61,7 @@ class PDBWriter():
         self.pdb_str = pdb_str
         self.out_filepath = out_filepath
         self.outfile = None
-        self.n_frames = 10
+        self.n_frames = 25
 
     def write_header(self):
         """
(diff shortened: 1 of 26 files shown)


proteneer:
new_schedule

In particular, one of the *really* nice properties about this method is that it guarantees that dU/dl is zero at the alchemical endpoints (for both insertion and deletion!)

fe/parameterize_simple.py
```diff
@@ -105,27 +119,37 @@ def close(self):
...
 
-    num_gpus = 1
-    pool = multiprocessing.Pool(num_gpus)
+    print('Creating multiprocessing pool with',args.num_gpus, 'gpus and', args.jobs_per_gpu, 'jobs per gpu')
+    pool = multiprocessing.Pool(num_workers)
 
     all_du_dls = []
 
-    lambda_schedule = np.linspace(0.00001, 0.99999, num=T)
+    start = 1e3
```
proteneer: @chuantian  This is the new lambda schedule

proteneer: It starts by placing the 4th dimension at +1000, then moves it quickly to 1nm in about 500 steps. Then it linearly moves it down to zero. I think we should probably sample the 0.1-0.3 sample more aggressively as well.

proteneer:
I'm going to merge this in and deal with numerical issues in a separate PR.

PR #155, opened 2020-03-03
Bugfixes for BAR

proteneer:
Hot fix for this incredibly stupid bug on my end.

Suppose we were computing fwd_dG by integrating from [a,b] and bkwd_dG by integrating from [b,a], we need to flip the sign of bkwd_dG so that it's equivalent to integrating from [a,b] before passing it in to BAR.

This resolves the issue of not being able to train on a single conformer, as well as maintaining the ability to train on multiple conformers.


proteneer:
@chuantian I'm hot-merging this

PR #162, opened 2020-03-06
Gb batch training

chuantian:
This PR implement batch training for GBSA, only small molecule force field is optimized in this case. @proteneer I think this need your expertise on making sure everything is correct, such as CV, etc.


fe/parameterize_batch.py
```diff
@@ -0,0 +1,387 @@
...
+    parser.add_argument('--num_gpus', type=int, required=True)
+    parser.add_argument('--jobs_per_gpu', type=int, required=True)
+    parser.add_argument('--num_conformers', type=int, required=True)
+    parser.add_argument('--true_dG_file', type=str, nargs='*')
+
+    args = parser.parse_args()
+
+    assert os.path.isdir(args.frames_dir)
+
+    true_dG_list = []
```
proteneer: Is this even used?

fe/parameterize_batch.py
```diff
@@ -0,0 +1,387 @@
...
+        # epoch_params = get_params(opt_state)
+        # guest_params = epoch_params
+
+        train.shuffle()
+        print(train.data[0])
+
+###### Training ######
+        for i in range(len(train.data)):
+
+            for guest_mol in train.data[i]:
```
proteneer: Why is each element of data itself a list?

fe/parameterize_batch.py
```diff
@@ -0,0 +1,387 @@
...
+    assert T % 2 == 0
+    cas = np.ones(T)*0.93
+    epoch = 0
+    lr = 5e-4
+
+    ligand_list = []
+    for ligand_sdf_file in args.ligand_sdf_file:
+        f = open(ligand_sdf_file, 'r').readlines()
+        for file_name in f:
+            ligand_list.append(Chem.SDMolSupplier(file_name.rstrip('\n'), removeHs=False))
```
proteneer: I would make a single ligand sdf file with all the molecules, and attach the true_dG as a property. Then feed in a tuple of (mol, true_dG) as elements of data.

fe/parameterize_batch.py
```diff
@@ -0,0 +1,387 @@
...
+    host_conf = np.array(host_conf)
+    host_name = "complex"  
+
+    #initial_params = sim.combined_params
+    #opt_init, opt_update, get_params = optimizers.adam(lr)
+    #opt_state = opt_init(initial_params)
+
+    num_epochs = 100
+    # torsion, charge, gb radii, gb scaling factor, torsion
+    vary_allowed_groups = [4]*10 + [17]*20 + [18]*20 + [19]*20 + [4]*30
```
proteneer: This will need to be made more flexible later on.

fe/parameterize_batch.py
```diff
@@ -0,0 +1,387 @@
...
+    for epoch in range(num_epochs):
+        # sample from the rdkit DG distribution (this can be changed later to another distribution later on)
+        # epoch_params = get_params(opt_state)
+        # guest_params = epoch_params
+
+        train.shuffle()
+        print(train.data[0])
+
+###### Training ######
+        for i in range(len(train.data)):
```
proteneer: loop over the data using iterbatches(i), this was designed initially to iterate over any batch size (including 1).

```
for data in train.iterbatches(i):
      guest_mol = data[0]
```

fe/parameterize_batch.py
```diff
@@ -0,0 +1,387 @@
...
+          
+            for du_dls in all_du_dls:
+                fwd = du_dls[:T//2]
+                bkwd = du_dls[T//2:]
+
+            error = error_fn(all_du_dls, T, lambda_schedule, true_dG)
+
+            print("---EPOCH", epoch, "---- LOSS of one ligand in training ----", error)
+
+            error_grad = loss_grad_fn(all_du_dls, T, lambda_schedule, true_dG)
```
proteneer: not sure what true_dG here is pointing to since you don't use the list of true_dGs anywhere

fe/parameterize_batch.py
```diff
@@ -0,0 +1,387 @@
...
+    elif args.precision == 'double':
+        precision = np.float64
+    else:
+        raise Exception("precision must be either single or double")
+
+    num_gpus = args.num_gpus
+    num_workers = args.num_gpus*args.jobs_per_gpu
+
+    print('Creating multiprocessing pool with',args.num_gpus, 'gpus and', args.jobs_per_gpu, 'jobs per gpu')
+    pool = multiprocessing.Pool(num_workers)
```
proteneer: You need to call pool.close() at some point.

fe/parameterize_batch.py
```diff
@@ -0,0 +1,387 @@
...
+    #opt_state = opt_init(initial_params)
+
+    num_epochs = 100
+    # torsion, charge, gb radii, gb scaling factor, torsion
+    vary_allowed_groups = [4]*10 + [17]*20 + [18]*20 + [19]*20 + [4]*30
+
+    assert len(vary_allowed_groups) == num_epochs
+
+    ligand_dataset = dataset.Dataset(ligand_list)
+    ligand_dataset.iterbatches(1)
```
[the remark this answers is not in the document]
chuantian: Right, I'll remove this line.

fe/parameterize_batch.py
chuantian: Got it!

fe/parameterize_batch.py
chuantian: I changed both training and test accordingly.

fe/parameterize_batch.py
```diff
@@ -0,0 +1,381 @@
...
+
+    parser = argparse.ArgumentParser(description='Quick Test')
+    parser.add_argument('--frames_dir', type=str, required=True)
+    parser.add_argument('--precision', type=str, required=True)    
+    parser.add_argument('--complex_pdb', type=str, required=True)
+    parser.add_argument('--ligand_sdf_file', type=str, nargs='*')
+    parser.add_argument('--num_gpus', type=int, required=True)
+    parser.add_argument('--jobs_per_gpu', type=int, required=True)
+    parser.add_argument('--num_conformers', type=int, required=True)
+    parser.add_argument('--true_dG_file', type=str, nargs='*')
```
proteneer: I would add a docstring clarifying the sign of the dG values (either insertion or deletion) and if they're in kJ/mol or kcal/mol. This (along with ligand_sdf_file) should be required.

fe/parameterize_batch.py
```diff
@@ -0,0 +1,385 @@
...
+    backward_schedule = forward_schedule[::-1]
+    lambda_schedule = np.concatenate([forward_schedule, backward_schedule])
+
+    T = lambda_schedule.shape[0]
+    assert T % 2 == 0
+    dt = 0.001
+    step_sizes = np.ones(T)*dt
+    assert T % 2 == 0
+    cas = np.ones(T)*0.93
+    epoch = 0
```
proteneer: this epoch is not used. the for loop below redefines epoch

fe/parameterize_batch.py
```diff
@@ -0,0 +1,385 @@
...
+
+    guest_mol = ligand_list[0][0]
+        
+    off = smirnoff.ForceField("test_forcefields/smirnoff99Frosst.offxml")
+    _, (guest_params, _), _ = forcefield.parameterize(guest_mol, off, dimension=4)
+    
+    opt_init, opt_update, get_params = optimizers.adam(lr)
+    opt_state = opt_init(guest_params)
+
+    np.savetxt("original_off_params.txt", guest_params)
```
proteneer: you probably want to call this again after the end of every epoch.

proteneer: for your own sake, probably best to also save the corresponding guest_param_groups

chuantian:
I think we can skip this version

PR #163, opened 2020-03-08
Forcefield Updates

proteneer:
This PR implements the following

- [x] Upgrade to the most recent smirnoff 1.1 release.
- [x] Implement a converter from current smirnoff xml files to a specialized format that we can add GB and other *highly experimental* forcefields to.
- [x] Better grouping support so we can select which parameters to modify easily.
- [x] Generates indices into an opaque array of parameters (which can then be used by fancier optimizers such as jax.experimental.optimizers/scipy.optimize)
- [x] Allow for easy serialization *back* into a human readable form.
- [x] Avoid the need to parameterize a "dummy variable" initially
- [x] Cleans up the improper torsion code
- [x] Delay choice of dimensions and precision until the last minute
- [x] System class is now easily serializable, since its comprised of only PoDs.
- [x] Removes the need for doing a second forward pass by using pipes
- [x] Changes the small molecule forcefield groups to:
```
        group_map = {
            "Angle": (0,1),
            "Bond": (2,3),
            "Improper": (4,5,6), # k, phase, period
            "Proper": (7,8,9), # k, phase, period
            "vdW": (10,11),  # sigma, epsilon
            "GBSA": (12,13), # radii, scale factor
            "SimpleCharges": (14,)
        }
```
- [x] The protein forcefield is the above except they're offset by +100
- [x] Cleanup multigpu, multiconformer, and multi jobs per gpu support.

We will still need to implement a specialized jacobian when system parameters become a function forcefield parameters. (Eg. AM1BCC)

This code is *not* meant to be used outside of the timemachine since it very deeply assumes OpenMM's unit system for the sake of simplicity. A back-adapter into SMIRNOFF proper will be provided later on.


jchodera:
> Implement a converter from current smirnoff xml files to a specialized format that we can add GB and other highly experimental forcefields to.

The whole point of SMIRNOFF and its `openforcefield` toolkit parameter handlers is to make it easily extensible to try new things without having to change the toolkit code. If you find you need to convert it to _another_ format for this purpose, we've utterly failed in meeting these goals, and feedback on why this is easier---and what we'd need to do to make it easier for you to use the toolkit---would be very much appreciated!

Tagging @j-wags

proteneer:
> > Implement a converter from current smirnoff xml files to a specialized format that we can add GB and other highly experimental forcefields to.
> 
> The whole point of SMIRNOFF and its `openforcefield` toolkit parameter handlers is to make it easily extensible to try new things without having to change the toolkit code. If you find you need to convert it to _another_ format for this purpose, we've utterly failed in meeting these goals, and feedback on why this is easier---and what we'd need to do to make it easier for you to use the toolkit---would be very much appreciated!
> 
> Tagging @j-wags

I understand where you're coming from. This is a temporary stop-gap because there are too many things we need to try in the near future that aren't going to be fully rigorous w.r.t. your spec. When prototyping functional form development, we don't have time to:

1. Define a spec that 5 PIs need to agree on.
2. Build a parser with all the kinks worked out for {AMBER/OpenMM/GROMACS/TimeMachine} x {RDKit/OEChem}.
3. Easily be able to tinker with things like differentiable BCCs *without* obsessing over units and every miniscule detail.

Once we've **validated** our parameters, we'd be more than happy to contribute back a proper spec upstream to OFF. But until we've done so, we need to move quickly.

ff/forcefield.py
```diff
@@ -0,0 +1,118 @@
...
+        group_map = {
+            "Angle": (0,1),
+            "Bond": (2,3),
+            "Improper": (4,5,6),
+            "Proper": (7,8,9),
+            "vdW": (10,11),
+            "GBSA": (12,13),
+            "SimpleCharges": (14,)
+        }
+
```
chuantian: Just to confirm, this indexing will conflict with protein ones right?

proteneer: No - the protein forcefield groups will be the small molecule force groups + 100. I haven't added the protein converter code yet.

PR #165, opened 2020-03-10
Implement batch training for new open ff codebase

chuantian:


fe/parameterize_batch.py
```diff
@@ -0,0 +1,523 @@
...
+                lambda_schedule,
+                lambda_idxs,
+                precision
+            )
+
+            epoch_train_params = get_params(opt_state)
+
+            epoch_train_ff_params = copy.deepcopy(open_ff)
+            epoch_train_ff_params.params = epoch_train_params[len(host_system.params):]
+            fname = "epoch_"+str(epoch)+"_dG_"+str(true_dG)+"_params"
```
proteneer: why do we put the true_dG in the filename?

fe/parameterize_batch.py
```diff
@@ -0,0 +1,523 @@
...
+                du_dls = pc.recv()
+                all_du_dls.append(du_dls)
+
+            all_du_dls = np.array(all_du_dls)
+            loss_grad_fn = jax.grad(error_fn, argnums=(0,)) 
+
+            error = error_fn(all_du_dls, T, lambda_schedule, true_dG)
+
+            print("---EPOCH", epoch, "---Train---LOSS", error)
+            with open(args.out_dir+"/loss.txt", "a+") as f:
```
proteneer: always do os.path.join(args.out_dir, "loss.txt")

proteneer: (Also do this for the earlier check to see if the file exists)

fe/parameterize_batch.py
```diff
@@ -0,0 +1,523 @@
...
+
+            guest_mol = copy.deepcopy(data[0][0])
+            true_dG = data[0][1] # in kJ/mol and positive sign 
+
+            init_conf = guest_mol.GetConformer(0)
+            init_conf = np.array(init_conf.GetPositions(), dtype=np.float64)
+            init_conf = init_conf/10 # convert to md_units
+            conf_com = com(init_conf)
+
+            init_mol = Chem.Mol(guest_mol)
```
proteneer: Now I see why you need to use the deepcopy!

fe/parameterize_batch.py
```diff
@@ -0,0 +1,523 @@
...
+            amber_ff = app.ForceField('amber99sbildn.xml', 'amber99_obc.xml')
+            host_system = amber_ff.createSystem(host_pdb.topology,
+                nonbondedMethod=app.NoCutoff,
+                constraints=None,
+                rigidWater=False)
+
+            host_system = openmm_converter.deserialize_system(host_system)
+            num_host_atoms = len(host_system.masses)
+
+            # open_ff = forcefield.Forcefield("ff/smirnoff_1.1.0.py")
```
proteneer: we can probably remove this comment

fe/parameterize_batch.py
```diff
@@ -0,0 +1,523 @@
...
+                        print("derivs", g_idx, '\t group', gp, '\t', g, '\t adjusted to', g*pf, '\t old val', sim.system.params[g_idx])
+                else:
+                    filtered_grad.append(0)
+
+            filtered_grad = np.array(filtered_grad)
+            opt_state = opt_update(epoch, filtered_grad, opt_state)
+
+        for data in test.iterbatches(1):
+
+            guest_mol = copy.deepcopy(data[0][0])
```
proteneer: The test set has a *lot* of duplicate code shared with the training set - can you find a way to refactor them out? If we change something in the training code we'd also have to remember to change it exactly the same way in the test code. 

For example
- the host_pdb code is identical, and can be moved entirely out of the training loop
- lambda schedules etc are the same.

fe/parameterize_batch.py
```diff
@@ -0,0 +1,523 @@
...
+
+    for ligand_sdf_file in args.ligand_sdf:
+        for rd_mol in Chem.SDMolSupplier(ligand_sdf_file):
+            dG_value = rd_mol.GetProp("_dG")
+            ligand_list.append((rd_mol, float(dG_value)))
+
+    ligand_dataset = dataset.Dataset(ligand_list)
+    train, test = ligand_dataset.split(0.7)
+
+    if os.path.exists(args.out_dir+"/loss.txt"):
```
proteneer: what I've been doing is just pipe runs to to a file then grep for LOSS afterwards

fe/parameterize_batch.py
```diff
@@ -0,0 +1,523 @@
...
+                p = Process(target=sim.run_forward_and_backward, args=arg)
+                p.daemon = True
+                processes.append(p)
+                p.start()
+
+            all_du_dls = []
+            for pc in parent_conns:
+                du_dls = pc.recv()
+                all_du_dls.append(du_dls)
+
```
proteneer: you need to close parent conns with pc.close() here else it will leak processes (run nvidia-smi and you should see memory usage start to creep up). You will also run out of fds on your os. 

I think you will also have to wrap the simulation.py code with:
``` python
simulation.py: run_forward_and_backward()

try:
        du_dl_adjoints = pipe.recv()
        stepper.set_du_dl_adjoint(du_dl_adjoints)
        ctxt.set_x_t_adjoint(np.zeros_like(x0))
        start = time.time()
        ctxt.backward_mode()
        print("bkwd run time", time.time() - start)
        dL_dp = ctxt.get_param_adjoint_accum()
        pipe.send(dL_dp)
except EOFError:
        pass
pipe.close()
return
```

fe/parameterize_batch.py
chuantian: I use ligand_id just for now, but it won't associate with one particular ligand every epoch since we do shuffle.

fe/parameterize_batch.py
chuantian: Yes, that's okay too. I save this just for monitoring the loss more easily.

fe/parameterize_batch.py
```diff
@@ -0,0 +1,523 @@
...
+                parent_conn, child_conn = Pipe()
+                parent_conns.append(parent_conn)
+                # writer can be None if we don't care about vis
+                all_args.append([x0, v0, conf_idx % num_gpus, writer, child_conn])
+
+            processes = []
+
+            for arg in all_args:
+                p = Process(target=sim.run_forward_and_backward, args=arg)
```
proteneer: Process() called once here

fe/parameterize_batch.py
```diff
@@ -0,0 +1,523 @@
...
```
proteneer: Process() called second time here

fe/parameterize_batch.py
```diff
@@ -0,0 +1,487 @@
...
+
+            all_du_dls = []
+            for pc in parent_conns:
+                du_dls = pc.recv()
+                all_du_dls.append(du_dls)           
+                pc.close()
+
+            # terminate all the processes
+            for p in processes:
+                p.terminate()
```
chuantian: I have to do p.terminate in the test batch instead of p.join because the process doesn't join properly.

proteneer: hm why doesn't it join properly? that's a little concerning

proteneer:
Thanks for the excellent contribution. Merging

PR #168, opened 2020-03-18
Improper Torsion + Charge Fix.

proteneer:
Improper Torsions are *not* symmetric, so we can't use sorted atom_idxs from the valence dict.


proteneer:
@chuantian Can you let me know if this looks okay to you?

ff/forcefield.py
```diff
@@ -272,7 +267,7 @@ def make_vd():
 
             elif force_type == 'Improper':
 
-                for atom_idxs, p_idx in vd.items():
+                for _, (p_idx, atom_idxs) in vd.items():
```
chuantian: Here, the unsorted atom_idxs are preserved.

PR #173, opened 2020-03-25
Exp avg

proteneer:
@chuantian


proteneer:
@chuantian I need to do some cleanup tomorrow (or today, since its 2:30am). But this protocol was very stable on the merck benchmark hif2a.

chuantian:
Awesome job, Yutong! We finally get this stability issue out of the way.

PR #179, opened 2020-04-02
Revert back to the standard OBC1/2 parameters.

proteneer:
We were using @maxentile's optimized GB parameters (based on his work parameterizing FreeSolv). Unfortunately they are extremely unstable for our protein-ligand systems. So we're reverting back to the original OBC2 parameters taken from:

https://github.com/openforcefield/openforcefield/blob/master/openforcefield/data/test_forcefields/GBSA_OBC2-1.0.offxml

@chuantian


jchodera:
Would love to hear more about what you found, especially in informing what additional targets we should include in GB parameter optimization.

proteneer:
@jchodera I'm investigating. This is mainly a fail-safe to remove another source of error.

proteneer:
(ps don't merge this yet - I need to still hunt some more things down. It's 2AM and I probably can't sleuth very well)

chuantian:
By the way, if we are reverting, we should probably do OBC1 for now. This is related to the protein charge dependency and consistency issue brought about in another discussion. In addition, the ff/smirnoff_1.1.0.py in master branch is loading OBC1 alpha/beta/gamma parameters.

proteneer:
OBC1 and 2 in the openforcefield test files have exactly the same atom type and parameters. It's only the alpha/beta/gamma that differ. Can you confirm this? @chuantian

chuantian:
Yes, I compared GBSA_OBC1-1.0.offxml and GBSA_OBC2-1.0.offxml in openforcefield/openforcefield/data/test_forcefields/
They have exactly same parameters.

proteneer:
I meant confirming with the AMBER OBC parameters - I already checked the OFF params.

ff/smirnoff_1.1.0.py
```diff
@@ -343,15 +343,17 @@
...
+       ['[#1:1]~[#7]', 0.13,  0.85],
+       ['[#6:1]', 0.17,  0.72],
+       ['[#7:1]', 0.155,  0.79],
+       ['[#8:1]', 0.15,  0.85],
+       ['[#9:1]', 0.15,  0.88],
+       ['[#14:1]', 0.21,  0.8],
+       ['[#15:1]', 0.185,  0.86],
+       ['[#16:1]', 0.18,  0.96],
+       ['[#17:1]', 0.17,  0.8],
     ],
```
chuantian, line 357: These all match with the "really old" implementation in Amber (located in $AMBERHOME//AmberTools/src/leap/src/leap/unitio.c). 
In addition, the openMM implementation makes some stupid mistakes: H-N hydrogen GB radii is mistakenly switched with H-C hydrogen GB radii, etc. I am probably wrong, but they are really messed up in openmm/wrappers/python/simtk/openmm/app/data/amber*_obc.xml

jchodera: Any chance you can open an issue to point out these problems?
We need to revisit how we apply GBSA parameters anyway!

http://github.com/openmm/openmm/issues

chuantian: Yeah I'll do that!

PR #181, opened 2020-04-07
Bugfixes for parameterize_exp and simulation py

proteneer:
@chuantian 

This still uses a fixed random seed between epochs - we need to be able to remove it eventually

### DIFF
--- a/fe/parameterize_exp.py
+++ b/fe/parameterize_exp.py
@@ -7,6 +7,8 @@ import itertools
 import os
 import sys
 
+from timemachine.integrator import langevin_coefficients
+
 from jax.config import config as jax_config
 # this always needs to be set
 jax_config.update("jax_enable_x64", True)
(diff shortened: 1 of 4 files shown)


fe/parameterize_exp.py
```diff
@@ -67,66 +69,49 @@ def get_masses(m):
...
-    all_du_dls = jnp.array(all_du_dls)
-
-    bkwd = all_du_dls[:, T:]
-    bkwd_sched = schedule[T:]
+            print("conf_idx", conf_idx, "dG None")
 
-    dG_bkwd = math_utils.trapz(bkwd, bkwd_sched) # integral from 0 to inf
-    dG_bkwd = -dG_bkwd
+    all_du_dls = jnp.array(all_du_dls)
+    work_insertion = math_utils.trapz(all_du_dls, schedule) # integral from 0 to inf
```
chuantian, line 87: from inf to 0?

fe/parameterize_exp.py
```diff
@@ -228,6 +168,13 @@ def error_fn(all_du_dls_raw, T, schedule, true_dG):
     host_conf = np.array(host_conf)
     host_name = "complex"
 
+    conformer = guest_mol.GetConformer(0)
```
chuantian, line 171: this is the pre-supplied conformer

fe/parameterize_exp.py
```diff
@@ -282,32 +270,27 @@ def error_fn(all_du_dls_raw, T, schedule, true_dG):
...
+        for trial_idx in range(args.num_trials):
 
-            v0 = np.zeros_like(x0)
+            if epoch % 10 == 0:
+                combined_pdb = Chem.CombineMols(Chem.MolFromPDBFile(host_pdb_file, removeHs=False), guest_mol)
+                combined_pdb_str = StringIO(Chem.MolToPDBBlock(combined_pdb))
+                out_file = os.path.join(args.out_dir, "epoch_"+str(epoch)+"_insertion_"+str(trial_idx)+".pdb")
+                writer = PDBWriter(combined_pdb_str, out_file)
+            else:
+                writer = None
```
chuantian, line 286: we write PDB every 10 epochs to save storage

fe/parameterize_exp.py
proteneer: you're right - inf to zero.

PR #185, opened 2020-04-15
Thermodynamic Integration

proteneer:


chuantian:
I really like this functionality!

chuantian:
In eac990e, only non-bonded coupling/decoupling is done, right? no bond/angle/torsion

proteneer:
Correct only nonbonded is done. We need to implement energy calculations to compute du/dl of bonded terms. I'm going to merge this then starting working on RBFE this week.

PR #187, opened 2020-04-21
Limit Nonbonded Calculation

proteneer:
This allows us to limit the nonbonded calculations to that of a subset of atoms (basically masking out the non-interacting parts entirely.)

- [x] Nonbonded Calculations
- [x] GB
- [x] Tests for both


proteneer:
@chuantian This is one of the changes necessary for us to test the valence free energy code. It's an extremely easy of implementing exclusions for GB if we *never* have to move them.

PR #188, opened 2020-04-21
Minor neighborlist improvement.

proteneer:
This is a minor change to the neighborlist code. The old version of the code was still correct, just slightly less efficient. It was estimating the box size to be 2 times bigger than it actually was, thereby skipping less number of blocks than is optimal.


proteneer:
closing this - since the new neighborlist code will incorporate this.

PR #189, opened 2020-04-22
Initial commit of the bonded Alchemical merge algorithm for Chuan to …

proteneer:
…scaffold off of.


proteneer:
@PatWalters @bp-kelley if you guys have any better ideas on how the hell to do this please let me know.

proteneer:
TL;DR is that this takes two molecules with a common core, and different R-groups R_A and R_B, and figures out how build an alchemical "hybrid". Note that we can't just do rdkit.CombineMols since the resulting molecule will be un-parameterizable due to the system, so we need to parameterize, then merge and do the interpolation.

bp-kelley:
@proteneer Going to need a bit more guidance on this.  Could we iterate on a Jupyter notebook until we have something more concrete?

proteneer:
stale

PR #191, opened 2020-04-23
Full refactor of alchemical support + quasi periodic nonbonded

proteneer:
This will be a very long pull request with lots of WIPs with the ultimate goal of support alchemical RBFE on the valence terms.

**Main goals:**

1. Move the alchemical semantics from the stepper/simulation contexts to the individual force classes.
2. Deprecate the need to carry 4D semantics when doing integration.
3. Gradients now formally map from R^(3N+P+1) -> R^(3N+P+1). The implied content of the mapping is (x, p, lambda) -> (du_dx, _, du_dlambda)

The lambda idxs carry the following semantics:

***Bonded terms***

lambda_idx = 0, prefactor = 1
lambda_idx = 1, prefactor = lambda
lambda_idx =-1, prefactor = (1-lambda)

***Nonbonded terms***

lambda_idx = 0, w = 0
lambda_idx = 1, w = lambda
lambda_idx =-1, w = cutoff + lambda

Distances in the 4th dimension is always computed with periodic boundary conditions.

***Quirks***

The PBC is quasi periodic because the first three dimensions is non-periodic. We also only ever compute bonded terms using the first 3 dimensions.

The nonbonded neighborlist is done only in non-periodic 3D. It is guaranteed that that dij_3D < dij_4D, so the neighborlist will only ever underestimate the distance between two bounding boxes, never overestimate. This means that we will do slightly more work as slightly more tiles will pass the dij < cutoff test, but the actual nonbonded calculation later on will correctly compute dij_4D and bypass as needed. 

First pass:

- [x] HarmonicBond Inference
- [x] HarmonicBond JVP
- [x] Implement Alchemical Stepper
- [x] BPTT tests

Second pass:

- [x] Nonbonded Inference
- [x] Nonbonded JVP
- [x] BPTT test with bonded + nonbonded
- [x] GBSA Inference
- [x] GBSA JVP

Third pass:

- [x] HarmonicAngle Inference
- [x] HarmonicAngle JVP
- [x] PeriodicTorsion Inference
- [x] PeriodicTorsion JVP

Fourth pass:

- [x] Forcefield class fixes
- [x] Clean up steppers
- [x] Clean non-alchemical execution
- [x] Revert fixed point


proteneer:
Just doing the GBSA Inference only took all morning! JVP in the afternoon....

proteneer:
This is done!

PR #203, opened 2020-05-03
Initial bonded parameter interpolation

proteneer:
Bonded parameter interpolation is implemented as follows:

The gradients now take an additional _pi of the original parameters. For example, the harmonic bond forces now takes (k_idxs, b_idxs, k_idxs_pi, b_idxs_pi) where the _pi indicate the parameter that we interpolate into. For the force constants (eg. k_idxs) we add a special semantic to avoid having to explicitly add a zero-force constant forcefield parameter. Instead, when k_idxs[bond_idx] == -1 we simply assume that the endpoint is zero. This means we don't have to worry about computing derivatives in the backwards pass w.r.t. 0 force constants.

- [x] Bonds
- [ ] Angles
- [ ] Torsions
- [ ] Interpolater
- [ ] Test RBFE


proteneer:
closing this for now

PR #210, opened 2020-06-22
Add ability to set Morse potential alpha through cmd line.

proteneer:
Also modifies set default num gpus when debugging lambdas to 8.


proteneer:
@chuantian  I'm hot merging this - it's just an extra command line arg so we can set alpha in Morse potentials.

PR #211, opened 2020-06-22
Initial PoC of forcefield jacobians

proteneer:
cc @mattwthompson

Goals of the PR:

1. Allow the parameterization code to emit a (sys_params, vjp_fn) pair. sys_params is used by the various functional forms to compute energies, forces, etc. vjp_fn always take in adjoints with shape equivalent to sys_params, and returns an object with shape equal to that of ff_params. See tests for additional info on how this is meant to work.

2. Make it easy for classes that don't have differentiable parameters (eg. AM1BCC) to still be able to generate charges without having to manually introduce new forcefield parameters (we were doing some really wonky stuff before.) 

3. Separate out the classes into individual classes as opposed to having a single giant forcefield class try and take care of everything.

Items:
- [x] PoC done
- [x] HarmonicBond
- [x] MorseRestraint
- [x] Simple integrator test
- [x] Nonbonded
- [x] GBSA
- [x] HarmonicAngle
- [x] PeriodicTorsion
- [x] Small Molecule Forcefield Parameterization
- [x] Small Molecule Forcefield Serialization/Deserialization
- [x] Small Molecule Forcefield Props
- [x] Protein conversion
- [x] Inference code - parameterize ABFE
- [x] Revert fixed point
- [x] Cleanup


tests/test_handlers.py
```diff
@@ -0,0 +1,332 @@
...
+        assert bond_params.shape == (mol.GetNumBonds(), 2)
+
+
+        bonded_param_adjoints = np.random.randn(*bond_params.shape)
+
+        # test that we can use the adjoints
+        ff_adjoints = bond_vjp_fn(bonded_param_adjoints)[0]
+
+        # if a parameter is > 99 then its adjoint should be zero (converse isn't necessarily true since)
+        mask = np.argwhere(bond_params > 90)
```
mattwthompson, line 124: 90 or 99? Also below

tests/test_handlers.py
```diff
@@ -0,0 +1,332 @@
...
+            ['[#35X0-1:1]', 99., 999.],
+            ['[#53X0-1:1]', 99., 999.],
+        ]
+
+        smirks = [x[0] for x in patterns]
+        params = np.array([[x[1], x[2]] for x in patterns])
+
+        ljh = nonbonded.LennardJonesHandler(smirks, params)
+
+        mol = Chem.MolFromSmiles("C1CNCOC1F")
```
mattwthompson, line 301: Should this (and other similar mols in tests) have hydrogens?

tests/test_handlers.py
```diff
@@ -0,0 +1,332 @@
...
+
+        NL = mol.GetNumAtoms()
+        NP = 13
```
mattwthompson, line 304: Just for my understanding, this is arbitrarily 13, right?

ff/handlers/utils.py
```diff
@@ -0,0 +1,44 @@
+from rdkit import Chem
+
+
+def sort_tuple(arr):
```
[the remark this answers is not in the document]
proteneer: this is correct for bonds, angles, and proper torsions. We only inspect arr[0] and arr[-1]. We use a different scheme for improper torsions.

ff/handlers/utils.py
```diff
@@ -0,0 +1,44 @@
...
+def match_smirks(mol, smirks):
+    
+    # Make a copy of the molecule
+    rdmol = Chem.Mol(mol)
+    # Use designated aromaticity model
+    Chem.SanitizeMol(rdmol, Chem.SANITIZE_ALL ^ Chem.SANITIZE_SETAROMATICITY)
+    Chem.SetAromaticity(rdmol, Chem.AromaticityModel.AROMATICITY_MDL)
+    
+    # Set up query.
+    qmol = Chem.MolFromSmarts(smirks)  #cannot catch the error
```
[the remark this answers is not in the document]
proteneer: I was more concerned about SMIRKS that we write ourselves which may be malformed. Usually the end result in the event of an exception is qmol = None

ff/handlers/utils.py
```diff
@@ -0,0 +1,44 @@
...
+        smirks_index = atom.GetAtomMapNum()
+        if smirks_index != 0:
+            idx_map[smirks_index - 1] = atom.GetIdx()
+    map_list = [idx_map[x] for x in sorted(idx_map)]
+
+    # Perform matching
+    matches = list()
+    for match in rdmol.GetSubstructMatches(qmol, uniquify=False):
+        mas = [match[x] for x in map_list]
+        matches.append(tuple(mas))
```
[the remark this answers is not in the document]
proteneer: ImproperTorsions require unsorted patterns - so we don't do the sorting here.

tests/test_handlers.py
proteneer: this is just me being sloppy, it's all the same.

tests/test_handlers.py
proteneer: Apparently for testing purposes it doesn't matter. It just adds the number of atoms I need to manually check. But you're right that in production we need to do something like AddHs

tests/test_handlers.py
proteneer: my lucky number

proteneer:
Initial proof of concept implemented on the CUDA side for harmonic bond and morse potential jacobians

proteneer:
Nonbonded refactor implemented and tests fully passing. GB next!

proteneer:
GB on the CUDA side has been re-implemented and simplified.

proteneer:
It's fairly straightforward for us to chain the vjps so we can separate out the concerns.

```
import jax
import jax.numpy as np
def f(x):
    return x*x
def g(x):
    return np.exp(x)
fx, f_vjp_fn = jax.vjp(f, 1.0)
gfx, g_vjp_fn = jax.vjp(g, fx)
print(gfx)
print(f_vjp_fn(g_vjp_fn(0.3)[0]))
def gof(x):
    return g(f(x))
gfx, gof_vjp_fn = jax.vjp(gof, 1.0)
print(gfx)
print(gof_vjp_fn(0.3))
```

proteneer:
Protein conversion code is done. All the major pieces are almost ready to go - putting everything together tomorrow!

proteneer:
Serialization support is done - with the caveat of not supporting props yet.

proteneer:
Props done - time for inference testing!

proteneer:
Had to fix the ProperTorsion forcefield parameterization/serialization code due to the use of sparse matrices that would've made the vjps a massive nightmare to deal with later on.

proteneer:
This is nearly done save for the cleanup code - the training code will be done in a separate pass to enable running on 100s of GPUs.

proteneer:
This PR is finally complete. The actual re-parameterization code will come in a separate PR. This sets up the foundation for all our future efforts and greatly improves flexibility since we've now decouple forcefield parameters (P) from the system parameters (Q) and implemented the jacobian bridge via jax between them. This basically means that if we change how we generate Q downstream, we no longer need to modify or hack the C++ code, all we need to do is modify the jacobians on the python side!

proteneer:
This PR also adds many of the missing tests to the small molecule force field classes to ensure things like the vjps are operating correctly.

PR #213, opened 2020-07-09
Distributed multi-node training WIP

proteneer:
This PR will implement our final distributed training protocol, scaled up to 256 gpus inside a cloud environment.

The protocol is as follows:

1. One master node acts as AsyncClients to 256 workers. This semantically matches better with our multiprocessing pattern. The IP of the master node is fixed.
2. Upon launch, the workers broadcast their IPs to the master node via some undetermined mechanism. 
3. Once the master node detects that all the workers are online, the job is initiated. The messages are sent using the defined .proto schema. 
4. Workers can operate in either a stateless inference mode, or a stateful training mode. In the stateful mode the intermediates required for the backward pass is stored. 
5. The two points are synchronization on the master node are: computation of final deltaGs via thermodynamic integration and computation of the gradient update.

- [x] Define protobufs and core gRPC routes. We'll be using proto3 throughout.
- [x] Define server/client semantics (runners are here are ran as servers, and clients are the parameter trainers). This reduces the REQ/REP pairs from 3->2 and avoids the need to build an explicit queue. 
- [x] Lambda scheduling/job preparation
- [x] Inference Mode
- [x] Training Mode
- [x] Update requirements with frozen protobuf, grpc

Worker discovery/registration will be done separately.


proteneer:
This is the .proto file to be used:

``` protobuf
syntax = "proto3";

service Worker {
	rpc ForwardMode(ForwardRequest) returns (ForwardReply) {}
 	rpc BackwardMode(BackwardRequest) returns (BackwardReply) {}
}

// The request message containing the user's name.
message ForwardRequest {
	bool inference = 1;
	bytes system = 2; // pickle object
    string precision = 3;
}

message ForwardReply {
	bytes du_dls = 1; // numpy array
    bytes energies = 2; // numpy array
}

// The response message containing the greetings
message BackwardRequest {
	bytes adjoint_du_dls = 1; // pickle
}

message BackwardReply {
	bytes dl_dps = 1; // pickle
}
```

proteneer:
Update batch training code  to precompute systems for each lambda window, added code to do a single pass reduce of dL/dadjoints derivatives.

proteneer:
This is adequately implemented for a first pass. Merging and doing a clean up PR afterwards.

PR #216, opened 2020-07-11
gRPC based service discovery

proteneer:


training/abfe.py
```diff
@@ -87,24 +88,29 @@ def convert_uIC50_to_kJ_per_mole(amount_in_uM):
...
-        # 50009
-    ]
 
+    ip_list = []
+    expected_workers = 6
+    stop_event = threading.Event()
+
+    registry = registration.Registry(ip_list, expected_workers, stop_event)
+
+    server = grpc.server(futures.ThreadPoolExecutor(max_workers=16))
```
bp-kelley, line 98: I’m finding the combination that combination of concurrent and threading odd.  I’m assuming there is a way to use a process pool executor in case you run into GIL issues.

proteneer: This is actually the recommended pattern. The server code is only blocked by IO, not CPU/compute. 

See this thread for a more in-depth discussion:

https://github.com/grpc/grpc/issues/18535

proteneer:
After some thought this method of discovery is a little bit too unwieldy. I'm closing this in favor of another soon to come PR.

proteneer:
Re-opening this

proteneer:
closing this - let's keep the two parts decoupled for now.

PR #217, opened 2020-07-13
some minor refactoring to cleanup dependencies

proteneer:
This PR:

1. Removes worker dependencies on openmm and rdkit. 
2. Adds a config file as opposed to trying to pass everything in command line and expanding configurability.
3. Allow specification of worker hosts in config file.
4. Pulls pybind11 using https as opposed to git (remove the need to setup ssh keys explicitly).
5. lambda schedules need to be fully specified, with stage 0 being monotonically decreasing, and stage 1 and 2 being monotonically increasing.


proteneer:
This is ready to be merged. Multiplexing will be done next.

PR #219, opened 2020-07-15
add handler for bcc using Simon's parameters

chuantian:


tests/test_handlers.py
```diff
@@ -243,6 +243,42 @@ def test_am1_bcc():
 
     assert vjp_fn(charges_adjoints) == None
 
+def test_am1_ccc():
+    
+    patterns = [
```
proteneer, line 248: I'd expand the test on a slightly more set of parameters and molecule.

tests/test_handlers.py
```diff
@@ -243,6 +243,42 @@ def test_am1_bcc():
...
+    mol = Chem.AddHs(Chem.MolFromSmiles("CC#C"))
+    AllChem.EmbedMolecule(mol)
+    es_params, es_vjp_fn = am1h.parameterize(mol)
+
+    ligand_params = np.array([
+        -1.65,
+        -1.65,
+        -2.61,
+        1.10,
+        1.10,
```
proteneer, line 267: AM1 is still symmetric?

ff/handlers/nonbonded.py
```diff
@@ -184,3 +190,102 @@ def vjp_fn(*args, **kwargs):
...
+
+        if result is False:
+            raise Exception('Unable to assign charges')
+
+        am1_charges = [] 
+        for index, atom in enumerate(oemol.GetAtoms()):
+            q = atom.GetPartialCharge()*np.sqrt(constants.ONE_4PI_EPS0)
+            am1_charges.append(q)
+
+        bond_idxs = []
```
proteneer, line 244: are bonds lexicographically unique? i.e. if a bond (a_idx, b_idx) is in bond_idxs, will you ever have a situation where (b_idx, a_idx) is also in bond_idxs? should that be allowed?

tests/test_handlers.py
chuantian: they are not exactly the same, only up to 2 decimal places

tests/test_handlers.py
chuantian: I could use the same molecule as in the rest of the tests.

ff/handlers/nonbonded.py
proteneer: to follow up on this, can we add an assertion at the end with:

```
unique_bond_idxs = set()
for src_idx, dst_idx in bond_idxs:
     src_idx, dst_idx = sort((src_idx, dst_idx))
     unique_bond_idxs.add((src_idx, dst_idx))
assert len(unique_bond_idxs) == len(bond_idxs)
```

proteneer: disregard this - this isn't needed after discussing offline.

proteneer:
ps - don't forget to add a serialization test to this - it's straightforward and you can follow the examples in test_serialization.py

chuantian:
This can be closed, since the #228 will do the same but with updated master branch

chuantian:
close this

PR #229, opened 2020-07-18
Enable multiplexing in inference mode.

proteneer:
This PR enables to multiplex on inference mode since the workers become stateless.


proteneer:
merging this in so i can test ccc on cluster

PR #232, opened 2020-07-20
stage_dGs

proteneer:
Print the stage dGs during training.


proteneer:
Merging this in then working on the adam optimizer.

PR #235, opened 2020-08-03
Add Jenkinsfile

tony-garcia:
Test PR for CI pipeline

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,11 @@
+pipeline {
+  agent any
+  stages {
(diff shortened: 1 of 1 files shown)


bp-kelley:
@proteneer I think we can add some azure gpu CI, is that something I should look into?

proteneer:
open to it but last time i checked azure doesn't have the same gpus that we use on the  cloud (T4) - can azure spin up/down instances automatically?

bp-kelley:
It looks like they have p40s and m60s as well as the v100s

PR #247, opened 2020-08-10
Scaffolding for Harmonic Centroid Restraints.

proteneer:
This is an implementation of harmonic centroid restraints nearly identical to the RadiallySymmetric restraints in Yank. 

- [x] Forward mode implementation (slow)
- [x] Backward mode implementation (also slow)
- [x] Stage 0 - Attach restraint tests (Alchemical, lambda 0->1)
- [x] Stage 1 - Decoupling tests (Geometric, lambda 0nm->1.5nm)
- [x] Stage 2 - Detach restraint tests (Analytical)
- [x] vjp tests need to be restored
- [x] test restraint code.


proteneer:
Unfortunately the converge with purely CoM restraints is quite poor. I will work on implementing the somewhat standard BoreschLike restraints at this point, so at least we have a point of reference.

proteneer:
Boresch restraints and the haversine based derivatives implemented on the GPU - took a good hour or so to get tests passing.

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -316,11 +317,104 @@ void declare_restraint(py::module &m, const char *typestr) {
...
+    .def(py::init([](
+        const py::array_t<int, py::array::c_style> &group_a_idxs,
+        const py::array_t<int, py::array::c_style> &group_b_idxs,
+        const py::array_t<double, py::array::c_style> &masses,
+        double kb,
+        double b0,
+        int lambda_flag,
+        int lambda_offset
+    ){
+        std::vector<int> vec_group_a_idxs(group_a_idxs.size());
```
bp-kelley, line 345: it's hard to beat 
```
std::vector<int> vec_group_a_idxs = group_a_idxs;
```

bp-kelley: I'm also confused why these need to be copied at all, CentroidRestraint takes const refs

bp-kelley: Ah, the inputs aren't vectors here, they are 
```
const py::array_t<int, py::array::c_style>
```

bp-kelley: You can do tree the numpy array as a pointer and do:

```
std::vector<int> vec_group_a_idxs(group_a_idxs.data(),
                                                       group_a_idxs.data() + group_a_idx.data().size());
```

proteneer: Yeah I think later on we definitely need to clean this up. I was thinking of adding Eigen as dependency into the C++ later (which pybind11 has many excellent adapters for). Though I do like the alternate constructor to avoid having to call memcpy explicitly.

proteneer:
Looks like I will need to defer the Boresch stuff later... yet again. There are some extremely weird instabilities unless we're very careful about how we define the 6 anchor atoms. I will revisit this later. CoM restraints also seem okay-ish if we can increase the number of lambda windows dramatically. We're now at around a ~1kcal/mol fluctuation under different thermal seeds.

PR #251, opened 2020-08-13
Add some initial code for pose docking.

proteneer:
@schmolly


proteneer:
@schmolly I'm going to force merge this since it's just an initial prototype.

PR #252, opened 2020-08-13
Multiplexed backwards mode done

proteneer:
This PR:

- Stores multiple states on the workers
- Guards the self.states dictionary with mutexes to prevent most(?) type of race conditions
- Modifies the client so that backward mode jobs are distributed evenly
- We *do not* check to see if the GPUs have sufficient memory for multiple jobs. This is up to the end-user.

### DIFF
--- a/training/worker.py
+++ b/training/worker.py
@@ -13,21 +13,26 @@ import grpc
 import service_pb2
 import service_pb2_grpc
 
+from threading import Lock
 
 from timemachine.lib import custom_ops, ops
 
 class Worker(service_pb2_grpc.WorkerServicer):
 
     def __init__(self):
-        self.state = None
+        self.states = {}
+        self.mutex = Lock()
 
     def ResetState(self, request, context):
-        self.state = None
+        self.mutex.acquire()
+        self.states = {}
+        self.mutex.release()
         reply = service_pb2.EmptyMessage()
         return reply
 
     def ForwardMode(self, request, context):
-        assert self.state is None
+        # assert self.state is None
+        assert request.key not in self.states
 
         if request.precision == 'single':
             precision = np.float32
@@ -66,8 +71,10 @@ class Worker(service_pb2_grpc.WorkerServicer):
         )
 
         start = time.time()
-        ctxt.forward_mode()
 
+        # ensure only one GPU can be running at given time.
+        self.mutex.acquire()
+        ctxt.forward_mode()
         full_du_dls = stepper.get_du_dl() # [FxT]
         energies = stepper.get_energies()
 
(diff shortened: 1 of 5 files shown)


training/worker.py
```diff
@@ -13,21 +13,26 @@
...
 
     def __init__(self):
-        self.state = None
+        self.states = {}
+        self.mutex = Lock()
 
     def ResetState(self, request, context):
-        self.state = None
+        self.mutex.acquire()
+        self.states = {}
```
bp-kelley, line 28: self.states.clear() 

Might be better here

training/worker.py
```diff
@@ -13,21 +13,26 @@
...
+        self.mutex.release()
         reply = service_pb2.EmptyMessage()
         return reply
 
     def ForwardMode(self, request, context):
-        assert self.state is None
+        # assert self.state is None
+        assert request.key not in self.states
```
bp-kelley, line 35: This should be inside the mutex

training/worker.py
```diff
@@ -66,8 +71,10 @@ def ForwardMode(self, request, context):
         )
 
         start = time.time()
-        ctxt.forward_mode()
 
+        # ensure only one GPU can be running at given time.
+        self.mutex.acquire()
```
bp-kelley, line 76: Prob doesn’t matter here but

try:
  mutex.acquire()
finally:
  mutex.release()

Will prevent errors holding locks

training/worker.py
proteneer: good call

training/worker.py
proteneer: I think i'd want the error to be raised and propagated back to the host

training/worker.py
proteneer: yeah i think it's a little clearer - not sure GCing the old object works when the ref exits, but probably have a single object is better.

training/worker.py
proteneer: I'm actually going to ditch this assertion entirely

proteneer:
We need to modify the learning rate schedule for LJ so that it's based on a trust radius.

training/worker.py
bp-kelley: It will be, The finally block just ensures the mutex is released

Whatever is in “finally” is always run whether an exception is thrown or not.

You could also use a “with mutex” lock idiom as well.

proteneer: Ah I see what you mean: you're concerned about code in between the acquire() and the release() throwing an exception deadlocking the code. Yeah it's a definite possibility. I'll change it to a context manager.

proteneer:
@bp-kelley thanks for the comments. Merging this once tests pass.

PR #259, opened 2020-08-21
[WIP] Differentiable minimization

proteneer:
For some ligands, the starting pose will likely clash. So we need to do a minimization protocol (that's fully differentiable at the start of the simulation).

- [ ] Inference Mode
- [ ] Backwards Mode


proteneer:
It's unclear how to do this with the thermodynamic derivative. Closing.

PR #273, opened 2020-10-01
Exclusion list changes

proteneer:
This PR uses a trick to improve the way we compute exclusions. Previously blocks that contained exclusions will exhibit a significant loss of accuracy due to the accumulation of floating point numbers.

1. Given 4 floating point numbers, where a,c,d, are well-behaved, and b is an exclusion term, we compute (a+b+c+d)-b
2. (a+b+c+d) is casted to fixed point via unsigned long longs (ull), as well as -b
3. However, this method will result in massive loss of precision for most 1-2 exclusions and many 1-3 exclusions since calculations inside the parentheses are still accumulated using regular floating point arithmetic.
4. Instead we do every accumulation, even that of inside the parentheses, in fixed point. Meaning, we compute (a_ull + b_ull + c_ull + d_ull) - b_ull
5. ulls formally are a closed group under modular arithmetic, as any overflow will "circle back". This means we can perfectly cancel out large exclusion terms without fundamentally affecting the accuracy of the remaining operations.

This is very different from how OpenMM and AMBER handles exclusion lists, as they typically require construction of specialized neighborlists and nonbonded kernels that can handle tiles with exclusions. I'm not sure which approach is superior in the long run, but this implementation was hacked together in about 4 hours. Note that the PR still needs some cleanup and optimizations.

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -131,18 +131,13 @@ def prepare_nb_system(
 
     atom_idxs = np.arange(N)
 
-    if E == 1:
-        exclusion_idxs = np.array([[0, 1]], dtype=np.int32)
-    # exclusion_idxs = np.random.choice(atom_idxs, size=(E, 2), replace=False)
-    # exclusion_idxs = np.array(exclusion_idxs, dtype=np.int32).reshape(-1, 2)
+    exclusion_idxs = np.random.choice(atom_idxs, size=(E, 2), replace=False)
+    exclusion_idxs = np.array(exclusion_idxs, dtype=np.int32).reshape(-1, 2)
 
-
-        scales = np.array([[1, 1]], dtype=np.int32)
-
-        # scales = np.stack([
-        #     np.random.rand(E),
-        #     np.random.rand(E)
-        # ], axis=1)
+    scales = np.stack([
+            np.random.rand(E),
+            np.random.rand(E)
+    ], axis=1)
 
     beta = 2.0
 
--- a/tests/common.py
+++ b/tests/common.py
@@ -206,8 +206,8 @@ def prepare_nb_system(
     exclusion_idxs = np.array(exclusion_idxs, dtype=np.int32).reshape(-1, 2)
 
     scales = np.stack([
-            np.random.rand(E),
-            np.random.rand(E)
+        np.random.rand(E),
+        np.random.rand(E)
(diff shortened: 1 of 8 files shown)


bp-kelley:
Not sure it matters, but if you add two NaNs this way they end up equalling 0 not NaN so this will only cancel out one NaN I think.

```
#include <iostream>
#include <cmath>

#define FIXED_EXPONENT             0x100000000


unsigned long long FLOAT_TO_FIXED(double v) {
     return static_cast<unsigned long long>((long long)(v*FIXED_EXPONENT));
 }

double FIXED_TO_FLOAT(unsigned long long v) {
     return static_cast<double>(static_cast<long long>(v))/FIXED_EXPONENT;
}

void test(double a, double b) {
  unsigned long long ca = FLOAT_TO_FIXED(a);
  unsigned long long cb = FLOAT_TO_FIXED(b);

  std::cout << a+b << " " << FIXED_TO_FLOAT(ca+cb) << std::endl;
}


int main(int, char**) {
  double a = 3.4;
  double b = 2.234234;
  test(a,b);

  test(NAN, NAN);
}
```

proteneer:
This is the behavior I'm testing for:


``` cpp
#include <iostream>
#include <cmath>
#define FIXED_EXPONENT             0x100000000

unsigned long long FLOAT_TO_FIXED(double v) {
     return static_cast<unsigned long long>((long long)(v*FIXED_EXPONENT));
 }
double FIXED_TO_FLOAT(unsigned long long v) {
     return static_cast<double>(static_cast<long long>(v))/FIXED_EXPONENT;
}
void test(double a, double b, double c, double d) {
  unsigned long long ca = FLOAT_TO_FIXED(a);
  unsigned long long cb = FLOAT_TO_FIXED(b);
  unsigned long long cc = FLOAT_TO_FIXED(c);
  unsigned long long cd = FLOAT_TO_FIXED(d);

  std::cout << a+b+c+d-b << " " << FIXED_TO_FLOAT(ca+cb+cc+cd-cb) << std::endl;

}

int main(int, char**) {
  double a = 3.4;
  double b = 0;
  double c = 1.2347;
  double d = -1.234;
  test(a,b,c,d);
  test(a,NAN,c,d);
}
```

proteneer:
Merging this.

PR #275, opened 2020-10-02
Restore CoM restraints

proteneer:
This restores functionality for center of mass restraints, but removing the lambda parameter. The alchemical lambda parameter has been removed as it will be done property later on through an AlchemicalGradient class.


proteneer:
Merging this so we can try out separated topologies

PR #283, opened 2020-10-17
clean up pose_dock

schmolly:
Changes to dock_setup.py were mostly formatting (I used the Black python formatter), but I also add the ability to hold selected atoms constant by increasing their mass (by a lot).

Most of the changes to pose_dock.py were due to adding and accommodating different options.

### DIFF
--- a/docking/dock_setup.py
+++ b/docking/dock_setup.py
@@ -14,9 +14,7 @@ def combine_parameters(guest_q, guest_lj, host_qlj):
     return np.concatenate([host_qlj, guest_qlj])
 
 
-def combine_potentials(
-    guest_ff_handlers, guest_mol, host_system, precision, constant_atoms=[]
-):
+def combine_potentials(guest_ff_handlers, guest_mol, host_system, precision):
     """
     This function is responsible for figuring out how to take two separate hamiltonians
     and combining them into one sensible alchemical system.
@@ -47,9 +45,6 @@ def combine_potentials(
     host_potentials, host_masses = openmm_deserializer.deserialize_system(
         host_system, precision, cutoff=1.0
     )
-    for atom_num in constant_atoms:
-        host_masses[atom_num - 1] += 50000
-
     host_nb_bp = None
 
     combined_potentials = []
--- a/docking/dock_setup.py
+++ b/docking/dock_setup.py
@@ -6,19 +6,17 @@ from ff.handlers import bonded, nonbonded, openmm_deserializer
 
 from timemachine.lib import potentials
 
-def combine_parameters(guest_q, guest_lj, host_qlj):
-    guest_qlj = np.concatenate([
-        np.reshape(guest_q, (-1, 1)),
-        np.reshape(guest_lj, (-1, 2))
-    ], axis=1)
 
+def combine_parameters(guest_q, guest_lj, host_qlj):
+    guest_qlj = np.concatenate(
+        [np.reshape(guest_q, (-1, 1)), np.reshape(guest_lj, (-1, 2))], axis=1
+    )
     return np.concatenate([host_qlj, guest_qlj])
 
+
 def combine_potentials(
-    ff_handlers,
-    guest_mol,
-    host_system,
-    precision):
+    guest_ff_handlers, guest_mol, host_system, precision, constant_atoms=[]
+):
     """
     This function is responsible for figuring out how to take two separate hamiltonians
     and combining them into one sensible alchemical system.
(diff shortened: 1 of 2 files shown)


docking/dock_setup.py
```diff
@@ -47,15 +45,14 @@ def combine_potentials(
     """
 
     host_potentials, host_masses = openmm_deserializer.deserialize_system(
-        host_system,
-        precision,
-        cutoff=1.0
+        host_system, precision, cutoff=1.0
     )
+    for atom_num in constant_atoms:
```
proteneer, line 50: Two notes:

1. I'm going to eventually refactor this function soonish so there will be a single combine_potentials function used across multiple workflows (abfe/rbfe) - so it may make sense to do the constant atoms juggling outside of this function.

2. Are idxs in constant_atoms 1-idxed for the protein atoms?

docking/pose_dock.py
```diff
@@ -15,130 +16,183 @@
...
+    """Poses guests into a host by running short simulations in which the guests phase in over time
+
+    Parameters
+    ----------
+
+    guests_sdfile: path to input sdf with guests to pose/dock
+    host_pdbfile: path to host pdb file to dock into
+    outdir: where to write output (will be created if it does not already exist)
+    n_steps: how many total steps of simulation to do (recommended: 20000)
+    lowering_steps: how many steps to lower the guest over (recommended: 10000)
```
proteneer, line 41: 10000 might be a bit too aggressive for a LangevinIntegrator with a default friction of 1. I was seeing the ligand insert itself into all kinds of weird geometries - but I think this is probably target/ligand specific.

docking/pose_dock.py
```diff
@@ -15,130 +16,183 @@
...
+    start_percent: how visible the guest should start out (recommended: 0.75)
+    random_rotation: whether to apply a random rotation to each guest before beginning the simulation
+    constant_atoms: atom numbers from the host_pdbfile to hold mostly fixed across the simulation
```
proteneer, line 44: May want to note that this is 1-idxed?

docking/pose_dock.py
```diff
@@ -15,130 +16,183 @@
...
+            new_lambda_schedule
+        ):
+            ctxt.step(lamb)
+            if step % 1000 == 0:
+                print("step", step, "lamb", lamb, "nrg", ctxt.get_u_t())
+                combined_pdb_str = StringIO(Chem.MolToPDBBlock(combined_pdb))
+                pdb_writer = PDBWriter(
+                    combined_pdb_str,
+                    os.path.join(
+                        outdir, f"{guest_name}_{str(step).zfill(len(str(n_steps)))}.pdb"
```
proteneer, line 132: I had no idea zfill existed

docking/pose_dock.py
```diff
@@ -15,130 +16,183 @@
...
+    )
+    parser.add_argument("--nsteps", type=int, default=20000, help="simulation length (1 step = 1.5 femtoseconds")
+    parser.add_argument(
+        "--lowering_steps",
+        type=int,
+        default=10000,
+        help="how many steps to take while phasing in the guest",
+    )
+    parser.add_argument(
+        "--start_percent",
```
proteneer, line 165: So the start percentage isn't exactly a good measure here. The lambda idx really corresponds to the 4D plane on which the ligand resides in. Anything more than 3x then length of the vdW radii will result result in a close to a 0% visibility. It's better to just be explicit about and call this --start_lambda and simplify the construction of the lambda schedule

```
        new_lambda_schedule = np.concatenate(
            [
                np.linspace(args.start_lambda, 0.0, lowering_steps),
                np.zeros(n_steps - lowering_steps),
            ]
        )
```

docking/pose_dock.py
```diff
@@ -15,130 +16,183 @@
...
+        box = np.eye(3) * 100
+        v0 = np.zeros_like(x0)
+
+        impls = []
+        for b in bps:
+            p_impl = b.bound_impl()
+            impls.append(p_impl)
+
+        ctxt = custom_ops.Context(x0, v0, box, intg, impls)
+
```
proteneer, line 114: Should probably add a sanity assertion that n_steps >= lower_steps, otherwise np.zeros(-1) will throw a cryptic error message:

```
>>> np.zeros(-1)
  File "<stdin>", line 1, in <module>
ValueError: negative dimensions are not allowed
```

proteneer (commented):
This looks great :) some minor details but overall a fantastic set of additions.

docking/dock_setup.py
schmolly: 1. moved this back to pose_dock.py
2. Yes, because PDB files start at ATOM 1

docking/pose_dock.py
schmolly: I'd love for you to explain this to me in more detail in our next meeting :)

proteneer:
LGTM - feel free to squash merge when you're ready.

proteneer (approved):

PR #285, opened 2020-10-30
Allow plane_idxs to be set in order to make ligand non-interacting.

proteneer:
This PR modifies nonbonded interaction to allow setting atom i's 4th dimensional coordinate to cutoff*lambda_plane_idxs[i]

Before, the 4th dimension is computed as:

``` python
d4 = lambda_offset_idxs[i]*lamb
```
It is now modified to:

```
d4 = cutoff*lambda_plane_idxs[i] + lambda_offset_idxs[i]*lamb
```
pose_dock.py has also been modified - though the behavior is identical to that of before.

The use case for this is being able to be able to implement an alchemical restraint stage where a ligand with K atoms that is non-interacting with the rest of the system *without* having to introduce exclusions via K against (N-K) pairlist.

### DIFF
--- a/docking/dock_setup.py
+++ b/docking/dock_setup.py
@@ -131,10 +131,15 @@ def combine_potentials(guest_ff_handlers, guest_mol, host_system, precision):
 
     # allow the ligand to be alchemically decoupled
     # a value of one indicates that we allow the atom to be adjusted by the lambda value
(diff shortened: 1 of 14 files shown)


proteneer:
@schmolly I don't think this breaks your code - I made changes and tested it with default settings and things seem to be fine. We should consider putting the pose_dock stuff under testing in the near future.

proteneer:
tests passing - squash merging.

PR #289, opened 2020-11-02
Add the ability to grab full du_dl

proteneer:
@schmolly we should discuss this offline as well.


proteneer:
Closing this since @schmolly will open a new one.

PR #290, opened 2020-11-02
Reference shape force implementation.

proteneer:
This PR is a WIP of a gaussian shape based force implementation. It is not designed for batch operations right now, but rather aim is to use it inside an MD inner loop. Some notes:

1. The jax implementation is intentionally slow and does a full NxN calculation with a lots of block diagonal waste.  
2. The overlap volume is strictly between 0 and 1 following the approach in Grant's 1996 paper. The potential energy is defined as 1/volume.
3. The p,a parameters will be respectively converted from the eps and sig parameters of the off LennardJones directly. I'm not sure how this conversion will go yet, but probably something like radii ~= sig^(1/3) using the highly sophisticated technique of eyeballing. The eps to p conversion is a direct 1 to 1 since they're both pre-factors.


proteneer:
Turns out U=1/V is a terrible idea, U=k*(V-1)**2 is much much better. Long live harmonic oscillators!

proteneer:
MERGED!

PR #293, opened 2020-11-11
Inertial restraints and its derivative

proteneer:
Implements orientational restraints based on alignment of the inertial tensor.


proteneer:
Initial C++ (non-CUDA) version implemented - this was insanity. CUDA version next.

PR #294, opened 2020-11-12
Full du dl

schmolly:
Yutong's code LGTM -- calculation of the du_dl was already happening, so these changes just allow us to observe those values and use the trapezoidal rule to calculate the work.

I added the options to do deletions and skip errors (useful when processing many guests), and did some general cleanup.


docking/pose_dock.py
```diff
@@ -1,57 +1,67 @@
...
 
-    A pdb file every 1000 steps (outdir/<guest_name>_<step>.pdb)
-    stdout every 1000 steps noting the step number, lambda value, and energy
+    A pdb file every 100 steps (outdir/<guest_name>_<step>.pdb)
+    stdout every 100 steps noting the step number, lambda value, and energy
+    stdout for each guest noting the work of transition
+
+    Note
+    ----
+    If the energy becomes positive, the simulation for that guest will stop and the work will not be
```
proteneer, line 58: I'm not sure if positive energy is the best way to detect im(ex)ploded systems

docking/pose_dock.py
```diff
@@ -80,9 +89,20 @@ def pose_dock(
...
-        bps, masses = dock_setup.combine_potentials(
-            guest_ff_handlers, guest_mol, host_system
-        )
+        if skip_errors:
+            try:
+                bps, masses = dock_setup.combine_potentials(
+                    guest_ff_handlers, guest_mol, host_system
+                )
+            except Exception as err:
+                print(f"Error: There was a problem setting up {guest_name}")
```
proteneer, line 98: Curious to know what were some of the errors you've encountered?

docking/pose_dock.py
```diff
@@ -119,27 +140,75 @@ def pose_dock(
...
+                        f"{guest_name}_{str(step).zfill(len(str(n_steps)))}_guest.sdf",
+                    )
                 )
-                pdb_writer.write_header()
-                pdb_writer.write(ctxt.get_x_t() * 10)
-                pdb_writer.close()
+                writer.write(guest_mol, conf_id)
+                writer.close()
+
+        if calc_work:
```
proteneer, line 209: we should add a sanity check that endpoints in du_dl_obs.full_du_dl() is approximately 0 (eg. absolute value of 1e-3 or less)

proteneer (changes requested):
Overall this looks good - looks like you're relying on solvation etc. to the end-user.

docking/pose_dock.py
proteneer: the total energy is a usually system-dependent number - better way would be to check that the norm of force per atom is no more than say 10000kJ/(mol*nm)

docking/pose_dock.py
schmolly: Usually it was when I had problems with my input guest structures -- missing hydrogens or atoms too close together.

proteneer: Ah okay - would be nice to have this be more informative but we can add more diagnostic information later.

proteneer:
@schmolly Did you want to make anymore changes? You may need to do a squash rebase soon as master is quickly moving ahead.

docking/pose_dock.py
schmolly: done

docking/pose_dock.py
schmolly: done

docking/pose_dock.py
```diff
@@ -119,27 +139,86 @@ def pose_dock(
...
+            if step % 100 == 0 or step == len(new_lambda_schedule) - 1:
+                print(
+                    f"guest_name: {guest_name}\t"
+                    f"step: {str(step).zfill(len(str(n_steps)))}\t"
+                    f"lambda: {lamb:.2f}\t"
+                    f"energy: {ctxt.get_u_t():.2f}"
+                )
+                forces = ctxt.get_du_dx_t()
+                norm_forces = np.linalg.norm(forces, axis=-1)
+                if any(x > 10000 for x in norm_forces):
```
proteneer, line 175: if np.any(norm_forces > 10000) would also work

docking/pose_dock.py
```diff
@@ -119,27 +139,86 @@ def pose_dock(
...
+                    os.path.join(
+                        outdir,
+                        f"{guest_name}_{str(step).zfill(len(str(n_steps)))}_guest.sdf",
+                    )
+                )
+                writer.write(guest_mol, conf_id)
+                writer.close()
+
+        if (
+            abs(du_dl_obs.full_du_dl()[0] - 0) > 0.001
```
proteneer, line 211: nothing quite like subtracting a zero

proteneer (approved):
lgtm - think you can do a squash rebase merge? (if not I can show you how at a later point and you can just squash merge it this time around)

docking/pose_dock.py
schmolly: whoopslol :rofl:

schmolly:
I'll try it!

PR #299, opened 2020-11-19
Misc. minor adjustments in tests/common, tests/test_bonded, and tests/test_shape

maxentile:
* PyCharm auto-formatting
* replace a float literal with a reference to constants.py
* extract "relative_tolerance_at_precision" specification
* add ~2-line descriptions of tests
* move some test parameters into function signature, for later randomization
* fix an "original np" vs. "jax np" import error in test_shape.py (question: should we standardize these throughout to onp and jnp? or onp and np? or np and jnp?)
* add some type annotations where clear (question: what is the appropriate type annotation for ROMol?)


tests/test_bonded.py
```diff
@@ -1,31 +1,37 @@
...
-        N = 10
+class TestBonded(GradientTest):
 
-        for precision, rtol in [(np.float64, 1e-9), (np.float32, 2e-5)]:
+    def test_centroid_restraint(self, n_particles=10, n_A=4, n_B=3):
+        """Randomly define subsets A and B of a larger collection of particles,
+        generate a centroid restraint between A and B, and then validate the resulting CentroidRestraint force"""
 
-            x_primal = self.get_random_coords(N, 3)
+        kb = 5.4  # TODO: is this constant important?
```
proteneer, line 24: Not important, might as well as be a random number.

tests/test_bonded.py
```diff
@@ -1,31 +1,37 @@
...
+from jax.config import config
+
+config.update("jax_enable_x64", True)
 import functools
 
 from common import GradientTest
 from timemachine.lib import potentials
 from timemachine.potentials import bonded
 
+relative_tolerance_at_precision = {
```
proteneer, line 12: In general, at least for nonbonded forces, we probably can't do this. Different forces can achieve different levels of precision depending on the dynamic range. We generally want it to be as tight as possible for a particular force.

tests/test_bonded.py
```diff
@@ -62,34 +66,30 @@ def test_centroid_restraint(self):
                 precision=precision
             )
 
+    def test_harmonic_bond(self, n_particles=64, n_bonds=35, dim=3):
+        """Randomly connect pairs of particles, then validate the resulting HarmonicBond force"""
+        np.random.seed(125)  # TODO: where should this seed be set?
```
proteneer, line 71: Seeds are a difficult topic, setting them to a fixed number has the advantage of making it easy to debug (eg. when tests fail) as it makes everything reproducible, but has the downside of making the method not truly random. Maybe a hybrid approach where we roll a "global" seed (eg. 2020) once a year is the way to go. The annoyances of false positives in testing (failed test that failed because of a random seed in a different part of the code) can't be understated.

tests/test_bonded.py
proteneer: but just remember to seed a fixed seed and test a couple of combinations

proteneer (changes requested):

tests/test_bonded.py
maxentile: Hmm, I think we would want to (1) set a random seed so that the execution of the entire test program is deterministic, (2) reduce any sensitivity to the choice of random seed.

To achieve (2), I think we should be running multiple randomized tests rather than a single randomized test, if cheap enough. (The unit tests here are cheap enough to run many times, but some of the integration tests are not.)

To achieve (1), setting the random seed once at the top of the file is sufficient, but may not be convenient (e.g. if you want to reproduce a specific test failure in a file containing many tests). To make it easier to reproduce behavior of specific tests that depend on random state, we can modify the test function signature to accept a seed or random state as an argument.

tests/test_bonded.py
maxentile: That makes sense -- so if we were to have a single definition somewhere, rtol would depend on both `force_name` and `fp_precision`, not just `fp_precision` as I had encoded here

tests/test_bonded.py
maxentile: Thanks! I will move this into the test argument (with 5.4 as default value), for later randomization

tests/test_bonded.py
proteneer: How does pytest work with test_xxx functions that have different default kwargs? Are you able to pass in different values into pytest? (Or is the intention for this to be called from another test_ function? since pytest will automatically run any functions that begin with test_)

maxentile: Yep! PyTest provides a nice way to "parametrize" a test, as in this example https://docs.pytest.org/en/stable/parametrize.html

```python
import pytest


@pytest.mark.parametrize("test_input,expected", [("3+5", 8), ("2+4", 6), ("6*9", 42)])
def test_eval(test_input, expected):
    assert eval(test_input) == expected
```

Initially, my intention would be to do something similar to this example, but using a vector of generated random seeds rather than a vector of hand-written inputs. (That way each individual test "knows" what random state it used, and the whole collection of generated tests should run deterministically.) It may be a good idea for me to take a quick look for other examples of modern pythonic randomized tests.

proteneer: Okay that makes sense, feel free to re-request a review when you think its ready for a second pass (I'll probably just approve it without comments if the changes are mainly clerical)

tests/test_bonded.py
proteneer: Wouldn't it be better then to just have the various precision/tolerances local to the function itself (as how it used to be) as opposed to having to scroll up and down to see what they are?

maxentile: I'll revert this change -- I had noticed that the same (precision, rtol) pairs were used for all forces I had looked at in this file, but since they will in general be different across files and force types, there's not much sense in extracting the definition.

tests/test_bonded.py
maxentile: Thanks! I think the tricky issue of test randomization would better addressed in a dedicated subsequent PR -- wasn't intending to make substantive changes to the testing logic in the present PR.

tests/test_bonded.py
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/299/commits/72decf1a0194c811d528bb1f7caf1f00fdda5db6

proteneer (approved):
lgtm.

excellent changes to improve code health.

proteneer:
(feel free to squash merge when tests pass)

maxentile:
Thank you for the reviews and discussion!

PR #301, opened 2020-11-30
[WIP] Reference Rigid Shape Restraints

proteneer:
This pull request implements rigid orientational restraints using the SHAPE overlap as opposed to using principal moments of inertia. PMI-based approaches have unstable eigenvector derivatives that naturally arise throughout the simulation due to symmetries in the inertia tensor (which result in degenerate eigenvalues).

The general protocol is as follows:

1. We compute a rotational quaternion q_opt(x_a, x_b) that computes the maximum overlap between two molecules x_a and x_b.
2. The rotational quaternion is typically solved using an iterative method. In order to satisfy the rotational requirement, the norm of the quaternion must be 1. This is done with a re-parameterization of q(a,b,c)=(1-a^2-b^2-c^2, a, b, c).
3. We iterate over multiple rounds of BFGS-B by optimizing the enclosing box bounds under a sphere in order to satisfy the unit norm requirement.
4. The initial quaternion is {1,0,0,0}, and is intentionally chosen to be identical to that of the reference quaternion used in the definition of the loss function.
4. Unlike other popular packages, we actually check and enforce that dOverlap/dq has a max norm of 1e-6 at q=q_opt. Most constrained optimizers (eg: SLSQP, or other renormalization techniques) do not actually generate zero norms in the derivative.
5. A zero norm derivative enables us to do an extremely cool trick: we can use the implicit function theorem to solve for dq_opt/dx_{a,b} without having to backpropagate through BFGS. dq/dx_a is just -M.H^{-1} evaluated at the optimum. This enables us to generate forces.
6. The analytical solution is done by inverting a 4x4 nearly PD hessian.

We can use this to alchemically switch between rigid shape restraints to a non-rigid shape restraint with probably maximum overlap in phase space.

### DIFF
--- a/examples/overlap_test.py
+++ b/examples/overlap_test.py
@@ -300,31 +300,40 @@ def convergence(args):
     # re-seed since forking 
     onp.random.seed(int.from_bytes(os.urandom(4), byteorder='little'))
 
-    # this cannot be jit'd right now
-    rigid_grad_fn = jax.grad(rigid_restraint_fn)
+    rigid_restraint_fn = rigid_restraint_fn # does nothing, kept here for clarity
+    rigid_restraint_grad_fn = jax.grad(rigid_restraint_fn)
+
+    shape_restraint_fn = jax.jit(shape_restraint_fn)
+    shape_restraint_grad_fn = jax.jit(jax.grad(shape_restraint_fn))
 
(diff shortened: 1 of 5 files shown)


proteneer:
I'm going to close this PR so it stops triggering tests.

PR #303, opened 2020-12-04
Show computation of du_dp in AHFE

proteneer:
Same deal as last time


proteneer:
I'm going to close this PR. I'm reworking the vjp system so this will be outdated.

PR #304, opened 2020-12-08
Linearly rescale lambda between 0 and 1 in nonbonded ixns

proteneer:
This PR modifies the w (4d) coordinate in nonbonded calculations from:
```
w = lambda
```
to
```
w = cutoff*lambda
```
This is to enable a uniform treatment of lambda with alchemical terms (eg. restraint attachment etc.) as PR will now ensure lambda to be between [0, 1). This was a point of confusion for some people, and this should make the bounds less confusing and more consistent.

@schmolly This change will affect you - existing code should divide old lambda values by cutoff.

### DIFF
--- a/tests/test_md.py
+++ b/tests/test_md.py
@@ -46,7 +46,7 @@ class TestContext(unittest.TestCase):
             lambda_offset_idxs,
             p_scale=3.0,
             # cutoff=0.5,
-            cutoff=100.0
+            cutoff=1.5
         )
 
         masses = np.random.rand(N)
--- a/timemachine/potentials/jax_utils.py
+++ b/timemachine/potentials/jax_utils.py
@@ -1,13 +1,15 @@
 import jax.numpy as np
 
-
-# def lambda_to_w(lamb, plane_idxs, offset_idxs, cutoff):
-#     d4 = cutoff*plane_idxs + offset_idxs*lamb
-#     # d4 = offset_idxs*lamb
-#     return d4
-
 def convert_to_4d(x3, lamb, lambda_plane_idxs, lambda_offset_idxs, cutoff):
-    # d4 = lambda_to_w(lamb, lambda_plane_idxs, lambda_offset_idxs, cutoff)
+
+    # (ytz): this initializes the 4th dimension to a fixed plane adjust by an offset
+    # followed by a scaling by cutoff.
+
+    # lambda_plane_idxs are typically 0 or 1 and allows us to turn off an interaction
+    # independent of the lambda value.
+
+    # lambda_offset_idxs are typically 0 and 1, and allows us to adjust the w coordinate
+    # in a lambda-dependent way.
     d4 = cutoff*(lambda_plane_idxs + lambda_offset_idxs*lamb)
     d4 = np.expand_dims(d4, axis=-1)
     x4 = np.concatenate((x3, d4), axis=1)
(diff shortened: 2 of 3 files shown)


proteneer:
Additional note: this is the base case, there are more efficient ways of doing this, especially in the context of relative free energy calculations to improve overlap.

timemachine/potentials/jax_utils.py
```diff
@@ -8,7 +8,7 @@
 
 def convert_to_4d(x3, lamb, lambda_plane_idxs, lambda_offset_idxs, cutoff):
     # d4 = lambda_to_w(lamb, lambda_plane_idxs, lambda_offset_idxs, cutoff)
-    d4 = cutoff*lambda_plane_idxs + lambda_offset_idxs*lamb
```
maxentile, line 11: @proteneer : Could you please add a line comment on what this updated conversion function should do?

tests/test_md.py
```diff
@@ -46,7 +46,7 @@ def test_fwd_mode(self):
             lambda_offset_idxs,
             p_scale=3.0,
             # cutoff=0.5,
-            cutoff=100.0
```
maxentile, line 49: Indicates we may need to modify any instances where cutoff may have been `None` or very large

maxentile:
Looks good! Keeping the lambda range [0,1] for the nonbonded interactions should reduce the number of things a user has to think about, and will also make it easier in the future to mix-and-match with other ways to construct intermediate potentials u(x; lambda).

Points I'll want to look at more carefully soon:
- [ ] Double-checking for any possibility that cutoff is ambiguous / defined in more than one place
- [ ] Updating comments and examples where lambda > 1.0

tests/test_md.py
proteneer: This was accidentally introduced originally, and I'm restoring it to something more reasonable so we can test different cutoff.

proteneer:
> Looks good! Keeping the lambda range [0,1] for the nonbonded interactions should reduce the number of things a user has to think about, and will also make it easier in the future to mix-and-match with other ways to construct intermediate potentials u(x; lambda).

Agreed!

> Points I'll want to look at more carefully soon:

Some of the uses are rather application specific - I'm explicitly letting @schmolly know that she can change her scripts in pose_dock to bound lambda between 0 and 1.

proteneer:
To add a little more color commentary on the ```lambda_plane_idxs``` and ```lambda_offset_idxs``` parameters to show why it's useful:

Suppose we wanted to do a 1-stage decoupling with interpolation on the nonbonded terms.

Let lambda=0 be R_A fully interacting, R_B non-interacting
Let lambda=1 be R_B fully interacting, R_A non-interacting.

We need to move R_A from w=0 to cutoff, and R_B from w=-cutoff to w=0. This is accomplished by setting:

-lambda_plane_idxs for R_A to be 0, and lambda_plane_idxs for R_B to be -1.
-lambda_offset_idxs for R_A and R_B to be 1

Alternatively, we can add exclusions between R_A and R_B and do:

-lambda_plane_idxs for R_A to be 0, and lambda_plane_idxs for R_B to be 1.
-lambda_offset_idxs for R_A to be 1, and lambda_offset_idxs for R_B to be -1

schmolly:
Should lambda be in [0, 1) or [0, 1]?

proteneer:
[0, 1] Sorry that was a typo.

schmolly:
Would the lambda_offset_idxs ever be anything other than -1 or 1? Or is their intended use just to indicate the direction of movement along w?

proteneer:
I think most typical use cases are -1, 0, and 1. It's currently typed to an int, so setting it to 2 would make w increase to cutoff "twice as fast", i.e. by lambda=0.5

schmolly (approved):
LGTM

maxentile (approved):
Looks great!

PR #305, opened 2020-12-08
Parameter Interpolation

proteneer:
This PR implements parameter interpolation:

```
new_param = (1-lambda)*params[i] + lambda*params[i]
```

This will allow us to alchemically interpolate charges etc.

### DIFF
--- a/tests/test_parameter_interpolation.py
+++ b/tests/test_parameter_interpolation.py
@@ -1,0 +1,87 @@
+from jax.config import config; config.update("jax_enable_x64", True)
+
+import functools
+import numpy as np
+
+from common import GradientTest
+from timemachine.lib import potentials
+
+from common import prepare_water_system
+
+def interpolated_potential(conf, params, box, lamb, u_fn):
+    assert params.size % 2 == 0
+
+    CP = params.shape[0]//2
+    new_params = (1-lamb)*params[:CP] + lamb*params[CP:] 
+
+    return u_fn(conf, new_params, box, lamb)
+
+
+class TestInterpolatedPotential(GradientTest):
+
+    def test_nonbonded(self):
+
+        # we test nonbonded terms to ensure that we're doing the chain rule with du_dl correctly since
+        # bonded terms do not depend on lambda.
+
+        np.random.seed(4321)
+        D = 3
+
+        cutoff = 1.0
+        size = 36
+
+        water_coords = self.get_water_coords(D, sort=False)
+        coords = water_coords[:size]
+        padding = 0.2
+        diag = np.amax(coords, axis=0) - np.amin(coords, axis=0) + padding
+        box = np.eye(3)
+        np.fill_diagonal(box, diag)
+
+        N = coords.shape[0]
+
+        lambda_plane_idxs = np.random.randint(low=0, high=2, size=N, dtype=np.int32)
+        lambda_offset_idxs = np.random.randint(low=0, high=2, size=N, dtype=np.int32)
+
+        for lamb in [0.0, 0.2, 1.0]:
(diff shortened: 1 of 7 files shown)


proteneer:
The only tricky part of this PR was implementation of idempotent du/dl calls - which resorts to using a pseudo-associative reduction from CUB.

maxentile:
Looks great! I'll be a bit longer with my line-by-line review, but for now I have one high-level question.

This currently hard-codes that the parameters are to be interpolated linearly, and adds CUDA code to perform the linear interpolation. However, linearly interpolating the parameters can be significantly less efficient than other informed interpolation schemes, even for tasks like changing the location and force constant of a spring. For example, see Figure 2 of (Gelman, Meng, 1998) http://www.stat.columbia.edu/~gelman/research/published/path2.pdf  for an illustration and analysis of optimal paths in parameter space for annealing between two 1D Gaussians.

Allowing flexibility for user-defined interpolation functions will likely be important.

Question: would it be much worse if the parameter-interpolating logic were performed on the Python side instead?

jchodera:
Could you metaprogram this, and have the Python build the kernels, to get the best of both worlds?

proteneer:
100% agreed on there being better/more efficient interpolation schemes. I took a brief look inside the code for perses beforehand and most of what I saw was simple linear interpolation. If all you're doing is computing energies and forces, you can already just interpolate on the python side and just use the plain, unwrapped classes.

The key issue is how often we want du/dl and du/dparams (if we want them at all). In principle one can post-process the trajectory and compute the chain rule dot product terms on the python side explicitly (i.e. d/dl u = du/dl + du/dp.dp/u, and d/dp u = du/dp.dp/dpi), but this does get unwieldy for large systems (especially if we want to optimize protein/water parameters later).

Perhaps a middle ground is to do prototyping of other interpolation schemes using the slow way, and once we find a better scheme we can implement an optimized version on the GPU?

proteneer:
> Could you metaprogram this, and have the Python build the kernels, to get the best of both worlds?

Probably - this might be a good prototype of Zygote/Enzyme's autodiff machinery to try out.

maxentile:
>Perhaps a middle ground is to do prototyping of other interpolation schemes using the slow way, and once we find a better scheme we can implement an optimized version on the GPU?

This sounds good to me -- my guess is we'll be able to get a good amount of mileage out of the linear parameter interpolation approach (which has been used successfully in many prior studies, as in eqs. 9,10,11 of http://theory.rutgers.edu/resources/pdfs/Giese_JChemTheoryComput_2018_ASAP.pdf ).

And a good amount of further mileage could probably be gotten out of hard-coding a few nonlinear parameter interpolation schemes for the various standard forcefield terms individually, after prototyping and testing a few options on the Python side. (Probably next steps there would be to interpolate Harmonic{Bond|Angle}Force parameters using something like the scheme in eqs. 70,71 of the (Gelman, Meng, 1998) reference, to do a literature search for guidance on how to do LJ parameter interpolation, and to think a bit about how to do PeriodicTorsionForce parameter interpolation.)

>Could you metaprogram this, and have the Python build the kernels, to get the best of both worlds?

That could be appealing, if Python-based parameter interpolation proves too slow for prototyping, and if there's a simple enough way to implement this. I imagine a template-based code-generator could probably get the job done here, assuming the interpolation functions are simple symbolic expressions defined in terms of params_A, params_B, and lambda, and assuming we use something like sympy to do the symbolic derivatives. There's probably a broader discussion on what parts of the CUDA implementations here might be amenable to some form of code generation.

maxentile (approved):
Looks great! This will directly expand the range of transformations that can be performed using timemachine, and provides a solid basis for further elaborations.

Issue #306, opened 2020-12-10
Make u, du_dx and x_t internally consistent


proteneer:
/facepalm

proteneer:
Not sure about observables right now, but these three things should either be internally consistent or we should prevent access to them (and instead rely on stand alone forces).

@maxentile thoughts?

maxentile:
Good catch! I think it would be ideal if these all behaved as much as possible like pure functions of x. The programming style encouraged in Jax is pretty functional, and I think having any of these functions behave off-by-a-timestep can cause surprises.

On the other hand, I'm not sure I fully appreciate the performance reasons for having some of these functions reference results computed the previous timestep.

proteneer:
So after digging into this a little deeper - there's a mismatch between how we do things and how OpenMM does this. OpenMM has a flag that checks to see if forces and energies have been computed yet, and internal state is updated accordingly. However, our step(lambda) function takes in a lambda parameter to make NEQ calculations simpler, and this would require that our get_u_t() and get_du_dx_t() function *also* take in a lambda parameter since the Context does not store lambda internally as part of its state (unlike OpenMM).

What do you think about us just simply explicitly renaming the getter functions to properly match their semantics:

```
get_u_t() -> get_u_t_minus_one()
get_du_dx_t() -> get_du_dx_t_minus_one()
```

maxentile:
>What do you think about us just simply explicitly renaming the getter functions to properly match their semantics:

This seems tricky, and I probably don't quite appreciate the value added by these stateful getters over the  `u(x, lambda)` functions.

I'll note a possible remaining naming ambiguity -- to compute nonequilibrium works, we'll frequently need to compute `u(x, lambda_t)` and `u(x, lambda_t_minus_1)` (same `x`, different `lambda`), which we can do unambiguously with the python wrapped potential objects. Misremembering whether `get_u_t_minus_1` returns `u(x_t_minus_1, lambda_t)` or `u(x_t_minus_1, lambda_t_minus_1)` is foreseeable and could lead to subtle errors.

maxentile:
Resolved by https://github.com/proteneer/timemachine/pull/311

PR #311, opened 2021-01-06
Rename and deprecate du_dx_t and u_t functions.

proteneer:
Addresses #306 based on offline comments with @maxentile 

I think this is explicit enough. We will remove these functions entirely when we re-work the integrator system.


maxentile (approved):
Nice work! Looks good to me

proteneer:
Thanks for catching the missing changes!

Issue #313, opened 2021-01-08
Alchemically scale exclusions or turn off nonbonded interactions for multiple R-groups


proteneer:
Suppose we wish to turn off the interaction of R_A, where R_A is comprised to *two* separate R-groups (on different ends of the molecule) that are decoupled/annihilated.  In this particular case, the nonbonded terms will introduce an **intramolecular** force between the two groups, which will then violate the assumption that dummy atoms are non-interacting. 

We can resolve this by ensuring that parameter interpolation *also* turns off the charge and the epsilon value, so that intramolecular interactions on the R-groups are fully annihilated.

proteneer:
Note that the theoretical justification is identical to that of Boresch's paper on orientational restraints, whereby, adding additional, ring forming bonds, will violate the ability to factorize the partition function in a clean way.

proteneer:
If we *only* have a single a R-group that is being deleted, then we don't have to do this. So we should be okay most of the time already, but we should nevertheless becareful.

PR #314, opened 2021-01-08
Fixes GHI #313 regarding nonbonded parameters.

proteneer:
This PR fixes issue #313 where charges and epsilons were not turned off at the end points.

### DIFF
--- a/fe/topology.py
+++ b/fe/topology.py
@@ -438,8 +438,7 @@ class SingleTopology():
         """
         Interpolate two sets of per-particle parameters.
 
-        This can be used to interpolate nonbonded parameters,
-        coordinates, etc.
+        This can be used to interpolate masses, coordinates, etc.
 
         Parameters
         ----------
@@ -470,6 +469,54 @@ class SingleTopology():
 
         return jnp.array(src_params), jnp.array(dst_params)
 
+    def interpolate_nonbonded_params(self, params_a, params_b):
+        """
+        Special interpolation method for nonbonded parameters. For R-group atoms,
+        their charges and vdw eps parameters are scaled to zero. Vdw sigma
+        remains unchanged. This method is needed in order to ensure that R-groups
+        that branch from multiple distinct attachment points are fully non-interacting
+        to allow for factorization of the partition function. In order words, this function
+        implements essentially the non-softcore part of parameter interpolation.
+
+        Parameters
+        ----------
+        params_a: np.ndarray, shape [N_A, 3]
+            Nonbonded parameters for the mol_a
+
+        params_b: np.ndarray, shape [N_B, 3]
+            Nonbonded parameters for the mol_b
+
+        Returns
+        -------
+        tuple: (src, dst)
+            Two np.ndarrays each of shape [N_C, ...]
+
+        """
+
+        src_params = [None]*self.get_num_atoms()
+        dst_params = [None]*self.get_num_atoms()
+
+        # src -> dst is turning off the parameter
+        for a_idx, c_idx in enumerate(self.a_to_c):
+            params = params_a[a_idx]
+            src_params[c_idx] = params
+            if self.c_flags[c_idx] != 0:
+                assert self.c_flags[c_idx] == 1
+                dst_params[c_idx] = jnp.array([0, params[1], 0]) # q, sig, eps
+
+        # b is initially decoupled
+        for b_idx, c_idx in enumerate(self.b_to_c):
+            params = params_b[b_idx]
+            dst_params[c_idx] = params
+            # this will already be processed when looping over a
+            if self.c_flags[c_idx] == 0:
+                assert src_params[c_idx] is not None
+            else:
+                assert self.c_flags[c_idx] == 2
+                src_params[c_idx] = jnp.array([0, params[1], 0]) # q, sig, eps
+
+        return jnp.array(src_params), jnp.array(dst_params)
+
 
     def parameterize_nonbonded(self, ff_q_params, ff_lj_params):
         # Nonbonded potentials combine through parameter interpolation, not energy interpolation.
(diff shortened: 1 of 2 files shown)


fe/topology.py
```diff
@@ -470,6 +469,54 @@ def interpolate_params(self, params_a, params_b):
 
         return jnp.array(src_params), jnp.array(dst_params)
 
+    def interpolate_nonbonded_params(self, params_a, params_b):
+        """
+        Special interpolation method for nonbonded parameters. For R-group atoms,
+        their charges and vdw eps parameters are scaled to zero. Vdw sigma
+        remains unchanged. This method is needed in order to ensure that R-groups
+        that branch from multiple distinct attachment points are fully non-interacting
```
maxentile, line 477: Sorry, I'm having trouble following the current description of the problem in this docstring. It's unclear to me why the issue would only arise when considering multiple R-groups that branch from multiple distinct attachment points, but not when a single R group is large enough to contain atom pairs that interact via nonbonded terms. Perhaps it would be clarifying to add even a bit more description of the desired behavior here-- or add a pointer to the behavior of a failing test case when the issue is not handled properly.

tests/test_topology.py
```diff
@@ -124,27 +124,104 @@ def test_nonbonded(self):
...
+        np.testing.assert_array_equal(qlj_b[6:], params_dst[7:])
+        np.testing.assert_array_equal(np.array([
+            [0, qlj_b[6][1], 0],
+            [0, qlj_b[7][1], 0]]), params_src[7:]
+        )
+
+
+    def test_nonbonded_optimal_map(self):
+
+        # map benzene H to phenol O, leaving a daggling phenol H
```
maxentile, line 170:
```suggestion
        # map benzene H to phenol O, leaving a dangling phenol H
```

tests/test_topology.py
```diff
@@ -124,27 +124,104 @@ def test_nonbonded(self):
...
+        # test interpolation of parameters
+        bt_a = topology.BaseTopology(self.mol_a, self.ff)
+        qlj_a, pot_a = bt_a.parameterize_nonbonded(self.ff.q_handle.params, self.ff.lj_handle.params)
+        bt_b = topology.BaseTopology(self.mol_b, self.ff)
+        qlj_b, pot_b = bt_b.parameterize_nonbonded(self.ff.q_handle.params, self.ff.lj_handle.params)
+
+        qlj_c = np.mean([params[:len(params)//2], params[len(params)//2:]], axis=0)
+
+        params_src = params[:len(params)//2]
+        params_dst = params[len(params)//2:]
```
maxentile, line 214: I think recomputing the indices more than a couple times using the `[:len(params)//2]`, `[len(params)//2:]` pattern can be error-prone. Consider replacement with something like

```suggestion
        # compute indices for source and destination parameters
        middle_ind = len(params)//2
        src_inds = np.arange(0, middle_ind)
        dst_inds = np.arange(middle_ind, len(params))
        
        # fetch source and destination parameters
        params_src = params[src_inds]
        params_dst = params[dst_inds]
        
        # perform computation on source and destination parameters
        qlj_c = np.mean([params_src, params_dst], axis=0)
```

tests/test_topology.py
```diff
@@ -124,27 +124,104 @@ def test_nonbonded(self):
...
+
+        # test r-group in B
+        np.testing.assert_array_equal(qlj_b[6:], params_dst[7:])
+        np.testing.assert_array_equal(np.array([
+            [0, qlj_b[6][1], 0],
+            [0, qlj_b[7][1], 0]]), params_src[7:]
+        )
+
+
+    def test_nonbonded_optimal_map(self):
```
maxentile, line 168: Add one-line statement to docstring of the assertion being tested

```suggestion
    def test_nonbonded_optimal_map(self):
        """assert that coordinates and nonbonded parameters can be averaged in benzene -> phenol transformation"""
```

tests/test_topology.py
```diff
@@ -124,27 +124,104 @@ def test_nonbonded(self):
...
+        np.testing.assert_array_equal(x_b[-1], x_avg[7]) # dangling H
+
+        params, vjp_fn, pot_c = jax.vjp(
+            st.parameterize_nonbonded,
+            self.ff.q_handle.params,
+            self.ff.lj_handle.params,
+            has_aux=True
+        )
+
+        vjp_fn(np.random.rand(*params.shape))
```
maxentile, line 201: Should we check other aspects of the result of calling vjp_fn here?

maxentile (approved):
Nice work! Appears to address an issue that could impact the correctness of relative binding free energy estimates when parameter interpolation rather than energy interpolation is used for nonbonded interactions. Made some minor nitpick comments, but overall looks good to me.

fe/topology.py
proteneer: I'm actually not 100% sure for the case of a large single R-group, I know that with Boresch restraints for ABFE, we do not need to turn off the ligand's nonbonded interactions, even though they branch off from a single virtual bond. I think it's always safer to turn them off than to leave them on, and it's more consistent with the behavior seen in perses.

tests/test_topology.py
proteneer: Will fix

tests/test_topology.py
proteneer: This is mainly to check that the vjp is actually callable, the correctness part I'm relying on the tests within jax. This is mainly for making sure we don't actually have a onp (original numpy) call as opposed to a jnp (jax numpy) call within the code.

tests/test_topology.py
proteneer: Will fix

tests/test_topology.py
proteneer: I'll see if I can make this require less mental arithmetic - but there are other parts of the code where this sort of style shows up.

PR #315, opened 2021-01-13
[WIP] Low hanging fruits of nonbonded kernel optimizations

proteneer:
This PR attempts to address some of the performance regressions that we've been seeing. In particular, it comes close to matching the speed of computeNonbonded kernel in OpenMM. However, we do not do any crazy JIT compilation. All in all, there's about two dozen various tricks used and is probably too much to list here, though feel free to ping me offline. Note that this PR does *not* address the neighborlist issue yet, that will be dealt with in a separate PR later.

This PR does introduce one major change to the parameters expected by the nonbonded kernel, namely that the epsilons are expected to be in sqrt form, and the sigmas should be halved. This is to avoid an expensive sqrt calculation in the inner loop of the kernel. The openmm system deserializer and the off forcefield handlers have been updated to address this.


proteneer:
The tests probably won't pass, but it's getting late and I will deal with them tomorrow.

PR #316, opened 2021-01-13
Update neighborlist construction

proteneer:
This PR reconstructs the neighborlist and the hilbert ordering only when the distance of any particle has moved beyond padding/2. Matches behavior as that of OpenMM, and will soon get close to parity in terms of performance.

Main changes;

- Implements a context.multiple_steps() to save on device synchronization overhead
- Neighborlist is now padded with a 1A offset and rebuilt roughly once every 5-10 steps. However, the actual non-bonded computation is now roughly 10% more expensive.
- Hilbert reordering is done only when neighborlist is rebuilt since Radix sweeping gets expensive.
- Removed unnecessary calls to cudaDeviceSynchronize()
- Modifed the fixed pointe accumulation order in the nonbonded kernels.

The "holy grail" of implementations is an incremental/delta based neighborlist algorithm. But as far as I know, no one's figured out how to do this on the GPU yet.

```
 GPU activities:   52.41%  11.1187s     50000  222.37us  219.23us  277.54us  void k_nonbonded_du_dx<float, bool=0>(int, double const *, double const *, double const *, double, int const *, int const *, double, double, int const *, unsigned int const *, __int64*)
                    9.18%  1.94830s      8876  219.50us  199.52us  251.71us  void k_find_blocks_with_ixns<float>(int, double const *, double const *, double const *, double const *, unsigned int*, int*, unsigned int*, unsigned int*, double)
                    5.19%  1.10064s     50500  21.794us  12.640us  25.343us  void k_nonbonded_exclusions<float>(int, double const *, double const *, double const *, double, int const *, int const *, int const *, double const *, double, double, __int64*, __int64*, __int64*, __int64*)
                    4.51%  957.47ms     44380  21.574us  20.800us  42.688us  void cub::RadixSortScanBinsKernel<cub::DeviceRadixSortPolicy<unsigned int, unsigned int, int>::Policy700, int>(unsigned int*, int)
                    3.27%  693.96ms     50000  13.879us  13.408us  65.696us  void timemachine::update_forward<double>(int, int, double, timemachine::update_forward<double> const *, timemachine::update_forward<double> const , timemachine::update_forward<double> const , timemachine::update_forward<double>*, timemachine::update_forward<double> const *, __int64 const *, timemachine::update_forward<double>)
                    2.93%  622.50ms    101000  6.1630us  5.4400us  58.080us  void k_permute<double>(int, unsigned int const *, double const *, double*)
                    2.62%  554.84ms     26628  20.836us  16.448us  49.087us  void cub::DeviceRadixSortDownsweepKernel<cub::DeviceRadixSortPolicy<unsigned int, unsigned int, int>::Policy700, bool=1, bool=0, unsigned int, unsigned int, int>(cub::DeviceRadixSortPolicy<unsigned int, unsigned int, int>::Policy700 const *, cub::DeviceRadixSortDownsweepKernel<cub::DeviceRadixSortPolicy<unsigned int, unsigned int, int>::Policy700, bool=1, bool=0, unsigned int, unsigned int, int>*, bool=1 const *, cub::DeviceRadixSortDownsweepKernel<cub::DeviceRadixSortPolicy<unsigned int, unsigned int, int>::Policy700, bool=1, bool=0, unsigned int, unsigned int, int>**, bool=0*, cub::DeviceRadixSortDownsweepKernel<cub::DeviceRadixSortPolicy<unsigned int, unsigned int, int>::Policy700, bool=1, bool=0, unsigned int, unsigned int, int>**, int, int, cub::GridEvenShare<cub::DeviceRadixSortDownsweepKernel<cub::DeviceRadixSortPolicy<unsigned int, unsigned int, int>::Policy700, bool=1, bool=0, unsigned int, unsigned int, int>**>)
                    2.38%  504.88ms     50500  9.9970us  3.6160us  48.928us  void k_harmonic_angle_inference<float, int=3>(int, double const *, double const *, int const *, __int64*, double*, double*)
                    2.33%  494.80ms     50500  9.7980us  4.7360us  50.687us  void k_periodic_torsion<float, int=3>(int, double const *, double const *, int const *, __int64*, double*, double*)
                    1.96%  415.69ms     50000  8.3130us  7.6480us  44.832us  void gen_sequenced<curandStateXORWOW, double2, normal_args_double_st, __operator_&__(double2 curand_normal_scaled2_double<curandStateXORWOW>(curandStateXORWOW*, normal_args_double_st)), rng_config<curandStateXORWOW, curandOrdering=101>>(curandStateXORWOW*, double2*, unsigned long, unsigned long, normal_args_double_st)
```

### DIFF
--- a/timemachine/cpp/src/kernels/k_nonbonded.cuh
+++ b/timemachine/cpp/src/kernels/k_nonbonded.cuh
@@ -74,6 +74,62 @@ void __global__ k_coords_to_kv(
 
 }
 
+template <typename RealType>
+void __global__ k_check_rebuild_box(
+    const int N,
+    const double *new_box,
+    const double *old_box,
+    int *rebuild) {
+
+    const int idx = blockIdx.x*blockDim.x + threadIdx.x;
+
+    if(idx > 9) {
+        return;
+    }
+
+    // (ytz): box vectors have exactly 9 components
+    // we can probably derive a looser bound later on.
+    if(old_box[idx] != new_box[idx]) {
+        rebuild[0] = 1;
+    }
+
+}
+
+template <typename RealType>
+void __global__ k_check_rebuild_coords(
+    const int N,
+    const double *new_coords,
+    const double *old_coords,
+    double padding,
+    int *rebuild) {
+
+    const int atom_idx = blockIdx.x*blockDim.x + threadIdx.x;
+
+    if(atom_idx >= N) {
+        return;
+    }
+
+    RealType xi = old_coords[atom_idx*3+0];
+    RealType yi = old_coords[atom_idx*3+1];
+    RealType zi = old_coords[atom_idx*3+2];
+
+    RealType xj = new_coords[atom_idx*3+0];
+    RealType yj = new_coords[atom_idx*3+1];
+    RealType zj = new_coords[atom_idx*3+2];
+
+    RealType dx = xi - xj;
+    RealType dy = yi - yj;
+    RealType dz = zi - zj;
+
+    RealType d2ij = dx*dx + dy*dy + dz*dz;
+    if(d2ij > static_cast<RealType>(0.25)*padding*padding) {
+        // (ytz): this is *safe* but technically is a race condition
+        rebuild[0] = 1;
+    }
+
+
+}
+
 
 template <typename RealType>
 void __global__ k_permute(
@@ -324,6 +380,7 @@ void __global__ k_nonbonded_du_dx(
         // is set to cutoff are non-interacting.
         if(d2ij < cutoff_squared  && atom_j_idx > atom_i_idx && atom_j_idx < N && atom_i_idx < N) {
 
+
             // electrostatics
             RealType inv_dij = rsqrt(d2ij);
             RealType dij = d2ij*inv_dij;
@@ -333,8 +390,6 @@ void __global__ k_nonbonded_du_dx(
             RealType es_prefactor = qij*inv_dij*real_es_factor(real_beta, dij, inv_d2ij);
 
             // lennard jones force
-
-            // RealType lj_prefactor = 0;
             RealType delta_prefactor = es_prefactor;
             if(eps_i != 0 && eps_j != 0) {
                 RealType eps_ij = eps_i * eps_j;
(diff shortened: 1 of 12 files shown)


proteneer:
This currently runs about 253ns/day on the DHFR tests/benchmark.py script.

proteneer:
@maxentile I think this is ready to go

maxentile:
Thanks! I'll finish reviewing by end of tomorrow.

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -74,6 +74,41 @@ void __global__ k_coords_to_kv(
...
+    RealType dz = zi - zj;
+
+    RealType d2ij = dx*dx + dy*dy + dz*dz;
+    if(d2ij > static_cast<RealType>(0.25)*padding*padding) {
+        // (ytz): this is *safe* but technically is a race condition
+        rebuild[0] = 1;
+    }
+
+
+}
```
maxentile, line 110: Should this function also check if the box has changed? (Or does changing the box automatically trigger a neighborlist rebuild elsewhere in the code)?

(Consider a chain of neighboring particles x[i] in the center of a very large box. If I shrink the box so that the borders are close to x[0] and x[N], then (x[0], x[N]) could now be neighbors, even though none of the particle coordinates x[i] changed.)

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -345,29 +379,33 @@ void __global__ k_nonbonded_du_dx(
...
             gi_z += FLOAT_TO_FIXED(delta_prefactor*delta_z);
 
-            gj_x -= FLOAT_TO_FIXED(delta_prefactor*delta_x);
-            gj_y -= FLOAT_TO_FIXED(delta_prefactor*delta_y);
-            gj_z -= FLOAT_TO_FIXED(delta_prefactor*delta_z);
+            // (ytz): the cast relies on undefined behavior since the exclusion terms will overflow.
+            // Thus far, the current gen GPUs are deterministic in the resulting value. So long
+            // as the UB stays deterministic then this should be fine. There's no guarantee though that
+            // NVIDIA will keep the behavior consistent. Furthermore, do *not* attempt to swap signs via:
+            // gj_* -= FLOAT_TO_FIXED(delta_prefactor*delta_*); This is incorrect
```
maxentile, line 391: This still makes me a bit nervous.

Perhaps after this PR we can discuss offline if there are any other options for handling exclusions that could be nearly as fast as this approach

maxentile (commented):
Nice work! Clarification question: is all of the logic required to check if the neighbor list needs to be rebuilt defined in `k_check_rebuild`, or is additional logic defined in another function?

proteneer:
> Nice work! Clarification question: is all of the logic required to check if the neighbor list needs to be rebuilt defined in `k_check_rebuild`, or is additional logic defined in another function?

The neighborlist only needs to be rebuilt if any particle has moved by more than padding/2 (which is checked by k_check_rebuild)

timemachine/cpp/src/kernels/k_nonbonded.cuh
proteneer: Agreed - there's other assumptions we make in the code that's somewhat similar. Eg. we assume that the warp size is always 32 threads though NVIDIA doesn't guarantee this.

timemachine/cpp/src/kernels/k_nonbonded.cuh
proteneer: Hm. Currently we're running only in NVT, with NPT this definitely would be a problem, and we'd need to rebuild the neighborlist if the box size changes.

maxentile: Even without running an NPT simulation, I guess we could also run into an issue if the neighbor list was last computed with some box size, and we later query energies/forces/... for a different, smaller box size. (For example, if we also allow to optimize the box vectors during energy minimization.)

Perhaps we should add a check that `old_box` and `new_box` are the same before calling `k_check_rebuild`?

proteneer: This seems like a reasonable thing to add. I'll fix.

proteneer:
@maxentile I added a new commit that adds a check for this behavior. It's not optimal (as we should only need to recheck when any of the box dimensions shrink, but this would only apply to orthogonal box vectors) but is formally correct.

proteneer:
@maxentile  Let me know if there's anything else that's missing before you approve and I can merge

maxentile:
Nice work! The check for box size changes was the only thing I noticed missing -- looks great to merge now.

(Unrelated to the content of this PR, I did notice a couple things I'd like to update about the benchmark.py script to improve the reliability of timing results, which I'll address early next week.)

maxentile (approved):

PR #317, opened 2021-01-14
update pose_dock, add rigorous work and dock_and_equilibrate

schmolly:


docking/dock_and_equilibrate.py
```diff
@@ -0,0 +1,387 @@
...
+        x0 = np.concatenate([solvated_host_coords, orig_guest_coords])
+        v0 = np.zeros_like(x0)
+        print(
+            f"SYSTEM", f"guest_name: {guest_name}", f"num_atoms: {len(x0)}",
+        )
+
+        for atom_num in constant_atoms:
+            combined_masses[atom_num - 1] += 50000
+
+        seed = 2020
```
proteneer, line 163: Maybe we should make this 2021 :)

docking/dock_and_equilibrate.py
```diff
@@ -0,0 +1,387 @@
...
+                        outdir,
+                        guest_name,
+                        str(step).zfill(len(str(insertion_steps))),
+                        f"ins",
+                    )
+                if too_much_force(ctxt, combined_bps, host_box, u_impls, lamb):
+                    calc_work = False
+                    break
+
+        if (
```
proteneer, line 204: Note that this condition is only true for ABFE - if we do RBFE this condition would no longer hold.

docking/dock_and_equilibrate.py
```diff
@@ -0,0 +1,387 @@
...
+    print(
+        f"{stage}\t"
+        f"guest_name: {guest_name}\t"
+        f"step: {str(step).zfill(len(str(n_steps)))}\t"
+        f"lambda: {lamb:.2f}\t"
+        f"energy: {energy:.2f}"
+    )
+
+
+def too_much_force(ctxt, bps, box, u_impls, lamb):
```
proteneer, line 259: This function should be called sparingly. It is *hideously* expensive due to all the memcpys. Even once every 100 steps might be too often.

docking/pose_dock.py
```diff
@@ -66,11 +70,40 @@ def pose_dock(
...
+    host_nb_bp = None
+    for bp in host_potentials:
+        if isinstance(bp, potentials.Nonbonded):
+            # (ytz): hack to ensure we only have one nonbonded term
+            assert host_nb_bp is None
+            host_nb_bp = bp
+        else:
+            final_potentials.append(bp)
+
+    padding = 0.1
```
proteneer, line 98: note to self and @maxentile  we should really fix this later on. This padding was done to address the particles that are too close to the boundary.

docking/pose_dock.py
```diff
@@ -281,18 +309,16 @@ def pose_dock(
         "--max_lambda",
         type=float,
         default=1.1,
```
proteneer, line 311: Lambda is now strictly between 0 and 1.

docking/dock_and_equilibrate.py
```diff
@@ -0,0 +1,387 @@
...
+    constant_atoms=[]
+):
+    """Solvates a host, inserts guest(s) into solvated host, equilibrates
+
+    Parameters
+    ----------
+
+    host_pdbfile: path to host pdb file to dock into
+    guests_sdfile: path to input sdf with guests to pose/dock
+    max_lambda: lambda value the guest should insert from or delete to
```
proteneer, line 40: Lambda is now strictly between 0 and 1.

tests/test_recipe.py
```diff
@@ -22,7 +22,7 @@ def test_recipe_from_rdkit():
```
proteneer, line 22: Once this is merged I will delete this file and the recipe class altogether.

docking/pose_dock.py
maxentile: Agreed!

maxentile (approved):
Great work! And thanks for walking through this on Thursday. Looks good to me.

proteneer (approved):

proteneer:
LGTM. Feel free to merge.

PR #318, opened 2021-01-19
Update README and delete unused files

proteneer:
Summary:

-No longer need to set BUILD_TEST=ON for tests. This was needed before we switch to fixed point accumulators in the nonbonded kernel. But now it's no longer necessary.
-Updated README
-Removed unused/deprecated files (eg. Recipe class is gone, and replaced by *Topology classes)


maxentile (approved):
Nice work! Great improvements to the top-level readme, and it looks safe to remove these files now

PR #319, opened 2021-01-19
Update timer logic in benchmark.py

maxentile:
The MD speed estimate reported by `tests/benchmark.py` was based on a running average that included a costly one-time setup step. This PR updates the timer to report instantaneous ("per-batch") speed estimates rather than running averages, and not start the timer until after one-time setup is complete.

An example of the difference in reported speed estimates is shown below.

speed_estimate_comparison

### DIFF
--- a/tests/benchmark.py
+++ b/tests/benchmark.py
@@ -91,29 +91,37 @@ def benchmark_dhfr():
 
     lamb = 0.0
 
-    start = time.time()
-
     writer = PDBWriter([host_pdb.topology], "dhfr.pdb")
 
     num_batches = 100
     steps_per_batch = 1000
+    start = time.time()
+
+    lambda_schedule = np.ones(steps_per_batch)*lamb
+
+    seconds_per_day = 86400
 
     for batch in range(num_batches):
-        lambda_schedule = np.ones(steps_per_batch)*lamb
-        ctxt.multiple_steps(lambda_schedule)
 
-        delta = time.time()-start
-        steps_per_second = (batch+1)*steps_per_batch/delta
-        seconds_per_day = 86400
+        # time the current batch
+        batch_start = time.time()
+        ctxt.multiple_steps(lambda_schedule)
+        batch_end = time.time()
+
+        delta = batch_end - batch_start
+        steps_per_second = steps_per_batch / delta
+
         steps_per_day = steps_per_second*seconds_per_day
         ps_per_day = dt*steps_per_day
         ns_per_day = ps_per_day*1e-3
 
-        print((batch+1)*steps_per_batch, "steps @ ", ns_per_day, " ns/day")
+        print(f'steps per second: {steps_per_second:.3f}')
+        print(f'ns per day: {ns_per_day:.3f}')
+
         # coords = recenter(ctxt.get_x_t(), box)
         # writer.write_frame(coords*10)
 
-    print("total time", time.time() - start)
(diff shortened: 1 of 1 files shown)


tests/benchmark.py
```diff
@@ -91,29 +91,37 @@ def benchmark_dhfr():
...
     for batch in range(num_batches):
-        lambda_schedule = np.ones(steps_per_batch)*lamb
+
+        # time the current batch
+        batch_start = time.time()
         ctxt.multiple_steps(lambda_schedule)
+        batch_end = time.time()
+
+        delta = batch_end - batch_start
+        steps_per_second = steps_per_batch / delta
```
proteneer, line 112: This is always using steps_per_batch to estimate the ns/day. This value likely is not going to converge with only steps_per_batch steps. The neighborlist code sometimes will sometimes result in more interactions than other parts of the code, and the statistics need to be averaged over a large # of samples.

In fact, simply moving the ```start = time.time()``` to right before ```for batch in range(num_batches):``` would probably be sufficient. The overhead is most likely coming from the PDBWriter constructor, which does a ton of tempfile, parsing, etc.

``` python
class PDBWriter():

    def __init__(self, objs, out_filepath): 
        assert len(objs) > 0

        rd_mols = []
        # super jank
        for obj in objs:
            if isinstance(obj, app.Topology):
                with tempfile.NamedTemporaryFile(mode='w') as fp:
                    # write
                    PDBFile.writeHeader(obj, fp)
                    PDBFile.writeModel(obj, np.zeros((obj.getNumAtoms(), 3)), fp, 0)
                    PDBFile.writeFooter(obj, fp)
                    fp.flush()
                    # read
                    rd_mols.append(Chem.MolFromPDBFile(fp.name, removeHs=False))

            if isinstance(obj, Chem.Mol):
                rd_mols.append(obj)

        combined_mol = rd_mols[0]
        for mol in rd_mols[1:]:
            combined_mol = Chem.CombineMols(combined_mol, mol)

        with tempfile.NamedTemporaryFile(mode='w') as fp:
            # write
            Chem.MolToPDBFile(combined_mol, fp.name)
            fp.flush()
            # read
            combined_pdb = app.PDBFile(fp.name)
            self.topology = combined_pdb.topology

        self.out_handle = open(out_filepath, "w")
        PDBFile.writeHeader(self.topology, self.out_handle)
        self.topology = self.topology
        self.frame_idx = 0
```

maxentile: Agreed -- the per-batch estimates with batch_size = 1000 steps fluctuated by +/- 1 percent from their mean value, and are expected to continue fluctuating. Will re-add running average, skipping PDBWriter constructor...

maxentile: speed_estimate_comparison

tests/benchmark.py
```diff
@@ -91,29 +91,44 @@ def benchmark_dhfr():
...
+        batch_end = time.time()
+
+        delta = batch_end - batch_start
+
+        batch_times.append(delta)
 
-        delta = time.time()-start
-        steps_per_second = (batch+1)*steps_per_batch/delta
-        seconds_per_day = 86400
+        steps_per_second = steps_per_batch / np.mean(batch_times)
```
proteneer, line 119: Interesting way of computing steps_per_second, seems right to me. note that mean takes O(N) take, but in practice this doesn't matter.

proteneer (approved):
Approving

tests/benchmark.py
proteneer: That's about as -1/sqrt(x) as I've ever seen

PR #322, opened 2021-01-21
Unify gRPC and multiprocess APIs

proteneer:
This PR attempts unify the gRPC client and the multiprocess client to the same API. The second pass will focus on integration with existing code. @maxentile  we will need to dispatch the staticmethod _simulate using this pattern. See tests for example usage.


proteneer:
fucked this up - need to rebase

PR #323, opened 2021-01-21
re-opened: gRPC and Multiprocess unification

proteneer:
This PR:

1. Adds a client.py file with classes that form a consistent API for both multiprocess and gRPC based parallelization
2. Adds three classes: a generic ProcessPoolClient, a CUDA specific CUDAPoolClient, and a gRPC based GRPCClient
3. Adds example usage via tests: tests/test_parallel.py
4. Cleans up the old protocol buffer files to make it more extensible, though also more dependent on pickling. Once we have a fixed pattern, we can formalize this a bit more.

gRPC client usage pattern is:
1. Launch workers on nodes.
2. Launch client after all the workers are alive.

Note that service_*.py files are generated automatically by protocol buffers and gRPC. They are vendored in to facilitate deployment.


parallel/client.py
```diff
@@ -0,0 +1,140 @@
...
+
+        client = ConcreteClient()
+
+        futures = []
+        for arg in args:
+            fut = client.submit(fn, *args)
+            futures.append(fut)
+
+        for fut in futures:
+            res = fut.result()
```
maxentile, line 49: The [examples in `tests/test_parallel.py`](https://github.com/proteneer/timemachine/blob/caa7a76311c17dafe22113543886486d1b284a8a/tests/test_parallel.py#L27-L34) look a bit more like this:

```suggestion
        futures = []
        for arg in args:
            fut = client.submit(task_fn, arg)
            futures.append(fut)
            
        res = []
        for fut in futures:
            res.append(fut.result())
```

tests/test_parallel.py
```diff
@@ -0,0 +1,114 @@
...
+    return os.environ['CUDA_VISIBLE_DEVICES']
+
+class TestCUDAPoolClient(unittest.TestCase):
+
+    def setUp(self):
+        max_workers = 4
+        self.cli = client.CUDAPoolClient(max_workers)
+
+    def test_submit(self):
+        arr = np.linspace(0, 1.0, 10)
```
maxentile, line 48: Unused
```suggestion
```

maxentile (approved):
Looks great! This will definitely streamline several of the scripts in `examples/`. I've also confirmed that the tests run successfully on the cluster. Shall I update the example scripts to use `CUDAPoolClient`?

proteneer:
Yes please give it a shot - I'm going to merge so you can rebase/merge your branch into/from master.

PR #324, opened 2021-01-22
Remove unused files and clean up documentation

proteneer:
This PR removes a couple of unused files and fixes some minor typos.


tests/test_nonbonded.py
```diff
@@ -417,59 +415,44 @@ def test_nonbonded(self):
...
 
-        # test numerical accuracy on a box of water
+            _, coords, box, _ = builders.build_water_system(6.2)
+            coords = coords/coords.unit
+            coords = coords[:size]
 
-        for size in [33, 231, 1050]:
+            N = coords.shape[0]
 
-            if not benchmark:
```
proteneer, line 426: There's other benchmarking tests now in test_nonbonded so this jank is no longer needed.

proteneer:
This is mostly a clean-up PR to remove old/unused code.

old_tests/invariances.py
```diff
@@ -1,89 +0,0 @@
-from jax.test_util import check_grads
-from scipy.stats import special_ortho_group
-import numpy as np
-import jax
-
-def assert_potential_invariance(energy_fn, x0, params, box=None):
-    # note: all potentials must support non-periodic implementations,
-    # defined by box being None
```
maxentile, line 8: Is it redundant to test translational and rotational invariance given the other ways the potentials are currently tested?

old_tests/test_bar.py
```diff
@@ -1,60 +0,0 @@
...
-            w[i][j] += 0.5*delta
-            left_edge = bar.EXP(w)
-            w[i][j] -= delta
-            right_edge = bar.EXP(w)
-            fd = (left_edge - right_edge)/delta
-            fd_exp[i][j] = fd
-            w[i][j] += 0.5*delta
-    return fd_exp
-
-class TestFreeEnergyDerivatives(unittest.TestCase):
```
maxentile, line 40: Is there a similar test we should add for TI derivatives?

maxentile (approved):
Looks good to me! Scanned over the deleted files and noted a couple places where it wasn't immediately clear to me what had obviated the things being deleted, if you could add a brief comment while the decisions are fresh in mind.

old_tests/invariances.py
proteneer: Most of the old_tests were for an older implementation before we had PBCs. Unfortunately PBCs remove the rotational invariance of potential energy function for the nonbonded potential. (Bonded terms do not consider PBCs, ever, however, the exclusions *do* consider the PBCs). So this test needs to be re-written in a slightly more clever way.

The old_tests weren't really being run, and be done purely on the jax layer. So I'll probably add these back in later on.

old_tests/test_bar.py
proteneer: I think we should look into it. We can either test the numerical TI calculation itself, or define a custom vjp that "differentiates under the integral".

proteneer:
I'm going to merge this - and let's address the issues in a separate PR.

PR #325, opened 2021-01-22
Update multi-node example `rbfe_single.py`

maxentile:
Not yet ready to merge.

Attempting to update `examples/rbfe_single.py` and `examples/rhfe_single.py` to use the `CUDAPoolClient` introduced in https://github.com/proteneer/timemachine/pull/323, first focusing on `examples/rbfe_single.py`.

Current issue: although the example starts running, it's hanging / behaving as if each of the calls to `rfe.host_edge` are never returning.

### DIFF
--- a/examples/rbfe_single.py
+++ b/examples/rbfe_single.py
@@ -27,6 +27,12 @@ def convert_uIC50_to_kJ_per_mole(amount_in_uM):
     return 0.593*np.log(amount_in_uM*1e-6)*4.18
 
 
+def wrap_method(args, fxn):
+    # TODO: is there a more functools-y approach to make
+    #   a function accept tuple instead of positional arguments?
+    return fxn(*args)
+
+
 def run_epoch(client, ff, mol_a, mol_b, core):
     # build the protein system.
     complex_system, complex_coords, _, _, complex_box, _ = builders.build_protein_system('tests/data/hif2a_nowater_min.pdb')
@@ -70,8 +76,9 @@ def run_epoch(client, ff, mol_a, mol_b, core):
 
         futures = []
         for arg in host_args:
-            fut = client.submit(rfe.host_edge, arg)
-            futures.append(fut)
+            do_work = functools.partial(wrap_method, fxn=rfe.host_edge)
+            future = client.submit(do_work, arg)
+            futures.append(future)
 
         results = []
         for fut in futures:
--- a/examples/rbfe_single.py
+++ b/examples/rbfe_single.py
@@ -29,7 +29,7 @@ def convert_uIC50_to_kJ_per_mole(amount_in_uM):
 
 def run_epoch(client, ff, mol_a, mol_b, core):
     # build the protein system.
-    complex_system, complex_coords, _, _, complex_box = builders.build_protein_system('tests/data/hif2a_nowater_min.pdb')
+    complex_system, complex_coords, _, _, complex_box, _ = builders.build_protein_system('tests/data/hif2a_nowater_min.pdb')
     complex_box += np.eye(3)*0.1 # BFGS this later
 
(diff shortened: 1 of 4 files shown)


examples/rbfe_single.py
```diff
@@ -183,8 +191,7 @@ def run_epoch(ff, mol_a, mol_b, core):
 
     cmd_args = parser.parse_args()
 
-    multiprocessing.set_start_method('spawn') # CUDA runtime is not forkable
```
proteneer, line 186: @maxentile You should not remove this. This also affects the launch mode of ProcessPoolExecutors. It will default to 'fork' which is incompatible with CUDA, and as it turns out, jax.

maxentile: Ahh, thank you for catching this! Would this be okay to move into the `ProcessPoolClient` or `CUDAPoolClient` constructor then? (I had mistakenly assumed it was safe to remove from this file because there was no longer a direct call to any other multiprocessing function defined in the script.)

proteneer: I think that's fine, an less intrusive alternative would be to assert in the constructor of the PoolClients:
```
multiprocessing.get_start_method() == 'spawn'
```
This way the end-user is still conscientious of the start mode. Generally speaking, library code that set globally hidden singletons is a bad idea, and these types of decisions (like setting jax's precision) should be done immediately and explicitly at the start of the application script. Trying to debug this later on (because some other code might assume forking) will be a goddamn nightmare. Some other fun bits: on windows the default start mode is actually already 'spawn', but on linux for some reason the default mode 'fork'.

maxentile:
(Note that the [`mp_context` argument was only exposed in 3.7](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor), so [the latest commit](https://github.com/proteneer/timemachine/pull/325/commits/f506a62a7b0bd7da833b2bb5ca768f24b090c232) is incompatible with Python 3.6)

proteneer:
@maxentile I removed the jank_executor and added a test specifically for jax as well.

proteneer:
Looks like the tests are still failing and @tony-garcia reverted the 3.7 commit - I'm going to restore the jank_executor and let you guys figure it out.

examples/rbfe_single.py
```diff
@@ -71,10 +72,17 @@ def run_epoch(ff, mol_a, mol_b, core):
         # solvent leg
         host_args = []
         for lambda_idx, lamb in enumerate(lambda_schedule):
-            gpu_idx = lambda_idx % cmd_args.num_gpus
-            host_args.append((gpu_idx, lamb, host_system, minimized_host_coords, host_box, cmd_args.num_equil_steps, cmd_args.num_prod_steps))
-        
-        results = pool.map(functools.partial(wrap_method, fn=rfe.host_edge), host_args, chunksize=1)
+            host_args.append((lamb, host_system, minimized_host_coords, host_box, cmd_args.num_equil_steps, cmd_args.num_prod_steps))
```
proteneer, line 75: Last nitpick: we should simplify this into:

``` python
futures = []
for lambda_idx, lamb in enumerate(lambda_schedule):
    futures.append(client.submit(rfe.host_edge, lamb, host_system, minimized_host_coords, host_box, cmd_args.num_equil_steps, cmd_args.num_prod_steps))
```

proteneer (changes requested):

proteneer (approved):
LGTM

examples/rbfe_single.py
maxentile: Reflected in https://github.com/proteneer/timemachine/pull/325/commits/b0374bd313e34de2246359efbe8273e316735d6f

PR #327, opened 2021-01-28
Fix gte sign for k_check_rebuild_box

proteneer:
This PR fixes an off by one error. It will not affect the correctness of previous results, only the speed, as the worst case behavior is triggering and extra nblist rebuild.


maxentile (approved):
Good catch!

PR #328, opened 2021-01-28
[DNM] Testing updated Jenkins script

maxentile:
Updated Python 3.6 -> 3.7  on Jenkins server with @tony-garcia.

(Do not merge -- pointing to a temporary version of fabric file)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,7 +3,7 @@ pipeline {
   stages {
     stage('Execute CI') {
       steps {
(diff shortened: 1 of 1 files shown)


maxentile:
It's alive!

proteneer:
awesome! Does this mean we can remove the jank_executor?

maxentile:
I believe so!

([The issue](https://github.com/proteneer/timemachine/pull/325#issuecomment-766921015) was that passing each `ProcessPoolClient` its own multiprocessing context required a version-specific Python feature that was introduced between Python 3.6 and 3.7. The `jank_executor` was a workaround to incorporate that feature while also keeping the Jenkins tests running on Python 3.6. We have now upgraded the Jenkins test runner to use Python 3.7, so it should be safe to remove this workaround.)

PR #329, opened 2021-01-29
Remove jank_executor and upgrade to python 3.7

proteneer:
1) We add a note requiring python 3.7 in README
2) Remove jank executor
3) Remove the notice requiring DBUILD_TEST=ON for  CMake


maxentile (approved):
Looks good!

PR #330, opened 2021-01-29
[WIP] RBFE optimization

proteneer:
Edit - okay this PR turned into an unfortunate mega PR:
- [x] Drastically improve performance of code close to that of the non-alchemical system
- [x] Pushes lambda argument into the bonded potentials
- [x] Updates topology handlers to generate a single fused set of potentials
- [x] Gets rid of the manual VJP system entirely  
- [x] Separates preparation of the system from execution of the system
- [x] Implements a differentiable estimator.py::deltaG function that allows jax to do the chain rule
- [x] Fixes a super annoying pickle bug with the InterpolatedPotential class (__getattr__).

This PR attempts to address some of the issues seen #326. So far the diagnosis seems to be:

We're reaching the limit of atomicAdd collisions, as ```k_reduce_add_force_buffer``` in lambda_potential.cu is taking up close to 9% of the total runtime. Part of this is because for every small molecule bonded term, its split into 3 parts (lhs, core, rhs) to implement the linear rescaling. It also dramatically increases the number of kernels required to deal with the extra buffering. 

So we're going to push the ```(1-lamb)*U0 and lamb*U1``` to the inner loop to see if it helps. This is backwards compatible, so we can modify the combining code for the SingleTopology class in a separate pass if needed.

i.e. This allows for U_bonded(x, lamb, params) to be a proper function of lambda. Where as before we relied on the LambdaPotential to do the transformation.

Some technical parts:
1) This will require a upgrade to C++17 (which is fortunately compatible with CUDA10+)
2) Pybind had to be upgrade to support std::optional in order to support default args in not super painful way
3) I will be slowly converting FLOAT_TO_FIXED code

Potentials
- [x] HarmonicBond
- [x] HarmonicAngle
- [x] PeriodicTorsion

PR #331, opened 2021-02-02
Train to multiple RBFEs

maxentile:
This PR builds out from `examples/rbfe_single.py` in the direction of fitting to more than one RBFE label simultaneously.

The main two files to review are:
* `examples/hif2a/generate_star_map.py`
* `examples/hif2a/fit_to_multiple_rbfes.py`

This PR also
* Adds hif2a dataset curated by the authors of https://github.com/MCompChem/fep-benchmark/releases/tag/v1.0 ~(accounting for most of the lines of changes!)~ [update: 2020/02/08 - imported using git submodule instead of copying]
* Modifies the return type of simulation functions in the free energy module, to return the entire trajectory of du/dlambdas for later processing (rather than returning mean(du/dlambda) and stddev(du/dlambda)).
* Adds a script to measure how varying two simulation settings (number of lambda windows, number of MD steps per lambda window) impacts the wall time, precision, and bias of the estimates

Still to-do:
- [ ] Factor out any generally usable convenience classes or functions that were added in the example scripts `examples/hif2a/generate_star_map.py`, `examples/hif2a/fit_to_multiple_rbfes.py`, and add tests for them
- [x] Decide if we should instead use git submodule or something, rather than copying the fep-benchmark files into datasets/
- [ ] Update `examples/estimator_variance.py` for efficiency (generate a single long trajectory and bootstrap resample from it, rather than looping over trajectory lengths and replicates)
- [ ] Investigate and resolve why this [factorizability assertion](https://github.com/proteneer/timemachine/blob/2eb956f9f8ce62287cc531188d1d1481832c5e96/fe/topology.py#L381) is failing for some transformations


examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -0,0 +1,431 @@
...
+def _mean_du_dlambda(result):
+    """summarize result of rfe.host_edge into mean du/dl
+
+    TODO: refactor where this analysis step occurs
+    """
+    bonded_du_dl, nonbonded_du_dl, _ = result
+    return np.mean(bonded_du_dl + nonbonded_du_dl)
+
+
+def _update_combined_handle_and_grads(ghs, combined_handle_and_grads):
```
proteneer, line 147: Looking back, I wish there was a slightly more elegant way to compute these derivatives. One possible solution is to just treat all of this the jax/AD way, i.e. define a custom vjp function on observables of system parameters.

Let ```p``` be forcefield parameters, and ```q(p)``` be system parameters derived from ```p```. Let ```G(q)``` be our free energy estimate. Let ```L(G)``` be our loss function defined relative to some label.

Given:
```
L(G(q(p)))
```
Its derivative is
```
dL/dG . dG/dq . dq/dp
```
We can use jax's vjp system to directly compute outer ```dL/dG``` and ```dq/dp```. However, the inner ```dG/dq``` is done currently using a custom_op of sorts. In theory we can declare ```G(q)``` a primitive and augment it with a defvjp so that the chain rule can be done automatically.

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -0,0 +1,431 @@
...
+        """
+        # unpack result tuple
+        bonded_du_dl, nonbonded_du_dl, grads_and_handles = result
+
+        # save du/dlambda trajectories
+        du_dl_filename = path_to_results.joinpath(
+            f"du_dlambda trajectories ({name}, lambda={lamb:.3f})")
+        print(f'saving bonded and nonbonded du/dlambda trajectories to {du_dl_filename}...')
+        np.savez(du_dl_filename, bonded_du_dl=bonded_du_dl, nonbonded_du_dl=nonbonded_du_dl, rfe=rfe,
+                 configuration=configuration)
```
proteneer, line 183: kudos for saving the configuration!

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -0,0 +1,431 @@
...
+    """
+
+    # build the complex system
+    complex_system, complex_coords, _, _, complex_box, _ = builders.build_protein_system(
+        path_to_protein)
+    # TODO: optimize box
+    complex_box += np.eye(3) * 0.1  # BFGS this later
+
+    # build the water system
+    solvent_system, solvent_coords, solvent_box, _ = builders.build_water_system(4.0)
```
proteneer, line 207: FYI: don't actually know if 4nm is enough or too much here - we may need to revisit this later.

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -0,0 +1,431 @@
...
+    client specifies a multi-node distributed computing client environment
+
+    TODO: add an argument for force field parameters, then register this function with Jax as value_and_grad or similar...
+    """
+
+    # build the complex system
+    complex_system, complex_coords, _, _, complex_box, _ = builders.build_protein_system(
+        path_to_protein)
+    # TODO: optimize box
+    complex_box += np.eye(3) * 0.1  # BFGS this later
```
proteneer, line 204: We definitely will need to fix this - this will affect the densities for sure. A crazy idea/hack without resorting to BFGS is actually that we can find the offending water molecules (not atoms) and just lower them from 4D like the we do for the ligand.

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -0,0 +1,431 @@
...
+# TODO: define more flexible update rules here, rather than update parameters
+step_sizes = {
+    nonbonded.AM1CCCHandler: 1e-3,
+    nonbonded.LennardJonesHandler: 1e-3,
+    # ...
+}
+
+gradient_clip_thresholds = {
+    nonbonded.AM1CCCHandler: 0.001,
+    nonbonded.LennardJonesHandler: np.array([0.001, 0]),  # TODO: allow to update epsilon also?
```
proteneer, line 271: I've never had any success with training with epsilons, but that might be because I don't know what the learning rate should be.

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -0,0 +1,431 @@
...
+
+    The gradient descent update would be
+        update = - step_size * grad(x[k]),
+
+    and to avoid instability, we clip the absolute values of all components of the update
+        update = - clip(step_size * grad(x[k]))
+
+    TODO: menu of other, fancier update functions
+    """
+    return - np.clip(step_size * gradient, -clip_threshold, clip_threshold)
```
proteneer, line 288: In practice, the gradient clipping will basically "binarize" the updates, the relative scales of the gradients are mostly removed.

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -0,0 +1,431 @@
...
+
+    return parameter_updates
+
+
+def _save_forcefield(filename):
+    # TODO: update path
+
+    with open(path_to_results.joinpath(filename), 'w') as fh:
+        fh.write(step_params)
+
```
proteneer, line 378: Note that this is referencing a global variable (defined in the ```__name__ ==  "__main__"```) block, it may be slightly safer/easier to read to just pass this in directly as an argument to the ```_save_forcefield``` function.

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -0,0 +1,431 @@
...
+    # TODO: refactor loss definition out of this function
+    loss = np.abs(pred - label)
+    unit = "kJ/mol"
+
+    message = f"""
+    pred:   {pred:.3f} {unit}
+    target: {label:.3f} {unit} 
+    loss:   {loss:.3f} {unit}
+    """
```
proteneer, line 331: Would it be possible to include timings? Would be useful for us to assess speed/performance later on.

examples/hif2a/generate_star_map.py
```diff
@@ -0,0 +1,264 @@
...
+from ff import Forcefield
+from ff.handlers.deserialize import deserialize_handlers
+
+path_to_ff = str(root.joinpath('ff/params/smirnoff_1_1_0_ccc.py'))
+with open(path_to_ff) as f:
+    ff_handlers = deserialize_handlers(f.read())
+forcefield = Forcefield(ff_handlers)
+
+# 1. Get ligands
+# TODO: possibly Git submodule to fep-benchmark inside datasets/, rather than copying fep-benchmark into datasets
```
proteneer, line 22: I have a love-hate (mostly hate) relationship with git submodule, but there may be times when its appropriate.

examples/hif2a/generate_star_map.py
```diff
@@ -0,0 +1,264 @@
...
+        """
+        x_i = mol1.GetConformer(0).GetPositions()[atom1]
+        x_j = mol2.GetConformer(0).GetPositions()[atom2]
+        return bool(np.linalg.norm(x_i - x_j) <= 0.5)  # must convert from np.bool_ to Python bool!
+
+
+def mcs_map(a, b):
+    """Find the MCS map of going from A to B"""
+    params = rdFMCS.MCSParameters()
+    params.AtomTyper = CompareDist()
```
proteneer, line 69: Note that later on we may want to build compound queries to improve the precision of the MCS.

``` python
mcs_params = rdFMCS.MCSParameters()
mcs_params.AtomTyper = CompareDist()
mcs_params.BondCompareParameters.CompleteRingsOnly = 1 
```

However, we will need to update the RDKit to do so due to a bug I'd filed last year (and has since been resolved)

https://github.com/rdkit/rdkit/issues/3635

examples/hif2a/generate_star_map.py
```diff
@@ -0,0 +1,264 @@
...
+	which is 0 when A = B
+
+    * Alternatives considered but not tried:
+        * min(n_A, n_B) - n_MCS
+        * max(n_A, n_B) - n_MCS
+        * 1.0 - (n_MCS / min(n_A, n_B))
+
+    TODO: move this into a utility module or the free energy module
+    """
+    return (n_A + n_B) - 2 * n_MCS
```
proteneer, line 96: aka the XOR

examples/hif2a/generate_star_map.py
```diff
@@ -0,0 +1,264 @@
...
+        for j, b in enumerate(matches_b):
+            cost[i, j] = np.linalg.norm(conf_a[np.array(a)] - conf_b[np.array(b)], axis=1).sum()
+
+    # find (i,j) = argmin cost
+    min_i, min_j = np.unravel_index(np.argmin(cost, axis=None), cost.shape)
+    print(f'argmin of {n_a} x {n_b} cost matrix: {(min_i, min_j)} ')
+    # TODO: maybe also print the difference between min(cost) and cost[0,0],
+    #   to see how big of a difference it made to pick the default
+
+    # TODO: is there a way to use the matching from MCS directly?
```
proteneer, line 201: Unfortunately no, the MCS code actually constructs a queryMol incrementally now that I think about it. It was designed for the multiple ligand use case (i.e. the FMCS code works on >2 number of ligands).

examples/hif2a/generate_star_map.py
```diff
@@ -0,0 +1,264 @@
...
+from fe.free_energy import RelativeFreeEnergy
+from fe.topology import AtomMappingError
+
+def get_mol_id(mol):
+    return mol.GetProp('ID')
+
+
+def convert_uIC50_to_kJ_per_mole(amount_in_uM):
+    """
+    TODO: move this into a utility module
```
proteneer, line 223: agreed with both TODOs.

examples/train_to_rbfe.py
```diff
@@ -0,0 +1,247 @@
+# This script computes the relative binding free energy of a single edge.
```
proteneer, line 1: We may want to conslidate parts of this with the multi version

ff/handlers/nonbonded.py
```diff
@@ -390,4 +390,8 @@ def static_parameterize(params, smirks, mol):
         deltas = params[bond_idx_params]
         incremented = ops.index_add(am1_charges, bond_idxs[:, 0], deltas)
         decremented = ops.index_add(incremented, bond_idxs[:, 1], -deltas)
-        return decremented
\ No newline at end of file
+
+        q_params = decremented
+        assert (q_params.shape[0] == mol.GetNumAtoms()) # check that return shape is consistent with input mol
```
proteneer, line 395: nice!

examples/hif2a/fit_to_multiple_rbfes.py
proteneer: PS - definitely don't try and actually do this right now. We should save it for another PR after extensive offline discussions.

proteneer (commented):
Did the first pass! Let's talk over some more bits tomorrow.

examples/estimator_variance.py
```diff
@@ -0,0 +1,233 @@
...
+    # num_prod_steps ...
+    num_gpus = 10
+
+    # TODO: can I just get all of the du/dlambda time-series information
+    #   so that we can guess the variance of dG_estimate without having to run
+    #   multiple times from scratch?
+    num_repeats = 5
+
+    multiprocessing.set_start_method('spawn')  # CUDA runtime is not forkable
+    pool = multiprocessing.Pool(num_gpus)
```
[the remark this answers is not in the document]
maxentile: Good point! Addressed in https://github.com/proteneer/timemachine/pull/331/commits/617716111367211908c79f9491639781dbd9d38d

examples/estimator_variance.py
```diff
@@ -0,0 +1,233 @@
...
+        num_steps_per_lambda=1000,
+        num_equil_steps=1000,
+    )
+
+    num_steps_grid = [10,20,30,40,50,100,250,500,750,1000,2500,5000,10000,50000]
+
+    results = dict()
+    times = dict()
+
+    for num_steps in num_steps_grid:
```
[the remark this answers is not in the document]
maxentile: Yes, that is definitely the intent! For now, I will leave this as-is for the purpose of not further delaying / cluttering this PR, but I intend to address this in a subsequent update

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: Saving the `rfe` object in this archive was definitely redundant here, as I ended up storing `2 * n_lambdas` copies of a serialized `rfe` per TI calculation... (These were also stored in a way that required `np.load(..., allow_pickle=True)`, which isn't ideal.)

Since we have already stored the serialized `rfe`'s in an ordered list in the same directory, it suffices to store the integer index of the `rfe` used in each optimizer step, addressed here: https://github.com/proteneer/timemachine/pull/331/commits/666e4cd91808cbfae314bb9ac4262772808b1ef2

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: Multiplying the gradient estimate by a small `step_size` before applying the clipping operation means that the updates will not be binarized as often,  so the relative scales are retained more often.

As discussed off-line, on iterations where the gradient estimate can be flagged as highly unreliable (e.g. if it accompanies a deltaG estimate so large it might as well be NaN), we may also want to set the `step_size` or `clip_threshold` to 0. Setting the `clip_threshold` to a small number safeguards against having occasional nonsense gradients cause immediate training instability.

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: Woops, good catch! This was an oversight -- addressed in https://github.com/proteneer/timemachine/pull/331/commits/f0de0da68ae7b63706f481f9af00e7bbdf5c6567

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/331/commits/89cbff0da504dad1efd20a0cd1c4ad637448b723

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: We have discussed this a bit offline, and @proteneer has implemented this approach in https://github.com/proteneer/timemachine/pull/330 ! I think this represents an immediate improvement, reducing the amount that needs to be managed explicitly in application scripts like the ones developed in the current PR. It also improves our flexibility in prototyping loss functions (and other differentiable computations that accept a free energy estimate as an argument). We should probably merge this tomorrow as far as I'm concerned!

However, I do have a reservation about the approach I would like to bookmark for when we possibly revisit this design down the road.

I think being able to iterate quickly on design choices made in the free-energy-estimation layer represents a large source of overall performance gains, and I worry that this approach might hinder design changes at this layer, to the extent that it "locks in" a specific free-energy-estimation protocol.

The approach I would favor in the long run is to provide accelerated wrappers for the essential, expensive parts of any free energy computation that are deterministic and amenable to autodiff -- keeping any experimental / noisy / biased approximations in prototyping-friendly functions that call these (rather than pushing any of these approximations into custom chain rule code). Specifically, I would want to make sure we have something like "`U(x, lam, params)`, registering the handlers required for `grad(U, argnums=[0,1,2])`, `value_and_grad(U, argnums=[0,1,2])`, and perhaps `vmap(U, argnums=[0, 1])`". Most differentiable free energy estimators I can think of only require to interact with these, and can be implemented in prototyping-friendly (autodiff-able!) Python / Jax in terms of `U` and its derivatives (possibly in accelerator-friendly batches via `vmap`).

Whenever we find that having the free energy estimators written in prototyping-friendly Python / Jax on top of this stable core of functions is too slow, then I think there would be clearer benefit of pushing more of the current best-performing free-energy-estimation pipeline into a primitive function like `G(q)` and accompanying it with custom chain rule code.

proteneer: Some notes on the speed/perf side of things since I've tried to make it fast with pure jax:

1) Nonbonded ```U``` and their derivatives will require a neighborlist for any realistic system with > 1000 atoms.
2) The current reference implementations do not implement a proper neighborlist, since writing a performant neighborlist implemention via jax/numpy is **hard**. We're literally talking about 100x slower. 
3) There's not much to be gained with ```vmap```, since the system sizes are practically so large that a single calculation will probably saturate the GPU.

However, doing a batched variant of the potential functions via a custom_vjp is not difficult, and you could more or less amortize out the cost of the memory allocation.

proteneer: On the more general side of things:

It would not be hard to modify the current ```deltaG``` estimator from one that computes ```xs``` and ```dU/dx, dU/dparams, dU/dlambda``` on the fly to one that takes in a pre-computed ```xs``` and post-processes the ```dU/dparams```. We just need to be cognizant of the GPU RAM requirements when computing ```dU/dparams``` for each ```x``` in ```xs```.

maxentile: Sorry for my unclarity! I didn't mean to suggest reverting to `U(x, lam, params)` or its derivatives in pure Jax (which would incur a significant slow-down for the reasons you mention here, and that you've mentioned before!). I was suggesting to wrap the accelerated implementations of` U` and its derivatives you've already implemented, in a way that I think would allow for simpler access in other Jax application code that calls `U` or its derivatives.

For example, if we were implementing the forward EXP estimator of a free energy difference, I would want the calls to `U(xs, 0.0, params)` and `U(xs, 1.0, params)` to be accelerated in timemachine, but let Jax autodiff machinery handle any subsequent chain-rule stuff involved in actually computing the EXP estimate from the potential energy differences, and in any downstream computations. The approach I thought you were proposing, if applied to this example, would be to implement a Jax primitive `G(params)` that runs a simulation at lam=0.0, performs the reweighting from lam=0.0 to lam=1.0, and implements the chain-rule for logsumexp, all as a single unit.

maxentile:
Currently a bit slowed down debugging this kind of error, which appears for some (but not all) edges

```
The above exception was the direct cause of the following exception:

  File "examples/hif2a/fit_to_multiple_rbfes.py", line 275, in <module>
    loss, loss_grads = binding_estimate_and_grad_fxn(ordered_params, rfe.mol_a, rfe.mol_b, rfe.core, rfe.label)
  File "/home/jfass/miniconda3/envs/timemachine_379/lib/python3.7/site-packages/jax/_src/traceback_util.py", line 139, in reraise_with_filtered_traceback
    return fun(*args, **kwargs)
  File "/home/jfass/miniconda3/envs/timemachine_379/lib/python3.7/site-packages/jax/api.py", line 808, in value_and_grad_f
    ans, vjp_py = _vjp(f_partial, *dyn_args)
  File "/home/jfass/miniconda3/envs/timemachine_379/lib/python3.7/site-packages/jax/api.py", line 1914, in _vjp
    out_primal, out_vjp = ad.vjp(flat_fun, primals_flat)
  File "/home/jfass/miniconda3/envs/timemachine_379/lib/python3.7/site-packages/jax/interpreters/ad.py", line 114, in vjp
    out_primals, pvals, jaxpr, consts = linearize(traceable, *primals)
  File "/home/jfass/miniconda3/envs/timemachine_379/lib/python3.7/site-packages/jax/interpreters/ad.py", line 101, in linearize
    jaxpr, out_pvals, consts = pe.trace_to_jaxpr(jvpfun_flat, in_pvals)
  File "/home/jfass/miniconda3/envs/timemachine_379/lib/python3.7/site-packages/jax/interpreters/partial_eval.py", line 506, in trace_to_jaxpr
    jaxpr, (out_pvals, consts, env) = fun.call_wrapped(pvals)
  File "/home/jfass/miniconda3/envs/timemachine_379/lib/python3.7/site-packages/jax/linear_util.py", line 166, in call_wrapped
    ans = self.f(*args, **dict(self.params, **kwargs))
  File "/home/jfass/Documents/GitHub/timemachine/fe/model.py", line 137, in loss
    pred_ddG = self.predict(ff_params, mol_a, mol_b, core)
  File "/home/jfass/Documents/GitHub/timemachine/fe/model.py", line 85, in predict
    unbound_potentials, sys_params, masses, coords = rfe.prepare_host_edge(ff_params, host_system, min_host_coords)
  File "/home/jfass/Documents/GitHub/timemachine/fe/free_energy.py", line 215, in prepare_host_edge
    final_params, final_potentials = self._get_system_params_and_potentials(ff_params, hgt)
  File "/home/jfass/Documents/GitHub/timemachine/fe/free_energy.py", line 51, in _get_system_params_and_potentials
    combined_params, combined_potential = fn(*params)
  File "/home/jfass/Documents/GitHub/timemachine/fe/topology.py", line 119, in parameterize_periodic_torsion
    guest_params, guest_potential = self.guest_topology.parameterize_periodic_torsion(proper_params, improper_params)
  File "/home/jfass/Documents/GitHub/timemachine/fe/topology.py", line 834, in parameterize_periodic_torsion
    proper_params, proper_potential = self.parameterize_proper_torsion(proper_params)
  File "/home/jfass/Documents/GitHub/timemachine/fe/topology.py", line 825, in parameterize_proper_torsion
    return self._parameterize_bonded_term(ff_params, self.ff.pt_handle, potentials.PeriodicTorsion)
  File "/home/jfass/Documents/GitHub/timemachine/fe/topology.py", line 807, in _parameterize_bonded_term
    raise (RuntimeError(message))
RuntimeError: len(core_idxs_a) != len(core_idxs_b)
                
                core_idxs_a.shape:     (107, 4)
                core_idxs_b.shape:     (103, 4)
                unique_idxs_r.shape:   (2, 4)
                
                core_params_a.shape:   (107, 3)
                core_params_b.shape:   (103, 3)
                unique_params_r.shape: (2, 3)
```

datasets/readme.md
```diff
@@ -0,0 +1,9 @@
+## `freesolv`
+`freesolv.sdf` contains a snapshot of the FreeSolv database of hydration free energies curated by David Mobley, Peter Guthrie, and colleagues here: https://github.com/MobleyLab/FreeSolv
+
+TODO(YTZ) : note which version of FreeSolv is stored, and if any other filtering or processing has been applied
```
proteneer, line 4:
```
15:35 $ git log
commit ff0961a3177863c8002e8487ff9718c513974138 (HEAD -> master, origin/master, origin/HEAD)
Merge: 9d8a809 e332a44
Author: David L. Mobley <l5a9i0q4q3p4w2q5@mobleylab.slack.com>
Date:   Fri Jan 26 10:38:04 2018 -0800
```

datasets/readme.md
```diff
@@ -0,0 +1,9 @@
+## `freesolv`
+`freesolv.sdf` contains a snapshot of the FreeSolv database of hydration free energies curated by David Mobley, Peter Guthrie, and colleagues here: https://github.com/MobleyLab/FreeSolv
+
+TODO(YTZ) : note which version of FreeSolv is stored, and if any other filtering or processing has been applied 
+
+## `fep-benchmark`
+Datasets in the `fep-benchmark` folder are imported from https://github.com/MCompChem/fep-benchmark/releases/tag/v1.0 (distributed under MIT license), created by Christina Schindler and Daniel Kuhn, Merck KGaA, Darmstadt, Germany, December 2018 , accompanying https://chemrxiv.org/articles/preprint/Large-Scale_Assessment_of_Binding_Free_Energy_Calculations_in_Active_Drug_Discovery_Projects/11364884 (Schindler et al., 2019)
+
+Note: This folder may appear empty when you clone the timemachine repository. Clone the repository using `--recursive` flag, or update the dataset submodules after cloning using `git submodule update --init`.
```
proteneer: After offline discussion let's just vendor it and remove this line.

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -0,0 +1,308 @@
...
+        complex_schedule=construct_lambda_schedule(configuration.num_complex_windows),
+        solvent_system=solvent_system,
+        solvent_coords=solvent_coords,
+        solvent_box=solvent_box,
+        solvent_schedule=construct_lambda_schedule(configuration.num_solvent_windows),
+        equil_steps=configuration.num_equil_steps,
+        prod_steps=configuration.num_prod_steps,
+    )
+
+    import jax
```
proteneer, line 261: Did you want to import this here?

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -0,0 +1,308 @@
...
+
+    # in each optimizer step, look at one transformation from relative_transformations
+    for step, rfe_ind in enumerate(step_inds):
+        rfe = relative_transformations[rfe_ind]
+
+        # compute a step, measuring total wall-time
+        t0 = time()
+
+        # TODO: delete when done debugging
+        loss_0 = binding_model.loss(ordered_params, rfe.mol_a, rfe.mol_b, rfe.core, rfe.label)
```
proteneer, line 277: Did you want to remove this?

examples/train_to_rbfe.py
proteneer: How does this differ from examples/rbfe_single.py?

fe/topology.py
```diff
@@ -450,30 +450,36 @@ def __init__(self, mol_a, mol_b, core, ff):
 
         self.assert_factorizability()
 
-    def assert_factorizability(self):
+    def _identify_offending_core_indices(self):
+        """Identifies atoms involved in violations of a factorizability assumption,
```
proteneer, line 454: nice!

proteneer (approved):
lgtm save for some minor comments. Feel free to merge when ready.

datasets/readme.md
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/331/commits/67f52752af14ed8e18a39f9c1d7bdba96d6be750

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: Oops, yes! Resolved in https://github.com/proteneer/timemachine/pull/331/commits/9dacdd8384c6b4678a2449715eaf8ba3a47a1fa0

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/331/commits/9dacdd8384c6b4678a2449715eaf8ba3a47a1fa0

datasets/readme.md
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/331/commits/0b3617ad14baf0486020c94c128f88fabbcee6e3

maxentile:
Thanks Yutong!

For some reason commenting appears disabled on one of your comments:

>How does this differ from examples/rbfe_single.py?

It's outdated and redundant, and has now been removed in https://github.com/proteneer/timemachine/pull/331/commits/c8180a6dbc28e744a72d74161cac95e42cf820a4

Issue #266, opened 2020-09-18
Installation troubleshooting


maxentile:
Thank you for making this project open source. I look forward to contributing to this project in the near future.

Documenting here a few installation hiccups I've encountered, and some possible workarounds. If these workarounds sound okay, I intend to open a small PR in the next few days reflecting these changes.

Following these build instructions: https://github.com/proteneer/timemachine#running-tests 

- [ ] Problem: jaxlib XLA extension has a GLIBC incompatibility on the system where I'm attempting to build the project. (`ImportError: /lib64/libm.so.6: version 'GLIBC_2.23' not found (required by .../envs/timemachine/lib/python3.7/site-packages/jaxlib/xla_extension.so)`)
  * Workaround: downgrade Python version, as suggested in https://github.com/tensorflow/tensorflow/issues/28660 . (Observation: This GLIBC incompatibility is encountered in a fresh python 3.7.3 environment `conda create -n timemachine`, but not in a fresh python 3.6.8 environment `conda create -n timemachine_368 python=3.6.8`)
  * Possible solution: specify compatible python versions

- [ ] Problem: cmake error encountered while executing `cmake -DBUILD_TEST=ON -DCUDA_ARCH=sm_70 ../`
    ```
    CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
    CMake 3.5 or higher is required. You are running version 2.8.12.2
    ```
  * Workaround: `pip install cmake --upgrade`
  * Possible solution: specify minimum `cmake` version in requirements.txt

- [ ] Problem: namespace errors encountered while executing `make -j4 install` : `.../timemachine/timemachine/cpp/src/nonbonded.cu(29): error: namespace "std" has no member "runtime_error"`
   * Workaround: add `#include <stdexcept>` to each of `timemachine/cpp/src/electrostatics.cu, timemachine/cpp/src/harmonic_angle.cu, timemachine/cpp/src/harmonic_bond.cu, timemachine/cpp/src/lennard_jones.cu, timemachine/cpp/src/nonbonded.cu, timemachine/cpp/src/periodic_torsion.cu, timemachine/cpp/src/restraint.cu, timemachine/cpp/src/stepper.cu`

- [ ] Problem: `ModuleNotFoundError`s when executing pytests: `timemachine.lib`, `rdkit`, `simtk` not found
  * Workaround: Append `path/to/timemachine` to `sys.path`, and `conda install -c rdkit rdkit`, `conda install -c omnia openmm`.
[...]

PR #334, opened 2021-02-03
Update CMakeLists to build python 3.7 wrappers

proteneer:
How did our python CI tests ever work? It seems like we were building 3.6 wrappers

### DIFF
--- a/timemachine/cpp/CMakeLists.txt
+++ b/timemachine/cpp/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(PYBIND11_PYTHON_VERSION 3.7 CACHE STRING "Which version of python we're building wrappers against")
 
--- a/timemachine/cpp/CMakeLists.txt
+++ b/timemachine/cpp/CMakeLists.txt
@@ -1,12 +1,12 @@
 cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(PYBIND11_PYTHON_VERSION 3.7 CACHE STRING "Which version of python we're building wrappers against")
+
 project(timemachine LANGUAGES CXX CUDA)
 
-find_package(PythonInterp 3.6 REQUIRED)
-find_package(PythonLibs 3.6 REQUIRED)
-
-set(CUDA_ARCH sm_70 CACHE STRING "Specific CUDA architecture we're targetting.")
+find_package(PythonInterp 3.7 REQUIRED)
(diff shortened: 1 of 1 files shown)


proteneer:
Looking back at an older build, we were indeed building the 3.7 wrappers successfully.

```
out: -- Installing: /home/ubuntu/jenkins/timemachine/8bfb20/PR-331/timemachine/lib/custom_ops.cpython-37m-x86_64-linux-gnu.so
```
It looks like it was because the CMakeFlags were merely specifying the minimum version:

```
out: -- Found PythonInterp: /home/ubuntu/anaconda3/envs/timemachine_379/bin/python (found suitable version "3.7.9", minimum required is "3.6") 
out: -- Found PythonLibs: /home/ubuntu/anaconda3/envs/timemachine_379/lib/libpython3.7m.so (found suitable version "3.7.9", minimum required is "3.6") 
```

proteneer:
I'd still like to get this merged in though, as the new CUDAPoolClient now requires 3.7.

maxentile:
>How did our python CI tests ever work? It seems like we were building 3.6 wrappers
>Looking back at an older build, we were indeed building the 3.7 wrappers successfully.

The readme instructions and the build script have gotten out of sync in at least a couple ways.

To inform CMake of the desired Python in our conda environment on the build server, @tony-garcia and I updated the Jenkins script to use the following CMake command, which differs from the one listed in the readme by the addition of environment-specific `DPYTHON_LIBRARY`, `DPYTHON_INCLUDE_DIR`, and `DPYTHON_EXECUTABLE` flags:

```python
conda_env = "timemachine_379"
conda_path = "/home/ubuntu/anaconda3/envs/{}".format(conda_env)
python_version = "3.7m"

cmake_command = "cmake -DBUILD_TEST=ON -DCUDA_ARCH=sm_70 -DPYTHON_LIBRARY={}/lib/libpython{}.so -DPYTHON_INCLUDE_DIR={}/include/python{}/ -DPYTHON_EXECUTABLE={}/bin/python ../".format(
    conda_path, python_version, conda_path, python_version, conda_path)

print(cmake_command)
# cmake -DBUILD_TEST=ON -DCUDA_ARCH=sm_70 -DPYTHON_LIBRARY=/home/ubuntu/anaconda3/envs/timemachine_379/lib/libpython3.7m.so -DPYTHON_INCLUDE_DIR=/home/ubuntu/anaconda3/envs/timemachine_379/include/python3.7m/ -DPYTHON_EXECUTABLE=/home/ubuntu/anaconda3/envs/timemachine_379/bin/python ../
```

Should we add a note in the readme installation instructions with these flags? The CMake line in the readme assumes we're building with the system Python in mind, but the Jenkins build script assumes we're building with a conda environment in mind.

Another point of difference between the readme instructions and what's run on Jenkins is that, in order to run the tests, we also need a few "heavy" dependencies that we don't want to list in requirements.txt, for reasons you described here: https://github.com/proteneer/timemachine/issues/266#issuecomment-695115687 . When the tests are run on Jenkins, these are assumed to be already installed in the build environment.

In any case, we should definitely have updated the CMakeFlags to enforce the minimum Python version -- thanks for catching and fixing that!

maxentile (approved):

proteneer:
Thanks for the detailed explanation. i think simply mentioning that Python 3.7 is required in the README is sufficient. I was just a little surprised.

PR #336, opened 2021-02-05
Minimize host against both molecules

proteneer:
This PR should fix some of the implosions that @maxentile was seeing. The issue was that minimizing the host w.r.t. to only ligand A doesn't resolve clashes in ligand B. This is why we were mainly seeing implosions at lambda=1.0 but not at lambda=0.0

The question is, does minimizing w.r.t. A then B result in the same host conf as minimizing w.r.t. B then A? If not, we should probably use  joint minimization approach where A and B are minimized simultaneously with exclusions between A and B.

### DIFF
--- a/fe/model.py
+++ b/fe/model.py
@@ -76,6 +76,7 @@ class RBFEModel():
 
             print("Minimizing the host structure to remove clashes.")
             min_host_coords = minimizer.minimize_host_4d(mol_a, host_system, host_coords, self.ff, host_box)
+            min_host_coords = minimizer.minimize_host_4d(mol_b, host_system, min_host_coords, self.ff, host_box)
 
             rfe = free_energy.RelativeFreeEnergy(mol_a, mol_b, core, self.ff)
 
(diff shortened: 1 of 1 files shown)


maxentile (approved):
Great catch!

>The question is, does minimizing w.r.t. A then B result in the same host conf as minimizing w.r.t. B then A? If not, we should probably use joint minimization approach where A and B are minimized simultaneously with exclusions between A and B.

My guess would be "no" in general, since the minimizations aren't globally converged, but my guess is that any starting-configuration dependent "hysteresis" this step introduces between `RFE(a, b)` and `RFE(b, a)` would be small relative to other sources of bias and run-to-run variability.

I'd say we should incorporate the proposed fix for now, adding a to-do note to revisit and implement a more symmetric initialization scheme (like joint minimization) if needed.

proteneer:
> I'd say we should incorporate the proposed fix for now, adding a to-do note to revisit and implement a more symmetric initialization scheme (like joint minimization) if needed.

Sounds good to me. I added a note and will revisit later.

maxentile:
Another thing that could be relatively cheap and informative, if we're ever suspicious about this step in the future, is to **run the minimizations in both orders** ("a then b" vs. "b then a") and print a user warning if the resulting geometries (or maybe just energies?) are substantially different.

PR #341, opened 2021-02-05
Make the minimizer more robust

proteneer:
This PR:

1) Modifies the minimizer to take in a list of mols, currently supporting length 1 and length 2, but fixes the API so that we can in theory add multiple ligands for simultaneous docking
2) length == 2 is conveniently taken care of by the DualTopology class, which automatically adds exclusions between the two ligands
3) Removes the thermal noise by override the cc coefficient in the langevin equation so that the minimization is deterministic, as this made bug tracking rather difficult last night
4) Adds a final sanity assertion


tests/test_minimizer.py
```diff
@@ -0,0 +1,27 @@
...
+    mol_a = all_mols[1]
+    mol_b = all_mols[4]
+
+
+    ff = Forcefield(deserialize_handlers(open('ff/params/smirnoff_1_1_0_ccc.py').read()))
+
+    # these methods will throw if the minimization failed
+    minimized_coords = minimizer.minimize_host_4d([mol_a, mol_b], complex_system, complex_coords, ff, complex_box)
+    minimized_coords = minimizer.minimize_host_4d([mol_a], complex_system, complex_coords, ff, complex_box)
+    minimized_coords = minimizer.minimize_host_4d([mol_b], complex_system, complex_coords, ff, complex_box)
```
maxentile, line 27: Minor possible improvement: We could have each of these three calls act as a separate pytest, using pytest.mark.parameterize on the first argument.

md/minimizer.py
```diff
@@ -15,15 +15,15 @@ def get_romol_conf(mol):
     return guest_conf/10 # from angstroms to nm
 
 
-def minimize_host_4d(romol, host_system, host_coords, ff, box):
+def minimize_host_4d(mols, host_system, host_coords, ff, box):
     """
     Insert romol into a host system via 4D decoupling under a Langevin thermostat.
```
maxentile, line 20: Nitpick: using thermostat should keep system at constant temperature, but minimization should effectively bring system to zero temperature

fe/topology.py
```diff
@@ -375,11 +373,20 @@ def parameterize_harmonic_bond(self, ff_params):
...
+    def parameterize_periodic_torsion(self, proper_params, improper_params):
+        """
+        Parameterize all periodic torsions in the system.
+        """
+        proper_params, proper_potential = self.parameterize_proper_torsion(proper_params)
+        improper_params, improper_potential = self.parameterize_improper_torsion(improper_params)
+        combined_params = jnp.concatenate([proper_params, improper_params])
+        combined_idxs = np.concatenate([proper_potential.get_idxs(), improper_potential.get_idxs()])
+        combined_potential = potentials.PeriodicTorsion(combined_idxs)
+        return combined_params, combined_potential
```
maxentile, line 385: Confirming if this is part of the current PR?

maxentile (changes requested):
Looks great to me! 

>Removes the thermal noise by override the cc coefficient in the langevin equation so that the minimization is deterministic, as this made bug tracking rather difficult last night

I think there's another preferable way to make the minimization deterministic (temperature to zero, rather than friction to zero)

fe/topology.py
proteneer: yes - I forgot to add this for the DualTopology class

md/minimizer.py
```diff
@@ -70,15 +81,21 @@ def minimize_host_4d(romol, host_system, host_coords, ff, box):
...
         300.0,
         1.5e-3,
         1.0,
         combined_masses,
         seed
-    ).impl()
+    )
+
+    # remove the randomness completely
+    intg.ccs = np.zeros_like(intg.ccs)
```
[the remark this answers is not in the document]
proteneer: ccs isn't the friction coefficient, but its derived from it the cas, cbs, ccs are computed via:

``` python
# timemachine/integrator.py

def langevin_coefficients(
    temperature,
    dt,
    friction,
    masses):
    """
    Compute coefficients for langevin dynamics

    Parameters
    ----------
    temperature: float
        units of Kelvin

    dt: float
        units of picoseconds

    friction: float
        frequency in picoseconds

    masses: array
        mass of each atom in standard mass units

    Returns
    -------
    tuple (ca, cb, cc)
        ca is scalar, and cb and cc are n length arrays
        that are used during langevin dynamics

    """
    vscale = np.exp(-dt*friction)
    if friction == 0:
        fscale = dt
    else:
        fscale = (1-vscale)/friction
    kT = BOLTZ * temperature
    nscale = np.sqrt(kT*(1-vscale*vscale)) # noise scale
    invMasses = 1.0/masses
    sqrtInvMasses = np.sqrt(invMasses)

    ca = vscale
    cb = fscale*invMasses
    cc = nscale*sqrtInvMasses
    return ca, cb, cc
```

So there's still drag

tests/test_minimizer.py
proteneer: Cool! I need to look up the documentation on this, didn't know about it

md/minimizer.py
proteneer: Agreed - this isn't accurate if we set ccs=0 and I need to update the docstring

maxentile (approved):

md/minimizer.py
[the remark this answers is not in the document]
maxentile: Ahh, thanks for clarifying! I had mis-remembered this. Marking resolved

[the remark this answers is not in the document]
maxentile: It may still be clearer to set the temperature to 0 in the integrator constructor, rather than overriding these coefficients, but the effect should be the same

tests/test_minimizer.py
proteneer: okay I actually tried this, but for this particular use case it's a little tricky since loading in the values as an arg to parameterize was difficult, and using idxs requires too much janking around broadcasting (since python lists aren't broadcastable). I think it's probably clear/easier to read as-is.

```

import numpy as np

from rdkit import Chem

from md import minimizer, builders

from ff.handlers.deserialize import deserialize_handlers
from ff import Forcefield

import pytest

@pytest.mark.parametrize("mol_idxs", [[0, 1], [0], [1]])
def test_minimizer(mol_idxs):

    complex_system, complex_coords, _, _, complex_box, _ = builders.build_protein_system('tests/data/hif2a_nowater_min.pdb')
    complex_box += np.eye(3)*0.1 # BFGS this later

    suppl = Chem.SDMolSupplier('tests/data/ligands_40.sdf', removeHs=False)
    all_mols = [x for x in suppl]
    # mol_a = all_mols[1]
    # mol_b = all_mols[4]


    ff = Forcefield(deserialize_handlers(open('ff/params/smirnoff_1_1_0_ccc.py').read()))

    # these methods will throw if the minimization failed
    # can't easily broadcast this
    minimized_coords = minimizer.minimize_host_4d(all_mols[mol_idxs], complex_system, complex_coords, ff, complex_box)
`````

File: md/minimizer.py (hunk 4 above)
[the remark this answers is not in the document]
Comment (proteneer) [reply md/minimizer.py:96]: Good call! 
```
>>> integrator.langevin_coefficients(300.0, 1.5e-3, 1.0, np.array([1.0]))
(0.9985011244377109, array([0.00149888]), array([0.08643984])) # ca cb cc
>>> integrator.langevin_coefficients(0.0, 1.5e-3, 1.0, np.array([1.0]))
(0.9985011244377109, array([0.00149888]), array([0.])) # ca cb cc
```

I like this better

PR #346, opened 2021-02-10
Improved FF parameter optimization

maxentile:
Exploring a couple routes for improving the force field parameter optimization logic. As a placeholder, we had been using SGD with clipped gradients, [with different clipping thresholds for each parameter type](https://github.com/proteneer/timemachine/blob/12bcc2e8d590ea6eded7124aa316ce14a6bfc703/examples/rbfe_single.py#L123-L141).

This was motivated in part by (1) the observation that the gradient magnitudes for some parameter types were much larger than for others, and (2) the goal of "stabilizing" the optimization, so that we don't jump to points in FF parameter space that would cause immediate simulation instability.

This PR
- [x] Views the result of binding model value_and_grad as providing a local Taylor approximation `f_prime` of the loss function `f`. Knowing that `f` we're optimizing can't get lower than 0, it becomes natural to truncate a proposed optimizer step if the local approximation is saying `f_prime(x_proposed) < 0`. This move is directly influenced by https://arxiv.org/abs/1903.08619 .
- [x] Replaces the global bound 0 with a more conservative heuristic: `0.95 * current_loss`. I.e. try to take a step that's projected to decrease the loss by 5% at most. This will have the nice effect of making the steps smaller when the loss is already close to 0. However, the 0.95 threshold was something I just made up -- we could tune this later, or find another heuristic that doesn't require to tune this threshold.
- [x] Adds a parameter flattening / unflattening utility, to allow the gradient-processing logic to be more disentangled from the ff_handlers structure.
- [x] Saves trajectories of loss, flat parameters, flat gradients to a `.npz` archive.
- [x] Adds type hints in the current example file. (Can remove if they hinder readability.)
- [x] Removes parameter type specific gradient clipping thresholds. (In particular, it relaxes the part that clipped LJ epsilon parameter updates to 0!)
-----------
In the future, we could consider further improvements obtained by:
* replacing queries to this single local surrogate with a hierarchy of surrogates considered in OpenFF property-estimator and related projects (esp. reweighting-based surrogates)
* using a smarter search direction than the raw -gradient (incorporating knowledge of the previous iterates (as in BFGS or "momentum"), or computing a cheap approximation of the Newton direction (as in Newton-CG))
* incorporating more elaborate trust-region definitions
-----------
Some changes that might be nice to make in the current PR:
- [x] Move gradient-processing functions out of example script into a new module `timemachine/optimizers` or similar?
- [ ] Try to make the steps less conservative?
- [ ] Apply same optimizer approach to multi-edge training?
- [ ] [`flatten`](https://github.com/HIPS/autograd/blob/master/autograd/misc/flatten.py) was a really nice feature of HIPS/autograd , but I'm not sure if it lives on somewhere in Jax. Does Jax have a similar function? If so, I would like to delete the model-specific flatten/unflatten function I added here.

-----------
Plotting the results of running the updated script for a few hours with the following settings: `python examples/rbfe_single.py --num_gpus 10 --num_complex_windows 60 --num_solvent_windows 60 --num_equil_steps 10000 --num_prod_steps 10000`

rbfe_single_loss
rbfe_single_parameter_traj
rbfe_single_parameter_increment

I think this is encouraging, since the total distance traveled in FF parameter space is much smaller than the distance covered by a single optimizer step in the previous approach, and it appears to make the loss small and to "converge" nicely.

I was initially concerned that the optimization is not making gradient norm small, but @proteneer points out that this is actually unsurprising, since we're using L1 loss! (Because loss is abs(pred - target), norm(grad(loss)) = norm(grad(free_energy)) everywhere except where pred = target.)

rbfe_single_gradient_norm


examples/rbfe_single.py
```diff
@@ -15,11 +16,89 @@
...
+    """
+    def f_prime(y: array) -> float:
+        return f_x + np.dot(grad, y - x)
+
+    return f_prime
+
+
+def _smart_clip(
+        x: array, f_x: float, grad: array,
+        step_size: float=0.1, search_direction: Optional[array]=None,
```
proteneer, line 45: When should we use the default values for step_size and step_lower_bound?

examples/rbfe_single.py
```diff
@@ -120,31 +199,102 @@ def wrap_method(args, fxn):
     ordered_params = forcefield.get_ordered_params()
     ordered_handles = forcefield.get_ordered_handles()
 
-    gradient_clip_thresholds = {
-        nonbonded.AM1CCCHandler: 0.05,
-        nonbonded.LennardJonesHandler: np.array([0.001, 0])
-    }
+    handle_types_being_optimized = [AM1CCCHandler, LennardJonesHandler]
```
proteneer, line 202: We should have a way to avoid optimizing a particular component of the parameters, eg. for torsions, params are:

(k, phase, period), but we really dont want to optimize the period.

examples/rbfe_single.py
```diff
@@ -120,31 +199,102 @@ def wrap_method(args, fxn):
...
         print("epoch", epoch, "loss", loss)
 
-        for loss_grad, handle in zip(loss_grad, ordered_handles):
-            assert handle.params.shape == loss_grad.shape
+        # note: unflatten_grad and unflatten_theta have identical definitions for now
+        flat_loss_grad, unflatten_grad = flatten(loss_grad)
+        flat_theta, unflatten_theta = flatten(ordered_params)
+
+        step_lower_bound = loss * relative_improvement_bound
+        theta_increment = _smart_clip(flat_theta, loss, flat_loss_grad, step_lower_bound=step_lower_bound)
```
proteneer, line 264: Why does the smart_clip have to run over a single flattened view over the individual handler's .params? Is it a speed issue?

proteneer: PS if you're confident that a single step_size/lower_bound pair is sufficient across all of our handlers - then maybe this is still okay. It just seems a little unwieldy to flatten -> compute updates -> unflatten -> apply updates

examples/rbfe_single.py
maxentile: Will add some usage notes. `step_size` is used to generate an initial proposal `x_proposed`. If `f_prime(x_proposed) < step_lower_bound`, then the step will be truncated.

The default `step_lower_bound=0` corresponds to a suggestion in the cited study, incorporating the knowledge that the loss is bounded below by 0. In the script, I passed in a non-default argument to the `step_lower_bound` to make the behavior of the method more conservative.

The default value `step_size=0.1` isn't very precisely chosen. The behavior of the method will be insensitive to picking `step_size` anywhere between like 1e-3 and +inf for our problems, since this will trigger the step-truncating logic on most every step. If the `step_size` is chosen sufficiently small that it rarely produces proposals that violate `step_lower_bound`, then that will start to have an effect on the behavior of the optimizer.

examples/rbfe_single.py
maxentile: Not speed, but flexibility. (I think speed of this operation is pretty irrelevant -- the cost of essentially doing a "-=" (or anything up to the order of a few seconds of thinking to decide what to do with the current gradient estimate) is negligible compared with cost of running a free energy calculations.)

I think providing a flat view of the function improves flexibility, and our ability to try different optimization strategies. Generic optimizers typically assume access to a function in terms of queries to `f(x)`, `grad(f)(x)`, ... , where `x` is a flat array. Adding a parameter flattening / unflattening utility is intended to allow the gradient-processing logic to be more disentangled from the ff_handlers structure. The definition of the current step function `_smart_clip`, like the definition of most any optimizer we may import from another library, does not need to know anything about how we're currently organizing ff_handlers, and will not need to be refactored if we decide to update how parameters are stored in ff_handlers. (For example, if we later use a different charge model than `AM1CCC`, or represent torsion terms in a way that associates only one term to each torsion tuple.)

examples/rbfe_single.py
maxentile: Good point -- I should rework this. I haven't checked if the returned gradient w.r.t. discrete parameters like period is currently nan or 0, but we certainly will want to avoid updating discrete parameters like this when we start optimizing valence parameters.

I think a modular way to support masking by parameter type might be to zero out the relevant entries of the `search_direction`...

examples/rbfe_single.py
proteneer: Okay - we can try this approach for now then. Just be aware of our need to

1) sometimes set certain components to zero (eg. periods in torsions), which might be difficult to identify/mark in a flattened view
2) if we ever want to take an approach like forcebalance to convert between mathematical and physical parameters by attaching units. 

but either way, I'm comfortable with letting you make the call on this (with the assumption that you'll be able to figure out a good solution to the above should the need arise!)

examples/rbfe_single.py
proteneer: Thanks - that description definitely helped me out!

proteneer (approved):
Looks good to me

examples/rbfe_single.py
maxentile: Added these clarifications to docstring in https://github.com/proteneer/timemachine/pull/346/commits/d3a31d2e60b7732ab8b254094cefa2cb67835093

maxentile:
To get an initial sense of how much differently the optimizer behaves when provided with d free_energy_loss / d forcefield_params estimates that have lower bias and variance -- but are ~10x more computationally expensive -- here are results of running the same procedure as before, but with the # of production MD steps per lambda window increased from 10,000 to 100,000. (Truncated after about 70 optimizer steps, due to impatience.)

rbfe_single_loss

rbfe_single_parameter_increment

rbfe_single_gradient_norm

Further study is needed here, but this may be a hint that, for this kind of task, early in optimization, we may be able to get away with quite cheap and noisy gradient estimates.

(Later, I imagine it will be productive to apply other variance reduction strategies than brute-force "increase sampling" strategies, but for now I'm encouraged that it appears we can do something here even before taking further pains to reduce the gradient-estimator variance.)

PR #348, opened 2021-02-10
Make (shortened versions of) benchmarks discoverable by pytest

maxentile:
Addressing an issue @proteneer just ran into, where I had broken the `tests/benchmark.py` script by changing the `RelativeFreeEnergy` constructor, but tests were still passing because the benchmark script was not being automatically collected.

To reduce the cost, the benchmarks should perform only a small number of MD steps when collected by pytest. When run via `__main__` as before, the benchmark script should perform the same large number of MD steps as before.

Taking a fresh look at benchmark.py:benchmark, I notice that the `lamb` parameter wasn't being set as probably intended in benchmark_hif2a() -- I haven't corrected this yet, in the interest of preserving the current behavior, but I should probably
- [x] fix lambda parameter for benchmarking apo vs. rbfe speed.


tests/test_benchmark.py
```diff
@@ -33,19 +34,26 @@ def recenter(conf, box):
 
     return np.array(new_coords)
 
-
 def benchmark(
     label,
     masses,
-    lamb,
+    lamb, # TODO: note lamb unused
```
proteneer, line 40: oops - I think for RBFE the intention was for this to be set to 0.5

maxentile: Resolved in https://github.com/proteneer/timemachine/pull/348/commits/f836d025c18667b2caeee21c13dcb6ec12454193

proteneer (approved):
lgtm - feel free to merge once tests are back up and running

maxentile:
Woohoo! Tests are up and running again. This PR does increase the time to run the tests by a bit (~ a second to run MD, and ~ less than a minute to do the minimize_4d and other setup, bringing the total build/test time from ~9:50 to ~10:40). We will probably be revisiting soon if there are ways to trim down the time required to run the test suite.

PR #349, opened 2021-02-11
Revert hif2a benchmark to use _sc forcefield rather than _ccc

maxentile:
Reverts a change introduced in https://github.com/proteneer/timemachine/pull/348/commits/1654551f130541bcfac2b781aba781db90d4987e that inadvertently changed the force field used for hif2a benchmark from _sc to _ccc , which @proteneer reports is causing a portability issue. This should now look the same [as before](https://github.com/proteneer/timemachine/blob/12bcc2e8d590ea6eded7124aa316ce14a6bfc703/tests/benchmark.py#L193-L195).

Later, we should add a way to label when tests aren't allowed to use openeye toolkits

### DIFF
--- a/tests/test_benchmark.py
+++ b/tests/test_benchmark.py
@@ -6,6 +6,8 @@ import time
 import numpy as np
 
 from ff.handlers import openmm_deserializer
+from ff.handlers.deserialize import deserialize_handlers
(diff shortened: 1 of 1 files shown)


proteneer (approved):
lgtm

PR #351, opened 2021-02-19
Selective Execution Improvements

proteneer:
This PR begins the work on selective execution in hopes of improving performance for du_dl computation. This is needed for faster non-equilibrium calculations which require du_dl on every step. For RBFE calculations with interpolated nonbonded kernels, du_dl is chained via:

```
d/dl u(lambda, params(lambda)) = du/dlambda + du/dparams.dparams/dlambda
```

Currently, with the InterpolatedPotential class, we need to compute a per-particle du/dparams (even though the vast majority of them have ```dparams/dlambda=0```, since the host parameters are not interpolated). However, it is not clear how to pass the ```dparams/dlambda``` information to the underlying Nonbonded class, while still keeping the abstraction barriers clear. So this PR modifies the Nonbonded class to be able to directly do parameter interpolation.

In order to improve performance, we note that we generally require 4 outputs, but not all at once:
```
u (rarely needed)
du/dp (needed at the endpoints)
du/dl (always needed for NEQ, and periodically needed for EQ)
du/dx (always needed)
```
Previously, we had a special ```execute_du_dx_only()``` method to indicate that we only needed ```du_dx```. However, with NEQ support coming down the road, we will also need ```du_dl``` in a way that is performant. I really did not want to add a third ```execute_du_dx_and_du_dl_only()``` class, and specialize a third kernel just for this computation. So with some template magic, we pre-generate all 2^4 choices on the kernel level, and trigger specialized paths depending on which of the outputs we need.

So the kernel code is now templatized with:

```
template <
    typename RealType,
    bool ALCHEMICAL,
    bool COMPUTE_U,
    bool COMPUTE_DU_DX,
    bool COMPUTE_DU_DL,
    bool COMPUTE_DU_DP
>
void __global__ k_nonbonded_unified(...)
```
And specialized code is generated for each combination. The ALCHEMICAL flag is used to partition interacting tiles into two sets, one non-alchemical set where we guarantee that ```du_dl==0``` and ```dp_dl==0``` for every particle in the tile, and one alchemical set where we need to compute the full chain. Note that 95% of the interactions will follow the fast, non-alchemical route. 

Some baseline performance numbers, assuming every tile is alchemical, is as follows:

```
   Avg                                    alchem     u     du_dx   du_dl   du_dp
554.34us  void k_nonbonded_unified<float, bool=1, bool=1, bool=1, bool=1, bool=1>
533.58us  void k_nonbonded_unified<float, bool=1, bool=0, bool=1, bool=1, bool=1>
457.73us  void k_nonbonded_unified<float, bool=1, bool=1, bool=1, bool=1, bool=0>
439.58us  void k_nonbonded_unified<float, bool=1, bool=0, bool=1, bool=1, bool=0> <- du_dx and du_dl
426.43us  void k_nonbonded_unified<float, bool=1, bool=1, bool=1, bool=0, bool=1>
403.66us  void k_nonbonded_unified<float, bool=1, bool=1, bool=0, bool=1, bool=1>
393.92us  void k_nonbonded_unified<float, bool=1, bool=0, bool=1, bool=0, bool=1>
389.28us  void k_nonbonded_unified<float, bool=1, bool=0, bool=0, bool=1, bool=1>
332.31us  void k_nonbonded_unified<float, bool=1, bool=1, bool=1, bool=0, bool=0>
316.21us  void k_nonbonded_unified<float, bool=1, bool=1, bool=0, bool=1, bool=0>
296.56us  void k_nonbonded_unified<float, bool=1, bool=0, bool=0, bool=1, bool=0>
256.41us  void k_nonbonded_unified<float, bool=1, bool=0, bool=1, bool=0, bool=0> <- du_dx only
247.96us  void k_nonbonded_unified<float, bool=1, bool=1, bool=0, bool=0, bool=1>
224.96us  void k_nonbonded_unified<float, bool=1, bool=0, bool=0, bool=0, bool=1>
170.47us  void k_nonbonded_unified<float, bool=1, bool=1, bool=0, bool=0, bool=0>
```

Our goal is to bring the performance of ```du_dx and du_dl``` to nearly that of ```du_dx``` only. Note that the ```alchem``` template parameter is an implementation detail.

Another nice side effect of this (hopefully) is that we also get a nice speed up from not having to compute host-host interactions in 3D (even for the du_dx only) case.


maxentile:
This is great! The added flexibility of being able to toggle any of these combinations should also be a huge help down the road for other applications besides the ones you listed, as we need only some of these combinations for certain kinds of post-processing. For example, reweighting requires `u` only (or in some cases `u` and `du_dl` only), which already appears significantly sped up in the table you posted.

proteneer:
```
547.11us  void k_nonbonded_unified<float, bool=1, bool=1, bool=1, bool=1>
524.21us  void k_nonbonded_unified<float, bool=0, bool=1, bool=1, bool=1>
520.10us  void k_nonbonded_unified<float, bool=1, bool=1, bool=0, bool=1>
490.51us  void k_nonbonded_unified<float, bool=0, bool=1, bool=0, bool=1>
398.44us  void k_nonbonded_unified<float, bool=1, bool=1, bool=1, bool=0>
379.36us  void k_nonbonded_unified<float, bool=1, bool=1, bool=0, bool=0>
309.35us  void k_nonbonded_unified<float, bool=1, bool=0, bool=1, bool=1>
298.79us  void k_nonbonded_unified<float, bool=0, bool=1, bool=1, bool=0> <-- du_dx and du_dl
298.50us  void k_nonbonded_unified<float, bool=1, bool=0, bool=0, bool=1>
290.65us  void k_nonbonded_unified<float, bool=0, bool=1, bool=0, bool=0> <-- du_dx only
274.72us  void k_nonbonded_unified<float, bool=0, bool=0, bool=1, bool=1>
268.50us  void k_nonbonded_unified<float, bool=0, bool=0, bool=0, bool=1>
189.92us  void k_nonbonded_unified<float, bool=1, bool=0, bool=1, bool=0>
186.81us  void k_nonbonded_unified<float, bool=1, bool=0, bool=0, bool=0>
```

proteneer:
```
master
hif2a-apo: N=8805 speed: 512.80ns/day dt: 1.5fs (ran 25000 steps in 6.32s)
hif2a-rbfe-with-du-dp: N=8840 speed: 408.32ns/day dt: 1.5fs (ran 25000 steps in 7.94s)
hif2a-rbfe-du-dl-freq-0: N=8840 speed: 413.40ns/day dt: 1.5fs (ran 25000 steps in 7.84s)
hif2a-rbfe-du-dl-freq-1: N=8840 speed: 230.82ns/day dt: 1.5fs (ran 25000 steps in 14.04s)
hif2a-rbfe-du-dl-freq-5: N=8840 speed: 357.52ns/day dt: 1.5fs (ran 25000 steps in 9.06s)

this branch
hif2a-apo: N=8805 speed: 501.13ns/day dt: 1.5fs (ran 25000 steps in 6.47s)
hif2a-rbfe-with-du-dp: N=8840 speed: 416.04ns/day dt: 1.5fs (ran 25000 steps in 7.79s)
hif2a-rbfe-du-dl-freq-0: N=8840 speed: 416.81ns/day dt: 1.5fs (ran 25000 steps in 7.78s)
hif2a-rbfe-du-dl-freq-1: N=8840 speed: 396.17ns/day dt: 1.5fs (ran 25000 steps in 8.18s)
hif2a-rbfe-du-dl-freq-5: N=8840 speed: 412.09ns/day dt: 1.5fs (ran 25000 steps in 7.86s)
```
72% speed-up for hif2a-rbfe-du-dl-freq-1 (computing du/dl on every step), and the overall performance is within 5% of hif2a-rbfe-du-dl-freq-0 (only computing forces!)

proteneer:
List of optimizations implemented in this PR
- Generate nonbonded kernels to compute selectively based on template args that elide unnecessary arguments
- Separates each tile into a fast "non-alchemical" tile, and a slow "alchemical" tile based on doing a fast check at the start of the calculation
- Modifies du_dl computation and du_dp computations to require slightly less floating point operations, in particular, an extra erfc was removed in the du_dl part by re-using a pre-computed number. (erfcs are one of the slowest operations we can do). 
- Modifies the kernel code to always accumulate du_dl and u, which are scalars, to instead a buffered array. The final array is reduced using cub in a single, final pass, as opposed to relying on atomic instructions.

docking/pose_dock.py
```diff
@@ -170,29 +165,34 @@ def pose_dock(
...
-                guest_coords = ctxt.get_x_t()[len(host_conf) :] * 10
-                report.write_frame(host_coords, host_mol, guest_coords, guest_mol, guest_name, outdir, step, 'pd')
-            if step in (0, int(n_steps/2), n_steps-1):
-                if report.too_much_force(ctxt, lamb, box, bps, impls):
-                    calc_work = False
-                    break
+        # (ytz): we gotta figure out how to batch this code, tbd: batch this
+
+        # collect a du_dl calculation every step
+        subsample_freq = 1
```
proteneer, line 171: @schmolly I wasn't sure how to work this in batches - so I just do the check at the last frame. Is this okay?

proteneer:
reviewers:
@maxentile is the primary reviewer
@schmolly is mainly for the docking changes
@badisa is mainly for auditing/learning

docking/pose_dock.py
schmolly: Yep!

tests/common.py
```diff
@@ -493,52 +493,50 @@ def compare_forces(
...
-
-        if isinstance(test_potential, potentials.Nonbonded):
-            np.testing.assert_array_equal(test_du_dp, test_du_dp_2)
-            np.testing.assert_array_equal(test_du_dl, test_du_dl_2)
-            np.testing.assert_array_equal(test_u, test_u_2)
+        for combo in range(2**4):
+            compute_du_dx = combo & 1 << 0
+            compute_du_dp = combo & 1 << 1
+            compute_du_dl = combo & 1 << 2
+            compute_u = combo & 1 << 3
```
maxentile, line 504: Another way to do this, if we prefer itertools-ing to bit-shifting

```suggestion
        from itertools import product

        for combo in product([False, True], repeat=4):
            (compute_du_dx, compute_du_dp, compute_du_dl, compute_u) = combo
```

tests/common.py
```diff
@@ -493,52 +493,50 @@ def compare_forces(
...
+                params,
+                box,
+                lamb,
+                compute_du_dx,
+                compute_du_dp,
+                compute_du_dl,
+                compute_u
+            )
+
+            np.testing.assert_array_equal(test_du_dx, test_du_dx_2)
```
maxentile, line 537: Should we add comment here (or in `execute_selective` docstring) about which of these are deterministic, and approximately how many non-deterministic bits to expect for the others? (Not sure if the new approach to du_dl accumulation is now deterministic, for example.)

tests/test_nonbonded.py
```diff
@@ -247,11 +246,38 @@ def test_benchmark(self):
...
+
+        for combo in range(2**4):
+
+            print("COMBO", combo)
 
-        for _ in range(100):
+            compute_du_dx = combo & 1 << 0
+            compute_du_dp = combo & 1 << 1
+            compute_du_dl = combo & 1 << 2
+            compute_u = combo & 1 << 3
```
maxentile, line 266: Can also use itertools product here

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -208,54 +226,6 @@ void declare_avg_partial_u_partial_param(py::module &m) {
...
-        return new timemachine::FullPartialUPartialLambda(bps, freq);
-    }))
-    .def("full_du_dl", [](timemachine::FullPartialUPartialLambda &obj) -> py::array_t<double, py::array::c_style> {
-        int count = obj.count();
-        py::array_t<double, py::array::c_style> py_full_du_dl({count});
-        obj.full_du_dl(py_full_du_dl.mutable_data());
-        return py_full_du_dl;
-    });
-}
-
```
maxentile, line 257: Double-check for any usages of `.full_du_dl()` this API change might impact in scripts `dock_and_equilibrate.py`, `rigorous_work.py`

maxentile (commented):
Looks great to me so far, and the performance increases are fantastic! Being able to simulate nonequilibrium switches with du_dlambda computed each timestep for the same cost as vanilla MD is a major speed improvement compared with other implementation approaches I'm aware of. Many details remain for me to review: left some first-pass comments and will return soon.

tests/common.py
proteneer: I like your solution way more! I should really learn how to use itertools better

tests/common.py
proteneer: I hadn't realized this but du/dl and u are now fully deterministic, which only leaves du_dp because it still uses double *buffers

tests/test_nonbonded.py
proteneer: willfix

timemachine/cpp/src/wrap_kernels.cpp
proteneer: Good call - let me see if I can use them. @schmolly we should probably put these three files under testing at some point.

proteneer: Fixed

docking/dock_and_equilibrate.py
```diff
@@ -265,7 +245,7 @@ def main():
     parser.add_argument(
         "--max_lambda",
         type=float,
-        default=0.25,
+        default=1.0,
```
proteneer, line 248: @schmolly  Had to change this to 1.0 for the du_dl endpoint checks to pass.

docking/rigorous_work.py
```diff
@@ -375,46 +335,50 @@ def do_deletion(x0, v0, combined_bps, combined_masses, box, guest_name, leg_type
...
-                guest_name,
-                TRANSITION_STEPS,
-                f"{leg_type.upper()}_DELETION",
-            )
-        if step in (0, int(TRANSITION_STEPS/2), TRANSITION_STEPS-1):
-            if report.too_much_force(ctxt, lamb, box, combined_bps, u_impls):
-                calc_work = False
-                return
+    subsample_freq = 1
+    full_du_dls = ctxt.multiple_steps(deletion_lambda_schedule, subsample_freq)
```
schmolly, line 349: Can/should we use the ctxt.multiple_steps() for the insertion & equilibration, too?

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -139,7 +139,24 @@ void declare_context(py::module &m) {
     }))
     .def("add_observable", &timemachine::Context::add_observable)
     .def("step", &timemachine::Context::step)
-    .def("multiple_steps", &timemachine::Context::multiple_steps)
+    .def("multiple_steps", [](timemachine::Context &ctxt,
+        const py::array_t<double, py::array::c_style> &lambda_schedule,
+        std::optional<int> store_du_dl_freq) -> py::array_t<double, py::array::c_style> {
+        // (ytz): I hate C++
```
badisa, line 146: :+1:

docking/rigorous_work.py
```diff
@@ -182,25 +153,14 @@ def calculate_rigorous_work(
...
-            params, potential = fn(handle.params)
-            combined_bps.append(potential.bind(params))
-        nb_params, nb_potential = wgt.parameterize_nonbonded(ff.q_handle.params, ff.lj_handle.params)
-        combined_bps.append(nb_potential.bind(nb_params))
-        guest_masses = [a.GetMass() for a in guest_mol.GetAtoms()]
-        combined_masses = np.concatenate([water_masses, guest_masses])
-        start_time = time.time()
+        afe = free_energy.AbsoluteFreeEnergy(guest_mol, ff)
+
+        ups, sys_params, combined_masses, _ = afe.prepare_host_edge(ff.get_ordered_params(), water_system, orig_water_coords)
```
schmolly, line 158: Is there an afe.prepare_water_edge()?

tests/test_lambda_potential.py
```diff
@@ -28,6 +29,7 @@ def lambda_potential(conf, params, box, lamb, multiplier, offset, u_fn):
 
 class TestLambdaPotential(GradientTest):
 
+    @unittest.skip("not supported")
```
badisa, line 32: Are unittest markers preferred over the pytest markers?

proteneer: I'm actually not sure, I think unittest markers always get picked up by pytest runners, but pytest markers will not get picked up by unittest runners

timemachine/cpp/src/context.cu
```diff
@@ -47,19 +61,60 @@ void Context::add_observable(Observable *obs) {
...
-    cudaDeviceSynchronize();
+    int buffer_size = (lambda_schedule.size() + store_du_dl_freq - 1) / store_du_dl_freq;
+    try {
+        // indicator so we can set it to a default arg.
+        gpuErrchk(cudaMalloc(&d_du_dl_buffer, buffer_size*sizeof(*d_du_dl_buffer)));
+        gpuErrchk(cudaMemset(d_du_dl_buffer, 0, buffer_size*sizeof(*d_du_dl_buffer)));
+        for(int i=0; i < lambda_schedule.size(); i++) {
+            // decide if we need to store the du_dl for this step
+            unsigned long long *du_dl_ptr = nullptr;
+            if(i % store_du_dl_freq == 0) {
```
badisa, line 80: Is there any concern that lambda_schedule.size() -1 % != 0? IE Should the last step report du_dl?

docking/rigorous_work.py
proteneer: resolved per offline discussions

maxentile (approved):
Looks great to me! This represents a major improvement in simplicity, flexibility, and performance. I also think the API change for how to access the full du/dlambda trajectory will enable cleaner application code. The deletions of restraint and GBSA code seem fine, but should probably be noted in a change log. Should be good to merge when we're happy with any immediate changes this PR requires in `docking/`

badisa (approved):

timemachine/cpp/src/context.cu
proteneer: This is a great question, and often trips people up. But it's actually intended.

Running for 10 steps with a `store_du_dl_freq` of 5 should produce exactly 2 du_dls. We do not trigger the computations for `du/dx, du/dl` etc. unless we are in a step call.

This is actually why @schmolly  was running for 1001 and 10001 steps for parts of her code.

This is because of how the integrators currently work,

```
stp = step
x_t = coords at time t
ddl = if we compute du/dl
stp   0  1  2  3  4  5  6  7  8  9
x_t  0->1->2->3->4->5->6->7->8->9->10
ddl? u              u
```
The `du/dl` for `x_t=10` would be generated on `stp=10`, which is the 11th step in a 0-idx'd system.

proteneer:
Discussed with @schmolly offline re: docking changes. There's a lot of perf we can squeeze out still re: re-using contexts and what not, but in the interest of keeping scope limited we'll do that later.

PR #356, opened 2021-02-25
Use absolute path to examples path

badisa:
* Ensures that users can run the script from where ever
  they want


maxentile (approved):
Nice -- thanks for the fix!

PR #357, opened 2021-02-25
Add get_gpu_count

badisa:
* Avoid requesting too many GPUs
* Add warning to CUDAPoolClient if workers requested is
  greater than GPUs found
* Raise exception if /dev/ doesn't exist, IE it isn't linux

### DIFF
--- a/examples/rbfe_single.py
+++ b/examples/rbfe_single.py
@@ -18,6 +18,7 @@ from testsystems.relative import hif2a_ligand_pair
 from ff.handlers.serialize import serialize_handlers
 from ff.handlers.nonbonded import AM1CCCHandler, LennardJonesHandler
 from parallel.client import CUDAPoolClient
+from parallel.utils import get_gpu_count
 
 from typing import Union, Optional, Iterable, Any, Tuple, Dict
 
@@ -44,7 +45,7 @@ if __name__ == "__main__":
         "--num_gpus",
         type=int,
         help="number of gpus",
-        required=True
+        default=get_gpu_count()
     )
 
     parser.add_argument(
--- a/parallel/client.py
+++ b/parallel/client.py
@@ -7,6 +7,7 @@ import pickle
 import grpc
 
 from parallel import service_pb2_grpc, service_pb2
+from parallel.utils import get_gpu_count
 
 from concurrent import futures
 
@@ -93,6 +94,10 @@ class CUDAPoolClient(ProcessPoolClient):
     will run on a different GPU modulo num workers, which should be set to
     the number of GPUs.
     """
+    def __init__(self, max_workers):
+        super().__init__(max_workers)
+        gpus = get_gpu_count()
+        assert self.max_workers <= gpus, f"More workers '{self.max_workers}' requested than GPUs '{gpus}'"
 
     @staticmethod
     def wrapper(idx, fn, *args):
--- a/parallel/utils.py
+++ b/parallel/utils.py
@@ -1,0 +1,9 @@
+import os
+from subprocess import check_output
+
+
+def get_gpu_count() -> int:
+    output = check_output(["nvidia-smi", "-L"])
+    # Expected to return a line delimited summary of each GPU
+    return len([x for x in output.split(b"\n") if len(x)])
+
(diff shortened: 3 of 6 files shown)


parallel/client.py
```diff
@@ -93,6 +94,10 @@ class CUDAPoolClient(ProcessPoolClient):
     will run on a different GPU modulo num workers, which should be set to
     the number of GPUs.
     """
+    def __init__(self, max_workers):
+        super().__init__(max_workers)
+        if self.max_workers > get_gpu_count():
```
badisa, line 99: I ran into a case where things die if you have workers greater than GPU count, so this seemed helpful to me though maybe it could be noisy in some situation?

parallel/utils.py
```diff
@@ -0,0 +1,9 @@
+import os
+from glob import glob
+
+
+def get_gpu_count() -> int:
+    if not os.path.isdir("/dev/"):
+        raise RuntimeError("No /dev/ directory, can't detect number of GPUs")
+    return len(glob("/dev/nvidia[0-9]*"))
```
badisa, line 8: This is what I have had the most success in using to determine GPU count, though if someone knows a preferable way let me know

parallel/utils.py
```diff
@@ -0,0 +1,9 @@
+import os
+from glob import glob
+
+
+def get_gpu_count() -> int:
+    if not os.path.isdir("/dev/"):
```
badisa, line 6: If there is no dev dir could also return 0, though I don't think this is suppose to be able to run without having /dev/nvidia* paths

maxentile (approved):
Nice! This looks great to me, reducing the amount of environment-specific stuff defined in application script and avoiding problems that arise when workers > GPU count.

Two possible follow-ups: (1) I should edit these examples so user can also request fewer than the maximum number of GPUs available. (2) In `fit_to_multiple_rbfes.py` it is currently possible to define a job with fewer independent simulation windows than there are available GPUs (although there would be no further parallelism exploited if reserving `NUM_GPUS` greater than `num_complex_windows + num_solvent_windows`). Not sure if there might be an automatic way to prevent user from reserving more GPUs than the job can use simultaneously.

examples/rbfe_single.py
```diff
@@ -44,6 +45,7 @@ def wrap_method(args: Iterable[Any], fxn: callable):
         "--num_gpus",
         type=int,
         help="number of gpus",
+        default=get_gpu_count(),
         required=True
```
proteneer, line 49: How does required=True work with default values?

parallel/client.py
proteneer: I would probably assert in this case. The code probably wouldn't even run as it'd be setting `CUDA_VISIBLE_DEVICES` to garbage.

parallel/utils.py
proteneer:
```
nvidia-smi -L
```
is probably more preferred as 
1) it checks to make sure that drivers are present and functional (grepping dev just checks PCIE slots IIRC)
2) generates UUIDs for the GPUs in case they're useful

tests/test_parallel.py
```diff
@@ -54,6 +55,22 @@ def test_jax(self):
 def environ_check():
     return os.environ['CUDA_VISIBLE_DEVICES']
 
+class TestGPUCount(unittest.TestCase):
+
+    @patch("parallel.utils.os.path.isdir")
+    @patch("parallel.utils.glob")
+    def test_get_gpu_count(self, mock_glob, mock_dir):
```
proteneer, line 62: nice!

proteneer (changes requested):

examples/rbfe_single.py
badisa: It still prompts you, so should probably remove the required=True

parallel/utils.py
badisa: @proteneer Running `nvidia-smi -L` seems to take about 8 seconds to run on a p3.16xlarge, is that acceptable? It does verify that drivers exist, however it will still die if cuda is incorrect configured.

proteneer: That's fine - it's negligible relative to the MD simulation.

proteneer:
@badisa looks like tests on this are still failing - how did you want to proceed?

proteneer (approved):
LGTM!

PR #358, opened 2021-02-25
Relative docking

schmolly:
Add relative docking code and tests for docking/ files


docking/dock_and_equilibrate.py
```diff
@@ -177,28 +185,34 @@ def dock_and_equilibrate(
...
 
         if calc_work:
-            work = np.trapz(
-                full_du_dls, insertion_lambda_schedule[::subsample_freq]
-            )
+            work = np.trapz(full_du_dls, insertion_lambda_schedule[::subsample_freq])
             print(f"guest_name: {guest_name}\tinsertion_work: {work:.2f}")
 
         # equilibrate
         for step in range(eq_steps):
```
proteneer, line 202: we may want to also replace this ctxt.multiple_steps() as well later on

docking/relative_docking.py
```diff
@@ -0,0 +1,326 @@
...
+from ff.handlers.deserialize import deserialize_handlers
+from timemachine.lib import custom_ops, LangevinIntegrator
+
+from testsystems.relative import hif2a_ligand_pair
+
+from docking import report
+
+MAX_LAMBDA = 1.0
+MIN_LAMBDA = 0.0
+TRANSITION_STEPS = 501
```
proteneer, line 23: An interesting question is how many transition steps we need to still get reasonable enrichment (even though the estimates themselves may be quite poor)

proteneer (approved):
LGTM - i take it you're saving the BAR stuff for a separate PR?

schmolly:
> LGTM - i take it you're saving the BAR stuff for a separate PR?

Yep!

maxentile:
Great work! I'm pretty excited about this feature. Reviewing now...

docking/rigorous_work.py
```diff
@@ -122,64 +157,49 @@ def calculate_rigorous_work(
         )
         ff = Forcefield(guest_ff_handlers)
 
-
         conformer = guest_mol.GetConformer(0)
         mol_conf = np.array(conformer.GetPositions(), dtype=np.float64)
         mol_conf = mol_conf / 10  # convert to md_units
```
maxentile, line 162: `mol_conf` may be unused

docking/relative_docking.py
```diff
@@ -0,0 +1,326 @@
...
+
+    all_works = {}
+    for system, coords, box, label in zip(
+        [solvated_host_system, water_system],
+        [solvated_host_coords, water_coords],
+        [host_box, water_box],
+        ["protein", "solvent"],
+    ):
+        # minimize w.r.t. both mol_a and mol_b?
+        min_coords = minimizer.minimize_host_4d([mol_a], system, coords, ff, box)
```
maxentile, line 137: I think it'll be interesting to revisit this point later!

Rather than running an equilibrium simulation at lambda=0 and spinning off 0→1 noneq trajectories, could run an equilibrium simulation at lambda=0.5 and spin off 0.5→1 and 0.5→0 trajectories.

(This could be favorable in some cases, if the 0.5→1, 0.5→0 edges are each at least 2x easier to estimate than the 0->1 edge directly. However, this may not be favorable if it's much more expensive to generate decorrelated equilibrium snapshots at lambda=0.5 than at the lambda=0,1 endstates (e.g. if we have pre-computed a bunch of endstate samples, or if mixing is much slower at lambda=0.5 state).)

maxentile (approved):
Looks great to me! This is a powerful feature, that could nicely trade-off throughput and precision as needed.

docking/rigorous_work.py
schmolly: good catch!

PR #359, opened 2021-02-26
Task/improve grpc client

badisa:


examples/rbfe_single.py
```diff
@@ -77,7 +83,20 @@ def wrap_method(args: Iterable[Any], fxn: callable):
...
+        channel = grpc.insecure_channel('0.0.0.0:'+str(port),
+            options = [
+                ('grpc.max_send_message_length', 500 * 1024 * 1024),
+                ('grpc.max_receive_message_length', 500 * 1024 * 1024)
+            ]
+        )
+        stub = parallel.service_pb2_grpc.WorkerStub(channel)
+        stubs.append(stub)
+
+    client = GRPCClient(stubs=stubs)
```
badisa: I will probably just drop this commit before merging

examples/rbfe_single.py
```diff
@@ -77,7 +83,9 @@ def wrap_method(args: Iterable[Any], fxn: callable):
 
     cmd_args = parser.parse_args()
 
-    client = CUDAPoolClient(max_workers=cmd_args.num_gpus)
+    hosts = ["0.0.0.0:2021", "0.0.0.0:2022"]
+    client = GRPCClient(hosts=hosts)
+    # client = CUDAPoolClient(max_workers=cmd_args.num_gpus)
```
maxentile, line 88: Should we make the choice of client an argparse argument for this script, or perhaps add a separate example script for CUDAPoolClient vs GRPCClient? (Or does GRPCClient replace CUDAPoolClient?)

maxentile (approved):
Looks good to me!

examples/rbfe_single.py
badisa: It should be a parameter, which is why I was thinking of dropping this commit initially and put more work in later to swap them out cleanly

proteneer (approved):

PR #360, opened 2021-02-26
Improve multi rbfe training

maxentile:
Completed:
- [x] Add configurable loss function in `examples/hif2a/fit_to_multiple_rbfes.py` (now choose among L1, flat-bottom, or Pseudo-Huber)
- [x] Add another safeguard in optimizer loop. If loss >= 100, assume something went wrong in the simulation, don't trust gradient estimate, and don't take a step that iteration
- [x] Move atom-mapping from application script into `fe/atom_mapping.py` module, and elaborate a couple strategies
- [x] Update `examples/hif2a/generate_star_map.py` to allow multiple cores, filter by transformation size, and attempt multiple atom mapping strategies. (Later, want to define a series of graceful fallbacks from "brittle, but efficient" to "robust, but less efficient" atom maps.)
- [x] Save du_dls and other intermediate results, by passing a callback into `binding_model.predict(...) -> float` (this was a bit more roundabout than before simulator was abstracted as a custom_vjp-equipped, float-valued black-box function, but a callback seems to work fine for now)

To-do before removing [WIP] label:
- [x] Fix any tests of `fe/model.py` and `fe/estimator.py` this PR breaks
- [x] Add tests for new functionality in `fe/loss.py` and `fe/atom_mapping.py`


fe/estimator.py
```diff
@@ -99,29 +135,41 @@ def _deltaG(model, sys_params):
...
-@functools.partial(jax.custom_vjp, nondiff_argnums=(0,))
-def deltaG(model, sys_params):
-    return _deltaG(model, sys_params)[0]
+@functools.partial(jax.custom_vjp, nondiff_argnums=(0,2))
+def deltaG(model, sys_params, callback=None) -> float:
+    return _deltaG(model=model, sys_params=sys_params, callback=callback)[0]
 
-def deltaG_fwd(model, sys_params):
-    return _deltaG(model, sys_params)
+def deltaG_fwd(model, sys_params, callback=None) -> Tuple[float, gradient]:
```
proteneer, line 161: Someone posted a reply in this issue, so we might be able to make the aux stuff less janky
```
https://github.com/google/jax/issues/3806
```


``` python
import dataclasses
import jax
import jax.numpy as jnp

@dataclasses.dataclass
class Foo:
    jax_thing: jnp.array
    non_jax_thing: int

def flatten(foo):
    return (foo.jax_thing,), (foo.non_jax_thing,)

def unflatten(aux_data, children):
    non_jax_thing, = aux_data
    jax_thing, = children
    return Foo(jax_thing, non_jax_thing)

jax.tree_util.register_pytree_node(Foo, flatten, unflatten)
```

proteneer (commented):
First pass done

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -43,8 +55,8 @@
 
 intermediate_configuration = Configuration(
     num_gpus=10,
-    num_complex_windows=30,
-    num_solvent_windows=30,
+    num_complex_windows=60,
```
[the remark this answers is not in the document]
maxentile: Not sure about efficiency, intent is to ensure quadrature bias is the same for production and intermediate configuration. (Remaining differences are sampling variation and initial configuration bias.)

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -261,44 +184,176 @@ def _save_forcefield(filename, ff_params):
...
+
+
+    relative_improvement_bound = 0.8
+
+    def _compute_step_lower_bound(loss):
+        """problem this addresses: on a small fraction of steps, the simulation returns a prediction > 100 kJ/mol
+        away from target, typically indicating an instability was encountered.
+        detect if this occurs, and don't allow a step."""
+
+        blown_up = loss > 100 # loss should never be > 100, unless something has gone very wrong!
```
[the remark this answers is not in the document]
maxentile: Good call -- I wanted to check `loss > 100`, assuming `loss = abs(pred_dG - expt_dG)`.

Could also check `abs(pred_dG) > 100` kJ/mol, `stddev(du_dl, axis=1).max() > 100` kJ/mol, or similar (generally simulation reliability should probably be something that the free energy estimator is responsible for reporting). These would be less brittle to future changes in loss definition.

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -261,44 +184,176 @@ def _save_forcefield(filename, ff_params):
...
 
-        # save updated forcefield files after every gradient step
+        # also save information about this step's parameter gradient and parameter update
+        # checkpoint results to npz (overwrite
+        flat_theta_traj.append(np.array(flat_theta))
+        flat_grad_traj.append(flat_loss_grad)
+        loss_traj.append(loss)
+
+        def _results_to_arrays(results: List[SimulationResult]):
+            #result = SimulationResult(xs=xs, du_dls=full_du_dls, du_dps=grads)
```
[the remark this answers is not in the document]
maxentile: Will move it into docstring -- just a reminder of what fields are available in SimulationResult

fe/estimator.py
```diff
@@ -54,31 +65,56 @@ def simulate(lamb, box, x0, v0, final_potentials, integrator, equil_steps, prod_
...
+        full_du_dls = ctxt.multiple_steps(prod_schedule, du_dl_freq)
+
+    else:
+        xs = [] # in nanometers
+        full_du_dls = []
+
+        # simulate in du_dl_freq-sized chunks
+        t = 0
+        while t < len(prod_schedule):
+            full_du_dls.append(ctxt.multiple_steps(prod_schedule[t: (t + du_dl_freq)], du_dl_freq))
```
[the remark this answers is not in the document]
maxentile: >This needs to be re-worked, or at least you'd need to triple check performance.
>ctxt.get_x_t() is pretty expensive as it requires a memcpy from the GPU to the CPU. Even something like du_dl_freq = 5 this is going to adversely affect speed.

Routes to re-work this:
1. Is there a preferred way to report an MD trajectory with timemachine? If so, will switch to it
2. Print warning if user requests trajectory with `du_dl_freq` thinning interval < 100 steps or something
3. Add parameter to control trajectory thinning interval separately from `du_dl_freq` thinning interval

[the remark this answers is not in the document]
proteneer: 1) Nope, the only jank we have is the PDBWriter right now, which is terribly inefficient. We probably want to switch to an xtc format later on.
2/3) I think we just want du_dl_freq and frame storage frequencies to be decoupled. Generally, I'd avoid write out a frame maybe at most once every 1000 steps. du_dl_freq I think is something like 5 or so IIRC. Usually frame write frequencies >>> du_dl_freq, and for the sake of convenience we can just floor divide into something appropriate.

examples/hif2a/fit_to_multiple_rbfes.py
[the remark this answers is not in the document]
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/360/commits/0e0c776832eb044baf5069e0d66f756dedb76a69

examples/hif2a/fit_to_multiple_rbfes.py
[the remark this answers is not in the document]
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/360/commits/aa8ee07e7a9464f4208c393c52a1b03928ad6c9f

fe/estimator.py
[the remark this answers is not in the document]
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/360/commits/d56a1acbaad90dad242ba927a8e2764ed9c5707b

maxentile:
Ready for another review -- noting a few items came up but will be saved for future PRs:
- [ ] Consider refactoring from using callbacks to using aux returns to access internal state in deltaG call
- [ ] Spring-cleaning in fe/loss.py
- [ ] Reduce code duplication and divergence between related `examples/` scripts
- [ ] Add check in `RelativeFreeEnergy` constructor for whether atom map would cause clashes when (mol_a, mol_b) conformer coordinates are linearly interpolated, and later provide alternatives to linearly interpolating conformers. (Alternatives must still satisfy (a->b, b->a) symmetry requirement.)
- [ ] Improved atom-mapping and edge selection

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -24,8 +26,18 @@
 from collections import namedtuple
 
 from pickle import load
-from typing import List, Union
-Handler = Union[nonbonded.AM1CCCHandler, nonbonded.LennardJonesHandler] # TODO: relax this assumption
+from typing import List, Union, Tuple
+
+from typing import Union, Optional, Iterable, Any, Tuple, Dict
```
proteneer, line 31: Some duplicate imports here from above

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -45,9 +57,9 @@
 )
 
 intermediate_configuration = Configuration(
-    num_gpus=NUM_GPUS,
-    num_complex_windows=30,
-    num_solvent_windows=30,
+    num_gpus=10,
```
proteneer, line 60: I think we can keep this as is for now but @badisa will later on need to change this for the gRPC client.

examples/hif2a/generate_star_map.py
```diff
@@ -39,297 +38,183 @@
...
 
 # SMARTS courtesy of YTZ (Jan 26, 2021)
-eyeballed_smarts_pattern = '[*]~1~[*]~[*]~2~[*]~[*]~[*]~[*](~[#8]~[*]~3~[*]~[*]~[*]~[*]~[*]~3)~[*]~2~[*]~1'
-bicyclic_query_mol = Chem.MolFromSmarts(eyeballed_smarts_pattern)
+bicyclic_smarts_pattern = '[*]~1~[*]~[*]~2~[*]~[*]~[*]~[*](~[#8]~[*]~3~[*]~[*]~[*]~[*]~[*]~3)~[*]~2~[*]~1'
+bicyclic_query_mol = Chem.MolFromSmarts(bicyclic_smarts_pattern)
 
 # filter matches
-mols_with_core_1 = [mol for mol in mols if mol.HasSubstructMatch(bicyclic_query_mol)]
+has_bicycle = lambda mol: len(mol.GetSubstructMatches(bicyclic_query_mol)) == 1
```
proteneer, line 45: shouldn't this be >= 1? It's possible for there to be duplicate matches

examples/rbfe_single.py
```diff
@@ -124,25 +117,24 @@ def wrap_method(args: Iterable[Any], fxn: callable):
         cmd_args.num_prod_steps
     )
 
-    vg_fn = jax.value_and_grad(binding_model.loss, argnums=0)
+    vg_fn = jax.value_and_grad(binding_model.loss, argnums=0, has_aux=True)
```
proteneer, line 120: nice the `has_aux=True` works now?

fe/estimator.py
```diff
@@ -53,32 +65,57 @@ def simulate(lamb, box, x0, v0, final_potentials, integrator, equil_steps, prod_
         ctxt.add_observable(obs)
 
     prod_schedule = np.ones(prod_steps)*lamb
-    du_dl_freq = 5
-    full_du_dls = ctxt.multiple_steps(prod_schedule, du_dl_freq)
 
-    du_dl_mean, du_dl_std = np.mean(full_du_dls), np.std(full_du_dls)
+    # run MD, optionally pausing every du_dl_freq steps to extract x snapshot
+    if not get_trajectory:
```
proteneer, line 70: reminder to myself to just add a `x_t_freq` as well to multiple_steps later on so this manual chunking doesn't have to happen. It seems to be a common enough use case to get trajectories.

fe/estimator.py
```diff
@@ -53,32 +65,57 @@ def simulate(lamb, box, x0, v0, final_potentials, integrator, equil_steps, prod_
...
     grads = []
     for obs in du_dp_obs:
         grads.append(obs.avg_du_dp())
 
-    return (du_dl_mean, du_dl_std), grads
+    result = SimulationResult(xs=xs, du_dls=full_du_dls, du_dps=grads)
+    return result
+
 
+class FreeEnergyModel:
```
proteneer, line 101: Note that classes make the FreeEnergyModel mutable, (namedtuple version was immutable), I'm guessing this is to manually set the callback later on?

fe/estimator.py
```diff
@@ -99,29 +136,41 @@ def _deltaG(model, sys_params):
 
         results = [x.result() for x in futures]
 
-    du_dls = []
+    if not (callback is None):
```
proteneer, line 139: easier to read for me would be:
```
if callback is not None
```
(less parentheses)

fe/model.py
```diff
@@ -60,11 +69,16 @@ def predict(self, ff_params: list, mol_a: Chem.Mol, mol_b: Chem.Mol, core: np.nd
         core: np.ndarray
             N x 2 list of ints corresponding to the atom mapping of the core.
 
+        callback: function
+            accepts a list of SimulationResults and a string, and doesn't return anything. may save to disk
+            TODO: is there a way to save intermediate results to disk that doesn't require passing a callable a few layers deep here...
```
proteneer, line 74: maybe document that link to the jax page that shows how to use jax registration of objects to actually do this properly so we can clean this up later easier?

proteneer (approved):

examples/rbfe_single.py
proteneer: ah I see - this is done through outside of the custom_vjp via the callback trick.

proteneer:
Looks mostly good to me save for some clerical comments.

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: Oops, this was a mistake when resolving merge conflicts in https://github.com/proteneer/timemachine/commit/b71f4d28cce4eecf1cb9060ca05fb585401aa651 -- will revert

examples/hif2a/generate_star_map.py
maxentile: No, should be == 1. (We specifically wanted to exclude the molecule in this set that matches this pattern twice) Will document

fe/estimator.py
maxentile: Ahh, this wasn't for mutability, but to make callback optional. I didn't realize that namedtuple already supports default constructor arguments (https://docs.python.org/3/library/collections.html#collections.namedtuple), and forgot that the constraint that this needs to be immutable. Will revert

examples/hif2a/generate_star_map.py
maxentile: image

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/360/commits/e359676e356dfd7d201425f8a08b56654b3769c7

examples/rbfe_single.py
maxentile: Good catch -- that was in error! Reverted in https://github.com/proteneer/timemachine/pull/360/commits/ca31e56995de24c34fcfe5eaa6f086824857df50

fe/estimator.py
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/360/commits/ed646fbccbe0f70dc6244fd8c16df8feb95d368c

fe/model.py
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/360/commits/9e72088cc1131049a840d0a4b58f596e05cb4532

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/360/commits/573747ab81b782756cae19f4171220a36dd1ba57

fe/estimator.py
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/360/commits/19b3bb60f7e4f5aa39b854a62b4c230b6eec05b0

examples/hif2a/generate_star_map.py
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/360/commits/67cd114fca7e7635e2d0ed2aa33c3616c098e2b2

badisa (approved):

PR #363, opened 2021-03-04
Task/add argparse to hif2a example

badisa:


examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -184,7 +193,6 @@ def _save_forcefield(filename, ff_params):
         prod_steps=configuration.num_prod_steps,
     )
 
-    # TODO: use binding_model.predict rather than binding_model.loss
```
badisa, line 187: @maxentile Based on your last changes I think this comment is no longer relevant?

maxentile: Good catch -- thank you for spotting and correcting that!

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -147,14 +156,14 @@ def _save_forcefield(filename, ff_params):
     # update the forcefield parameters for a few steps, each step informed by a single free energy calculation
 
     # compute and save the sequence of relative_transformation indices
-    num_epochs = int(np.ceil(num_parameter_updates / len(relative_transformations)))
-    np.random.seed(2021)
+    num_epochs = int(np.ceil(args.param_updates / len(relative_transformations)))
+    np.random.seed(args.seed)
```
maxentile, line 160: Nice! Good idea to make the random seed configurable as well (e.g. when we wish to run replicates that differ only in the random seed)

maxentile (approved):
Looks great!

Makes several important parameters configurable rather than hard-coded, and makes quality-of-life improvements along the way.

Following up after this PR, we may also want to make some the remaining hard-coded parameters configurable (by argparse initially, and then if / when these become too nested, considering yaml or similar):
* Parameters governing optimizer step size
* Parameters governing loss function shape

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -125,20 +122,32 @@ def _save_forcefield(filename, ff_params):
 
 
 if __name__ == "__main__":
+    parser = ArgumentParser(description="Fit Forcefield parameters to hif2a")
+    parser.add_argument("--num-gpus", default=NUM_GPUS, help="Number of GPUs to run against")
+    parser.add_argument("--hosts", nargs="*", default=None, help="Hosts running GRPC worker to use for compute")
```
proteneer, line 127: We should probably indicate and check that --hosts and --num_gpus is an XOR relationship

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -125,20 +122,32 @@ def _save_forcefield(filename, ff_params):
...
-    # set up multi-GPU client
-    client = CUDAPoolClient(max_workers=configuration.num_gpus)
+    configuration = None
+    if args.config == "intermediate": # goldilocks
+        configuration = intermediate_configuration
+    elif args.config == "test":
+        configuration = testing_configuration # a little
+    else:
+        configuration = production_configuration # a lot
+    assert configuration is not None, "No configuration provided"
```
proteneer, line 143: when would this ever trigger? One of the three cases for the if statement would always be true; no?

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -125,20 +122,32 @@ def _save_forcefield(filename, ff_params):
...
-    # TODO: make this configurable also
-
-    # set up multi-GPU client
-    client = CUDAPoolClient(max_workers=configuration.num_gpus)
+    configuration = None
+    if args.config == "intermediate": # goldilocks
+        configuration = intermediate_configuration
+    elif args.config == "test":
+        configuration = testing_configuration # a little
+    else:
```
proteneer, line 141: I would probably keep the assertion below and change this to an elif for production.

proteneer (approved):
lgtm - I'd change the logic slightly for some of the cases to be more explicit.

examples/hif2a/fit_to_multiple_rbfes.py
badisa: You are correct in that it would always be true as written. I added it in case the code changed and the logic was no longer valid

proteneer:
lgtm - feel free to merge when you think its ready

PR #364, opened 2021-03-08
Specialize casting for 86

proteneer:
Per @dmclark17 this PR specializes the static_cast convention only for CUDA_ARCH=sm_86. Roughly a 23% speed-up.

Before

```
dhfr-apo: N=23558 speed: 254.43ns/day dt: 1.5fs (ran 100000 steps in 50.95s)
building a protein system with 1758 protein atoms and 7047 water atoms
hif2a-apo: N=8805 speed: 482.40ns/day dt: 1.5fs (ran 100000 steps in 26.87s)
WARNING:absl:No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)
hif2a-rbfe-with-du-dp: N=8840 speed: 402.42ns/day dt: 1.5fs (ran 100000 steps in 32.21s)
hif2a-rbfe-du-dl-freq-0: N=8840 speed: 402.28ns/day dt: 1.5fs (ran 100000 steps in 32.22s)
hif2a-rbfe-du-dl-freq-1: N=8840 speed: 382.35ns/day dt: 1.5fs (ran 100000 steps in 33.90s)
hif2a-rbfe-du-dl-freq-5: N=8840 speed: 396.94ns/day dt: 1.5fs (ran 100000 steps in 32.66s)
solvent-apo: N=6282 speed: 612.97ns/day dt: 1.5fs (ran 100000 steps in 21.15s)
solvent-rbfe-with-du-dp: N=6317 speed: 496.21ns/day dt: 1.5fs (ran 100000 steps in 26.12s)
solvent-rbfe-du-dl-freq-0: N=6317 speed: 498.67ns/day dt: 1.5fs (ran 100000 steps in 25.99s)
solvent-rbfe-du-dl-freq-1: N=6317 speed: 474.17ns/day dt: 1.5fs (ran 100000 steps in 27.34s)
solvent-rbfe-du-dl-freq-5: N=6317 speed: 483.54ns/day dt: 1.5fs (ran 100000 steps in 26.81s)
```

after

```
dhfr-apo: N=23558 speed: 313.43ns/day dt: 1.5fs (ran 100000 steps in 41.36s)
building a protein system with 1758 protein atoms and 7047 water atoms
hif2a-apo: N=8805 speed: 561.24ns/day dt: 1.5fs (ran 100000 steps in 23.10s)
[...]

PR #365, opened 2021-03-08
Task/add grpc status request

badisa:


examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -155,6 +155,7 @@ def _save_forcefield(filename, ff_params):
     else:
         # Setup GRPC client
         client = GRPCClient(hosts=args.hosts)
+    client.verify()
```
badisa, line 158: @maxentile Added you to the PR so you were aware of the (potential) change to adding a verify method to the client

parallel/client.py
```diff
@@ -57,6 +57,17 @@ def submit(self, task_fn, *args):
...
         raise NotImplementedError()
 
+    def verify(self):
+        """Verify performs any necessary checks to verify the client is ready to
+        handle calls to submit.
+
+        Raises
+        ------
+        Exception
+            If verfication fails
```
badisa, line 67: typo: verification*

proto/parallel/service.proto
```diff
@@ -6,8 +6,19 @@ syntax = "proto3";
...
 message PickleData {
     bytes binary = 1;
+}
+
+// Incase we want to send data in the future
+message StatusRequest{
+
+}
+
+message StatusResponse {
```
badisa, line 21: @proteneer Let me know if anything else jumps out as important to verify.

parallel/client.py
```diff
@@ -186,3 +187,16 @@ def submit(self, task_fn, *args):
         future = self.stubs[self._idx].Submit.future(request)
         self._idx = (self._idx + 1) % len(self.stubs)
         return BinaryFutureWrapper(future)
+
+    def verify(self):
```
badisa, line 191: Currently this does does not verify the host that is running the client. This seemed helpful so if you wanted to modify your local 'analysis' code while still using a homogenous set of workers to generate data.

parallel/client.py
```diff
@@ -89,6 +100,12 @@ def submit(self, task_fn, *args):
         self._idx = (self._idx + 1) % self.max_workers
         return future
 
+    def verify(self):
+        """
+        See abstract class for documentation.
+        """
+        return
```
proteneer, line 107: Better to raise `NotImplementedError` here instead.

parallel/client.py
```diff
@@ -164,3 +187,16 @@ def submit(self, task_fn, *args):
...
+        """
+        See abstract class for documentation.
+        """
+        futures = [stub.Status.future(service_pb2.StatusRequest()) for stub in self.stubs]
+        workers_status = [x.result() for x in futures]
+        for field in ("nvidia_driver", "git_sha",):
+            # All fields should be the same
+            host_values = {self.hosts[i]: getattr(x, field) for i, x in enumerate(workers_status)}
+            uni_vals = set(host_values.values())
+            assert len(uni_vals) == 1, f"Not all hosts agreed for {field}: {host_values}"
```
proteneer, line 201: Nice!

parallel/client.py
proteneer: This makes sense to me. Homogeneity amongst the workers is the main thing we care about.

proto/parallel/service.proto
proteneer: I think theses two are the big ones. There's more compiler specific things we may check later on:

- check the CUDA version that was used to compile the code (nvcc --version)
- check the CUDA_ARCH used in CMake

A more stringent/all-or-nothing approach might be to just check the hash of - I *think* they should generate identical byte code if the images are built identically.

``` bash
> md5sum custom_ops.cpython-37m-x86_64-linux-gnu.so
ae4aaa3d10e850372fdc0d51906000c7  custom_ops.cpython-37m-x86_64-linux-gnu.so
```

parallel/client.py
badisa: But doesn't that make it that we can't treat the non-abstract classes as the same? As I see it there is no validation to do here, it is 'good'.

proteneer (approved):
Excellent diagnostic checks to prevent frustration during run-time! Only a minor comment re: compiled code (address as you see necessary, I'd be happy with the PR as-is).

parallel/client.py
badisa: My thought is you can do the following

```
if client_type == "process":
   client = ProcessPoolClient()
elif client_type == "cuda":
   client = CudaClient()
 elif client_type == "grpc":
   client = GrpcClient()
 client.verify()
 ```

proteneer: Sorry you're totally right this was a brainfart on my end - for some reason I thought the ProcessPoolClient was the abstract class! (I hardly ever use it)

proto/parallel/service.proto
badisa: I will investigate this further separate from this. We may run into issues with running this on heterogenous hardware as well, which perhaps this indicates is a bad idea?

proteneer: Agreed. Let's re-visit this later (if at all).

There are very small/subtle differences that may or may not matter between different GPUs/compilers, in fact one of those difference pertain to a technically incorrect but mostly inconsequential rounding issue I'm currently trying to address in #364. In general though, we probably can't guarantee bit-wise identical results across different GPUs (part of this probably has to do with our usage of cub).

PR #367, opened 2021-03-08
Improve saving intermediate results

maxentile:
Instead of accumulating a trajectory of intermediate results in memory (and repeatedly overwriting a growing `.npz` file containing the whole history), saves a snapshot of each step's results as a separate file in output_path

### DIFF
--- a/examples/hif2a/fit_to_multiple_rbfes.py
+++ b/examples/hif2a/fit_to_multiple_rbfes.py
@@ -27,11 +27,10 @@ from parallel.utils import get_gpu_count
 from collections import namedtuple
 
 from pickle import load
-from typing import List, Union, Tuple, Dict
 
 from optimize.step import truncated_step
 
-from typing import List, Union
+from typing import Tuple, Dict, List, Union
 from pathlib import Path
 from time import time
 
@@ -74,22 +73,9 @@ testing_configuration = Configuration(
 #   and a "training configuration"
 #       (which describes the overall training loop)
 
-# don't make assumptions about working directory
-
 # locations relative to project root
 root = Path(__file__).absolute().parent.parent.parent
 path_to_protein = str(root.joinpath('tests/data/hif2a_nowater_min.pdb'))
-path_to_ff = str(root.joinpath('ff/params/smirnoff_1_1_0_ccc.py'))
-
-# locations relative to example folder
-path_to_results = Path(__file__).absolute().parent
-path_to_transformations = str(path_to_results.joinpath('relative_transformations.pkl'))
-
-# load and construct forcefield
-with open(path_to_ff) as f:
-    ff_handlers = deserialize_handlers(f.read())
-
-forcefield = Forcefield(ff_handlers)
 
 
 class ParameterUpdate:
@@ -101,12 +87,11 @@ class ParameterUpdate:
 
     # TODO: def __str__
 
-    def save(self, name='parameter_update.npz'):
-        """save numpy arrays to path/to/results/{name}.npz"""
-        parameter_update_path = path_to_results.joinpath(name)
-        print(f'saving parameter updates to {parameter_update_path}')
+    def save(self, fname='parameter_update.npz'):
+        """save numpy arrays to fname"""
+        print(f'saving parameter updates to {fname}')
         np.savez(
-            file=parameter_update_path,
+            file=fname,
             before=self.before,
             after=self.after,
             gradient=self.gradient,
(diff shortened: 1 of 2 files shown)


examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -101,33 +89,46 @@ def __init__(self, before, after, gradient, update):
...
-    # TODO: update path
-
-    with open(path_to_results.joinpath(filename), 'w') as fh:
+def _save_forcefield(fname, ff_params):
+    with open(fname, 'w') as fh:
         fh.write(ff_params)
 
 
 if __name__ == "__main__":
+    import datetime
```
badisa, line 110: nit: group imports together at top?

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -74,22 +74,10 @@
 #   and a "training configuration"
 #       (which describes the overall training loop)
 
-# don't make assumptions about working directory
 
 # locations relative to project root
 root = Path(__file__).absolute().parent.parent.parent
```
badisa, line 79: Not really for this PR, but realized reading through this on my own, was that we probably want to do something like...

```
import timemachine
root = Path(timemachine.__file__).parent
```

This way this file can move and the implication of root stays the same

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -349,40 +346,33 @@ def save_in_memory_callback(results, stage):
...
-
-        # TODO: update path
-        path_to_du_dls = 'du_dls_checkpoint.npz'
+        # save du_dls snapshot
+        path_to_du_dls = output_path.joinpath(f'du_dls_snapshot_{step}.npz')
         print(f'saving du_dl trajs to {path_to_du_dls}')
         du_dls_dict = dict() # keywords here must be strings
-        for key in results_traj:
-            du_dls_dict[_stringify_key(key)] = _results_to_arrays(results_traj[key])[1] # TODO: make this less gross
+        for stage in {'solvent', 'complex'}:
```
badisa, line 358: nit: `for stage in results_this_step.keys()`

badisa (commented):
Should we modify rbfe_single.py as well? It looks to have the same accumulation issues

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -292,22 +300,11 @@ def _blew_up(results: List[SimulationResult]):
...
-
-    results_traj = dict() # indexed by (step, stage) tuples # TODO: proper type hint
-
+    results_this_step = dict() # {stage : result} pairs # TODO: proper type hint
 
     def save_in_memory_callback(results, stage):
-        # TODO: replace with saving this to disk rather than accumulating in memory
-        global results_traj
-        results_traj[(step, stage)] = results
+        global results_this_step
```
proteneer, line 306: Random Q: do you actually need to declare the global here? Doesn't this just form a closure around the variable anyways?

examples/hif2a/fit_to_multiple_rbfes.py
proteneer: +1

Only time imports should be conditional is if there's dependency issues

badisa (approved):

proteneer (approved):
lgtm - good catch on the leak

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/367/commits/a24361543e280abcd444aa08f756fb96489d3557

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/367/commits/a24361543e280abcd444aa08f756fb96489d3557

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: Probably not, just emphasizing the intent to have this function modify in-place a variable in main scope, even though it's going to be called a [couple](https://github.com/proteneer/timemachine/blob/0ce5b98dcadf5eb8ee54b1d069d22f6bf8792e7d/fe/model.py#L131) [layers](https://github.com/proteneer/timemachine/blob/0ce5b98dcadf5eb8ee54b1d069d22f6bf8792e7d/fe/estimator.py#L112) deep... (This callback workaround is temporary, until we settle on a Jax-idiomatic return type for deltaG() that more fully captures what we want to access from a free energy simulation.)

maxentile:
>Should we modify rbfe_single.py as well? It looks to have the same accumulation issues

I'd like to postpone to a later PR where we consolidate these two examples. Resolving a few arbitrary divergences that I've accumulated between these two scripts, `rbfe_single.py` should probably become "call `fit_to_rbfes.py` with `--path_to_edges` pointing to a length-1 list of RBFE definitions rather than a length-2+ list  of RBFE definitions."

PR #368, opened 2021-03-08
Feature/yaml configuration

badisa:
Just to get some eyes on what I am understanding fro the generate_star_map example

### DIFF
--- a/examples/hif2a/hifa2_config.yaml
+++ b/examples/hif2a/hifa2_config.yaml
@@ -1,0 +1,22 @@
+---
+map_generation:
+  atom_mapping_strategy: geometry
+  transformation_threshold: 3
+  protein: datasets/fep-benchmark/hif2a/5tbm_prepared.pdb
+  forcefield: ff/params/smirnoff_1_1_0_ccc.py
+  ligands:
+    - datasets/fep-benchmark/hif2a/ligands.sdf
+  # Cores and hubs must be in the same order. 
+  cores:
+    - method: smarts
+      smarts: "[*]~1~[*]~[*]~2~[*]~[*]~[*]~[*](~[#8]~[*]~3~[*]~[*]~[*]~[*]~[*]~3)~[*]~2~[*]~1"
+    - method: smarts
+      smarts: "[*]1~[*]~[*]~[*]~[*]~[*]1~[#8]~[*]2~[*]~[*]~[*]~[*]~[*]2"
+  networks:
+    - method: manual
+      property: "ID"
+      value: "165"
+    - method: manual
+      property: "ID"
+      value: "41"
+  output: relative_transformations.pkl
--- a/fe/atom_mapping.py
+++ b/fe/atom_mapping.py
@@ -1,7 +1,11 @@
 import numpy as np
 from rdkit import Chem
 from rdkit.Chem import rdFMCS
+
+import matplotlib.pyplot as plt
+
 from fe.topology import AtomMappingError
+from fe.utils import core_from_distances, simple_geometry_mapping
 
 
 class CompareDist(rdFMCS.MCSAtomCompare):
@@ -22,7 +26,7 @@ class CompareDist(rdFMCS.MCSAtomCompare):
         return bool(np.linalg.norm(x_i - x_j) <= self.threshold)  # must convert from np.bool_ to Python bool!
 
 
-def mcs_map(a, b, threshold=0.5):
+def mcs_map(a, b, threshold: float = 0.5):
     """Find the MCS map of going from A to B"""
     params = rdFMCS.MCSParameters()
     params.AtomTyper = CompareDist(threshold=threshold)
@@ -53,7 +57,7 @@ def transformation_size(n_A, n_B, n_MCS):
     return (n_A + n_B) - 2 * n_MCS
 
 
-def compute_all_pairs_mcs(mols):
+def compute_all_pairs_mcs(mols, threshold: float = 0.5):
     """Generate square matrix of MCS(mols[i], mols[j]) for all pairs i,j
 
     Parameters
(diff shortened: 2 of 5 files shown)


badisa:
Core 0 ([*]~1~[*]~[*]~2~[*]~[*]~[*]~[*](~[#8]~[*]~3~[*]~[*]~[*]~[*]~[*]~3)~[*]~2~[*]~1)
Generates 16 edges and 9 errors

Core 1([*]1~[*]~[*]~[*]~[*]~[*]1~[#8]~[*]2~[*]~[*]~[*]~[*]~[*]2)
Generates 6 edges and 1 error

This also 'abandons' one molecule that doesn't match either core (ID 224)

examples/hif2a/hifa2_config.yaml
```diff
@@ -0,0 +1,15 @@
+---
+map-generation:
+  strategy: geometry
```
maxentile, line 3: Should we specify this further as `atom_mapping_strategy` or similar? It's foreseeable we would could elaborate on this to reference parameterized strategies for other tasks (e.g. an `edge_selection_strategy`).

maxentile (approved):
This is great! I noted a few possible elaborations, but overall I think this nicely abstracts what had been hard-coded in the `generate_star_map.py` example, and paves the way toward a more complete yaml specification of the settings controlling relative free energy calculations in timemachine.

Side note: The yaml specification @andrrizzi introduced for absolute free energy calculations in yank greatly eased usability and enabled concise expression of combinatorial experiments, showcased here https://github.com/choderalab/yank/tree/master/docs/yank-yaml-cookbook . The configuration choices that should be exposed in this part of timemachine aren't quite the same, but yank yaml is a good example to study for inspiration.

examples/hif2a/hifa2_config.yaml
```diff
@@ -0,0 +1,15 @@
...
+  protein: datasets/fep-benchmark/hif2a/5tbm_prepared.pdb
+  forcefield: ff/params/smirnoff_1_1_0_ccc.py
+  ligands:
+    - datasets/fep-benchmark/hif2a/ligands.sdf
+  # Cores and hubs must be in the same order. 
+  cores:
+    - "[*]~1~[*]~[*]~2~[*]~[*]~[*]~[*](~[#8]~[*]~3~[*]~[*]~[*]~[*]~[*]~3)~[*]~2~[*]~1"
+    - "[*]1~[*]~[*]~[*]~[*]~[*]1~[#8]~[*]2~[*]~[*]~[*]~[*]~[*]2"
+  hubs:
+    - "165"
```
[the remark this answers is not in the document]
badisa: Definitely doable, and if we expand the manual selection we could probably easily allow users to modify the property in question they are specifying rather than using out convention of looking at ID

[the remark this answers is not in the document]
badisa: I reworked it so that we can more easily add new methods (though it does risk the schema being too flexible). I believe this covers most use cases by partitioning the molecules by cores and running the different network generation methods (what were previously called hubs) on each set of cores.

Will likely want to write some more validation code to ensure the schema doesn't get out of control

fe/atom_mapping.py
```diff
@@ -71,8 +75,10 @@ def compute_all_pairs_mcs(mols):
     mcs_s = np.zeros((len(mols), len(mols)), dtype=object)
 
     for i in range(len(mols)):
-        for j in range(len(mols)):
-            mcs_s[i, j] = mcs_map(mols[i], mols[j])
+        for j in range(i, len(mols)):
```
maxentile, line 78: Nice! Can mark the to-do note resolved

examples/hif2a/hifa2_config.yaml
maxentile: Resolved by @badisa in https://github.com/proteneer/timemachine/pull/368/commits/dde1d3b5fd571338c99187718fcf800628e55d27

examples/hif2a/hifa2_config.yaml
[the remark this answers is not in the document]
maxentile: Nice! I think the current approach you've taken (dictionary of keyword arguments) is good for this stage of development, while we are adding several atom-mapping + edge-selection strategies.

timemachine/parser.py
```diff
@@ -0,0 +1,50 @@
...
+class MapGenerationConfig:
+
+    protein: str
+    ligands: List[str]
+    transformation_threshold: int = 3
+    atom_mapping_strategy: str = "geometry"
+    forcefield: str = str(ROOT_DIR.joinpath("ff/params/smirnoff_1_1_0_ccc.py"))
+    output: Optional[str] = None
+    cores: Optional[List[Dict]] = None
+    networks: Optional[List[Dict]] = None
```
maxentile: This looks good to me, although I think we might adjust a bit later if these settings should be nested

I think the top-level things that we want to specify here are:
* `protein` target
* small molecule `forcefield`
* `ligands` list
* `output` path
* `atom_mapping_strategy` -- all information required to specify a function `generate_atom_map(mol_a, mol_b) -> atom_map`. (In this case, something like "geometry" with keyword argument "threshold = 0.5".)
* `edge_selection_strategy` -- all information required to specify a function `generate_edges(ligands) -> list_of_pairs_of_ligands`.
   * Basic case is "generate a star, including (reference_ligand, other_ligand) pairs, where reference_ligand is selected from `ligands` either manually or using a named strategy."
   * We may want to apply "pre-filters" to the set of ligands, and  "post-filters" to the set of candidate edges.
      * A pre-filter applied in this case is something like "ligands must match user-specified core smarts".
      * Post-filters might include "an edge may add or remove no more than `transformation_size_threshold` atoms", and "an edge may not change ring sizes".
    * Edge selection strategies may be combined (e.g. if you have a complete strategy tailored to one core, and a complete strategy tailored to another core, we may want to apply them both to a ligand set and take the union the edges they select).

PR #371, opened 2021-03-15
Prepare protein prior to building the system

badisa:
* Use OESpruce

### DIFF
--- a/examples/hif2a/fit_to_multiple_rbfes.py
+++ b/examples/hif2a/fit_to_multiple_rbfes.py
@@ -109,11 +109,11 @@ def _save_forcefield(fname, ff_params):
 
 if __name__ == "__main__":
     default_output_path = f"results_{str(datetime.datetime.now())}"
-    parser = ArgumentParser(description="Fit Forcefield parameters to hif2a")
-    parser.add_argument("--num-gpus", default=None, type=int,
+    parser = ArgumentParser(description="Fit Forcefield parameters to multiple RBFEs")
+    parser.add_argument("--num_gpus", default=None, type=int,
                         help=f"Number of GPUs to run against, defaults to {NUM_GPUS} if no hosts provided")
     parser.add_argument("--hosts", nargs="*", default=None, help="Hosts running GRPC worker to use for compute")
-    parser.add_argument("--param-updates", default=1000, type=int, help="Number of updates for parameters")
+    parser.add_argument("--param_updates", default=1000, type=int, help="Number of updates for parameters")
     parser.add_argument("--seed", default=2021, type=int, help="Seed for shuffling ordering of transformations")
     parser.add_argument("--config", default="intermediate", choices=["intermediate", "production", "test"])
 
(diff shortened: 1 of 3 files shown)


ff/handlers/bcc_aromaticity.py
```diff
@@ -25,6 +26,18 @@ def __init__(self, *args, smirks: str, **kwargs):
         super(InvalidSmirksError, self).__init__(*args, **kwargs)
         self.smirks = smirks
 
+@contextmanager
+def openeye_log_wrapper():
```
proteneer, line 30: I'm pretty unfamiliar with OEChem's context manager and logging this, where does this end up going?

md/builders.py
```diff
@@ -1,15 +1,59 @@
+from tempfile import NamedTemporaryFile
+
 import numpy as np
 from simtk import unit
 from simtk.openmm import app, Vec3
 
+from openeye.oechem import (
+    oemolistream,
```
proteneer, line 8: My first time seeing this kind of import setup - pretty nice!

md/builders.py
```diff
@@ -1,15 +1,59 @@
...
+from openeye.oespruce import (
+    OEStructureMetadata,
+    OEMakeDesignUnitOptions,
+    OEMakeBioDesignUnits,
+)
+
+from ff.handlers.bcc_aromaticity import openeye_log_wrapper
+
+
+def prepare_protein(path: str, output_path: str):
```
proteneer, line 24: Do you mind documenting what this actually does in terms of structures prep?

md/builders.py
```diff
@@ -1,15 +1,59 @@
...
-def build_protein_system(host_pdbfile):
-
+def build_protein_system(host_pdbfile: str, prepare: bool = True):
     host_ff = app.ForceField('amber99sbildn.xml', 'tip3p.xml')
-    host_pdb = app.PDBFile(host_pdbfile)
+    if not prepare:
+        host_pdb = app.PDBFile(host_pdbfile)
+    else:
+        with NamedTemporaryFile(suffix=".pdb") as temp:
+            prepare_protein(host_pdbfile, temp.name)
```
proteneer, line 55: oof - this can't handle string-like handles? was thinking at first we could do stringio or something but not sure if the API supports it.

md/builders.py
```diff
@@ -1,15 +1,59 @@
...
+            host_pdb = app.PDBFile(temp.name)
```
proteneer, line 56: We should double check to make sure there's nothing internal inside app.PDBFile that might need this file later on - doesn't the context manager delete the file?

ff/handlers/bcc_aromaticity.py
badisa: Its the same behavior as previously, its dumping it to an oestream and returning it as a context manager. Just allowed me to capture the output the same way in two places without having to check a boolean return value

md/builders.py
badisa: I much prefer explicit imports to module imports, unless absolutely necesary

md/builders.py
badisa: We could make it work with a handle, would just need to fill up the oemolistream just need to refresh myself on how to do that.

md/builders.py
badisa: The context manager does delete this file. Will verify that it isn't storing the file reference somewhere

maxentile (approved):
Looks great to me!

proteneer (approved):
lgtm

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -122,9 +121,13 @@ def _save_forcefield(fname, ff_params):
     parser.add_argument("--path_to_edges", default="relative_transformations.pkl",
                         help="Path to pickle file containing list of RelativeFreeEnergy objects")
     parser.add_argument("--output_path", default=default_output_path, help="Path to output directory")
+    parser.add_argument("--protein-path", default=str(root.joinpath("datasets/fep-benchmark/hif2a/5tbm_prepared.pdb")), help="Path to protein")
```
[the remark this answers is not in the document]
proteneer: +1 to this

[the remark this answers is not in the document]
proteneer: ps my choice for underscore is purely double clicking convenience, double clicking foo_bar selects the the entire string, double clicking on foo-bar on the foo vs bar parts selects only either foo or bar.

badisa:
Will revisit after minimization work

PR #372, opened 2021-03-17
Task/reduce likelihood of minimizer exploding

badisa:
Runs L-BFGS as part of `minimize_host_4d`.

Had originally thought about adding the first phase minimizing to the boxes, but this seemed consistent with what users want to do. If that is mistaken I can take another look at minimizing the boxes after construction, though that probably needs to be done against regardless if injecting anything into the system.

### DIFF
--- a/docking/dock_and_equilibrate.py
+++ b/docking/dock_and_equilibrate.py
@@ -7,7 +7,7 @@ import numpy as np
 
 from rdkit import Chem
 
-from md import builders
+from md import builders, minimizer
 from fe import pdb_writer, free_energy
 from ff import Forcefield
 from ff.handlers.deserialize import deserialize_handlers
@@ -86,15 +86,23 @@ def dock_and_equilibrate(
         solvated_topology,
     ) = builders.build_protein_system(host_pdbfile)
 
-    # sometimes water boxes are sad. Should be minimized first; this is a workaround
-    host_box += np.eye(3) * 0.1
-
     solvated_host_pdb = os.path.join(outdir, "solvated_host.pdb")
     writer = pdb_writer.PDBWriter([solvated_topology], solvated_host_pdb)
     writer.write_frame(solvated_host_coords)
     writer.close()
     solvated_host_mol = Chem.MolFromPDBFile(solvated_host_pdb, removeHs=False)
     os.remove(solvated_host_pdb)
+
+    guest_ff_handlers = deserialize_handlers(
+        open(
+            os.path.join(
+                os.path.dirname(os.path.abspath(__file__)),
+                "..",
+                "ff/params/smirnoff_1_1_0_ccc.py",
+            )
+        ).read()
+    )
+    ff = Forcefield(guest_ff_handlers)
 
     # Run the procedure
     print("Getting guests...")
@@ -105,28 +113,20 @@ def dock_and_equilibrate(
         guest_conformer = guest_mol.GetConformer(0)
         orig_guest_coords = np.array(guest_conformer.GetPositions(), dtype=np.float64)
         orig_guest_coords = orig_guest_coords / 10  # convert to md_units
-        guest_ff_handlers = deserialize_handlers(
-            open(
-                os.path.join(
-                    os.path.dirname(os.path.abspath(__file__)),
-                    "..",
-                    "ff/params/smirnoff_1_1_0_ccc.py",
-                )
-            ).read()
-        )
-        ff = Forcefield(guest_ff_handlers)
+
+        minimized_coords = minimizer.minimize_host_4d([guest_mol], solvated_host_system, solvated_host_coords, ff, host_box)
 
         afe = free_energy.AbsoluteFreeEnergy(guest_mol, ff)
 
         ups, sys_params, combined_masses, _ = afe.prepare_host_edge(
-            ff.get_ordered_params(), solvated_host_system, solvated_host_coords
+            ff.get_ordered_params(), solvated_host_system, minimized_coords
         )
 
         combined_bps = []
         for up, sp in zip(ups, sys_params):
             combined_bps.append(up.bind(sp))
 
-        x0 = np.concatenate([solvated_host_coords, orig_guest_coords])
+        x0 = np.concatenate([minimized_coords, orig_guest_coords])
         v0 = np.zeros_like(x0)
         print(
             f"SYSTEM",
--- a/examples/hif2a/fit_to_multiple_rbfes.py
+++ b/examples/hif2a/fit_to_multiple_rbfes.py
@@ -266,13 +266,9 @@ if __name__ == "__main__":
         # build the complex system
         # note: "complex" means "protein + solvent"
         complex_system, complex_coords, _, _, complex_box, _ = builders.build_protein_system(prot_path)
-        # TODO: optimize box
-        complex_box += np.eye(3) * 0.1  # BFGS this later
 
         # build the water system
         solvent_system, solvent_coords, solvent_box, _ = builders.build_water_system(4.0)
-        # TODO: optimize box
-        solvent_box += np.eye(3) * 0.1  # BFGS this later
 
         systems[prot_path] = RBFEModel(
             client=client,
--- a/examples/rhfe_dual.py
+++ b/examples/rhfe_dual.py
@@ -65,7 +65,7 @@ ff = Forcefield(ff_handlers)
 
 # for RHFE we need to insert the reference ligand first, before inserting the
 # decoupling ligand
-minimized_coords = minimizer.minimize_4d(romol_a, system, host_coords, ff, box)
+minimized_coords = minimizer.minimize_host_4d([romol_a], system, host_coords, ff, box)
 
 # note the order in which the coordinates are concatenated in this step --
 #   in a later step we will need to combine recipes in the same order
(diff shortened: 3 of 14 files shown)


docking/dock_and_equilibrate.py
```diff
@@ -86,16 +86,24 @@ def dock_and_equilibrate(
...
-    host_box += np.eye(3) * 0.1
-
     solvated_host_pdb = os.path.join(outdir, "solvated_host.pdb")
     writer = pdb_writer.PDBWriter([solvated_topology], solvated_host_pdb)
     writer.write_frame(solvated_host_coords)
     writer.close()
     solvated_host_mol = Chem.MolFromPDBFile(solvated_host_pdb, removeHs=False)
     os.remove(solvated_host_pdb)
 
+    guest_ff_handlers = deserialize_handlers(
```
badisa, line 96: ABC: Moved this out of the loop

examples/rhfe_dual.py
```diff
@@ -65,7 +65,7 @@ def get_romol_conf(mol):
 
 # for RHFE we need to insert the reference ligand first, before inserting the
 # decoupling ligand
-minimized_coords = minimizer.minimize_4d(romol_a, system, host_coords, ff, box)
+minimized_coords = minimizer.minimize_host_4d([romol_a], system, host_coords, ff, box)
```
badisa, line 68: This doesn't fix this script (IE it doesn't run), but at least it isn't try to call a non-existent API

md/minimizer.py
```diff
@@ -103,21 +108,36 @@ def minimize_host_4d(mols, host_system, host_coords, ff, box):
...
+        x0 = x0.reshape((-1, 3))
+        u = 0
+        du_dx_sum = np.zeros_like(x0)
+        for impl in u_impls:
+            du_dx, _, u_bond = impl.execute(x0, box, lamb)
+            u += u_bond
+            du_dx_sum += du_dx
+        return u, du_dx_sum.reshape(-1)
+
+    options = {"disp": verbose, "maxiter": 25}
```
badisa, line 122: 25 iterations based on the CDK8 set I was running into issues with before. Most of the energy reduction took place in the first 20 iterations and if it ran for too long it would end up exploding. Could probably avoid the latter by adding constraints

badisa:
Intend to hold off merging this until I have confirmed that the CDK8 set I have runs to completion through `fit_to_multiple_rbfe.py`, though seems to be working in all of my previous test cases.

maxentile (commented):
Nice work @badisa ! I do have one substantive comment: I notice this PR deletes the temporary box padding workaround. (Good riddance!) However, unless I missed something, the PR does not currently replace this workaround with an optimization of the box. Optimizing the box would be the ideal replacement I think (at least until we settle on a barostat!). To optimize the box during energy minimization, I think we would like to estimate `dU/dbox` and provide this to the scipy optimize routine. Initially, we can estimate `dU/dbox` via finite difference, append it to the `dU/dcoords` you collect from `impl.execute`, and have scipy L-BFGS now optimize over a slightly bigger variable `coords_and_box`, rather than optimizing over `coords` only with fixed box. Is this roughly how we would like to replace the box padding workaround? If so, I can open a modification to this PR that reflects this.

examples/rhfe_dual.py
proteneer: Don't worry about it - I will fix this later.

md/minimizer.py
```diff
@@ -8,14 +8,16 @@
...
     """Coordinates of mol's 0th conformer, in nanometers"""
     conformer = mol.GetConformer(0)
     guest_conf = np.array(conformer.GetPositions(), dtype=np.float64)
     return guest_conf/10 # from angstroms to nm
 
 
-def minimize_host_4d(mols, host_system, host_coords, ff, box):
+def minimize_host_4d(mols, host_system, host_coords, ff, box, verbose: bool = False):
     """
     Insert mols into a host system via 4D decoupling using a 0 Kelvin Langevin integrator.
```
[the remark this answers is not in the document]
proteneer: We need this function to be deterministic and independent of any random seeds. Velocity rescaling methods would be okay here though.

md/minimizer.py
proteneer: IIRC when we debug this together we mainly saw line searching blow up.

docking/dock_and_equilibrate.py
schmolly: docking/pose_dock.py could use similar treatment

proteneer:
> Nice work @badisa ! I do have one substantive comment: I notice this PR deletes the temporary box padding workaround. (Good riddance!) However, unless I missed something, the PR does not currently replace this workaround with an optimization of the box.

The box work-around was always pure-jank. The density of the water box that gets copied over from OpenMM is "roughly correct", and is always preferred to the slightly increased box size.

> Optimizing the box would be the ideal replacement I think (at least until we settle on a barostat!).

Minimizing the energy is not the proper objective. You can imagine a single particle with a +1 charge would optimize the box into an infinitely large box in order to minimize U. You need to optimize the box into an appropriate pressure, so what you'd want really is still to compute current virial and compare it against a target virial. The box minimization PR be done in a separate step, as there's a substantial amount of validation we need to do.

md/minimizer.py
badisa: Correct. My sense is that without constraints it is creating a 'bad' state that makes the gradients/energies no longer meaningful. The line searching would fail after big jumps in energy

maxentile:
>The density of the water box that gets copied over from OpenMM is "roughly correct", and is always preferred to the slightly increased box size.
>The box minimization PR be done in a separate step, as there's a substantial amount of validation we need to do.

Thank you for clarifying!

(I didn't have a good sense for this, but my intuition was that energy-minimizing the box with inserted mols present would be a preferable way to get an approximately NPT-typical box volume for the new system, compared with retaining box dimensions assumed to be approximately NPT-typical for a system without the inserted mols. As you point out, however, optimizing the box could lead to arbitrarily NPT-atypical boxes. Propose to move further discussion of how to handle box equilibration to a new Issue, since it sounds out of scope for the current PR: https://github.com/proteneer/timemachine/issues/373)

badisa:
The CDK8 edges I ran exploded on minimization again, though it was after 920 steps. Need to investigate what edges triggered the failures

md/fire.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def init_fn(R: np.array, **kwargs) -> FireDescentState:
+        V = jnp.zeros_like(R)
+        n_pos = jnp.zeros((), jnp.int32)
+        F = force(R, **kwargs)
+        return FireDescentState(
+            R, V, F, dt_start, alpha_start, n_pos
+        )  # pytype: disable=wrong-arg-count
+
+    def apply_fn(state: FireDescentState, **kwargs) -> FireDescentState:
```
badisa, line 114: Note that the LAMMPS paper users a modified algorithm which seems to be built around a different integrator: https://arxiv.org/pdf/1908.02038.pdf

badisa:
These changes worked against all of the edges that previously failed, need to rerun the intermediate config to be sure.

md/minimizer.py
```diff
@@ -101,40 +144,18 @@ def minimize_host_4d(mols, host_system, host_coords, ff, box, verbose: bool = Fa
     x0 = combined_coords
     v0 = np.zeros_like(x0)
 
-    u_impls = []
-
-    for bp in final_potentials:
-        fn = bp.bound_impl(precision=np.float32)
-        u_impls.append(fn)
-
+    x0 = fire_minimize(x0, u_impls, box, np.ones(100))
```
badisa, line 147: Selected 100 as rough starting point. The sandwich optimization adds ~2 seconds per optimization on a G4DN (T4 gpu) as it stands. Intend to see if this works for the 1000 opt steps and if that looks good use the generated FFs to evaluate a faster set of steps.

I have tried just using minimize with a lambda schedule of 2k steps (from 1.0 to 0) and that was much slower than taking MD steps

badisa: Recent change to remove jax numpy in favor of pure numpy has reduced that to ~1 second. This is comparing 25 steps of Scipy optimize to 100 steps of Fire

proteneer: Nice!

maxentile (approved):
Nice work!

md/fire.py
```diff
@@ -0,0 +1,146 @@
+# Refer to License below. Code taken from https://github.com/google/jax-md/blob/master/jax_md
```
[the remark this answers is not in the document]
badisa: That would require making it a dependency, which generally we don't need?

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -187,13 +187,9 @@ def _save_forcefield(fname, ff_params):
     # build the complex system
     complex_system, complex_coords, _, _, complex_box, _ = builders.build_protein_system(
         path_to_protein)
-    # TODO: optimize box
```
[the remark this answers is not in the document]
badisa: Correct, this is not adjusting the box dimensions. Though ti does make the padding no longer necessary to prevent blow ups.

md/fire.py
badisa: An advantage of FIRE over scipy optimize (at least L-BFGS) is that it is faster. Definitely worth a comparison in the future.

PR #376, opened 2021-03-24
Allow multiple_steps to also return the coordinates.

proteneer:
Note that this changes the signature of multiple_steps return type to a tuple `(du_dls, xs)`

### DIFF
--- a/fe/estimator.py
+++ b/fe/estimator.py
@@ -63,7 +63,7 @@ def simulate(lamb, box, x0, v0, final_potentials, integrator, equil_steps, prod_
         no frames are returned.
 
     du_dl_interval: int
-        how often we store du_dls. if du_dl_interval == 5 then
+        how often we store du_dls. if du_dl_interval == 0 then
         no du_dls are returned
 
     Returns
--- a/fe/estimator.py
+++ b/fe/estimator.py
@@ -27,13 +27,49 @@ def unflatten(aux_data, children):
 
 jax.tree_util.register_pytree_node(SimulationResult, flatten, unflatten)
 
-def simulate(lamb, box, x0, v0, final_potentials, integrator, equil_steps, prod_steps, get_trajectory=False,
+def simulate(lamb, box, x0, v0, final_potentials, integrator, equil_steps, prod_steps,
     x_interval=1000, du_dl_interval=5):
     """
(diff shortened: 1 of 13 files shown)


fe/estimator.py
```diff
@@ -27,13 +27,49 @@ def unflatten(aux_data, children):
...
 
-def simulate(lamb, box, x0, v0, final_potentials, integrator, equil_steps, prod_steps, get_trajectory=False,
+def simulate(lamb, box, x0, v0, final_potentials, integrator, equil_steps, prod_steps,
     x_interval=1000, du_dl_interval=5):
     """
     Run a simulation and collect relevant statistics for this simulation.
 
     Parameters
     ----------
+    lamb: float
```
badisa, line 37: 👍

timemachine/cpp/src/context.cu
```diff
@@ -61,46 +63,66 @@ void Context::add_observable(Observable *obs) {
...
+    int store_x_freq) {
     unsigned long long *d_du_dl_buffer = nullptr;
     // try catch block is to deal with leaks in d_du_dl_buffer
-    if(store_du_dl_freq==0) {
-        store_du_dl_freq = lambda_schedule.size();
+    if(store_du_dl_freq <= 0) {
+        throw std::runtime_error("store_du_dl_freq <= 0");
+    }
+    if(store_x_freq <= 0) {
+        throw std::runtime_error("store_x_freq <= 0");
```
maxentile, line 76: Should we also catch here if the user requests an x buffer that would be too big? (say if len(lambda_schedule) = 1e7 and store_x_freq = 1)

maxentile (approved):
Looks great!

timemachine/cpp/src/context.cu
proteneer: The limits are going to be GPU memory size dependent. These types of checks are best left on the application code side of things (there's lots of other ways to shoot one self in the foot). The default behavior here is not storing any frames, so I think we're pretty safe.

badisa (approved):

maxentile (approved):
Nice, +1 on including the `*_freq` -> `*_interval` change!

proteneer:
Thanks for reviews, squash merging.

PR #379, opened 2021-03-26
Feature/train validation split random

badisa:


examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -277,93 +299,141 @@ def unflatten(theta: array) -> Dict[Handler, array]:
...
-            loss=loss
-        )
+    step_inds = []
+    for epoch in range(num_epochs):
+        inds = np.arange(len(training.data))
+        np.random.shuffle(inds)
+        step_inds.append(inds)
 
-        # TODO: same for xs and du_dps snapshots
+    np.save(output_path.joinpath('step_indices.npy'), np.hstack(step_inds)[:args.param_updates])
```
badisa, line 315: step indices is now only the training steps, with the validation set never shifting order (correct me if I am wrong in thinking we don't need to shuffle the validation set)

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -277,93 +299,141 @@ def unflatten(theta: array) -> Dict[Handler, array]:
...
+
+            # save updated forcefield .py files after every gradient step
+            step_params = serialize_handlers(ff_handlers)
+            # TODO: consider if there's a more modular way to keep track of ff updates
+            _save_forcefield(output_path.joinpath(f"forcefield_checkpoint_{step}.py"), ff_params=step_params)
+            step += 1
+            if step >= args.param_updates:
+                break
+        if len(validation) > 0:
+            val_loss = np.zeros(len(validation))
```
badisa, line 413: Saving out significantly less for the validation, wasn't clear to me if we want all of the save information or not.

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: This sounds right

training/dataset.py
```diff
@@ -14,16 +16,82 @@ def __len__(self):
...
+        left = set(left)
+        right = set(right)
+        indices = set(range(len(self)))
+        if len(left.intersection(right)) > 0:
+            raise ValueError("Left and right indices contain overlap")
+        if len(left.union(right)) != len(indices) or len(left.union(right).difference(indices)) > 0:
+            raise ValueError("Indices provided don't match dataset indices")
+        left_split = [self.data[i] for i in left]
+        right_split = [self.data[i] for i in right]
+        return Dataset(left_split), Dataset(right_split)
```
maxentile, line 97: Nice!
* Could you mention in docstrings when to use `split` vs `random_split` vs `indice_split`?
* In addition to returning the result of the split, calling `dataset.random_split()` will modify `dataset` in place. Not sure if any later step might assume the order of items within `dataset` was the same as at initialization. If so, it might be safer instead to do something like `train_inds, validation_inds = ...; return Dataset(self.data[train_inds]), Dataset(self.data[validation_inds])`?

tests/test_training.py
```diff
@@ -0,0 +1,56 @@
...
+
+import numpy as np
+
+from training.dataset import Dataset
+
+
+class TestDataset(unittest.TestCase):
+
+	def test_split(self):
+		ds = Dataset(list(range(100)))
```
maxentile, line 11: Formatting: indentation

fe/model.py
```diff
@@ -80,7 +80,7 @@ def predict(self, ff_params: list, mol_a: Chem.Mol, mol_b: Chem.Mol, core: np.nd
             ("complex", self.complex_system, self.complex_coords, self.complex_box, self.complex_schedule),
             ("solvent", self.solvent_system, self.solvent_coords, self.solvent_box, self.solvent_schedule)]:
 
-            print("Minimizing the host structure to remove clashes.")
+            print(f"Minimizing the {stage} host structure to remove clashes.")
```
maxentile, line 83: Nice! This improves the informativeness of the terminal output

fe/free_energy.py
```diff
@@ -115,9 +115,10 @@ def prepare_host_edge(self, ff_params, host_system, host_coords):
 # this class is serializable.
 class RelativeFreeEnergy(BaseFreeEnergy):
 
-    def __init__(self, single_topology: topology.SingleTopology, label=None):
+    def __init__(self, single_topology: topology.SingleTopology, label=None, complex_path=None):
         self.top = single_topology
         self.label = label
+        self.complex_path = complex_path
```
maxentile, line 121: Nice! Question: is it better for this to be a path, or a different kind of object? (Perhaps making any other choice here would require refactoring `builders.build_protein_system` to accept something other than a path though...)

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: I think what you've done looks good. (No need to save params or gradient here, but it is helpful to save du_dls here, as you've implemented.) I think this has validation-set predictions computed after every epoch (rather than after every optimizer step), which I think makes sense as well.

maxentile (approved):
Nice work! Providing the option to compute validation-set predictions on the fly in this way -- rather than "manually" in a separate pass afterwards -- is a clear improvement

fe/free_energy.py
badisa: I decided on a path to avoid the pickle files being overly large and to avoid any potential issues with changing how we build protein systems. Will play with this.

maxentile: Makes sense!

training/dataset.py
```diff
@@ -14,16 +19,88 @@ def __len__(self):
...
+            The split datasets, the first dataset containing the fractional percentage
+            and the second with the leftover
+
+        """
+        inds = np.arange(len(self))
+        np.random.shuffle(inds)
+        split_idx = int(len(self.data)*frac)
+        return self.indice_split(inds[:split_idx], inds[split_idx:])
+
+    def indice_split(self, left: List[int], right: List[int]) -> Tuple["Dataset", "Dataset"]:
```
proteneer, line 76: index (singular) indices (plural) - for sake of my OCD please name this either index_split or indices_split

training/dataset.py
```diff
@@ -14,16 +19,88 @@ def __len__(self):
...
+            Indices of dataset to construct second dataset from.
+
+        Returns
+        -------
+        tuple(Dataset, Dataset)
+            The split datasets, the first dataset containing the objects at the indices
+            provided by left and the second with the right indices.
+
+        """
+        left = set(left)
```
proteneer, line 97: Do we want to check for uniqueness of the lists here? ie. len(set(right)) == len(right)? I can see arguments for both

training/dataset.py
```diff
@@ -14,16 +19,88 @@ def __len__(self):
...
+        right = set(right)
+        indices = set(range(len(self)))
+        if len(left.intersection(right)) > 0:
+            raise ValueError("Left and right indices contain overlap")
+        if len(left.union(right)) != len(indices) or len(left.union(right).difference(indices)) > 0:
```
proteneer, line 102: Just checking that we want to enforce that left union right == len(self)? (i.e. we probably don't want to ever construct a Datasets with elements we don't train or test over).

tests/test_training.py
```diff
@@ -0,0 +1,75 @@
...
+            self.assertAlmostEqual(len(rhs) / count, 1.0 - frac)
+            if len(lhs) and len(rhs):
+                self.assertNotEqual(lhs.data, ds.data[:len(lhs)])
+                self.assertNotEqual(rhs.data[0], ds.data[len(lhs)])
+        # Verify that original dataset didn't get shuffled in the process
+        for i in range(count):
+            self.assertEqual(ds.data[i], i)
+
+
+    def test_indice_split(self):
```
proteneer, line 49: excellent tests

proteneer (approved):

training/dataset.py
badisa: I believe so, which is why I check that the possible indices (`indices = set(range(len(self)))`) are equal to `left.union(right)`

PR #383, opened 2021-04-04
Restore centroid restraints

proteneer:
This PR restores the center of mass restraints that were disabled during the fixed point conversion PR. The mass argument has been intentionally removed (for now) to avoid subtle inconsistencies that may arise when used in conjunction with the RMSD-based rotational restraint.

This also fixes a singularity associated with `b0==0`, similar to the harmonic bond case.

### DIFF
--- a/tests/test_bonded.py
+++ b/tests/test_bonded.py
@@ -61,6 +61,59 @@ class TestBonded(GradientTest):
                 precision=precision
             )
 
+    def test_centroid_restraint_singularity(self):
+        # test singularity is stable when dij=0 and b0 = 0
+        box = np.eye(3) * 100
+
+        # specific to centroid restraint force
+        relative_tolerance_at_precision = {np.float32: 2e-5, np.float64: 1e-9}
+
+        n_particles = 5
+
+        for precision, rtol in relative_tolerance_at_precision.items():
+            x0 = self.get_random_coords(n_particles, 3)
+            coords = np.concatenate([x0, x0])
+
+            gai = np.arange(5).astype(np.int32)
+            gbi = (np.arange(5) + 5).astype(np.int32)
+
+            masses = np.random.rand(n_particles)
+
+            kb = 10.0
+            b0 = 0.0
+
+            ref_nrg = jax.partial(
+                bonded.centroid_restraint,
+                group_a_idxs=gai,
+                group_b_idxs=gbi,
+                kb=kb,
+                b0=b0
+            )
+
+            # we need to clear the du_dp buffer each time, so we need
+            # to instantiate test_nrg inside here
+            test_nrg = potentials.CentroidRestraint(
+                gai,
+                gbi,
+                kb,
+                b0
+            )
+
+            params = np.array([], dtype=np.float64)
+            lamb = 0.3  # doesn't matter
+
+            self.compare_forces(
+                coords,
+                params,
+                box,
+                lamb,
+                ref_nrg,
+                test_nrg,
+                rtol,
+                precision=precision
+            )
+
+
     def test_harmonic_bond(self, n_particles=64, n_bonds=35, dim=3):
         """Randomly connect pairs of particles, then validate the resulting HarmonicBond force"""
         np.random.seed(125)  # TODO: where should this seed be set?
--- a/tests/test_bonded.py
+++ b/tests/test_bonded.py
@@ -12,54 +12,54 @@ from timemachine.potentials import bonded
 
 class TestBonded(GradientTest):
 
-    # def test_centroid_restraint(self, n_particles=10, n_A=4, n_B=3, kb=5.4, b0=2.3):
-    #     """Randomly define subsets A and B of a larger collection of particles,
-    #     generate a centroid restraint between A and B, and then validate the resulting CentroidRestraint force"""
-    #     box = np.eye(3) * 100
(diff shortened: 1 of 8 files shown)


proteneer:
*do not merge* there is a bad test case I missed for when `k !=0, b==0, and dij != 0`

proteneer:
Corner case is fixed.

tests/test_bonded.py
```diff
@@ -12,54 +12,110 @@
...
+        generate a centroid restraint between A and B, and then validate the resulting CentroidRestraint force"""
+        box = np.eye(3) * 100
+
+        # specific to centroid restraint force
+        relative_tolerance_at_precision = {np.float32: 2e-5, np.float64: 1e-9}
+
+        for precision, rtol in relative_tolerance_at_precision.items():
+            x_primal = self.get_random_coords(n_particles, 3)
+
+            gai = np.random.randint(0, n_particles, n_A, dtype=np.int32)
```
maxentile, line 26: Note this will sample with replacement `n_A` times, so it's possible for `gai` to contain fewer than `n_A` unique points

timemachine/potentials/bonded.py
```diff
@@ -1,18 +1,26 @@
 import jax.numpy as np
 
-def centroid_restraint(conf, params, box, lamb, masses, group_a_idxs, group_b_idxs, kb, b0):
+def centroid_restraint(conf, params, box, lamb, group_a_idxs, group_b_idxs, kb, b0):
```
maxentile, line 4:
```suggestion
    """Computes kb  * (r - b0)**2 where r is the distance between the centroids of group_a and group_b
    
    Notes
    ------
    * Geometric centroid, not mass-weighted centroid
    * Gradient undefined when `r - b0 == 0` and `b0 != 0` (explicitly stabilized in case `b0 == 0`)
    """
```

Add docstring

timemachine/potentials/bonded.py
```diff
@@ -1,18 +1,26 @@
...
+    d2ij = np.sum(dx*dx)
+    d2ij = np.where(d2ij == 0, 0, d2ij) # stabilize derivative
+    dij = np.sqrt(d2ij)
     delta = dij - b0
 
-    return kb*delta*delta
+    # when b0 == 0 and dij == 0
+    return np.where(
+        b0 == 0,
+        kb * d2ij,
```
maxentile, line 20: Here `kb = 0.5 * force_constant` as an optimization?

(I note that [in `restraint`](https://github.com/proteneer/timemachine/blob/6e9a150bca9154442d924e4a6486a39ef27bf46a/timemachine/potentials/bonded.py#L74) the same convention is used but [in `harmonic_bond`](https://github.com/proteneer/timemachine/blob/6e9a150bca9154442d924e4a6486a39ef27bf46a/timemachine/potentials/bonded.py#L135) the `kb = force_constant` convention is used. )

timemachine/potentials/bonded.py
```diff
@@ -1,18 +1,26 @@
...
+        kb * np.power(dij - b0, 2.0)
```
maxentile, line 21:
```suggestion
        kb * delta**2
```

timemachine/potentials/bonded.py
```diff
@@ -1,18 +1,26 @@
 import jax.numpy as np
 
-def centroid_restraint(conf, params, box, lamb, masses, group_a_idxs, group_b_idxs, kb, b0):
+def centroid_restraint(conf, params, box, lamb, group_a_idxs, group_b_idxs, kb, b0):
```
maxentile, line 3: Note `params`, `box`, `lamb` also unused -- can these arguments be safely omitted here, or is it important for these to be retained (e.g. so that the signature partially matches that of `harmonic_bond`)?

timemachine/potentials/bonded.py
```diff
@@ -1,18 +1,26 @@
 import jax.numpy as np
 
-def centroid_restraint(conf, params, box, lamb, masses, group_a_idxs, group_b_idxs, kb, b0):
```
maxentile, line 3: Nice! Making the centroid definitions independent of the system masses is a good idea, I think.

Not necessarily something to adjust in the current PR, but rather than depending on physical `masses`, what would you think about having this depend on a normalized `weights` parameter?

maxentile (approved):
Nice work!

timemachine/potentials/bonded.py
maxentile: Ahh! Intended as documented here: https://github.com/proteneer/timemachine/pull/382/files#diff-d0b9851a09ca44d39e9ef2f006a931da3a68103d2a2781370f5157cb05ee4038R88-R95 . Marking resolved

tests/test_bonded.py
proteneer: Yes - this is intentional, since duplicate indices may appear in the RMSD Restraint.

timemachine/potentials/bonded.py
proteneer: I don't think there's a consistent form here. OpenMM Harmonic Bond and Angles both use 0.5*k, but periodic torsions use 1*k, and AMBER consistently uses 1*k across the board, so I just chose the simplest convention.

timemachine/potentials/bonded.py
proteneer: definitely doable, we just need to also support the same behavior in the RMSD restraint.

proteneer:
Thanks for the comments and prompt review.

PR #384, opened 2021-04-05
Optimized svd force stability tests

maxentile:
Addressing https://github.com/proteneer/timemachine/pull/382#issuecomment-813420233

Marking as draft, since there appears to be some mysterious GitHub behavior when I target the WIP branch. (Seems only to see commits up to 5 days ago, but the PR https://github.com/proteneer/timemachine/pull/382 of course has more recent commits.)

May be best to add this to main instead, after PR #382 is merged, rather than try to sort out this issue.

### DIFF
--- a/tests/optimize_testcases.py
+++ b/tests/optimize_testcases.py
@@ -1,0 +1,106 @@
+# Generate test cases that make the force magnitude large
+
+import numpy as np
+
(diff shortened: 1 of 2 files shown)


maxentile:
Will re-open this PR, except targeting main branch, after #382 is merged

PR #385, opened 2021-04-05
Prepare protein prior to building system

badisa:
* Simplifies workflow when dealing with new systems

### DIFF
--- a/md/builders.py
+++ b/md/builders.py
@@ -1,15 +1,91 @@
+from tempfile import NamedTemporaryFile
+
 import numpy as np
 from simtk import unit
 from simtk.openmm import app, Vec3
+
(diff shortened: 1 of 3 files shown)


md/builders.py
```diff
@@ -1,15 +1,91 @@
...
+from ff.handlers.bcc_aromaticity import openeye_log_wrapper
+
+
+def prepare_protein(path: str, output_path: str):
+    """
+    Prepares protein
+
+    Performs the following preparation:
+    * Add missing hydrogens
+    * Remove everything except for the protein
```
badisa, line 30: One thing I just realized is that it can also change the charge of the system. Not sure how to determine whether the charge assigned is better or worse. Will look at disabling it

Issue #386, opened 2021-04-07
record FMA compiler bug


proteneer:
Recording the FMA bug recently discovered here, so we can reference this issue in a patch. Thanks to @scottlegrand for identifying that the issue is because of the fma on denormed floats

``` cuda
#include <cstdio>

#define FIXED_EXPONENT             0x1000000000

void __global__ k_test(float a, float b, float c, float d, double *out) {


    float res = (a*b)+(c*d);

    unsigned long long cast = static_cast<unsigned long long>(llrint(res*FIXED_EXPONENT));

    printf("gpu a*b+c*d %llu\n", cast);

    out[0] = a+b+c+d;

}


void h_test(float a, float b, float c, float d) {

    float res = (a*b)+(c*d);

    unsigned long long cast = static_cast<unsigned long long>(llrint(res*FIXED_EXPONENT));

    printf("gpu a*b+c*d %llu\n", cast);

}


int main() {

    double *res;

    auto err = cudaMalloc(&res, sizeof(*res));#include <cstdio>

#define FIXED_EXPONENT             0x1000000000

void __global__ k_test(float a, float b, float c, float d, double *out) {


    float res = (a*b)+(c*d);

    unsigned long long cast = static_cast<unsigned long long>(llrint(res*FIXED_EXPONENT));

    printf("gpu a*b+c*d %llu\n", cast);

    out[0] = a+b+c+d;

}


void h_test(float a, float b, float c, float d) {

    float res = (a*b)+(c*d);

    unsigned long long cast = static_cast<unsigned long long>(llrint(res*FIXED_EXPONENT));

    printf("gpu a*b+c*d %llu\n", cast);

}


int main() {

    double *res;

    auto err = cudaMalloc(&res, sizeof(*res));

    // denorms
    float eps_j = 0.12814913690090179443;
    float deps_dl_i =-0.12525972723960876465;
    float eps_i = 0.12525972723960876465;
    float deps_dl_j = 0.12814913690090179443;

    k_test<<<1,1>>>(eps_j,deps_dl_i,eps_i,deps_dl_j,res);
    k_test<<<1,1>>>(eps_i,deps_dl_j,eps_j,deps_dl_i,res); // not bitwise identical

    cudaDeviceSynchronize();

    h_test(eps_j,deps_dl_i,eps_i,deps_dl_j);
    h_test(eps_i,deps_dl_j,eps_j,deps_dl_i); // not bitwise identical

    // no denorms case seems to be working fine
[...]

PR #387, opened 2021-04-07
Fix FMA compiler bug

proteneer:
Addresses  #386. This was one hell of a chase.


timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -351,6 +351,14 @@ float __device__ __forceinline__ real_es_factor(float real_beta, float dij, floa
     return -inv_d2ij*(static_cast<float>(TWO_OVER_SQRT_PI)*beta_dij*exp_beta_dij_2 + erfc_beta_dij);
 }
 
+float __device__ __forceinline__ fix_nvidia_fmad(float a, float b, float c, float d) {
```
bp-kelley, line 354: I think a comment is worth it here saying that you are fixing issues in the HW fused multiply and add

bp-kelley (changes requested):

bp-kelley:
Denormals are hard :)

timemachine/cpp/src/kernels/k_nonbonded.cuh
proteneer: Done!

proteneer:
Merging this.

maxentile:
Nice! And to confirm, are there other places where it is important to use `fix_nvidia_fmad(a, b, c, d)` in place of `a*b + c*d`? I.e. is it expected that checking for discrepancies between nonbonded `uimpl` with and without `.disable_hilbert_sort` would cause all instances of this issue to surface? While this issue is fresh, I would like to make sure we expand the tests here.

I would have expected the behavior now triggered by `test_fma_compiler_bug` also to be triggered here in `test_nblist_hilbert`  -- perhaps `test_nblist_hilbert` (and possibly other tests I'm not thinking of?) should loop over precisions `np.float32` and `np.float64`.

https://github.com/proteneer/timemachine/blob/3937fbcac54342a47419efad052c98cd64674b4f/tests/test_nonbonded.py#L82-L118 ,

proteneer:
I'm not convinced that this is a compiler bug specific to the form of `a_j*b_i + c_i*d_j` or not yet. 

The nblist hilbert test above is done with a regular Nonbonded potential, where the dp/dl is necessary zero and there's no way for the behavior to trigger `b_i and d_j` are always zero.

I will add a test though expanding coverage to both precisions.

maxentile:
>The nblist hilbert test above is done with a regular Nonbonded potential, where the dp/dl is necessary zero and there's no way for the behavior to trigger b_i and d_j are always zero.

Ahh, had missed that `nb_bp` from `tests/repro.pkl.gz` wasn't a regular Nonbonded potential, but that makes sense. (Otherwise the test would have used a regular constructor!)

>I'm not convinced that this is a compiler bug specific to the form of `a_j*b_i + c_i*d_j` or not yet.

The root cause here does seem murky to me. Seems unlikely to be specific to floats vs. doubles, seems unlikely to be specific to "denormalized floats," and seems unlikely to be specific to a single expression like "`a*b + c*d`" independent of context.

In principle, I guess any factor that influences the low-level instruction selection for the parts of the program that rely on bitwise exact cancellations might pose a risk of significantly changing the behavior of this code in the future, which I am still uncomfortable about. For example, changing subsequent lines that influence how repeated subexpressions are reused in a kernel (as in https://github.com/proteneer/timemachine/issues/386#issuecomment-815379478 ) can influence whether the compiler decides to use `fma` to compute that expression.

I don't know enough to assess this confidently, but if you think it's possible this part of the code relies on any similar assumptions about compiler behavior that aren't explicitly guaranteed by NVIDIA, I guess it would be prudent to be even more conservative in our testing here!

PR #388, opened 2021-04-07
Minimize before running lamb=0 in tests/test_benchmark.py:benchmark_hif2a

maxentile:
Lets minimizer resolve box boundary clashes


tests/test_benchmark.py
```diff
@@ -187,21 +187,17 @@ def benchmark_hif2a(verbose=False, num_batches=100, steps_per_batch=1000):
             cutoff=1.0
         )
 
-        host_conf = []
-        for x,y,z in host_coords:
-            host_conf.append([to_md_units(x),to_md_units(y),to_md_units(z)])
-        host_conf = np.array(host_conf)
+        # resolve host clashes
+        min_host_coords = minimizer.minimize_host_4d([mol_a, mol_b], host_system, host_coords, ff, host_box)
```
badisa, line 191: Is ln 197 still correct in the stage being apo? Not sure it matters for the benchmark

badisa (approved):

tests/test_benchmark.py
maxentile: `host_coords.shape == min_host_coords.shape`, although the optimization is informed by `mol_a`, `mol_b`, I think. Not sure if there's a good way to call the minimizer without passing in one or two ligands? (Passing in an empty list in place of `[mol_a, mol_b]` raises a value error.)

proteneer: doable in principle - only a minor change is probably needed in the minimizer code to support an empty list.

proteneer:
LGTM!

proteneer (approved):

proteneer:
Did we want to merge this?

PR #390, opened 2021-04-08
Improve test coverage over two sets of precision.

proteneer:
Per comments from the other PR.


proteneer:
Thanks for catching the inconsistent tests @maxentile this helped me identify two issues:

1) dangling `double` type in the double precision version of `fix_nvidia_fmad`
2) another case of `a*b+c*d` in the charge derivatives.

maxentile (approved):
Nice!

PR #391, opened 2021-04-08
Modify RMSD restraint to be over proper rotations

proteneer:
This PR modifies the RMSD restraint to always return a proper rotation matrix.

### DIFF
--- a/tests/test_bonded.py
+++ b/tests/test_bonded.py
@@ -117,7 +117,6 @@ class TestBonded(GradientTest):
                 )
 
 
-
     def test_rmsd_restraint(self):
         # test the RMSD force by generating random coordinates.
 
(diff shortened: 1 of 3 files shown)


tests/test_bonded.py
```diff
@@ -128,46 +127,121 @@ def test_rmsd_restraint(self):
...
+                    ref_u,
+                    test_u,
+                    rtol,
+                    atol=atol,
+                    precision=precision
+                )
+
+
+    def test_rmsd_restraint_singularity(self):
+        # test the RMSD force by generating random coordinates.
```
maxentile, line 179: Could we expand this test's docstring to describe which of [the two identified singularities](https://github.com/proteneer/timemachine/pull/391/files#diff-d0b9851a09ca44d39e9ef2f006a931da3a68103d2a2781370f5157cb05ee4038R131) it is intended to hit, and what the expected behavior is? 

Although it looks to me like this may not quite be testing what you intend, it's not immediately obvious to me how to repair it.

Since this uses `n_A=n_B=3` I think the correlation matrices will always be rank deficient -- and thus `ref_u = 0` based on the check you had added in the previous version. I think this condition is tripped not just for small perturbations of specially chosen `x0, x1` (as currently written), but for most any choice of 6 input coordinates.

(I.e. replacing [L225-L233](https://github.com/proteneer/timemachine/pull/391/files#diff-726b2c071a1c59bd5858e946421e900c75db412d33d202ecd3f285b42bda8e60R225-R233) with `xs = [np.random.randn(6, 3) for _ in range(20)]` doesn't change whether the test passes, and `ref_u` still evaluates to 0 for all `x` in `xs`.)

maxentile (changes requested):

proteneer:
Thanks for the comments - I've modified the tests to be explicit. One case is non-full rank correlation matrix as before, and the other case I've explicitly generated a "large force" conformer from the blow-up forces notebook you sent me.

maxentile:
Nice! I'd like to poke around a bit more to look for discontinuities before completing my review. Initially looks good.

maxentile:
It looks like energies and forces for this restraint potential are discontinuous at conformers `conf=vstack([x1, x2])` that cause `correlation_matrix(x1, x2)` to have repeated singular values, see https://gist.github.com/maxentile/94f367347bdb69015f8e07696fb9ab1f 

The computed force magnitudes at any conformer are probably now controlled, but energies and forces do look like "step-functions" near any conformers that trigger [this condition](https://github.com/proteneer/timemachine/pull/391/files#diff-d0b9851a09ca44d39e9ef2f006a931da3a68103d2a2781370f5157cb05ee4038R144).

Conformers that trigger this condition can be found by numerically optimizing for singular value degeneracy, which I had initially encoded as something like
```python
def loss(conf):
    x1, x2 = conf[:n], conf[n:]
    S = compute_spectrum_of_correlation_matrix(x1, x2)
    return np.min(np.abs(np.diff(S)))
```

proteneer:
That's correct - the step function behavior is not ideal but may be okay depending on how often it triggers. I'm going to dive a little deeper to see when/why the backprop derivatives blow-up.

proteneer:
After some offline discussion, it seems too dangerous to form the rotation matrix by anything other than the raw U@V_T. I've updated the PR so that the identity matrix is returned (and evaluates to a zero potential energy), if the rotation matrix is improper.

proteneer:
Unfortunately turning off the rotation entirely for the reflective case creates large discontinuities:

Screenshot from 2021-04-09 15-21-14

proteneer:
Closing this since we're now using rejection sampling

PR #392, opened 2021-04-08
docking improvements

schmolly:
mostly relative_docking improvements
fix test outdirs
fix dock_and_equilibrate bug


docking/dock_and_equilibrate.py
```diff
@@ -181,21 +181,23 @@ def dock_and_equilibrate(
...
                 str(step).zfill(len(str(insertion_steps))),
-                f"ins",
+                "ins",
             )
 
         if report.too_much_force(ctxt, lamb, host_box, combined_bps, u_impls):
             print("Not calculating work (too much force)")
             calc_work = False
-            break
+            continue
```
proteneer, line 190: Just checking that you intended to continue as opposed to break

schmolly: yep, that was the bug! there's one spot where it should break and a different spot where it should continue. The bug was that if one compound in an sdf failed, the rest of the compounds in the set wouldn't run at all

tests/test_docking.py
```diff
@@ -35,18 +35,18 @@ def test_pose_dock(self):
...
+        self.assertTrue(Path(f"{outdir}/43/43_pd_1000_guest.sdf").exists())
 
-        for f in Path("test_pd_outdir").glob("*/*.*"):
+        for f in Path(outdir).glob("*/*.*"):
             Path.unlink(f)
-        for d in Path("test_pd_outdir").glob("*"):
+        for d in Path(outdir).glob("*"):
             d.rmdir()
-        Path("test_pd_outdir").rmdir()
+        Path(outdir).rmdir()
```
proteneer, line 49: if the assertion/tests fail do the directories still get cleaned up? not sure if you need a context manager or a use a context managed tempfile/tempdir?

schmolly: good question! I'll have to google that

docking/rigorous_work.py
```diff
@@ -21,28 +21,34 @@
 DELETION_MAX_LAMBDA = 1.0
 MIN_LAMBDA = 0.0
 INSERTION_STEPS = 501
-DELETION_STEPS = 501
 EQ1_STEPS = 5001
```
maxentile, line 24: Would it be helpful to expose any other of these variables as CLI arguments? (Maybe `INSERTION_STEPS` and `EQ1_STEPS`?)

docking/rigorous_work.py
```diff
@@ -424,12 +445,20 @@ def main():
     parser.add_argument(
         "--no_outfiles", action="store_true", help="write no output pdb/sdf files"
     )
+    parser.add_argument(
+        "--deletion_steps", type=int, default=501, help="how many steps to delete over"
+    )
+    parser.add_argument(
+        "--num_deletions", type=int, default=10, help="number of deletions to do"
+    )
```
maxentile, line 453: Suggest to swap order of these to match the Python signature (although not technically required!)

docking/relative_docking.py
```diff
@@ -18,24 +27,19 @@
 
 from docking import report
 
+
```
maxentile, line 30: Can remove testsystems import on L26

maxentile (approved):
Nice work! These all look like clear improvements.

Replacing the `uuid`-named directory in tests with a `tempfile`-managed temporary directory, as @proteneer suggested, would probably be even a bit safer.

docking/rigorous_work.py
schmolly: might as well!

proteneer (approved):
@schmolly  did you want to merge this? I don't remember where we left off on this...

schmolly:
yes, I'm ready to merge! :)

proteneer:
Great - feel free to merge once tests pass!

PR #393, opened 2021-04-09
Task/optimize rmsd restraints

badisa:
Previously the implementation would slow down as the total number of particles increased due to copying d_x to host. The GPU implementation is approximately 60 microseconds with a system containing 32 atoms with 5 mapped atoms. Increasing the system to 20k atoms, it is the same. The original implementation would take about 40 microseconds for the 32 atom system, and at 20k atoms it would take a bit over 400 microseconds.

### DIFF
--- a/tests/test_bonded.py
+++ b/tests/test_bonded.py
@@ -128,55 +128,51 @@ class TestBonded(GradientTest):
         n_particles_a = 25
         n_particles_b = 7
 
-        relative_tolerance_at_precision = {np.float32: 2e-5, np.float64: 1e-9}
+        for _ in range(100):
+            coords = self.get_random_coords(n_particles_a + n_particles_b, 3)
 
-        for precision, rtol in relative_tolerance_at_precision.items():
+            n = coords.shape[0]
+            n_mapped_atoms = 5
 
-            for _ in range(100):
-                coords = self.get_random_coords(n_particles_a + n_particles_b, 3)
+            atom_idxs_a = np.arange(n_particles_a)
+            atom_idxs_b = np.arange(n_particles_b)
 
-                n = coords.shape[0]
-                n_mapped_atoms = 5
+            atom_map = np.stack([
+                np.random.randint(0, n_particles_a, n_mapped_atoms, dtype=np.int32),
+                np.random.randint(0, n_particles_b, n_mapped_atoms, dtype=np.int32) + n_particles_a
+            ], axis=1)
 
-                atom_idxs_a = np.arange(n_particles_a)
-                atom_idxs_b = np.arange(n_particles_b)
+            atom_map = atom_map.astype(np.int32)
 
-                atom_map = np.stack([
-                    np.random.randint(0, n_particles_a, n_mapped_atoms, dtype=np.int32),
-                    np.random.randint(0, n_particles_b, n_mapped_atoms, dtype=np.int32) + n_particles_a
-                ], axis=1)
+            params = np.array([], dtype=np.float64)
+            k = 1.35
+            lamb = 0.0
 
-                atom_map = atom_map.astype(np.int32)
+            for precision, rtol, atol in [(np.float64, 1e-6, 1e-6), (np.float32, 1e-4, 1e-6)]:
 
-                params = np.array([], dtype=np.float64)
-                k = 1.35
-                lamb = 0.0
+                ref_u = functools.partial(bonded.rmsd_restraint,
+                    group_a_idxs=atom_map[:, 0],
+                    group_b_idxs=atom_map[:, 1],
+                    k=k
+                )
 
-                for precision, rtol, atol in [(np.float64, 1e-6, 1e-6), (np.float32, 1e-4, 1e-6)]:
+                test_u = potentials.RMSDRestraint(atom_map, n, k)
 
-                    ref_u = functools.partial(bonded.rmsd_restraint,
-                        group_a_idxs=atom_map[:, 0],
-                        group_b_idxs=atom_map[:, 1],
-                        k=k
-                    )
-
-                    test_u = potentials.RMSDRestraint(atom_map, n, k)
-
-                    # note the slightly higher than usual atol (1e-6 vs 1e-8)
-                    # this is due to fixed point accumulation of energy wipes out
-                    # the low magnitude energies as some of test cases have
-                    # an infinitesmally small absolute error (on the order of 1e-12)
-                    self.compare_forces(
-                        coords,
-                        params,
-                        box,
-                        lamb,
-                        ref_u,
-                        test_u,
-                        rtol,
-                        atol=atol,
-                        precision=precision
-                    )
+                # note the slightly higher than usual atol (1e-6 vs 1e-8)
+                # this is due to fixed point accumulation of energy wipes out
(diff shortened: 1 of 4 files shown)


tests/test_bonded.py
```diff
@@ -128,55 +128,51 @@ def test_rmsd_restraint(self):
         n_particles_a = 25
         n_particles_b = 7
 
-        relative_tolerance_at_precision = {np.float32: 2e-5, np.float64: 1e-9}
+        for _ in range(100):
+            coords = self.get_random_coords(n_particles_a + n_particles_b, 3)
 
-        for precision, rtol in relative_tolerance_at_precision.items():
```
badisa, line 133: This just made the test slower and the values weren't used.

timemachine/cpp/src/rmsd_restraint.cu
```diff
@@ -74,87 +102,82 @@ void RMSDRestraint<RealType>::execute_device(
...
-    x2 = x2.rowwise() - h_b_centroid.transpose();
+    x1 = x1.rowwise() - h_a_centroid_mean.transpose();
+    x2 = x2.rowwise() - h_b_centroid_mean.transpose();
 
-    Eigen::MatrixXd c = x2.transpose() * x1;
+    // Keep Matrices in doubles to avoid bad behavior in float32
+    Eigen::Matrix<double, 3, 3> c = (x2.transpose().template cast<double> () * x1.template cast<double> ());
 
-    Eigen::JacobiSVD<Eigen::MatrixXd> svd(c, Eigen::ComputeFullU | Eigen::ComputeFullV);
+    Eigen::JacobiSVD<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic>> svd(c, Eigen::ComputeFullU | Eigen::ComputeFullV);
```
badisa, line 141: If the following matrix operations are done using float32, things misbehave. Everything up to this point is numerically the same, but afterwards it blows up.

timemachine/cpp/src/rmsd_restraint.cu
```diff
@@ -173,52 +196,41 @@ void RMSDRestraint<RealType>::execute_device(
...
+    Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic> rhs_1 = smat*(F.cwiseProduct(v.transpose()*v_a - v_a.transpose()*v))*v.transpose();
+    Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic> rhs_2 = smat_inv*v_a.transpose()*(eye - v*v.transpose());
+    Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic> rhs = u*(rhs_1 + rhs_2);
+    Eigen::Matrix<RealType, Eigen::Dynamic, Eigen::Dynamic> c_adjoint = (lhs + rhs).cast<RealType>();
+    // Eigen::RowMajor required to be able to copy out of device
+    Eigen::Matrix<RealType, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> x2_adjoint = (c_adjoint*x1.transpose()).transpose();
+    Eigen::Matrix<RealType, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> x1_adjoint = x2*c_adjoint;
+
+    // Reusing d_centroid_a_ and d_centroid_b_ for x1 and x2 adjoints
+    gpuErrchk(cudaMemcpy(d_centroid_a_, x1_adjoint.data(), B*3*sizeof(RealType), cudaMemcpyHostToDevice));
```
badisa, line 212: Reuses the centroid memory to update forces

timemachine/cpp/src/rmsd_restraint.cu
```diff
@@ -173,52 +196,41 @@ void RMSDRestraint<RealType>::execute_device(
...
+
+    RealType squared_term = term*term;
+
+    k_finalize<<<grid_blocks, tpb, 0, stream>>>(
+        d_atom_map_,
+        d_centroid_a_,
+        d_centroid_b_,
+        B,
+        k_,
+        squared_term,
```
badisa, line 223: Rather than passing a buffer, i pass the RealType directly. Seemed cleaner, though perhaps there is a performance impact?

tests/test_bonded.py
```diff
@@ -128,55 +128,51 @@ def test_rmsd_restraint(self):
...
+                self.compare_forces(
+                    coords,
+                    params,
+                    box,
+                    lamb,
+                    ref_u,
+                    test_u,
+                    rtol,
+                    atol=atol,
+                    precision=precision
```
proteneer, line 174: isn't the precision being used here?

tests/test_bonded.py
proteneer: yes the PR from the rmsd_patch branch also addressed this. I'll probably need to resolve a merge conflict with the RMSD patch later though, but I think the performance looks good.

timemachine/cpp/src/rmsd_restraint.cu
proteneer: Thats... concerning. Is it just the SVD itself that needs double? or everything afterwards? does doing a double precision SVD and single precision backprop work?

timemachine/cpp/src/rmsd_restraint.cu
```diff
@@ -173,52 +196,41 @@ void RMSDRestraint<RealType>::execute_device(
...
-    gpuErrchk(cudaMemcpy(d_du_dx_buf_, &du_dx[0], N*3*sizeof(*d_du_dx_buf_), cudaMemcpyHostToDevice));
-
-    const int blocks= (N_+32-1)/32;
-    const int tpb = 32;
-
-    k_finalize<<<blocks, tpb, 0, stream>>>(
-        N,
-        d_du_dx_buf_,
-        d_u_buf_,
+    Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic> v_a = v_a_t.transpose();
```
proteneer, line 199: I think some of these expression can be optimized a bit here more (eg. v_a_t.transpose() = v_a), but we can do this in a later pass if you'd like.

badisa: Happy to do that in this PR, especially since I will probably wait until you merge your updates to this logic anyways.

timemachine/cpp/src/rmsd_restraint.cu
badisa: Will dig into it and find where it breaks, the eigen templating was giving me a hard time

badisa: It was the SVD, everything else works as it should.

timemachine/cpp/src/rmsd_restraint.cu
badisa: Done in latest

timemachine/cpp/src/rmsd_restraint.cu
badisa: I had thought i had nailed it down, but looked like everything past the SVD needs to be double. Wasn't able to nail down what line was spitting out bad numbers.

proteneer:
Let's put this on-hold - the derivatives are not quite stable and we will revisit this at later point in time.

PR #394, opened 2021-04-14
Task/run validation pre training

badisa:


examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -311,8 +339,13 @@ def unflatten(theta: array) -> Dict[Handler, array]:
     np.save(output_path.joinpath('step_indices.npy'), np.hstack(step_inds)[:args.param_updates])
 
     step = 0
+    epoch = 0
     # in each optimizer step, look at one transformation from relative_transformations
-    for epoch, steps in enumerate(step_inds):
+    for steps in step_inds:
+        # Run Validation edges at start of epoch. Unlike NNs we have a reasonable starting
+        # point that is worth knowing
+        run_validation_edges(validation, ordered_params, systems, epoch)
```
badisa, line 347: Running the validation after training made it hard to determine the training without rerunning RBFE calculations.

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -333,15 +366,17 @@ def unflatten(theta: array) -> Dict[Handler, array]:
...
             print(f"at optimizer step {step}, loss={loss:.3f}")
 
             # check if it's probably okay to take an optimizer step on the basis of this result
             # TODO: move responsibility for returning error flags / simulation uncertainty estimates further upstream
-            blown_up = any(_blew_up(results) for results in results_this_step.values())
+            blown_up = False
+            for stage, results in results_this_step.items():
+                if _blew_up(results):
+                    blown_up = True
+                    print(f"step {step} blew up in {stage} stage")
```
badisa, line 379: Here just to make it clear where exactly the simulation blew up.

examples/hif2a/fit_to_multiple_rbfes.py
proteneer: +1, running validation before hand is the standard way to do this.

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -164,6 +164,36 @@ def unflatten(theta: array) -> Dict[Handler, array]:
     return theta, unflatten
 
 
+def run_validation_edges(validation: Dataset, params, systems, epoch):
+    if len(validation) <= 0:
+        return
+    val_loss = np.zeros(len(validation))
+    for i, rfe in enumerate(validation.data):
+        if getattr(rfe, "complex_path", None) is not None:
```
proteneer, line 172: This looks a little janky

fe/model.py
```diff
@@ -93,7 +93,7 @@ def predict(self, ff_params: list, mol_a: Chem.Mol, mol_b: Chem.Mol, core: np.nd
 
             x0 = coords
             v0 = np.zeros_like(coords)
-            box = np.eye(3, dtype=np.float64)*100 # note: box unused
+            # box = np.eye(3, dtype=np.float64)*100 # note: box unused
```
proteneer, line 96: We can probably just remove this line as opposed to commenting it out.

proteneer (approved):

examples/hif2a/fit_to_multiple_rbfes.py
badisa: It is, and can probably go away. Had it as I had older transformations that were generated that didn't have an associated path so I would default to the one provided by the script. The same logic happens for the training code

maxentile (approved):
Nice! This will allow us to address questions about the difference in performance before and after training without having to choose between the hassle of benchmarking the initial force field in a separate pass, or making the unsafe assumption that "performance before training" = "performance after 1 epoch"

PR #395, opened 2021-04-14
Ensure RDKit saves properties when pickling mols

badisa:
* Ensures we can access properties later on

### DIFF
--- a/examples/hif2a/generate_star_map.py
+++ b/examples/hif2a/generate_star_map.py
@@ -196,9 +196,13 @@ if __name__ == "__main__":
     forcefield = Forcefield(ff_handlers)
 
     mols = []
+    # Ensure we pickle up mol properties
+    Chem.SetDefaultPickleProperties(Chem.PropertyPickleOptions.AllProps)
     for lig_path in map_config.ligands:
         supplier = Chem.SDMolSupplier(lig_path, removeHs=False)
-        mols.extend(list(supplier))
+        # Prefer the list comp over casting to a list, seems to be inconsistent
+        # on OSX/linux
+        mols.extend([mol for mol in supplier])
 
     # In the future hopefully we can programmatically find the cores rather specifying
     cores = map_config.cores
--- a/timemachine/lib/potentials.py
+++ b/timemachine/lib/potentials.py
@@ -4,8 +4,6 @@ from timemachine.lib import custom_ops
 # (ytz): classes in this class wrap custom_ops but have the added benefit
 # of being pickleable.
 
-BoundPotential = custom_ops.BoundPotential
-
 class CustomOpWrapper():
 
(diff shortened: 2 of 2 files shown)


timemachine/lib/potentials.py
```diff
@@ -4,8 +4,6 @@
 # (ytz): classes in this class wrap custom_ops but have the added benefit
 # of being pickleable.
 
-BoundPotential = custom_ops.BoundPotential
```
proteneer, line 7: Is the removal of this alias intentional?

examples/hif2a/generate_star_map.py
```diff
@@ -196,9 +196,11 @@ def mcs_star_map(threshold: float = 0.5):
     forcefield = Forcefield(ff_handlers)
 
     mols = []
+    # Ensure we pickle up mol properties
+    Chem.SetDefaultPickleProperties(Chem.PropertyPickleOptions.AllProps)
     for lig_path in map_config.ligands:
         supplier = Chem.SDMolSupplier(lig_path, removeHs=False)
-        mols.extend(list(supplier))
+        mols.extend([mol for mol in supplier])
```
proteneer, line 203: I'm fine either way, but out of curiosity, does the list(supplier) break?

timemachine/lib/potentials.py
badisa: Woops

examples/hif2a/generate_star_map.py
badisa: It appears to work on linux, but breaks on OSX. Will add a comment. The latter seems more explicit and seems to generally be safer when it comes to C++ wrapped code in my experience

timemachine/lib/potentials.py
badisa: Actually looks like it isn't used. I had removed it to be able to run the `generate_star_map.py`. Do we use this alias?

maxentile (approved):
Nice! I didn't realize there was a setting for this. Will prevent dropping important mol properties like mol ID and affinity when pickling

timemachine/lib/potentials.py
proteneer: I think it's safe to drop this then. Was the idea that generate_star_map can be ran entirely without the custom_ops?

proteneer: I think at one point we had an instance(potentials.BoundPotential as an alias) - but its probably better to just directly reference the class in the custom_ops to make parts of the code more portable.

badisa: This was so I could run it on my mac without it, though hadn't expected to keep this change. Since it looks unused, I agree that we should reference classes via `custom_ops`

PR #397, opened 2021-04-21
Endpoint correction

proteneer:
This PR implements the endpoint correction for converting an intractable set of restraints to a tractable one using the RMSD via rejection sampling. This is jointly done with lots of help from @maxentile. I'll point out a couple of "conventions" that hopefully we can agree on directly in the commit.

### DIFF
--- a/fe/endpoint_correction.py
+++ b/fe/endpoint_correction.py
@@ -1,0 +1,218 @@
+import functools
+
+from typing import Tuple
+import jax
+import numpy as np
+import jax.numpy as jnp
+from timemachine.potentials import bonded, rmsd
+from scipy.stats import special_ortho_group
+
+def exp_u(rotation, k, beta):
+    return jnp.exp(-beta*rmsd.psi(rotation, k))
+
+exp_batch = jax.jit(jax.vmap(exp_u, (0, None, None)))
+
+def sample_multiple_rotations(k, beta, size):
+    num_batches = 500
+    batch_size = 10000
+
+    samples = []
+
+    for batch_attempt in range(num_batches):
+        Rs = special_ortho_group.rvs(3, size=batch_size)
+        tests = np.random.rand(batch_size)
+        M = np.pi**2 # volume of SO(3)
+
+        # (detailed explanation by jfass re: normalizing comments)
+        # In rejection sampling, we need an upper bound M on the ratio q_target(x) / q_proposal(x),
+        # so that q_target(x) / M q_proposal(x) <= 1 everywhere.
+        # In our case we can get a tight bound from looking at q_target and q_proposal independently.
+        #     q_target(x) = exp(-U(x)), and we know min_x U(x) = 0, so we know max_x q_target(x) = 1.
+        #     Our proposal is uniform over rotations, so q_proposal(x) is a constant.
+
+        # If we used normalized_q_proposal(x) = 1 / pi^2, then M=pi^2 would be an appropriate constant
+        # to use, guaranteeing q_target(x) / M normalized_q_proposal(x) <= 1 everywhere.
+
+        # We could just as well set q_proposal(x) = 1 and M=1, which has exactly the same effect.
+
+        # However, mixing and matching these could lead to errors.
+
+        # For example, it would be incorrect to use normalized_q_proposal(x) = 1 / pi^2, M=1, as we
+        # could now sample x where q_target(x) / M normalized_q_proposal(x) > 1. Setting M too small
+        # like this would impact the correctness of the sampler, but would also be easy to detect by
+        # runtime assertion.
+
+        # On the other hand, setting M larger than needed does not impact the correctness of the
+        # sampler, although it impacts the efficiency. I think a previous version of this code set M ~10x
+        # larger than it needed to be, by implicitly using q_proposal(x) = 1, and picking M = pi^2.
+        # Setting M=1 here doesn't change the correctness of the sampler, but increases its acceptance
+        # rate relative to that version.
+        M = 1
+        acceptance_prob = exp_batch(Rs, k, beta)/M
+        locations = np.argwhere(tests < acceptance_prob).reshape(-1)
+
+        samples.append(Rs[locations])
+        if sum([len(x) for x in samples]) > size:
+            break
+
+    result = np.concatenate(samples)[:size]
+    assert len(result) == size
+
+    return result
+
+
+def estimate_delta_us(
+    k_translation,
+    k_rotation,
+    core_idxs,
+    core_params,
+    beta,
+    lhs_xs,
+    rhs_xs):
+    """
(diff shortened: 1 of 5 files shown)


fe/endpoint_correction.py
```diff
@@ -0,0 +1,199 @@
...
+    rhs_xs: np.array [T, N, 3]
+        Samples from the non-interacting, fully unrestrained right hand state.
+
+    beta: 1/kT
+        inverse kT
+
+    Returns
+    -------
+    4-tuple
+        4-tuple array of lhs delta_us, rhs delta_us, sampled rotations, and sampled translations.
```
proteneer, line 73: I think this convention for lhs_du and rhs_du is now consistent with what @maxentile has in mind (and that of pymbar).

tests/test_endpoint_correction.py
```diff
@@ -0,0 +1,226 @@
...
+        if step % sampling_frequency == 0 and step > equilibrium_steps:
+            rhs_xs.append(x_t)
+
+    lhs_xs = np.array(lhs_xs)
+    rhs_xs = np.array(rhs_xs)
+
+    k_translation = 200.0
+
+    results = []
+    for k_rotation in [0.0, 50.0, 1000.0]:
```
proteneer, line 183: I think we can expand this later on to brute force over a grid of k_translation/k_rotation values to try and find the optimal overlap.

tests/test_endpoint_correction.py
```diff
@@ -0,0 +1,226 @@
...
+        lamb=None
+    )
+
+    core_idxs = []
+    core_params = []
+
+    for a in mol_a.GetAtoms():
+        if a.IsInRing():
+            core_idxs.append((a.GetIdx(), a.GetIdx() + mol_a.GetNumAtoms()))
+            # if the force constant is too high, no k/t combination can generate
```
proteneer, line 75: If we set the harmonic core restraints to too high of a value, then there's seems to be no suitable choice of k_translation/k_rotation that can maximize the overlap.

fe/endpoint_correction.py
```diff
@@ -0,0 +1,199 @@
...
+    sample_size = rhs_xs.shape[0]
+    rotation_samples = sample_multiple_rotations(k_rotation, beta, sample_size)
+    covariance = np.eye(3)/(2*beta*k_translation)
+    translation_samples = np.random.multivariate_normal((0,0,0), covariance, sample_size)
+
+    rhs_xs_aligned = []
+    rhs_du = []
+
+    for x, r, t in zip(rhs_xs, rotation_samples, translation_samples):
+        # this can be batched as well, at some point.
```
proteneer, line 135: I think this loop is now the bottleneck, will look into batching this via vmap tomorrow!

fe/endpoint_correction.py
```diff
@@ -0,0 +1,199 @@
...
+    num_batches = 500
+    batch_size = 10000
+
+    samples = []
+
+    for batch_attempt in range(num_batches):
+        Rs = special_ortho_group.rvs(3, size=batch_size)
+        tests = np.random.rand(batch_size)
+        M = np.pi**2 # volume of SO(3)
+        # acceptance_prob = exp_batch(Rs, k)/M
```
proteneer, line 25: @maxentile I think we agreed to simply set M=1, but just wanted to triple check this.

maxentile: Yep! I think there was just some ambiguity about where we were putting this constant.

In rejection sampling, we need an upper bound `M` on the ratio `q_target(x) / q_proposal(x)`, so that `q_target(x) / M q_proposal(x) <= 1` everywhere.

In our case we can get a tight bound from looking at `q_target` and `q_proposal` independently.
* `q_target(x) = exp(-U(x))`, and we know `min_x U(x) = 0`, so we know `max_x q_target(x) = 1`.
* Our proposal is uniform over rotations, so `q_proposal(x)` is a constant.

If we used `normalized_q_proposal(x) = 1 / pi^2`, then `M=pi^2` would be an appropriate constant to use, guaranteeing `q_target(x) / M normalized_q_proposal(x) <= 1` everywhere.

We could just as well set `q_proposal(x) = 1` and `M=1`, which has exactly the same effect.

However, mixing and matching these could lead to errors.

For example, it would be incorrect to use `normalized_q_proposal(x) = 1 / pi^2`, `M=1`, as we could now sample `x` where `q_target(x) / M normalized_q_proposal(x) > 1`. Setting M too small like this would impact the correctness of the sampler, but would also be easy to detect by runtime assertion.

On the other hand, setting M larger than needed does not impact the correctness of the sampler, although it impacts the efficiency. I think a previous version of this code set M ~10x larger than it needed to be, by implicitly using `q_proposal(x) = 1`, and picking `M = pi^2`. Setting `M=1` here doesn't change the correctness of the sampler, but increases its acceptance rate relative to that version.

proteneer: Thanks for the detailed explanation, I'm going to document this in the code directly.

maxentile (approved):
Great work! I think this sampling approach will allow us to get the best of two worlds regarding orientational restraints. During production MD, we can define orientation restraints that are convenient to run MD on but that introduce an unknown contribution to the free energy (e.g. using shape or a collection of harmonic bonds). Then, we can correct for this contribution in a postprocessing step that requires no MD, and that does not assume the endpoint restraint potential to be smooth (e.g. SVD-based restraints).

This separation could afford additional flexibility for designing orientational restraints with convenient properties, without worrying if they would introduce challenges when used as a force in production MD.

fe/endpoint_correction.py
```diff
@@ -0,0 +1,199 @@
...
+        acceptance_prob = exp_batch(Rs, k, beta)
+        locations = np.argwhere(tests < acceptance_prob).reshape(-1)
+
+        samples.append(Rs[locations])
+        if sum([len(x) for x in samples]) > size:
+            break
+
+    return np.concatenate(samples)[:size]
```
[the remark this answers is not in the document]
proteneer: Good call I think it should raise an exception.

fe/endpoint_correction.py
```diff
@@ -0,0 +1,199 @@
...
+        x_new[restr_group_idxs_a] = x_a
+        x_new[restr_group_idxs_b] = x_b
+        rhs_xs_aligned.append(x_new)
+
+    rhs_xs_aligned = np.array(rhs_xs_aligned)
+    rhs_du = delta_u_batch(rhs_xs_aligned)
+
+    return lhs_du, -rhs_du, rotation_samples, translation_samples
+
+def rmsd_align(x1, x2):
```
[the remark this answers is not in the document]
proteneer: good call, I will refactor this.

proteneer:
Thanks for the detailed comments!

PR #398, opened 2021-04-21
Fix test to support RDKit 2021.03.1

badisa:
* EmbedMolecule conf generation changes the charges a bit,
  this hardcodes the SDF definition


badisa:
Tests all pass with rdkit=2021.03.1

tests/test_handlers.py
```diff
@@ -290,10 +290,57 @@ def test_am1_ccc():
     smirks = [x[0] for x in patterns]
     params = np.array([x[1] for x in patterns])
     props = None
-
+    mol_sdf = """
```
badisa, line 293: New RDKit changes the conf generation

proteneer (approved):
good catch!

PR #399, opened 2021-04-21
Differentiable cycle closure

maxentile:
Uses [`cvxpylayers`](https://github.com/cvxgrp/cvxpylayers)'s new Jax backend to express a maximum-likelihood estimate of free energies, `fs`, as a differentiable function of a collection of relative free energy calculations, `simulated_rbfes`.

The likelihood currently implemented is a special case of the Gaussian noise model analyzed in [DiffNet](https://github.com/forcefield/DiffNet). (Something like `log_likelihood(rbfe_ij; fs, sigma_ij) = norm.logpdf(rbfe_ij, loc=fs[j] - fs[i], scale=sigma_ij)`, with `sigma_ij=1`, summed over all available RBFE calculations (`i`, `j`) .)

The maximum likelihood estimate is a collection of `fs` variables that maximizes this function, given fixed values of the problem parameters `rbfe_ij` and `sigma_ij`. cvxpylayers provides generic capabilities for differentiating the results of convex optimization problems like this with respect to the parameters governing the problem. In our setting, we would be interested in differentiating the maximum-likelihood estimate w.r.t. the individual simulated RBFEs. (Something like "compute how the MLE on all the nodes would change if the RBFE calculated on each edge were slightly different.")

Compared with DiffNet, where the emphasis is on differentiating a measure of the overall estimator quality w.r.t. allocation of simulation effort, the emphasis in this PR is on differentiating the MLE estimator itself w.r.t. the RBFE calculations. Since we can in turn differentiate each RBFE calculation w.r.t. force field parameters, this should allow us to differentiate losses that are defined in terms of cycle-closure-corrected free energies w.r.t. force field parameters. We also retain the flexibility to swap in other noise models as appropriate (e.g. heavy-tailed or correlated noise), without any model-specific work beyond writing down the noise model's likelihood function.


proteneer:
Nice! Some basic questions before I dive deeper into the implementation:

0) I noticed that we're not supporting variable gaussian noises (i.e. every RBFE estimate is predicted to have the same amount of error). I think this is okay for a first pass, but we may need to modify this later on.
1) Does this give the same results as that of the MLEstimate in DiffNet up to a constant offset (assuming constant noise, no diagonsl, etc.)? We should maybe add a test for this?
2) Should including diagonal comparison inds (0,0), (1,1), etc throw an exception since the documentation states that ABFE estimates are not supported?
3) It looks like we're not supporting the ability to feed in a priori values/uncertainties deltaGs (which correspond to experimental values), and this is fine. I always viewed using a priori information as sort of cheating
4) It looks like you're verifying the correctness of the gradient by seeing if you can optimize with it, is there a less "intense" way of verifying the correctness? (eg. via finite difference or something?)

requirements.txt
```diff
@@ -10,3 +10,4 @@ jax==0.2.9
 jaxlib==0.1.59
 hilbertcurve==1.0.5
 networkx==2.5
+cvxpylayers==0.1.5
```
proteneer, line 13: note to self to ensure that whatever future version of jax we upgrade to is compatible with this version of cvxpylayers...

maxentile:
0. Wrote with this in mind, but had initially omitted for clarity. Addressed in https://github.com/proteneer/timemachine/pull/399/commits/ed7e02bbe4d79cb21b6b048713d74635595cfab8 by deleting "sigma = 1" and adding vector of sigmas to constructor.

1. It had better! We check that we exactly recover random ground-truth `fs` when presented noiseless `fs[j] - fs[i]` comparisons. Could add a test that the ground-truth `fs` also corresponds to MLE returned by DiffNet on these instances, although I don't immediately see much added value for the current testing purposes.

2. Good catch, self-comparisons `rbfe(i, i) = nonzero` would indeed be invalid input, for which the constructor should raise an error. Addressed in https://github.com/proteneer/timemachine/pull/399/commits/9e1c8cf0ce0becb42128dad6b0b758ad654ee7aa .

3. Wrote with this in mind, but would prefer to omit from this PR for clarity. To support this in the future, we would replace `argmax_{trial_fs} log_likelihood(trial_fs)` with `argmax_{trial_fs} log_prior(trial_fs) + log_likelihood(trial_fs)`, for some suitable definition of log_prior. There are many directions to go with the definition of log_prior, but as long as it is convex in trial_fs, adding it should not require major surgery (just modify [this line](https://github.com/proteneer/timemachine/blob/9e1c8cf0ce0becb42128dad6b0b758ad654ee7aa/fe/cycles.py#L99)).

4. [This](https://github.com/proteneer/timemachine/blob/4e74fc866e2d21a8256ffe9cc2ae684cf54989af/tests/test_cycle_closure.py#L52-L88) is indeed a pretty demanding integration test that several things are working together properly for the intended use case (gradients of a cycle-closure-corrected loss can be computed, a resulting optimization problem is solvable from a random starting point, the solution of this optimization problem is self-consistent up to machine precision, ...) and the test runs in about 250 milliseconds locally. Added ~half a dozen finite-difference grad_checks in https://github.com/proteneer/timemachine/pull/399/commits/70db7b4477b5b7e040f3dafbc42fd5b8b8d6c633 , which together take about 400 milliseconds locally.

proteneer:
> Good catch, self-comparisons rbfe(i, i) = nonzero would indeed be invalid input, for which the constructor should raise an error. Addressed in 9e1c8cf .

One more nitpick, what is the behavior if
1) duplicate values are provided? 
2) both ji and ij are provided and ddG_ij != - ddj_ji

maxentile:
Repeated independent measurements of a pair (i, j) are fine for this implementation (in contrast to DiffNet).

Here, if a pair (i,j) appears multiple times in comparison_inds, this will be treated as multiple independent observations of the underlying difference fs[j] - fs[i], corrupted by different noise.

```python
n_nodes = 2

true_fs = np.array([0,5])

n_comparisons = 1000
comparison_inds = np.array([[0,1]]*n_comparisons)
print(f'{n_comparisons} independent simulations of pair {comparison_inds[0]}')
true_rbfes = onp.array([true_fs[j] - true_fs[i] for (i, j) in comparison_inds])
sigmas = np.ones(len(comparison_inds))

# independent Normal(0, sigma) "measurement noise"
onp.random.seed(2021)
noise = onp.random.randn(len(true_rbfes)) * sigmas
simulated_rbfes = true_rbfes + noise

predict_fs = construct_mle_layer(n_nodes, comparison_inds, sigmas)
fs = predict_fs(simulated_rbfes)
fs -= fs[0]

print('true fs', true_fs)
print('predicted fs', fs)
```

```
1000 independent simulations of pair [0 1]
true fs [0 5]
predicted fs [0.         5.00009006]
```

For the currently implemented case of independent Gaussian observations case, repeated independent measurements of the same pair (i,j) will have the same effect as a single measurement with a different mean and smaller variance. However, for other likelihood models, the effect could be more complicated.

tests/test_cycle_closure.py
```diff
@@ -0,0 +1,202 @@
...
+    predict_fs = construct_mle_layer(n_nodes, rbfe_inds)
+    fs = predict_fs(simulated_rbfes)
+
+    assert (np.isclose(true_fs, fs).all())
+
+
+def test_deadlock_triangle(verbose=True):
+    """Construct a 3-cycle with a set of edge labels and initial edge predictions that could result in a cancellation of
+    gradients in some situations. Assert that the gradient of the cycle-corrected edge loss is non-zero for this set of
+    initial predictions, and assert that this loss can be minimized."""
```
proteneer, line 53: nice!

fe/cycles.py
```diff
@@ -0,0 +1,155 @@
...
+    inds_l, inds_r = rbfe_inds.T
+    if (inds_l == inds_r).any():
+        raise AssertionError(f'invalid rbfe_inds -- includes self-comparisons: {rbfe_inds[(inds_l == inds_r)]}')
+
+    # check that the "map" is connected
+    if len(set(rbfe_inds.flatten())) != n_nodes:
+        present = set(rbfe_inds.flatten())
+        expected = set(np.arange(n_nodes))
+        missing_inds = list(expected.difference(present))
+        raise AssertionError(f'invalid rbfe_inds -- missing comparisons for nodes: {missing_inds}')
```
proteneer, line 102: Is this a hard requirement? I think for many datasets this won't be necessarily true.

proteneer (approved):

proteneer:
This PR looks great! Once tests pass feel free to merge.

fe/cycles.py
maxentile: On second thought, I think we might need to meet an even more stringent condition than the one I actually checked here!

If we have `rbfe_inds = [[0,1], [2,3]]`, `simulated_rbfes = [1, 1]`, then any `fs = [0, 1, offset, offset + 1]` for arbitrary value of `offset` would equally well maximize the likelihood of `simulated_rbfes`. The information provided by `simulated_rbfes` provides no basis for comparing connected component `[0,1]` with connected component `[2,3]`, and I think we would not want to return an answer `fs` that could arbitrarily corrupt the between-component comparisons `fs[2] - fs[1]`, `fs[2] - fs[0]`, `fs[3] - fs[1]`, `fs[3] - fs[0]`.

To avoid this issue, I think it only makes sense to call this function on comparison networks containing a single connected component. In cases where the comparison network is disconnected, but we still wish to say something about comparisons _within_ each connected component, we should analyze each component independently, rather than altogether.

maxentile: Possibly addressed in https://github.com/proteneer/timemachine/pull/399/commits/7e4ca4890feb72638d814511e129e64ee282d322 ?

maxentile: Added a utility function to get connected components https://github.com/proteneer/timemachine/pull/399/commits/4d36492598730001dccc65e58a09f45910b7f4b9 .

If we have a bag of relative and absolute calculations that contains isolated "islands" (e.g. if we had two non-overlapping star maps, with no absolute calculations included), this function would tell us to analyze the two non-overlapping star maps independently. If we had at least one absolute calculation in each non-overlapping star map, then this function would tell us it's okay to analyze them jointly.

jchodera:
This would be awesome if it works stably! 

Dumb question: If you assume the changes in the $\sigma_{ij}$ are negligible in the parameter derivatives, isn't the gradient of the vector of free energies `x = pinv(F) * z` just `dx/dtheta = pinv(F) * dz/dtheta` since `F` only contains `\sigma_{ij}`? That is, isn't the gradient you seek just linear in `z`, which contains the free energies? Do you really need all this machinery?

maxentile:
>This would be awesome if it works stably!

Thanks! In the random tests so far it appears numerically stable, but I have not yet stress-tested this very stringently. Do you have suggestions for instances that would be likely to trigger numerical instabilities you'd be most concerned about here?

(Perhaps picking sigmas such that the condition number of the Fisher matrix `cond(F)` is large?)

>Dumb question: If you assume the changes in the $\sigma_{ij}$ are negligible in the parameter derivatives, isn't the gradient of the vector of free energies x = pinv(F) * z just dx/dtheta = pinv(F) * dz/dtheta since F only contains \sigma_{ij}? That is, isn't the gradient you seek just linear in z, which contains the free energies? 

Good question!

I decided to go with `cp.Minimize(- log_likelihood)` rather than the closed-form solution `dot(pinv(F), zs)` because there appeared to be potential benefit in terms of modeling flexibility, at little cost in terms of implementation complexity.

I think the `dot(pinv(F), zs)` approach would no longer be applicable if our edge or node likelihoods weren't all assumed i.i.d. Gaussian. However, the `cp.Minimize(- log_likelihood)` approach would later allow us to swap in any choice of likelihood such that `-log_likelihood` is still convex, without requiring us to provide a closed-form solution.

In particular, this allows to consider later extensions such as:
* Relaxing assumption that simulation errors are zero-mean Gaussian, e.g. assuming certain kinds of simulations are biased, or assuming simulation errors are heavier-tailed. If we had directional work values on the edges, I think we could also replace Gaussian likelihood with BAR likelihood, if we wanted.
* Relaxing assumption that edge errors are independent. (For example, if some edge estimates reuse starting structures or endpoint samples.)
* Replacing `log_likelihood` with `log_prior + log_likelihood`, where `log_prior` is again not restricted to be Gaussian.

Later experimenting with extensions like this would not require any model-specific work beyond writing down the model, because we've outsourced the responsibility of actually solving the model.

>Do you really need all this machinery?

I think the bulkiest parts of this PR are things I would want whether I used `cp.Minimize(- log_likelihood)` or `dot(pinv(F), zs)` (such as input validation).

Replacing `cp.Minimize(- log_likelihood)` with `dot(pinv(F), zs)` would indeed let me delete the lines that define `log_likelihood`, but would require me to add similar lines to construct `F` and `zs`. For some cases that the `cp.Minimize` approach handles gracefully (such as directed edges, multiple measurements per edge, or large sparse comparison networks), some additional case-specific code would be required to convert the inputs into the format of `F` and `zs` (e.g. we would need logic to collapse a collection of Gaussian measurements for directed edges (i,j), (j,i) into a single Gaussian measurement on undirected edge (i,j)). Overall I think the net number of lines would be similar.

requirements.txt
badisa: That would make a good comment actually

fe/cycles.py
```diff
@@ -0,0 +1,155 @@
...
+
+from fe.utils import validate_map
+
+from typing import Optional
+
+
+def construct_mle_layer(n_nodes: int,
+                        rbfe_inds: np.array, rbfe_sigmas: Optional[np.array] = None,
+                        abfe_inds: Optional[np.array] = None, abfe_sigmas: Optional[np.array] = None) -> callable:
+    """Construct a differentiable function predict_fs(simulated_rbfes, simulated_abfes) -> fs
```
badisa, line 19: Does `fs` here stand for free energies?

maxentile: The naming in this function could stand to be improved -- probably best to get this right early!

Would something like `rbfe_inds` -> `relative_inds`, `abfe_inds` -> `absolute_inds`, `fs` -> `free_energies` be an improvement? I'm also considering to replace the paired arguments `(rbfe_inds, rbfe_sigmas)` with an "RBFEInfo" object, `(abfe_inds, abfe_sigmas)` with an "ABFEInfo" object, or similar, which may improve readability.

requirements.txt
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/399/commits/6afd987a2cb955d5e773b3a59303de425a6210a8

fe/cycles.py
badisa: The only one that was opaque to me was `fs`, the others seem in our wheel house already

badisa (approved):

fe/cycles.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/399/commits/d604200579747051844e9f3ba9fa4bb4ed9c6c90

maxentile:
Thank you for the helpful comments and reviews! Will merge in a moment, and possibly revisit to design more "adversarial" tests if we get more suspicious that some important source of numerical instability was missed in the initial randomized tests.

PR #400, opened 2021-04-23
Clean-up estimator code

proteneer:
This PR cleans-up the estimator code in preparation for much more involved logic regarding multi-substage protocols since I was struggling trying to figure out how to incorporate endpoint correction dGs, caching etc. Namely, the generation of the deltaGs and the deltaG_grads (and the backprop logic0 is now separated from the distributed execution of the simulation itself. 

Some niceties:

1) `jax.lax.stop_gradient` is used to stop `sys_params` during simulation.
2) the results of the simulation(s) are then used to explicit form a list of dGs and dG_grads, which may be comprised of multiple sub-stages. 
3) no more need for dataclasses or simulation results objects via the `@dataclass` decorator, we can directly pass back-out the results.


fe/estimator.py
```diff
@@ -1,170 +1,28 @@
...
-
-    for result in results:
-        # (ytz): figure out what to do with stddev(du_dl) later
-        mean_du_dls.append(np.mean(result.du_dls))
-        all_grads.append(result.du_dps)
-
-    dG = np.trapz(mean_du_dls, model.lambda_schedule)
     dG_grad = []
-    for rhs, lhs in zip(all_grads[-1], all_grads[0]):
+    for lhs, rhs in zip(model.dG_grads[0][0], model.dG_grads[-1][-1]):
```
maxentile, line 21: Document expected nesting of `model.dG_grads`

maxentile (commented):
This seems like a step in a direction I'm quite happy with, keeping more separation between the expensive distributed computations required to collect samples vs. the differentiable functions computed on those samples.

I have some residual confusion about the new signatures in `fe/estimator.py`, in particular if `sys_params` is ever used there now? Probably should also check in `tests/test_estimator.py` and any application scripts that use `RBFEModel` for appearance of `has_aux=True` etc., and update these to match the new signatures and return types.

fe/estimator.py
```diff
@@ -1,170 +1,28 @@
...
-
 FreeEnergyModel = namedtuple(
     "FreeEnergyModel",
-    ["unbound_potentials", "client", "box", "x0", "v0", "integrator", "lambda_schedule", "equil_steps", "prod_steps"]
+    ["dGs", "dG_grads"]
 )
 
 gradient = List[Any] # TODO: make this more descriptive of dG_grad structure
 
 def _deltaG(model, sys_params) -> Tuple[Tuple[float, List], np.array]:
```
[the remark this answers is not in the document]
proteneer: This is needed to properly compute the gradient, otherwise backpropagation would break. deltaG is still a function that takes in `sys_params` and returns a `dG`, and is just `dG/dsys_params`. If `sys_params` were not an argument, then the gradient isn't even defined!

fe/estimator.py
```diff
@@ -178,8 +36,7 @@ def deltaG_bwd(model, residual, grad) -> Tuple[np.array]:
     """Note: nondiff args must appear first here, even though one of them appears last in the original function's signature!
     """
```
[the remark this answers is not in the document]
proteneer: Willfix

fe/estimator.py
proteneer: willfix

proteneer:
> I have some residual confusion about the new signatures in fe/estimator.py, in particular if sys_params is ever used there now?

To elaborate a little more, the way that jax implements the vjp is through transposition of the jvp. When `jax.grad` is called on a function, the differentiated parameters (argnums in jax.grad) are traced along its sequence of transformations. The traced value must lead to the output of the function. If the chain of `L=f(g(h(...k(p)))` breaks, then the derivative `dL/dp` is no longer computable. We break the chain with `jax.lax.stop_gradient`, but restore it later through a dummy function `deltaG` that implements the thermodynamic derivative. The function signature is `deltaG(sys_params) -> scalar`. I designed the signature in mind with multi-stage protocols, where the deltaG of a given stage is comprised of multiple sub-stages, and each sub-stage's deltaG is a function of `sys_params`
```
stage: type
complex0: conversion,
complex1: decoupling,
complex2: endpt correction
```
Each stage is also associated with a list of dU/dsys_params:
```
stage: derivatives
complex0: du/dps of size num_lambda_windows,
complex1: du/dps of size num_lambda_windows, 
complex2: du/dps of size 2 
``` 
The thermodynamic derivative is implemented as the last stages's last window subtracted by the first stage's first window. As opposed to introducing additional error by hoping that intermediate, bordering windows cancel out their du_dps (i.e. hoping that last lambda window of complex0 has equal du_dp to that of first window of complex1). Mechanically, it also makes it easier for us to write distributed code later on, without having to worry about the intricacies of the vjp system (and dealing with stuff like `aux=True`). 

> Probably should also check in tests/test_estimator.py and any application scripts that use RBFEModel for appearance of has_aux=True etc., and update these to match the new signatures and return types.

Willfix

proteneer:
PS I'm not fully happy with this design yet - there are some kinks I don't know how to resolve, eg. how to avoid compute du/dp if we're only doing inference. Previously it was possible to trigger a special case when going through _deltaG_fwd, but now it's not so clear.

proteneer:
I've also verified that this change implements the gradient correctly over a training run:

```
epoch 0 loss 15.900586568460632
epoch 1 loss 16.848918690021492
epoch 2 loss 14.681580014999366
epoch 3 loss 14.033311147514318
epoch 4 loss 15.916431833336908
epoch 5 loss 14.461758360124852
epoch 6 loss 12.928816789834254
epoch 7 loss 12.772812854007531
epoch 8 loss 11.780748201299982
epoch 9 loss 10.373853521621339
epoch 10 loss 9.796049498478514
```

proteneer:
Found a trick for dealing with dynamically detecting if we're tracing or just doing regular inference:

```
# dynamically detect if we need derivatives or not.
assert np.all([isinstance(x, jax.interpreters.ad.JVPTracer) for x in sys_params])

compute_grad = False

for x in sys_params:
    if isinstance(x, jax.interpreters.ad.JVPTracer):
        compute_grad = True
        break
```

fe/model.py
```diff
@@ -104,19 +209,66 @@ def predict(self, ff_params: list, mol_a: Chem.Mol, mol_b: Chem.Mol, core: np.nd
                 seed
             )
 
+            # dynamically detect if we need derivatives or not.
+            assert np.all([isinstance(x, jax.interpreters.ad.JVPTracer) for x in sys_params])
```
badisa, line 213: Do we need this line and the loop on 217? Does this not imply the second unless sys_params is empty??

badisa:
Do we need to change the `fit_to_multiple_rbfe.py` example to work with these changes since SimulationResult is no longer returned?

fe/model.py
proteneer: yeah this is buggy - i removed this assertion. I'm going to rework this logic a little more - I'm not super happy with run-time reflection to see if we need to compute derivatives or not.

proteneer:
Closing this one for now, not totally happy about the treatment of the gradients, and with the differentiable re-weighting layer we may change our strategy more so later on. So I'd rather we do this correctly.

PR #404, opened 2021-04-27
Differentiable reweighting

maxentile:
This is essentially a stripped-down version of `pymbar.MBAR.computePerturbedFreeEnergies`, organized to allow Jax to compute derivatives of a perturbed free energy estimate w.r.t. `params` automatically if it knows how to compute derivatives of the potential energy function w.r.t. `params`. Notably, this requires the implementation of no free-energy-specific custom derivative rules!

Given cached samples `x_k` from several reference thermodynamic states, we can use MBAR to represent the samples as a collection of weighted samples from a single big mixture of the reference states, summarized as a "flat" list of samples `xs` and accompanying `reference_log_weights`.

We can then estimate free energies and expectations in new, unsampled thermodynamic states `u_fxn(x, lam, params)`, by computing importance weights `log_importance_weight(xs[i], lam, params) = - u_fxn(xs[i], lam, params) - reference_log_weights[i]` for the cached samples in the new unsampled state w.r.t. the sampled mixture. Note that the "numerator" here is a differentiable function of `params`, and the "denominator" here has been precomputed once.

Assuming that `u_fxn(x, lam, params)` is differentiable w.r.t. `params`, these importance weights are also differentiable w.r.t. `params`, and so too are any resulting estimates of free energies and expectations.

This generic approach of expressing free energies etc. in terms of differentiable importance weights computed on cached samples complements the current approach of expressing free energies etc. in terms of distributed simulators that compute custom derivative rules.

The main immediate motivation is to allow us to take more than one parameter optimization step using the configurations sampled by an expensive free energy calculation.

The main concerns are computational cost and estimator reliability.
* Computing the importance weights naively requires `2 * len(xs)` calls to the potential energy function. This PR makes no attempt to reduce the cost of computing a vector of potential energies, since this is rarely going to exceed the cost of simulating from scratch. (However, strategies are available to reduce the number of calls, and to reduce the cost of each call, should this become a blocker.)
* An estimator that relies on reweighting samples from a reference distribution to a target distribution will become unreliable if the target and reference distributions are very different from each other. To inform assessment of estimator reliability, this PR prints a user warning if the estimated number of effective samples available for reweighting drops below a threshold. (Notably, this estimate of the effective sample size is itself a differentiable function of `params`, which @wiederm and I have discussed using to automatically "steer away from" distributions that would deviate too far from the reference distribution.)

Later extensions could allow:
* Gradient-based protocol optimization for TI, by expressing the TI estimator variance of a trial lambda schedule in terms of differentiable importance weights.
* Gradient-based optimization of intermediate states for MBAR, by expressing the MBAR overlap matrix for a collection of trial intermediate states in terms of differentiable importance weights.
* More efficient computation of importance weights `vmap(u_fxn, in_axes=(0, None, None))(xs, lam, params)` at new `params` by also caching pair lists or other information associated with the cached samples `xs`.


jchodera:
Super excited to see this! 🚀

fe/reweighting.py
```diff
@@ -0,0 +1,138 @@
...
+        compute u_fxn(x, lam=0, params), u_fxn(x, lam=1, params) and their gradients w.r.t. params on all cached xs.
+
+        References
+        ----------
+        * Shirts MR and Chodera JD. Statistically optimal analysis of samples from multiple equilibrium states.
+            J. Chem. Phys. 129:124105, 2008. http://dx.doi.org/10.1063/1.2978177
+        * Shirts MR. Reweighting from the mixture distribution as a better way to describe the Multistate Bennett Acceptance Ratio.
+            arXiv preprint, 2017. https://arxiv.org/abs/1704.00891
+        * pymbar implementation of computePerturbedFreeEnergies , which has an associated uncertainty estimate as well!
+            https://github.com/choderalab/pymbar/blob/3c4262c490261110a7595eec37df3e2b8caeab37/pymbar/old_mbar.py#L1106-L1212
```
badisa, line 55: We already discussed this, just a reminder to point to the `mbar.py` implementation rather than `old_mbar.py`

maxentile: Addressed in https://github.com/proteneer/timemachine/pull/404/commits/8b81efe1d8e92122cb1cda2107bd58930c5fadd5

maxentile:
Based on discussions with @badisa and @proteneer , there are a few usage examples that would be helpful at add, some of which have been investigated in the context of [a 1D testsystem here](https://github.com/proteneer/timemachine/blob/f6a795807ec308722e6bcba3da30f6e3f27b6372/examples/free_energy_gradients_comparison/comparison_on_1d_system.ipynb) :

- [x] Use cached samples to form a deterministic, differentiable approximation of a free energy based loss function, and minimize this loss using generic methods
- [x] Use reweighting as a lower variance estimator of value and grad from fresh samples, without reusing cached samples
- [x] Check the effective sample size available for estimating value and grad at a new point in parameter space, given cached samples at an initial point in parameter space

Are there other important aspects we should investigate already in a 1D example?

Next steps are to add usage examples based on more realistic test systems:

- [ ] Minimize a loss function given by the RBFE edge analyzed in `rbfe_single.py`
- [ ] Minimize a loss function defined in terms of multiple RBFE edges, including cycles
- [ ] (Stretch goal: Minimize a loss function given by a single challenging RBFE that has large variance, currently hampering optimization)

Doing so will require a thin Jax-y wrapper around a collection of timemachine unbound potentials to produce a "pure" function with the desired signature. (The examples in the notebook are written in a generic way, assuming access to a differentiable function with signature `u_fxn(x, lam, params) -> float`. The 1D-specific parts of the notebook are mainly the visualizations and "ground-truth" reference computations.)

maxentile:
Closing for now to reduce branch-indexing CI load, will reopen when internal validation is complete

Issue #409, opened 2021-05-02
Debug poor cancellation of NaNs


proteneer:
I've encountered another case of poor cancellation of NaNs - will provide a more detailed repro come Monday

proteneer:
Resolved.

PR #410, opened 2021-05-03
Add bad test for du_dl

proteneer:
This is the repro for #409. The behavior only affects du_dl, and only in single precision (double precision is fine).


proteneer:
Furthermore, turning off the hilbert ordering this time around has no effect.

proteneer:
Fixed. This was pretty tricky. This is due to:

1) two particles i,j are extremely close to each other, if this computation proceeds it would result in a nan.
2) in v_nonbonded computation, their du_dl is guaranteed to be zero via the is_vanilla check, so the computation is bypassed, and the nan is skipped from accumulation via the template `ALCHEMICAL=false`.
3) however, the exclusions did not check for this, and proceeds with computation, and a nan is accumulated.

The PR ensures that the the exclusions are computed consistently with v_nonbonded

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -901,6 +901,19 @@ void __global__ k_nonbonded_exclusions(
 
     unsigned long long energy = 0;
 
+    int is_vanilla = (
+        lambda_offset_i == 0 &&
+        lambda_plane_i == 0 &&
+        dq_dl_i == 0 &&
+        dsig_dl_i == 0 &&
+        deps_dl_i == 0 &&
+        lambda_offset_j == 0 &&
```
badisa, line 910: You mentioned offline that when the 4D coords are the same (which I believe the lambda values influence) we get the singularity. Do we need to check equality of the lambda values rather than just checking both are zero?

proteneer: The 4D coordinates are still computed on the fly in master, once #403 is merged this check will change.

badisa (approved):

PR #412, opened 2021-05-04
Lambda improvements

proteneer:
This small PR does two things:

1) Enables one to use ternary operators by adding a default single arg constructor to the Surreal class
2) Modifies the kernel_cache_ to be bound to a nonbonded kernel, as opposed to having static scope. Previously, this would cause problems as importing the custom_op will occur *before* CUDA_VISIBLE_DEVICES is set, which will result in the driver using whatever the default is.


badisa (approved):
Good catch!

proteneer:
Thanks for the spot review!

PR #414, opened 2021-05-05
Make exceptions explicit to instance

badisa:
* grpc errors don't mention the ip address as it attempts to batch
  up the requests, which is bad for verification

### DIFF
--- a/parallel/client.py
+++ b/parallel/client.py
@@ -192,11 +192,16 @@ class GRPCClient(AbstractClient):
         """
         See abstract class for documentation.
         """
-        futures = [stub.Status.future(service_pb2.StatusRequest()) for stub in self.stubs]
-        workers_status = [x.result() for x in futures]
-        for field in ("nvidia_driver", "git_sha",):
-            # All fields should be the same
-            host_values = {self.hosts[i]: getattr(x, field) for i, x in enumerate(workers_status)}
-            uni_vals = set(host_values.values())
-            assert len(uni_vals) == 1, f"Not all hosts agreed for {field}: {host_values}"
-            assert all(uni_vals), f"Missing values for {field}: {host_values}"
+        prev_vals = {}
+        for host, stub in zip(self.hosts, self.stubs):
+            try:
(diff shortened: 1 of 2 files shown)


parallel/client.py
```diff
@@ -192,11 +192,16 @@ def verify(self):
         """
         See abstract class for documentation.
         """
-        futures = [stub.Status.future(service_pb2.StatusRequest()) for stub in self.stubs]
-        workers_status = [x.result() for x in futures]
```
badisa, line 196: Previously an exception could be raised here that said "not all workers could be connected to", which made it difficult to debug if an instance was acting up. This will have the ip address in the exception allowing for easier debugging of borked instances

parallel/client.py
```diff
@@ -192,11 +192,16 @@ def verify(self):
...
-        for field in ("nvidia_driver", "git_sha",):
-            # All fields should be the same
-            host_values = {self.hosts[i]: getattr(x, field) for i, x in enumerate(workers_status)}
-            uni_vals = set(host_values.values())
-            assert len(uni_vals) == 1, f"Not all hosts agreed for {field}: {host_values}"
-            assert all(uni_vals), f"Missing values for {field}: {host_values}"
+        prev_vals = {}
+        for host, stub in zip(self.hosts, self.stubs):
+            try:
+                status = stub.Status(service_pb2.StatusRequest())
```
badisa, line 198: This has a default timeout of 20 seconds, could reduce it to 5/10, though might cause us some headache later.

parallel/client.py
proteneer: Nice! this was intentionally designed this way so you can replace the list comprehension with something more useful later :)

parallel/client.py
proteneer: What are common reasons for timing out?

badisa: - No connection/No server
- Slow connection <- This one might bite us if we set it too low

proteneer (approved):

PR #415, opened 2021-05-05
Shrink Nonbonded Interactions

proteneer:
This PR implements a ligand shrinking technique initially described by Gilson et al used for docking. This only applies to *intermolecular* nonbonded interactions.

The idea is to create virtual sites on every ligand atom, whose positions are adjusted relative to the centroid. At lambda=0, the ligand is full size, and at lambda=1, every ligand atom is located the centroid. The intuition for this is to allow for the environment atoms (water, side chains, etc.) to slowly flood to avoid phase-like transitions. 

This method has two neat properties:

1) automatically neutralizing the ligand
2) the ligand still has fully functioning bonded terms, and intramolecular nonbonded terms.
3) one can possibly get away with only center of mass restraints using this techniques, since it reduces the ligand to be featureless blob that is independent of orientation. Though we can still use endpoint-correction like before.

I don't particularly like the implementation, as its quite sub-optimal, so marking this as a draft until we've ran some experiments with it.

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -163,7 +163,7 @@ def prepare_water_system(
 
     params = np.stack([
         (np.random.rand(N).astype(np.float64) - 0.5)*np.sqrt(138.935456), # q
-        np.random.rand(N).astype(np.float64)/5.0, # sig
+        np.random.rand(N).astype(np.float64)/6.0, # sig # fix me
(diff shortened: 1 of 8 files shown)


proteneer:
This doesn't feel right in the end, so we'll re-visit this later.

PR #418, opened 2021-05-06
Issue 416: Don't run bound potential if there are zero params

badisa:
* Unable to launch 0 blocks on the GPU and the values computed
  using the bound potentials are never used


proteneer (approved):
LGTM!

PR #419, opened 2021-05-06
Update MCS search with latest RDKit capabilities

badisa:
* Finds same number of hif2a edges
* Only finds a single edge for CDK8, which indicates more work
  is necessary.
* Minor clean up because OCD

I had hoped that this would generate a completely star graph for CDK8, though in my own hand curated generation of the current edges I was only able to get some triangles.


maxentile (approved):
Nice! Clear improvements all around.

PR #420, opened 2021-05-07
Variadic decoupling

proteneer:
This PR expands the flexilibity of the lambda scheduling to allow it vary on a per-particle basis. For example, one can introduce per-particle decoupling, so we can pull out (or even alchemically scale) particles at different rates. As an illustration, setting transform_w in the nonbonded constructor to:

```
args.append("""
    double offset = atom_idx / (2.0*N);
    if(lambda < offset) {
        return 0.0;
    } else if (lambda > (0.5 + offset)) {
        return 1.0;
    } else {
        NumericType term = sin((lambda - offset)*PI);
        return term*term;
    }

""")
```
will generate a decoupling schedule that allows the atoms in the ligand to be decoupled sequentially:
Figure_1

### DIFF
--- a/tests/test_parameter_interpolation.py
+++ b/tests/test_parameter_interpolation.py
@@ -121,8 +121,27 @@ class TestInterpolatedPotential(GradientTest):
         def transform_e(lamb):
             return jnp.where(lamb < 0.5, jnp.sin(lamb*np.pi)*jnp.sin(lamb*np.pi), 1)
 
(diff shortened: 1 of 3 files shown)


proteneer:
edit: do not merge until du_dp is fixed

badisa (approved):

proteneer:
Closing this for now to reduce CI load.

PR #424, opened 2021-05-10
nonbonded docs

maxentile:
Adds docstrings for `nonbonded_v3` and `convert_to_4d`.

Possible adjustments:

- [x] Remove `scales` argument?
- [ ] Add reaction-field reference?

### DIFF
--- a/timemachine/potentials/nonbonded.py
+++ b/timemachine/potentials/nonbonded.py
@@ -131,5 +131,5 @@ def nonbonded_v3(
 def validate_coulomb_cutoff(cutoff=1.0, beta=2.0, threshold=1e-2):
     """check whether f(r) = erfc(beta * r) <= threshold at r = cutoff
     following https://github.com/proteneer/timemachine/pull/424#discussion_r629678467"""
-    if erfc(beta * cutoff) <= threshold:
+    if erfc(beta * cutoff) > threshold:
         print(UserWarning(f"erfc(beta * cutoff) = {erfc(beta * cutoff)} > threshold = {threshold}"))


timemachine/potentials/nonbonded.py
```diff
@@ -25,6 +20,45 @@ def nonbonded_v3(
...
+            where w = cutoff * (lambda_plane_idxs + lambda_offset_idxs * lamb)
+    params : (N, 3) np.array
+        columns [charges, sigmas, epsilons], one row per particle
+    box : Optional 3x3 np.array
+    lamb : float
+    charge_rescale_mask : (N, N) np.array
+        the Coulomb contribution of pair (i,j) will be multiplied by charge_rescale_mask[i,j]
+    lj_rescale_mask : (N, N) np.array
+        the Lennard-Jones contribution of pair (i,j) will be multiplied by lj_rescale_mask[i,j]
+    scales
```
proteneer, line 42: I think this can be removed.

timemachine/potentials/nonbonded.py
```diff
@@ -25,6 +20,45 @@ def nonbonded_v3(
...
+        unused # TODO: remove?
+    beta : float
```
proteneer, line 44: in OpenMM I think this is called alpha (real part PME), I'd be happy to be consistent with OpenMM

timemachine/potentials/nonbonded.py
```diff
@@ -25,6 +20,45 @@ def nonbonded_v3(
...
+        the charge product q_ij will be multiplied by erfc(beta*d_ij)
+    cutoff : Optional float
```
proteneer, line 46: at some point we should add a sanity check that U at cutoff is less than some threshold

timemachine/potentials/nonbonded.py
```diff
@@ -25,6 +20,45 @@ def nonbonded_v3(
...
+
+    Returns
+    -------
+    energy : float
+
+    References
+    ----------
+    * Rodinger, 2005, J. Chem. Phys. "Absolute free energy calculations by thermodynamic integration in four spatial
+        dimensions" https://aip.scitation.org/doi/abs/10.1063/1.1946750
+    * TODO: Add a reference for the reaction field treatment?
```
proteneer, line 60: I need to dig this up again - it's really just the real part of PME right now.

timemachine/potentials/nonbonded.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/424/commits/eb64bf66c7ecd38516517f9e0ada9424bbffef5e (removing references) and https://github.com/proteneer/timemachine/pull/424/commits/fb5079847da403a7ab537d62a3c96cfb5bafd829 (modifying the signature)

PR #429, opened 2021-05-13
Jax wrap unbound impls

maxentile:
Adds a convenience function that abstracts a collection of `unbound_impl`s as a single function `U`, where Jax has been informed how to compute `grad(U, argnums=(0,1,3))(coords, sys_params, box, lam)`.

This is intended to reduce the amount of "manual" packing and unpacking of force field parameter derivatives. This should be helpful in increasing prototyping flexibility in general, and especially on the way towards compact implementation of a few specific planned features such as [differentiable importance weights](https://github.com/proteneer/timemachine/pull/404), which would benefit from having Jax know how to use timemachine's custom derivative rules for `U`. This should also make it more convenient for us to define inexpensive experiments on aspects of optimizer logic that depend on talking to a force field but that don't depend on talking to a full-blown free energy calculation -- a direction that @badisa and I have discussed.

The approach taken in this PR doesn't yet play nice with other Jax transformations such as `vmap`. (For example, `vmap(U, in_axes=(None, None, None, 0))(coords, sys_params, box, np.linspace(0,1,60))` doesn't fall back to a simple Python loop, but instead raises an error.) To extend in this direction we would probably go the Jax primitive route instead of the Jax custom jvp route. Note also that the current PR doesn't touch the `ff_params -> sys_params` step.


badisa (approved):

proteneer:
These changes look good to me, and thanks for adding the extra tests to check for linearizability of the jvp.

proteneer (approved):

PR #431, opened 2021-05-17
Fixes #428: Modify include paths dynamically

### LINKED ISSUE #428: Runtime compilation errors from jitify if run from directories other than `timemachine/`
To reproduce: place [`quick_test.py`](https://gist.github.com/maxentile/e5ac3e7aa7dbdc814597c8bcb73f1a57) in `timemachine/examples/`. Running `python quick_test.py` from `timemachine/examples/` will produce some jitify errors about not being able to find `surreal.cuh` and `k_fixed_point.cuh`, but running `python examples/quick_test.py` from `timemachine/` will run without error.

Troubleshooting steps suggested by @proteneer and @badisa : 
- [x] Confirm that [the path to](https://github.com/proteneer/timemachine/blob/803d417b9dd7defe5dc91b958ea5963d2b0751d6/timemachine/cpp/src/wrap_kernels.cpp#L908-L909) `k_lambda_transformer_jit.cuh` is absolute rather than relative
- [x] Specify absolute paths to these headers using `--pre-include`
- [ ] Embed source files directly using [this macro](https://github.com/NVIDIA/jitify/blob/2f65aa60c5ebabe28d254feb092e727a5a754a25/jitify.hpp#L45)

badisa:
A bit of a hack, but its consistent with how we are injecting the dynamic code otherwise.

### DIFF
--- a/tests/test_jit.py
+++ b/tests/test_jit.py
@@ -1,0 +1,27 @@
+import os
+from tempfile import TemporaryDirectory
+from unittest import TestCase
+
+from md.builders import build_water_system
+from md.minimizer import minimize_host_4d
+
+from fe.free_energy import RelativeFreeEnergy
+
+from testsystems.relative import hif2a_ligand_pair
+
+
+class TestJIT(TestCase):
+
+    def test_random_directory(self):
+        with TemporaryDirectory(prefix="timemachine") as temp_dir:
+            orig_dir = os.getcwd()
+            os.chdir(temp_dir)
+            try:
+                # build a pair of alchemical ligands in a water box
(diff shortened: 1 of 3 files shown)


maxentile (approved):

tests/test_jit.py
```diff
@@ -0,0 +1,27 @@
...
+from md.minimizer import minimize_host_4d
+
+from fe.free_energy import RelativeFreeEnergy
+
+from testsystems.relative import hif2a_ligand_pair
+
+
+class TestJIT(TestCase):
+
+    def test_random_directory(self):
```
badisa, line 15: This test fails when the .so is compiled without these changes

Issue #373, opened 2021-03-17
NPT support?


maxentile:
@proteneer : Could you please comment on any design constraints that affect how `timemachine` could support NPT simulation?

I think earlier there were more stringent design constraints that would have ruled out methods like the `MonteCarloBarostat` used in OpenMM. Currently, however, I'm not aware of design constraints that would rule out `MonteCarloBarostat`.

There may be different sets of requirements for two tasks:
* "NPT equilibration" - get a single box volume suitable for NVT production simulation
* "NPT production" - sample box fluctuations during production simulation

maxentile:
Following up on [a comment from the other thread](https://github.com/proteneer/timemachine/pull/372#issuecomment-801406388):
>Minimizing the energy is not the proper objective. You can imagine a single particle with a +1 charge would optimize the box into an infinitely large box in order to minimize U. You need to optimize the box into an appropriate pressure, so what you'd want really is still to compute current virial and compare it against a target virial. 

If we minimize the NPT reduced potential `u(coords, box) + p V(box)`, rather than the potential energy `U(coords, box)`, that would rule out infinitely large-volume boxes from appearing as minimizers in this example. Would the NPT reduced potential be a suitable objective to minimize for (possibly deterministic) NPT equilibration?

maxentile:
Summarizing offline discussions:
* The priority is to solve the "NPT equilibration" task, before the "NPT production" task. A shared solution to both tasks would be preferable, if possible.
* A solution to the "NPT equilibration" task must address a possible deficiency of the Monte Carlo barostat, namely slow equilibration times.
* If needed, a solution can introduce a dependence on the virial. 

I'd imagine the default solution is to use OpenMM MC barostat. I've plotted the equilibration behavior of the OpenMM MC barostat on a TIP3P waterbox, initialized with a padded box, as we vary the proposal frequency: https://gist.github.com/maxentile/930d0686f9927d683f650c9c67cdb575

![image](https://user-images.githubusercontent.com/5759036/111790394-86a07000-8898-11eb-8a7d-e35f4fde6ccb.png)

Note that increasing the barostat proposal frequency increases the wall time per step, up to a factor of ~3x if we make 1 proposal per MD step. I think this suggests the possible "slow equilibration time" deficiency of the MC barostat can be addressed by adjusting this setting.

Switching the x axis from # MD steps to wall time, and focusing on the fastest barostat proposal frequencies:

[...]

PR #433, opened 2021-05-19
NPT support

maxentile:
Closes https://github.com/proteneer/timemachine/issues/373 by cloning MonteCarloBarostat implementation from OpenMM.

- [x] Add [`NVTEnsemble`](https://github.com/proteneer/timemachine/blob/3e20a5a18d4b8aaeba144001fd044a08b03417cb/md/ensembles.py#L52-L78) and [`NPTEnsemble` ](https://github.com/proteneer/timemachine/blob/3e20a5a18d4b8aaeba144001fd044a08b03417cb/md/ensembles.py#L81-L121) definitions.
- [x] Add [`MonteCarloMove`](https://github.com/proteneer/timemachine/blob/3e20a5a18d4b8aaeba144001fd044a08b03417cb/md/moves.py#L6-L31)s for temperature control ([`UnadjustedLangevinMove`](https://github.com/proteneer/timemachine/blob/3e20a5a18d4b8aaeba144001fd044a08b03417cb/md/thermostat/moves.py#L7-L34)) and pressure control ([`MonteCarloBarostat`](https://github.com/proteneer/timemachine/blob/3e20a5a18d4b8aaeba144001fd044a08b03417cb/md/barostat/moves.py#L86-L196)). NPT simulation is performed by [composing](https://github.com/proteneer/timemachine/blob/3e20a5a18d4b8aaeba144001fd044a08b03417cb/md/utils.py#L25) these moves. Later extensions of timemachine might update the current UnadjustedLangevinMove to use the LangevinMiddleIntegrator scheme from OpenMM, or add wrappers for external implementations of more advanced HMC moves.
- [x] Add [usage example](https://github.com/proteneer/timemachine/blob/3e20a5a18d4b8aaeba144001fd044a08b03417cb/examples/npt_equilibration.py) applying NPT equilibration to each window of an absolute free energy calculation.
- [x] Add [tests](https://github.com/proteneer/timemachine/blob/3e20a5a18d4b8aaeba144001fd044a08b03417cb/tests/test_ensembles.py) of the reduced potential computation in NVT, NPT ensembles
- [x] Add [test](https://github.com/proteneer/timemachine/blob/3e20a5a18d4b8aaeba144001fd044a08b03417cb/tests/test_barostat.py#L56-L155) that barostat'd molecular ideal gas volume is consistent with expectation
- [x] Add [slow test](https://github.com/proteneer/timemachine/blob/3e20a5a18d4b8aaeba144001fd044a08b03417cb/slow_tests/test_barostat_slow.py) that barostat'd density of water is very close to 1.0 kg/L
- [x] Add [convenience function](https://github.com/proteneer/timemachine/blob/3e20a5a18d4b8aaeba144001fd044a08b03417cb/md/thermostat/utils.py#L6-L19) for sampling according to temperature, and delete vestigial [`set_velocities_to_temperature`](https://github.com/proteneer/timemachine/commit/1f976288230b5c429cf0568387d855f873bf6c70#diff-c8d76629b43f57a44bd424b60d79e55e4bf555c7031210b0f15bf35783071bd7L8-L13)
- [x] Delete vestigial [barostat/barostat.py](https://github.com/proteneer/timemachine/blob/0ff865c8c84d43c8a76e2a1ff0e3c178dc5fb7ad/barostat/barostat.py), [barostat/potentials.py](https://github.com/proteneer/timemachine/blob/0ff865c8c84d43c8a76e2a1ff0e3c178dc5fb7ad/barostat/potentials.py)


slow_tests/test_barostat_slow.py
```diff
@@ -0,0 +1,111 @@
...
+from md.states import CoordsVelBox
+from md.utils import simulate_npt_traj
+from md.thermostat.moves import UnadjustedLangevinMove
+from md.thermostat.utils import sample_velocities
+
+from timemachine.lib import LangevinIntegrator
+
+from functools import partial
+
+# simulation parameters
```
badisa, line 27: Wrapping this in a unittest.TestCase seems cleaner than having this as a script

badisa (approved):

slow_tests/test_barostat_slow.py
badisa: Actually trying to run this, it took ~5 minute so perhaps best not as a test? Might be worth deciding on some convention between a 'test' and a 'benchmark'. As the slow tests still get run in CI

maxentile: I had mistakenly thought CI was running  `slow_tests/` less frequently than `tests/` (but I see that the actual behavior of `slow_tests/` is documented in [this PR description](https://github.com/proteneer/timemachine/pull/405) -- tests in this folder are run without cuda-memcheck so that they can be run at the same frequency as `tests/`, just faster). Happy to move this wherever is appropriate.

Do you think we should move this from `slow_tests/` into a new `benchmarks/` folder or something that doesn't get discovered by CI at all?

Another alternative might be to [mark](https://docs.pytest.org/en/6.2.x/mark.html) this and possibly other tests as slow, telling pytest to skip these most of the time, but periodically still run the slow tests through pytest.

badisa: I say we leave this for now, as it is, this won't cause any issues. Probably warrants more discussion than a PR would allow

examples/npt_equilibration.py
```diff
@@ -0,0 +1,128 @@
...
+initial_waterbox_width = 3.0 * unit.nanometer
+timestep = 1.5 * unit.femtosecond
+collision_rate = 1.0 / unit.picosecond
+n_moves = 2000
+barostat_interval = 5
+seed = 2021
+
+# thermodynamic parameters
+temperature = 300 * unit.kelvin
+pressure = 1.0 * unit.atmosphere
```
proteneer, line 41: Should we be consistent with OpenMM for sanity purposes? IIRC the default was 1 bar (which is very close but not exactly to 1 atm)

md/barostat/moves.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def displace_by_group(self, coords, displacements):
+        return coords + displacements[self.scatter_inds]
+
+    def scale_centroids(self, coords, center, scale):
+        """
+
+        Notes
+        -----
+        * Currently ignores particle mass -- the centroid of a group of
```
proteneer, line 74: Does the OpenMM centroid computation take mass into account? @badisa

proteneer: (PS if we decide to support this, `np.average` as opposed to `np.mean` allows one to pass in a list of weights.

md/barostat/moves.py
```diff
@@ -0,0 +1,196 @@
...
+        Notes
+        -----
+        TODO: refactor so that this doesn't overwrite the global barostat.n_proposed, barostat.n_accepted counters,
+            instead using something like barostat.n_proposed_since_last_adaptation?
+
+        References
+        ----------
+        Direct clone of https://github.com/openmm/openmm/blob/d8ef57fed6554ec95684e53768188e1f666405c9/openmmapi/src/MonteCarloBarostatImpl.cpp#L103-L113
+        with the following tweaks:
+            * introduces a lower_bound (so that the proposal scale can't go below some preset value)
```
proteneer, line 160: I like these tweaks!

md/barostat/moves.py
badisa: Based on my reading of the OpenMM, it does not take the masses into account. That said, the kernel logic that they have is  fairly opaque to me. 

https://github.com/openmm/openmm/blob/be19e0222ddf66f612016a3c1f687161a53c2396/platforms/common/src/CommonKernels.cpp#L7625

md/barostat/utils.py
```diff
@@ -0,0 +1,60 @@
+import numpy as np
+import networkx as nx
+from scipy.spatial.distance import pdist
+from typing import List
+from timemachine.lib.potentials import HarmonicBond
+
+
+def compute_box_volume(box: np.ndarray) -> float:
+    assert box.shape == (3, 3)
+    return np.linalg.det(box)
```
proteneer, line 10: Nice! I hadn't realized `det(box)` was the proper way to generalize volumes to triclinic systems, but after some searching it appears you are correct.

https://journals.iucr.org/a/issues/2011/01/00/au5114/au5114.pdf

md/barostat/utils.py
```diff
@@ -0,0 +1,60 @@
...
+    # NOTE: this assumes all bonds are represented by harmonic bond force
+    bond_list = list(map(tuple, harmonic_bond_potential.get_idxs()))
+    # TODO: if we add HBond constraints, be sure to add these to the bond_list!
+    alchemical_topology = nx.Graph(bond_list)
+    connected_components = list(map(list, nx.algorithms.connected_components(alchemical_topology)))
+
+    return connected_components
+
+
+def merge_big_groups(group_indices: List[np.array]) -> List[np.array]:
```
proteneer, line 36: is this still being used?

md/barostat/utils.py
```diff
@@ -0,0 +1,60 @@
...
+    assert np.linalg.norm(box - np.diag(np.diag(box))) == 0
+
+    center = np.sum(box / 2, axis=0)
+
+    assert center.shape == (3,)
+
+    return center
+
+
+def get_group_indices(harmonic_bond_potential: HarmonicBond) -> List[np.array]:
```
proteneer, line 25: Note that in certain cases (eg. core restraints for ABFE), we will actually have two harmonic bond terms, one for the regular valences, and one for the restraining term. It's probably more flexible to just modify the input of this function to either

1) a list of HarmonicBonds
or perhaps slightly cleaner:
2) have it take in a simple Bx2 list of idxs (and leave it up to the caller to pass merge the indices).

md/thermostat/utils.py
```diff
@@ -0,0 +1,19 @@
...
+
+def sample_velocities(masses: Quantity, temperature: Quantity) -> np.array:
+    """Sample Maxwell-Boltzmann velocities ~ N(0, sqrt(kB T / m)"""
+    n_particles = len(masses)
+    spatial_dim = 3
+
+    v_unscaled = np.random.randn(n_particles, spatial_dim)
+
+    # intended to be consistent with timemachine.integrator:langevin_coefficients
+    sigma = np.sqrt(BOLTZ * temperature.value_in_unit(kelvin)) * np.sqrt(1 / masses)
```
proteneer, line 14: I double checked this against https://github.com/choderalab/openmm-validation/blob/master/tests/test_custom_integrators.py#L62

and it seems consistent! nice.

proteneer:
One last thing I would add (and probably something you two are already thinking of) - can you work with @badisa on figuring what unit tests for his C++ code should be for correctness purposes?

proteneer (approved):
LGTM - some clerical comments but otherwise looks fantastic. This also sets up a good framework in case we want to introduce other types of MC moves in the future! (eg. random rigid rotation move, etc.)

examples/npt_equilibration.py
maxentile: Updated the examples from atm to bar in https://github.com/proteneer/timemachine/pull/433/commits/0f483831b931ee8ed3844391a9aa5add835290ca. (Note that we don't set a default temperature or pressure in the `barostat` or `ensembles` code itself, and allow user inputs in any simtk pressure unit.)

md/barostat/utils.py
maxentile: Ahh, good catch -- this had been used in an earlier version where I wanted to have the barostat displace the protein:ligand complex as a single unit (rather than displacing the protein and ligand separately). Examples don't currently exercise this option, and I don't currently see a compelling reason to do it one way or the other, so I removed this function in https://github.com/proteneer/timemachine/pull/433/commits/03952ce62b88fecf1a7c43408413db7ea9796175 .

md/barostat/utils.py
maxentile: Updated the signature as suggested here https://github.com/proteneer/timemachine/pull/433/commits/954937ce6ba2a2b4c44a2c6d36c5efc59c9318e7 . Later, might be worthwhile to remove both of these and replace with an "alchemical topology" object or similar, that helps keep track of what bonds are present regardless of whether they're represented as harmonic bond terms, distance constraints, etc. down the road.

tests/test_barostat.py
```diff
@@ -0,0 +1,155 @@
...
+            u, du_dx = ensemble.reduced_potential_and_gradient(x, box, lam)
+            return u
+
+        thermostat = UnadjustedLangevinMove(integrator_impl, potential_energy_model.all_impls, lam, n_steps=barostat_interval)
+        barostat = MonteCarloBarostat(partial(reduced_potential_fxn, lam=lam), group_indices, max_delta_volume=3.0)
+
+        v_0 = sample_velocities(masses * unit.amu, temperature)
+
+        # rescale the box to be approximately the desired box volume already
+        rescaler = CentroidRescaler(group_indices)
```
badisa, line 138: If you disable the barostat in this test, this test still passes

maxentile: Explainable (but undesirable!) behavior, since the test is initialized at the expected volume here https://github.com/proteneer/timemachine/pull/433/commits/5f9cf6dccd2021060ec6033c260cc346dc0adc71 .

The initial version of the test took a long time to run, since it required to equilibrate to a far larger box size than initialized (like, initialized at ~10-100 nm^3, with ~10000 nm^3 expected). However, changing the initialization to be _exactly_ the expected box volume has the undesirable effect of making the test only check whether running the barostat causes the simulation to _drift away from the correct answer._ We of course also would want this test to fail if the barostat is doing nothing!

The behavior of the test should be fixed in https://github.com/proteneer/timemachine/pull/433/commits/05817c851e676ccb40042ca2b121adbfa367213b , where the barostat is initialized from a box that's instead 10% too large.

maxentile: And with this aspect of the test behavior fixed, it appears the test is no longer passing!

This needs to be resolved before merging.

Possibilities:
* there are uncaught bugs with the barostat itself
* the test is not behaving as intended for another reason, e.g. because simulation with such a giant box is causing other numerical issues or slow thermostat convergence. (Note that the OpenMM version of this test system has only 64 particles, and much smaller expected box sizes...)

maxentile: [Increasing the pressure by a factor of 100](https://github.com/proteneer/timemachine/pull/433/commits/4886d6ec176d26d70a8bdc2ee70aa9e951e89781) so that the expected volume is within a couple orders of magnitude of water box simulation cells we typically consider, the test is failing but with small relative differences from expectation.

```
Mismatched elements: 1 / 3 (33.3%)
Max absolute difference: 4.03597219
Max relative difference: 0.01881098
 x: array([107.001069, 218.590044, 360.868921])
 y: array([107.277036, 214.554072, 357.59012 ])
```

Will adjust to run this test longer...

maxentile: Adjusting the molecular ideal gas tests to [run 5x longer](https://github.com/proteneer/timemachine/pull/433/commits/ba0e6676d8198b23244425e0c52a73963e207938) causes this to pass, reducing my level of suspicion of an uncaught bug in the barostat itself. However, this adds another 3 minutes to the tests...

jchodera: Could limiting `max_delta_volume=3.0` here be a source of slow convergence? An optimal choice (or adaptively identified choice) would presumably converge pretty quickly if you're just testing the mean.

You can also cut your convergence time in half by not discarding the first half to equilibration, but instead starting at the analytical volume and seeing if it drifts away?

badisa (approved):
Looks good. Only thought would be to move the barostat test to the slow tests to avoid running under cuda memcheck as you had mentioned previously.

tests/test_barostat.py
maxentile: Thank you! These comments are partially addressed in https://github.com/proteneer/timemachine/pull/433/commits/64520756b9f155edef466049d3f1f75e087895bd by removing the `max_delta_volume` argument, and initializing in a way that should require less equilibration time.

A previous version of this test did initialize at the analytical volume and wait to see if it drifts away, but this led to false passes. Here I've opted to initialize just slightly outside of the relative tolerance, so that the required convergence time is reduced, but the test will still fail if the barostat does nothing.

maxentile:
>move the barostat test to the slow tests to avoid running under cuda memcheck as you had mentioned previously.

Thank you! Addressed in https://github.com/proteneer/timemachine/pull/433/commits/fef7fe1a01c913cd6fb13952b7eee093c2219d1e

badisa:
Running the molecular ideal gas test on my dev box I see a failure:

```
E       AssertionError: 
E       Not equal to tolerance rtol=0.01, atol=0
E       
E       Mismatched elements: 1 / 3 (33.3%)
E       Max absolute difference: 2.78290284
E       Max relative difference: 0.0107958
E        x: array([107.813074, 216.870393, 360.373084])
E        y: array([107.277054, 214.554109, 357.590181])
```

maxentile:
>Running the molecular ideal gas test on my dev box I see a failure:

Ahh, off by 1.07% where the tolerance was 1%... Reverting from `n_moves=2000` to `n_moves=10000` in https://github.com/proteneer/timemachine/pull/433/commits/c234ff4b60cac00a86a1415205714779e4ef13d3 fixes this for now, and later we can pick the relative tolerance in a bit more of a principled way based on the expected variability.

PR #436, opened 2021-05-25
Add C++ MonteCarloBarostat

badisa:
* Aproximately 30x faster than python version
* Still need to get this on the GPU for an even bigger speed up


badisa:
This hasn't been ported to CUDA yet, and wanted to get feedback on the general C++ side of things along with implementation.

maxentile (commented):
Nice! Do you have an initial sense of what's most responsible for the speed-up in the current version, and what fraction of time is now spent in the barostat compared with the thermostat if the barostat_interval is, say, 5 or 10?

(In the Python version, there was a lot of overhead in [looping over the `bound_impls`](https://github.com/proteneer/timemachine/blob/5f645aea8c9271b98e9313ce7f2f026e037c055e/md/ensembles.py#L36-L49) to compute the energies required for the accept-reject step, for example, so that the cost of a single barostat step was ~5x the cost of a regular MD step.)

slow_tests/test_barostat_molecular_ideal_gas.py
```diff
@@ -79,6 +79,16 @@ def test_molecular_ideal_gas():
...
     n_molecules = complex_top.getNumResidues()
 
+    bound_potentials = []
+    for params, unbound_pot in zip(sys_params, unbound_potentials):
+        bp = unbound_pot.bind(np.asarray(params))
+        bound_potentials.append(bp)
+
+    u_impls = []
+    for bp in bound_potentials:
+        bp_impl = bp.bound_impl(precision=np.float32)
```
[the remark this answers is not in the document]
badisa: I will give this a shot with float64 and see if it differs.

I don't have the NaN catching logic that you do, and have seen very reasonable (1k-2k) energies in this test. Will verify that energies are blowing up in float64, though would guess that float64 is less likely to blow up than float32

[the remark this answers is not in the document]
proteneer: IIRC,

FWIW even in "float64" its only about roughly the precision of a float42 or something due to the fixed point conversion. Things that blow-up in float64 are probably as likely to blow up in float32.

timemachine/cpp/src/barostat.cu
```diff
@@ -0,0 +1,224 @@
...
+    gpuErrchk(cudaMemcpy(&h_box[0], d_box, 9*sizeof(d_box), cudaMemcpyDeviceToHost));
+
+    double volume = h_box[0*3+0]*h_box[1*3+1]*h_box[2*3+2];
+    double center_x = h_box[0*3+0] * 0.5;
+    double center_y = h_box[1*3+1] * 0.5;
+    double center_z = h_box[2*3+2] * 0.5;
+
+    if(volume_scale_ == 0) {
+        volume_scale_ = 0.01*volume;
+    }
```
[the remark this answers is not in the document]
badisa: Should likely move this initialization to the constructor.

Are you saying that the barostat should accept providing JIT functions to adapt or have a python API to the barostat to modify the volume_scale?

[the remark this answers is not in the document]
proteneer: I think I'd written this initially as a way to not have to explicitly pass the volume in the constructor (not something that's always known), so this is in practice only ran on the very first call to inplace_move.

timemachine/cpp/src/barostat.cu
```diff
@@ -0,0 +1,224 @@
...
+    // Scale centroids
+    //const int tpb = 32;
+    //const int blocks = (num_molecules + tpb - 1) / tpb;
+    //scale_centroids<<<blocks, tpb, 0, stream>>>(num_molecules, d_group_idxs_, d_x_after_, d_box_after_);
+    //gpuErrchk(cudaPeekAtLastError());
+    for( std::vector<int> mol_atoms : group_idxs_) {
+        double centroid_x = 0;
+        double centroid_y = 0;
+        double centroid_z = 0;
+        for( int idx : mol_atoms ) {
```
proteneer, line 135: your inner python has taken over

timemachine/cpp/src/barostat.cu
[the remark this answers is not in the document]
maxentile: >Are you saying that the barostat should accept providing JIT functions to adapt or have a python API to the barostat to modify the volume_scale?

No, sorry for confusion -- meant to suggest moving the "if volume_scale_ is too small, set to 0.01" comparison (which currently appears at L92 but is not checked later) into the scale adapting logic between L210 and L219 (which currently compares to a maximum volume scale threshold but not a minimum volume_scale_ threshold).

timemachine/cpp/src/context.cu
```diff
@@ -46,6 +48,16 @@ Context::Context(
...
 };
 
+Context::Context(
+    int N,
+    const double *x_0,
+    const double *v_0,
+    const double *box_0,
+    Integrator* intg,
+    std::vector<BoundPotential *> bps) :
+    Context(N, x_0, v_0, box_0, intg, nullptr, bps) {};
```
proteneer, line 58: Did you intend to disable the definition? Invoking this constructor would instantiate an invalid object since the internals aren't initialized.

proteneer:
Some additional things to consider and test for:

1) Create barostat at 10 atm
2) Run MD for say 1000 steps
3) Modify barostat pressure to 1 atm
4) Re-run for 10000 steps

Would you need to re-initialize the internals? (eg. volume_scale_, etc.) to accomodate for the change in the barostat? Or would a brand new context need to be created?

timemachine/cpp/src/context.cu
badisa: I did not. Was trying to use a delegated constructor, but sounds like I borked it

badisa:
The GPU implementation as I have done it seems to have a negligible impact on performance on the slow barostat test. Need to do some benchmarking on systems of varying size to see if the performance impact grows larger with larger number of molecules.

slow_tests/test_barostat_molecular_ideal_gas.py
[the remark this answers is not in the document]
badisa: No discernible difference between float32/64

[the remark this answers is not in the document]
maxentile: Nice! Will mark this as resolved then -- I had mistakenly thought the threshold for when large overflown energies would appear as large negative energies during MC trials was substantially different in the two precision modes.

timemachine/cpp/src/barostat.cu
```diff
@@ -0,0 +1,267 @@
...
+
+    gpuErrchk(cudaMalloc(&d_u_before_, N_*sizeof(*d_u_before_)));
+    gpuErrchk(cudaMalloc(&d_u_after_, N_*sizeof(*d_u_after_)));
+
+    gpuErrchk(cudaMalloc(&d_mol_idxs_, num_mols*sizeof(*d_mol_idxs_)));
+
+    int mol_idxs[num_mols];
+    for (int i = 0; i < num_mols; i++) {
+        std::vector<int> atoms = group_idxs[i];
+        // Sort the array in ascending order
```
proteneer, line 57: Why does each mol need have its atoms be sorted? To help with faster mem loading?

badisa: Because the group indices are not always ordered so you could get something like...

```
[[1,2,3], [6,5,4]]
```
which would result in the logic thinking that the atom is of size 5 rather than 3. Could also find the min of the next mol, which would be more consistent in its run time.

timemachine/cpp/src/barostat.cu
```diff
@@ -0,0 +1,267 @@
...
+        std::sort(atoms.begin(), atoms.end());
+        mol_idxs[i] = atoms[0];
```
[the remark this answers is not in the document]
badisa: Correct, should I add a check for this case, or is there a use case for running the barostat with nothing in it?

timemachine/cpp/src/barostat.cu
```diff
@@ -0,0 +1,267 @@
...
+    double center_z = box[2*3+2] * 0.5;
+
+    const int starting_atom_idx = mol_idxs[mol_idx];
+    const int ending_atom_idx = mol_idx + 1 >= num_molecules ? N : mol_idxs[mol_idx + 1];
+    const int num_atoms = ending_atom_idx - starting_atom_idx;
+
+    double centroid_x = 0;
+    double centroid_y = 0;
+    double centroid_z = 0;
+    for (int i = starting_atom_idx; i < ending_atom_idx; i++) {
```
proteneer, line 98: This is actually an inefficient algorithm esp. for the thread that processes the protein group. It will block every other thread. Did you benchmark the perf on a protein system?

timemachine/cpp/src/barostat.hpp
```diff
@@ -0,0 +1,88 @@
...
+    const int N_;
+    double pressure_;
+    const double temperature_;
+    int interval_;
+    const int seed_;
+    const std::vector<std::vector<int> > group_idxs_;
+
+    // stuff that deals with RNG
+    std::mt19937 mt_; // Expected to be seeded with seed_
+    std::uniform_real_distribution<double> dist_; // guaranteed unbiased
```
maxentile, line 67: Question: should this use the same RNG approach as the thermostat? https://github.com/proteneer/timemachine/blob/6658dde849574fe4b5b6eb3d84e7b9422bd20deb/timemachine/cpp/src/integrator.cu#L29-L31

The barostat consumes far less randomness than the thermostat so I think this is fine either way.

tests/test_barostat.py
```diff
@@ -0,0 +1,224 @@
...
+    # Set the pressure to 1 bar
+    baro.set_pressure((1 * unit.atmosphere).value_in_unit(unit.bar))
+    # Changing the barostat interval resets the barostat step.
+    baro.set_interval(2)
+    start = time.time()
+    while time.time() - start < 15:
+        ctxt.multiple_steps(np.ones(1000)*lam)
+        atm_box = ctxt.get_box()
+        if np.sum(atm_box) > ten_atm_box_sum:
+            break
```
maxentile, line 105: Can we remove this check? (With it present, I think the assertion `ten_atm_box_sum < np.sum(atm_box)` can pass also in the case of the box size randomly fluctuating, not just in the case of the box size consistently increasing)

timemachine/cpp/src/barostat.hpp
badisa: Since as it stands the random variables are used in C++ space, the C++ RNG seemed better suited. It has about 2 microsecond over head for the two random numbers generated per move, which is only a small fraction of the total time.

timemachine/cpp/src/barostat.cu
```diff
@@ -0,0 +1,295 @@
...
+    gpuErrchk(cudaMalloc(&d_u_before_, N_*sizeof(*d_u_before_)));
+    gpuErrchk(cudaMalloc(&d_u_after_, N_*sizeof(*d_u_after_)));
+
+    gpuErrchk(cudaMalloc(&d_centroids_, num_mols*3*sizeof(*d_centroids_)));
+    gpuErrchk(cudaMalloc(&d_mol_idxs_, N_*sizeof(*d_mol_idxs_)));
+    gpuErrchk(cudaMalloc(&d_mol_offsets_, (num_mols+1)*sizeof(*d_mol_offsets_)));
+
+    int offset = 0;
+    int mol_offsets[num_mols+1];
+    int mol_idxs[N_];
```
proteneer, line 60: You probably don't want to use N_, but to recompute this based on the sum of group.size() in group_idxs...

tests/test_barostat.py
badisa: Removed in 2601796

timemachine/cpp/src/barostat.cu
badisa: If N_ != sum([g.size() for g in group_idxs]) this will bork when indexing in the kernels, should I add an assertion that the two values are the same?

timemachine/cpp/src/barostat.cu
```diff
@@ -0,0 +1,295 @@
...
+    coords[atom_idx*3+2] += displacement_z;
+    if (atom_idx == 0) {
+        scaled_box[0*3+0] *= length_scale;
+        scaled_box[1*3+1] *= length_scale;
+        scaled_box[2*3+2] *= length_scale;
+    }
+}
+
+void __global__ find_group_centroids(
+    const int N, // Number of coords
```
proteneer, line 126: Are you sure this should be number of coords as opposed to sum of group_idxs size?

maxentile (approved):

timemachine/cpp/src/barostat.cu
badisa: With latest changes it is still N, but for mol idxs that are invalid the value is set to -1

timemachine/cpp/src/barostat.cu
badisa: After changes benchmarked per on protein vs water and it was the same. Great catch, should have noticed this

proteneer (approved):

proteneer:
Yazhu!

timemachine/cpp/src/barostat.cu
```diff
@@ -0,0 +1,328 @@
...
+    ) {
+
+    step_++;
+    if (step_ % interval_ != 0) {
+        return;
+    }
+    std::vector<double> h_box(9);
+    gpuErrchk(cudaMemcpy(&h_box[0], d_box, 9*sizeof(d_box), cudaMemcpyDeviceToHost));
+
+    double volume = h_box[0*3+0]*h_box[1*3+1]*h_box[2*3+2];
```
jchodera, line 191: Not sure if you folks will only ever support rectilinear boxes, but this volume expression will only work for rectilinear boxes (where all box angles are`pi/2`). You can simply replace this with `abs(det(h_box))` if you want it to be fully general for paralellepipeds.

PR #438, opened 2021-05-25
Stein control variates

maxentile:
The variance of a T.I. estimate is determined by the variance of the individual estimates for the averages `E_{x ~ pi(x; lam)} [ f(x) ]` where `f(x) = du_dlam(x; lam)`. 

One way to reduce the variance in these estimates is to introduce control variates, which are functions `g` whose mean in the distribution `pi` is 0: `E_{x ~ pi(x; lam)} [ g(x) ] = 0`.

If we have a zero-mean function `g`, then replacing the uncontrolled estimator `mean([f(x) for x in traj])` with the controlled estimator `mean([f(x) - g(x) for x in traj])` will be valid. If `stddev(f - g)` is smaller than `stddev(f)`, then it will also be beneficial. 

One recipe for constructing a parameterized family of zero-mean functions `g` is given by the "scalar Langevin-Stein operator." Given a flexible twice-differentiable `trial_fxn(x; params)` (e.g. defined by a neural network), we can turn it into an equally flexible family of valid control variates `g(x; params)` using [this one-liner](https://github.com/proteneer/timemachine/blob/c127d8cf34dac2f1403e6ae8f0764a91ff38d8eb/control_variates/stein.py#L42) in terms of derivatives of `trial_fxn` and `log_pi`. This is a specific instance of a versatile collection of approaches based on Stein's method -- see [references](https://github.com/proteneer/timemachine/blob/c127d8cf34dac2f1403e6ae8f0764a91ff38d8eb/control_variates/stein.py#L29-L36).

Given a flexible family of valid control variates, we can apply autodiff to minimize `stddev([f(x) - g(x; params) for x in traj])` and obtain a good control variate, as in [Si, Oates, Duncan, Carin, Briol, 2020] Scalable Control Variates for Monte Carlo Methods via Stochastic Optimization https://arxiv.org/abs/2006.07487 .

Note that this postprocessing approach can achieve variance reduction alone, and also in combination with reweighting. Where we would compute a reweighting-based estimate of the expectation of `f`, we can swap in a reweighting-based estimate of the expectation of `f - g`. (Multistate reweighting provides a different, complimentary kind of variance reduction when estimating expectations in a given lambda window, by optimally "borrowing" weighted samples from other lambda windows.)

Later extensions might:
* Construct more flexible trial functions than the restricted trial functions implemented here (which are [sums of pair functions](https://github.com/proteneer/timemachine/blob/c127d8cf34dac2f1403e6ae8f0764a91ff38d8eb/control_variates/trial_fxns/pair.py#L14-L30)). If the class of trial functions is sufficiently flexible, then reducing the variance all the way to 0 is possible in principle. For example, a direction I've discussed with @dominicrufa is to construct trial functions using flexible functional forms that have been successfully applied in the construction of ML potentials.
* Precompute a bank of good control variates, and then optimize linear combinations of these. Rather than optimizing _all_ the parameters of the control variates, we can form a fixed bank of control variates, and then in a much cheaper step find their optimal linear combination. A similar approach is taken in [Coles, Mangaud, Frenkel, Rotenberg, 2021] Reduced variance analysis of molecular dynamics simulations by linear combination of estimators https://arxiv.org/abs/2104.05038 .
* Consider importing or taking inspiration from more advanced implementations of control variates based on Stein's method, such as the R package ZVCV [South, 2020] Zero-variance control variates https://CRAN.R-project.org/package=ZVCV
* Apply control variates to estimation of other averages than `du_dlam`. For example, we compute the "thermodynamic gradient" in terms of averages of `du_params` in the lambda=0 and lambda=1 endstates. Estimation of these averages should also be amenable to variance reduction.


proteneer (approved):
I think we need to re-visit some of the design choices (laplacian operator, tanh bases) later on if we decide to use this in production. I'm happy to merge this as-is right now though.

maxentile:
Thanks!

>I think we need to re-visit some of the design choices (laplacian operator, tanh bases) later on if we decide to use this in production. 

Tanh basis functions were an arbitrary conservative placeholder -- other smooth trial_fxns are almost certainly to be preferred, and there are a couple routes available to construct more expressive trial_fxns.

The inclusion of the Laplacian isn't quite as free of a design choice here, but note that the PR did not compute the Laplacian as `trace(hessian(trial_fxn)(x))`. Instead, this PR provided [a basic utility](https://github.com/proteneer/timemachine/blob/c127d8cf34dac2f1403e6ae8f0764a91ff38d8eb/control_variates/trial_fxns/pair.py#L33-L61) for computing the Laplacian in a way that scales linearly with the dimension of `x` for a restricted class of user-specified trial_fxns. Although this is still a big(!) constant factor slower than it could be, I think it's nearly fast enough for practical use.

(Overall, postprocessing each lambda window in the example takes about 3-5 seconds on the CPU, to yield a variance reduction of [about an order of magnitude for the highest-variance windows](https://github.com/proteneer/timemachine/blob/d4802d1aae33e0cc14dc3ab678a11049475edbe6/scripts/controlled_ti/figures/variance_comparison.png).)

>I'm happy to merge this as-is right now though.

Before merging, I will finish adding a couple tests I should have included in the first draft

maxentile:
Closing since we are moving towards using BAR / MBAR, and variance is less of a concern.

PR #439, opened 2021-05-25
Set random seed for MC barostat test initial velocities

maxentile:
I introduced a test in https://github.com/proteneer/timemachine/pull/433 that appears to be failing stochastically by small margins on unrelated PRs... Although the integrator seed is set, I missed setting the numpy random seed for initial velocities. This annoyance will be avoided completely when the slow barostat test is replaced with the 30x faster barostat implementation in https://github.com/proteneer/timemachine/pull/436 !

### DIFF
--- a/slow_tests/test_barostat_molecular_ideal_gas.py
+++ b/slow_tests/test_barostat_molecular_ideal_gas.py
@@ -1,4 +1,5 @@
 import numpy as np
+np.random.seed(2021)
 from simtk import unit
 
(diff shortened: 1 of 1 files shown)


badisa:
Should we favor #436 over this change?

maxentile:
>Should we favor #436 over this change?

Yes -- this was just intended to remove randomness in CI for unrelated PRs until #436 is merged. Will close since #436 will be merged soon.

PR #442, opened 2021-06-02
Sequential adaptation of nonequilibrium lambda schedule

maxentile:
This PR adds a more conservative approach to protocol optimization than the one in https://github.com/proteneer/timemachine/pull/437. Here we sequentially adapt the _number and spacing of lambda windows_ along a predefined 1D path, but don't change the _shape_ of the path.

- [x] Add [function](https://github.com/proteneer/timemachine/blob/a63b7af6facf523f79c04db0e22964c13ec3ff9d/scripts/noneq_adaptation/adapt_noneq.py#L97-L143) for sequentially adapting the spacing of lambda windows for nonequilibrium switches to keep the rate of increase of stddev(work) approximately constant over the course of the protocol. Nonequilibrium switches are implemented by alternating ["propagation" ("running MD at a fixed value of lambda")](https://github.com/proteneer/timemachine/blob/a63b7af6facf523f79c04db0e22964c13ec3ff9d/scripts/noneq_adaptation/adapt_noneq.py#L140) and ["perturbation" ("incrementing lambda")](https://github.com/proteneer/timemachine/blob/a63b7af6facf523f79c04db0e22964c13ec3ff9d/scripts/noneq_adaptation/adapt_noneq.py#L135). Adaptation is performed by monitoring a population of in-progress nonequilibrium switching trajectories, and using information from the population to determine the size of the next lambda increment. Works here are computed numerically "the A.I.S. way", as a sum of `u(x, lam[t+1]) - u(x, lam[t])` increments. (This way of computing the works encodes an assumption that each "propagation" step is a p(x; lam)-reversible markov kernel. Thermostatted MD targeting exp(-u(x; lam)) generally violates this assumption, potentially by a lot, but some thermostats violate this assumption less than others.)
- [x] Add [demo script](https://github.com/proteneer/timemachine/blob/a63b7af6facf523f79c04db0e22964c13ec3ff9d/scripts/noneq_adaptation/deploy.py#L28-L46) for running nonequilibrium MD using `ctxt.multiple_steps(lambda_schedule)` with `lambda_schedule` linearly interpolating the sequence of lambdas obtained by sequential adaptation. (Instead of making a big increment to lambda every hundred MD steps or so, make a small lambda increment every step.) Works are computed numerically "the nonequilibrium TI way", by [`trapz(du_dl, lambda_schedule)`](https://github.com/proteneer/timemachine/blob/a63b7af6facf523f79c04db0e22964c13ec3ff9d/scripts/noneq_adaptation/deploy.py#L115-L116). This is consistent with how such a protocol would be deployed in the nonequilibrium docking workflow. (In addition to the earlier assumption, this way of computing the work relies on the curvature d^2u/dlam^2 being small enough to avoid incurring much T.I. quadrature error.)
- [x] Add [plotting script](https://github.com/proteneer/timemachine/blob/a63b7af6facf523f79c04db0e22964c13ec3ff9d/scripts/noneq_adaptation/plot.py), including a diagnostic figure of [stddev(work increments) per MD step](https://github.com/proteneer/timemachine/blob/a63b7af6facf523f79c04db0e22964c13ec3ff9d/scripts/noneq_adaptation/plot.py#L27-L39). The resulting comparison of default vs. optimized protocols, for various settings of the tuning parameter that governs the "speed" of the optimized protocol:

default_vs_optimized

This figure indicates that lambda spacings obtained using a sequential adaptation method may provide an efficiency improvement when transferred to the nonequilibrium docking workflow, although it doesn't perfectly equalize the rate at which stddev(work) increases (esp. near lambda=1).

-----------------

- [ ] A niche investigation we may become interested in later is to compare these two numerical approaches to computing the work. I expect they should agree with each other in the limit of slow switching, but can arbitrarily disagree with each other in limits of fast switching or high curvature. They also have different computational costs (1 `du_dl(x_t)` call per MD step vs 2 `u(x_t, lam)` calls per perturbation step), and are probably impacted differently by timestep error. A related observation is that the stddev of the work when computed the nonequilibrium TI way seems to saturate around ~6-8 kT for this system -- even at very fast switching, when the work distribution is shifted >50kT. This may be a hint of some sort of stabilizing bias associated with the nonequilibrium TI way of computing work, but I don't have a good feel for this. (There are also more general ways to compute nonequilibrium work that are unbiased and that expose opportunities for variance reduction, such as [the CFT way](https://www.pnas.org/content/108/45/E1009) and [the SMC way](https://www.cs.ubc.ca/~arnaud/delmoral_doucet_jasra_smcsamplers_jrssb.pdf).)

- [ ] A future opportunity for a significant speed-up is to supply a fast version of [`[u(x) for x in xs]`](https://github.com/proteneer/timemachine/blob/a63b7af6facf523f79c04db0e22964c13ec3ff9d/scripts/noneq_adaptation/adapt_noneq.py#L36-L38), so that we don't incur the overhead of copying back and forth to the Python layer each loop iteration. (Computing potential energies in a Python loop takes about 2 milliseconds per iteration, about ~10x longer than [it should probably take](https://github.com/proteneer/timemachine/pull/351#issue-576524289) in timemachine's float32 mode. If it were cheaper to evaluate batches of energies through timemachine's Python bindings, that would reduce the performance penalty associated with performing more frequent or precise schedule adaptation from Python, among other uses.)


.gitignore
```diff
@@ -6,3 +6,5 @@ timemachine/cpp/build/
 *.DS_Store
 oe_license.txt
 *.swp
+*.npy
```
badisa, line 9: 👍

requirements.txt
```diff
@@ -10,4 +10,5 @@ jax==0.2.9
 jaxlib==0.1.59
 hilbertcurve==1.0.5
 networkx==2.5
-cvxpylayers==0.1.5 # this depends on jax -- upgrade when jax / jaxlib are upgraded
+cvxpylayers==0.1.5  # this depends on jax -- upgrade when jax / jaxlib are upgraded
+tqdm                # progress bars
```
badisa, line 14: Would be good to pin this version to be safe

scripts/noneq_adaptation/adapt_noneq.py
```diff
@@ -0,0 +1,174 @@
+from md.thermostat.utils import sample_velocities
```
badisa, line 1: Perhaps a bit of an philosophical question, but what is the difference between things that go in the `scripts` dir vs `examples`?

maxentile: Not sure -- we could set a more deliberate convention here. `scripts/` seems appropriate for obtaining a numerical result we may reuse, or developing more "aspirational" features / use cases. `examples/` seems more appropriate for demonstrating intended usage of core functionality.

requirements.txt
proteneer: +1 on pinning this before merging.

maxentile: Addressed in https://github.com/proteneer/timemachine/pull/442/commits/06562616bffe6e0a0c64e04d1697c6efd4221235

jchodera:
>  A niche investigation we may become interested in later is to compare these two numerical approaches to computing the work. I expect they should agree with each other in the limit of slow switching, but can arbitrarily disagree with each other in limits of fast switching or high curvature. They also have different computational costs (1 du_dl(x_t) call per MD step vs 2 u(x_t, lam) calls per perturbation step), and are probably impacted differently by timestep error. A related observation is that the stddev of the work when computed the nonequilibrium TI way seems to saturate around ~6-8 kT for this system -- even at very fast switching, when the work distribution is shifted >50kT. This may be a hint of some sort of stabilizing bias associated with the nonequilibrium TI way of computing work, but I don't have a good feel for this. (There are also more general ways to compute nonequilibrium work that are unbiased and that expose opportunities for variance reduction, such as the CFT way and the SMC way.)

This is a very interesting observation! We know that the "nonequilibrium TI way" is incorrect, though---it should give provably wrong estimates, while the CFT way (presumably including shadow work) should give exact estimates and the SMC way (including just protocol work) should give pretty close estimates if BAOAB is used for propagation.

If you are able to separately compute _just_ the potential components associated with the alchemical subsystem and its interactions, this should be a very small fraction of the overall computational cost of computing the total system potential energy, so you should be able to compute the contribute to protocol work (`u(x, lam[t+1]) - u(x, lam[t])`) at negligible additional cost. This should give you near-exact estimates at minimal overall cost.

maxentile:
Thanks!

>If you are able to separately compute just the potential components associated with the alchemical subsystem and its interactions, this should be a very small fraction of the overall computational cost of computing the total system potential energy, so you should be able to compute the contribute to protocol work (u(x, lam[t+1]) - u(x, lam[t])) at negligible additional cost.

I had overlooked that possibility when commenting on difference in costs! I think adding a fast function for computing just the alchemical contributions would also accelerate several other tasks I'm interested in, where I want to compute `u(x, lam, params)` for fixed `x` and `params` but several values of `lam` (aka provide Jax a fast batching [rule for](https://jax.readthedocs.io/en/latest/jax.html#jax.vmap) `vmap(u, in_axes=(None, 0, None))(x, lam, params)`). This isn't blocking the current implementation, but is something I plan to add support for in the near future, and for which @proteneer has suggested a fast implementation approach.

>This is a very interesting observation! We know that the "nonequilibrium TI way" is incorrect, though---it should give provably wrong estimates, while the CFT way (presumably including shadow work) should give exact estimates and the SMC way (including just protocol work) should give pretty close estimates if BAOAB is used for propagation.

I should clarify what I meant by the "AIS way", "SMC way", "CFT way", and "nonequilibrium TI way", and add a bit more commentary that may help orient any others following along!

The way I see it, there are two main ways to think about computing nonequilibrium work: "importance sampling on path space" or "nonequilibrium T.I."  In the first approach, "nonequilibrium works" are identified with "- log importance weights" of sampled "forward" paths evaluated in a "reverse" path ensemble, and we are concerned both with computing these pathwise importance weights, and with estimating things like free energies using them. In the second, we use the numerical machinery from T.I., but apply it to trajectories that violate T.I.'s assumption of being in equilibrium.

A friendly introduction to the "importance sampling on path space" approach, written for a broad ML audience, is given in [[Habeck, 2017] Model evidence from nonequilibrium simulation](https://papers.nips.cc/paper/2017/hash/4da04049a062f5adfe81b67dd755cecc-Abstract.html). 

Within the "importance sampling on path space" view, I think of 3 main sub-approaches to defining and computing these importance weights, all of which can lead to consistent, unbiased estimators of normalizing constants:
* The "SMC way" appears the most general to me. Importance weights in sequential Monte Carlo samplers are computed between a forward path distribution and a reverse path distribution that can be defined independently (as in eq. 12 of [[Del Moral et al, 2005] "SMC samplers"](https://www.cs.ubc.ca/~arnaud/delmoral_doucet_jasra_smcsamplers_jrssb.pdf)). Independently defined forward and reverse kernels don't have any physical significance I can think of, but they do open the door to variance reduction options. (E.g. it is possible to reduce the variance by optimizing free parameters associated with the reverse path distribution.)
* The "CFT way" can be viewed as a special case of the SMC way in this context, where the reverse path distribution is defined using the time-reversal of the forward path distribution, as in the Crooks Fluctuation Theorem. In a computer simulation, the forward path distribution could be defined on paths that have "sub-timestep resolution" (by formally keeping track of every position and velocity update), and the reverse process is obtained by time-reversing all of the integrator substeps. (Many examples of this are included in https://www.pnas.org/content/108/45/E1009 .) For some ways to construct Langevin integrators, the terms that appear in the resulting log importance weight are convenient to compute, and have a physical interpretation as "protocol work" and a housekeeping "heat" (or "shadow work").
* The "AIS way" is a special case, where the forward kernels are reversible MCMC moves, as in Annealed Importance Sampling. This cancels out all of the forward and reverse terms in the log importance weights, so that it can be computed using just the "protocol work" terms `u(x, lam[t+1]) - u(x, lam[t])` with no additional bookkeeping.

It is common practice in the MD world to compute works the "AIS way" _even when the forward kernels are irreversible_, by neglecting the degree to which the forward kernels violate the assumption of being reversible. This introduces _bias_ relative to the more complete bookkeeping approaches. As shown in [[Sivak, Chodera, Crooks 2013]](https://arxiv.org/abs/1107.2967), there are situations when this bias dominates the free energy estimate and must be eliminated or corrected for.

It's worth noting that a biased approach can still lead to estimates with lower _overall error_ at a fixed number of samples, if the bias correction term would have contributed "more variance than it's worth."  This can be the case if the timestep error is negligible to start with, or if we have a small sample budget and the bias-correcting "shadow work" term has much higher variance than the "protocol work" term. This bias can be reduced by any method that reduces or eliminates timestep error. Options include: running thermostatted MD with a small timestep, running MD with a nicer thermostat even at a large timestep, running Metropolis-corrected MD, etc... There are complicated trade-offs here, but I think a good trade-off in some situations (adopted in BLUES, for example) is to run with a particularly nice thermostat (the "BAOAB" thermostat by Leimkuhler and Matthews) and then not worry about the timestep error. A few years ago, a couple of us in John's lab had looked into this bias with an emphasis on differences among thermostats -- the plots in this issue comment may be help illustrate this point https://github.com/choderalab/neq-switching-shadow-work/pull/1#issue-196746762 . (Summary: Because a thermostat like BAOAB makes very small sampling error, its behavior is often practically indistinguishable from an unbiased MCMC move, so the bias is pretty much negligible to start with. A thermostat that makes larger sampling errors will incur larger bias.) I think the default approach in the MD literature has been to omit correction for this bias, which I think can be more or less safe depending on how much timestep error is being made.

--------

The nonequilibrium T.I. approach to computing work, on the other hand, is something I'm far less familiar with, and perhaps @proteneer , @schmolly , or @badisa may want to chime in here. Speculatively, I wonder if the biases of the nonequilibrium T.I. approach may have the effect of "stabilizing" the resulting work distribution for the protocol in `timemachine`, given that the stddev of the work doesn't increase as dramatically as I would expect as we take the limit of very fast switching.

I will note that the nonequilibrium T.I. approach has been used successfully in some recent publications, such as https://doi.org/10.1038/s42004-021-00498-y and https://pubs.rsc.org/en/content/articlelanding/2020/sc/c9sc03754c#!divAbstract by Gapsys and colleagues.

If I were to construct counterexamples where the nonequilibrium T.I. approach gives clearly incorrect results, I think I would take the limit of fast switching or high curvature. (E.g. construct a high-curvature "round-trip" alchemical path where `u(x, lam=0) = u(x, lam=1)` everywhere, but `|du_dlam(x, lam)| >> 0`. If we take the limit of short switching trajectories, we can make the nonequilibrium TI estimate have any bias and variance we want, while the importance sampling estimates remain unbiased, and can even have zero variance if we want.)

It may be useful at some point to measure the behavior of this numerical bias for practical switching lengths and protocol curvature, to assess if we would benefit in practice from adopting one of the unbiased (or less-biased) options.

For the purposes of this PR though, I just wanted to highlight the difference between the way I computed works during protocol adaptation and the way we compute works elsewhere in `timemachine`, since I think that difference helps explain why the stddev(work increment) plots aren't completely flattened out by this style of protocol adaptation.

scripts/noneq_adaptation/utils.py
```diff
@@ -0,0 +1,27 @@
...
+from typing import Callable, List
+from md.states import CoordsVelBox
+
+Lambda = float
+Energies = np.array
+VectorizedPotentialEnergy = Callable[[List[CoordsVelBox], Lambda], Energies]
+Samples = List[CoordsVelBox]
+
+
+def compute_work_increments(
```
badisa, line 12: Do we want to have imports between files in the scripts dir? The dir name would imply to me that everything is 'stand alone'.

maxentile: What would you prefer? (Intent of [splitting](https://github.com/proteneer/timemachine/pull/442/commits/39e2eebe27fa48e005bf01bfdfafe0d7f759bdca) into 3 scripts rather than 1 was to ease later extraction to wherever in the `docking` or `fe` modules would be most appropriate, if we decide to use this style of schedule adaptation)

badisa: It would see that if the logic is used by three scripts and intended to be reused, should live in a meaningful timemachine module.

badisa: Since this seems more applicable to docking, perhaps that is where it should go? Or having an `optimize.protocol` module.

This seems within scope, just don't want to 'lose' production code to a scripts directory

maxentile: Sounds good! Adding to the `optimize` module makes more sense than the alternatives I had initially thought of.

scripts/noneq_adaptation/adapt_noneq.py
badisa: After our discussion in person, it seemed that `scripts/` was for 'infrastructure' utilities rather than exploratory work. Perhaps this could be our first 'attic' item???

maxentile: Following your suggestion here https://github.com/proteneer/timemachine/pull/442#discussion_r649370902 , I've moved the reusable components into `optimize.protocol`. Will move the usage example into `tests/`, `examples/`, or a new `attic/` folder tomorrow as you prefer.

badisa: Whatever you think is best, can see the argument for any of those.

maxentile:
Closing for now to reduce branch-indexing CI load, may reopen if protocol optimization becomes a higher priority again

PR #443, opened 2021-06-02
Task/add barostat to training

badisa:


fe/estimator.py
```diff
@@ -139,29 +148,32 @@ def _deltaG(model, sys_params) -> Tuple[Tuple[float, List], np.array]:
         bp = unbound_pot.bind(np.asarray(params))
         bound_potentials.append(bp)
 
+    func_args = []
+    for lamb in model.lambda_schedule:
```
badisa, line 152: I pulled this out as I found it clumsy to modify two function calls any time I changed the function signature

fe/model.py
```diff
@@ -91,19 +93,34 @@ def predict(self, ff_params: list, mol_a: Chem.Mol, mol_b: Chem.Mol, core: np.nd
...
             integrator = LangevinIntegrator(
-                300.0,
+                temperature,
                 1.5e-3,
                 1.0,
                 masses,
                 seed
             )
 
+            barostat = MonteCarloBarostat(
```
badisa, line 115: I need to rethink the API for the barostat needs to be rethought, right now it feels clumsy

fe/estimator.py
maxentile: Nice!

Another way we might refactor this is to avoid the need for two code paths predicted on `model.client is None`, by adding a `NullClient` or similar in [parallel/client.py](https://github.com/proteneer/timemachine/blob/e8773e6456cdc70141ab9d17e2773d4fc1f2f111/parallel/client.py) ?

tests/test_free_energy.py
```diff
@@ -170,14 +186,31 @@ def vacuum_model(ff_params):
...
             masses,
             seed
         )
 
+        barostat = MonteCarloBarostat(
+            x0.shape[0],
+            pressure,
+            temperature,
+            group_idxs,
+            25,
```
proteneer, line 211: We may want to use a higher frequency during equilibration (eg. 5), and use only 25 for production.

fe/estimator.py
proteneer: +1

fe/model.py
proteneer: Not that bad imo

maxentile (approved):

fe/estimator.py
maxentile: Nicely resolved in https://github.com/proteneer/timemachine/commit/4def6196c6a5d75116c76f830bd8910bfb0e2c58

badisa:
I am going to delay merging this until I get some variance measures to be sure that this didn't add some subtle behavior that the existing tests don't catch

proteneer (approved):
lgtm!

PR #444, opened 2021-06-02
Return Box dimensions from multiple_steps

badisa:


docking/dock_and_equilibrate.py
```diff
@@ -153,7 +153,7 @@ def dock_and_equilibrate(
         # collect a du_dl calculation once every other step
         subsample_interval = 1
 
-        full_du_dls, _ = ctxt.multiple_steps(
+        full_du_dls, _, host_box = ctxt.multiple_steps(
```
badisa, line 156: @schmolly I used the box here as the report_step call relies on the box coords. This might need more detailed review to ensure this behaves as you expect it too. Same with the docking/pose_dock.py change

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -167,7 +167,10 @@ void declare_context(py::module &m) {
         py::array_t<double, py::array::c_style> out_x_buffer({F, N, D});
         std::memcpy(out_x_buffer.mutable_data(), result[1].data(), result[1].size()*sizeof(double));
 
-        return py::make_tuple(out_du_dl_buffer, out_x_buffer);
+        py::array_t<double, py::array::c_style> box_buffer({D, D});
```
maxentile, line 170: Question: do we want this to be of shape `{F, D, D}` (store box vectors at every frame for which we return coordinates), or `{D, D}` (final box vectors only)?

proteneer: +1 - we need this to be `{F, D, D}`.

proteneer: In particular, `F` should be identical to the `F` in out_x_buffer

docking/dock_and_equilibrate.py
badisa: I undid this as the desired behavior was to return the boxes for each x interval, which would give you a different box.

timemachine/cpp/src/wrap_kernels.cpp
maxentile: Nicely resolved in https://github.com/proteneer/timemachine/pull/444/commits/66f1375e0c73256a7dea74a59d371cd803ca3b4d

tests/test_md.py
```diff
@@ -210,8 +210,10 @@ def integrate_once_through(
             test_xs,
             ref_all_xs[::x_interval]
         )
+        self.assertEqual(test_boxes.shape[0], test_xs.shape[0])
+        self.assertEqual(test_boxes.shape[1], D)
+        self.assertEqual(test_boxes.shape[2], test_xs.shape[2])
```
maxentile, line 215: Could we also add an assertion that `test_boxes[0]` and `test_boxes[-1]` are  consistent with `ctxt_2.get_box()` before and after trajectory?

maxentile (approved):
Nice! This behaves as expected in local testing, and I think all relevant usages of `ctxt.multiple_steps` have been updated.

Could probably also replace usages like this https://github.com/proteneer/timemachine/blob/be9db7961ab8d0007740ebc7bb764c519f3f26c5/tests/test_barostat.py#L171-L175 if desired

tests/test_md.py
proteneer: I don't think this is guaranteed since boxes, like coords, use "pre-step" saving:

``` context.cu

        for(int i=0; i < lambda_schedule.size(); i++) {
            // decide if we need to store the du_dl for this step
            unsigned long long *du_dl_ptr = nullptr;
            if(i % store_du_dl_interval == 0) {
                // pemdas but just to make it clear we're doing pointer arithmetic
                du_dl_ptr = d_du_dl_buffer + (i / store_du_dl_interval);
            }

            if(i % store_x_interval == 0) {
                gpuErrchk(cudaMemcpy(
                    &h_x_buffer[0] + (i / store_x_interval)*N_*3,
                    d_x_t_,
                    N_*3*sizeof(double),
                    cudaMemcpyDeviceToHost)
                );
            }

            double lambda = lambda_schedule[i];
            this->_step(lambda, du_dl_ptr);
        }
```

badisa: It is guaranteed, at least in this test, as there is no barostat.

proteneer (approved):
merge away!

PR #446, opened 2021-06-07
Add pytest.ini

badisa:
* Makes the pytest configuration more standard

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,5 @@
+[pytest]
+addopts = -svx
+testpaths =
+    tests/
+    slow_tests/


badisa:
Mostly to see if the changes made to CI are working.

pytest.ini
```diff
@@ -0,0 +1,5 @@
+[pytest]
+addopts = -sv
```
proteneer, line 2: Did you want to enable -x as well? (This way the test fails immediately as opposed to running everything)

badisa: Will add

proteneer (approved):

PR #447, opened 2021-06-09
Change default precision of Python wrappers from F64 -> F32

maxentile:
There appears to be about [a 10-15x performance difference](https://gist.github.com/maxentile/c7a3624c3d9579e281f7787834c709ab) between precision modes for `ctxt.multiple_steps`, but [no apparent accuracy impact](https://github.com/proteneer/timemachine/pull/436#discussion_r639239726) for uses so far, so this would be a more sensible default. (Also, the default throughout the rest of timemachine has been to use the float32 implementations, so this was an outlier.)

The performance difference between precision modes for the wrapped unbound potentials is less dramatic, [maybe 20-40%](https://gist.github.com/maxentile/520e2a04c76f424619b068d8d7e889e7), but I updated these as well for consistency.

### DIFF
--- a/fe/functional.py
+++ b/fe/functional.py
@@ -37,7 +37,7 @@ def wrap_impl(impl):
     return U
 
 
-def construct_differentiable_interface(unbound_potentials, precision=np.float64):
+def construct_differentiable_interface(unbound_potentials, precision=np.float32):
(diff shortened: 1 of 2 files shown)


badisa (approved):
LGTM!

PR #448, opened 2021-06-15
Possibly JIT-safer `delta_r`

maxentile:
On more recent jax/jaxlib versions than the versions pinned in requirements.txt, I noticed that `jit(delta_r)(ri, rj, box)` sometimes produces different results from `delta_r(ri, rj, box)`.

[Replacing](https://github.com/proteneer/timemachine/blob/1ed3176c56205726aebaff6b24d8ed1616f432ec/timemachine/potentials/jax_utils.py#L50-L56) the Python for-loop over `dims` with a [`lax.fori_loop`](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.fori_loop.html) seems to safeguard against this possibility.

### DIFF
--- a/tests/test_jax_utils.py
+++ b/tests/test_jax_utils.py
@@ -8,12 +8,26 @@ onp.random.seed(2021)
 from timemachine.potentials.jax_utils import delta_r
 
 
-def test_jitted_delta_r_symmetric():
-    """assert jit(delta_r)(ri, rj, box) == - jit(delta_r)(rj, ri, box)"""
-    ri, rj = onp.random.randn(2, 1000, 3)
-    box = np.eye(3)
+def test_delta_r():
+    """assert that
+    * delta_r(ri, rj, box) == - delta_r(rj, ri, box)
+    * and that delta_r agrees with jit(delta_r)
+    on a few random inputs of varying size
+    """
 
-    dr_ij = jit(delta_r)(ri, rj, box)
-    dr_ji = jit(delta_r)(rj, ri, box)
+    for _ in range(5):
+        n_atoms = onp.random.randint(50, 1000)
+        dim = onp.random.randint(3, 5)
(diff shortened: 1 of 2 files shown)


maxentile:
I haven't yet attempted to find which jax/jaxlib versions might be associated with the behavior, but this might turn out to be something to report to the Jax developers.

For posterity, here's [a colab notebook](https://colab.research.google.com/drive/1hPUE_66VWJ6KZrZLrrm368cI39rp5sRE?usp=sharing) that shows unexpected behavior with the Python-loop implementation of delta_r and expected behavior with the lax.fori_loop implementation, in a fresh colab environment (with `jax==0.2.13`, `jaxlib==0.1.66+cuda110`)

proteneer:
```jit(delta_r)(ri, rj, box) sometimes produces different results from delta_r(ri, rj, box)```

jit is not guaranteed to produce bitwise identical results as non-jit version. I noticed you were doing `==` checks in the notebook.

maxentile:
>jit is not guaranteed to produce bitwise identical results as non-jit version. I noticed you were doing == checks in the notebook.

Thanks for catching this -- I didn't print the magnitude of the differences in the notebook. The differences between d(i, j) and d(j, i) in the colab notebook are within machine precision, ~1e-16.

Note that the colab tests symmetry between d(i, j) and d(j, i) using ==, but does not directly compare delta_r and jit(delta_r).

The tests [use](https://github.com/proteneer/timemachine/blob/8fa4d85113a3634d468122aca8d96ae67782ded4/tests/test_jax_utils.py#L32) numpy.testing.assert_allclose for all comparisons.

I'll close this until more confident this has anything to do with the initial observation (occasional large differences in du_deps between a refactored version of nonbonded_v3 and the current version of nonbonded_v3 (although energies and other derivatives appeared identical)).

PR #449, opened 2021-06-15
Task/train on dg of cycles

badisa:
Curious to see what you think @maxentile regarding the addition of an inference mode along with the attempt at a 'generic' method to train on batches. 

Breaking Change:
* The step indices are no longer a flat list


examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -120,9 +125,36 @@ def _blew_up(results: List[SimulationResult]) -> bool:
...
+        indices.append(list(index.get_transform_indices(rfe)))
+        pred_ddG, stage_results = model.predict(ff_params, rfe.mol_a, rfe.mol_b, rfe.core)
+        all_results.extend(list(stage_results))
+        preds.append(pred_ddG)
+    # If the edges in the batch are within a cycle, correct the ddGs
+    indices = jnp.asarray(indices)
+    ind_l, ind_r = indices.T
+    try:
+        layer = construct_mle_layer(len(index), indices)
+        corrected_dg = layer(jnp.asarray(preds))
```
badisa, line 148: Perhaps we should make this be the identity function if the edges are disconnected?

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -330,40 +366,42 @@ def run_validation_edges(validation: Dataset, params, systems, epoch):
...
-        epoch += 1
-        print(f"Epoch: {epoch}")
-        for i in steps:
-            rfe = training.data[i]
-            if getattr(rfe, "complex_path", None):
-                model = systems[rfe.complex_path]
-            else:
-                model = systems[protein_path]
-            # compute a step, measuring total wall-time
+        run_validation_edges(validation, ordered_params, systems, epoch+1, inference=args.inference_only)
```
badisa, line 388: A bit hacky to run all edges as validation in inference mode, but was quick and easy. Perhaps the function should just be renamed?

fe/estimator.py
```diff
@@ -163,7 +163,7 @@ def _deltaG(model, sys_params) -> Tuple[Tuple[float, List], np.array]:
     futures = []
     for lamb in model.lambda_schedule:
         args = (lamb, model.box, model.x0, model.v0, bound_potentials, model.integrator, model.equil_steps, model.prod_steps, model.barostat)
-        kwargs = {}  # Unused for now
+        kwargs = {"x_interval": 0}  # Unused for now
```
badisa, line 166: Remove, was hitting memory limits on longer simulations thanks due to the default being 1000. Perhaps we should default to a proportion of the prod steps?

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -120,9 +125,36 @@ def _blew_up(results: List[SimulationResult]) -> bool:
...
+        layer = construct_mle_layer(len(index), indices)
+        corrected_dg = layer(jnp.asarray(preds))
+        cycle_corrected_rbfes = corrected_dg[ind_r] - corrected_dg[ind_l]
+    except AssertionError:
+        # Provided non connected graph, ie its not a cycle
+        cycle_corrected_rbfes = jnp.asarray(preds)
+    labels = jnp.asarray([rfe.label for rfe, _ in batch])
+    loss = jnp.mean(jnp.abs(cycle_corrected_rbfes - labels))
+    # Aggregate using mean, then compute psuedo huber loss
+    loss = pseudo_huber_loss(loss)
```
maxentile, line 156: Could you comment a bit more on this step?

Would have expected to aggregate in the other order, perhaps as

```python
residuals = cycle_corrected_rbfes - labels
loss_terms = pseudo_huber_loss(residuals)
loss = mean(loss_terms)
```
or similar

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: Doing this could be more convenient, would just require changing the output type of `layer`.

Note that the input and output shapes of `layer` are not the same. (As input it can accept arrays of as many `simulated_rbfes` and `simulated_abfes` as are available (which may be `>= n_nodes`), and as output it always returns an array of `n_nodes` absolute dgs.)

At this check, https://github.com/proteneer/timemachine/blob/48623960f6dd114e0d2b05b33f55b9278b3074de/fe/cycles.py#L110-L113 we currently raise an assertion error. If instead of raising an error we wanted to return a function that passed its arguments through unmodified, then I think the function would need to return arrays of `corrected_ddgs` and `corrected_dgs` of the same shapes as the input `simulated_rbfes` and `simulated_abfes`.

(I think I would prefer to call `construct_mle_layer` only on connected comparison networks and keep the output type as-is.)

maxentile (commented):
Nice! I have a minor comment about the loss aggregation but otherwise looks like a great improvement.

examples/hif2a/fit_to_multiple_rbfes.py
badisa: You are totally right, woops!

examples/hif2a/fit_to_multiple_rbfes.py
badisa: I am happy with this as it is, just remember the non-differentiable version supporting this case so wasn't sure if we wanted to unify it.

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: Nice! Addressed in https://github.com/proteneer/timemachine/pull/449/commits/d89ff27d1e7a2a3a3507353c099a6164b8378867

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: Follow-up: Do we generally want to define the loss in terms of [edges](https://github.com/proteneer/timemachine/blob/48623960f6dd114e0d2b05b33f55b9278b3074de/tests/test_cycle_closure.py#L67-L77), or in terms of [nodes](https://github.com/proteneer/timemachine/blob/48623960f6dd114e0d2b05b33f55b9278b3074de/tests/test_cycle_closure.py#L110-L116)?

maxentile (approved):

examples/hif2a/fit_to_multiple_rbfes.py
badisa: Thats a good question, and something I don't have a good answer for. It seems like `nodes` would provide better generalizability (at least to ABFE), and that is what our actual underlying data is.

proteneer: If we're running cycle closure, we should probably define it in terms of dGs (nodes).

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -120,9 +125,36 @@ def _blew_up(results: List[SimulationResult]) -> bool:
...
+    try:
+        layer = construct_mle_layer(len(index), indices)
+        corrected_dg = layer(jnp.asarray(preds))
+        cycle_corrected_rbfes = corrected_dg[ind_r] - corrected_dg[ind_l]
+    except AssertionError:
+        # Provided non connected graph, ie its not a cycle
+        cycle_corrected_rbfes = jnp.asarray(preds)
+    labels = jnp.asarray([rfe.label for rfe, _ in batch])
+    loss = pseudo_huber_loss(cycle_corrected_rbfes - labels)
+    # Aggregate the pseudo huber los using mean
```
proteneer, line 155: typo los

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -120,9 +125,36 @@ def _blew_up(results: List[SimulationResult]) -> bool:
...
+        all_results.extend(list(stage_results))
+        preds.append(pred_ddG)
+    # If the edges in the batch are within a cycle, correct the ddGs
+    indices = jnp.asarray(indices)
+    ind_l, ind_r = indices.T
+    try:
+        layer = construct_mle_layer(len(index), indices)
+        corrected_dg = layer(jnp.asarray(preds))
+        cycle_corrected_rbfes = corrected_dg[ind_r] - corrected_dg[ind_l]
+    except AssertionError:
```
proteneer, line 150: We may want to subclass this exception at some point - not sure if this might silently squash actual , non-connectivity related errors.

badisa: Currently it could, I can add in a more explicit exception

examples/hif2a/fit_to_multiple_rbfes.py
badisa: Is it reasonable to train on ddG for non-cyclic edges and training on dG for cyclic edges?

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -330,40 +366,42 @@ def run_validation_edges(validation: Dataset, params, systems, epoch):
...
     np.random.seed(args.seed)
 
+    batch_size = args.batch_size
     step_inds = []
     for epoch in range(num_epochs):
         inds = np.arange(len(training.data))
         np.random.shuffle(inds)
-        step_inds.append(inds)
+        batched_inds = []
+        num_steps = int(np.ceil(len(inds) / batch_size))
```
proteneer, line 375: is this actually robust against floating point errors? not sure if len(inds) / batch_size will round off to 2.0000001 or something and

fe/estimator.py
proteneer: Alternative is to strip out waters, but then again we may want to keep the waters in case we want to debug sampling problems.

examples/hif2a/fit_to_multiple_rbfes.py
proteneer: If the edges are non-cyclic, you can trivially induce ddGs to dGs. But from a practical standpoint it should be equivalent.

badisa: How would I induce the dGs in this case?

examples/hif2a/fit_to_multiple_rbfes.py
badisa: Replace with integer arithmetic rather than floating point

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -120,9 +125,36 @@ def _blew_up(results: List[SimulationResult]) -> bool:
...
+        preds.append(pred_ddG)
+    # If the edges in the batch are within a cycle, correct the ddGs
+    indices = jnp.asarray(indices)
+    ind_l, ind_r = indices.T
+    try:
+        layer = construct_mle_layer(len(index), indices)
+        corrected_dg = layer(jnp.asarray(preds))
+        cycle_corrected_rbfes = corrected_dg[ind_r] - corrected_dg[ind_l]
+    except DisconnectedEdgesError:
+        # Provided non connected graph, ie its not a cycle
```
maxentile, line 151: Nit: Should we modify this comment slightly to distinguish between "connected" and "contains cycles"? (a minimum spanning tree for example would be connected but not contain any cycles)

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: >If the edges are non-cyclic, you can trivially induce ddGs to dGs.

@proteneer : Could you clarify if this comment applies also to a _disconnected_ comparison network, or only to a connected comparison network that contains no cycles?

I'm not sure how to handle the case that the comparison network is disconnected.

(If disconnected, would expect to run into issues illustrated by the [example in this comment](https://github.com/proteneer/timemachine/pull/399#discussion_r618787870).)

proteneer: I think there are two aspects

1) Disconnected Components, which we can decompose into individual independent components. 
2) A given component may or may not have cycles. 

Regardless of whether or not the component has cycles, the cycle closure correction (in the absence of experimental dGs), should still be able to induce predicted dGs, shifted by some arbitrary amount (or picking an arbitrary node as zero). You can't really compare the these predict dGs without a bias (or some sort of shifting to the mean against the label dGs). You can add a constant bias on these predicted dGs as a free parameter for fitting. If there's multiple disconnected components, then *each* cycle will necessarily need its only shifted bias (which can be terrible as it adds more parameters to the model). 

If you don't want to deal with the bias corrections and training said biases, then just compare against the cycle closure corrected ddGs. But either way, the presence of a cycle vs no cycle should not affect the choice of predicting dG vs ddG, probably best to be consistent through out.

PR #450, opened 2021-06-17
Topology improvements for RABFE.

proteneer:
This PR adds support for 3 additional topology classes used in our RABFE protocol:

**BaseTopologyConversion** - used for converting molecules from a forcefield dependent to a forcefield independent state. This is used for both complex and solvent stages. During training, we would only re-run this stage.
**BaseTopologyStandardDecoupling** - used for decoupling the standard ligand from the environment. This is used in the solvent stage.
**DualTopologyStandardDecoupling** - used for decoupling one of the two standard ligands from the environment. This is used in the complex stage, and is repeated. Note that the epsilons of the standardized ligand is halved in the lambda=0 state, and interpolated to the standard strength at lambda=1. 

There are two accessory classes:
**DualTopoloyMinimization** - Used by the minimizer.py code. This doesn't do any fancy indexing, it just sets the lambda_idxs to sane values
**DualTopologyRHFE** - Used for testing the RHFE code.

Most of this PR just involves setting up the various lambda_idxs, with some parameter interpolation mixed in.

### DIFF
--- a/fe/topology.py
+++ b/fe/topology.py
@@ -589,13 +589,21 @@ class DualTopologyRHFE(DualTopology):
 
 
 def get_ring_membership(mol):
+    """
+    Get the membership of each atom in the mol. The algorithm
+    finds islands that are formed connected by bond bridges.
 
-    # for _ in mol_a.
+    Let N = mol.GetNumAtoms()
+
+    The returned array is of length N, where each atom is marked
+    by a membership class K such that 0 < K < N.
+    """
+
     g = networkx.Graph()
     for bond in mol.GetBonds():
         g.add_edge(bond.GetBeginAtomIdx(), bond.GetEndAtomIdx())
 
-    # find bridges and remove them
+    # find bridges and remove them to form islands
     for bridge in networkx.bridges(g):
         g.remove_edge(*bridge)
 
@@ -606,6 +614,7 @@ def get_ring_membership(mol):
             membership[atom_idx] = group_idx
 
     return membership
+
 
 class DualTopologyStandardDecoupling(DualTopology):
     """
--- a/fe/topology.py
+++ b/fe/topology.py
@@ -1,3 +1,7 @@
+import networkx
+from abc import ABC
+from rdkit import Chem
+
 import numpy as np
 import jax
 import jax.numpy as jnp
@@ -10,6 +14,52 @@ _SCALE_13 = 1.0
 _SCALE_14 = 0.5
 _BETA = 2.0
 _CUTOFF = 1.2
+
+
+_STANDARD_CHARGE = 0.0
+
+# ff independent "standard" sig/eps
+def simple_lj_typer(mol):
+
+    sig_eps = []
+
+    for atom in mol.GetAtoms():
+        a_num = atom.GetAtomicNum()
+        if a_num == 1:
+            assert len(atom.GetNeighbors()) == 1
+            neighbor = atom.GetNeighbors()[0]
+            b_num = neighbor.GetAtomicNum()
+            if b_num == 6:
+                val = (0.25, 0.25)
+            elif b_num == 7:
+                val = (0.10, 0.25)
+            elif b_num == 8:
+                val = (0.05, 0.02)
+            elif b_num == 16:
+                val = (0.10, 0.25)
+            else:
+                val = (0.10, 0.25)
+        elif a_num == 6:
+            val = (0.34, 0.6)
+        elif a_num == 7:
+            val = (0.32, 0.8)
+        elif a_num == 8:
+            val = (0.30, 0.9)
+        elif a_num == 9:
+            val = (0.3, 0.5)
+        elif a_num == 15:
+            val = (0.37, 0.9)
+        elif a_num == 16:
+            val = (0.35, 1.0)
+        else:
+            assert 0
+
+        # sigmas need to be halved
+        sig_eps.append((val[0]/2, val[1]))
+
+    sig_eps = np.array(sig_eps)
+
+    return sig_eps
 
 
 class AtomMappingError(Exception):
@@ -219,7 +269,7 @@ class BaseTopology():
             lambda_offset_idxs,
             beta,
             cutoff
-        ) 
+        )
 
         params = jnp.concatenate([
             jnp.reshape(q_params, (-1, 1)),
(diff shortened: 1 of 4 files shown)


fe/topology.py
```diff
@@ -390,6 +555,163 @@ def parameterize_improper_torsion(self, ff_params):
...
+    for bridge in networkx.bridges(g):
+        g.remove_edge(*bridge)
+
+    # find connected components after bridge removal
+    membership = np.zeros(mol.GetNumAtoms(), dtype=np.int32)
+    for group_idx, atom_list in enumerate(networkx.connected_components(g)):
+        for atom_idx in atom_list:
+            membership[atom_idx] = group_idx
+
+    return membership
```
maxentile, line 616: Nice! Documentation request: add comment about when to use this function vs. rdkit ringinfo https://www.rdkit.org/docs/GettingStartedInPython.html#ring-information ?

fe/topology.py
```diff
@@ -390,6 +555,163 @@ def parameterize_improper_torsion(self, ff_params):
...
+
+class DualTopologyStandardDecoupling(DualTopology):
+    """
+    Standardized variant, where both ligands A and B have their charges, sigma, and epsilons are set
+    to standard, forcefield-independent values. There is no parameter interpolation. lambda=0 has both
+    Ligand A and B fully in the pocket. lambda=1 has ligand b fully decoupled, while ligand a is kept
+    in place.
+
+    Furthermore, the ligand's non-ring torsions are turned off at the standard state to improve
+    sampling.
```
maxentile, line 627:
```suggestion
    Standardized variant, where both ligands A and B have their charges, sigmas, and epsilons set
    to standard, forcefield-independent values. There is no parameter interpolation. lambda=0 has both
    ligand A and B fully in the pocket. lambda=1 has ligand B fully decoupled, while ligand A is kept
    in place.

    Furthermore, the ligand's non-ring torsions are turned off at the standard state to improve
    sampling.
```

fe/topology.py
```diff
@@ -390,6 +555,163 @@ def parameterize_improper_torsion(self, ff_params):
...
+        # we don't actually need derivatives for this stage.
+
+        # zero out/stop grad
+        ff_q_params = np.zeros_like(ff_q_params)
+        ff_lj_params = np.zeros_like(ff_lj_params)
+
+        qlj_params, nb_potential = super().parameterize_nonbonded(ff_q_params, ff_lj_params)
+        qlj_params = jax.ops.index_update(qlj_params, jax.ops.index[:, 0], _STANDARD_CHARGE)
+        mol_c = Chem.CombineMols(self.mol_a, self.mol_b)
+        sig_eps = simple_lj_typer(mol_c)
```
maxentile, line 665: Similar comment as above -- is there a way to expose this choice in either the `parameterize_nonbonded` function signature, or in the `DualTopologyStandardDecoupling` constructor?

fe/topology.py
```diff
@@ -390,6 +555,163 @@ def parameterize_improper_torsion(self, ff_params):
...
+        qlj_params, nb_potential = super().parameterize_nonbonded(ff_q_params, ff_lj_params)
+
+        combined_lambda_plane_idxs = np.zeros(
+            self.mol_a.GetNumAtoms() + self.mol_b.GetNumAtoms(),
+            dtype=np.int32
+        )
+        combined_lambda_offset_idxs = np.concatenate([
+            np.ones(self.mol_a.GetNumAtoms(), dtype=np.int32),
+            np.ones(self.mol_b.GetNumAtoms(), dtype=np.int32)
+        ])
```
maxentile, line 708:
```suggestion
        n_A, n_B = self.mol_a.GetNumAtoms(), self.mol_b.GetNumAtoms()
        combined_lambda_plane_idxs = np.zeros(n_A + n_B, dtype=np.int32)
        combined_lambda_offset_idxs = np.ones(n_A + n_B, dtype=np.int32)
```

fe/topology.py
```diff
@@ -390,6 +555,163 @@ def parameterize_improper_torsion(self, ff_params):
...
+    def parameterize_nonbonded(self,
+        ff_q_params,
+        ff_lj_params):
+
+        # both mol_a and mol_b are standardized.
+        # we don't actually need derivatives for this stage.
+
+        # zero out/stop grad
+        ff_q_params = np.zeros_like(ff_q_params)
+        ff_lj_params = np.zeros_like(ff_lj_params)
```
maxentile, line 660: Perhaps add a line comment about what would happen if these weren't zero'd out?

fe/topology.py
```diff
@@ -12,6 +16,52 @@
 _CUTOFF = 1.2
 
 
+_STANDARD_CHARGE = 0.0
+
+# ff independent "standard" sig/eps
+def simple_lj_typer(mol):
```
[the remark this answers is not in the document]
proteneer: I'll move the charge typing in here for consistency.

The current design is pretty basic, and probably has significant room for improvement. But I'll leave that for a future PR.

fe/topology.py
```diff
@@ -253,11 +303,125 @@ def parameterize_periodic_torsion(self, proper_params, improper_params):
...
+
+    def parameterize_nonbonded(self,
+        ff_q_params,
+        ff_lj_params):
+
+        # mol is standardized into a forcefield independent state.
+        qlj_params, nb_potential = super().parameterize_nonbonded(ff_q_params, ff_lj_params)
+        qlj_params = jax.ops.index_update(qlj_params, jax.ops.index[:, 0], _STANDARD_CHARGE)
+
+        sig_eps = simple_lj_typer(self.mol)
```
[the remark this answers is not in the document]
proteneer: I'd rather not introduce `standardized_lj_params`.

`ff_q_params, ff_lj_params` are supposed to be trainable parameters. We definitely don't want to `standardized_lj_params` to be trainable. We could probably just add support for a customer typer function later on.

fe/topology.py
proteneer: rdkit RingInfo does something totally different. It is populated by either FastFindRings or SSSRs (ring basis). Neither is quite what we want.

fe/topology.py
proteneer: So I'm not actually how to design this to be a little bit more elegant. I don't want this function actually use the input args, as this topology handler should *not* be dependent on the forcefield at all. But for various ugly reasons, we need to still generate the nonbonded potential using the parameterization code. So this is just a sanity check indicating that ff_q_params and ff_lj_params can be whatever.

fe/topology.py
```diff
@@ -253,11 +303,125 @@ def parameterize_periodic_torsion(self, proper_params, improper_params):
         improper_params, improper_potential = self.parameterize_improper_torsion(improper_params)
         combined_params = jnp.concatenate([proper_params, improper_params])
         combined_idxs = np.concatenate([proper_potential.get_idxs(), improper_potential.get_idxs()])
-        combined_potential = potentials.PeriodicTorsion(combined_idxs)
+
+        # dual topology needs this fix too... probably
```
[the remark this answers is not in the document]
proteneer: This is an oversight on my part, definitely something I forgot to translate over. Will fix in this patch.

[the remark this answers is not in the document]
proteneer: PS in case you were wondering why the tests I showed earlier still worked fine, in my rabfe_jank branch, I used direct parameter replacement (i.e. setting k_torsion == 0). But while refactoring this code, I wanted to be consistent with the BaseTopology and instead modified this to use lambda_offset indices etc, and I only patched the BaseTopology parameterize_periodic_torsion, not the DualTopology variant.

Thanks for catching this!

fe/topology.py
proteneer: I'm actually not sure I want to expose this to the end-user yet. Part of this is because we need to take care that the user supplies the same simple_qlj_typer in all three classes (you're not really allowed to mix and match them in order for the end states to be connected).

fe/topology.py
proteneer: Will edit this in directly with a minor change on capitalization of the N

fe/topology.py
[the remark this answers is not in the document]
proteneer: Fixed

fe/topology.py
proteneer: I'm just going to remove this line for now, since the refactored code makes it slightly more obvious that the input qlj_params don't really do anything

``` python
    def parameterize_nonbonded(self,
        ff_q_params,
        ff_lj_params):

        # mol is standardized into a forcefield independent state.
        _, nb_potential = super().parameterize_nonbonded(ff_q_params, ff_lj_params)
        qlj_params = standard_qlj_typer(self.mol)

        return qlj_params, nb_potential
```

badisa (approved):

maxentile (approved):

PR #451, opened 2021-06-18
Feature/pre equilibrate rbfe systems

badisa:


fe/model.py
```diff
@@ -10,10 +12,23 @@
...
+        return np.abs(masses[i] - 1.00794) < 1e-3
+
+    for i, j in bond_list:
+        i, j = np.array([i, j])[np.argsort([masses[i], masses[j]])]
+        if is_hydrogen(i) and not is_hydrogen(j):
+            # H-X
+            # order of operations is important!
+            masses[j] -= multiplier*masses[i]
+            masses[i] += multiplier*masses[i]
+    return masses
```
maxentile, line 31: Question: Should this utility live in `fe/` or in `md/`?

Documentation request: add reference for which HMR scheme is used.

fe/pdb_writer.py
```diff
@@ -38,6 +38,34 @@
 
 from rdkit import Chem
 
+
+def convert_single_topology_mols(coords: np.ndarray, topo: "SingleTopology") -> np.ndarray:
```
maxentile, line 42: Question about type-hint syntax: Should this be `SingleTopology` or `"SingleTopology"`?

fe/pdb_writer.py
```diff
@@ -38,6 +38,34 @@
...
+
+
+    Example
+    -------
+
+        >>> writer = PDBWriter(...)
+        >>> lig_coords = convert_single_topology_mols(x0[num_host_coords:], topo)
+        >>> new_coords = np.concatenate((x0[:num_host_coords], lig_coords), axis=0)
+        >>> writer.write_frame(new_coords*10)
+        >>> writer.close()
```
maxentile, line 59: Nice! This will be helpful.

fe/topology.py
```diff
@@ -412,11 +412,15 @@ def __init__(self, mol_a, mol_b, core, ff):
         ff: ff.Forcefield
             Forcefield to be used for parameterization.
 
+        minimize: bool
+            Whether both R groups to be interacting at lambda=0.5
```
maxentile, line 416:
```suggestion
            Whether both R groups should be interacting at lambda=0.5
```

fe/model.py
```diff
@@ -81,23 +203,51 @@ def predict(self, ff_params: list, mol_a: Chem.Mol, mol_b: Chem.Mol, core: np.nd
...
 
             harmonic_bond_potential = unbound_potentials[0]
-            group_idxs = get_group_indices(get_bond_list(harmonic_bond_potential))
-
-            x0 = coords
-            v0 = np.zeros_like(coords)
+            bond_list = get_bond_list(harmonic_bond_potential)
+            if self.hmr:
+                masses = apply_hmr(masses, bond_list)
+                time_step = 2.5e-3
```
maxentile, line 249: I don't see a much better alternative, but I wonder if there's a way to avoid `timestep = literal; if self.hmr: timestep = different_literal`. Perhaps we want to move the responsibility for changing the masses into `rfe.prepare_host_edge(..., hmr=True)`, and make `time_step` an explicit argument in the RBFEModel constructor.

fe/pdb_writer.py
badisa: Just didn't want to do the import, can add in the import. Not sure what impact the string style has on mypy

fe/model.py
badisa: I put this in `fe/model.py` as I had seen that @proteneer had this implementation in `fe/model_utils.py`. I would think the `md` directory is the right place, and if you could comment @proteneer so that we can have one place for our two branches

proteneer: I'm okay with this going where-ever - I'll just import it. I think `md/` better as well.

proteneer: Use this link: https://pubmed.ncbi.nlm.nih.gov/26574392/

fe/pdb_writer.py
proteneer: Yeah this extra import dependency drove me crazy

fe/model.py
proteneer: We're going to refactor these classes like crazy soon with all the dual topology stuff soon to come.

fe/estimator.py
```diff
@@ -29,8 +34,44 @@ def unflatten(aux_data, children):
...
+        intg_impl,
+        all_impls,
+        barostat=baro_impl,
+    )
+
+    # equilibration
+    equil_schedule = np.ones(equil_steps)*lamb
+    ctxt.multiple_steps(equil_schedule)
+
+    return ctxt.get_x_t(), ctxt.get_v_t(), ctxt.get_box()
```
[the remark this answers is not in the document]
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/451/commits/876cfcfaa836cc0e5077aa966357a992ad9d19cf

fe/model.py
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/451/commits/5e594209bb81d9f9ed725df0af710e53fa8577d7

maxentile (approved):

PR #453, opened 2021-06-22
Jax nonbonded on specific interacting pairs

maxentile:
To accelerate applications that require many batched computations of `[u(x, lam, params) - u_ref(x) for x in xs]`, and its derivatives, it would be helpful to be able to compute just the small part of the potential that depends on lam. (Intended applications include: https://github.com/proteneer/timemachine/pull/404, https://github.com/proteneer/timemachine/pull/437, and https://github.com/proteneer/timemachine/pull/442.)

The part of the nonbonded potential that depends on lam usually contains just a small subset of interacting pairs. (E.g. just the (ligand atom, environment atom) pairs that are within a distance cutoff of each other.)

This PR adds a function [`nonbonded_v3_on_specific_pairs`](https://github.com/proteneer/timemachine/blob/6c6a081ecec395c7714dda299b8b03536fe7fb05/timemachine/potentials/nonbonded.py#L158-L182). In the proposed function, a list of specific interacting pairs are passed in as index arrays `inds_l`, `inds_r` (similar to OpenMM CustomBondForce). The main intended use involves calling the `jit`- and `vmap`-transformed versions of the function, so this is [exercised](https://github.com/proteneer/timemachine/blob/6c6a081ecec395c7714dda299b8b03536fe7fb05/tests/test_jax_nonbonded.py#L110-L151) in the tests.

Along the way, I extracted a number of small helper functions into `jax_utils.py`. A helper function is added to [compute distances on specific pairs](https://github.com/proteneer/timemachine/blob/6c6a081ecec395c7714dda299b8b03536fe7fb05/timemachine/potentials/jax_utils.py#L109-L118). A few helper functions get index arrays corresponding to common scenarios: [all-pairs](https://github.com/proteneer/timemachine/blob/6c6a081ecec395c7714dda299b8b03536fe7fb05/timemachine/potentials/jax_utils.py#L11-L19) (n x n), [each pair from two collections](https://github.com/proteneer/timemachine/blob/6c6a081ecec395c7714dda299b8b03536fe7fb05/timemachine/potentials/jax_utils.py#L22-L31) (n x m), and [each pair within cutoff](https://github.com/proteneer/timemachine/blob/6c6a081ecec395c7714dda299b8b03536fe7fb05/timemachine/potentials/jax_utils.py#L126-L157). 

Other bits to note:
* ~[Replaced](https://github.com/proteneer/timemachine/blob/6c6a081ecec395c7714dda299b8b03536fe7fb05/timemachine/potentials/jax_utils.py#L102-L104) a Python for-loop with a Jax.lax fori_loop~ (Reverted this in https://github.com/proteneer/timemachine/pull/453/commits/061647abcb9fd0859b8011bd4d6aa83093afa092 , since incompatible with current Jax version.)
* [Added](https://github.com/proteneer/timemachine/blob/6c6a081ecec395c7714dda299b8b03536fe7fb05/timemachine/potentials/nonbonded.py#L121-L122) an optional `runtime_validate` argument for the current nonbonded_v3, which can be set to False when used with Jax JIT (to avoid concretization type errors from the if-else statements in the validation)
* [Expanded](https://github.com/proteneer/timemachine/blob/6c6a081ecec395c7714dda299b8b03536fe7fb05/timemachine/potentials/jax_utils.py#L36-L71) `convert_to_4d` into two stages, with an eye towards later modification of the first stage

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -135,15 +135,16 @@ def prepare_reference_nonbonded(
         lj_rescale_mask[i][j] = 1 - exc
         lj_rescale_mask[j][i] = 1 - exc
 
-    ref_total_energy = functools.partial(
+    ref_total_energy = jax.jit(functools.partial(
         nonbonded.nonbonded_v3,
         charge_rescale_mask=charge_rescale_mask,
         lj_rescale_mask=lj_rescale_mask,
         beta=beta,
         cutoff=cutoff,
         lambda_plane_idxs=lambda_plane_idxs,
-        lambda_offset_idxs=lambda_offset_idxs
-    )
+        lambda_offset_idxs=lambda_offset_idxs,
+        runtime_validate=False,
+    ))
 
     return ref_total_energy
 
@@ -210,15 +211,16 @@ def prepare_water_system(
         lj_rescale_mask[i][j] = 1 - exc
         lj_rescale_mask[j][i] = 1 - exc
 
-    ref_total_energy = functools.partial(
+    ref_total_energy = jax.jit(functools.partial(
         nonbonded.nonbonded_v3,
         charge_rescale_mask=charge_rescale_mask,
         lj_rescale_mask=lj_rescale_mask,
         beta=beta,
         cutoff=cutoff,
         lambda_plane_idxs=lambda_plane_idxs,
-        lambda_offset_idxs=lambda_offset_idxs
-    )
+        lambda_offset_idxs=lambda_offset_idxs,
+        runtime_validate=False,
+    ))
 
     return params, ref_total_energy, test_potential
 
@@ -271,15 +273,16 @@ def prepare_nb_system(
         lj_rescale_mask[i][j] = 1 - exc
         lj_rescale_mask[j][i] = 1 - exc
 
-    ref_total_energy = functools.partial(
+    ref_total_energy = jax.jit(functools.partial(
         nonbonded.nonbonded_v3,
         charge_rescale_mask=charge_rescale_mask,
         lj_rescale_mask=lj_rescale_mask,
         beta=beta,
         cutoff=cutoff,
         lambda_plane_idxs=lambda_plane_idxs,
-        lambda_offset_idxs=lambda_offset_idxs
-    )
+        lambda_offset_idxs=lambda_offset_idxs,
+        runtime_validate=False
+    ))
 
     return params, ref_total_energy, test_potential
 
--- a/tests/common.py
+++ b/tests/common.py
@@ -136,16 +136,15 @@ def prepare_reference_nonbonded(
         lj_rescale_mask[i][j] = 1 - exc
         lj_rescale_mask[j][i] = 1 - exc
 
-    ref_total_energy = jax.jit(functools.partial(
+    ref_total_energy = functools.partial(
         nonbonded.nonbonded_v3,
         charge_rescale_mask=charge_rescale_mask,
         lj_rescale_mask=lj_rescale_mask,
         beta=beta,
         cutoff=cutoff,
         lambda_plane_idxs=lambda_plane_idxs,
-        lambda_offset_idxs=lambda_offset_idxs,
-        runtime_validate=False,
-    ))
+        lambda_offset_idxs=lambda_offset_idxs
+    )
 
     return ref_total_energy
 
@@ -212,7 +211,7 @@ def prepare_water_system(
         lj_rescale_mask[i][j] = 1 - exc
         lj_rescale_mask[j][i] = 1 - exc
 
-    ref_total_energy = jax.jit(functools.partial(
+    ref_total_energy = functools.partial(
         nonbonded.nonbonded_v3,
         charge_rescale_mask=charge_rescale_mask,
         lj_rescale_mask=lj_rescale_mask,
@@ -221,7 +220,7 @@ def prepare_water_system(
         lambda_plane_idxs=lambda_plane_idxs,
         lambda_offset_idxs=lambda_offset_idxs,
         runtime_validate=False,
-    ))
+    )
 
     return params, ref_total_energy, test_potential
 
@@ -274,7 +273,7 @@ def prepare_nb_system(
         lj_rescale_mask[i][j] = 1 - exc
         lj_rescale_mask[j][i] = 1 - exc
 
-    ref_total_energy = jax.jit(functools.partial(
+    ref_total_energy = functools.partial(
         nonbonded.nonbonded_v3,
         charge_rescale_mask=charge_rescale_mask,
         lj_rescale_mask=lj_rescale_mask,
@@ -283,7 +282,7 @@ def prepare_nb_system(
         lambda_plane_idxs=lambda_plane_idxs,
         lambda_offset_idxs=lambda_offset_idxs,
         runtime_validate=False
-    ))
+    )
 
     return params, ref_total_energy, test_potential
 
--- a/tests/common.py
+++ b/tests/common.py
@@ -1,7 +1,6 @@
 import unittest
 import numpy as np
 import jax
-jax.config.update("jax_enable_x64", True)
 import jax.numpy as jnp
 import functools
 from timemachine.constants import ONE_4PI_EPS0
--- a/tests/common.py
+++ b/tests/common.py
@@ -1,6 +1,7 @@
 import unittest
 import numpy as np
 import jax
+jax.config.update("jax_enable_x64", True)
 import jax.numpy as jnp
 import functools
 from timemachine.constants import ONE_4PI_EPS0
--- a/tests/test_jax_nonbonded.py
+++ b/tests/test_jax_nonbonded.py
@@ -6,7 +6,7 @@ jax.config.update("jax_enable_x64", True)
 import numpy as onp
 from numpy.random import randn, rand, randint
 from jax import numpy as np, value_and_grad, jit
-from timemachine.potentials.nonbonded import nonbonded_v3, nonbonded
+from timemachine.potentials.nonbonded import nonbonded_v3, _nonbonded_v3_clone
 
 
 def generate_random_inputs(n_atoms, dim=3):
@@ -53,8 +53,8 @@ def compare_two_potentials(u_a, u_b, args, differentiate_wrt=(0, 1, 3)):
 
 
 def test_jax_nonbonded():
-    u_a, u_b = nonbonded_v3, nonbonded
+    u_a, u_b = nonbonded_v3, _nonbonded_v3_clone
     args = generate_random_inputs(20, 3)
 
     compare_two_potentials(u_a, u_b, args)
-    compare_two_potentials(u_a, jit(u_b), args)
+    compare_two_potentials(jit(u_a), jit(u_b), args)
--- a/tests/test_jax_nonbonded.py
+++ b/tests/test_jax_nonbonded.py
@@ -37,7 +37,7 @@ def generate_random_inputs(n_atoms: int, dim: int = 3) -> NonbondedArgs:
     params = rand(n_atoms, 3)
     params[:, 0] -= np.mean(params[:, 0])
 
-    box = np.diag(0.5 + 0.5 * rand(3)) * 10 * n_atoms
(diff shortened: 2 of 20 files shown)


tests/common.py
```diff
@@ -1,6 +1,7 @@
 import unittest
 import numpy as np
 import jax
+jax.config.update("jax_enable_x64", True)
```
proteneer, line 4: Did we want to enable this here? Most of the tests have their own enable - and I'd rather for things that get imported to not silently include this

tests/test_jax_nonbonded.py
```diff
@@ -0,0 +1,151 @@
...
+Conf = Params = Box = ChargeMask = LJMask = LambdaPlaneIdxs = LambdaOffsetIdxs = np.array
+Lamb = Beta = Cutoff = Energy = float
+
+nonbonded_args = Conf, Params, Box, Lamb, ChargeMask, LJMask, Beta, Cutoff, LambdaPlaneIdxs, LambdaOffsetIdxs
+NonbondedArgs = Tuple[nonbonded_args]
+NonbondedFxn = Callable[[*nonbonded_args], Energy]
+
+
+def generate_random_inputs(n_atoms: int, dim: int = 3) -> NonbondedArgs:
+    # jittered
```
proteneer, line 26: jittered?

tests/test_jax_nonbonded.py
```diff
@@ -0,0 +1,151 @@
...
+    offsets = onp.array([onp.arange(n_atoms)] * dim).T
```
proteneer, line 27: Whats the need for these offsets in generating the conformation?

tests/test_jax_nonbonded.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def compare_two_potentials(u_a: NonbondedFxn, u_b: NonbondedFxn, args: NonbondedArgs, differentiate_wrt=(0, 1, 3)):
+    """Assert that energies and derivatives w.r.t. request argnums are close"""
+    value_and_grads = partial(value_and_grad, argnums=differentiate_wrt)
+    energy_a, gradients_a = value_and_grads(u_a)(*args)
+    energy_b, gradients_b = value_and_grads(u_b)(*args)
+
+    assert energy_a == energy_b
+    onp.testing.assert_almost_equal(energy_a, energy_b)
```
proteneer, line 65: you probably want to specify the tolerances for these based on the precision of the method, I think the defaults are for float32.

timemachine/potentials/nonbonded.py
```diff
@@ -1,24 +1,50 @@
...
+        params,
+        box,
+        lamb,
+        charge_rescale_mask,
+        lj_rescale_mask,
+        beta,
+        cutoff,
+        lambda_plane_idxs,
+        lambda_offset_idxs,
+        runtime_validate=False,
```
proteneer, line 46: When is this set to True? I noticed that most of the tests have this set to False

tests/test_jax_nonbonded.py
```diff
@@ -0,0 +1,151 @@
...
+    offsets = onp.array([onp.arange(n_atoms)] * dim).T
+    assert (offsets.shape == (n_atoms, dim))
+    for i in range(dim):
+        onp.random.shuffle(offsets[:, i])
+    conf = (0.5 * randn(n_atoms, dim) + offsets)
+    conf += conf.min()
+    params = rand(n_atoms, 3)
+    params[:, 0] -= np.mean(params[:, 0])
+
+    box = np.diag(0.5 + 0.5 * rand(3)) * 10 * n_atoms
```
proteneer, line 36: I would add a test using a box that actually triggers the PBC wrapping behavior. I think this creates a really large box most of the time such that the system is effectively aperiodic.

tests/test_jax_nonbonded.py
```diff
@@ -0,0 +1,151 @@
...
+
+    # make 4th dimension of box large enough so its roughly aperiodic
+    if box is not None:
+        box_4d = np.eye(4) * 1000
+        box_4d = index_update(box_4d, index[:3, :3], box)
+    box = box_4d
+
+    # TODO: break this into more manageable blocks
+    inds_i, inds_j = get_all_pairs_indices(N)
+    # n_interactions = len(inds_i)
```
proteneer, line 101: extra comment?

timemachine/potentials/jax_utils.py
```diff
@@ -36,19 +89,74 @@ def rescale_coordinates(
...
+
+def batched_neighbor_inds(confs, inds_l, inds_r, cutoff, boxes):
+    """Given candidate interacting pairs (inds_l, inds_r),
+        inds_l.shape == n_interactions
+    exclude most pairs whose distances are >= cutoff (neighbor_inds_l, neighbor_inds_r)
+        neighbor_inds_l.shape == (len(confs), max_n_neighbors)
+        where the total number of neighbors returned for each conf in confs is the same
+        max_n_neighbors
+
+    This padding causes some amount of wasted effort, but keeps things nice and fixed-dimensional
```
proteneer, line 134: Can you confirm that (None,) leading dimensions still trigger a separate JIT pass? I thought if the first dim were variadic, the CUDA back-end at least would avoid repeated jitting.

proteneer (commented):
First pass looks good - some minor comments below.

tests/common.py
maxentile: Reverted in https://github.com/proteneer/timemachine/pull/453/commits/59fff43ddf47efafadcd3d3ec78b917a3e8c9b74 .

I missed if there's a way to set the Jax precision in a more fine-grained way.

In the current version of `test_nonbonded.py`, the precision is set to float64 [globally](https://github.com/proteneer/timemachine/blob/74efe28bfe4fe72a995a764bf3afe635b01bfc73/tests/test_nonbonded.py#L9). `compare_forces` sets the precision for the custom_ops implementation but [doesn't appear](https://github.com/proteneer/timemachine/blob/74efe28bfe4fe72a995a764bf3afe635b01bfc73/tests/common.py#L424-L448) to set the precision for the reference Jax implementation.

tests/test_jax_nonbonded.py
maxentile: Clarified in https://github.com/proteneer/timemachine/pull/453/commits/cb49add53a032f0671e82f98ce9de54ebf216217

tests/test_jax_nonbonded.py
maxentile: Clarified in https://github.com/proteneer/timemachine/pull/453/commits/cb49add53a032f0671e82f98ce9de54ebf216217 -- goal is to avoid generating a random configuration where a pair of points is right on top of each other. There are more direct ways to do this

tests/test_jax_nonbonded.py
maxentile: I'm not sure how to set the precision of a Jax function. Could you point me to how to do this? (I thought Jax default precision is set globally)

timemachine/potentials/nonbonded.py
maxentile: Ahh, I changed the default behavior here. Previous behavior corresponds to runtime_validate=True , following https://github.com/proteneer/timemachine/pull/424#discussion_r629678467 . Will revert the default behavior to True

tests/test_jax_nonbonded.py
proteneer: Thanks for clarifying - I think this is fine then.

tests/test_jax_nonbonded.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/453/commits/a28424d2bd60338587d7eedc92bf0573fdb2460a

tests/test_jax_nonbonded.py
maxentile: Consolidated in https://github.com/proteneer/timemachine/pull/453/commits/fcf2a98def1bd877cd8f310a495ecb31decd5099

timemachine/potentials/jax_utils.py
maxentile: Sorry, I'm not sure how to confirm this! Is there an example you can point me to, illustrating where to provide (None,) leading dimensions?

[This comment](https://github.com/google/jax/issues/2521#issuecomment-604759386) on the Jax issue tracker suggested that it does need to be shape-specialized for now, and later comments suggest that the Jax masking approach to emulating variable-length behavior with fixed-length arrays is being reworked.

tests/test_jax_nonbonded.py
proteneer: Hm, so far we've never used the 32-bit version of jax, since we use it mainly as a *reference* platform where speed isn't of the paramount importance. That is, previously, we've always compared 32-bit and 64-bit custom_ops against *only* the 64bit version of jax. I think comparing the 64-bit nonbonded_v3 vs the 64-bit nonbonded_v3_specific pairs is probably sufficient, but you're right that there's no easy way to test the 32-bit versions except to globally enable it in another file.

timemachine/potentials/jax_utils.py
proteneer: You're right - XLA doesn't support None still. Oof.

timemachine/potentials/nonbonded.py
```diff
@@ -119,13 +146,40 @@ def nonbonded_v3(
...
+def nonbonded_v3_on_specific_pairs(conf, params, box, inds_l, inds_r, beta: float, cutoff: Optional[float] = None):
+    """See nonbonded_v3 docstring for more details"""
+
+    # distances and cutoff
+    dij = distance_on_pairs(conf[inds_l], conf[inds_r], box)
+    if cutoff is None:
+        cutoff = np.inf
+    keep_mask = dij <= cutoff
+
+    def apply_cutoff(x):
```
badisa, line 167: nit: Seems like this nested function is unnecessary since it is calling a single function?

maxentile: Personal style preference, but unnecessary. Can revert.

(I need to do `np.where(keep_mask, x, 0)` a few times in the next couple lines, and it seemed less "noisy" to be calling a function with a single argument. I don't think there's any overhead, since Jax JIT will effectively inline this.)

timemachine/potentials/nonbonded.py
maxentile: Addressed in https://github.com/proteneer/timemachine/commit/840c9a09be58b075aab760ce1b71d0d3df33eb5c . Default behavior (True) now matches previous behavior, and non-default behavior (False) remains used in tests that require JIT.

proteneer (approved):
I think this LGTM - we should probably figure out how to do parameter interpolation at some point before this can be fully utilized. I don't think it should be hard.

maxentile:
The latest two test failures look unrelated to the proposed changes (one a segfault in an unrelated test, and one a small violation of an assertion in the barostat pressure jump test). 

>I think this LGTM - we should probably figure out how to do parameter interpolation at some point before this can be fully utilized. I don't think it should be hard.

Agreed, but perhaps best for a separate PR. Nonbonded parameter interpolation happens before params are passed to the Jax function `nonbonded_v3` I think. To support parameter interpolation I think we would just need to increase the number of pairs considered (`n x m` -> `n x m + m x m` if the parameters of `m` particles are being interpolated), and call [the Jax implementation](https://github.com/proteneer/timemachine/blob/8c6846ebd48002530590d5907de23efffa03567c/tests/test_parameter_interpolation.py#L12-L18) of parameter interpolation.

timemachine/potentials/nonbonded.py
```diff
@@ -1,24 +1,50 @@
+import jax
+jax.config.update("jax_enable_x64", True)
```
proteneer, line 2: You probably want to remove this - let the top level code decide whether or not to enable this.

badisa (approved):

timemachine/potentials/nonbonded.py
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/453/commits/0489b746e68c13000963809cdf750c936da1bbb9

PR #455, opened 2021-06-25
Relative Hydration Improvements

proteneer:
This PR makes two major changes:

1) The estimator_rabfe code now returns the BAR error, and the final error is properly accumulated all the way to the final value.
2) For the purposes of correctness testing, and to avoid sampling problems, we turn off the non-ring torsions for *both* the absolute calculation and the relative calculation. This helps improve the overlap of the endpoint correction and removes one source of sampling problems. In binding free energies, we do a similar procedure (except we explicitly compute the deltaG of turning off the torsions). This isn't needed for this particular test case because we're mainly testing for consistency of the restraints themselves.

### DIFF
--- a/examples/validate_relative_hydration.py
+++ b/examples/validate_relative_hydration.py
@@ -152,14 +152,15 @@ if __name__ == "__main__":
                 mol_a = dataset.data[i]
                 mol_b = dataset.data[j]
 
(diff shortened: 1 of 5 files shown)


maxentile (approved):
Looks good!

proteneer:
Thanks for the quick review, merging.

PR #458, opened 2021-06-28
Bugfix/#456 make barostat and centroid deterministic

badisa:


tests/test_barostat.py
```diff
@@ -167,15 +167,12 @@ def test_barostat_varying_pressure():
...
-    start = time.time()
-    volumes = []
-    while time.time() - start < 15:
-        ctxt.multiple_steps(np.ones(1000)*lam)
-        atm_box = ctxt.get_box()
-        volumes.append(compute_box_volume(atm_box))
 
+    ctxt.multiple_steps(np.ones(5000)*lam)
+    atm_box = ctxt.get_box()
+    np.testing.assert_equal(compute_box_volume(atm_box), 7.838565243228307)
```
proteneer, line 173: we probably don't want to make it *this* precise, because if we run the test on a different driver/GPU the numbers will be off. CUB only guarantees local-determinism (w.r.t. GPU/driver combo)

badisa:
Still working on the test, appears that the return value is not consistent across hardware.

timemachine/cpp/src/barostat.cu
```diff
@@ -134,9 +136,9 @@ void __global__ rescale_positions(
 
     const double num_atoms = static_cast<double>(mol_offsets[mol_idx+1] - mol_offsets[mol_idx]);
 
-    const double centroid_x = centroids[mol_idx*3+0] / num_atoms;
-    const double centroid_y = centroids[mol_idx*3+1] / num_atoms;
-    const double centroid_z = centroids[mol_idx*3+2] / num_atoms;
+    const double centroid_x = FIXED_TO_FLOAT<double>(centroids[mol_idx*3+0]) / num_atoms;
```
proteneer, line 139: you probably don't want to use this particular FIXED_TO_FLOAT, it has a precision designed for the nonbonded forces.

tests/test_barostat.py
badisa: Looks like even between T4s, this can differ by 0.4 which is more significant than I had expected...

proteneer: hm... it looks like you're doing atomicAdds actually, I think even with atomicAdds we can only guarantee run-to-run determinism.

timemachine/cpp/src/barostat.cu
```diff
@@ -157,17 +159,17 @@ void __global__ find_group_centroids(
...
     const int idx = blockIdx.x*blockDim.x + threadIdx.x;
     if (idx >= N) {
         return;
     }
     const int atom_idx = atom_idxs[idx];
     const int mol_idx = mol_idxs[idx];
-    atomicAdd(centroids + mol_idx*3+0, coords[atom_idx*3+0]);
-    atomicAdd(centroids + mol_idx*3+1, coords[atom_idx*3+2]);
-    atomicAdd(centroids + mol_idx*3+2, coords[atom_idx*3+2]);
+    atomicAdd(centroids + mol_idx*3+0, FLOAT_TO_FIXED<double>(coords[atom_idx*3+0]));
```
proteneer, line 170: I just realized that this will probably be horrendously slow for the protein - since the group sizes for the centroid is huge. We can address this in a separate PR by doing reduction on this probably.

timemachine/cpp/src/barostat.cu
badisa: Which fixed precision function is preferable? Do we want another level of fixed precision for positions?

tests/test_barostat.py
proteneer: Have you confirmed that its deterministic on the same T4 instance?

badisa: It was deterministic on a T4 and on my RTX 2080, both passing the tests.

tests/test_barostat.py
```diff
@@ -1,3 +1,4 @@
+import platform
```
proteneer, line 1: wow did not know this was a thing

tests/test_barostat.py
```diff
@@ -167,15 +168,22 @@ def test_barostat_varying_pressure():
...
+    platform_version = platform.version().lower()
+    # OpenEye's AM1 Charging values are OS platform dependent. To ensure that we have deterministic values
+    # we check against our two most common OS versions, Ubuntu 18.04 and 20.04.
+    box_vol = 7.594815077349339
+    box_diff = 0.26980723017633856
+    if "ubuntu" not in platform_version:
+        print("Test expected to run under ubuntu 20.04 or 18.04")
+    if "20.04" in platform_version:
+        box_vol = 7.838565243228307
+        box_diff = 0.5460385300302901
```
maxentile, line 183: Request: Could we add an environment-dependent assertion about starting conditions as well? (Starting conditions (namely, the partial charge parameters), but not any of the later simulation logic, are expected to be environment-dependent. This should imply currently tested values, I think, but asserting both would make it faster to track down whether the OS-dependent starting condition assumption or the OS-independent simulation logic assumption change in the future.)

maxentile: Thanks! Resolved in https://github.com/proteneer/timemachine/pull/458/commits/413f8770c1f88521658c9b5ab2b4cd34f6ac8b7a

proteneer (approved):
lgtm!

PR #459, opened 2021-06-28
Langevin improvements

maxentile:
- [x] Modifies the `langevin_coefficients` function to produce coefficients for BAOAB [Leimkuhler, Matthews, 2012] https://arxiv.org/abs/1203.5428
https://github.com/proteneer/timemachine/blob/9a13918a31b00f7c637fde748ef683a3d06f9f1b/timemachine/integrator.py#L44-L46
- [x] Modifies the integrator `update_fwd` function to implement a half-timestep rotation of BAOAB https://github.com/proteneer/timemachine/blob/9a13918a31b00f7c637fde748ef683a3d06f9f1b/timemachine/cpp/src/integrator.cu#L71-L74
- [x] Removes unused `brownian_coefficients` function (which can now safely be emulated by passing `friction=np.inf` to the `langevin_coefficients` function)

-------

Motivation: BAOAB makes less timestep error than the currently implemented Langevin integrator, and is more robust to changing the `friction`. (At the extreme, note that the current integrator becomes completely "Hamiltonian-independent" in the limit of friction to +inf. It also makes more configuration-space error than BAOAB as a function of step size `dt`. Robustness of the current and proposed integrators with respect to changing `dt` and `friction` are compared on a 1D example in this notebook: https://gist.github.com/maxentile/3a8cefe97a8ed14366105bb25d24ef22 .)


tests/test_md.py
```diff
@@ -91,8 +81,13 @@ def integrate_once_through(
...
                 all_du_dxs.append(du_dx)
                 all_xs.append(x_t)
-                v_t = ca*v_t + np.expand_dims(cbs, axis=-1)*du_dx
-                x_t = x_t + v_t*dt
+
+                noise = np.random.randn(*v_t.shape)
+
+                v_mid = v_t + np.expand_dims(cbs, axis=-1) * du_dx
+
+                v_t = ca * v_mid + np.expand_dims(ccs, axis=-1) * noise
```
proteneer, line 89: note that I wasn't sure how to test providing an artificial set of rngs.... maybe we can call into cuRAND and give it the same seed? (XORWOW in particular)

proteneer:
Looks great! just a minor comment re: testing, but can't wait to try this out once tests pass.

tests/test_md.py
maxentile: Ahh, good call.

I don't immediately see a convenient way to emulate cuRAND from numpy, to construct a test for identical trajectories. (For example, each thread gets its own seed https://docs.nvidia.com/cuda/curand/host-api-overview.html#order , so the emulator would have to be at least that detailed... If we find we need to do something like this elsewhere, perhaps the CUDA simulator in Numba may be handy? https://numba.pydata.org/numba-doc/dev/cuda/simulator.html )

Perhaps it would be good to add a short test that we can use MD to estimate known ensemble averages to within some tolerance, for a couple choices of timestep and collision rate.

For now I've reverted the test to not use random numbers in https://github.com/proteneer/timemachine/pull/459/commits/17cde2ab069cdeba1254ca945a5e86f76228c94e .

proteneer: https://pyculib.readthedocs.io/en/latest/curand.html

Might work if you use XORWOW, since its a fully deterministic algorithm (should be easy to try). But I'm also okay with just removing the noise for the test.

proteneer (approved):

badisa (approved):

tests/test_md.py
maxentile: Thanks!

Will leave as-is for now, but later I plan to revisit how to test for accurate sampling.

PR #460, opened 2021-06-30
Add barostat to benchmarks

badisa:
Adds about 20 seconds to the test variants of the benchmarks.

Points out that the barostat is really slow, having about a 15% performance impact at an interval of 25 and 40% at an interval of 5.

Will need to speed up the barostat in the near future.

### DIFF
--- a/slow_tests/test_benchmark.py
+++ b/slow_tests/test_benchmark.py
@@ -11,14 +11,14 @@ from ff import Forcefield
 
 from simtk.openmm import app
 
-from timemachine.lib import custom_ops
-from timemachine.lib import LangevinIntegrator
+from timemachine.lib import custom_ops, LangevinIntegrator, MonteCarloBarostat
 
 from fe.utils import to_md_units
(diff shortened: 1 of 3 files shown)


maxentile (approved):

slow_tests/test_benchmark.py
```diff
@@ -47,7 +47,9 @@ def benchmark(
     num_batches=100,
     steps_per_batch=1000,
     compute_du_dp_interval=100,
-    compute_du_dl_interval=0
+    compute_du_dl_interval=0,
+    barostat=False,
+    barostat_interval=25,
```
proteneer, line 52: 1) did we want to use barostat_interval = 0 as a signal for False?
2) what happens if we actually passed in a barostat interval of 0 to the custom_ops class, does actually do a mod zero of something?

badisa: 1) That would work for me

2) It faults, will fix it up to raise a more meaningful error

PR #465, opened 2021-07-12
print MBAR values

proteneer:
Not sure if we want to merge this - but mainly to showcase using MBAR instead of BAR for the final energy estimate.


proteneer:
Closing this in favor of #476

PR #466, opened 2021-07-14
Task/optimize barostat

badisa:
Fundamentally this does not speed things up that much. In the case where the interval is 1, you see a ~40% improvement in speed (DHFR 54 ns/day -> 70ns/day), however as the interval gets up to 25 it falls off to being negligible (193ns/day -> 195, and that is not a meaningful difference). 

The issue is that *every* barostat move triggers up to 2 rebuilds of the neighborlist. The first when you shift the box and if that shift failed, it has to recompute the original neighborlist. At this point the most obvious solution is to speed up the neighborlist construction, though that seems to be a separate issue. Another idea would be to give the Barostat a different set of potentials that could reduce the number of neighborlist reconstructions are needed. This would happen because as the simulation runs longer the number of accepted moves decreases, which means the original neighborlist can be reused more regularly.

### DIFF
--- a/timemachine/cpp/src/barostat.cu
+++ b/timemachine/cpp/src/barostat.cu
@@ -4,6 +4,7 @@
 #include <algorithm>
 #include <stdio.h>
 #include <set>
+#include <cub/cub.cuh>
 
 #include "kernels/k_fixed_point.cuh"
 
@@ -14,7 +15,7 @@ namespace timemachine {
 
 MonteCarloBarostat::MonteCarloBarostat(
     const int N,
-    const double pressure,  // Expected in Bar
+    const double pressure, // Expected in Bar
     const double temperature, // Kelvin
     const std::vector<std::vector<int> > group_idxs,
     const int interval,
@@ -27,9 +28,8 @@ MonteCarloBarostat::MonteCarloBarostat(
     bps_(bps),
     group_idxs_(group_idxs),
     num_grouped_atoms_(0),
-    volume_scale_(0),
-    num_attempted_(0),
-    num_accepted_(0),
+    d_sum_storage_(nullptr),
+    d_sum_storage_bytes_(0),
     seed_(seed),
     step_(0) {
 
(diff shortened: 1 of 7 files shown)


timemachine/cpp/src/context.cu
```diff
@@ -161,16 +164,13 @@ void Context::_step(double lambda, unsigned long long *du_dl_out) {
...
 
-    gpuErrchk(cudaMemset(d_du_dx_t_, 0, N_*3*sizeof(*d_du_dx_t_)));
+    gpuErrchk(cudaMemsetAsync(d_du_dx_t_, 0, N_*3*sizeof(*d_du_dx_t_), stream));
 
     if(du_dl_out) {
-        gpuErrchk(cudaMemset(d_du_dl_buffer_, 0, N_*sizeof(*d_du_dl_buffer_)));
+        gpuErrchk(cudaMemsetAsync(d_du_dl_buffer_, 0, N_*sizeof(*d_du_dl_buffer_), stream));
     }
 
-    auto start = std::chrono::high_resolution_clock::now();
```
badisa, line 170: Was unused

timemachine/cpp/src/context.cu
```diff
@@ -151,6 +152,8 @@ void Context::_step(double lambda, unsigned long long *du_dl_out) {
     // the observables decide on whether or not to act on given
     // data (cheap pointers in any case)
 
+    cudaStream_t stream = static_cast<cudaStream_t>(0);
```
badisa, line 159: Seemed cleaner to reuse the stream through out a step, to enable async all the way through

timemachine/cpp/src/context.cu
```diff
@@ -179,7 +188,7 @@ void Context::_step(double lambda, unsigned long long *du_dl_out) {
             d_du_dx_t_, // we only need the forces
             nullptr,
             du_dl_out ? d_du_dl_buffer_ : nullptr,
-            nullptr,
+            compute_u ? d_u_buffer_ : nullptr,
```
badisa, line 191: If the barostat will attempt a move, compute the energy immediately and pass it to the barostat.

badisa:
Tests are failing on the test runner with nans, likely something is fundamentally off. Would appreciate any feedback on the general approach while I investigate.

timemachine/cpp/src/barostat.cu
```diff
@@ -172,76 +198,170 @@ void __global__ find_group_centroids(
...
+        return; // Only a single thread needs to perform this operation
+    }
+    const double volume = d_box[0*3+0] * d_box[1*3+1] * d_box[2*3+2];
+    if(d_volume_scale[0] == 0) {
+        d_volume_scale[0] = 0.1 * volume;
+    }
+    const double delta_volume = d_volume_scale[0]*2*(volume_scaling-0.5);
+    const double new_volume = volume+delta_volume;
+    d_volume_delta[0] = delta_volume;
+    d_length_scale[0] = pow(new_volume/volume, 1.0/3.0);
```
proteneer, line 219: cbrt()

timemachine/cpp/src/barostat.cu
```diff
@@ -251,66 +371,59 @@ void MonteCarloBarostat::inplace_move(
...
-    for (int i = 0; i < N_; i++) {
-        u_init_agg += initial_energy[i];
-        u_final_agg += final_energy[i];
-    }
-
-    double u_init = FIXED_TO_FLOAT<double>(u_init_agg);
-    double u_final = FIXED_TO_FLOAT<double>(u_final_agg);
-
-    double w = u_final-u_init + pressure*delta_volume - num_molecules*kT*std::log(new_volume/volume);
+    const double metropolis_condition = dist_(mt_);
```
proteneer, line 402: we probably want to also do this on the GPU too

badisa:
Has some fundamental issues, going to close until it is fixed up.

PR #467, opened 2021-07-14
Improve Conversion Protocol

proteneer:
This PR turns off the non-ring torsions within the core region only. Previously we turned off all non-ring torsions in the conversion stage, regardless of whether or not it was part of the core. We noticed this introduces more error than we'd like. This PR also requires the user to pre-specify a SMARTS pattern to define the core (and checks to make sure that it's connected).

### DIFF
--- a/fe/topology.py
+++ b/fe/topology.py
@@ -342,23 +342,23 @@ class BaseTopologyConversion(BaseTopology):
     FastFindRings or SSSRs (ring basis), of which neither is what we want.
     """
 
-    def parameterize_proper_torsion(self, ff_params):
-        # alchemically turn off proper torsions.
-        torsion_params, torsion_potential = super().parameterize_proper_torsion(ff_params)
-        membership = get_ring_membership(self.mol)
+    # def parameterize_proper_torsion(self, ff_params):
+    #     # alchemically turn off proper torsions.
+    #     torsion_params, torsion_potential = super().parameterize_proper_torsion(ff_params)
+    #     membership = get_ring_membership(self.mol)
 
-        num_torsions = torsion_params.shape[0]
+    #     num_torsions = torsion_params.shape[0]
 
-        lambda_mult_idxs = np.zeros(num_torsions, dtype=np.int32)
-        lambda_offset_idxs = np.ones(num_torsions, dtype=np.int32)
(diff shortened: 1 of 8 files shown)


fe/topology.py
```diff
@@ -690,23 +697,38 @@ class DualTopologyStandardDecoupling(DualTopology):
...
+        NA = self.mol_a.GetNumAtoms()
+
+        for torsion_idx, (i, b, c, j) in enumerate(torsion_potential.get_idxs()):
+
+            # does this torsion belong to mol_a?
+            if b < NA and c < NA and i < NA and j < NA:
+                if (membership_a[b] != membership_a[c]) and (b in core_idxs_a and c in core_idxs_a):
+                    lambda_offset_idxs[torsion_idx] = 0
+            # are we in in b?
+            if b > NA and c > NA and i > NA and j > NA:
```
badisa: nit: elif

badisa: Should this be `b >= NA`? If I understand this, shouldn't the first atom of Mol B be `idx == NA`?

proteneer: I think you were looking at a stale/outdated page - this was updated last night

badisa: You are right, looks good now!

fe/topology.py
```diff
@@ -690,23 +697,38 @@ class DualTopologyStandardDecoupling(DualTopology):
...
+                if (membership_b[b] != membership_b[c]) and (b in core_idxs_b and c in core_idxs_b):
+                    lambda_offset_idxs[torsion_idx] = 0
+            # we dun goofed, torsion spans two different mols
+            else:
+                assert 0
```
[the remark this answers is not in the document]
proteneer: This is just such a bad situation that I don't want to handle the exception ever. It should just terminate hard. (Usually I only use exceptions if I'd actually want to somehow recover), `assert 0` usually indicates something really bad.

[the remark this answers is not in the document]
badisa: I just meant something like `assert 0, "Torsion spans two different molecules"`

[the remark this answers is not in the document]
proteneer: wilfix

proteneer:
Closing this in favor of #473

PR #474, opened 2021-08-02
Improve parameter flattening

maxentile:
A basic utility for parameter optimization is to "flatten" a structured collection of parameter arrays into a single vector, and later "unflatten" a vector to recover the same structure. Optimization libraries like scipy.optimize usually assume they're talking to a function that accepts a flat parameter vector as input.

It would be good to avoid hard-coding a simple optimizer's logic in terms of elementwise operations applied in-place to a parameter structure we do not flatten, and to avoid hard-coding details of flattening / unflattening a specific parameter structure.

This PR:
- [x] Adds [a utility function](https://github.com/proteneer/timemachine/blob/5b668ffa0848a77f2558b8fceaf39bf87802f288/optimize/utils.py#L6-L41) for flattening and unflattening a generic structure of parameters, using Jax's pytree utilities.
- [x] Adds [two tests](https://github.com/proteneer/timemachine/blob/5b668ffa0848a77f2558b8fceaf39bf87802f288/tests/test_optimize.py#L18-L63) checking for correct flattening/unflattening round-trip behavior, when applied either to a nested structure or to an `ordered_params` structure
- [x] Removes two duplicates of a hardcoded `flatten(params, handles)` function from examples/


tests/test_optimize.py
```diff
@@ -1,13 +1,68 @@
...
+
+    example_pytree = dict(
+        Bonds=onp.random.randn(50, 2),
+        Angles=onp.random.randn(60, 3),
+        Extras=dict(
+            GlobalVars=[
+                jnp.arange(70),
+                onp.random.randn(80, 4)
+            ]
+        )
```
proteneer, line 46: Possible to also test a scalar?
eg. Bias=1.2, Multiplier=2.5

optimize/utils.py
```diff
@@ -0,0 +1,41 @@
...
+        f_flat(x) where x is a flat Jax array
+
+    See also:
+    * HIPS autograd.misc.flatten
+        https://github.com/HIPS/autograd/blob/c6f630a5ec18bd30f1485bc0dbbccb8664c77510/autograd/misc/flatten.py#L9-L15
+        which has similar intended usage but a slightly different signature:
+        flatten(value) -> flat_value, unflatten_fxn
+    """
+
+    tree_structure = tree_util.tree_structure(example_tree)
```
proteneer, line 21: nice usage of tree_util!

proteneer:
This LGTM for now - might want to consider adding support for adding per-type learning rates later on? (based on the maximum change in energy or something).

proteneer (approved):

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -447,33 +410,31 @@ def equilibrate_edges(datasets: List[Dataset], systems: List[Dict[str, Any]], nu
...
 
             # based on current estimate of (loss, grad, and simulation stability), return a conservative step to take in parameter space
             theta_increment = truncated_step(flat_theta, loss, flat_loss_grad,
                                              step_lower_bound=_compute_step_lower_bound(loss, blown_up))
-            param_increments = unflatten_theta(theta_increment)
+            param_increments = unflatten(theta_increment)
 
             # for any parameter handler types being updated, update in place
-            for handle in ordered_handles:
+            for (handle, increment) in zip(ordered_handles, param_increments):
```
badisa, line 422: nit: the `()` around handle and increment are unncessary

optimize/utils.py
```diff
@@ -0,0 +1,41 @@
+from jax import tree_util, numpy as np
+import numpy as onp
+from typing import Tuple, Callable
+
+
+def flatten_and_unflatten(example_tree) -> Tuple[Callable, Callable]:
```
badisa, line 6: nit: example_tree makes me think this is an example rather than production code

badisa (approved):

tests/test_optimize.py
maxentile: This assumes arrays -- would `Bias=np.array([1.2])` be okay to add for this test or should I modify flatten_and_unflatten to handle float-valued as well as array-valued leaves?

optimize/utils.py
maxentile: Do you have a preferred name for this argument? I chose `example_tree` after rejecting:
* `tree`, since that introduces a name collision with definitions within the function
* `template_tree`, since that suggests the argument is more generic or more template-y, rather than "just an example of what you're trying to flatten"
* `tree_structure`, since that has a specific meaning in jax.tree_util

Perhaps `tree_instance` would be preferred?

examples/hif2a/fit_to_multiple_rbfes.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/474/commits/98a5653f4a6584744dd136abd196e4c32b5ed386

optimize/utils.py
badisa: Perhaps `input_tree` or `base_tree`?

maxentile: Addressed in https://github.com/proteneer/timemachine/pull/474/commits/79586b0b247cdbfbb133a073bc23f817f4222db8

tests/test_optimize.py
maxentile: Added length-one arrays to this test in https://github.com/proteneer/timemachine/pull/474/commits/97ec17aed72058f08831e7183269d03267662860 .

Modifying flatten_and_unflatten to handle float-valued as well as array-valued leaves shortly...

maxentile: Float-valued leaves handled in https://github.com/proteneer/timemachine/pull/474/commits/9d258e34c667640f7afc3fed38d52aca24c0a6bf

PR #441, opened 2021-06-02
Increase number of argument names included in Python bindings

maxentile:
Increases informativeness of `help(Potential.execute_selective)` (and a couple others), following https://pybind11.readthedocs.io/en/stable/basics.html#keyword-args .

Consistent with how argument names have been supplied elsewhere in `wrap_kernels.cpp`, such as https://github.com/proteneer/timemachine/blob/be9db7961ab8d0007740ebc7bb764c519f3f26c5/timemachine/cpp/src/wrap_kernels.cpp#L144

-------------
Before:
```
Help on instancemethod in module timemachine.lib.custom_ops:

execute_selective(...)
    execute_selective(self: timemachine.lib.custom_ops.Potential, arg0: numpy.ndarray[numpy.float64], arg1: numpy.ndarray[numpy.float64], arg2: numpy.ndarray[numpy.float64], arg3: float, arg4: bool, arg5: bool, arg6: bool, arg7: bool) -> tuple
```

After:
```
Help on instancemethod in module timemachine.lib.custom_ops:

execute_selective(...)
    execute_selective(self: timemachine.lib.custom_ops.Potential, coords: numpy.ndarray[numpy.float64], params: numpy.ndarray[numpy.float64], box: numpy.ndarray[numpy.float64], lam: float, compute_du_dx: bool, compute_du_dp: bool, compute_du_dl: bool, compute_u: bool) -> tuple
```

PR #475, opened 2021-08-03
Fast delta_U computation

proteneer:
This PR implements the C++ side of computing Us (U terms that depend on lambda) efficiently. It follows roughly that of multiple_steps. However, it has a alot of nuances that make it rather difficult to merge into existing multiple_branches path. In particular, this PR enables one to quickly compute energies across a fixed number of lambda windows, and will be sped up *drastically* soon with @badisa's upcoming PR.


timemachine/cpp/src/context.hpp
```diff
@@ -57,6 +72,7 @@ class Context {
 
     unsigned long long *d_du_dx_t_; // du/dx [N,3]
     unsigned long long *d_du_dl_buffer_; // du/dl [N]
+    unsigned long long *d_u_buffer_; // du/dl [N]
```
maxentile, line 75: line comment may be out of sync

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -180,6 +180,61 @@ void declare_context(py::module &m) {
...
+    .def("multiple_steps_delta_U", [](timemachine::Context &ctxt,
+        const double lambda,
+        const int n_steps,
+        const py::array_t<double, py::array::c_style> &lambda_windows,
+        int store_u_interval,
+        int store_x_interval) -> py::tuple {
+        // (ytz): Collect energies across multiple snapshots.
+
+        // Let lambda_windows have shape [K].
+        // The return'd U matrix has F rows and K states. Where F == ceil(n_steps/store_u_interval).
```
maxentile, line 192:
```suggestion
        // The returned U matrix has F rows and K columns. Where F == ceil(n_steps/store_u_interval).
```

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -180,6 +180,61 @@ void declare_context(py::module &m) {
...
+        // We are guaranteed that for any row U[i, :], the energies are correct up to a constant.
+        // Terms that depend on lambda will always be computed. For example, we may choose to omit
+        // the bonded energies that are independent of lambda from this calculation.
+
+        // Currently however, the energies just so happen to be exact, but this is not
+        // strictly guaranteed. So you should avoid comparing energies down the columns U[:, i].
```
maxentile, line 198: These usage comments are very important -- probably will save some missteps if these are made more visible to the Python layer. Would it be okay to switch these from line comments (visible only when inspecting `wrap_kernels.cpp` source) to a docstring (visible interactively via `help(ctxt.multiple_steps_delta_U)`)?

I think this can be done by moving these comments into a string, and then passing that string as the third argument to `.def` https://pybind11.readthedocs.io/en/stable/advanced/misc.html#generating-documentation-using-sphinx

timemachine/cpp/src/context.cu
```diff
@@ -140,6 +142,163 @@ std::array<std::vector<double>, 3> Context::multiple_steps(
...
+                        stream
+                    );
+                    gpuErrchk(cudaPeekAtLastError());
+                }
+
+            }
+
+            // for(int i=0; i < streams_.size(); i++) {
+                // gpuErrchk(cudaStreamSynchronize(streams_[i]));
+            // }
```
maxentile, line 251: remove?

timemachine/cpp/src/context.hpp
proteneer: willfix

timemachine/cpp/src/wrap_kernels.cpp
proteneer: It's not clear to me if

1) We need to depend on sphinx explicitly in the build step actually generate the docstrings
2) How help() can look up the generated docstrings
3) What we'd really like is something that pylint can read into (as opposed to something that's only usable interactively). I tested the approach above and it doesn't seem to trigger.

PS I don't mind adding sphinx etc. but we should actually verify that it would actually does what we want to do, namely 3).

badisa (approved):

timemachine/cpp/src/wrap_kernels.cpp
maxentile: Ahh, I misunderstood -- will have to look further into how to supply docstrings to pybind.

Re: 1, 2 -- Sorry, didn't mean to suggest adding sphinx dependency, just this was the first search result I encountered when looking for a page in the pybind documentation that shows how to supply a docstring. Supplying a string as the third argument to `.def` and having it accessible via `help(...)` does appear to work in other parts of the pybind documentation e.g. https://pybind11.readthedocs.io/en/stable/basics.html#keyword-arguments , and we had [previously](https://github.com/proteneer/timemachine/pull/441) added annotations to be visible to `help(...)`

Re: 3 -- Hadn't considered pylint compatibility, but that would be helpful for whichever documentation convention we adopt here.

Perhaps "documentation of Python bindings" should be moved into a new issue to avoid derailing the current PR?

Main point I wanted to ensure is visible to the Python layer is that there are per-row offsets. Some applications could be sensitive or insensitive to these offsets, so documenting this somehow to the Python user seems important.

An alternative to a docstring might be to add a keyword argument with a descriptive name for this offset behavior, and have it only accept `True` for now.

maxentile: On second thought, can this offset behavior be simplified / removed altogether?

We want to return `delta_U(x, target_lam) = U(x, target_lam) - U(x, reference_lam)`.

The comment suggests we are instead possibly returning `offset_delta_U(x, target_lam) = delta_U(x, target_lam) + offset(x)`.

Could we instead return `offset_delta_U(x, target_lam) - offset_delta_U(x, reference_lam)` -- which simplifies to `delta_U(x, target_lam)` -- so there is no longer an unspecified offset?

maxentile (approved):
Looks good! Should accelerate the task of computing the nested loop `[[U(x, lam) - U(x, reference_lam) for lam in lambdas] for x in trajectory]`, as required for MBAR -- with further acceleration forthcoming. This could also accelerate the task of [optimizing the spacing of intermediates in nonequilibrium switching](https://github.com/proteneer/timemachine/pull/442), which could use `[U(x, lam) - U(x, reference_lam) for lam in trial_lambdas]` (where `x` is the last frame of the current trajectory, and we want to select one of `trial_lambdas` to simulate in the next trajectory).

timemachine/cpp/src/wrap_kernels.cpp
proteneer: > Perhaps "documentation of Python bindings" should be moved into a new issue to avoid derailing the current PR?

Ok - let's settle then on simply having the docstring be accessible via help() for now, and look for a more robust solution (and editor friendly) one in a separate issue/PR.

> Some applications could be sensitive or insensitive to these offsets.

Applications that depend on the total energy probably aren't free energy methods :)

> We want to return `delta_U(x, target_lam) = U(x, target_lam) - U(x, reference_lam)`.

I see that this is bad naming, this should probably be renamed to "offset_U" for clarity.

> `The comment suggests we are instead possibly returning offset_delta_U(x, target_lam) = delta_U(x, target_lam) + offset(x).`

This is correct, we're really just returning energies shifted by a constant amount across the rows.

> Could we instead return offset_delta_U(x, target_lam) - offset_delta_U(x, reference_lam)

Wouldn't returning `offset_delta_U(x, target_lam) - offset_delta_U(x, reference_lam)` require us to return a NxN matrix of energies for *each* frame x against all `target`, `reference` pair of lamba window combinations. Are you suggesting that we explicitly fix the `reference` lambda window  to that of the window used to run the MD simulation (`lambda` argument as opposed to `lambda_windows`)? Note that the reference lambda window *currently* isn't guaranteed to be an element of the "lambda_windows" we evaluate the energies at, but we can change this.

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -180,6 +180,61 @@ void declare_context(py::module &m) {
         ctxt.get_x_t(buffer.mutable_data());
         return buffer;
     })
+    .def("multiple_steps_delta_U", [](timemachine::Context &ctxt,
+        const double lambda,
```
proteneer, line 184: @maxentile  wasn't sure if this was what you meant by "reference"?

timemachine/cpp/src/wrap_kernels.cpp
proteneer: Not sure if we can pick an arbitrary `reference_lambda` here... since we want the delta_Us returned to be the same independent of the choice of `reference_lambda` (at least if they were to be used in MBAR/inf-switch).

`U` evaluated is clearly dependent on `lambda`: `U(x,lamb) = U_independent(x) + U_dependent(x, lamb)`. If we were to evaluate at two different values of `lambda`:

`U(x, lamb_ref) = U_independent(x) + U_dependent(x, lamb_ref)`
`U(x, lamb_target) = U_independent(x) + U_dependent(x, lamb_target)`
`delta_U = U_dependent(x, lamb_target) - U_dependent(x, lamb_ref)`

The exact value of the delta_Us we get back out would then be dependent on `lamb_ref`. The offsetting approach only shifts the `U_independent` component by a fixed constant, and is actually what we want.

maxentile: >Wouldn't returning offset_delta_U(x, target_lam) - offset_delta_U(x, reference_lam) require us to return a NxN matrix of energies for each frame x against all target, reference pair of lamba window combinations. 

No -- sorry for unclarity. There is one `reference_lam` (in the `lambda` argument) and multiple `target_lam`s (in the `lambdas` argument).

>Are you suggesting that we explicitly fix the reference lambda window to that of the window used to run the MD simulation (lambda argument as opposed to lambda_windows)?

Yes!

This wouldn't cost any generality, unless I'm missing something? This would just replace an unspecified behavior (how much is each row offset by) with a specified behavior (each row is offset by 0).

>I see that this is bad naming, this should probably be renamed to "offset_U" for clarity.

That might be better!

Summary of improvement options:
* Change definition to return `delta_U`
* Change name or add other user-facing hint that `offset_U` is returned

maxentile: Resolved in https://github.com/proteneer/timemachine/pull/475/commits/7c07f1846a8033ba8b354eb606cb6dbe309639ac by changing function name from `multiple_steps_delta_U` to `multiple_steps_U` and expanding comments into a docstring

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -180,6 +180,84 @@ void declare_context(py::module &m) {
...
+
+        store_x_interval: int
+            How often we store the frames
+
+        Returns
+        -------
+        3-tuple of energies, coordinates, boxes
+            Energies have shape (ceil(n_steps/store_u_interval), K)
+            Coordinates have shape (ceil(n_steps/store_x_interval), N, 3)
+            Boxes have shape (ceil(n_steps/store_x_interval), 3, 3)
```
maxentile, line 251: Nit:
```suggestion
            Energies have shape (F, K)
            Coordinates have shape (F, N, 3)
            Boxes have shape (F, 3, 3)
```
since `F = ceil(n_steps/store_u_interval)` is introduced and used earlier in this docstring

PR #479, opened 2021-08-05
Feature/energy of subset of system#472

badisa:


badisa:
This needs to be revisited after the reversion, closing to avoid load on CI

PR #480, opened 2021-08-05
Truncate residuals for "bottomed-out" labels

maxentile:
Ad hoc patch for comparing absolute predictions to absolute labels, if the labels are only considered reliable in some specified interval (aka fitting to "bottomed-out" assays, as in https://github.com/MCompChem/fep-benchmark/issues/6 ).

Where we compute `loss(predictions - labels)`, we can instead compute `loss(truncated_residuals(predictions, labels))`:
* If the label is in the reliable interval, compute the residual `predictions - labels` normally.
* If the label is outside the interval, truncate the residual to reflect that the label is only providing a `true_value <= some_threshold` signal, rather than a `true_value == label` signal.

Examples of this behavior: https://github.com/proteneer/timemachine/blob/1fccf1636a65489549f1815622918f5b73fe8d5e/fe/loss.py#L105-L109

Further elaborations might:
* add a similar patch, but for relative predictions -- `truncated_relative_residuals(rel_preds_ij, abs_labels_i, abs_labels_j)`
* consider using the `- log_likelihood(predictions, labels)` of an explicit model

### DIFF
--- a/tests/test_loss.py
+++ b/tests/test_loss.py
@@ -1,5 +1,5 @@
 import numpy as np
-from fe.loss import l1_loss, pseudo_huber_loss, flat_bottom_loss
+from fe.loss import l1_loss, pseudo_huber_loss, flat_bottom_loss, truncated_residuals
 
 
 def _assert_nonnegative_loss(loss_fxn: callable):
@@ -27,3 +27,35 @@ def test_pseudo_huber_loss():
 
 def test_flat_bottom_loss():
     _assert_basic_loss_properties(flat_bottom_loss)
+
+
+def _scalar_truncated_residual(prediction, label, reliable_interval):
+    lower, upper = reliable_interval
+    assert lower < upper
+
+    if (label >= lower) and (label <= upper):
+        residual = prediction - label
+    elif label < lower:
+        residual = max(0, prediction - lower)
+    elif label > upper:
+        residual = min(0, prediction - upper)
+    else:
+        raise (RuntimeError('unsatisfiable reliable_range'))
+
+    return residual
+
+
+def test_compute_residuals():
+    predictions = np.random.rand(10000) * 4 - 2
+    labels = np.random.rand(10000) * 4 - 2
+
+    reliable_interval = (-1, +1)
+
+    ref = np.array([
+        _scalar_truncated_residual(p, l, reliable_interval)
+        for (p, l) in zip(predictions, labels)
+    ])
+
+    test = truncated_residuals(predictions, labels, reliable_interval)
+
+    np.testing.assert_allclose(ref, test)
--- a/tests/test_loss.py
+++ b/tests/test_loss.py
@@ -45,12 +45,7 @@ def _scalar_truncated_residual(prediction, label, reliable_interval):
     return residual
 
 
-def test_compute_residuals():
-    predictions = np.random.rand(10000) * 4 - 2
-    labels = np.random.rand(10000) * 4 - 2
-
-    reliable_interval = (-1, +1)
-
+def assert_consistency_with_scalar_version(predictions, labels, reliable_interval):
     ref = np.array([
         _scalar_truncated_residual(p, l, reliable_interval)
         for (p, l) in zip(predictions, labels)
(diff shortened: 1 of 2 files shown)


badisa (approved):

proteneer:
did the CI worker OOM or something?

````
[10.78.114.37] out: tests/test_nonbonded.py::TestNonbondedDHFR::test_correctness ========= Error: process didn't terminate successfully
```

tests/test_loss.py
```diff
@@ -27,3 +27,35 @@ def test_pseudo_huber_loss():
...
+    labels = np.random.rand(10000) * 4 - 2
+
+    reliable_interval = (-1, +1)
+
+    ref = np.array([
+        _scalar_truncated_residual(p, l, reliable_interval)
+        for (p, l) in zip(predictions, labels)
+    ])
+
+    test = truncated_residuals(predictions, labels, reliable_interval)
```
proteneer, line 59: possible to also add a test for [-np.inf, +np.inf] interval?

tests/test_loss.py
```diff
@@ -27,3 +27,35 @@ def test_pseudo_huber_loss():
...
+    assert lower < upper
+
+    if (label >= lower) and (label <= upper):
+        residual = prediction - label
+    elif label < lower:
+        residual = max(0, prediction - lower)
+    elif label > upper:
+        residual = min(0, prediction - upper)
+    else:
+        raise (RuntimeError('unsatisfiable reliable_range'))
```
proteneer, line 43: When would this assertion be raised?

badisa:
> did the CI worker OOM or something?
> 
> ```
> [10.78.114.37] out: tests/test_nonbonded.py::TestNonbondedDHFR::test_correctness ========= Error: process didn't terminate successfully
> ```

Confirmed that the CI worker OOM'd, no obvious reason....

tests/test_loss.py
proteneer: sorry just realized this is in the test file! don't really care then

proteneer:
Can you clarify this particular test case for me:

``` python

def test_compute_residuals_sanity_check():
    reliable_interval = (-10, +np.inf)
    predictions = np.array([-100.0, 0.0])
    
    # assays bottom'd out
    labels = np.array([-15, -15])

    # if we predict -100, but our label of -15 bottomed out at -10, residual should be 0
    # if we predict    0, and our label of -15 bottomed out at -10, residual should be >=10
    expected_residual = np.array([0, 10])
    test_residual = truncated_residuals(predictions, labels, expected_residual)
    np.testing.assert_array_equal(expected_residual, test_residual) # fails
```

maxentile:
I think this might be a typo in the test case, as the defined `reliable_interval` is unused.

Replacing
``` python

def test_compute_residuals_sanity_check():
    reliable_interval = (-10, +np.inf)
    predictions = np.array([-100.0, 0.0])
    
    # assays bottom'd out
    labels = np.array([-15, -15])

    # if we predict -100, but our label of -15 bottomed out at -10, residual should be 0
    # if we predict    0, and our label of -15 bottomed out at -10, residual should be >=10
    expected_residual = np.array([0, 10])
    test_residual = truncated_residuals(predictions, labels, expected_residual)
    np.testing.assert_array_equal(expected_residual, test_residual) # fails, test_residual truncates to [0,0]
```


with the likely intended
``` python

def test_compute_residuals_sanity_check():
    reliable_interval = (-10, +np.inf)
    predictions = np.array([-100.0, 0.0])
    
    # assays bottom'd out
    labels = np.array([-15, -15])

    # if we predict -100, but our label of -15 bottomed out at -10, residual should be 0
    # if we predict    0, and our label of -15 bottomed out at -10, residual should be >=10
    expected_residual = np.array([0, 10])
    test_residual = truncated_residuals(predictions, labels, reliable_interval)
    np.testing.assert_array_equal(expected_residual, test_residual) # passes
```

causes the assertion to pass

tests/test_loss.py
maxentile: Didn't have a specific error in mind, just being conservative -- this line might be unreachable, and I can remove it.

(Wanted to hear something in case we later pass some combination of nans, infs, +infs, and reals in `label`, `lower`, `upper` that doesn't get caught by the if statements.)

proteneer:
Oops! You're 100% right!

proteneer:
Thanks, after some more sanity checks, this looks good!

proteneer (approved):

tests/test_loss.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/480/commits/f12351e2aa7e50010073c5819964427742d91396

maxentile:
Thanks! Incorporated the suggested new test cases, will merge in a moment.

PR #483, opened 2021-08-09
Add logging statements to worker script

badisa:
* Provides start time and time in seconds


badisa:
This will provide some information when determining what triggered a hang

proteneer (approved):

proteneer:
lgtm - may want to extract a more meaningful name later on but i'm okay with this as is.

PR #484, opened 2021-08-10
Extract some definitions from RABFE example script

maxentile:
Context and motivation: Aiming to add an RABFE training example based on [`examples/validate_relative_binding.py`](https://github.com/proteneer/timemachine/blob/7df21182fd0076030860d7ec3c6847161ba061e1/examples/validate_relative_binding.py). This new script should run in training mode (i.e.  load the expensive cached results and run only the cheaper conversion steps), while duplicating as little as possible of the original script. 

In this PR, two important definitions are extracted to the `fe` module, and a new function `RABFEResult.from_log` is added/tested.
- [x] Extract RABFE thermodynamic cycle definition
- [x] Extract preferred terminal log format for RABFE results
- [x] Add (and test) function to parse terminal log
- [x] Extract custom atom comparison to `fe.atom_mapping`

Further refactoring of the RABFE example script is in progress on the [`rabfe-training`](https://github.com/proteneer/timemachine/tree/rabfe-training) branch

### DIFF
--- a/fe/free_energy_rabfe.py
+++ b/fe/free_energy_rabfe.py
@@ -1,5 +1,5 @@
 from rdkit import Chem
-from typing import List
+from collections import namedtuple
 from jax.config import config; config.update("jax_enable_x64", True)
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


fe/free_energy_rabfe.py
```diff
@@ -11,6 +11,52 @@
...
+    dG_complex_decouple: float
+    dG_solvent_conversion: float
+    dG_solvent_decouple: float
+
+    def log(self):
+        """print stage summary"""
+        print("stage summary for mol:", self.mol_name,
+              "dG_complex_conversion (K complex)", self.dG_complex_conversion,
+              "dG_complex_decouple (E0 + A0 + A1 + E1)", self.dG_complex_decouple,
+              "dG_solvent_conversion (K complex)", self.dG_solvent_conversion,
```
proteneer, line 30: K solvent

proteneer (approved):
LGTM!

PR #485, opened 2021-08-10
Move docs and thermo_deriv into attic

maxentile:
A small quality-of-life improvement we discussed was to move some components of timemachine that are valuable for reference but that are no longer maintained into an `attic/` folder, to ease navigation of the live code.

This PR makes a couple of the easiest moves, where whole top-level folders appear safe to move: 
- [x] Move the backpropagation-through-time notes from `docs/` to `attic/docs/` -- these are valuable for reference, and we may want to use these techniques again in future steps, but this has gotten a bit out of sync with the current implementation
- [x] Move numerical experiments from `thermo_deriv/` to `attic/thermo_deriv/` -- these currently have some unsatisfied dependencies, and I wasn't immediately able to document the intent or result of each experiment -- perhaps @proteneer could add a few words to the readme about any noteworthy aspects we expect to reference later?
- [x] Add brief list of contents to `attic/readme.md`

Later, we might move any smaller bits and pieces we find ourselves tripping over.


proteneer:
Thanks!

> Move the backpropagation-through-time notes from docs/ to attic/docs/ -- these are valuable for reference, and we may want to use these techniques again in future steps, but this has gotten a bit out of sync with the current implementation

Sounds good, was fun while it lasted.

> Move numerical experiments from thermo_deriv/ to attic/thermo_deriv/ -- these currently have some unsatisfied dependencies, and I wasn't immediately able to document the intent or result of each experiment -- perhaps @proteneer could add a few words to the readme about any noteworthy aspects we expect to reference later?

Most of these were written while I was on "vacation"  last summer - I'd need to dig in to what they were doing or they convinced me that thermodynamics works at all later.

badisa (approved):
🚀

proteneer (approved):

Issue #486, opened 2021-08-11
Misc estimator things to cleanup


proteneer:
1) Fix gradients with endpoint correction to use the correct end states
2) Change barostat frequency from 5->25
3) Change du/dp frequency from 5->100 and measure perf differences.

proteneer:
Current Timings for 100k steps on hif2a on master

```
x_interval=1000
du_dp_interval=5
barostat_interval=5
Multiple_steps: 89 seconds
Delta_U's time: 6.5 seconds
```

proteneer:
branch: estimator_cleanup (using `ctxt.multiple_steps_U` as opposed to `ctxt.multiple_steps` and `delta_Us` separately)

`
x_interval=1000
du_dp_interval=5
barostat_interval=5
total time 89.58291339874268
`

(7% speedup roughly)

maxentile:
Closed by https://github.com/proteneer/timemachine/pull/487

PR #487, opened 2021-08-11
Optimize simulation settings and fix gradient bug

proteneer:
Addresses #486


fe/estimator_abfe.py
```diff
@@ -132,6 +132,10 @@ def simulate(lamb, box, x0, v0, final_potentials, integrator, barostat, equil_st
     equil_schedule = np.ones(equil_steps)*lamb
     ctxt.multiple_steps(equil_schedule)
 
+    # (ytz): intentionally hard-coded, I'd rather the end-user *not*
```
proteneer, line 135: (note that equilibration will still use whatever default barostat interval is used, but it's assumed that once we're at equilibrium, interval=25 should be sufficient).

maxentile (approved):
Looks good. One point you mentioned in linked issue https://github.com/proteneer/timemachine/issues/486 is to change the du/dp save interval -- should that be done in the current PR as well?

proteneer:
> Looks good. One point you mentioned in linked issue #486 is to change the du/dp save interval -- should that be done in the current PR as well?

It is done here in this PR:
```
        du_dp_obs.append(custom_ops.AvgPartialUPartialParam(impl, 25))
```

maxentile:
Ahh! Thanks, I had just looked for an interval of "100" and didn't see it. Looks like it addresses all 3 points then.

proteneer:
yeah, I re-ran with an interval of 25 and it didn't slow things down much compared to 100, so figured might as well as collect some more statistics!

PR #488, opened 2021-08-12
Add tests for estimator_abfe and fix a bug.

proteneer:
This PR replaces the old tests for estimator.py (which will soon be deprecated) for estimator_abfe.py instead. This also helped me catch a minor bug in the computation of the derivatives when the endpoint correction is turned on.

### DIFF
--- a/fe/estimator_abfe.py
+++ b/fe/estimator_abfe.py
@@ -361,7 +361,7 @@ def deltaG_fwd(model, sys_params) -> Tuple[Tuple[float, List], np.array]:
 def deltaG_bwd(model, residual, grad) -> Tuple[np.array]:
     """Note: nondiff args must appear first here, even though one of them appears last in the original function's signature!
     """
-     # residual are the partial dG / partial dparams for each term
+    # residual are the partial dG / partial dparams for each term
     # grad[0] is the adjoint of dG w.r.t. loss: partial L/partial dG
     # grad[1] is the adjoint of dG_err w.r.t. loss: which we don't use
     # grad[2] is the adjoint of simulation results w.r.t. loss: which we don't use
--- a/fe/estimator_abfe.py
+++ b/fe/estimator_abfe.py
@@ -305,16 +305,31 @@ def _deltaG(model, sys_params) -> Tuple[Tuple[float, List], np.array]:
     dG = bar_dG # use the exact answer
     dG_grad = []
 
+    # (ytz): results[-1].du_dps contain system parameter derivatives for the
+    # independent, gas phase simulation. They're usually ordered as:
+    # [Bonds, Angles, Torsions, Nonbonded]
+    #
+    # results[0].du_dps contain system parameter derivatives for the core
+    # restrained state. If we're doing the endpoint correction during
+    # decoupling stages, the derivatives are ordered as:
+
+    # [Bonds, Angles, Torsions, Nonbonded, RestraintBonds]
+    # Otherwise, in stages like conversion where the endpoint correction
+    # is turned off, the derivatives are ordered as :
+    # [Bonds, Angles, Torsions, Nonbonded]
+
+    # Note that this zip will always loop over only the
+    # [Bonds, Angles, Torsions, Nonbonded] terms, since it only
+    # enumerates over the smaller of the two lists.
     for rhs, lhs in zip(results[-1].du_dps, results[0].du_dps):
         dG_grad.append(rhs - lhs)
 
     if model.endpoint_correct:
-        # if we're doing endpoint correction, we need to fill in
-        # missing derivatives since zip() from above loops over
-        # the shorter array.
+        # (ytz): Fill in missing derivatives since zip() from above loops
+        # over the shorter array. The "rhs" from results[-1] is identically
+        # zero as the energies do not depend the core restraints.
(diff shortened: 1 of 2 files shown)


tests/test_estimator.py
```diff
@@ -74,29 +93,128 @@ def test_free_energy_estimator():
...
+        grad_fn = jax.grad(loss_fn)
+        grad = grad_fn(sys_params)
 
         assert len(grad) == 2
         assert grad[0].shape == sys_params[0].shape
+        assert grad[1].shape == sys_params[1].shape
+
+
+def test_free_energy_estimator_with_endpoint_correction():
+
```
maxentile, line 138:
```suggestion
def test_free_energy_estimator_with_endpoint_correction():
    """assert shape compatibility between gradients and system parameters"""
```

tests/test_estimator.py
```diff
@@ -74,29 +93,128 @@ def test_free_energy_estimator():
...
+        
+        value_and_grad_fn = jax.value_and_grad(loss_fn)
+        dG, sys_grad = value_and_grad_fn(sys_params) # run fwd, store result, and run bwd
+
+        grad_fn = jax.grad(loss_fn)
+        grad = grad_fn(sys_params)
+
+        assert len(grad) == 3
+        assert grad[0].shape == sys_params[0].shape
         assert grad[1].shape == sys_params[1].shape
```
maxentile, line 220: Comment on expectation for `grad[2].shape, sys_params[2].shape`?

maxentile (approved):

fe/estimator_abfe.py
```diff
@@ -303,10 +305,15 @@ def _deltaG(model, sys_params) -> Tuple[Tuple[float, List], np.array]:
...
 
-    # note this uses the full results, and not just sim_results
     for rhs, lhs in zip(results[-1].du_dps, results[0].du_dps):
         dG_grad.append(rhs - lhs)
 
+    if model.endpoint_correct:
+        # if we're doing endpoint correction, we need to fill in
+        # missing derivatives since zip() from above loops over
+        # the shorter array.
+        dG_grad.append(-results[0].du_dps[-1])
```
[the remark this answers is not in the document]
proteneer: Added more documentation on this.

tests/test_estimator.py
proteneer: Will expand on this a little in the code.

tests/test_estimator.py
proteneer: addressed in `9336d6c`

tests/test_estimator.py
proteneer: addressed in `9336d6c`

fe/estimator_abfe.py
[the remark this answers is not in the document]
proteneer: addressed in `9336d6c`

fe/estimator_abfe.py
```diff
@@ -303,11 +305,31 @@ def _deltaG(model, sys_params) -> Tuple[Tuple[float, List], np.array]:
...
+    # [Bonds, Angles, Torsions, Nonbonded] terms, since it only
+    # enumerates over the smaller of the two lists.
     for rhs, lhs in zip(results[-1].du_dps, results[0].du_dps):
         dG_grad.append(rhs - lhs)
 
     if model.endpoint_correct:
+        # (ytz): Fill in missing derivatives since zip() from above loops
+        # over the shorter array. The "rhs" from results[-1] is identically
+        # zero as the energies do not depend the core restraints.
+        dG_grad.append(-results[0].du_dps[-1])
```
maxentile, line 331: Ahh, thanks -- this clarifies the minus sign as well.

Rephrasing to reduce the number of ways this line appears different from the `dG_grad.append` statement above:
```suggestion
        lhs = results[0].du_dps[-1]
        rhs = 0 # zero as the energies do not depend the core restraints.
        dG_grad.append(rhs - lhs)
```

fe/estimator_abfe.py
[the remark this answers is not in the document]
maxentile: Thanks! The added comment clarifies the current bug fix, and suggests a possibly more self-documenting data structure for a future enhancement (e.g. change the code so that any names that appear only in the comment appear instead as keys or attributes).

fe/estimator_abfe.py
```diff
@@ -303,11 +305,31 @@ def _deltaG(model, sys_params) -> Tuple[Tuple[float, List], np.array]:
...
+    # [Bonds, Angles, Torsions, Nonbonded]
+
+    # Note that this zip will always loop over only the
+    # [Bonds, Angles, Torsions, Nonbonded] terms, since it only
+    # enumerates over the smaller of the two lists.
     for rhs, lhs in zip(results[-1].du_dps, results[0].du_dps):
         dG_grad.append(rhs - lhs)
 
     if model.endpoint_correct:
+        # (ytz): Fill in missing derivatives since zip() from above loops
```
maxentile, line 328: Could add a reminder in case we ever change the list of potentials that appears in lhs of endpoint correction, although unlikely

```suggestion
    if model.endpoint_correct:
        assert len(results[0].du_dps) - len(results[-1].du_dps) == 1
        # (ytz): Fill in missing derivatives since zip() from above loops
```

Issue #470, opened 2021-07-16
Add check that cuttoff * 2 < box_width


badisa:
We need to verify that when we run the non-bonded potential that the cutoff used is less than half the longest edge of the bounding box. If that constraint is invalidated, then a particle can interact with multiple copies of the same particle due to the wrapping of the box.

This check should only be performed every hundred steps or so to ensure that it doesn't have a performance overhead.

PR #489, opened 2021-08-13
Implements #470: Verifies that box width is valid

badisa:
* If cutoff * 2 is greater than the min box width then
  the neighborlist construction is poorly defined since
  a particle can overlap with another tile more than once

### DIFF
--- a/tests/test_parameter_interpolation.py
+++ b/tests/test_parameter_interpolation.py
@@ -28,15 +28,13 @@ class TestInterpolatedPotential(GradientTest):
         np.random.seed(4321)
         D = 3
 
-        cutoff = 0.25
+        cutoff = 1.0
         size = 36
 
         water_coords = self.get_water_coords(D, sort=False)
         coords = water_coords[:size]
         padding = 0.2
-        diag = np.amax(coords, axis=0) - np.amin(coords, axis=0) + padding
-        box = np.eye(3)
-        np.fill_diagonal(box, diag)
+        box = np.eye(3) * 3
 
         N = coords.shape[0]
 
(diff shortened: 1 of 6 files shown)


badisa:
Based on the slow_tests/test_benchmarks.py, this has no impact upon performance.

timemachine/cpp/src/nonbonded.cu
```diff
@@ -303,6 +303,23 @@ void Nonbonded<RealType, Interpolated>::execute_device(
...
+        // Verify that the cutoff and box size are valid together. If cutoff is greater than half the box
+        // then a particle can interact with multiple periodic copies.
+        std::vector<double> h_box(9);
+        gpuErrchk(cudaMemcpyAsync(&h_box[0], d_box, 3*3*sizeof(*d_box), cudaMemcpyDeviceToHost, stream));
+        const double twice_sq_cutoff = 2 * ((cutoff_+nblist_padding_) * (cutoff_+nblist_padding_));
+
+        cudaStreamSynchronize(stream);
+        // Use l2 norm on each row to determine if cutoff is too small
+        for (int i = 0; i < 3; i++) {
+            double row_norm_sq = h_box[i*3+0]*h_box[i*3+0] + h_box[i*3+1]*h_box[i*3+1] + h_box[i*3+2]*h_box[i*3+2];
```
proteneer, line 316: looks right to me, note that we should document somewhere that the box 3x3 matrix corresponds to:
```
[
   [u_x,u_y,u_z],
   [v_x,v_y,v_z], 
   [w_x,w_y,w_z],
]
```

where `u,v,w` correspond to be the box vectors. Different packages have different conventions.

tests/test_parameter_interpolation.py
```diff
@@ -102,9 +100,7 @@ def test_nonbonded_advanced(self):
         water_coords = self.get_water_coords(D, sort=False)
         coords = water_coords[:size]
         padding = 0.2
-        diag = np.amax(coords, axis=0) - np.amin(coords, axis=0) + padding
-        box = np.eye(3)
-        np.fill_diagonal(box, diag)
+        box = np.eye(3) * 3
```
proteneer, line 104: add a test for a non-orthogonal box just to sanity check that the test would fail?

badisa:
Closing as #498 made a rebase messy

PR #490, opened 2021-08-16
Parameter specific learning rates

maxentile:
Addresses https://github.com/proteneer/timemachine/pull/474#issuecomment-891246789  by adding two definitions to a new file `optimizer/precondition.py` : 
- [x] Dictionary of default learning rates, mapping each handler class to a vector of learning rates. By default, all bonded parameters have learning rates of 0. For consistency with other definitions in the `optimize` module, this also sets the learning rate for LJ sigma parameters to 0, but I'm not sure if this should be the default.
- [x] Convenience function that expands this dictionary to a list of `ordered_learning_rates` of the same shapes as `ordered_params`. This can be flattened the same way as ordered params, so it can be multiplied elementwise.

Intended use is [something like](https://github.com/proteneer/timemachine/blob/f74f02a54f190d9a076a180178e083730e55a6d3/slow_tests/test_rabfe_conversion_gradients.py#L42-L46)

```python
g = flatten(ordered_gradients)
lr = flatten(ordered_learning_rates)
raw_search_direction = - g
search_direction = raw_search_direction * lr
```


optimize/precondition.py
```diff
@@ -0,0 +1,33 @@
...
+from ff.handlers.bonded import HarmonicAngleHandler, HarmonicBondHandler, ImproperTorsionHandler, ProperTorsionHandler
+from ff.handlers.nonbonded import LennardJonesHandler, AM1CCCHandler
+
+default_learning_rates = {
+    HarmonicBondHandler:    np.zeros(2),    # k, length
+    HarmonicAngleHandler:   np.zeros(2),    # k, angle
+    ProperTorsionHandler:   np.zeros(3),    # k, phase, periodicity
+    ImproperTorsionHandler: np.zeros(3),    # k, phase, periodicity
+    AM1CCCHandler:          np.ones(1),     # charge increment
+    LennardJonesHandler:  np.array([0, 1]), # sigma, epsilon
```
proteneer, line 11: Intentionally disabling sigma? For some reason I thought it should be the other way around.... (disable epsilon but not sigma).

optimize/precondition.py
```diff
@@ -0,0 +1,33 @@
...
+    """Get a list of arrays of same shape as ordered_params, but containing
+    learning rates"""
+
+    ordered_learning_rates = []
+    for handle, params in zip(ordered_handles, ordered_params):
+        key = handle.__class__
+        lr_row = learning_rates[key]
+        lr_array = np.array([lr_row] * len(params))
+
+        if key == AM1CCCHandler:
```
proteneer, line 26: Would having AM1CCCHandler learning_rates be a scalar, i.e. `1.0`, as opposed to a `np.ones(1)` solve the shape mismatch?

tests/common.py
```diff
@@ -13,6 +17,17 @@
 import itertools
 
 
+def get_default_ff(default_ff='ff/params/smirnoff_1_1_0_ccc.py'):
```
proteneer, line 20: How would this change if we upgraded to say 1.3.0 later on? Should we rename this to be something more specific like 
`get_110_ccc_ff()`?

proteneer (approved):
mostly clerical comments but otherwise LGTM

optimize/precondition.py
maxentile: Unintentional, had things backwards. Addressed in https://github.com/proteneer/timemachine/pull/490/commits/c01ae0a5f901bc9a012e851f381a056e29a35ed8

optimize/precondition.py
maxentile: That should also do it -- addressed in https://github.com/proteneer/timemachine/pull/490/commits/f61b242c786b4c53544d57ff8ced21e836bbc66a

badisa (approved):

tests/common.py
maxentile: >How would this change if we upgraded to say 1.3.0 later on?

Same way we would change the other uses in the project-- this still pops up if we find and replace the literal `ff/params/smirnoff_1_1_0_ccc.py` in all project files. Later if we want to have this be more automatic, we could maybe set `DEFAULT_FF = "ff/params/smirnoff_1_1_0_ccc.py"` to be some project-level constant (which could replace the ~ two dozen duplicates.

>Should we rename this to be something more specific like get_110_ccc_ff()?

Applied this name change in https://github.com/proteneer/timemachine/pull/490/commits/c1243d724475754c8ededcb0b5aac02a25acb6db

maxentile:
Thanks for the constructive feedback.

Merging, but may want to revisit the `default_learning_rates` definition

PR #491, opened 2021-08-17
Task/port smirnoff 1.3.1#463

badisa:


ff/smirnoff_converter.py
```diff
@@ -85,95 +86,390 @@ def parse_quantity(number_string):
...
-                continue
-            elif 'scale' in key:
-                val = float(val)
-            elif key == 'switch_width':
-                continue
-            if key == "version":
-                continue
-            vdws["props"][key] = val 
-        forcefield['vdW'] = vdws
+AM1CCC_CHARGES = {'patterns': [('[#6X4:1]-[#16X1-1:2]', 0.9818644744993272),
```
badisa, line 89: Perhaps it is overkill to add this to the script, it just seemed cleaner than having to copy it over by hand

maxentile: Would it be preferable if this dictionary were stored in a separate named file that is loaded by `ff/smirnoff_converter.py`, rather than being stored inside `ff/smirnoff_converter.py`?

(I don't know if it's important to store information about `AM1CCC_CHARGES` origin or versions, or if these are constrained to be identical to the AM1CCC definitions in `ff/params`)

ff/compare_forcefields.py
```diff
@@ -0,0 +1,60 @@
...
+                    found = False
+                    for comp_pattern in comp_ff[key][subkey]:
+                        if pattern[0] == comp_pattern[0]:
+                            found = True
+                            if len(pattern) != len(comp_pattern):
+                                print(f"Mismatch of size for pattern {pattern[0]}")
+                            if any(ref_val != comp_val for ref_val, comp_val in zip(pattern, comp_pattern)):
+                                print(f"{key} pattern {pattern[0]} differs:")
+                                print(f"Reference  {pattern[1:]}")
+                                print(f"Comparison {comp_pattern[1:]}")
```
maxentile, line 51: Nit: Slightly ambiguous naming (`pattern[0]` contains SMIRKS pattern, `pattern[1:]` contains params). Maybe `smirks, params = item[0], item[1:]` or similar would be preferable.

ff/compare_forcefields.py
```diff
@@ -0,0 +1,60 @@
...
+                                print()
+                    if not found:
+                        print(f"Comp FF has no pattern {pattern[0]}")
+                # sorted_values = sorted(ref_ff[key][subkey], key=lambda x: x[0])
+                # print(sorted_values)
```
maxentile, line 56: Keep?

ff/compare_forcefields.py
```diff
@@ -0,0 +1,60 @@
...
+            else:
+                if ref_ff[key][subkey] != comp_ff[key][subkey]:
+                    print("Difference in {subkey} value: Reference value {ref_ff[key][subkey]} New Value {comp_ff[key][subkey]}")
```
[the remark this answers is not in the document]
badisa: Had missed that, doesn't actually trigger, was only a defensive check

ff/smirnoff_converter.py
badisa: @proteneer What do you think? I am for moving this to another file, though I don't have a good answer regarding if we need to store these charges and if there are versions?

maxentile (changes requested):
Two requests:

1. Could we add permalinks to the original files, either in the PR description, in the `ff/params` readme, or in comment blocks at the top of the ported files? (`openff_unconstrained-*.offxml` from https://github.com/openforcefield/openff-forcefields/tree/master/openforcefields/offxml --> `smirnoff_*.xml` in `ff/`)?

   I think there could already be an ambiguity about provenance -- based on the comments https://github.com/proteneer/timemachine/blob/409e858b9de5dfabef277334d9ff633e2e4b93d8/ff/smirnoff_1.3.1.xml#L5 I think these may have come from the default OpenFF versions assuming constraints (`openff-*.offxml`). Unsure if it may have been intended to start from the unconstrained versions instead, or if this distinction doesn't matter anymore. (In previous versions, constrained H bonds did not always get a HarmonicBond term assigned, for example, so it would have been important to get the `unconstrained` versions -- eyeballing, it looks like [this specific issue](https://github.com/proteneer/timemachine/blob/409e858b9de5dfabef277334d9ff633e2e4b93d8/ff/smirnoff_1.3.1.xml#L98-L102) doesn't affect the newly ported files, but I'm unsure if there are any other relevant differences.)

2. Could we add in the PR description (or ideally, the `ff/params` readme) a comment about whether the full specification used by 1.3.1 is supported in timemachine, and if not, whether there are any special cautions when applying the ported parameters?

   For example, I don't think we support WBO interpolation, referenced for [bonds](https://github.com/proteneer/timemachine/blob/409e858b9de5dfabef277334d9ff633e2e4b93d8/ff/smirnoff_1.3.1.xml#L12) and [torsions](https://github.com/proteneer/timemachine/blob/409e858b9de5dfabef277334d9ff633e2e4b93d8/ff/smirnoff_1.3.1.xml#L147) in `timemachine/ff/smirnoff_1.3.1.xml`. It is currently unclear if the intent is to implement WBO interpolation in timemachine, or to use the ported parameters without WBO interpolation.

ff/compare_forcefields.py
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/491/commits/409e858b9de5dfabef277334d9ff633e2e4b93d8

ff/compare_forcefields.py
maxentile: Resolved in https://github.com/proteneer/timemachine/pull/491/commits/409e858b9de5dfabef277334d9ff633e2e4b93d8

ff/readme.md
```diff
@@ -0,0 +1,31 @@
+# Timemachine Forcefields
```
proteneer, line 1: Nice Documentation!

ff/smirnoff_converter.py
```diff
@@ -85,95 +88,109 @@ def parse_quantity(number_string):
...
+            for s in itemlist:
+                patt = s.attributes['smirks'].value
+                epsilon = parse_quantity(s.attributes['epsilon'].value)
+                if "rmin_half" in s.attributes:
+                    rmin_half = parse_quantity(s.attributes['rmin_half'].value)
+                    sigma = 2. * rmin_half / (2.**(1. / 6.))
+                else:
+                    sigma = parse_quantity(s.attributes['sigma'].value)
+                # Take sqrt of epislon to avoid singularity in backprop
+                params.append([patt, sigma, np.sqrt(epsilon)])
```
proteneer, line 173: Hard to tell what the actual diffs are due to indentation, but I take it this is the only one that changed?

proteneer (approved):

ff/readme.md
```diff
@@ -0,0 +1,31 @@
...
+
+The forcefields that are bundled with Timemachine come from the [Open Forcefield Initiative](https://openforcefield.org/), using the constrained definitions. The following forcefields are currently included:
+
+* OpenForceField 1.1.0 - [original](https://github.com/openforcefield/openff-forcefields/blob/master/openforcefields/offxml/openff-1.1.0.offxml)
+* OpenForceField 1.3.1 - [original](https://github.com/openforcefield/openff-forcefields/blob/master/openforcefields/offxml/openff-1.3.1.offxml)
+* OpenForceField 2.0.0 - [original](https://github.com/openforcefield/openff-forcefields/blob/master/openforcefields/offxml/openff-2.0.0.offxml)
+
+## Forcefield Support
+
+Timemachine does not currently the full OpenForceField definition. The following are some features that are not currently supported:
```
maxentile, line 27:
```suggestion
Timemachine does not currently support the full OpenForceField definition. The following are some features that are not currently supported:
```

ff/readme.md
```diff
@@ -0,0 +1,31 @@
...
+
+* [Virtual Sites](https://open-forcefield-toolkit.readthedocs.io/en/latest/virtualsites.html)
+* Wiberg Bond Ordering Interpolation
```
maxentile, line 30:
```suggestion
* Fractional bond order Interpolation for [bonds](https://open-forcefield-toolkit.readthedocs.io/en/0.10.0/users/smirnoff.html#fractional-bond-orders) and [torsions](https://open-forcefield-toolkit.readthedocs.io/en/0.10.0/users/smirnoff.html#fractional-torsion-bond-orders)
```

ff/readme.md
```diff
@@ -0,0 +1,31 @@
...
+
+After conversion, it is possible to compare two converted forcefield to see the differences by running the following:
+
+```bash
+python compare_forcefields.py params/smirnoff_1_1_0_ccc.py params/smirnoff_2_0_0_ccc.py
+```
+
+## Built in Forcefields
+
+The forcefields that are bundled with Timemachine come from the [Open Forcefield Initiative](https://openforcefield.org/), using the constrained definitions. The following forcefields are currently included:
```
maxentile, line 19: Thanks -- this clarifies whether constrained or unconstrained was intended!

maxentile (approved):

ff/smirnoff_converter.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/491/commits/d755da86beb9fe268dc9f01c23ade01fbb18cd17

ff/smirnoff_converter.py
badisa: The output keys for the FF are also different. Previously they resulted in a dict that the keys had to be renamed. I don't know if that was necessary historically or if there is a secondary stage of conversion I was meant to run. Otherwise this is the only substantive change to the script.

PR #492, opened 2021-08-17
Increase message size to 1GB on worker side

badisa:
Attempting to rerun inference I ran into the workers complaining about the message size being larger than 500mb. This makes it consistent with the client side.

Rerunning with this change resolved the issue.


maxentile (approved):
Good catch!

PR #495, opened 2021-08-18
Return Binary XTC from RABFE Simulate

badisa:


badisa:
Have only tested this by running the FEP benchmark CDK8 in trivial (1000 steps) scenarios, intend to do a larger run soon.

fe/estimator_abfe.py
```diff
@@ -153,13 +160,16 @@ def simulate(lamb, box, x0, v0, final_potentials, integrator, barostat, equil_st
...
 
-    result = SimulationResult(
-        xs=xs,
-        boxes=boxes,
-        du_dps=grads,
-        lambda_us=full_us,
-    )
-
+    with NamedTemporaryFile(suffix=".xtc") as temp:
+        traj = mdtraj.Trajectory(xs, mdtraj.Topology.from_openmm(openmm_topo))
```
proteneer, line 164: You can set topology=None here:

```
mdtraj.Trajectory(xs, topology=None)
```

ex:

``` python
xyz=np.random.rand(100,50,3)
boxes = np.random.rand(100,3,3)
traj = mdtraj.Trajectory(xyz=xyz, topology=None)
traj.unitcell_vectors=boxes
>>> traj.save_xtc('debug.xtc')
```

xtc format does not store anything about the topology.

fe/estimator_abfe.py
```diff
@@ -80,6 +84,9 @@ def simulate(lamb, box, x0, v0, final_potentials, integrator, barostat, equil_st
     lambda_windows: list of float
         lambda windows we evaluate energies at.
 
+    openmm_topo: openmm.app.Topology
```
proteneer, line 87: not really needed, see below.

fe/model_rabfe.py
```diff
@@ -285,6 +284,14 @@ def _predict_a_to_b(
             seed
         )
 
+        # uncomment if we want to visualize.
```
maxentile, line 287:
```suggestion
```

maxentile (approved):

fe/estimator_abfe.py
```diff
@@ -338,14 +346,26 @@ def _deltaG(model, sys_params) -> Tuple[Tuple[float, List], np.array]:
         k_translation = 200.0
         k_rotation = 100.0
         start = time.time()
+        md_topo = mdtraj.Topology.from_openmm(model.openmm_topo)
+        with NamedTemporaryFile(suffix=".xtc") as temp:
+            with open(temp.name, "wb") as ofs:
+                ofs.write(results[-2].frames)
+            frames = mdtraj.load_xtc(temp.name, top=md_topo)
```
proteneer, line 353: ugh... does load_xtc really need a top?

fe/model_rabfe.py
```diff
@@ -285,6 +284,14 @@ def _predict_a_to_b(
...
+        combined_topology = model_utils.generate_imaged_topology(
```
proteneer, line 288: Note that this does rely on the image_molecules() call inside mdtraj, which has an extremely small, but non-zero, possibility of segfaulting. Not sure if this is truly safe or not.

fe/estimator_abfe.py
```diff
@@ -338,14 +346,26 @@ def _deltaG(model, sys_params) -> Tuple[Tuple[float, List], np.array]:
...
         lhs_du, rhs_du, rotation_samples, translation_samples = endpoint_correction.estimate_delta_us(
             k_translation=k_translation,
             k_rotation=k_rotation,
             core_idxs=core_restr.get_idxs(),
             core_params=core_restr.params.reshape((-1,2)),
             beta=model.beta,
-            lhs_xs=results[-2].xs,
-            rhs_xs=results[-1].xs
+            lhs_xs=lhs_xs,
+            rhs_xs=rhs_xs
```
proteneer, line 368: Note that this is saved with 3 decimal places of precision. So if we're in units of nanometers, we can be precise up to 1/1000th of a nanometer, or 1/100th of an angstroms. Since the endpoint correction only looks at the energy differences in harmonic angles, forces, and orientational restraints, I think this is pretty safe.

I know that @maxentile  had concerns re: loss of precision in the nonbonded forces, but this doesn't matter for the endpoint correction.

fe/estimator_abfe.py
badisa: Yup.... It does...

proteneer (approved):

badisa:
Going to run a few stages of inference with production settings against this to verify before merging.

badisa:
Going to reopen with float32 as we had discussed.

PR #501, opened 2021-08-24
Implements #470: Verifies that box width is valid

badisa:
* If cutoff * 2 is greater than the min box width then the neighborlist construction is poorly defined since a particle can overlap with another tile more than once


timemachine/cpp/src/nonbonded.cu
```diff
@@ -336,12 +336,28 @@ void Nonbonded<RealType, Interpolated>::execute_device(
...
+        const double db_cutoff = (cutoff_+nblist_padding_) * 2;
+        cudaStreamSynchronize(stream);
+        // Verify that box is ortholinear and the width of the box in all dimensions is greater than twice the cutoff
+        for (int i = 0; i < 9; i++) {
+            if (i == 0 || i == 4 || i == 8) {
+                if (h_box[i] < db_cutoff) {
+                    throw std::runtime_error("Cutoff with padding is more than half of the box width, neighborlist is no longer reliable");
+                }
+            } else if (h_box[i] != 0.0) {
+                throw std::runtime_error("Provided non-ortholinear box, unable to compute nonbonded energy");
```
proteneer, line 354: i think orthogonal is the proper term here

proteneer (approved):
approved once tests pass!

PR #502, opened 2021-08-31
Feature/map cores by dist

badisa:


tests/test_free_energy_rabfe.py
```diff
@@ -23,3 +31,45 @@ def test_rabfe_result_to_from_log():
...
+    smi_a = "CCCONNN"
+    smi_b = "CCCNNN"
+    mol_a = Chem.MolFromSmiles(smi_a)
+    mol_a = Chem.AddHs(mol_a)
+    mol_b = Chem.MolFromSmiles(smi_b)
+    mol_b = Chem.AddHs(mol_b)
+    for mol in [mol_a, mol_b]:
+        AllChem.EmbedMolecule(mol, randomSeed=seed)
+
+    core = setup_relative_restraints_by_distance(mol_a, mol_b)
```
proteneer, line 46: we should add a test asserting that all distances from this core are less than cutoff

proteneer: (and probably a good idea to test for more cutoffs)

fe/free_energy_rabfe.py
```diff
@@ -302,6 +302,68 @@ def setup_relative_restraints(
...
+            if not terminal and b.GetDegree() == 1:
+                continue
+            if np.linalg.norm(ligand_coords_a[idx] - ligand_coords_b[b_idx]) < cutoff:
+                core_idxs_a.append(idx)
+                core_idxs_b.append(b_idx)
+    assert len(core_idxs_a) == len(core_idxs_b), "Core sizes were inconsistent"
+
+    ri = np.expand_dims(ligand_coords_a[core_idxs_a], 1)
+    rj = np.expand_dims(ligand_coords_b[core_idxs_b], 0)
+    rij = np.sqrt(np.sum(np.power(ri-rj, 2), axis=-1))
```
maxentile, line 352:
```suggestion
    from scipy.spatial.distance import cdist
    rij = cdist(ligand_coords_a[core_idxs_a], ligand_coords_b[core_idxs_b])
```

maxentile (approved):
Looks good!

Question: Should we also remove/replace any of the following?
* [`fe.free_energy_rabfe.setup_relative_restraints`](https://github.com/proteneer/timemachine/blob/35a810f3e9738993a82d9c0a1fc54bba91a0834b/fe/free_energy_rabfe.py#L261-L303)
* [`fe.atom_mapping.get_core_by_matching`](https://github.com/proteneer/timemachine/blob/35a810f3e9738993a82d9c0a1fc54bba91a0834b/fe/atom_mapping.py#L212-L220)
* [`fe.atom_mapping.get_core_by_geometry`](https://github.com/proteneer/timemachine/blob/35a810f3e9738993a82d9c0a1fc54bba91a0834b/fe/atom_mapping.py#L223-L231)
* [`fe.utils.core_from_distances`](https://github.com/proteneer/timemachine/blob/35a810f3e9738993a82d9c0a1fc54bba91a0834b/fe/utils.py#L102-L115)
* [`fe.utils.simple_geometry_mapping`](https://github.com/proteneer/timemachine/blob/35a810f3e9738993a82d9c0a1fc54bba91a0834b/fe/utils.py#L118-L148)

proteneer (approved):
lgtm!

fe/free_energy_rabfe.py
proteneer: I think for this case its fine, but in general I wouldn't advise going down the cdist route since its not jax.grad/vjpable.

proteneer:
> fe.free_energy_rabfe.setup_relative_restraints
is probably stale, I don't see any good uses for it so I think we can remove it. Not sure about the others, and not sure if I authored them.

badisa:
setup_relative_restraints is used within `validation_binding_hydration`, should that just be replaced by the distance match?

proteneer:
Hm let's keep the function then - and do the swapping out validation in another PR.

PR #504, opened 2021-09-01
Update Jaxlib to 0.1.61

badisa:
* 0.1.59 no longer available on pypi


maxentile (approved):
Looks good!

We might need to increment the pinned jax/jaxlib version again in 1-2 months. It looks like only the most recent 12 releases are on pypi https://pypi.org/project/jaxlib/#history , and historically new releases of jaxlib have appeared every few weeks, so I guess we should expect 0.1.61 to drop off pypi soon, too.

PR #505, opened 2021-09-09
Task/clean up rabfe script

badisa:


examples/validate_relative_binding.py
```diff
@@ -22,27 +22,25 @@
...
 
 import timemachine
 from timemachine.potentials import rmsd
 from md import builders, minimizer
 from rdkit.Chem import rdFMCS
-from fe.atom_mapping import CompareDistNonterminal
 
 if __name__ == "__main__":
 
-    multiprocessing.set_start_method('spawn')
```
badisa, line 44: @proteneer Based on my understanding of multiprocessing, we can rely on the `get_context` in the `CUDAPoolClient`. If this was added for another use case let me know

maxentile (approved):

examples/validate_relative_binding.py
proteneer: IIRC this method must be called only once, and at the start of the run. This was added at the start to basically detect if anything else (imports etc.) either called set_start_method before (in which this line would fail), or after (in which the later line would fail).

examples/validate_relative_binding.py
```diff
@@ -392,17 +391,18 @@ def pred_fn(params, mol, mol_ref):
 
 
     for epoch in range(cmd_args.epochs):
-        epoch_params = serialize_handlers(ordered_handles)
-        # dataset.shuffle()
+        dataset.shuffle()
```
proteneer, line 394: Did we want to shuffle for inference (eg. property_field is None)

badisa: In that case it seems like it shouldn't matter? Ordering in inference, with or without the labels, is inconsequential as far as I can tell

proteneer: would it be okay for the default be to not-shuffle, and add an argument to enable it?

badisa: Works for me

PR #506, opened 2021-09-10
Add Bromine to standard QLJ typer

badisa:
* Allows Bromine to be used within RABFE

### DIFF
--- a/fe/topology.py
+++ b/fe/topology.py
@@ -21,6 +21,12 @@ def standard_qlj_typer(mol):
     in a relatively simple and forcefield independent way. The
     parameters here roughly follow the Smirnoff 1.1.0 Lennard Jones
     types.
+
+    These values are taken from ff/params/smirnoff_1_1_0_cc.py, rounding down
+    to two decimal places for sigma and one decimal place for epsilon.
+
+    returns:
+        list of tuples containing (charge, sigma, epsilon)
     """
 
     standard_qlj = []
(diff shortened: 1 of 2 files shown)


fe/topology.py
```diff
@@ -21,6 +21,12 @@ def standard_qlj_typer(mol):
     in a relatively simple and forcefield independent way. The
     parameters here roughly follow the Smirnoff 1.1.0 Lennard Jones
     types.
+
+    These values are taken from ff/params/smirnoff_1_1_0_cc.py, rounding down
+    to two decimal places for sigma and one decimal place for epsilon.
```
maxentile, line 26: Nice! Confirmed this is an accurate description of how the previous parameters were obtained? https://github.com/proteneer/timemachine/pull/450#discussion_r653976062

proteneer (approved):
lgtm, we may rework how this is done later on pending reweighting results.

fe/topology.py
```diff
@@ -55,6 +61,8 @@ def standard_qlj_typer(mol):
             val = (0.0, 0.35, 1.0)
         elif a_num == 17:
             val = (0.0, 0.35, 1.0)
+        elif a_num == 35:
```
maxentile, line 64: Are there other elements `standard_qlj_typer` should cover? (I imagine this should eventually have the same elemental coverage as OpenFF 1.1. A previous commit expanded `standard_qlj_typer` to cover chlorine. Perhaps a future commit would expand to cover iodine.)

maxentile (approved):

PR #508, opened 2021-10-04
Implement CPU optimal alignment

proteneer:
This PR implements RMSD alignment on the CPU of two point clouds, partially using code from rmsd_restraint.cu (which is mostly deprecated).

### DIFF
--- a/tests/test_rmsd_align.py
+++ b/tests/test_rmsd_align.py
@@ -1,0 +1,38 @@
+# test the custom_op for rmsd alignment
+
+from timemachine.lib import custom_ops
+
+from timemachine.potentials import rmsd
+import numpy as np
+from testsystems import relative
+from scipy.stats import special_ortho_group
+
+
+def test_rmsd_align():
+
+    N = 25
+
+    for _ in range(100):
+
+        x1 = np.random.rand(N,3)
+        random_t = np.random.rand(3)
+        random_R = special_ortho_group.rvs(3)
+        x2 = x1@random_R + random_t
+
+        assert np.linalg.norm(x2 - x1) > 1e-6
+
+        x2_aligned_reference = rmsd.align_x2_unto_x1(x1, x2)
+
+        # com2 should equal to that of com1
+        np.testing.assert_almost_equal(np.mean(x1, axis=0), np.mean(x2_aligned_reference, axis=0))
+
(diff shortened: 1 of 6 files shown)


tests/test_rmsd_align.py
```diff
@@ -0,0 +1,38 @@
...
+
+from timemachine.lib import custom_ops
+
+from timemachine.potentials import rmsd
+import numpy as np
+from testsystems import relative
+from scipy.stats import special_ortho_group
+
+
+def test_rmsd_align():
```
maxentile, line 11:
```suggestion
def test_rmsd_align():
    """Assert that the same optimal alignment is recovered by both
    reference (`rmsd.align_x2_unto_x1`) and CPU (`custom_ops.rmsd_align`),
    for cases where `x1` and `x2` differ by a rigid transformation"""
```

tests/test_rmsd_align.py
```diff
@@ -0,0 +1,38 @@
...
+
+    N = 25
+
+    for _ in range(100):
+
+        x1 = np.random.rand(N,3)
+        random_t = np.random.rand(3)
+        random_R = special_ortho_group.rvs(3)
+        x2 = x1@random_R + random_t
```
maxentile, line 20: Question: should this also exercise the case where `is_reflection` is true? https://github.com/proteneer/timemachine/blob/1ba4b20ea3efc4631f77cb637661ba2c78bb47eb/timemachine/cpp/src/rmsd_align.cpp#L51-L54

(If so, changing the test to use `ortho_group` instead of `special_ortho_group` might do the trick.)

proteneer: I'd thought of doing this initially, but I wanted to initially write a version where I can assert that the optimally aligned RMSD is zero. In the case of an improper rotation projected from O(3) into SO(3), the RMSD would no longer be zero. But you're right that we should still add a test for this, at least to ensure that the RMSD is smaller and that the python/cpp results are consistent.

proteneer: PS thanks a lot for this comment! this actually helped me catch a bug in how I implemented the reflection case.

maxentile (approved):

badisa (approved):
LGTM

PR #514, opened 2021-10-20
Improve Jax nonbonded test case

maxentile:
One way the Jax nonbonded potentials are tested is by generating random values for all [10 arguments](https://github.com/proteneer/timemachine/blob/a69cf9c1d417d0996f23ec1946e137029f33ee1d/tests/test_jax_nonbonded.py#L20), and asserting self-consistency between two implementations `implementation_a(*random_args) == implementation_b(*random_args)`.

[Previously](https://github.com/proteneer/timemachine/blob/f216353677e91a8948a404b200a5747a0e644f8c/tests/test_jax_nonbonded.py#L25-L60), the randomized conformer and box arguments produced large, sparse boxes containing few if any interactions within cutoff, meaning the test was insensitive, as @proteneer noted.

Worryingly, fixing the test to generate a more crowded box (with [clashes resolved](https://github.com/proteneer/timemachine/blob/a69cf9c1d417d0996f23ec1946e137029f33ee1d/tests/test_jax_nonbonded.py#L25-L69)) caused the test to fail!

An easier variant of the test, where only the conformer is randomized but the rest of the arguments are held fixed, passed.

To isolate which difficult feature of the test generator was causing the test to fail, I considered toggling randomization of each argument independently. The root cause of the test failure was a second [bug within the test](https://github.com/proteneer/timemachine/commit/2f9d16c70c2ffb71e67a7081a5a261d9d760093f), related to how the random `lj_rescale_mask` and `charge_rescale_mask` arguments were generated. (The randomized masks could be asymmetric. The two nonbonded implementations will have different behavior for asymmetric `rescale_masks`, since one uses the upper right triangle, and the other uses the whole matrix.)

- [x] Remove "sparsely populated" box from test
- [x] Add clash-free "crowded" box to test
- [x] Refactor test to allow randomization of each argument to be toggled independently (default behavior is still to randomize all arguments)
- [x] Generate symmetric random `rescale_mask`s in test
- [x] Make arbitrary changes to distributions of some test arguments (e.g. `sig ~ Unif([0,1])` -> `sig ~ Unif([0, 0.1])`)
- [x] Remove an `==` comparison (use `assert_almost_equal` instead)
- [x] Add waterbox to test

### DIFF
--- a/tests/test_jax_nonbonded.py
+++ b/tests/test_jax_nonbonded.py
@@ -154,9 +154,6 @@ def generate_random_inputs(n_atoms, dim, instance_flags=difficult_instance_flags
         if instance_flags['randomize_lj_rescale_mask']:
             lj_rescale_mask[i, j] = lj_rescale_mask[j, i] = 0.0
 
-    if instance_flags['randomize_charge_rescale_mask']:
-        charge_rescale_mask
-
     beta = 2.0
     if instance_flags['randomize_beta']:
         beta += rand()


proteneer:
That looks like a singularity in the `du/deps` parameter derivative calculation?

proteneer:
Okay, I think this is then a simple case of just skipping the interaction entirely if either eps is zero to avoid to the singularity.

PR #521, opened 2021-10-29
Make equilibration cache files more human readable

badisa:
* Allows running multiple inference runs in the same dir without
  colliding

### DIFF
--- a/examples/validate_relative_binding.py
+++ b/examples/validate_relative_binding.py
@@ -45,6 +45,47 @@ from timemachine.potentials import rmsd
 from md import builders, minimizer
 from rdkit.Chem import rdFMCS
 
+
+def cache_wrapper(cache_path: str, fxn: callable, overwrite: bool = False) -> callable:
+    """Given a path and a function, will either write the result of the function call
+    to the file or read form the file. The output of the function is expected to be picklable.
+
+    Parameters
+    ----------
+    cache_path: string
+        Path to write pickled results to and to read results from
+
+
+    fxn: callable
+        Function whose result to cache. Used for expensive functions whose results
+        should be stored between runs.
+
+    overwrite: boolean
+        Overwrite the cache if the file already exists. Useful is the cache file was incorrectly
+        generated or the implementation of fxn has changed.
+
+    Returns
+    -------
+
+    wrapped_function: callable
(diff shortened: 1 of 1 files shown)


examples/validate_relative_binding.py
```diff
@@ -45,6 +45,26 @@
 from md import builders, minimizer
 from rdkit.Chem import rdFMCS
 
+
+def cache_wrapper(cache_path: str, f):
```
badisa, line 49: Feels like we might want to put this somewhere else so we can use this sort of logic in other scripts.

maxentile, line 49: At first glance I thought `f` was a file object.

Naming suggestions:
* retain single letter name, add type hint `(cache_path: str, f)` -> `(cache_path: str, f: callable)`
* expand single letter name, so type hint would be redundant `(cache_path: str, f)` -> `(cache_path: str, fxn)`

examples/validate_relative_binding.py
```diff
@@ -181,28 +201,24 @@
     pressure = 1.0
     dt = 2.5e-3
 
+    cached_equil_path = "equil_{}_blocker_{}.pkl".format(
+        os.path.basename(cmd_args.protein_pdb).split(".")[0].replace(" ", "_"), cmd_args.blocker_name.replace(" ", "_")
```
maxentile, line 205: Seems a safe assumption that no directory names in path/to/proteinpdb contain a " ", not sure if we might want to do any other string -> filename sanitization here

examples/validate_relative_binding.py
```diff
@@ -45,6 +45,26 @@
 from md import builders, minimizer
 from rdkit.Chem import rdFMCS
 
+
+def cache_wrapper(cache_path: str, f):
+    """Given a path and a function, will either write the result of the function call
+    to the file or read form the file. The output of the function is expected to be pickable
```
maxentile, line 51: Spelling

examples/validate_relative_binding.py
maxentile, line 49: Should we add an optional argument `overwrite=False`? (Perhaps a brief comment about when and how we should invalidate the cache would be helpful too!)

maxentile (approved):
Looks good!

PR #524, opened 2021-10-29
Add enhanced sampling tests for the condensed phase.

proteneer:
This extends #518 to the condensed phase by using Multiple Try Metropolis (https://en.wikipedia.org/wiki/Multiple-try_Metropolis). Theory and prototyping was worked out with @maxentile over the past few weeks. 

On the biphenyl case, using K=100, we can achieve an acceptance rate as high as 77%, with about half the samples 
 "enhancing sampling" in that it flips the torsion of interest. Note that this still working with decharged ligands, and introduces a decharged version of the forcefield.


proteneer:
Note to self: tests will break until #520 gets merged.

maxentile (approved):

md/enhanced.py
```diff
@@ -351,3 +351,98 @@ def sample_from_log_weights(weighted_samples, log_weights, size):
     assert np.abs(np.sum(weights) - 1) < 1e-5
     idxs = np.random.choice(np.arange(len(weights)), size=size, p=weights)
     return weighted_samples[idxs]
+
+
+from fe import free_energy
```
badisa, line 356: Do we care about these being at the top of the file?

md/enhanced.py
```diff
@@ -351,3 +351,98 @@ def sample_from_log_weights(weighted_samples, log_weights, size):
...
+
+    lamb = 0.0
+    equil_schedule = np.ones(50000) * lamb
+    equil_ctxt.multiple_steps(equil_schedule)
+
+    x0 = equil_ctxt.get_x_t()
+    v0 = np.zeros_like(x0)
+
+    # production
+    intg = lib.LangevinIntegrator(temperature, dt, friction, masses, 2021)
```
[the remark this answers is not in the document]
proteneer: hm good call - best I move the seed into the args and let the caller explicitly set it. This actually may be why I was getting identical histograms in the repeat runs for the vanilla simulation.

tests/test_enhanced_sampling.py
```diff
@@ -91,4 +102,203 @@ def get_torsion(x_t):
...
+
+    @jax.jit
+    def get_torsion(x_t):
+        torsion_idxs = np.array([5, 6, 7, 8])
+        cijkl = x_t[torsion_idxs]
+        return bonded.signed_torsion_angle(*cijkl)
+
+    batch_get_torsion = jax.jit(jax.vmap(get_torsion))
+    vacuum_weights = np.exp(log_weights - logsumexp(log_weights))
+    assert np.abs(np.sum(vacuum_weights) - 1) < 1e-5
```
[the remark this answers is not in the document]
proteneer: This was just arbitrary I think I can bring this down to `1e-10` in double precision.

md/enhanced.py
```diff
@@ -351,3 +351,98 @@ def sample_from_log_weights(weighted_samples, log_weights, size):
...
+from timemachine import lib
+from timemachine.lib import custom_ops
+from md import builders
+from md.barostat.utils import get_group_indices, get_bond_list
+
+
+def get_solvent_phase_system(mol, ff):
+    masses = np.array([a.GetMass() for a in mol.GetAtoms()])
+    water_system, water_coords, water_box, water_topology = builders.build_water_system(3.0)
+    water_box = water_box + np.eye(3) * 0.5  # add a small margin around the box for stability
```
badisa, line 367: I thought minimization removed our need for the margin, are there still cases that we need it? Or is stability specific to the enhanced sampling?

tests/test_enhanced_sampling.py
```diff
@@ -91,4 +102,203 @@ def get_torsion(x_t):
...
+    xa = x_solvent[-num_atoms:]
+    xb = x_gas
+
+    assert xa.shape == xb.shape
+
+    xb_new = rmsd.align_x2_unto_x1(xa, xb)
+    return xb_new
+
+
+def test_condensed_phase_mtm():
```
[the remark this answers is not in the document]
proteneer: let's talk about how to refactor this into a proper class.

tests/test_enhanced_sampling.py
```diff
@@ -91,4 +102,203 @@ def get_torsion(x_t):
...
+    for iteration in range(num_batches):
+
+        x_solvent, box_solvent = sample_generator.send(next_x)
+        before_U_k_sample = before_U_k(x_solvent, box_solvent)
+        log_pi_x = -before_U_k_sample / kT
+
+        solvent_torsion = get_torsion(x_solvent[-num_ligand_atoms:])
+        enhanced_torsions.append(solvent_torsion)
+
+        # Multiple Try Metropolis (MTM) on K samples
```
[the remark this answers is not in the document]
proteneer: This all sounds great to me!

proteneer:
@maxentile I refactor the MTM code into its own class, to hopefully follow stylistically to your initial code in `moves.py`. Can you take a look? 

Let's defer the `lambda(x,y)` enhancements to a separate PR later on?

md/enhanced.py
proteneer: yes, will move soon!

md/moves.py
```diff
@@ -56,3 +56,45 @@ def n_accepted(self):
...
+    def __init__(self, K):
+        self.K = K
+        self.n_accepted = 0
+        self.n_proposed = 0
+
+    def batch_propose(self, x: CoordsVelBox):
+        raise NotImplementedError()
+
+    def batch_log_prob(self, proposals):
+        raise NotImplementedError()
```
maxentile, line 74: Suggestion: Would expose `batch_propose` and `batch_log_prob` as constructor arguments.

md/moves.py
```diff
@@ -56,3 +56,45 @@ def n_accepted(self):
     @property
     def n_proposed(self):
         return np.sum(self.n_proposed_by_move)
+
+
+from scipy.special import logsumexp
+
+
+class MultipleTryMetropolis:
+    def __init__(self, K):
```
maxentile, line 65: Would add reference to https://www.stat.purdue.edu/~fmliang/papers/2000/Mtry.pdf in docstring, and a comment that implicitly `lambda(x, y) = 1 / proposal_pdf(x, y)`

md/moves.py
```diff
@@ -56,3 +56,45 @@ def n_accepted(self):
...
+    def move(self, x: CoordsVelBox) -> CoordsVelBox:
+        self.n_proposed += 1
+        # assume proposals themselves have water and ligand aligned and combined
+        numerator_proposals = self.batch_propose(x)
+        numerator_log_probs = self.batch_log_prob(numerator_proposals)
+        assert self.K == len(numerator_proposals)
+        assert self.K == len(numerator_log_probs)
+        normalized_numerator_log_probs = np.exp(numerator_log_probs - logsumexp(numerator_log_probs))
+
+        new_sample = numerator_proposals[np.random.choice(np.arange(self.K), p=normalized_numerator_log_probs)]
```
maxentile, line 85: naming nit: "new_sample" -> "x_proposed", to be parallel to "x"

tests/test_enhanced_sampling.py
```diff
@@ -91,4 +103,207 @@ def get_torsion(x_t):
...
+    plt.show()
+
+    vanilla_samples_lhs, _ = np.histogram(vanilla_torsions, bins=50, range=(-np.pi, 0), density=True)
+
+    # check for consistency with vanilla samples
+    assert np.mean((enhanced_torsions_lhs - vanilla_samples_lhs) ** 2) < 5e-2
+
+
+if __name__ == "__main__":
+    test_condensed_phase_mtm()
```
badisa, line 309: Is this to imply this test shouldn't be part of the regular test suite?

badisa (approved):

tests/test_enhanced_sampling.py
```diff
@@ -120,6 +121,45 @@ def align_sample(x_gas, x_solvent):
...
+
+        self.vacuum_samples = vacuum_samples
+        self.vacuum_log_weights = vacuum_log_weights
+        self.batch_log_prob_fn = jax.vmap(log_prob_fn)
+
+        super().__init__(K)
+
+    def batch_propose(self, x: CoordsVelBox) -> List[CoordsVelBox]:
+
+        ligand_samples = enhanced.sample_from_log_weights(self.vacuum_samples, self.vacuum_log_weights, self.K)
```
maxentile, line 139: Suggestion: Another way to split this up would be to have a `AlignedLigandResampler(samples, log_weights)` or similar that knows how to do `batch_propose`, and then pass the `resampler.batch_propose` function as an argument to the `MultipleTryMetropolis(batch_log_prob, batch_propose)` constructor.

tests/test_enhanced_sampling.py
```diff
@@ -120,6 +121,45 @@ def align_sample(x_gas, x_solvent):
...
+class AlignedMover(MultipleTryMetropolis):
+    def __init__(self, K, vacuum_samples, vacuum_log_weights, log_prob_fn):
+
+        self.vacuum_samples = vacuum_samples
+        self.vacuum_log_weights = vacuum_log_weights
+        self.batch_log_prob_fn = jax.vmap(log_prob_fn)
+
+        super().__init__(K)
+
+    def batch_propose(self, x: CoordsVelBox) -> List[CoordsVelBox]:
```
maxentile, line 137: Docstring should note assumption that ligand is in the last `num_ligand_atoms` indices of `x` -- in future, we might want to replace coordinates of a different subsystem.

maxentile:
>@maxentile I refactored the MTM code into its own class, to hopefully follow stylistically to your initial code in moves.py. Can you take a look?

Looks great overall! Will leave a couple follow-up comments inline, nothing major.

>Let's defer the lambda(x,y) enhancements to a separate PR later on?

Agreed! Unclear if there's practical benefit from making this flexible, and it introduces complexity.

md/enhanced.py
proteneer: I think there are still cases where waters get added at the PBC boundary - the barostat will equilibrate the box size down to the correct size though.

tests/test_enhanced_sampling.py
proteneer: I will remove this in a final pass!

tests/test_enhanced_sampling.py
proteneer: `AlignedLigandResampler` would also need to know the water coords to generate correct samples, and the `x_solvent` changes on each frame.

md/moves.py
proteneer: willfix

md/moves.py
proteneer: willfix

tests/test_enhanced_sampling.py
proteneer: willfix

tests/test_enhanced_sampling.py
maxentile: Yes, it would have a `.batch_propose(x)` method, exactly as implemented here! Just it would be its own object that could "lend" out its `batch_propose` method to a `MultipleTryMetropolis` constructor, rather than being a definition that lives inside a `MultipleTryMetropolis` subclass.

proteneer: ah okay - I think I see what you're saying. I can probably make that work.

maxentile: (The same ligand-alignment functionality is used also in a postprocessing step you had implemented -- seems like the same functionality could be defined once and used in both places.)

md/moves.py
maxentile: Based on discussion this morning, my suggestion that implicitly "lambda(x, y) = 1 / proposal_pdf(x, y)" was incorrect -- this is not implied by the current implementation of `w(x, y)`. Will re-visit this more carefully, to confirm that the `lambda(x, y)` implied by the implementation is symmetric!

proteneer:
I've made some changes to this PR after some offline discussions with @maxentile.

I've implemented a `ReferenceMTMMove` class that accepts the full, log'd, signature for `pi(x)`, `Q(x,y)`, and `lambda(x,y)`, as well as the proposal function.
I've also modified the code to take deterministic keys when sampling, so that the `acceptance_probability(x, K, key)` method now takes in an extra PRNG key (using jax) such that we can track the state of the RNG in a much cleaner way.
I've also implemented an `OptimizedMTMMove` class that takes in simplified set of arguments, namely a `log_weight_fn` and a proposal function.
I then compare the results of the `OptimizeMTMMove::acceptance_probability` with the results of `ReferenceMTMMove::acceptance_probability` to ensure that I get back the same samples, and acceptance probabilities, when presented with identical keys.
The `OptimizedMTMMove` is about 10x faster than the `ReferenceMTMMove`.
I've also modified the condensed phase to proposal samples from a `decharged` potential, and whose conformations are then used for a `charged` potential to better mimic real world use cases.

md/enhanced.py
```diff
@@ -351,3 +351,171 @@ def sample_from_log_weights(weighted_samples, log_weights, size):
...
+    equil_schedule = np.ones(num_steps) * lamb
+    equil_ctxt.multiple_steps(equil_schedule)
+
+    x0 = equil_ctxt.get_x_t()
+
+    # (ytz): This has to be zeros_like for now since if we freeze ligand
+    # coordinates it would start to move during rejected moves.
+    # v0 = np.zeros_like(x0)
+    v0 = equil_ctxt.get_v_t()
+    v0[-15:] = 0
```
proteneer, line 454: note to self: remove -15

md/moves.py
```diff
@@ -56,3 +68,197 @@ def n_accepted(self):
...
+
+        return after_npt
+
+
+class DeterministicMTMMove:
+    def move(self, xvb: CoordsVelBox) -> CoordsVelBox:
+        self.n_proposed += 1
+        y_proposed, acceptance_probability, sub_key = self.acceptance_probability(xvb.coords, xvb.box, self.rng_key)
+        self.rng_key = sub_key
+        alpha = np.random.rand()
```
maxentile, line 128: Should this line also use / update `self.rng_key`? (`acceptance_probability` will be a deterministic function of `self.rng_key` state, but the outcome of the "biased coin toss" will depend on global numpy rng state)

md/moves.py
```diff
@@ -56,3 +68,197 @@ def n_accepted(self):
...
+    ):
+
+        intg = lib.LangevinIntegrator(temperature, dt, friction, masses, seed)
+        self.integrator_impl = intg.impl()
+        all_impls = [bp.bound_impl(np.float32) for bp in ubps]
+
+        bond_list = get_bond_list(ubps[0])
+        group_idxs = get_group_indices(bond_list)
+
+        barostat = lib.MonteCarloBarostat(len(masses), pressure, temperature, group_idxs, barostat_interval, seed + 1)
```
maxentile, line 95: May need to handle the `seed=None` case earlier, to avoid `None + 1`

md/moves.py
```diff
@@ -56,3 +68,197 @@ def n_accepted(self):
...
+        self.bound_impls = all_impls
+        self.barostat_impl = barostat_impl
+        self.lamb = lamb
+        self.n_steps = n_steps
+
+    def move(self, x: CoordsVelBox):
+        # note: context creation overhead here is actually very small!
+        ctxt = custom_ops.Context(
+            x.coords, x.velocities, x.box, self.integrator_impl, self.bound_impls, self.barostat_impl
+        )
```
maxentile, line 107: Nice! Yeah, I think we had measured this to be ~0.2 ms  -- negligible when `n_steps` is 100s of MD steps. May want to print a warning in the constructor if `n_steps` is very small (say, < 10?)

maxentile (approved):

ff/params/readme.md
```diff
@@ -1,3 +1,4 @@
 * smirnoff_1_1_0_am1bcc.py -- "bond charge corrections"
 * smirnoff_1_1_0_ccc.py -- "correctable charge corrections"
-* smirnoff_1_1_0_sc.py -- "simple charge"
\ No newline at end of file
+* smirnoff_1_1_0_sc.py -- "simple charge"
+* smirnoff_1_1_0_test_sampling.py -- "decharged model with LJ adjustments for enhanced sampling tests"
```
maxentile, line 4:
```suggestion
```

Can delete reference to new file, deleted in https://github.com/proteneer/timemachine/pull/524/commits/5ee7dd87a9c4e6e5672140e60d2d3a2c42aed92b

slow_tests/test_mtm_condensed.py
```diff
@@ -0,0 +1,200 @@
...
+from matplotlib import pyplot as plt
+
+
+def get_ff_am1cc():
+    ff_handlers = deserialize_handlers(open("ff/params/smirnoff_1_1_0_ccc.py").read())
+    ff = Forcefield(ff_handlers)
+    return ff
+
+
+def test_condensed_phase_mtm():
```
badisa, line 36: This test claims to run 499k steps, is that necessary?

proteneer:
> This test claims to run 499k steps, is that necessary?
This test is quite exhaustive - lower # of steps have had convergence issues

PR #525, opened 2021-11-01
Horrible hack example

badisa:


md/enhanced.py
```diff
@@ -445,23 +442,29 @@ def generate_solvent_phase_samples(
...
-        if mc_mover is not None:
-            xvb = mc_mover.move(CoordsVelBox(x_t, v_t, box))
-            ctxt.set_x_t(xvb.coords)
-            # don't have setters for x_t and box right now
-            np.testing.assert_array_equal(xvb.velocities, v_t)
-            np.testing.assert_array_equal(xvb.box, box)
-
-    return np.ndarray(all_xs)
+
+    class Hackery:
```
badisa, line 446: This is a horrible hack, but it demonstrates what @proteneer was trying to avoid.

We have to keep all of the impls referenced somewhere, else GC cleans them up and the context object seg faults.

proteneer:
yeah, this is pretty hackery

proteneer:
also known as a functor

badisa:
Discussed offline why this sort of thing is necessary

PR #526, opened 2021-11-04
Feature/queue all rabfe tasks

badisa:


maxentile (approved):
Looks good -- modifies `examples/validate_relative_binding.py` to use distance-based instead of smarts-based core restraints, avoid unnecessary accumulation of results in memory, add a function that returns a dictionary of futures, and launch more futures at once.

Clarification: Wasn't immediately clear to me whether this queues up all independent simulation tasks (all windows from all stages from all molecules), or if there are any remaining synchronization barriers (perhaps between stages).

badisa:
> Wasn't immediately clear to me whether this queues up all independent simulation tasks (all windows from all stages from all molecules), or if there are any remaining synchronization barriers (perhaps between stages).

@maxentile As long as the implementation in the client doesn't impose any constraints on the length of the queue or have its own synchronization barriers, this should allow all tasks to be queued. Currently all of our clients are implemented in a way as to allow infinite queuing (up to memory/disk limitations).

maxentile:
Nice -- thank you for clarifying!

examples/validate_relative_binding.py
```diff
@@ -333,81 +327,85 @@ def predict_dG(blocker, mol):
...
-            solvent_decouple_params,
-            solvent_decouple_model,
-            solvent_decouple_futures,
-        ) = binding_model_solvent_decouple.simulate_futures(
-            ordered_params, mol, solvent_x0, solvent_box0, prefix="solvent_decouple_" + mol_name + "_" + str(epoch)
-        )
 
+        suffix = f"{mol_name}_{epoch}"
+
+        return {
```
proteneer, line 347: No strong preference but I've seen @maxentile use a NamedTuple for these types of things. Do we have a preference?

proteneer (approved):

examples/validate_relative_binding.py
badisa: I would like to encapsulate all of these models into a larger RABFE model, as there is a lot of code in this script that seems specific to setting up and running RABFE simulations. This sort of thing should be a lot more hidden than it is.

proteneer: Makes sense to me - thanks.

PR #528, opened 2021-11-08
Feature/#527 filter frames

badisa:


maxentile (approved):
Looks good!

fe/frames.py
```diff
@@ -0,0 +1,19 @@
+from typing import List, Tuple, Generator
+
+
+FrameIterator = Generator[Tuple[int, "SimulationResult"], None, None]
```
proteneer, line 4: does `SimulationResult` here have to be a string? Is it a class name we can import?

badisa: It can be, but we currently have two SimulationResult objects. One in `fe/estimator.py` and one in `fe/estimator_abfe.py`, so I wanted to avoid locking into one of them. Right now it doesn't rely on the actual object, but going forward we might want to expect a certain interface.

PR #531, opened 2021-11-15
Swap default and override OS-dependent parameter values in barostat test

mcwitt:
Currently, we use parameter values that allow tests to pass on Ubuntu 18.04 in `test_barostat_is_deterministic`, with an override exception triggered when the OS is Ubuntu 20.04.

This PR swaps the default and override values so that if the OS is unrecognized, we use the defaults for Ubuntu 20.04. This allows tests to pass (with a warning) on some other recent OSes (e.g. NixOS 21.11).

### DIFF
--- a/tests/test_barostat.py
+++ b/tests/test_barostat.py
@@ -187,15 +187,15 @@ def test_barostat_is_deterministic():
 
     # OpenEye's AM1 Charging values are OS platform dependent. To ensure that we have deterministic values
     # we check against our two most common OS versions, Ubuntu 18.04 and 20.04.
-    box_vol = 26.711716908713402
+    box_vol = 26.869380588831582
     lig_charge_vals = np.array(
-        [1.4572377542719206, -0.37011462071257184, 1.1478267014520305, -4.920166483601927, 0.16985194917937935]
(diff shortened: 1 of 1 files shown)


maxentile (approved):
Looks good!

Question: If supporting more than 2-3 versions, should we:
1. match exactly the behavior expected per platform? (add a `box_vol_dict`, and assert `computed_box_vol == box_vol_dict[platform_version]`)
2. match one of the expected behaviors (add a list of `valid_box_mols`, and assert `computed_box_vol in valid_box_vols`)

Option 2 would miss if the behaviors of ubuntu 20.04 and 18.04 swapped for some reason

Alternately, we could refactor the barostat test to not depend on OpenEye (i.e. hard-code the ligand charges within this test to some constants), and test for platform-specific AM1 behavior in a separate test.

mcwitt:
> Question: If supporting more than 2-3 platform versions

My thought was that this would be more of a hack to allow the tests to pass on some unsupported platforms that we might be using for local development, rather than adding to the list of supported platforms.

> Alternately, we could refactor the barostat test to not depend on OpenEye

This sounds like an ideal long-term option to me. I wonder if it'd even make sense to omit the platform-specific AM1 test, since that sounds like more of a test of openeye?

PR #534, opened 2021-11-16
Add config for clang-format, pre-commit

mcwitt:
- Adds configuration for `clang-format`, chosen to be fairly close to the existing style (config options are documented [here](https://clang.llvm.org/docs/ClangFormatStyleOptions.html))
- Adds [pre-commit](https://pre-commit.com/) config to run `clang-format` in addition to `black` and misc fixups with a commit hook
  * this provides a convenient way to format and apply fixups to modified files with a single command: `pre-commit run`
  * to use as a pre-commit hook, run `pre-commit install` in the repo (can remove with `pre-commit uninstall`)
- Modifies CI to run `pre-commit` instead of `black`. This runs all pre-commit checks (including `clang-format` and `black`) and fails if any changes are needed (the diff of the required change is printed in the CI log, see [example](https://jenkins.relaytxinternal.com/blue/organizations/jenkins/timemachine/detail/PR-534/23/pipeline))
- Applies formatting/fixup changes to all matched files with `pre-commit run --all-files`


timemachine/cpp/src/hilbert.h
```diff
@@ -32,14 +32,14 @@ typedef uint32_t halfmask_t;
  * Inputs:
  *  nDims:      Number of coordinate axes.
  *  nBits:      Number of bits per axis.
- *  index:      The index, contains nDims*nBits bits (so nDims*nBits must be <= 8*sizeof(bitmask_t)).
- * Outputs:
- *  coord:      The list of nDims coordinates, each with nBits bits.
- * Assumptions:
- *      nDims*nBits <= (sizeof index) * (bits_per_byte)
+ *  index:      The index, contains nDims*nBits bits (so nDims*nBits must be <=
+ * 8*sizeof(bitmask_t)). Outputs: coord:      The list of nDims coordinates,
```
badisa, line 36: Breaking up comments like this isn't great, is there a way to prevent that?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,23 @@
+repos:
```
badisa, line 1: If we are going to add pre-commit, we will need to add a check to the CI to ensure that all of checks pass before running tests.

Also addition of pre-commit as a requirement.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,23 @@
...
+    -   id: black
+        exclude: '^(ff/params/|parallel/grpc/)'
+        args: [--line-length=120]
+
+-   repo: https://github.com/pre-commit/mirrors-clang-format
+    rev: v13.0.0
+    hooks:
+    -   id: clang-format
+        include: '^timemachine/cpp/src/'
+        exclude: '^timemachine/cpp/src/jitify.hpp'
```
badisa, line 23: We might also want to exclude the hilbert.c/.hpp files if the intention is to avoid modifying vendored code

badisa:
The style seems fine by me, besides breaking comments into multiple lines

Makefile
```diff
@@ -3,7 +3,6 @@ MKFILE_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 CPP_DIR := $(MKFILE_DIR)timemachine/cpp/
 INSTALL_PREFIX := $(MKFILE_DIR)timemachine/
 PYTEST_CI_ARGS := --cov=. --cov-report=term-missing
-BLACK_FLAGS := --extend-exclude 'ff/params/|parallel/grpc/' --line-length 120 .
```
badisa, line 6: With this removed should remove the `make fmt` command

.pre-commit-config.yaml
mcwitt: Good call; this is fixed now (rebased to keep formatting changes in one commit)

timemachine/cpp/src/hilbert.h
mcwitt: good catch, I hadn't noticed these issues. `hilbert.*` are no longer being formatted, but I'll look for other cases and see if we can disable comment formatting

mcwitt: I didn't see any other cases of this, and also don't see any obvious way to disable formatting in comments. If this comes up in the future, we could probably disable formatting locally with a pragma if there's no other workaround.

badisa: Looking at the clang-format syntax here: https://clang.llvm.org/docs/ClangFormatStyleOptions.html

I think we want the additional flag:
```
ColumnLimit: 120
```

mcwitt: good call, added to e98f119

mcwitt: Just a note that we might want to revisit `ColumnLimit`. Setting to `120` seems to result in some pretty long argument list lines, ex. https://github.com/proteneer/timemachine/pull/534/files#diff-87cdbf91d94f4b5f6791e9b057b3934bd68f1737aa57b7134c02a61f83c89bc2R25

Makefile
mcwitt: good call, I missed this. Fixed in c9bb1e1.

.pre-commit-config.yaml
mcwitt: Added to requirements in 8f43657. Added to CI in 2097e2d.

mcwitt:
> The style seems fine by me, besides breaking comments into multiple lines

This should be fixed now by adding the option `ReflowComments: false`

proteneer:
lgtm - but did we want to merge this first before the other PRs?

maxentile (approved):
Looks great!

badisa (approved):

timemachine/cpp/src/integrator.hpp
```diff
@@ -7,53 +7,31 @@ namespace timemachine {
...
 
     virtual void step_fwd(
-        double *d_x_t,
-        double *d_v_t,
-        unsigned long long *d_du_dx_t,
-        double *d_box_t_,
-        cudaStream_t stream
-    ) override;
-
+        double *d_x_t, double *d_v_t, unsigned long long *d_du_dx_t, double *d_box_t_, cudaStream_t stream) override;
```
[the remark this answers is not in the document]
mcwitt: I also prefer Black's behavior here; I just did a little more looking and found [ExperimentalAutoDetectBinPacking](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#:~:text=experimentalautodetectbinpacking), which seems like it might be exactly what we want. But it comes with a disclaimer:

> NOTE: This is an experimental flag, that might go away or be renamed. Do not use this in config files, etc. Use at your own risk.

It might still be worth trying out?

timemachine/cpp/src/nonbonded.hpp
```diff
@@ -76,32 +75,26 @@ class Nonbonded : public Potential {
...
     void set_nblist_padding(double val);
     void disable_hilbert_sort();
 
     Nonbonded(
-        const std::vector<int> &exclusion_idxs, // [E,2]
-        const std::vector<double> &scales, // [E, 2]
-        const std::vector<int> &lambda_plane_idxs, // N
+        const std::vector<int> &exclusion_idxs,     // [E,2]
+        const std::vector<double> &scales,          // [E, 2]
+        const std::vector<int> &lambda_plane_idxs,  // N
```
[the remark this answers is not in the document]
proteneer: +1 to this too

timemachine/cpp/src/integrator.hpp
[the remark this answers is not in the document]
maxentile: Hmm

>Each call can be bin-packed, one-per-line or inconclusive. If it is inconclusive, e.g. completely on one line, but a decision needs to be made, clang-format analyzes whether there are other bin-packed cases in the input file and act accordingly.

doesn't sound ideal to me -- would prefer to have the formatter apply an unambiguous rule (and ideally allow for a manual toggle), but would not want the formatter to make complicated decisions based on the rest of the file.

I'd be in favor of keeping this as you currently have it!

PR #535, opened 2021-11-17
Task/#530 refactor rabfe model

badisa:
The primary goal of this PR is to isolate a lot of the implementation details of the RABFE protocol from the `validate_relative_binding.py` script as it makes it more difficult to re-implement elsewhere without a lot of copy-pasta. Not entirely happy with it yet, but wanted to get some feedback on the approach.

Verified similar results using a compound from hif2a, but not extensive at this point.


examples/validate_relative_binding.py
```diff
@@ -247,215 +244,76 @@ def fn(*args, **kwargs):
...
-        temperature,
-        pressure,
-        dt,
-        cmd_args.num_complex_equil_steps,
-        cmd_args.num_complex_prod_steps,
-        frame_filter=frame_filter,
-    )
 
-    binding_model_complex_decouple = model_rabfe.RelativeBindingModel(
+    rabfe_model = model_rabfe.RABFEModel(
```
badisa, line 248: This RABFEModel is constructed using all of the static variables of the model, most notably: the blocker, the complex config and the solvent config.

examples/validate_relative_binding.py
```diff
@@ -247,215 +244,76 @@ def fn(*args, **kwargs):
...
-        )
-        rabfe_result.log()
-        dG_err = np.sqrt(
-            dG_complex_conversion_error ** 2
-            + dG_complex_decouple_error ** 2
-            + dG_solvent_conversion_error ** 2
-            + dG_solvent_decouple_error ** 2
+        # generate the core_idxs
+        core_idxs = setup_relative_restraints_by_distance(mol, blocker)
+        results = rabfe_model.simulate_futures(
```
badisa, line 294: The signature here is different from the other models which I don't love

examples/validate_relative_binding.py
```diff
@@ -247,215 +244,76 @@ def fn(*args, **kwargs):
...
+            ordered_params,
+            mol,
+            core_idxs,
+            suffix,
         )
-
-        return rabfe_result.dG_bind, dG_err
+        # ff_params, mol, core_idxs, complex_coords, complex_box, solvent_coords, solvent_box, suffix
```
badisa, line 300:
```suggestion
```

examples/validate_relative_binding.py
```diff
@@ -471,9 +329,9 @@ def predict_dG(results: dict):
                     assert 0, "Unknown property units"
             except Exception as e:
                 print(f"Unable to find property {cmd_args.property_field}: {e}")
-        print(f"Epoch: {epoch}, Processing Mol: {mol_name}, Label: {label_dG}")
+        print(f"Epoch: {epoch}, Processing Mol: {mol_name}, Label: {label_dG}, Seed {rabfe_future.seed}")
```
badisa, line 331: We now log the seed that the simulation was run with, allowing for reproducing the simulations

fe/model_rabfe.py
```diff
@@ -579,3 +606,326 @@ class RelativeBindingModel(RelativeModel):
...
+        )
+
+        return rabfe_result.dG_bind, dG_err
+
+    def predict(self, ff_params, mol, core_idxs, suffix) -> Tuple[float, float]:
+        futures = self.simulate_futures(ff_params, mol, core_idxs, suffix)
+        results = self.predict_from_futures(futures)
+        return results
+
+    def set_complex_conversion_schedule(self, schedule: List[int]):
```
badisa, line 897: All these setters could be done with `__getattr__` magic, but as a first pass just expanded them out.

tests/test_endpoint_correction.py
```diff
@@ -170,11 +170,21 @@ def test_endpoint_correction():
     k_translation = 200.0
 
     results = []
-    for k_rotation in [0.0, 50.0, 1000.0]:
+    for i, k_rotation in enumerate([0.0, 50.0, 1000.0]):
 
         lhs_du, rhs_du, rotations, translations = endpoint_correction.estimate_delta_us(
             k_translation, k_rotation, core_idxs, core_params, beta, lhs_xs, rhs_xs
         )
+        if i == 0:
```
badisa, line 178: Verify that the endpoint correction always produces the same results given the same inputs

tests/test_rabfe.py
```diff
@@ -22,14 +28,77 @@
 
 
 class TestRABFEModels(TestCase):
-    def test_predict_complex_decouple(self):
-        """Just to verify that we can handle the most basic complex decoupling RABFE prediction"""
+    def test_predict_rabfe_model(self):
```
badisa, line 31: This test is pretty slow, 80s. The other two tests in this file are 50s and 10s, and both are now covered implicitly by this test. Moving it to slow tests doesn't seem to make an impact on the time. Not sure what to do yet, but the test seems like a good one if we are to rely on the RABFE protocol

tests/test_rabfe.py
```diff
@@ -22,14 +28,77 @@
...
+        model.set_solvent_decouple_equil_steps(num_steps)
+        model.set_solvent_decouple_prod_steps(num_steps)
+        model.set_solvent_conversion_equil_steps(num_steps)
+        model.set_solvent_conversion_prod_steps(num_steps)
+
+        core_idxs = setup_relative_restraints_by_distance(mol_a, mol_b)
+        ordered_params = hif2a_ligand_pair.ff.get_ordered_params()
+        with temporary_working_dir() as temp_dir:
+            dG, dG_err = model.predict(ordered_params, mol_b, core_idxs, "suffix")
+            np.testing.assert_almost_equal(dG, 572.3966117085292)
```
badisa, line 93: Now getting deterministic model predictions from RABFE

tests/test_rabfe.py
```diff
@@ -87,14 +156,9 @@ def test_predict_complex_decouple(self):
...
-    def test_predict_complex_conversion(self):
+    def test_predict_solvent_conversion(self):
         """Just to verify that we can handle the most basic complex conversion RABFE prediction"""
-        complex_system, complex_coords, _, _, complex_box, complex_topology = builders.build_protein_system(
-            os.path.join(DATA_DIR, "hif2a_nowater_min.pdb")
-        )
-
-        # build the water system
-        solvent_system, solvent_coords, solvent_box, _ = builders.build_water_system(4.0)
+        solvent_system, solvent_coords, solvent_box, solvent_topology = builders.build_water_system(4.0)
```
badisa, line 161: Switched to solvent system as it is a bit faster to run

timemachine/cpp/src/kernels/k_neighborlist.cuh
```diff
@@ -234,7 +234,7 @@ void __global__ k_find_blocks_with_ixns(
 
     if (include_col_block) {
 
-        // Compute center of column box and extent coords.
+        // Compute center of column block and extent coords.
```
badisa, line 237: Some imprecise comments @mcwitt pointed out while going over nonbonded, could be pulled out into separate PR

fe/endpoint_correction.py
```diff
@@ -18,12 +18,11 @@ def exp_u(rotation, k, beta):
 def sample_multiple_rotations(k, beta, size):
     num_batches = 500
     batch_size = 10000
-
+    state = np.random.RandomState(2021)
```
mcwitt, line 21: Will we need to update this on Jan 1st? :stuck_out_tongue:

fe/model_rabfe.py
mcwitt: Actually, could this be a `dataclass`? I'm not sure I'm reading it correctly yet, but it seems like there's not a lot of internal mutable state?

fe/endpoint_correction.py
badisa: We generally have the convention of setting the random seed as the year we made the change, doesn't really matter as long as the function is deterministic

fe/model_rabfe.py
badisa: This could definitely be a dataclass

PR #536, opened 2021-11-17
Makes Fixed Point fix broader

badisa:
* In testing from 7.5 -> 8.6 this fix is crucial for performance.

Please make sure to test this on your machine to ensure you don't see a performance loss. This real to int64 change has been important on the following GPUs I have tested:

* T4
* RTX 2080
* A4000
* RTX 3090
* A10G

### DIFF
--- a/timemachine/cpp/src/kernels/k_fixed_point.cuh
+++ b/timemachine/cpp/src/kernels/k_fixed_point.cuh
@@ -17,7 +17,7 @@ RealType __device__ __forceinline__ FIXED_TO_FLOAT_DU_DP(unsigned long long v) {
 // (ytz): courtesy of @scottlegrand/NVIDIA, even faster conversion
 // This was original a hack to improve perf on Maxwell, that is now needed for Ampere
 long long __device__ __forceinline__ real_to_int64(float x) {
-#if __CUDA_ARCH__ == 860 || __CUDA_ARCH__ == 750
+#if __CUDA_ARCH__ >= 750
   float z = x * (float)0x1.00000p-32;
   int hi = __float2int_rz( z );                         // First convert high bits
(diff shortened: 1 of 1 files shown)


proteneer (commented):
I think we may need to exclude 80 from this, since A100 has full double precision capabilities

timemachine/cpp/src/kernels/k_fixed_point.cuh
```diff
@@ -17,7 +17,7 @@ RealType __device__ __forceinline__ FIXED_TO_FLOAT_DU_DP(unsigned long long v) {
 // (ytz): courtesy of @scottlegrand/NVIDIA, even faster conversion
 // This was original a hack to improve perf on Maxwell, that is now needed for Ampere
 long long __device__ __forceinline__ real_to_int64(float x) {
-#if __CUDA_ARCH__ == 860 || __CUDA_ARCH__ == 750
+#if __CUDA_ARCH__ >= 750
```
proteneer, line 20: Okay I'm confused, I double checked https://en.wikipedia.org/wiki/CUDA

```
T4 is 7.5
RTX 2080 is 7.5
A4000 is 8.6
RTX 3090 is 8.6
A10G is 8.6
```

Are you saying that A10G performs better with this *disabled*?

badisa: No, it performs better with it enabled. I just wanted to move away from two versions explicitly stated. But need to benchmark an A100 to be sure that this works, for some reason I thought we had benchmarked an 8.0 card already

proteneer: IIRC A100 performs *better* with this *disabled* since it has full double precision ALUs.

badisa: Been having a hard time getting A100s to test this on

Issue #533, opened 2021-11-16
Summed Potential


badisa:
### Background

A big performance bottle neck is copying data between the CPU and the GPU. Currently if we want to use the GPU potentials as a bound impl (IE as a python object we call `execute` on) we must sum the resulting values in python, resulting in N (the number of potentials) copies from the GPU to the CPU. We would like to instead combine a set of potentials into a potential that will perform the summing on the GPU and only copy data to the CPU once.

### Design Considerations

* Should function like any other potential, and be constructed from a set of other potentials.

badisa:
One good place to use the new summed potential would be in our minimizer code: https://github.com/proteneer/timemachine/blob/master/md/minimizer.py#L63

proteneer:
Resolved by #537

PR #537, opened 2021-11-18
Compute sum of potentials efficiently

mcwitt:
Addresses #533 

- Adds a new C++ custom op, `SummedPotential: Potential`
- Adds a version of `construct_differentiable_interface` (in `fe.functional`) that uses `SummedPotential` to do the sum over component potentials efficiently in C++
- Adds a test asserting that the new version produces equivalent results to the original

#### TODO
- [x] wrap with PyBind
- [x] add tests


proteneer:
I think you probably still want to use vector of potentials (not bound potentials) and compute the d_p offsets separately. While it's technically correct, it sort of breaks the API in that the `d_p` is no longer being used, and this SummedPotential is more like BoundSummedPotential (which again, could implemented slightly differently).

badisa:
@proteneer I had pointed @mcwitt to make it a vector of bound potentials. I am unclear of the utility of maintaining the API of using `d_p` as it is only called from the `BoundPotential` and always passes its own parameters. Combining the parameters from an arbitrary set of potentials and maintaining state seems more fragile than leveraging a pattern that we already use in the context.

proteneer:
Would it make more sense then for the `SummedPotential` class to inherit from the `BoundPotential` class instead? This way we wouldn't have to worry about a dead parameter `d_p` being present in `execute_device`

badisa:
I think I suggested that so we would be able to call `execute_selective`, but looking again we could just add that to the wrapping of the BoundPotentials. Though perhaps I am missing some subtlety

timemachine/lib/potentials.py
```diff
@@ -259,3 +259,16 @@ def unbound_impl(self, precision):
...
+
+
+class SummedPotential:
+    def __init__(self, potentials, params):
+        self._impl = custom_ops.SummedPotential(potentials, params)
+
+        # NOTE: the following is necessary to ensure that we maintain
+        # a reference to the wrapped Potential objects on the Python
+        # side; otherwise they may be freed prematurely
+        self._potentials = potentials
```
mcwitt, line 271: I found I needed this hack to avoid a segfault when calling `execute_device` on the wrapped `SummedPotential`. Without this, I think what's happening is that the wrapped `Potential`s are being freed prematurely (when there are no more references in Python). Curious if we've encountered this issue before and have a standard way of dealing with it?

tests/test_free_energy.py
```diff
@@ -373,3 +373,31 @@ def low_dim_f(perturb: float) -> float:
...
+    ff_params = hif2a_ligand_pair.ff.get_ordered_params()
+    unbound_potentials, sys_params, _, coords = hif2a_ligand_pair.prepare_vacuum_edge(ff_params)
+    box = np.eye(3) * 100
+    lam = 0.5
+
+    for precision in [np.float32, np.float64]:
+        U = construct_differentiable_interface(unbound_potentials, precision)
+        U_cpp = construct_differentiable_interface_cpp(unbound_potentials, sys_params, precision)
+        args = (coords, sys_params, box, lam)
+        assert np.allclose(U(*args), U_cpp(*args), rtol=1e-6)
```
mcwitt, line 392: Would we prefer `np.testing.assert_arrays_equal` here if possible?

tests/test_free_energy.py
```diff
@@ -373,3 +373,31 @@ def low_dim_f(perturb: float) -> float:
             for perturb in perturbations:
                 abs_err = check_grad(low_dim_f, grad(low_dim_f), perturb, epsilon=1e-4)
                 assert abs_err < 1e-3
+
+
+def test_consruct_differentiable_interface_cpp():
```
proteneer, line 378: typo in consruct

tests/test_free_energy.py
proteneer: For this test, I'd expect the results here to be bitwise identical.

fe/functional.py
```diff
@@ -50,3 +58,24 @@ def U(coords, params, box, lam):
         return np.sum(np.array([U_i(coords, p_i, box, lam) for (U_i, p_i) in zip(U_s, params)]))
 
     return U
+
+
+def construct_differentiable_interface_cpp(unbound_potentials, params, precision=np.float32):
```
proteneer, line 63: Did we want to rename this to something else? eg. _fast or _sum or _optimized?

tests/test_free_energy.py
proteneer: Actually - this is a little more complicated than I thought, since I think the old code actually does:

`
sum_floats(fixed_to_float(U_fixed(x_i)))
`

Where as the new code does

`
fixed_to_float(sum_ints(U_fixed(x_i)))
`

So I'm not fully sure that this commutes.

fe/functional.py
mcwitt: Thanks, I like all of those better than `_cpp`. If we don't have an existing convention maybe I'll choose `_fast`.

tests/test_free_energy.py
mcwitt: Thanks, fixed in 4f8bedb.

fe/functional.py
mcwitt: Changed in 1592167 (and 1ed2c2b)

tests/test_free_energy.py
mcwitt: Hmm, it does seem to pass with the equality assertion (9a32491). In light of your comment I'm not sure I understand how this works; will look into it more.

proteneer (approved):
lgtm!

timemachine/lib/potentials.py
maxentile: Ahh, good catch.

We have encountered a similar issue before, if you construct a context and don't keep around Python references to its `impl`s. Workarounds like https://github.com/proteneer/timemachine/pull/525#discussion_r740565792 can sometimes be needed.

(In the specific case of contexts though, another workaround is to frequently [construct new contexts](https://github.com/proteneer/timemachine/blob/8b0e4d2281c69b6aba2f62794349fe1e2f7950bf/md/thermostat/moves.py#L14-L27) whenever needed, since context creation overhead in timemachine is quite small.)

badisa (approved):

timemachine/cpp/src/summed_potential.cu
```diff
@@ -0,0 +1,47 @@
...
+namespace timemachine {
+
+SummedPotential::SummedPotential(std::vector<Potential *> potentials, std::vector<int> params_sizes)
+    : potentials_(potentials), params_sizes_(params_sizes) {
+    if (potentials_.size() != params_sizes_.size()) {
+        throw std::runtime_error("number of potentials != number of parameter sizes");
+    }
+};
+
+void SummedPotential::execute_device(
```
badisa, line 13: Do you know how much faster this is than the previous version?

mcwitt: I just ran a timing test (code [here](https://gist.github.com/mcwitt/21e992891eb2e51081fe974743503fef)). It looks like there's a modest ~20% speedup for computing U, and a ~50% speedup for computing the gradient:

```
===============================================================================
construct_differentiable_interface (precision=<class 'numpy.float32'>)
  construct: 12.506331214000966
  evaluate U  (1000 iters): 0.7602241589993355
  evaluate ∇U (1000 iters): 9.319278081000448
===============================================================================
construct_differentiable_interface_fast (precision=<class 'numpy.float32'>)
  construct: 12.518712601999141
  evaluate U  (1000 iters): 0.617156143000102
  evaluate ∇U (1000 iters): 4.233765714001493
```

proteneer: nice!

mcwitt: The above timing was only computing dU/dx. I updated to also compute dU/dp and dU/dl ([here](https://gist.github.com/mcwitt/4e65ef5e9356e264d8c450d45f329934)). The relative timing is pretty much unchanged:

```
===============================================================================
construct_differentiable_interface (precision=<class 'numpy.float32'>)
  construct: 12.515916177999316
  evaluate U  (1000 iters): 0.7570085630013637
  evaluate ∇U (1000 iters): 11.697148505998484
===============================================================================
construct_differentiable_interface_fast (precision=<class 'numpy.float32'>)
  construct: 12.499262368000927
  evaluate U  (1000 iters): 0.6442218980009784
  evaluate ∇U (1000 iters): 5.53595667800073
```

tests/test_free_energy.py
```diff
@@ -373,3 +373,31 @@ def low_dim_f(perturb: float) -> float:
...
+        np.testing.assert_array_equal(U(*args), U_ref(*args))
+
+        argnums = (0, 1, 3)
+        grad_U_ref = grad(U_ref, argnums=argnums)(*args)
+        grad_U = grad(U, argnums=argnums)(*args)
+
+        np.testing.assert_array_equal(grad_U[0], grad_U_ref[0])
+
+        for dU_dp, dU_dp_ref in zip(grad_U[1], grad_U_ref[1]):
+            np.testing.assert_array_equal(dU_dp, dU_dp_ref)
```
maxentile, line 401: Wouldn't hurt to add an assertion that `len(grad_U) == len(grad_U_ref)` (since this `zip` could in principle compare just `min(len(grad_U), len(grad_U_ref))` items)

tests/test_free_energy.py
```diff
@@ -373,3 +373,31 @@ def low_dim_f(perturb: float) -> float:
             for perturb in perturbations:
                 abs_err = check_grad(low_dim_f, grad(low_dim_f), perturb, epsilon=1e-4)
                 assert abs_err < 1e-3
+
+
+def test_construct_differentiable_interface_fast():
+    """Assert that the computation of U and its derivatives using the
+    C++ code path produces equivalent results to the doing the
```
maxentile, line 380:
```suggestion
    C++ code path produces equivalent results to doing the
```

maxentile (approved):
Looks great -- and obtains a nice speed-up!

tests/test_free_energy.py
mcwitt: good call, thanks! Added in 5641ab4.

PR #539, opened 2021-11-22
Upgrade jax and jaxlib

mcwitt:
The original motivation for the upgrade was to pull in this (non-essential) upstream fix: https://github.com/google/jax/pull/7606. There was some minor maintenance needed to get `py_compile` and tests passing:

- replace `jax.partial` with `functools.partial`, following [this point](https://github.com/google/jax/blob/main/CHANGELOG.md#:~:text=jax.partial%2C%20and%20jax.lax.partial%20were%20accidental%20exports%20that%20have%20now%20been%20removed.%20use%20functools.partial%20from%20the%20python%20standard%20library%20instead.) in the JAX changelog
- use `defjvp` instead of `defvjp` in a custom primitive implementation (`mybar`), since the latter was removed in this [commit](https://github.com/google/jax/commit/35fcf2e2fd5b4c56cbb591f4c8bf01222a23dfe5). It seems JAX is still able to derive reverse-mode autodiff from this, and the results and performance of `grad(mybar)` are unchanged.

#### Todo
- [ ] Update environment in CI


fe/loss.py
```diff
@@ -21,7 +21,7 @@ def mybar_impl(w):
     return A
 
 
-def mybar_vjp(g, w):
+def mybar_jvp(g, w):
```
proteneer, line 24: In general, a vjp is not implemented in the same way as a jvp. In particular, a vjp is a vector jacobian product, and a jvp is a jacobian vector product. The former is a left multiply, as shown here, where as the latter is a right multiply.

proteneer: Furthermore, `g` is usually the adjoint, i.e. the derivative of the loss with respect to the output for the underlying function. Wher as in jvps, the gradient is the derivative of the input of the underlying function with respect to the parameters.

mcwitt: Admittedly I was not very sure of what I was doing here.. From what I can gather the `defvjp` method we're using here was removed from jax ([commit](https://github.com/google/jax/issues/4918#issuecomment-728689031)). For untraceable code, I did find a [suggestion](https://github.com/google/jax/issues/4918#issuecomment-728689031) to define both jvp and a "transposition rule", but wasn't able to figure out the latter yet. However, I was surprised to find that `grad` still worked (with the same output and apparently performance) when I simply changed `jvp` to `vjp` :thinking: . I'll try to understand this better before moving this PR out of "draft" status.

proteneer: Consider a function `f(g(h(x)))`, where `f`,`g`,`h` are vector valued functions, and `x` is a vector - its derivative is `df/dg * dg/dh * dh/dx`and `*` denotes a matrix dot product. One way to evaluate this is:

`d/dx f = df/dg * (dg/dh * (dh/dx))`, i.e. inside out. This is "forward-mode", where `jv-product` here is comprised of: `j = dg/dh`, `v = dh/dx`, so `dg/dx = jv`
Then we apply the same `jvp` rule once more, this time setting `v` to the result of the above output:
`df/dg * dg/dx`, with `j=df/dg` and `v=dg/dx`. Here, the `v` is the intermediate that is carried along, or propagated forwards. In this case, `v` corresponds to the derivative of the intermediate with respect to the model parameters.

An alternative way is to evaluate it is outside in:

`d/dx f = ((df/dg) * dg/dh) * dh/dx`

Here we use reverse-mode, via `vj-products`, where `v=df/dg` and `j=dg/dh` in the first pass, and in the second pass `v=df/dh` and `j=dh/dx`. In this notation, the "adjoint" is refers to the `v` and is typically carried along (or back-propagated). In other words, in backwards mode, the `v` is the derivative of the final output "loss" with respect to the intermediate.

There are certain cases where the `vjp` and the `jvp` are identical. Eg. when the `j` is a symmetric matrix then any `v` would generate identical results. Or when `v` is a scalar. However, in your case, you simply misnamed the arguments.

`def_vjp` has the signature `foo(adjoint_grad, val)`, where as `def_jvp` has the signature `foo(primals, tangents)`, which is another way of saying `foo(val, grad)`. However, because the primitive `G(w): R^N - >R` emits a scalar valued output, its jacobian `dG/dw` is just a length N vector. It doesn't matter if it is "left multiplied" or "right multiplied" by a scalar.

mcwitt:
The following test fails on my local machine:

```
_____________________________________ test_optimized_MTM _____________________________________
…
>           np.testing.assert_array_equal(ref_yvb.coords, test_y)
E           AssertionError: 
E           Arrays are not equal
E           
E           Mismatched elements: 4 / 8049 (0.0497%)
E           Max absolute difference: 2.77555756e-17
E           Max relative difference: 1.50581625e-16
E            x: array([[-1.37349 , -0.387638, -0.139961],
E                  [-1.374389, -0.331702, -0.056808],
E                  [-1.441192, -0.459462, -0.129689],...
E            y: array([[-1.37349 , -0.387638, -0.139961],
E                  [-1.374389, -0.331702, -0.056808],
E                  [-1.441192, -0.459462, -0.129689],...

tests/test_mtm.py:164: AssertionError
```

fe/loss.py
mcwitt: IIRC, the (offline) resolution of this was that `ad.defjvp` and `ad.defvjp` both take the gradient as the first argument (confusingly different from `custom_jvp` and `custom_vjp`, whose arguments are swapped with respect to each other). As a result the above change is (accidentally) correct.

mcwitt:
> The following test fails on my local machine:

I'm not able to replicate this testing with the standard configuration and after bumping to 0.2.26/0.1.75 (all tests pass).

I will still investigate to figure out which of the two changes made the difference, but I think we could consider merging at this point.

maxentile (approved):
Looks good!

([This](https://github.com/proteneer/timemachine/pull/539#issuecomment-977073882) sounded kind of spooky, but doesn't seem to be an issue anymore.)

mcwitt:
I was able to reproduce the failure https://github.com/proteneer/timemachine/pull/539#issuecomment-977073882 by reverting c6897d7. With the present version, all tests pass.

mcwitt:
After more testing, it looks like test failure seen with 0.2.25/0.1.74 is an issue with `jaxlib`; the failure is reproduced with `jaxlib==0.1.74` and `jax` between 0.2.24 and 0.2.27. From my testing it looks like `jaxlib` between 0.1.70 and 0.1.74 (inclusive) are affected.

PR #541, opened 2021-12-01
[WIP] Test system for SMC

proteneer:
Test case for biphenyl toy-system, which conveniently uses the new differentiable_interface_fast from @mcwitt.

General outline:

1) Generate water samples
2) Generate ligand samples
3) Generate joint samples by randomly, independently sampling from water and ligand samples.
4) NPTMove is used to propagate

### DIFF
--- a/tests/test_smc.py
+++ b/tests/test_smc.py
@@ -145,8 +145,6 @@ def test_smc():
     npt_lamb = 1.0
     seed = int(time.time())
 
-    print(seed)
-
     mover = NPTMove(ubps, npt_lamb, masses, temperature, pressure, n_steps, seed)
 
     U_fn = functional.construct_differentiable_interface_fast(ubps, params)
@@ -156,8 +154,8 @@ def test_smc():
         xvb = CoordsVelBox(xi, np.zeros_like(xi), bi)
         xvb_new = mover.move(xvb)
 
-        # compare delta_Us when coords change
-        print(U_fn(xvb_new.coords, params, xvb_new.box, npt_lamb) - U_fn(xvb.coords, params, bi, npt_lamb))
+        # compare delta_Us when coords/box change (this will large and size extensive, probably not useful)
+        print(U_fn(xvb_new.coords, params, xvb_new.box, npt_lamb) - U_fn(xvb.coords, params, xvb.box, npt_lamb))
 
     # or, compare delta_Us when lambda changes
 
--- a/tests/test_smc.py
+++ b/tests/test_smc.py
@@ -58,15 +58,19 @@ def generate_ligand_samples(num_batches, mol, ff, temperature, seed):
 
 
 def generate_endstate_samples(num_samples, solvent_samples, ligand_samples, ligand_log_weights, num_ligand_atoms):
-    all_xbs = []
+    all_xvbs = []
     for _ in range(num_samples):
         choice_idx = np.random.choice(np.arange(len(solvent_samples)))
         solvent_x = solvent_samples[choice_idx].coords
-        ligand_x = enhanced.sample_from_log_weights(ligand_samples, ligand_log_weights, size=1)[0]
+        solvent_v = solvent_samples[choice_idx].velocities
+        ligand_xv = enhanced.sample_from_log_weights(ligand_samples, ligand_log_weights, size=1)[0]
+        ligand_x = ligand_xv[0]
+        ligand_v = ligand_xv[1]
         combined_x = np.concatenate([solvent_x[:-num_ligand_atoms], ligand_x], axis=0)
+        combined_v = np.concatenate([solvent_v[:-num_ligand_atoms], ligand_v], axis=0)
         combined_box = solvent_samples[choice_idx].box
-        all_xbs.append((combined_x, combined_box))
-    return all_xbs
+        all_xvbs.append((combined_x, combined_v, combined_box))
+    return all_xvbs
 
 
 def test_smc():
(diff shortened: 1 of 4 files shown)


tests/test_smc.py
```diff
@@ -0,0 +1,164 @@
...
+    npt_lamb = 1.0
+    seed = int(time.time())
+
+    mover = NPTMove(ubps, npt_lamb, masses, temperature, pressure, n_steps, seed)
+
+    U_fn = functional.construct_differentiable_interface_fast(ubps, params)
+
+    for xi, bi in all_xbs:
+        # (ytz): note to jfass, change v0 to not zeros_like
+        xvb = CoordsVelBox(xi, np.zeros_like(xi), bi)
```
maxentile, line 154: Something like 

```python
from md.thermostat.utils import sample_velocities
v0 = sample_velocities(masses * unit.amu, temperature)
```

?

tests/test_smc.py
```diff
@@ -0,0 +1,164 @@
...
+
+import jax
+from timemachine.potentials import bonded
+
+from fe import functional
+from ff import Forcefield
+from ff.handlers.deserialize import deserialize_handlers
+
+# (ytz): useful for visualization, so please leave this comment here!
+import asciiplotlib as apl
```
maxentile, line 24: Add `asciiplotlib` to `requirements.txt`?

tests/test_smc.py
```diff
@@ -0,0 +1,164 @@
...
+    fig.show()
+
+    bound_impls = []
+    for u, p in zip(ubps, params):
+        bound_impls.append(u.bind(p).bound_impl(np.float32))
+
+    # propagate with NPTMove
+    n_steps = 1000
+    npt_lamb = 1.0
+    seed = int(time.time())
```
maxentile, line 146: May want to freeze this to a constant in the test?
```suggestion
    seed = 2021
```

tests/test_smc.py
```diff
@@ -0,0 +1,164 @@
...
+        # (ytz): note to jfass, change v0 to not zeros_like
+        xvb = CoordsVelBox(xi, np.zeros_like(xi), bi)
+        xvb_new = mover.move(xvb)
+
+        # compare delta_Us when coords/box change (this will large and size extensive, probably not useful)
+        print(U_fn(xvb_new.coords, params, xvb_new.box, npt_lamb) - U_fn(xvb.coords, params, xvb.box, npt_lamb))
+
+    # or, compare delta_Us when lambda changes
+
+    U_fn = functional.construct_differentiable_interface_fast(ubps, params)
```
maxentile, line 162:
```suggestion
```

(already defined above)

maxentile (commented):
Nice! This will be a highly useful test system.

Clerical note: `tests/test_smc.py` does not immediately seem like a natural place for this code to live. (File currently contains no SMC-specific logic, and `test_smc()` currently contains no explicit assertions. I think we'll also want to be able to import this biphenyl test system in multiple contexts.) Should we move any of the definitions from `tests/test_smc.py` and `tests/test_ligands.py` into more generic locations like `testsystems`, `examples`, or `scripts`?

(We'll likely want to reuse the same system (`u(x, lam)`, `generate_nice_samples_at_lam_equals_one()`, `propagate(x, lam)`) with multiple sampling methods: (1) Null baseline: independent lambda windows. (2) Strong baseline: Hamiltonian replica exchange. (3) SMC variants obtained by toggling settings: with resampling vs. without resampling, adapted protocol vs. fixed protocol, etc.)

proteneer:
> Clerical note: tests/test_smc.py does not immediately seem like a natural place for this code to live. (File currently contains no SMC-specific logic, and test_smc() currently contains no explicit assertions. I think we'll also want to be able to import this biphenyl test system in multiple contexts.) Should we move any of the definitions from tests/test_smc.py and tests/test_ligands.py into more generic locations like testsystems, examples, or scripts?

This wasn't meant to be merged - more of an illustrative example. Let's work offline to see how we can move some of this into a testsystem file.

tests/test_smc.py
proteneer: I think there's some nuance here, with BAOAB I think we'd still need to adjust the velocities by half a step before we can do `set_v_t(sample_velocities())` - I was hoping we can get this detail correct.

tests/test_smc.py
proteneer: Again, not really meant to be merged, but rather an illustrative example. will remove later once the we work out the test case properly.

tests/test_smc.py
proteneer: Will freeze for actual test case - I need to think a bit more carefully about if we want each mover for a given lambda window to restart with the same seed.

tests/test_smc.py
maxentile: Good call.

Illustration of this issue in a 1D system, at a large stepsize: https://gist.github.com/maxentile/1bd50720fe09f0541be0598f2fe37ec9 

A possible workaround appears to be something like `half_step_back_v = equilibrium_v - 0.5 * dt * reduced_force(equilibrium_x) / masses`

**1. _Equilibrium_ distribution**
image
(Note that x and v are independent!)

**2. _Steady-state_ distribution sampled by our Langevin integrator**
image
(Note that x and v marginals are still accurate, but x and v are no longer independent!)

**3. Distribution sampled by running our Langevin integrator for 1 step, _initialized with (x, v) from equilibrium_**
image
(Note that x marginal is now pretty inaccurate!)

**4. Distribution sampled by running our Langevin integrator for 1 step, _initialized with x from equilibrium, v adjusted by half a step_**
image
(Note that x marginal is mostly recovered!)

------------

Other workarounds include:
* Running for more MD steps
* Running at a moderate stepsize

proteneer:
stale!

PR #542, opened 2021-12-01
Implement nonbonded exclusions as summed potential term

mcwitt:
Note: work in progress

Currently exclusions are computed internally in the nonbonded potential ([here](https://github.com/proteneer/timemachine/blob/54068f0e7cdfe063f067f684d6f1dfa3947db996/timemachine/cpp/src/kernels/k_nonbonded.cuh#L758)). This PR will move computation of the exclusion terms into a separate potential, and recover the existing nonbonded potential as a sum (using `SummedPotential` added in #537). This will increase flexibility and modularity of the code.

### DIFF
--- a/timemachine/cpp/src/kernels/k_nonbonded.cuh
+++ b/timemachine/cpp/src/kernels/k_nonbonded.cuh
@@ -763,7 +763,6 @@ void __global__ k_nonbonded_exclusions(
     const double *__restrict__ dp_dl,
     const double *__restrict__ coords_w, // 4D coords
     const double *__restrict__ dw_dl,    // 4D derivatives
-    const double lambda,
     const int *__restrict__ exclusion_idxs, // [E, 2] pair-list of atoms to be excluded
     const double *__restrict__ scales,      // [E]
     const double beta,
@@ -841,7 +840,6 @@ void __global__ k_nonbonded_exclusions(
     unsigned long long g_sigj = 0;
     unsigned long long g_epsj = 0;
 
-    RealType real_lambda = static_cast<RealType>(lambda);
     RealType real_beta = static_cast<RealType>(beta);
 
     RealType real_cutoff = static_cast<RealType>(cutoff);
--- a/timemachine/cpp/src/nonbonded.cu
+++ b/timemachine/cpp/src/nonbonded.cu
@@ -427,7 +427,6 @@ void Nonbonded<RealType, Interpolated>::execute_device(
             Interpolated ? d_unsorted_dp_dl_ : d_sorted_dp_dl_,
             d_w_,
             d_dw_dl_,
-            lambda,
             d_exclusion_idxs_,
             d_scales_,
             beta_,


timemachine/cpp/src/kernels/k_nonbonded_pairs.cuh
```diff
@@ -0,0 +1,200 @@
+#include "../fixed_point.hpp"
+#include "k_fixed_point.cuh"
+
+template <typename RealType>
+void __global__ k_nonbonded_pairs(
+    const int M, // number of pairs
```
mcwitt, line 6: Not sure if we have a convention for this already?

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -752,204 +752,3 @@ void __global__ k_nonbonded_unified(
             u_buffer);
     };
 }
-
-// tbd add restrict
```
mcwitt, line 756: Is this comment still relevant?

badisa: I don't think so

timemachine/cpp/src/kernels/k_nonbonded_pairs.cuh
proteneer: used to be called E, but I think anything works

mcwitt: Makes sense; I had assumed we used `E` for "exclusions", and in that case we'd probably want something else for the general case

timemachine/cpp/src/kernels/k_nonbonded_pairs.cuh
```diff
@@ -0,0 +1,202 @@
+#pragma once
```
proteneer, line 1: We may want to add upfront that the order of associative operations in this file must be consistent with that of the k_nonbonded.cuh file in order for the exclusions to cancel out correctly.

mcwitt: good call; done in 4a9d33d

timemachine/cpp/src/nonbonded_pairs.cu
```diff
@@ -0,0 +1,143 @@
...
+
+    gpuErrchk(cudaMalloc(&d_dp_dl_, N_ * 3 * sizeof(*d_dp_dl_)));
+
+    gpuErrchk(cudaMalloc(&d_scales_, M_ * 2 * sizeof(*d_scales_)));
+    gpuErrchk(cudaMemcpy(d_scales_, &scales[0], M_ * 2 * sizeof(*d_scales_), cudaMemcpyHostToDevice));
+
+    // construct identity permutation
+    if (Interpolated) {
+        gpuErrchk(cudaMalloc(&d_perm_, N_ * sizeof(*d_perm_)));
+        for (int i = 0; i < N_; i++) {
```
badisa, line 62: Can you set GPU memory this way? Pretty sure this needs to be done through code like this: https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/nonbonded.cu#L223-L229

timemachine/cpp/src/nonbonded_pairs.cu
```diff
@@ -0,0 +1,143 @@
...
+    const double lambda,
+    unsigned long long *d_du_dx,
+    double *d_du_dp,
+    unsigned long long *d_du_dl,
+    unsigned long long *d_u,
+    cudaStream_t stream) {
+
+    const int tpb = 32;
+
+    dim3 dimGrid((M_ + tpb - 1) / tpb, 1, 1);
```
badisa, line 99: You can pass `(M_ + tpb - 1) / tpb` to kernel calls without having to construct dimGrids if you want a 1d set of computations

timemachine/cpp/src/nonbonded_pairs.cu
mcwitt: makes sense - I guess we could also allocate and initialize on the host and copy? (since this is only happening at construction, I don't think it'd be a bottleneck?)

badisa: Allocating on the host would also work

timemachine/cpp/src/kernels/k_nonbonded_pairs.cuh
```diff
@@ -0,0 +1,208 @@
...
+    int lj_param_idx_sig_i = atom_i_idx * 3 + 1;
+    int lj_param_idx_eps_i = atom_i_idx * 3 + 2;
+
+    RealType qi = params[charge_param_idx_i];
+    RealType sig_i = params[lj_param_idx_sig_i];
+    RealType eps_i = params[lj_param_idx_eps_i];
+
+    unsigned long long g_qi = 0;
+    unsigned long long g_sigi = 0;
+    unsigned long long g_epsi = 0;
```
maxentile, line 68: I've eyeballed for local self-consistency (e.g. that L42-68 is identical to L70-96, and that the lines to zip (`x`, `+ 0`), (`y`, `+ 1`), (`z`, `+ 2`) etc. seem right) and this all looks look fine!

Not required, but I find long stretches of manually unrolled index arithmetic tedious to verify (and sometimes they contain subtle typos that sneak past review). Would be a small quality-of-life improvement if we can find a more concise idiom for `.x`, `.y`, `.z` that doesn't introduce runtime cost...

timemachine/cpp/src/kernels/k_nonbonded_pairs.cuh
```diff
@@ -0,0 +1,208 @@
...
+        gj_z -= FLOAT_TO_FIXED_NONBONDED(-delta_prefactor * delta_z);
+
+        // energy is size extensive so this may not be a good idea
+        energy -= FLOAT_TO_FIXED_NONBONDED(u);
+
+        g_qi -= FLOAT_TO_FIXED_DU_DP<RealType, FIXED_EXPONENT_DU_DCHARGE>(charge_scale * qj * inv_dij * ebd);
+        g_qj -= FLOAT_TO_FIXED_DU_DP<RealType, FIXED_EXPONENT_DU_DCHARGE>(charge_scale * qi * inv_dij * ebd);
+
+        real_du_dl -= delta_w * delta_prefactor * (dw_dl_i - dw_dl_j);
+        real_du_dl -= charge_scale * inv_dij * ebd * fix_nvidia_fmad(qj, dq_dl_i, qi, dq_dl_j);
```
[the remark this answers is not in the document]
mcwitt: Nice, I personally really like `strict_mul(a, b) + strict_mul(c, d)`. Maybe we could create an issue and implement this in a separate PR?

timemachine/cpp/src/kernels/k_nonbonded_pairs.cuh
mcwitt: This makes sense to me. To be clear, the implementation here was copied mostly unmodified from `k_nonbonded.cuh`, so I don't think there would be any new bugs introduced in the index arithmetic. But I'm definitely in favor of finding a zero (or acceptable) cost abstraction to make correctness more evident.

timemachine/cpp/src/kernels/k_nonbonded_pairs.cuh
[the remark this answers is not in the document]
maxentile: Migrated to https://github.com/proteneer/timemachine/issues/549 !

timemachine/cpp/src/nonbonded_pairs.cu
```diff
@@ -0,0 +1,149 @@
...
+    gpuErrchk(cudaMalloc(&d_scales_, M_ * 2 * sizeof(*d_scales_)));
+    gpuErrchk(cudaMemcpy(d_scales_, &scales[0], M_ * 2 * sizeof(*d_scales_), cudaMemcpyHostToDevice));
+
+    if (Interpolated) {
+
+        gpuErrchk(cudaMalloc(&d_scales_, M_ * 2 * sizeof(*d_scales_)));
+        gpuErrchk(cudaMemcpy(d_scales_, &scales[0], M_ * 2 * sizeof(*d_scales_), cudaMemcpyHostToDevice));
+
+        // construct identity permutation
+        int *perm = static_cast<int *>(malloc(N_ * sizeof(int)));
```
badisa, line 65: Since this is malloc'd don't we also need to free it? Using a std::vector<int> would avoid this

mcwitt: Good call, thanks. I think using `std::vector` makes sense.

timemachine/cpp/src/nonbonded_pairs.cu
mcwitt: Thanks, I think that makes sense. Changed in f4a07f0 (and also switched to using `ceil_divide` from `math_utils.h`).

timemachine/cpp/src/nonbonded_pairs.cu
mcwitt: Fixed in e849535.

timemachine/cpp/src/nonbonded_pairs.cu
mcwitt: Should be fixed to allocate on host and copy as of e849535.

timemachine/cpp/src/nonbonded_dense.cu
```diff
@@ -408,37 +395,6 @@ void Nonbonded<RealType, Interpolated>::execute_device(
         gpuErrchk(cudaPeekAtLastError());
     }
 
-    // exclusions use the non-sorted version
-    if (E_ > 0) {
```
mcwitt, line 412: TODO: might want to add this check back in the appropriate place

timemachine/cpp/src/kernels/k_nonbonded_pairs.cuh
```diff
@@ -0,0 +1,208 @@
...
+        gi_x -= FLOAT_TO_FIXED_NONBONDED(delta_prefactor * delta_x);
+        gi_y -= FLOAT_TO_FIXED_NONBONDED(delta_prefactor * delta_y);
+        gi_z -= FLOAT_TO_FIXED_NONBONDED(delta_prefactor * delta_z);
+
+        gj_x -= FLOAT_TO_FIXED_NONBONDED(-delta_prefactor * delta_x);
+        gj_y -= FLOAT_TO_FIXED_NONBONDED(-delta_prefactor * delta_y);
+        gj_z -= FLOAT_TO_FIXED_NONBONDED(-delta_prefactor * delta_z);
+
+        // energy is size extensive so this may not be a good idea
+        energy -= FLOAT_TO_FIXED_NONBONDED(u);
```
mcwitt, line 171: TODO: the implementation here is still specialized to exclusions (returns the _negative_ sum of pair potentials). We'll probably want flip the convention and return the positive potential since we intend to use this as a generic nonbonded pairs potential.

timemachine/cpp/src/kernels/k_nonbonded_pairs.cuh
```diff
@@ -0,0 +1,208 @@
...
+    delta_x -= box_x * nearbyint(delta_x * inv_box_x);
+    delta_y -= box_y * nearbyint(delta_y * inv_box_y);
+    delta_z -= box_z * nearbyint(delta_z * inv_box_z);
+
+    RealType delta_w = ci_w - cj_w;
+    RealType d2ij = delta_x * delta_x + delta_y * delta_y + delta_z * delta_z + delta_w * delta_w;
+
+    unsigned long long energy = 0;
+
+    int is_vanilla =
```
mcwitt, line 127: TODO: more descriptive name for this?

timemachine/cpp/src/kernels/nonbonded_common.cuh
```diff
@@ -0,0 +1,106 @@
+#pragma once
```
badisa, line 1: Since we now have a header and a source file, should we move the implementation over to the source file to have a consistent style?

timemachine/cpp/src/nonbonded_pairs.cu
```diff
@@ -0,0 +1,197 @@
...
+          kernel_cache_.program(kernel_src.c_str()).kernel("k_permute_interpolated").instantiate()),
+      compute_add_du_dp_interpolated_(
+          kernel_cache_.program(kernel_src.c_str()).kernel("k_add_du_dp_interpolated").instantiate()) {
+
+    if (pair_idxs.size() % 2 != 0) {
+        throw std::runtime_error("pair_idxs.size() must be exactly 2*M");
+    }
+
+    for (int i = 0; i < M_; i++) {
+        auto src = pair_idxs[i * 2 + 0];
```
badisa, line 31: @proteneer Do we use auto? I thought you had told me not to do that, but could be mistaken

timemachine/cpp/src/nonbonded_pairs.cu
```diff
@@ -0,0 +1,197 @@
...
+// TODO: this implementation is duplicated from NonbondedDense. Worth adding NonbondedBase?
+template <typename RealType, bool Negated, bool Interpolated>
+void NonbondedPairs<RealType, Negated, Interpolated>::du_dp_fixed_to_float(
+    const int N, const int P, const unsigned long long *du_dp, double *du_dp_float) {
+
+    // In the interpolated case we have derivatives for the initial and final parameters
+    const int num_tuples = Interpolated ? N * 2 : N;
+
+    for (int i = 0; i < num_tuples; i++) {
+        const int idx_charge = i * 3 + 0, idx_sig = i * 3 + 1, idx_eps = i * 3 + 2;
```
badisa, line 178: nit: Having this on three lines would be easier to read

badisa (approved):

timemachine/cpp/src/nonbonded.hpp
```diff
@@ -1,102 +1,32 @@
...
-    bool disable_hilbert_;
-
-    void hilbert_sort(const double *d_x, const double *d_box, cudaStream_t stream);
-
-    jitify::JitCache kernel_cache_;
-    jitify::KernelInstantiation compute_w_coords_instance_;
-    jitify::KernelInstantiation compute_permute_interpolated_;
-    jitify::KernelInstantiation compute_add_du_dp_interpolated_;
+    NonbondedDense<RealType, Interpolated> dense_;
+    NonbondedPairs<RealType, true, Interpolated> exclusions_; // implement exclusions as negated NonbondedPairs
```
maxentile, line 14: Subjective nit
```suggestion
    bool Negated = true;
    NonbondedPairs<RealType, Negated, Interpolated> exclusions_;
```

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -902,9 +903,61 @@ template <typename RealType, bool Interpolated> void declare_nonbonded(py::modul
...
+            py::arg("transform_lambda_charge") = "lambda",
+            py::arg("transform_lambda_sigma") = "lambda",
+            py::arg("transform_lambda_epsilon") = "lambda",
+            py::arg("transform_lambda_w") = "lambda");
+}
+
 template <typename RealType, bool Interpolated> void declare_nonbonded_pairs(py::module &m, const char *typestr) {
 
-    using Class = timemachine::NonbondedPairs<RealType, Interpolated>;
+    using Class = timemachine::NonbondedPairs<RealType, false, Interpolated>;
```
maxentile, line 960: Nice -- I think this makes sense to keep the wrapper specialized to `Negated=false`

timemachine/cpp/src/kernels/k_nonbonded_pairs.cuh
```diff
@@ -154,27 +158,27 @@ void __global__ k_nonbonded_pairs(
             g_epsi += FLOAT_TO_FIXED_DU_DP<RealType, FIXED_EXPONENT_DU_DEPS>(-eps_grad * eps_j);
             g_epsj += FLOAT_TO_FIXED_DU_DP<RealType, FIXED_EXPONENT_DU_DEPS>(-eps_grad * eps_i);
 
-            real_du_dl -= sig_grad * (dsig_dl_i + dsig_dl_j);
+            real_du_dl += sign * sig_grad * (dsig_dl_i + dsig_dl_j);
```
maxentile, line 161: Difficult to verify at a glance if `accumulator -= x` --> `accumulator += sign * x` is applied everywhere it needs to be -- should `sign` also be applied to the `g_sigi` and `g_epsi` increments?

maxentile (commented):
Looks great to me!

(I just want to stare a bit more at one stretch where sign-flip errors seem easy to make)

timemachine/cpp/src/kernels/k_nonbonded_pairs.cuh
maxentile: (Additionally, might apply `result = sign * accumulator` at the end, instead of to each of the increments)

timemachine/cpp/src/kernels/nonbonded_common.cuh
mcwitt: @badisa I might be mistaken but I think the templated functions and those declared with `__forceinline__` need to stay in the header file (templated functions because usage invokes template instantiation and `__forceinline__` so they can be inlined before optimization.

timemachine/cpp/src/wrap_kernels.cpp
mcwitt: Agreed -- I considered exposing this to the Python wrapper, but it seemed best to only wrap the non-negated version for now for simplicity.

timemachine/cpp/src/kernels/k_nonbonded_pairs.cuh
mcwitt: What do you think about factoring this into an inlined helper function as in 275d787?

timemachine/cpp/src/nonbonded_pairs.cu
mcwitt: I like it, changed in 935428b

timemachine/cpp/src/nonbonded.hpp
mcwitt: I like it, added in c299d45. (needed `static const` for this to compile)

timemachine/cpp/src/kernels/k_nonbonded_pairs.cuh
mcwitt: > Difficult to verify at a glance if accumulator -= x --> accumulator += sign * x is applied everywhere it needs to be

Agreed. I originally checked this by comparing the implementations in [`v_nonbonded_unified`](https://github.com/proteneer/timemachine/blob/669320ece66894d4204909f9359bc4e079de50cd/timemachine/cpp/src/kernels/k_nonbonded.cuh#L399) with [`k_nonbonded_exclusions`](https://github.com/proteneer/timemachine/blob/669320ece66894d4204909f9359bc4e079de50cd/timemachine/cpp/src/kernels/k_nonbonded.cuh#L747). But I think your idea of changing the sign at the final accumulation is more clearly correct.

mcwitt: Oh, I was too hasty, 275d787 doesn't actually work. I'm not exactly sure why, but I'll have to think more carefully about the order of the floating-to-fixed conversions.

mcwitt: > should sign also be applied to the g_sigi and g_epsi increments?

Good call, I missed these (they were accidentally correct for the `Negated = true` case, but we're currently missing a test for `Negated = false`). Fixed in 1ae5fa3.

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -1010,6 +1124,18 @@ PYBIND11_MODULE(custom_ops, m) {
     declare_nonbonded<double, false>(m, "f64");
     declare_nonbonded<float, false>(m, "f32");
 
+    declare_nonbonded_dense<double, true>(m, "f64_interpolated");
```
badisa, line 1127: We should add tests for the pairs and the dense, as this will be used but currently the wrappers aren't tested.

timemachine/cpp/src/kernels/k_nonbonded_pairs.cuh
maxentile: >Good call, I missed these (they were accidentally correct for the Negated = true case, but we're currently missing a test for Negated = false). Fixed in 1ae5fa3.

Nice -- that looks good!

>What do you think about factoring this into an inlined helper function as in 275d787?
>idea of changing the sign at the final accumulation is more clearly correct.

Current implementation looks good! (Although I have a slight preference for flipping sign at the end rather than during accumulation (and reverting to multiplying by a sign constant rather than applying an inlined function).)

maxentile: >I have a slight preference for flipping sign at the end rather than during accumulation

Ahh, @mcwitt points out an issue I had overlooked with this suggestion: the starting values could be nonzero!

timemachine/cpp/src/wrap_kernels.cpp
mcwitt: Good call. Since this PR is already getting big, I'm thinking to revert exposing `NonbondedDense` and `NonbondedPairs` to user code for now (5d286a5). Then we can test them individually when they're exposed in a future PR.

maxentile (approved):

timemachine/cpp/src/nonbonded_pairs.cu
mcwitt: (This is not new; it was copied from existing code in `nonbonded.cu`.)

mcwitt:
> lgtm pending mem-check/leak-check

Good call, I did find one issue, fixed in ab1edb8. There were also leaks detected in `master` when running `test_nonbonded`. I'll open a separate issue to deal with these.

```
========= LEAK SUMMARY: 7536 bytes leaked in 14 allocations
========= ERROR SUMMARY: 14 errors
```

mcwitt:
I added and reverted 38381b4 to demonstrate the test failure seen when implementing exclusions by passing negative scale parameters (this was the reason to switch to controlling negation with a template argument in fa0d85e).

PR #543, opened 2021-12-08
Make summed potential interface consistent with other custom ops

mcwitt:
Modifies the `SummedPotential` Python interface for consistency with other custom ops and easier composition of potentials:

- `SummedPotential` inherits from `CustomOpWrapper`
- constructor accepts a list of `CustomOpWrapper` objects
- sub-potentials initialized with consistent precision

This also removes the hack previously used to prevent deallocation of the sub-potentials when direct references from Python go out of scope, instead passing the [keep_alive call policy](https://pybind11.readthedocs.io/en/stable/advanced/functions.html#keep-alive) in the binding code.

This will be useful for implementing the full nonbonded potential as a sum of raw and exclusion components in #542.

### DIFF
--- a/timemachine/cpp/src/wrap_kernels.cpp
+++ b/timemachine/cpp/src/wrap_kernels.cpp
@@ -952,7 +952,8 @@ void declare_summed_potential(py::module &m) {
                 return new timemachine::SummedPotential(potentials, params_sizes);
             }),
             py::arg("potentials"),
-            py::arg("params_sizes"));
+            py::arg("params_sizes"),
+            py::keep_alive<1, 2>());
 }
 
 const py::array_t<double, py::array::c_style>
--- a/timemachine/lib/potentials.py
+++ b/timemachine/lib/potentials.py
@@ -36,25 +36,15 @@ class CustomOpWrapper:
         return custom_ops.BoundPotential(self.unbound_impl(precision), self.params)
 
 
-class SummedPotential:
+class SummedPotential(CustomOpWrapper):
     def __init__(self, potentials: List[CustomOpWrapper], params_init: List[NDArray]):
         self._potentials = potentials
         self._sizes = [ps.size for ps in params_init]
         self.params = None
 
-    def bind(self, params: List[NDArray]) -> "SummedPotential":
-        self.params = params
-        return self
-
-    def unbound_impl(self, precision: str) -> custom_ops.SummedPotential:
+    def unbound_impl(self, precision: str):
         impls = [p.unbound_impl(precision) for p in self._potentials]
         return custom_ops.SummedPotential(impls, self._sizes)
-
-    def bound_impl(self, precision: str) -> custom_ops.BoundPotential:
-        if self.params is None:
-            raise ValueError("This op has not been bound to parameters.")
-
-        return custom_ops.BoundPotential(self.unbound_impl(precision), self.params)
 
 
 # should not be used for Nonbonded Potentials
--- a/timemachine/lib/potentials.py
+++ b/timemachine/lib/potentials.py
@@ -42,7 +42,7 @@ class SummedPotential(CustomOpWrapper):
         self._sizes = [ps.size for ps in params_init]
         self.params = None
 
-    def unbound_impl(self, precision: str):
+    def unbound_impl(self, precision):
         impls = [p.unbound_impl(precision) for p in self._potentials]
         return custom_ops.SummedPotential(impls, self._sizes)
 
--- a/timemachine/lib/potentials.py
+++ b/timemachine/lib/potentials.py
@@ -44,7 +44,7 @@ class SummedPotential:
 
     def bind(self, params: List[NDArray]) -> "SummedPotential":
         self.params = params
-        pass
+        return self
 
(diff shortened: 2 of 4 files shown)


timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -952,7 +952,8 @@ void declare_summed_potential(py::module &m) {
                 return new timemachine::SummedPotential(potentials, params_sizes);
             }),
             py::arg("potentials"),
-            py::arg("params_sizes"));
+            py::arg("params_sizes"),
+            py::keep_alive<1, 2>());
```
badisa, line 956: This to avoid the potentials going out of scope and being cleaned up?

timemachine/lib/potentials.py
```diff
@@ -34,6 +36,27 @@ def bound_impl(self, precision):
         return custom_ops.BoundPotential(self.unbound_impl(precision), self.params)
 
 
+class SummedPotential:
```
badisa, line 39: Should this inherit from the wrapper?

timemachine/lib/potentials.py
```diff
@@ -34,6 +36,27 @@ def bound_impl(self, precision):
...
+    def __init__(self, potentials: List[CustomOpWrapper], params_init: List[NDArray]):
+        self._potentials = potentials
+        self._sizes = [ps.size for ps in params_init]
+        self.params = None
+
+    def bind(self, params: List[NDArray]) -> "SummedPotential":
```
badisa, line 45: Should return self based on the type hint

mcwitt: Ah, good catch. Fixed in 4d7e4a8.

timemachine/cpp/src/wrap_kernels.cpp
mcwitt: Yes, it avoids the need to use the hack I had before introducing a wrapper class whose only purpose was to keep a reference to the potentials.

timemachine/lib/potentials.py
mcwitt: I went back and forth on this; the method signature is the same as `CustomOpWrapper`, but the constructor and nontrivial parts of the implementation are different. I also thought about adding an abstract base class from which both inherit, but opted to keep it as simple as possible. I'll reconsider extending `CustomOpWrapper`

mcwitt: On further reflection, I think it does make sense to inherit from the wrapper, and simplifies things quite a bit! Changed in 4a26f7c.

timemachine/cpp/src/wrap_kernels.cpp
proteneer: When does it die?

badisa: Should we fix this more generally? By fixing it only for the summed potential it means the usage of potentials have to be differently based on the user's knowledge of the wrapping layer

mcwitt: > When does it die?
@proteneer Without `keep_alive` here, `test_construct_differentiable_interface_fast` fails with a [segfault](https://gist.github.com/mcwitt/f35f2fa2695f4ea1648f60758d4c13ba). I think the cleanup is happening when the [SummedPotential.unbound_impl](https://github.com/proteneer/timemachine/blob/eefac40e06025a49aa1a0ec18b799948aaee85d7/timemachine/lib/potentials.py#L45) returns, since at that point the only Python reference to the C++ `Potential` objects (`impls`) goes out of scope.

Storing a reference to the potentials in the `SummedPotential` object is a partial fix, but doesn't help in cases like [this](https://github.com/proteneer/timemachine/blob/eefac40e06025a49aa1a0ec18b799948aaee85d7/fe/functional.py#L63) where we keep only the implementation (not the full `SummedPotential` wrapper) in scope.

> Should we fix this more generally?
@badisa As far as I understand this particular issue is unique to `SummedPotential` because its implementation stores references to objects allocated on the Python side (which I don't think other existing potentials do?). But @maxentile [mentioned](https://github.com/proteneer/timemachine/pull/537#discussion_r756393280) that we sometimes have this issue elsewhere in the code. Would `keep_alive` be useful in the referenced case?

badisa: @mcwitt I see what you mean, because we construct the unbound potentials in the function. Thanks

timemachine/lib/potentials.py
```diff
@@ -34,6 +36,17 @@ def bound_impl(self, precision):
         return custom_ops.BoundPotential(self.unbound_impl(precision), self.params)
 
 
+class SummedPotential(CustomOpWrapper):
+    def __init__(self, potentials: List[CustomOpWrapper], params_init: List[NDArray]):
+        self._potentials = potentials
```
badisa, line 41: Would be good to verify that the number of potentials agrees with the number of parameter sets

mcwitt: Good call, added in 9d898d7 and tested in 6c9572b

badisa (approved):

PR #546, opened 2021-12-08
Add setters for velocities and box

maxentile:
Currently, to run MD from new initial conditions (x, v, box), we create a new context with the desired initial conditions. This is because the context doesn't have setters for v and box (although it has a setter for x).

The overhead for context creation in timemachine is very small (~0.2 ms), but it might still be preferable to have setters for v and box. Would this break anything / make any kinds of user errors more likely?

### DIFF
--- a/timemachine/cpp/src/wrap_kernels.cpp
+++ b/timemachine/cpp/src/wrap_kernels.cpp
@@ -232,6 +232,16 @@ void declare_context(py::module &m) {
                 ctxt.set_x_t(new_x_t.data());
             })
         .def(
+            "set_v_t",
+            [](timemachine::Context &ctxt, const py::array_t<double, py::array::c_style> new_v_t) {
+                ctxt.set_v_t(new_v_t.data());
+            })
+        .def(
+            "set_box",
+            [](timemachine::Context &ctxt, const py::array_t<double, py::array::c_style> new_box) {
+                ctxt.set_box(new_box.data());
+            })
+        .def(
             "get_x_t",
             [](timemachine::Context &ctxt) -> py::array_t<double, py::array::c_style> {
                 unsigned int N = ctxt.num_atoms();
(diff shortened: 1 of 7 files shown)


timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -231,6 +231,16 @@ void declare_context(py::module &m) {
             [](timemachine::Context &ctxt, const py::array_t<double, py::array::c_style> new_x_t) {
                 ctxt.set_x_t(new_x_t.data());
             })
+        .def(
+            "set_v_t",
```
badisa, line 235: For these two new wrapped functions should we verify that the velocities and box are of the correct shape? The box seems easiest, though making sure we don't corrupt the length of velocities seems handy.

This probably goes for all of our setters... though not necessary for this PR

maxentile: Good call. Runtime errors caused by passing in position, box, or parameter arrays of the wrong size can be catastrophic and should be prevented as early as possible.

I should fix in this PR, since the PR multiplies the opportunities to encounter this sort of problem!

(For example, if I pass in a 3 x 2 `box`, no error will be raised by the new setter, but the context will have a 3 x 3 `box` with zeros for the missing entries. If I pass in a 3 x 4 `box`, no error will be raised by the new setter but the context will have a 3 x 4 `box`...)

proteneer:
I intentionally avoided `set_v_t()` because it wasn't clear which integrator we were going to use. I think you should be clear that as is implemented, for BAOAB, the velocities that are set should be corrected by half a time step (esp if people are sampling v_t directly from Maxwell-Boltzmann). I'm still happy to keep accept the PR, but please document this behavior.

Maybe a better change later on would be to move the velocity to an internal state local to the integrator as opposed to the Context, so at least that way when we call `set_v_t()` on a BAOABIntegrator, the semantics are clear.

maxentile:
Closing until
* fix for https://github.com/proteneer/timemachine/issues/562 is finalized
* half-step velocity offset behavior mentioned here https://github.com/proteneer/timemachine/pull/546#issuecomment-988460644 is better understood (and possibly adopted wherever velocities are currently set by calling the constructor). (See also https://github.com/proteneer/timemachine/pull/541#discussion_r760207637 )

PR #548, opened 2021-12-13
Makes endpoint correction estimate_delta_us deterministic

badisa:
* Ensures deterministic dGs for RABFE if everything is seeded correctly


fe/endpoint_correction.py
```diff
@@ -18,12 +18,11 @@ def exp_u(rotation, k, beta):
 def sample_multiple_rotations(k, beta, size):
     num_batches = 500
     batch_size = 10000
-
+    state = np.random.RandomState(2021)
```
maxentile, line 21: Nit: Could possibly lead to subtle usage errors in the future if we later use `sample_multiple_rotations` in a new context that might assume independent rotations to be sampled on each call.

Might suggest either:
* adding a note in the docstring that it's a deterministic function of `(k, beta, size)`, or perhaps
* modifying the signature to include a `seed=None` or `random_state=None` optional argument, setting the seed deterministically from `estimate_delta_us`

tests/test_endpoint_correction.py
```diff
@@ -170,11 +170,21 @@ def test_endpoint_correction():
     k_translation = 200.0
 
     results = []
-    for k_rotation in [0.0, 50.0, 1000.0]:
+    for i, k_rotation in enumerate([0.0, 50.0, 1000.0]):
 
         lhs_du, rhs_du, rotations, translations = endpoint_correction.estimate_delta_us(
             k_translation, k_rotation, core_idxs, core_params, beta, lhs_xs, rhs_xs
         )
+        if i == 0:
```
maxentile, line 178: Any reason not to assert deterministic for `i > 0` too?

maxentile (approved):

tests/test_endpoint_correction.py
badisa: Only done to avoid the test taking too long. The test takes an extra ~30 seconds if you verify all of the iterations of the loop

maxentile: Thanks for clarifying! `i == 0` as implemented looks good then

mcwitt (approved):

fe/estimator_abfe.py
```diff
@@ -373,6 +373,7 @@ def _deltaG_from_results(model, results, sys_params) -> Tuple[Tuple[float, List]
             beta=model.beta,
             lhs_xs=results[-2].xs,
             rhs_xs=results[-1].xs,
+            seed=2021,
```
badisa, line 376: This commit may not make it, but would make the RABFE protocol deterministic all the way through. Note that the inputs still change as a function of the seed of the simulation.

proteneer (approved):

PR #550, opened 2021-12-17
Add getter methods to SummedPotential, BoundPotential Python bindings

mcwitt:
- Adds `BoundPotential::get_potential`, returning the unbound potential
- Adds `SummedPotential::get_potentials`, returning a vector of unbound component potentials
- Adds tests for the above, e.g. `BoundPotential(potential, ..).get_potential() === potential`
- Adds a test that the unbound potential referred to by `BoundPotential` is not deallocated prematurely (similar to the existing test for `SummedPotential` introduced in #552)

I've found this useful for debugging in cases where we have a list of `BoundPotential`s and no obvious way to distinguish them on the Python side. Similarly, I think it will be useful to be able to introspect `SummedPotential`s.

### DIFF
--- a/tests/test_potentials.py
+++ b/tests/test_potentials.py
@@ -1,11 +1,25 @@
 import pytest
+import numpy as np
 
-import timemachine.lib.potentials as ps
+from timemachine.lib import custom_ops, potentials
 
 
-def test_summed_potential_raises_on_inconsistent_lengths():
+@pytest.fixture
+def harmonic_bond():
+    return potentials.HarmonicBond(np.array([[0, 1], [0, 2]], dtype=np.int32)).bind([[1, 0], [1, 0]])
+
+
+def test_summed_potential_raises_on_inconsistent_lengths(harmonic_bond):
 
     with pytest.raises(ValueError) as excinfo:
-        ps.SummedPotential([ps.HarmonicBond()], [])
+        potentials.SummedPotential([harmonic_bond], [])
 
     assert str(excinfo.value) == "number of potentials != number of parameter arrays"
+
+
+def test_bound_potential_get_potential(harmonic_bond):
+
+    unbound_impl = harmonic_bond.unbound_impl(np.float32)
+    bound_impl = custom_ops.BoundPotential(unbound_impl, harmonic_bond.params)
+
+    assert unbound_impl is bound_impl.get_potential()
--- a/tests/test_potentials.py
+++ b/tests/test_potentials.py
@@ -6,7 +6,7 @@ from timemachine.lib import custom_ops, potentials
 
 @pytest.fixture
 def harmonic_bond():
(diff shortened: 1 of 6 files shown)


proteneer (commented):
looks simple, but maybe add a test on the python side just for sanity (and maybe check scope/memory cleanup)? not sure if this gets cleaned-up

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -523,6 +523,7 @@ void declare_bound_potential(py::module &m) {
             }),
             py::arg("potential"),
             py::arg("params"))
+        .def("get_potential", [](const timemachine::BoundPotential &bp) { return bp.potential; })
```
maxentile, line 526: A more explicit name might be `get_unbound_impl`. 

Cons: Would no longer be consistent with argument name `potential`.

Pros: Unambiguous whether it returns `harmonic_bond` potential or its `unbound_impl` in test https://github.com/proteneer/timemachine/blob/e01c188e6dc45df45816350e6b8215937611a127/tests/test_potentials.py#L27-L31

tests/test_potentials.py
```diff
@@ -1,11 +1,35 @@
...
-import timemachine.lib.potentials as ps
+from timemachine.lib import custom_ops, potentials
 
 
-def test_summed_potential_raises_on_inconsistent_lengths():
+@pytest.fixture
+def harmonic_bond():
+    bond_idxs = np.array([[0, 1], [0, 2]], np.int32)
+    params = np.array([[1, 0], [0, 1]], dtype=np.float32)
+    return potentials.HarmonicBond(bond_idxs).bind(params)
```
maxentile: Nice

maxentile (approved):
Looks good!

timemachine/cpp/src/wrap_kernels.cpp
mcwitt: Good point about the potential ambiguity. I had considered `get_unbound_impl`, but was thinking that might also be confusing because (as far as I know) the un/bound language is currently only used on the Python side

tests/test_potentials.py
```diff
@@ -11,24 +11,42 @@ def harmonic_bond():
...
+    lam = 1
+    bp.execute(coords, box, lam)
 
 
-def test_summed_potential_raises_on_inconsistent_lengths(harmonic_bond):
+def test_bound_potential_keeps_referenced_potential_alive(harmonic_bond):
+    bp = harmonic_bond.bound_impl(np.float32)
+
+    # segfaults if referenced potential has been deallocated prematurely
+    execute_bound_impl(bp)
```
mcwitt, line 25: Also checked that this does indeed segfault if the `smart_ptr` reference to the underlying potential in `BoundPotential` is replaced with a bare pointer (3515563)/

PR #551, opened 2021-12-17
Allow building with newer versions of cmake

mcwitt:
Modifies the build configuration for compatibility with recent versions of cmake (3.18+).

The cuda architecture is now specified using the [CUDA_ARCHITECTURES](https://cmake.org/cmake/help/git-stage/prop_tgt/CUDA_ARCHITECTURES.html) cmake property.

Also, the formatting of the version string has changed to not include the `sm_` prefix. This will require updating any build scripts where we override the default version in other repos (should be done before merging this).

### Todo
- [ ] Update CI build configuration to use cmake 3.18+ ([PR](https://git.relaytxinternal.com/fyork/timemachine-ansible/-/merge_requests/6))


proteneer:
Nice! Looks like a minor change overall

Makefile
```diff
@@ -6,7 +6,7 @@ PYTEST_CI_ARGS := --cov=. --cov-report=term-missing
 
 NPROCS = `nproc`
 
-CUDA_ARCH := "sm_70"
+CUDA_ARCH := "70"
```
badisa, line 9: Would be good to support the sm_ prefix and strip it off it is provided. Don't know that we will fix this everywhere immediately and its easy to solve with some sed
```
echo sm_75 | sed -e "s/^sm_//"
```

README.md
```diff
@@ -31,7 +31,7 @@ We currently support the following functional forms. Parameters that can be opti
 * OpenEye Toolkits
 * OpenMM
 * Cuda 11.0+
-* CMake 3.10.2 (newer versions may vary in the CMAKE_INSTALL_PREFIX and CUDA flags)
+* CMake 3.21 (newer versions may vary in the CMAKE_INSTALL_PREFIX and CUDA flags)
```
badisa, line 34: What is the minor version that you are using?

Makefile
mcwitt: I like it. Done in abe588d.

badisa:
It looks like it is passing on CI, we don't even need to upgrade cmake? Or does that get us something?

mcwitt:
@badisa 
> It looks like it is passing on CI, we don't even need to upgrade cmake? Or does that get us something?

Good call -- it seems like your suggestion to strip off the `sm_` prefix from the `CUDA_ARCH` env var makes this backwards-compatible with the existing cmake version in CI (3.10.2), so I don't think we'd need to upgrade.

mcwitt:
> makes this backwards-compatible with the existing cmake version in CI (3.10.2)

hmm, actually, on closer inspection this doesn't seem correct. The build succeeding since the change to strip the prefix, but `CUDA_ARCH` seems to be ignored -- for example [the last build](https://jenkins.relaytxinternal.com/blue/organizations/jenkins/timemachine/detail/PR-551/5/pipeline) used `sm_52`.

It looks like the `CUDA_ARCHITECTURES` cmake prop was only [added in 3.18](https://cmake.org/cmake/help/git-stage/prop_tgt/CUDA_ARCHITECTURES.html#:~:text=new%20in%20version%203.18.), so we'd still need to update the CI build to use 3.18+.

badisa (changes requested):
We should increase `cmake_minimum_required` to match the README if the build will build with the incorrect cuda arch

mcwitt:
> We should increase cmake_minimum_required to match the README if the build will build with the incorrect cuda arch

Thanks, I missed that. Done in d50b362.

badisa (approved):
Looks good, just got to wait on the ansible PR

mcwitt:
> If we aren't going to test with 3.18, why not just require it 3.22.1? Seems like we could break support for 3.18 without realizing if we don't test it.

My thought was just to provide the extra piece of information that it _should_ work with 3.18+ (since that's when the required flag was added). We also say "Cuda 11.0+", but I guess Cuda versions are generally backwards-compatible?

I don't have a strong preference; willing to remove the 3.18+ bit if that's potentially confusing.

badisa:
Since we have complete control we should require the version that we intend to test against. If in the future we want to support a wider range of Cmake this would make sense, but I don't want to say we support something we aren't going to test

mcwitt:
> I don't want to say we support something we aren't going to test

Fair enough, edited to list only cmake 3.22.1 in b389354.

Should we also tackle pinning the cuda version here?

badisa:
> > I don't want to say we support something we aren't going to test
> 
> Fair enough, edited to list only cmake 3.22.1 in [b389354](https://github.com/proteneer/timemachine/commit/b389354fdf81b16fb501dc137965f47afed1df6e).
> 
> Should we also tackle pinning the cuda version here?

We should tackle it in a separate PR when we move to the A10G GPUs

badisa (approved):

maxentile (approved):

PR #552, opened 2021-12-17
Use smart pointers for memory management in SummedPotential

mcwitt:
This is a more principled/robust fix for the issue of potentials referenced by a `SummedPotential` being deallocated before the `SummedPotential` instance is deallocated. (This seems to be what we're already doing for `BoundPotential`s).

### Issue

In our pybind11 wrappers, we declare `Potential` to have a ["holder" type](https://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html#std-shared-ptr) of `shared_ptr`. I think this means that Potentials will be deallocated when the shared pointer reference count (including Python + CXX references) reaches zero. `BoundPotential` seems to handle this properly by declaring its reference to the underlying Potential to be of type `shared_ptr`, so the underlying Potential should not be freed before BoundPotential.

But in the current implementation of `SummedPotential`, the vector of references to underlying potentials just holds bare pointers (type `vector<Potential *>`), which I think means that SummedPotential doesn't contribute to the reference count of the underlying potentials and the latter can be deallocated prematurely.

### Solution

This PR changes the type of `SummedPotential::potentials_` to use `shared_ptr`. This should prevent potentials from being prematurely deallocated. It also removes the `keep_alive` annotation, which is no longer necessary.

### DIFF
--- a/tests/test_potentials.py
+++ b/tests/test_potentials.py
@@ -1,11 +1,34 @@
 import pytest
+import numpy as np
 
-import timemachine.lib.potentials as ps
+from timemachine.lib import potentials
 
 
-def test_summed_potential_raises_on_inconsistent_lengths():
+@pytest.fixture
+def harmonic_bond():
+    bond_idxs = np.array([[0, 1], [0, 2]], dtype=np.int32)
+    params = np.ones(shape=(2, 2), dtype=np.float32)
+    return potentials.HarmonicBond(bond_idxs).bind(params)
+
+
+@pytest.fixture
+def summed_potential(harmonic_bond):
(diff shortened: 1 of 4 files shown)


maxentile (approved):

badisa:
Is there a test we could add to cover the case that caused this to occasionally segfault?

proteneer:
I think `shared_ptr` is fine for now, maybe later on we can consider proper ownership semantics via something like `unique_ptr`?

mcwitt:
@badisa good call. I added a failing test in e001880.

PR #553, opened 2021-12-21
[dontmerge] Disagreement with reference in the case of no exclusions

mcwitt:
Test against reference for nonbonded potential fails when exclusions are removed. This seems surprising.


mcwitt:
With @proteneer, determined that this is likely caused by (expected) loss of precision in fixed-to-floating conversion, exacerbated by large terms that are not back-subtracted when exclusions are removed. Closing since this is a known issue that shouldn't impact realistic use cases.

Issue #481, opened 2021-08-09
Detecting and handling numerical overflows


maxentile:
Currently, the potential energy function cannot return a nan or inf when numerical overflow is encountered, instead returning values around `-1e+7`.

This causes problems in two settings:
* Monte Carlo moves -- To compare the proposed state to the current state, we need to be able to compute `U(x_proposed) - U(x_current)`, where we can be ~certain that `U(x_current)` is valid, but it's likely that `U(x_proposed)` will overflow. If `U(x_proposed)` is actually ~+inf, but the returned value is instead `-1e+7`, then we can mistakenly accept a move that will crash the simulation.
* Importance weights -- We need to evaluate `U(x, lam_target) - U(x, lam_reference)`, where we can be ~certain that `U(x, lam_reference)` is valid, but it's likely that `U(x, lam_target)` will overflow. If `U(x, lam_target)` is actually ~+inf, but the returned value is instead `-1e+7`, then that can corrupt further analysis.
  
The [current workaround](https://github.com/proteneer/timemachine/blob/56e363d47c04b45abf4e0a143aea446757af465c/md/ensembles.py#L39-L40) in the Python layer is to compare `abs(U)` to an absolute `guard_threshold` of `1e+6`. Justification for this threshold: For the smallest systems (such as a single LJ pair) as well as practical systems, overflows appear as values around `-1e+7`, while valid energies are at least 2 orders of magnitude away from this threshold. However, if we considered larger systems, then valid typical energies could get even closer to this threshold.

After some discussion surrounding https://github.com/proteneer/timemachine/pull/476#discussion_r682934819 and [`sanitize_energies`](https://github.com/proteneer/timemachine/blob/8d23aa1bc9390c1aa8ede53da91557a145561db2/fe/utils.py#L260-L282), we may want to change this behavior.

Some considerations:
* Valid energies are size-extensive, but invalid energies do not appear to be size-extensive. Do we want to handle the case where `U_suspicious` and `U_valid` can have the same order of magnitude? This seems very difficult.
* Is it possible to detect overflow _during_ the computation, rather than _after_ the computation? I suspect the answer is no, but this would be ideal if possible.

Some possible changes:
[...]

PR #556, opened 2022-01-07
Use fixed point for du_dp in potentials

mcwitt:
Changes the signature of the `Potential::execute*` methods to use `unsigned long long` for `du_dp` as we do for other derivatives.

This is required for the refactor splitting `Nonbonded` into a sum of dense and exclusions potentials -- without these changes, `du_dp` is converted to floating point before summing, and back-subtraction of exclusions will not be exact.

#### Summary of changes

*Reviewer's note:* 849818a is a (partial) revert of #558 to implement a simpler interface. The effective changes start at 3e18d62.

- Partially reverts #558, removing `Potential::fixed_to_float`. For now `du_dp` is the only case where the conversion will be potential-specific, so we can simplify
- Adds `Potential::du_dp_fixed_to_float` to replace `fixed_to_float` and override implementations for `Nonbonded` and `SummedPotential`
- Removes `Observable` and `AvgPartialUParam`. These would've required an overhaul in with the change to fixed-point du/dp but are unlikely to be used going forward
- Removes usage of `AvgPartialUPartialParam` from estimators and tests. Since we don't use `jax.grad(deltaG)` anymore, this just removes the custom vjp rules rather than updating them to remove du/dps.

#### TODO
- [x] Fix for accumulation in nonbonded potential (requires #558)


timemachine/cpp/src/observable.cu
```diff
@@ -56,14 +58,19 @@ void AvgPartialUPartialParam::observe(int step, int N, double *d_x_t, double *d_
...
             nullptr,
             nullptr,
             stream // TBD: parallelize me!
         );
         count_++;
         const int tpb = 32;
         const int blocks = (size + tpb - 1) / tpb;
+
+        for (int i = 0; i < size; i++) {
+            d_du_dp_[i] = FIXED_TO_FLOAT<double>(d_du_dp_buffer_[i]);
```
mcwitt, line 71: todo: this should use potential-specific conversion

timemachine/cpp/src/fixed_point.hpp
```diff
@@ -2,6 +2,16 @@
 
 #define FIXED_EXPONENT 0x1000000000
 
+// we need to use a different level of precision for parameter derivatives
+#define FIXED_EXPONENT_DU_DCHARGE 0x1000000000
+#define FIXED_EXPONENT_DU_DSIG 0x2000000000
+#define FIXED_EXPONENT_DU_DEPS 0x4000000000 // this is just getting silly
```
maxentile, line 8: Naive question: Are there conditions when this list of exponents might need to change in the future?

(E.g. would we expand the list for other parameter types, or would the numeric values need to be adjusted if some assumption changed? Not sure if these currently assume we're going to call `du_dparams(x, params)` only when `u(x, params)` is in some well-behaved range -- perhaps the exponents would need to be different if we start calling `du_dparams` on highly "off-equilibrium" `(x, params)`?)

maxentile (commented):
Looks good -- left a few minor comments!

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -295,40 +295,40 @@ void declare_context(py::module &m) {
...
-            std::vector<int> shape = obj.shape();
-            py::array_t<double, py::array::c_style> buffer(shape);
+//                 return buffer;
+//             })
+//         .def("std_du_dp", [](timemachine::AvgPartialUPartialParam &obj) -> py::array_t<double, py::array::c_style> {
+//             std::vector<int> shape = obj.shape();
+//             py::array_t<double, py::array::c_style> buffer(shape);
 
-            obj.std_du_dp(buffer.mutable_data());
+//             obj.std_du_dp(buffer.mutable_data());
```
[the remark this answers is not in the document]
mcwitt: Good catch. The `Observable` and `AvgPartialUPartialParam` classes would've needed some updates in this PR, but @proteneer mentioned we might be able to deprecate them instead. I've added `pytest.mark.skip` decorators to the tests that depend on these in 16826f1, but I'm not sure these should all be deprecated.. Maybe we can update some of them to avoid needing the above instead?

timemachine/cpp/src/fixed_point.hpp
mcwitt: Hmm, it seems likely that we might want to change these at some point (maybe @proteneer will have a better idea). To be clear, this PR only pushes the existing conversions to the outer layer (the wrapper in `wrap_kernels.cpp`), so I think it should be mostly orthogonal to that issue. Might be worth opening another issue for discussion?

slow_tests/test_benchmark.py
```diff
@@ -313,10 +315,12 @@ def benchmark_hif2a(verbose=False, num_batches=100, steps_per_batch=1000):
             )
 
 
+@pytest.mark.skip("Uses deprecated AvgPartialUPartialParam")
```
mcwitt, line 318: todo: what to do about tests that use `AvgPartialUPartialParam`?

proteneer: Just disable them completely. We don't need to compute them on the fly anymore.

timemachine/cpp/src/fixed_point.hpp
proteneer: Let's not change those values, but document the dynamic range of the supported precision.

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -229,9 +229,33 @@ void __global__ k_inv_permute_assign_2x(
     array_2[perm[idx] * stride + stride_idx] = sorted_array_2[idx * stride + stride_idx];
 }
 
-template <typename RealType>
+// template <typename RealType>
```
proteneer, line 232: nit: can we remove this entirely?

timemachine/cpp/src/wrap_kernels.cpp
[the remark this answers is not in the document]
proteneer: Just deprecate/comment them out entirely.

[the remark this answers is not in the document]
mcwitt: Done in ac23026, 3a3610e, c8dac31, and d3f0ff1.

slow_tests/test_benchmark.py
mcwitt: Done in 3a3610e.

timemachine/cpp/src/kernels/k_nonbonded.cuh
mcwitt: Done in 5db9213.

timemachine/cpp/src/fixed_point.hpp
mcwitt: Should we delegate this to #481?

maxentile: Moved!

timemachine/cpp/src/wrap_kernels.cpp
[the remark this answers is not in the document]
maxentile: Oh, looks like one more reference in tested code: https://github.com/proteneer/timemachine/blob/bdc42fb2cd9c32b76642ca5b581271a2b2dffb0d/tests/test_frame_filters.py#L11 . I'll get any references from untested code when updating `examples/`.

fe/estimator.py
```diff
@@ -25,16 +24,15 @@
...
 
 
 def unflatten(aux_data, children):
-    xs, du_dls, du_dps = aux_data
-    return SimulationResult(xs, du_dls, du_dps)
+    xs, du_dls = aux_data
+    return SimulationResult(xs, du_dls)
 
 
 jax.tree_util.register_pytree_node(SimulationResult, flatten, unflatten)
```
maxentile, line 38: Tangential to this PR, but this might also now be unused (which would also let us remove `flatten` and `unflatten`)

fe/estimator_abfe.py
```diff
@@ -24,17 +24,16 @@
 class SimulationResult:
     xs: np.array
     boxes: np.array
```
maxentile, line 26: Tangential: noticed that `estimator_abfe` defines a simulation result containing `boxes`, but `estimator`'s definition does not includes `boxes`... Will open a separate issue

fe/estimator_abfe.py
```diff
@@ -24,17 +24,16 @@
...
 
 
 def unflatten(aux_data, children):
-    xs, boxes, du_dps, lambda_us = aux_data
-    return SimulationResult(xs, boxes, du_dps, lambda_us)
+    xs, boxes, lambda_us = aux_data
+    return SimulationResult(xs, boxes, lambda_us)
 
 
 jax.tree_util.register_pytree_node(SimulationResult, flatten, unflatten)
```
maxentile, line 39: Ditto, probably unused since `grad(delta_G)` is deprecated / removed

timemachine/cpp/src/wrap_kernels.cpp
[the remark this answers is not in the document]
mcwitt: Thanks! I removed this in beed645.

slow_tests/test_benchmark.py
```diff
@@ -98,14 +98,6 @@ def benchmark(
         barostat=baro_impl,
     )
 
-    # initialize observables
-    if compute_du_dp_interval > 0:
```
maxentile, line 102: `compute_du_dp_interval` can be removed from `benchmark` signature

maxentile (approved):
Looks good!

(There might be a few remaining references to the deleted observable classes / new opportunities for further code deletion. Commented on these as a reminder for later, but these should not block the current PR.)

slow_tests/test_benchmark.py
mcwitt: Removed in 3db9cff

fe/estimator_abfe.py
mcwitt: Removed in 3ab3b17

fe/estimator.py
mcwitt: Removed in 3ab3b17

proteneer (approved):
lgtm! Nice set of changes.

slow_tests/test_rbfe.py
```diff
@@ -1,6 +1,7 @@
 import os
 from tempfile import TemporaryDirectory
 from unittest import TestCase
+import pytest
```
badisa, line 4: This still used?

mcwitt: Nope, thanks. Removed in ad9db70

badisa (changes requested):
I am running into NVRTC compilation errors trying to run the RABFE script with this

badisa:
Trying to run the RABFE script using the following
```python -O examples/validate_relative_binding.py --num_gpus 1 --property_units 'uM' --num_complex_conv_windows 2 --num_complex_windows 2 --num_solvent_conv_windows 2 --num_solvent_windows 2 --num_complex_equil_steps 100 --num_complex_prod_steps 100 --num_solvent_equil_steps 20 --num_solvent_prod_steps 20 --num_complex_preequil_steps 100 --blocker_name 163  --protein_pdb datasets/fep-benchmark/hif2a/5tbm_prepared.pdb --ligand_sdf datasets/fep-benchmark/hif2a/ligands.sdf 
```

I see the following error:

```ERROR:examples/validate_relative_binding.py:Error simulating Mol: 25
  File "examples/validate_relative_binding.py", line 456, in <module>
    runs.append(simulate_pair(epoch, blocker_mol, mol))
  File "examples/validate_relative_binding.py", line 348, in simulate_pair
    [aligned_mol_coords, ref_coords],
  File "/home/fyork/proj/timemachine/md/minimizer.py", line 143, in minimize_host_4d
    u_impls = bind_potentials(hgt, ff)
  File "/home/fyork/proj/timemachine/md/minimizer.py", line 33, in bind_potentials
    u_impls.append(bp.bound_impl(precision=np.float32))
  File "/home/fyork/proj/timemachine/timemachine/lib/potentials.py", line 36, in bound_impl
    return custom_ops.BoundPotential(self.unbound_impl(precision), self.params)
  File "/home/fyork/proj/timemachine/timemachine/lib/potentials.py", line 30, in unbound_impl
    return custom_ctor(*self.args)
RuntimeError: NVRTC error: NVRTC_ERROR_COMPILATION
---------------------------------------------------
--- JIT compile log for jit_program ---
---------------------------------------------------
__nv_name_map(2): error: identifier "k_add_ull_to_real_interpolated" is undefined

__nv_name_map(2): error: Error in parsing name expression for lowered name lookup. Input name expression was: " k_add_ull_to_real_interpolated"

2 errors detected in the compilation of "jit_program".
```

We should fix this and add a test to cover the usage of the jit code

badisa (approved):
Sorry, hadn't cleaned my environment before testing!

PR #557, opened 2022-01-07
Removes MDTraj and its use of Imaging

badisa:


fe/model_utils.py
```diff
@@ -53,12 +54,24 @@ def is_hydrogen(i):
     return masses
 
 
-# def generate_openmm_topology(objs, host_coords, out_filename):
-def generate_openmm_topology(objs, coords, out_filename=None):
+def image_molecule(mol_coords, box):
+    """Given a set of coordinates for a single molecule and a box will return
+    the coordinates wrapped into the periodic box.
+    """
```
mcwitt, line 60: Should we mention the expected shape of the inputs?

fe/model_utils.py
```diff
@@ -53,12 +54,24 @@ def is_hydrogen(i):
...
+    centroid = np.mean(mol_coords, axis=0)
+    box_diag = np.diag(box)
```
mcwitt, line 62: nit: what do you think about `np.diagonal` instead here? I find it potentially confusing that `np.diag` works on both 1- and 2-d arrays (and does a different thing in each case).

fe/model_utils.py
```diff
@@ -53,12 +54,24 @@ def is_hydrogen(i):
...
+    new_center = box_diag * np.floor(centroid / box_diag)
+    return mol_coords - new_center
```
mcwitt, line 64: I might be missing a subtlety, but this seems equivalent to using the `%` operator elementwise. Would the following work?
```suggestion
    return centroid % box_diag
```

fe/model_utils.py
```diff
@@ -53,12 +54,24 @@ def is_hydrogen(i):
...
+
+
+def generate_openmm_topology(objs, coords, out_filename=None, box=None):
     rd_mols = []
 
     mol_sizes = []
+    # Convert nm into angstrom
     coords = coords * 10
+    if box is not None:
+        box = box * 10
```
mcwitt, line 74: nit: not necessary for this PR, but would it make sense to introduce `box_angstroms` instead of  overwriting?

fe/model_utils.py
```diff
@@ -96,13 +108,20 @@ def generate_openmm_topology(objs, coords, out_filename=None):
         else:
             end_idx = offsets[mol_idx + 1]
             mol_pos = coords[start_idx:end_idx]
+        if box is not None:
+            mol_pos = image_molecule(mol_pos, box)
 
-        assert mol_pos.shape[0] == mol.GetNumAtoms()
+        assert (
+            mol_pos.shape[0] == mol.GetNumAtoms()
+        ), f"Coordinates shape don't match {mol_pos.shape[0]} != {mol.GetNumAtoms()}"
```
mcwitt, line 116: nit: could this raise a `ValueError` instead?

requirements.txt
```diff
@@ -10,7 +10,6 @@ jax==0.2.9
 jaxlib==0.1.61
 hilbertcurve==1.0.5
 networkx==2.5
-mdtraj==1.9.6
```
mcwitt, line 13: :tada:

mcwitt (commented):
lgtm! Just had some minor comments

fe/model_utils.py
mcwitt: Ah, never mind, I think I was mixing up `centroid` and `mol_coords`. My suggestion doesn't make sense.

badisa: `centroid % box_diag` returns the desired center for the molecule after moved, while subtracting the current center actually places the coordinates in the box so that they are centered correctly. Will expand the test to make that a bit more clear, as having it as a single set of coordinates allows this to be a reasonable way to do it.

tests/test_fe_utils.py
```diff
@@ -32,3 +34,16 @@ def test_extract_delta_Us_from_U_knk():
     )
 
     np.testing.assert_almost_equal(expected_delta_Us, test_delta_Us)
+
+
+def test_image_molecules():
+    box_size = 10
+    box = np.eye(3) * box_size
+
+    np.random.seed(2022)
```
maxentile, line 43: Nit: Should random seed be set elsewhere in test file? (The test currently doesn't seem to use random numbers)

Request: Might also be good to test more cases, esp:
* offset by `(k * box_size) + eps` for some +ve and -ve integers `k`

maxentile (commented):
Looks good! Just might expand the test slightly

tests/test_fe_utils.py
badisa: Woops, the random is an artifact of an earlier version of the test.

Will expand to test those cases

maxentile: One more test elaboration: could we assert here that pairwise distances are the same before and after imaging? (want to ensure `np.isclose(distance(x, box), distance(x_imaged, box))` for all tested cases `x`)

requirements.txt
proteneer: nice!

tests/test_fe_utils.py
```diff
@@ -32,3 +34,27 @@ def test_extract_delta_Us_from_U_knk():
     )
 
     np.testing.assert_almost_equal(expected_delta_Us, test_delta_Us)
+
+
+def test_image_molecules():
```
proteneer, line 39: I would definitely bullet-proof this test a bit:

1) Use non-cube boxes, i.e. have diag of box be different
2) Also add a test case with a real molecule (whose center has been shifted to outside of the box.

``` python
mol = # some mol from tests folder with 3D coordinates
# ensure initial starting coordinates are somewhere in the box


# randomly modify coordinates by
for _ in range(1000):

   # shift each direction randomly, we may need to go beyond one image since waters can float very far away
   x_shift = np.rand.randint(-5, 5)
   y_shift = np.rand.randint(-5, 5)
   z_shift = np.rand.randint(-5, 5)
   
   offset = np.array([x_shift*box_size_0, y_shift*box_size_1, z_shift*box_size_2])
   offset = np.expand_dims(offset, axis=0) # make this broad castable to [N,3]
   new_mol_conf = old_conf + offset

   imaged_mol = image_molecule(new_mol_conf, box)
   # assert image_mol has same coordinates as imaged_mol

```

tests/test_fe_utils.py
proteneer: I'd actually just check that the conformations are equivalent (or least separated by a translation) - we don't want this to have any rigid rotational component (which pairwise distances cannot detect).

proteneer (changes requested):
lgtm but test should be improved

proteneer (approved):
New changes to test look good!

maxentile (approved):

PR #558, opened 2022-01-07
Add method for potential-specific fixed- to floating-point conversion

mcwitt:
Moves conversions from fixed- to floating-point that currently happen in the pybind11 wrapper to a new virtual method `Potential::fixed_to_floating`.

The motivation is that, once we transition to returning `du_dp`s in fixed point, we'll need potential-specific logic in the conversion (since parameter derivatives have different scales) and it will become awkward to handle the cases in the pybind11 wrapper. Since the fixed-to-floating conversion for parameter derivatives is currently handled in the Potential implementations, we don't need to override `fixed_to_floating` in subclasses yet (but we will need to do this in #556).

### DIFF
--- a/timemachine/cpp/src/potential.cu
+++ b/timemachine/cpp/src/potential.cu
@@ -139,16 +139,16 @@ void Potential::fixed_to_float(
         du_dp_out[i] = du_dp[i]; // TODO: just a pass-thru for now; update when du_dp is fixed-point
     }
 
-    unsigned long long du_dl_sum_ull = 0;
+    unsigned long long du_dl_sum_fixed = 0;
     for (int i = 0; i < N; i++) {
-        du_dl_sum_ull += du_dl[i];
+        du_dl_sum_fixed += du_dl[i];
     }
-    *du_dl_sum = FIXED_TO_FLOAT<double>(du_dl_sum_ull);
+    *du_dl_sum = FIXED_TO_FLOAT<double>(du_dl_sum_fixed);
 
-    unsigned long long u_sum_ull = 0;
+    unsigned long long u_sum_fixed = 0;
     for (int i = 0; i < N; i++) {
-        u_sum_ull += u[i];
+        u_sum_fixed += u[i];
     }
-    *u_sum = FIXED_TO_FLOAT<double>(u_sum_ull);
+    *u_sum = FIXED_TO_FLOAT<double>(u_sum_fixed);
 }
 } // namespace timemachine
--- a/timemachine/cpp/src/potential.cu
+++ b/timemachine/cpp/src/potential.cu
@@ -19,6 +19,8 @@ void Potential::execute_host(
     unsigned long long *h_du_dl, //
     unsigned long long *h_u) {
 
+    const int &D = Potential::D;
+
     double *d_x;
     double *d_p;
     double *d_box;
@@ -90,6 +92,8 @@ void Potential::execute_host_du_dx(
     const double lambda, // [1]
     unsigned long long *h_du_dx) {
 
+    const int &D = Potential::D;
+
     double *d_x;
     double *d_p;
     double *d_box;
@@ -130,6 +134,8 @@ void Potential::fixed_to_float(
     double *du_dp_out,
     double *du_dl_sum,
     double *u_sum) {
+
+    const int &D = Potential::D;
 
     for (int i = 0; i < N * D; i++) {
         du_dx_out[i] = FIXED_TO_FLOAT<double>(du_dx[i]);
--- a/timemachine/cpp/src/potential.cu
+++ b/timemachine/cpp/src/potential.cu
@@ -1,5 +1,6 @@
 #include <iostream>
 
+#include "fixed_point.hpp"
 #include "gpu_utils.cuh"
 #include "potential.hpp"
 #include "surreal.cuh"
@@ -21,8 +22,6 @@ void Potential::execute_host(
     double *d_x;
     double *d_p;
     double *d_box;
-
-    const int D = 3;
 
     gpuErrchk(cudaMalloc(&d_x, N * D * sizeof(double)));
     gpuErrchk(cudaMemcpy(d_x, h_x, N * D * sizeof(double), cudaMemcpyHostToDevice));
@@ -95,8 +94,6 @@ void Potential::execute_host_du_dx(
     double *d_p;
     double *d_box;
 
-    const int D = 3;
-
     gpuErrchk(cudaMalloc(&d_x, N * D * sizeof(double)));
     gpuErrchk(cudaMemcpy(d_x, h_x, N * D * sizeof(double), cudaMemcpyHostToDevice));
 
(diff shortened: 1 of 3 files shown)


timemachine/cpp/src/potential.cu
```diff
@@ -22,8 +23,6 @@ void Potential::execute_host(
     double *d_p;
     double *d_box;
 
-    const int D = 3;
```
mcwitt, line 25: took a small step toward DRYing these definitions (moved to a `static const` the header)

timemachine/cpp/src/potential.cu
```diff
@@ -122,4 +119,36 @@ void Potential::execute_host_du_dx(
...
+    double *du_dp_out,
+    double *du_dl_sum,
+    double *u_sum) {
+
+    for (int i = 0; i < N * D; i++) {
+        du_dx_out[i] = FIXED_TO_FLOAT<double>(du_dx[i]);
+    }
+
+    for (int i = 0; i < P; i++) {
+        du_dp_out[i] = du_dp[i]; // TODO: just a pass-thru for now; update when du_dp is fixed-point
```
mcwitt, line 139: In #556 , the subclass implementations will add the appropriate conversion here

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -374,33 +374,21 @@ void declare_potential(py::module &m) {
 
                 std::vector<unsigned long long> du_dx(N * D);
                 std::vector<double> du_dp(P);
-
-                // initialize to zero for the accumulator
```
mcwitt, line 378: I didn't understand why we were zeroing these, since it seems like they're overwritten -- maybe it's no longer needed?

timemachine/cpp/src/potential.cu
proteneer: Makes sense! It's probably actually good to be explicit about the conversion for each type of potential, since each functional form will have different sized (and different ranges) of du_dps.

proteneer: Though I think du_dx and du_dls may not need too much specialization, since on average the forces are fairly well behaved across the potentials (for MD, we definitely have an upper bound, otherwise the integrator will fall apart). du_dps and maybe u are more finnicky and variable across different classes.

timemachine/cpp/src/potential.hpp
```diff
@@ -10,6 +10,8 @@ class Potential {
 public:
     virtual ~Potential(){};
 
+    static const int D = 3;
```
proteneer, line 13: may want to either give this a suffix D_ (to indicate its a class variable, even though its static), or namespace above with Potential::D (unscoped single letter variables can be pretty hard to track donw).

timemachine/cpp/src/potential.cu
mcwitt: > Though I think du_dx and du_dls may not need too much specialization

Makes sense. Maybe can pull out the conversion for parameter derivatives into a separate method so that subclasses can override just that.

timemachine/cpp/src/potential.hpp
mcwitt: Hmm, what about something like `ndim` to move away from single-level names with class scope?

> may want to either give this a suffix D_

Would that still be conventional if `D` is public? (in cases I've seen so far the trailing `_` is used for private variables -- we could also just make `D` private).

timemachine/cpp/src/wrap_kernels.cpp
badisa: Kind of partial to keeping them zero'd, its more defensive than risking not overwriting everything. Or at least a comment that explicitly makes it clear that you must overwrite all values

timemachine/cpp/src/potential.cu
```diff
@@ -122,4 +119,36 @@ void Potential::execute_host_du_dx(
     gpuErrchk(cudaFree(d_box));
 };
 
+void Potential::fixed_to_float(
```
badisa, line 122: Should this be templated to support float and double?

timemachine/cpp/src/wrap_kernels.cpp
mcwitt: Hmm, thinking about this a bit more, I'd almost prefer the opposite, i.e. functions assume that output pointers initially point to uninitialized memory unless explicitly documented otherwise. In this view, something like what `execute_device` does (assuming `u` is initialized and just adding my energy) would be the exception to the rule and documented.

I think this could make the code a bit more readable because unused initializations can give the wrong impression about what a function does (e.g. I didn't initially understand that `execute_host` overwrites its outputs). What do you think?

timemachine/cpp/src/potential.hpp
proteneer: `D` is fine too. Then maybejust modify the usages to `Potential::D` so scoping is clear? C++/C has the unfortunate behavior where any un-namespaced include (and their includes) is globally visible, so as a sanity let's explicit use the namespace?

timemachine/cpp/src/wrap_kernels.cpp
proteneer: There's two types of behavior that we should consider:

1) Accumulators such as forces, energies, and du_dls, where different potentials accumulate to the same buffer. These need to be initialized to zero for correctness. 

2) Non-accumulators that rely on the underlying code to set the contents.

Usually when I read code that doesn't initialize, I *assume* that its contents will be set directly by the underlying call using an assignment operator somewhere. If it initializes to 0, then I assume that the underlying code will be accumulated some how (eg. via += or atomicAdds)

badisa: For cases where performance aren't important (which I think of anything not CUDA or called in a loop) setting to 0 is safer. You get sane 'bad' values rather than values that might seem valid. Definitely spent a bunch of time tracking down junk initial memory values thinking they were incorrect computation or clashes in my system.

@proteneer distinction is definitely important but I can't say I have the right mindset to notice that difference on a quick read of the code.

timemachine/cpp/src/potential.cu
mcwitt: Resolving this for now; I'll consider splitting out the du_dp conversion into a separate method when it comes time to override this in #556 .

timemachine/cpp/src/potential.hpp
mcwitt: Makes sense. Would something like this work to make the scoping clear while minimizing line noise? df9bfda

timemachine/cpp/src/wrap_kernels.cpp
mcwitt: @badisa what do you think about initializing with a junk value other than 0 as @proteneer suggested? E.g. f9d970c

badisa: Looks good!

badisa (approved):

maxentile (approved):

proteneer (approved):

PR #559, opened 2022-01-10
Verify RABFE Mol Pairs

badisa:
Covers some of the issues run into with RABFE simulations.

Also adds logging for where exactly exceptions came from, which helped find the issues that this attempts to address

### DIFF
--- a/fe/model_utils.py
+++ b/fe/model_utils.py
@@ -4,7 +4,20 @@ import numpy as np
 
 from simtk.openmm import app
 from rdkit import Chem
-from rdkit.Geometry import Point3D
+
+
+def assert_mol_has_all_hydrogens(mol: Chem.Mol):
+    atoms = mol.GetNumAtoms()
+    mol_copy = Chem.AddHs(mol)
+    assert atoms == mol_copy.GetNumAtoms(), "Hydrogens missing for mol"
+
+
+def verify_rabfe_pair(mol: Chem.Mol, blocker: Chem.Mol):
+    assert_mol_has_all_hydrogens(mol)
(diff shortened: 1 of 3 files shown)


fe/model_utils.py
```diff
@@ -4,7 +4,20 @@
...
+    mol_copy = Chem.AddHs(mol)
+    assert atoms == mol_copy.GetNumAtoms(), "Hydrogens missing for mol"
+
+
+def verify_rabfe_pair(mol: Chem.Mol, blocker: Chem.Mol):
+    assert_mol_has_all_hydrogens(mol)
+    assert_mol_has_all_hydrogens(blocker)
+    mol_charge = Chem.GetFormalCharge(mol)
+    blocker_charge = Chem.GetFormalCharge(blocker)
+    assert mol_charge == blocker_charge, f"Formal charge disagrees: blocker: {blocker_charge:d} ligand: {mol_charge:d}"
```
maxentile, line 20: Might comment if there are conditions we should make this line more or less stringent.

(Most stringent: `mol_charge == 0, blocker_charge == 0`, less stringent: `mol_charge == blocker_charge`, `abs(mol_charge - blocker_charge) < 2` etc.)

maxentile (approved):

fe/model_utils.py
badisa: Based on discussions, I thought we wanted to avoid any 'charge hopping'.

mcwitt (commented):
lgtm!

proteneer (approved):
lgtm! may be good to also check for RMSD later to ensure alignment

PR #566, opened 2022-01-13
Write out SDF results for RABFE Script

badisa:
This makes three sets of changes

- Sets reasonable defaults for RABFE, as it is too easy to fat finger the inputs
- Moves the computation of the dG error into the RABFEResult object, moves the logic out of the script
- Writes out an SDF file, with each compound containing all of the fields from the RABFEResult object

Can't attach an SDF file here (zipped to support uploading to github)
[example_hif2a.zip](https://github.com/proteneer/timemachine/files/7863802/example_hif2a.zip)

One field we may we want to tear off is the AM1 charge cache that we keep on the mol.

### DIFF
--- a/examples/validate_relative_binding.py
+++ b/examples/validate_relative_binding.py
@@ -402,7 +402,7 @@ if __name__ == "__main__":
             "epoch": epoch,
         }
 
-    def predict_dG(results: dict):
+    def predict_dG(results: dict) -> RABFEResult:
         dG_complex_decouple, dG_complex_decouple_error = binding_model_complex_decouple.predict_from_futures(
             results["complex_decouple"][0],
             results["mol"],
@@ -433,19 +433,16 @@ if __name__ == "__main__":
         rabfe_result = RABFEResult(
             mol_name=mol_name,
             dG_complex_conversion=dG_complex_conversion,
+            dG_complex_conversion_error=dG_complex_conversion_error,
             dG_complex_decouple=dG_complex_decouple,
+            dG_complex_decouple_error=dG_complex_decouple_error,
             dG_solvent_conversion=dG_solvent_conversion,
+            dG_solvent_conversion_error=dG_solvent_conversion_error,
             dG_solvent_decouple=dG_solvent_decouple,
+            dG_solvent_decouple_error=dG_solvent_decouple_error,
         )
         rabfe_result.log()
-        dG_err = np.sqrt(
-            dG_complex_conversion_error ** 2
-            + dG_complex_decouple_error ** 2
-            + dG_solvent_conversion_error ** 2
-            + dG_solvent_decouple_error ** 2
-        )
-
-        return rabfe_result.dG_bind, dG_err
+        return rabfe_result
 
     runs = []
     for epoch in range(cmd_args.epochs):
@@ -479,7 +476,10 @@ if __name__ == "__main__":
                 print(f"Unable to find property {cmd_args.property_field}: {e}")
         print(f"Epoch: {epoch}, Processing Mol: {mol_name}, Label: {label_dG}")
         try:
-            dG, dG_err = predict_dG(run)
-            print(f"Epoch: {epoch}, Mol: {mol_name}, Predicted dG: {dG}, dG Err:" f" {dG_err}, Label: {label_dG}")
+            result = predict_dG(run)
+            print(
+                f"Epoch: {epoch}, Mol: {mol_name}, Predicted dG: {result.dG_bind}, dG Err:"
+                f" {result.dG_bind_err}, Label: {label_dG}"
+            )
         except Exception:
             logger.exception(f"Error processing Mol: {mol_name}")
--- a/examples/validate_relative_binding.py
+++ b/examples/validate_relative_binding.py
@@ -94,7 +94,7 @@ def cache_wrapper(cache_path: str, fxn: callable, overwrite: bool = False) -> ca
 
 
 if __name__ == "__main__":
-    default_output_path = f"rabfe_{datetime.datetime.now().strftime('%Y_%m_%d_%H_%M_%S')}.sdf"
+    default_output_path = f"rabfe_{datetime.datetime.utcnow().isoformat(timespec='seconds').replace(':', '_')}.sdf"
 
     parser = argparse.ArgumentParser(
         description="Relatively absolute Binding Free Energy Testing",
--- a/examples/validate_relative_binding.py
+++ b/examples/validate_relative_binding.py
@@ -119,50 +119,50 @@ if __name__ == "__main__":
     parser.add_argument("--num_gpus", type=int, help="number of gpus", default=get_gpu_count())
 
     parser.add_argument(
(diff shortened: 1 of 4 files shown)


fe/free_energy_rabfe.py
```diff
@@ -32,20 +36,28 @@ def log(self):
...
+            self.dG_solvent_decouple_error,
         )
 
     @classmethod
     def from_log(cls, log_line):
         """parse log line"""
         mol_name, rest = log_line.split("stage summary for mol: ")[-1].split(" dG_complex_conversion (K complex) ")
         tokens = rest.split()
-        value_strings = tokens[0], tokens[9], tokens[13], tokens[16]
+        value_strings = [tokens[i] for i in (0, 2, 11, 13, 17, 19, 22, 24)]
```
maxentile, line 60: * Seems easy to make a typo here
* Confirm: this will raise an error if attempting to read from an incompatible log version?

tests/test_free_energy_rabfe.py
```diff
@@ -31,7 +31,17 @@ def capture_print(closure):
...
+    result = RABFEResult(
+        "my mol",
+        1.0,
+        0.0,
+        2.0,
+        0.0,
+        3.0,
+        0.0,
+        4.0,
+        0.0,
```
maxentile, line 43: Might as well make all 8 values distinct

fe/free_energy_rabfe.py
```diff
@@ -61,6 +61,41 @@ def from_log(cls, log_line):
...
+            {
+                "dG_bind": self.dG_bind,
+                "dG_bind_err": self.dG_bind_err,
+            }
+        )
+        for field, val in results_dict.items():
+            if field == "mol_name":
+                field_name = "_Name"
+            else:
+                field_name = self._convert_field_to_sdf_field(field)
```
maxentile, line 96: Would move this branch into `_convert_field_to_sdf_field`

maxentile (approved):
Looks good -- minor comments

fe/free_energy_rabfe.py
badisa: I have confirmed that this raises an error if attempting to read from the older log version, would you like it as a test?

maxentile: Test probably isn't necessary -- thanks for confirming!

tests/test_free_energy_rabfe.py
badisa: Also added a nan as I know @proteneer had wanted to know if they round tripped

maxentile: Nice -- resolved in https://github.com/proteneer/timemachine/pull/566/commits/920806dbfd283f2fbfa499dd3ad8d011a159037c and https://github.com/proteneer/timemachine/pull/566/commits/08ce6bf4a335f69498f03bd17a1b99d7f912510c

fe/free_energy_rabfe.py
maxentile: Nice -- resolved in https://github.com/proteneer/timemachine/pull/566/commits/7d9255f508f552def69a751680bf86b170d73504

fe/free_energy_rabfe.py
```diff
@@ -32,23 +37,75 @@ def log(self):
...
+    @classmethod
+    def _convert_field_to_sdf_field(cls, field_name: str) -> str:
+        if field_name == "mol_name":
+            cleaned_name = "_Name"
+        else:
+            cleaned_name = field_name.replace("_", " ")
+        return cleaned_name
+
+    @classmethod
+    def from_mol(cls, mol):
```
mcwitt, line 86: nit: type annotation

maxentile:
>One field we may we want to tear off is the AM1 charge cache that we keep on the mol.

I'd be inclined to keep it -- doesn't bloat the file too much, could save headaches later?

examples/validate_relative_binding.py
```diff
@@ -457,29 +458,42 @@ def predict_dG(results: dict):
...
+                    concentration = float(mol.GetProp(cmd_args.property_field))
+
+                    if cmd_args.property_units == "uM":
+                        label_dG = convert_uM_to_kJ_per_mole(concentration)
+                    elif cmd_args.property_units == "nM":
+                        label_dG = convert_uM_to_kJ_per_mole(concentration / 1000)
+                    else:
+                        assert 0, "Unknown property units"
+                except Exception as e:
+                    print(f"Unable to find property {cmd_args.property_field}: {e}")
```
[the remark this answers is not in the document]
badisa: the utility of logger.exception is to print out the traceback, in this case the traceback isn't useful in the way that the others could be coming from anywhere

examples/validate_relative_binding.py
```diff
@@ -93,12 +94,15 @@ def fn(*args, **kwargs):
 
 
 if __name__ == "__main__":
+    default_output_path = f"rabfe_{datetime.datetime.now().strftime('%Y_%m_%d_%H_%M_%S')}.sdf"
```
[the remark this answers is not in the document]
badisa: This is the convention here: https://github.com/proteneer/timemachine/blob/master/examples/hif2a/fit_to_multiple_rbfes.py#L188

Don't have a preference as this really is just to avoid rerunning the script and overwriting an existing file

fe/free_energy_rabfe.py
```diff
@@ -32,23 +37,75 @@ def log(self):
...
+            return NotImplemented
+        equal = True
+        for field in fields(self):
+            self_val = getattr(self, field.name)
+            other_val = getattr(other, field.name)
+            # Python doesn't consider nan == nan to be true
+            if field.type is float and math.isnan(self_val) and math.isnan(other_val):
+                continue
+            equal &= self_val == other_val
+        return equal
```
mcwitt, line 75: extreme nit: what do you think about a more functional style here, using `all`?
```suggestion
        def field_eq(field):
            self_val = getattr(self, field.name)
            other_val = getattr(other, field.name)
            if field.type is float and math.isnan(self_val) and math.isnan(other_val):
                return True
            return self_val == other_val

        return all(field_eq(field) for field in fields(self))
```

fe/free_energy_rabfe.py
```diff
@@ -32,23 +37,75 @@ def log(self):
...
 
     @classmethod
     def from_log(cls, log_line):
         """parse log line"""
         mol_name, rest = log_line.split("stage summary for mol: ")[-1].split(" dG_complex_conversion (K complex) ")
         tokens = rest.split()
-        value_strings = tokens[0], tokens[9], tokens[13], tokens[16]
-        values = list(map(float, value_strings))
+        values = [float(tokens[i]) for i in (0, 2, 11, 13, 17, 19, 22, 24)]
         return RABFEResult(mol_name, *values)
```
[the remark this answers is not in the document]
badisa: Just going to remove this code entirely as we won't be needing it anymore

schmolly (approved):
LGTM!

mcwitt (approved):

PR #570, opened 2022-01-19
Experiment/decouple with ff 1 1 0

badisa:


fe/model_rabfe.py
```diff
@@ -559,7 +564,20 @@ def setup_topology(self, mol_a, mol_b):
...
     def setup_topology(self, mol):
-        top = topology.BaseTopologyConversion(mol, self.ff)
+        with open(os.path.join(os.path.dirname(ff_path), "params/smirnoff_1_1_0_ccc.py")) as f:
+            ff_handlers = deserialize_handlers(f.read())
+
+        ref_forcefield = Forcefield(ff_handlers)
+
+        ff_params = ref_forcefield.get_ordered_params()
+
+        def ff_conversion(mol: Chem.Mol) -> NDArray:
```
badisa, line 574: @maxentile This might need to be jnp rather than np.

If you have any suggestions on how to do this in a cleaner way let me know.

tests/test_rabfe_topology.py
```diff
@@ -107,10 +107,8 @@ def test_base_topology_standard_decoupling():
 def test_dual_topology_standard_decoupling():
 
     # this class is used in double decoupling stages of the RABFE protocol. It modifies the
-    # DualTopology class in two ways:
-    # 1) the torsions between non-ring atoms are turned off
```
badisa, line 111: This seemed wrong at this point, I hope that I am not mistaken

tests/test_rabfe_topology.py
```diff
@@ -139,6 +137,7 @@ def test_dual_topology_standard_decoupling():
     assert isinstance(nonbonded_potential, potentials.NonbondedInterpolated)
 
     expected_qlj = topology.standard_qlj_typer(mol_c)
+    expected_qlj[:, 0] = expected_qlj[:, 0] / 2  # charges should be halved
```
badisa, line 140: Doesn't change anything since the charges are 0 in the test case, but seemed best to cover the example

tests/test_rabfe.py
```diff
@@ -21,6 +27,193 @@
...
+            conv_params = conv_nonbonded_params[num_host_atoms * 2 :]
+            decouple_params = decouple_nonbonded_params[num_host_atoms * 2 :]
+
+            assert conv_params.shape[0] == decouple_params.shape[0] - blocker.GetNumAtoms() * 2
+
+            # Verify the dest params of conv match the src params of decouple
+            np.testing.assert_array_equal(
+                conv_params[len(conv_params) // 2 :],
+                decouple_params[len(decouple_params) // 2 + blocker.GetNumAtoms() :],
+            )
```
maxentile, line 120: Nice!

fe/model_rabfe.py
maxentile: I think the `np.concatenate` needs to be `jnp.concatenate` (contains an array conversion, which jax won't like when tracing), but otherwise looks fine!

tests/test_rabfe_topology.py
maxentile: Double-checking: there is a test specific to ring torsions https://github.com/proteneer/timemachine/blob/4e4b861002a018b822d1311e54fc3a2e2be4e9d1/tests/test_rabfe_topology.py#L15-L43 , but I'll need to dig a bit more to confirm

maxentile: Correct, since https://github.com/proteneer/timemachine/pull/473

PR #571, opened 2022-01-19
Extract bond charge correction functions

maxentile:
Breaks up `AM1CCCHandler.static_parameterize` [into three pieces](https://github.com/proteneer/timemachine/blob/26ed49bcff38c187f1d6f46fb6f0699dce5523d9/ff/handlers/nonbonded.py#L450-L454), and adds a test for each piece.

* [`compute_or_load_am1_charges(mol)`](https://github.com/proteneer/timemachine/blob/26ed49bcff38c187f1d6f46fb6f0699dce5523d9/ff/handlers/nonbonded.py#L166), tested [here](https://github.com/proteneer/timemachine/blob/26ed49bcff38c187f1d6f46fb6f0699dce5523d9/tests/test_handlers.py#L566)
* [`bond_smirks_matches(mol, smirks_list)`](https://github.com/proteneer/timemachine/blob/26ed49bcff38c187f1d6f46fb6f0699dce5523d9/ff/handlers/nonbonded.py#L186), tested [here](https://github.com/proteneer/timemachine/blob/26ed49bcff38c187f1d6f46fb6f0699dce5523d9/tests/test_handlers.py#L593)
* [`apply_bond_charge_corrections(initial_charges, bond_idxs, deltas)`](https://github.com/proteneer/timemachine/blob/26ed49bcff38c187f1d6f46fb6f0699dce5523d9/ff/handlers/nonbonded.py#L221), tested [here](https://github.com/proteneer/timemachine/blob/26ed49bcff38c187f1d6f46fb6f0699dce5523d9/tests/test_handlers.py#L621)

Suggests that arbitrary other trainable functions could in principle be used to generate the inputs for `apply_bond_charge_corrections`.


ff/handlers/nonbonded.py
```diff
@@ -7,7 +7,8 @@
 import pickle
 
 from rdkit import Chem
-from ff.handlers.utils import match_smirks, sort_tuple
+from ff.handlers.utils import sort_tuple, match_smirks as rd_match_smirks
+from ff.handlers.bcc_aromaticity import match_smirks as oe_match_smirks
```
maxentile, line 11: May be better to apply these name changes to the definitions, and move these both into the same file somewhere

ff/handlers/nonbonded.py
```diff
@@ -161,6 +162,82 @@ def parameterize_ligand(params, param_idxs):
...
+        # with OpenEye is in order
+        # https://github.com/openforcefield/openff-toolkit/issues/983
+        am1_charges = list(oe_assign_charges(mol, "AM1"))
+
+        mol.SetProp(cache_key, base64.b64encode(pickle.dumps(am1_charges)))
+
+    else:
+        am1_charges = pickle.loads(base64.b64decode(mol.GetProp(cache_key)))
+
+    return np.array(am1_charges)
```
maxentile, line 182: Note that `oe_assign_charges(mol, "AM1")` is conformer dependent. Because the cache is stored onboard an object with mutable conformers, it is possible for the cache to become stale.

For example, if the function is called repeatedly between events that modify `mol`'s conformers in-place:

```python
from rdkit import Chem

from ff.handlers.nonbonded import compute_or_load_am1_charges

# from https://github.com/proteneer/timemachine/blob/a1877dfa2337816c9e8b753f23d558cc05fbb6d9/tests/test_handlers.py#L532
smi = "Clc1c(Cl)c(Cl)c(-c2c(Cl)c(Cl)c(Cl)c(Cl)c2Cl)c(Cl)c1Cl"

def get_mol():
    mol = Chem.MolFromSmiles(smi)
    mol = Chem.AddHs(mol)
    return mol


def precompute_conformers():
    mol = get_mol()
    AllChem.EmbedMultipleConfs(mol)
    return list(mol.GetConformers())[::5]

precomputed_conformers = precompute_conformers()


def set_conformer(mol, i):
    mol.RemoveAllConformers()
    mol.AddConformer(precomputed_conformers[i])

mol_0 = get_mol()
set_conformer(mol_0, 0)
charges_0 = compute_or_load_am1_charges(mol_0)

mol_1 = get_mol()
set_conformer(mol_1, 1)
charges_1 = compute_or_load_am1_charges(mol_1)

print('q(conf_1) - q(conf_0)\n', charges_1 - charges_0)

# modify conformer in place: mol_0 now contains conf_1
set_conformer(mol_0, 1)

# will return outdated charges
expect_charges_1 = compute_or_load_am1_charges(mol_0)
print('q_cached_0(conf_1) - q(conf_0)\n', expect_charges_1 - charges_0) # zeros
```

To protect against this, we might want to modify the caching logic / implementation, so that the computed charges are a deterministic function of the `mol` and `conformer`.

ff/handlers/nonbonded.py
```diff
@@ -161,6 +162,82 @@ def parameterize_ligand(params, param_idxs):
...
+
+    for type_idx, smirks in enumerate(smirks_list):
+        matched_bonds = []
+
+        matches = oe_match_smirks(smirks, oemol)
+
+        for matched_indices in matches:
+            a, b = matched_indices[0], matched_indices[1]
+            forward_matched_bond = [a, b]
+            reverse_matched_bond = [b, a]
```
mcwitt, line 207: Does forward/reverse order of bonds matter in the result? If not, could this be simplified by putting the lesser index first?
```suggestion
            matched_bond = sorted([a, b])
```

tests/test_handlers.py
```diff
@@ -562,6 +563,83 @@ def test_am1_differences():
...
+
+        # arbitrary: duplicates okay, reversals okay, symmetry not required
+        n_directed_bonds = np.random.randint(50, 100)
+        bonds = np.random.randint(0, n_atoms, size=(n_directed_bonds, 2))
+        deltas = np.random.randn(n_directed_bonds)
+
+        final_charges = nonbonded.apply_bond_charge_corrections(initial_charges, bonds, deltas)
+        final_net_charge = np.sum(final_charges)
+
+        np.testing.assert_almost_equal(final_net_charge, initial_net_charge)
```
mcwitt, line 640: Would it be worth also checking that some charges are changed?

ff/handlers/nonbonded.py
mcwitt: I'm not sure I fully understand this, but it sounds like modifying `cache_key` to include a substring that identifies the conformer could be a solution?

mcwitt (approved):
Looks good! Just had a couple of minor questions/comments.

ff/handlers/nonbonded.py
maxentile: The order does matter here, I think!

E.g. for a pair of atoms that is assigned this type

https://github.com/proteneer/timemachine/blob/d7acd339a681618fd2819b072f48f5aaafcf3b1b/ff/charges.py#L3

the partial charge of the carbon ("#6X4") will be adjusted in the [positive](https://github.com/proteneer/timemachine/blob/d1ed0e67ebe7a4b898b5104c3ab18bcfa81bd929/ff/handlers/nonbonded.py#L233) direction, but the partial charge of the sulfur "#16X1-1" will be adjusted in the [negative](https://github.com/proteneer/timemachine/blob/d1ed0e67ebe7a4b898b5104c3ab18bcfa81bd929/ff/handlers/nonbonded.py#L234) direction. Will add a note to clarify

ff/handlers/nonbonded.py
```diff
@@ -161,6 +162,82 @@ def parameterize_ligand(params, param_idxs):
...
+
+            already_matched = forward_matched_bond in matched_bonds or reverse_matched_bond in matched_bonds
+            already_assigned = forward_matched_bond in bond_idxs or reverse_matched_bond in bond_idxs
+
+            if not (already_matched or already_assigned):
```
mcwitt, line 212: Unless I'm missing something, `already_matched` implies `already_assigned`, so it seems like the former is not needed (nor is `matched_bonds`)?

tests/test_handlers.py
maxentile: Agreed -- addressed in https://github.com/proteneer/timemachine/pull/571/commits/ae9ebec7f37beb632c4aee98b42be2b3753c9a52 !

ff/handlers/nonbonded.py
maxentile: Added notes about this in two places: https://github.com/proteneer/timemachine/pull/571/commits/ff2c6b154e50991952cdc10911daf912a9fa30c4

ff/handlers/nonbonded.py
maxentile: Hmm, I think you're right -- will confirm!

Initially just wanted to group and name the conditions checked in https://github.com/proteneer/timemachine/blob/d7acd339a681618fd2819b072f48f5aaafcf3b1b/ff/handlers/nonbonded.py#L392-L397

tests/test_handlers.py
```diff
@@ -562,6 +563,84 @@ def test_am1_differences():
...
+    fresh_am1_charges = [nonbonded.compute_or_load_am1_charges(mol) for mol in all_mols]
+
+    # expect each mol to have AM1 cache now
+    for mol in all_mols:
+        assert mol.HasProp(cache_key)
+
+    # expect the same charges as the first time around
+    cached_am1_charges = [nonbonded.compute_or_load_am1_charges(mol) for mol in all_mols]
+    for (fresh, cached) in zip(fresh_am1_charges, cached_am1_charges):
+        np.testing.assert_allclose(fresh, cached)
```
badisa, line 590: Are these not exactly the same?

badisa (approved):

ff/handlers/nonbonded.py
maxentile: Applied in https://github.com/proteneer/timemachine/pull/571/commits/4de08487903e12e9eaed524bc2a986895714ebda

tests/test_handlers.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/571/commits/0dcf87ba0f5371d03cf7d23fadc1606e5a069f2e

ff/handlers/nonbonded.py
mcwitt: Makes sense, thanks for clarifying!

tests/test_handlers.py
```diff
@@ -562,6 +563,84 @@ def test_am1_differences():
...
+    fresh_am1_charges = [nonbonded.compute_or_load_am1_charges(mol) for mol in all_mols]
+
+    # expect each mol to have AM1 cache now
+    for mol in all_mols:
+        assert mol.HasProp(cache_key)
+
+    # expect the same charges as the first time around
+    cached_am1_charges = [nonbonded.compute_or_load_am1_charges(mol) for mol in all_mols]
+    for (fresh, cached) in zip(fresh_am1_charges, cached_am1_charges):
+        assert (fresh == cached).all()
```
mcwitt, line 590: nit: would `np.testing.assert_array_equal` be better here? (might give more helpful failure message?)

ff/handlers/nonbonded.py
```diff
@@ -161,6 +162,89 @@ def parameterize_ligand(params, param_idxs):
...
+
+        for matched_indices in matches:
+            a, b = matched_indices[0], matched_indices[1]
+            forward_matched_bond = [a, b]
+            reverse_matched_bond = [b, a]
+
+            already_assigned = forward_matched_bond in bond_idxs or reverse_matched_bond in bond_idxs
+
+            if not already_assigned:
+                bond_idxs.append(forward_matched_bond)
```
maxentile, line 214: @proteneer : could you double-check if there's anything lost by applying this simplification? 

Previously: https://github.com/proteneer/timemachine/blob/466dde7af309aee942e863403db1d738ef52e2e6/ff/handlers/nonbonded.py#L392-L401

ff/handlers/nonbonded.py
```diff
@@ -161,6 +162,89 @@ def parameterize_ligand(params, param_idxs):
...
+    return np.array(bond_idxs), np.array(type_idxs)
+
+
+def apply_bond_charge_corrections(initial_charges, bond_idxs, deltas):
+    """For an arbitrary collection of ordered bonds and associated increments `(a, b, delta)`,
+    update `charges` by `charges[a] += delta`, `charges[b] -= delta`
+
+    Notes
+    -----
+    * preserves sum(initial_charges) for arbitrary values of bond_idxs or deltas
```
maxentile, line 226: @proteneer : Could you double-check if there are any constraints on the inputs to this function?

(An antisymmetry constraint might have been suggested. Something along the lines of "for each bond `(a, b)`, we must provide either `(a, b, +delta)` or `(b, a, -delta)`.")

ff/handlers/nonbonded.py
proteneer: This looks correct, but I didn't write this originally (it was ported from Simon's BCC code originally). But semantically it seems to simply be: "matched either fwd or rev" and "not already assigned". At some point we may want to be consistent with the rest of OFF and have it "last match wins", though I'm not sure if simply reversing the SMIRKS patterns would accomplish this...

ff/handlers/nonbonded.py
proteneer: Should just sanity check with assertions that initial charges and final charges:

1) sum up to an integer
2) sum up to the same integer

proteneer: Maybe also sanity check that bond_idxs are unique? Though that probably doesn't affect correctness but more likely an indication of user error...

proteneer (approved):
LGTM! Save for some minor clerical comments.

ff/handlers/nonbonded.py
maxentile: Charge sum preservation is logically guaranteed and was tested with randomized inputs (including duplicates, etc.) here https://github.com/proteneer/timemachine/blob/71734541fa88054b8b121ad47cfa97bc5b561213/tests/test_handlers.py#L621-L641

Runtime assertions added in https://github.com/proteneer/timemachine/pull/571/commits/4984f6d5a08b961a35d558f3c5b05ab90b8790b4 , randomized tests restricted to integral net charge in https://github.com/proteneer/timemachine/pull/571/commits/00bc62871d3480ba22265ead71c731256f85b9ff , and warnings for duplicate bond_idxs added in https://github.com/proteneer/timemachine/pull/571/commits/26ed49bcff38c187f1d6f46fb6f0699dce5523d9 .

maxentile: https://github.com/proteneer/timemachine/pull/571#discussion_r791309783

Ahh, shoot -- should have read this as "integer multiple of sqrt(ONE_4PI_EPS0)" not "integer", as @badisa identified and fixed in https://github.com/proteneer/timemachine/pull/680

Issue #569, opened 2022-01-19
Fix memory leaks


mcwitt:
While testing for memory leaks in #542 using `cuda-memcheck --leak-check full`, I found existing leaks in master while running `test_nonbonded`.

```
========= LEAK SUMMARY: 7536 bytes leaked in 14 allocations
========= ERROR SUMMARY: 14 errors
```

Note: `--leak-check` requires calling `cudaDeviceReset()` at the end as in https://github.com/proteneer/timemachine/pull/542/commits/9b92c89131d6af471bf37d901107774b31a1e818

mcwitt:
Closed by #574

PR #574, opened 2022-01-21
Fix memory leak, add leak checking to CI

mcwitt:
Addresses  #569 

- Adds global test fixture calling `cuda_device_reset()` after each test. Each call triggers `cuda-memcheck` to detect any leaks
- Adds `--leak-check full --error-exitcode 1` to the `cuda-memcheck` invocation in CI. The effect is CI failure on any leak detection
- Fixes a memory leak in `barostat.cu` ([a5de263](https://github.com/proteneer/timemachine/pull/574/commits/a5de263aac7f5c925fc84934c119362a35c9f5e0))


Makefile
```diff
@@ -20,5 +20,5 @@ clean:
 ci:
 	pre-commit run --all-files --show-diff-on-failure && \
 	export PYTHONPATH=$(MKFILE_DIR) && \
-	cuda-memcheck pytest $(PYTEST_CI_ARGS) tests/ && \
+	cuda-memcheck --leak-check full pytest $(PYTEST_CI_ARGS) tests/ && \
```
badisa, line 23: Should we add `--error-exitcode 1`? Right now this will report memory leaks, which seems to be in one of the tests already, but doesn't trigger a failure

mcwitt: Good call, I didn't realize this was necessary. Added in c03150e.

Also interesting that we get a [failure](https://jenkins.relaytxinternal.com/job/timemachine/job/PR-574/1/display/redirect) even without this. This seems like an independent issue possibly exposed by adding the `cudaDeviceReset()` calls?

mcwitt: > Also interesting that we get a failure even without this

This looks to be resolved by 5cc6dd5. In an offline debugging session, we figured out that `cudaDeviceReset()` was running before GC (and hence destructors) in some cases. Adding an explicit `gc.collect()` before calling `custom_ops.cuda_device_reset()` seems to be a fix.

mcwitt:
Before 8d06402:
```
[10.78.114.37] out: ========= LEAK SUMMARY: 7617 bytes leaked in 32 allocations
[10.78.114.37] out: ========= ERROR SUMMARY: 32 errors
```
and after:
```
[10.78.114.37] out: ========= LEAK SUMMARY: 81 bytes leaked in 18 allocations
[10.78.114.37] out: ========= ERROR SUMMARY: 18 errors
```
The remaining leaks appear to be in `test_free_energy_estimator` and `test_free_energy_estimator_with_endpoint_correction`.

proteneer:
Found some additional leaks caused by the barostat's destructor (changes pushed)

tests/conftest.py
```diff
@@ -0,0 +1,20 @@
+# This file defines fixtures shared across multiple test modules.
+# https://docs.pytest.org/en/latest/how-to/fixtures.html#scope-sharing-fixtures-across-classes-modules-packages-or-session
```
maxentile, line 2: 👍

timemachine/cpp/src/barostat.cu
```diff
@@ -125,7 +125,10 @@ MonteCarloBarostat::~MonteCarloBarostat() {
     gpuErrchk(cudaFree(d_u_buffer_));
     gpuErrchk(cudaFree(d_init_u_));
     gpuErrchk(cudaFree(d_final_u_));
-    gpuErrchk(cudaFree(d_rand_)) gpuErrchk(cudaFree(d_length_scale_));
+    gpuErrchk(cudaFree(d_rand_));
+    gpuErrchk(cudaFree(d_length_scale_));
+    gpuErrchk(cudaFree(d_sum_storage_));
+    gpuErrchk(cudaFree(d_volume_));
```
maxentile, line 131: Ahh -- these had been missed for a long time!

(cc: https://github.com/proteneer/timemachine/issues/577 )

maxentile (approved):

badisa (approved):

Issue #565, opened 2022-01-12
Refactor `fe` module


maxentile:
Merge near-duplicate definitions between `estimator.py` and `estimator_abfe.py`. (Similarly for  `free_energy.py` and `free_energy_rabfe.py`, and to a lesser extent `model.py` and `model_rabfe.py`.)

- [ ] Probably want just one definition of `SimulationResult` https://github.com/proteneer/timemachine/pull/556#discussion_r783417969
- [ ] Review repeated definitions for `BaseFreeEnergy`, `AbsoluteFreeEnergy`, `RelativeFreeEnergy` between `free_energy.py` and `free_energy_rabfe.py`

Usability improvements:
- [ ] Provide named index arrays `top.host_idxs`, `top.mol_a_idxs`, etc. https://github.com/proteneer/timemachine/pull/575#discussion_r792002492 , remove index computations from user code where possible
- [ ] Provide streamlined way to generate alchemical `sys_params` -- currently requires a nearly complete simulation setup https://github.com/proteneer/timemachine/pull/575#discussion_r792007855
- [ ] Possibly annotate which potentials / parameters / etc are alchemically modified, to make it easier to verify thermodynamic cycles https://github.com/proteneer/timemachine/pull/575#discussion_r792017076

Flexibility improvements:
- [ ] Decouple thermodynamic cycle definition from estimation choices. Decouple both from parallel execution policy.

maxentile:
Another feature that would be useful to support in a later refactor of the `fe` module:

- [ ] Ability to initialize simulations from saved checkpoints

maxentile:
Closing -- details mostly stale, intent partially addressed in https://github.com/proteneer/timemachine/pull/936 , https://github.com/proteneer/timemachine/pull/961 , etc.

PR #575, opened 2022-01-21
Adds test to verify Conversion Params agree with Decouple Params

badisa:
These tests was suggested by @maxentile and it seems important to verifying the assumptions we make about our thermodynamic cycle for the RABFE protocol

### DIFF
--- a/tests/test_rabfe.py
+++ b/tests/test_rabfe.py
@@ -4,11 +4,17 @@ from unittest import TestCase
 import numpy as np
 
 from md import builders, minimizer
-from fe.model_rabfe import RelativeBindingModel, AbsoluteConversionModel
+from fe.model_rabfe import RelativeBindingModel, AbsoluteConversionModel, AbsoluteStandardHydrationModel
 from fe.free_energy import construct_lambda_schedule
-from fe.free_energy_rabfe import setup_relative_restraints_by_distance, get_romol_conf
+from fe.free_energy_rabfe import (
+    setup_relative_restraints_by_distance,
+    get_romol_conf,
+    RelativeFreeEnergy,
+    AbsoluteFreeEnergy,
+)
 from fe.frames import all_frames
 from timemachine.potentials import rmsd
+from timemachine.lib.potentials import NonbondedInterpolated, Nonbonded
 from testsystems.relative import hif2a_ligand_pair
 
 from parallel.client import CUDAPoolClient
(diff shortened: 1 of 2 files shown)


tests/test_rabfe.py
```diff
@@ -21,6 +27,193 @@
...
+            conv_params = conv_nonbonded_params[num_host_atoms * 2 :]
+            decouple_params = decouple_nonbonded_params[num_host_atoms * 2 :]
+
+            assert conv_params.shape[0] == decouple_params.shape[0] - blocker.GetNumAtoms() * 2
+
+            # Verify the dest params of conv match the src params of decouple
+            np.testing.assert_array_equal(
+                conv_params[len(conv_params) // 2 :],
+                decouple_params[len(decouple_params) // 2 + blocker.GetNumAtoms() :],
+            )
```
maxentile, line 120: Verified some of these index calculations but ran out of time with others
* `conv_params` is assumed concatenated in the order `[src, dest]`, verified consistent with [definition in BaseTopologyConversion](https://github.com/proteneer/timemachine/blob/06e2512425e70278538f1e440e9f5dd83b3af4af/fe/topology.py#L364)
* `decouple_params` is assumed concatenated in the order `[dest, blocker, src]`, but I'm having trouble inferring that from the definitions in [DualTopologyStandardDecoupling](https://github.com/proteneer/timemachine/blob/06e2512425e70278538f1e440e9f5dd83b3af4af/fe/topology.py#L601) or [HostGuestTopology](https://github.com/proteneer/timemachine/blob/06e2512425e70278538f1e440e9f5dd83b3af4af/fe/topology.py#L201-L255)

Strongly prefer if alchemical topology classes provided more hint of the memory layout that users must be aware of -- maybe named index arrays `top.host_idxs`, `top.mol_a_idxs`,  etc. -- will move this comment to https://github.com/proteneer/timemachine/issues/565

tests/test_rabfe.py
```diff
@@ -21,6 +27,193 @@
...
+            dt,
+            10,
+            50,
+            frame_filter=all_frames,
+        )
+
+        conv_topo = conv_model.setup_topology(ligand)
+        conv_ref = AbsoluteFreeEnergy(ligand, conv_topo)
+
+        conv_unbound_potentials, conv_sys_params, _ = conv_ref.prepare_host_edge(ff_params, conv_model.host_system)
```
maxentile, line 91: Observation: Dozens of lines of simulation detail (timestep, parallel client, frame filtering policy, lambda schedule, temperature, pressure, etc.) must be specified n order to obtain the `sys_params` needed for this test.

Ideally the `fe` module would allow a more direct way to prepare these -- will move this comment to https://github.com/proteneer/timemachine/issues/565

tests/test_rabfe.py
```diff
@@ -21,6 +27,193 @@
...
+
+        assert len(conv_sys_params) == len(decouple_sys_params)
+        seen_nonbonded = False
+        for i, decouple_pot in enumerate(decouple_unbound_potentials):
+            if not isinstance(decouple_pot, NonbondedInterpolated):
+                continue
```
maxentile, line 97: I think that's correct https://github.com/proteneer/timemachine/pull/570#discussion_r788237639 , but it's possible this test would need to be modified in future if RABFE is modified to soften torsions etc.

maxentile (approved):
Looks good! This should improve confidence.

(Noted some awkwardness for later `fe` module refactor.)

tests/test_rabfe.py
badisa: Will add a comment to explain it a bit more, as I think this is useful context

tests/test_rabfe.py
badisa: There is likely a more direct route that could be taken already. I tested at this level as it is the level at which I think about our thermo cycle. Say the AbsoluteConversionModel changes its topology, this test would capture that while testing it at the topology level would not.

proteneer: Since the conversion/decouple only affect ligand parameters - maybe there's a way to refactor this later on to test only on the ligand?

proteneer: Though having a system level test is still a good idea!

PR #576, opened 2022-01-22
Add DeviceBuffer, refactor execute_host

mcwitt:
Prototypes an RAII-like approach to managing device memory and uses it to fix a memory leak in `Potential::execute_host` (where device memory allocated for local variables is not freed if `execute_device` throws an exception).

A simpler fix for the above issue is 8d06402, but the approach in this PR might be desirable in the long run.

Note: there at least a few other places in the code that I think would benefit from `DeviceBuffer`, e.g.
- [Potential::execute_host_du_dx](https://github.com/proteneer/timemachine/blob/6593d22a23709b52117b3ff99f78d032c216ccb9/timemachine/cpp/src/potential.cu#L87)
- [BoundPotential::execute_host](https://github.com/proteneer/timemachine/blob/10f8d2b6ddd6d89807436f24a329774e9c99bf98/timemachine/cpp/src/bound_potential.cu#L20)
- [Context::multiple_steps](https://github.com/proteneer/timemachine/blob/10f8d2b6ddd6d89807436f24a329774e9c99bf98/timemachine/cpp/src/context.cu#L56)
- [Context::multiple_steps_U](https://github.com/proteneer/timemachine/blob/10f8d2b6ddd6d89807436f24a329774e9c99bf98/timemachine/cpp/src/context.cu#L131)

I opted to convert just `Potential::execute_host` for now as a proof of concept; once the interface is solidified, I'm open to converting the others to use DeviceBuffer as well in this PR or a follow-up.

#### Todo
- [x] Open enhancement issue to refactor memory management more broadly using `DeviceBuffer` (#577)

### DIFF
--- a/timemachine/cpp/src/device_buffer.cu
+++ b/timemachine/cpp/src/device_buffer.cu
@@ -15,6 +15,14 @@ DeviceBuffer<T>::DeviceBuffer(const std::size_t length) : size(length * sizeof(T
 
 template <typename T> DeviceBuffer<T>::~DeviceBuffer() { gpuErrchk(cudaFree(data)); }
 
+template <typename T> void DeviceBuffer<T>::copy_from(const T *host_buffer) const {
+    gpuErrchk(cudaMemcpy(data, host_buffer, size, cudaMemcpyHostToDevice));
+}
+
+template <typename T> void DeviceBuffer<T>::copy_to(T *host_buffer) const {
+    gpuErrchk(cudaMemcpy(host_buffer, data, size, cudaMemcpyDeviceToHost));
+}
+
 template class DeviceBuffer<double>;
 template class DeviceBuffer<unsigned long long>;
 } // namespace timemachine
--- a/timemachine/cpp/src/device_buffer.cu
+++ b/timemachine/cpp/src/device_buffer.cu
@@ -4,14 +4,14 @@
 
 namespace timemachine {
 
-template <typename T> T *allocate(const std::size_t size) {
+template <typename T> T *allocate(const std::size_t length) {
     T *buffer;
-    gpuErrchk(cudaMalloc(&buffer, size));
+    gpuErrchk(cudaMalloc(&buffer, length * sizeof(T)));
     return buffer;
 }
 
 template <typename T>
-DeviceBuffer<T>::DeviceBuffer(const std::size_t length) : size(length * sizeof(T)), data(allocate<T>(size)) {}
+DeviceBuffer<T>::DeviceBuffer(const std::size_t length) : size(length * sizeof(T)), data(allocate<T>(length)) {}
 
 template <typename T> DeviceBuffer<T>::~DeviceBuffer() { gpuErrchk(cudaFree(data)); }
 
--- a/timemachine/cpp/src/device_buffer.cu
+++ b/timemachine/cpp/src/device_buffer.cu
@@ -4,16 +4,16 @@
 
 namespace timemachine {
 
-template <typename T>
-DeviceBuffer<T>::DeviceBuffer(const std::size_t length) : size(length * sizeof(T)), data(allocate_(size)) {}
-
-template <typename T> DeviceBuffer<T>::~DeviceBuffer() { gpuErrchk(cudaFree(data)); }
-
-template <typename T> T *DeviceBuffer<T>::allocate_(const std::size_t size) {
+template <typename T> T *allocate(const std::size_t size) {
     T *buffer;
     gpuErrchk(cudaMalloc(&buffer, size));
     return buffer;
 }
+
+template <typename T>
+DeviceBuffer<T>::DeviceBuffer(const std::size_t length) : size(length * sizeof(T)), data(allocate<T>(size)) {}
+
+template <typename T> DeviceBuffer<T>::~DeviceBuffer() { gpuErrchk(cudaFree(data)); }
 
 template class DeviceBuffer<double>;
 template class DeviceBuffer<unsigned long long>;
--- a/timemachine/cpp/src/device_buffer.cu
+++ b/timemachine/cpp/src/device_buffer.cu
@@ -13,11 +13,11 @@ template <typename T> DeviceBuffer<T>::~DeviceBuffer() {
     }
 }
 
-template <typename T> void DeviceBuffer<T>::ensure_allocated() {
+template <typename T> void DeviceBuffer<T>::allocate() {
     if (!data_) {
         gpuErrchk(cudaMalloc(&data_, size));
     } else {
-        throw std::runtime_error("attempted to reallocate an allocated buffer");
+        throw std::runtime_error("called allocate on an allocated buffer");
     }
 }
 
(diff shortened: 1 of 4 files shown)


timemachine/cpp/src/device_buffer.cu
```diff
@@ -0,0 +1,36 @@
...
+#include <cstddef>
+
+namespace timemachine {
+
+template <typename T>
+DeviceBuffer<T>::DeviceBuffer(const std::size_t length) : size(length * sizeof(T)), data_(nullptr) {}
+
+template <typename T> DeviceBuffer<T>::~DeviceBuffer() {
+    if (data_) {
+        gpuErrchk(cudaFree(data_));
```
mcwitt, line 12: Unfortunately this will make the context reported `gpuErrchk` much less useful

timemachine/cpp/src/device_buffer.hpp
```diff
@@ -0,0 +1,25 @@
...
+namespace timemachine {
+
+template <typename T> class DeviceBuffer {
+
+public:
+    DeviceBuffer(const std::size_t length);
+
+    ~DeviceBuffer();
+
+    const size_t size;
```
proteneer, line 13: I would probably make this a bit more C++ safe, right now the caller can just directly set this. You probably want to move this to a private member, and *only* allow for a `size()` function that returns the size, but disable the setter (by simply not implementing one. I can't think of a good use case for why we'd want to the setter

proteneer: Actually sorry I confused, I didn't realize this was `const` - which means it can only be set in the constructor

timemachine/cpp/src/device_buffer.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+    void ensure_allocated();
+
+    T *data();
```
proteneer, line 17: i'd also make a const version of this for read-only access

`const T *data();`

timemachine/cpp/src/potential.cu
```diff
@@ -23,66 +25,63 @@ void Potential::execute_host(
...
-    gpuErrchk(cudaMemcpy(d_box, h_box, D * D * sizeof(double), cudaMemcpyHostToDevice));
+    gpuErrchk(cudaMemcpy(d_x.data, h_x, d_x.size, cudaMemcpyHostToDevice));
+    gpuErrchk(cudaMemcpy(d_p.data, h_p, d_p.size, cudaMemcpyHostToDevice));
+    gpuErrchk(cudaMemcpy(d_box.data, h_box, d_box.size, cudaMemcpyHostToDevice));
 
-    unsigned long long *d_du_dx = nullptr;
-    unsigned long long *d_du_dp = nullptr;
-    unsigned long long *d_du_dl = nullptr;
-    unsigned long long *d_u = nullptr;
+    std::unique_ptr<DeviceBuffer<unsigned long long>> d_du_dx;
```
mcwitt, line 36: `unique_ptr` will automatically call the referenced `DeviceBuffer`'s destructor when the pointer goes out of scope.

timemachine/cpp/src/device_buffer.cu
```diff
@@ -0,0 +1,20 @@
...
+
+namespace timemachine {
+
+template <typename T>
+DeviceBuffer<T>::DeviceBuffer(const std::size_t length) : size(length * sizeof(T)), data(allocate_(size)) {}
+
+template <typename T> DeviceBuffer<T>::~DeviceBuffer() { gpuErrchk(cudaFree(data)); }
+
+template <typename T> T *DeviceBuffer<T>::allocate_(const std::size_t size) {
+    T *data;
```
badisa, line 13: Should this be a different name, perhaps `buffer`? It is confusing that data is both a private member and a free variable in this context.

timemachine/cpp/src/potential.cu
proteneer: nice! that works too

proteneer (approved):

timemachine/cpp/src/potential.cu
```diff
@@ -23,66 +25,63 @@ void Potential::execute_host(
...
-    gpuErrchk(cudaMemcpy(d_p, h_p, P * sizeof(double), cudaMemcpyHostToDevice));
+    DeviceBuffer<double> d_x(N * D);
+    DeviceBuffer<double> d_p(P);
+    DeviceBuffer<double> d_box(D * D);
 
-    gpuErrchk(cudaMalloc(&d_box, D * D * sizeof(double)));
-    gpuErrchk(cudaMemcpy(d_box, h_box, D * D * sizeof(double), cudaMemcpyHostToDevice));
+    gpuErrchk(cudaMemcpy(d_x.data, h_x, d_x.size, cudaMemcpyHostToDevice));
+    gpuErrchk(cudaMemcpy(d_p.data, h_p, d_p.size, cudaMemcpyHostToDevice));
+    gpuErrchk(cudaMemcpy(d_box.data, h_box, d_box.size, cudaMemcpyHostToDevice));
```
maxentile, line 34: Would it be useful to move this line into a `DeviceBuffer::to_device` method or similar?

timemachine/cpp/src/potential.cu
```diff
@@ -23,66 +25,63 @@ void Potential::execute_host(
...
+        d_du_dp ? d_du_dp->data : nullptr,
+        d_du_dl ? d_du_dl->data : nullptr,
+        d_u ? d_u->data : nullptr,
+        static_cast<cudaStream_t>(0));
 
     // outputs
     if (h_du_dx) {
-        gpuErrchk(cudaMemcpy(h_du_dx, d_du_dx, N * D * sizeof(*h_du_dx), cudaMemcpyDeviceToHost));
-        gpuErrchk(cudaFree(d_du_dx));
+        gpuErrchk(cudaMemcpy(h_du_dx, d_du_dx->data, d_du_dx->size, cudaMemcpyDeviceToHost));
```
maxentile, line 74: Would it be useful to move this line into a `DeviceBuffer::to_host` method or similar?

maxentile (approved):
Nice! This seems like an improvement in safety and readability -- reducing the requirement for repetitive paired sequences of malloc / free CUDA runtime API calls by moving them into a constructor / destructor pair.

timemachine/cpp/src/device_buffer.cu
mcwitt: Agreed, fixed in 6bcef8b.

timemachine/cpp/src/potential.cu
mcwitt: Nice idea! I implemented `DeviceBuffer::copy_from` and `DeviceBuffer::copy_to` in 8d982bf, but am still undecided as to whether I actually like these names better than `to_device` and `to_host` (I guess I found the latter slightly awkward as members of `DeviceBuffer`, but the explicitness is also nice)

timemachine/cpp/src/device_buffer.cu
mcwitt: Added a todo comment for now.

timemachine/cpp/src/potential.cu
maxentile: I prefer the way you did it: `copy_from` / `copy_to`!

timemachine/cpp/src/potential.cu
mcwitt: Added in 8d982bf

maxentile (approved):

badisa (approved):

PR #545, opened 2021-12-08
Feature/neighborlist on host ligand

badisa:
This PR extends our Neighborlist to support construct from two sets of coordinates. This will be necessary for eventuating host-ligand interactions.

The performance is unchanged from what is currently on master.

Current Benchmarks on A4000
```
CUDA Kernel Statistics:

 Time(%)  Total Time (ns)  Instances  Average (ns)  Minimum (ns)  Maximum (ns)  StdDev (ns)                                                  Name                                                
 -------  ---------------  ---------  ------------  ------------  ------------  -----------  ----------------------------------------------------------------------------------------------------
    64.6        1,808,224         12     150,685.3         7,680       294,208    136,109.5  void k_find_blocks_with_ixns<double>(int, const double *, const double *, const double *, const dou…
    20.6          575,296         12      47,941.3         5,024       101,120     41,874.3  void k_find_blocks_with_ixns<float>(int, const double *, const double *, const double *, const doub…
     8.1          227,422         12      18,951.8        16,864        22,528      2,351.4  void k_find_block_bounds<double>(int, int, int, const double *, const double *, double *, double *) 
     4.3          121,057         12      10,088.1         7,104        13,600      3,051.8  void k_find_block_bounds<float>(int, int, int, const double *, const double *, double *, double *)  
     2.4           67,230         24       2,801.3         2,592         3,040        158.5  k_compact_trim_atoms(int, int, unsigned int *, unsigned int *, int *, unsigned int *)   

dhfr-apo: N=23558 speed: 216.82ns/day dt: 1.5fs (ran 100000 steps in 59.78s)
dhfr-apo-barostat-interval-25: N=23558 speed: 193.24ns/day dt: 1.5fs (ran 100000 steps in 67.07s)
dhfr-apo-hmr-barostat-interval-25: N=23558 speed: 319.22ns/day dt: 2.5fs (ran 100000 steps in 67.67s)
building a protein system with 1758 protein atoms and 7047 water atoms
WARNING:absl:No GPU/TPU found, falling back to CPU. (Set TF_CPP_MIN_LOG_LEVEL=0 and rerun for more info.)
hif2a-apo: N=8805 speed: 478.39ns/day dt: 1.5fs (ran 100000 steps in 27.10s)
hif2a-apo-barostat-interval-25: N=8805 speed: 418.24ns/day dt: 1.5fs (ran 100000 steps in 30.99s)
[...]

Issue #577, opened 2022-01-24
Refactor device memory management using DeviceBuffer


mcwitt:
E.g., the following implementations should probably all use `DeviceBuffer` (introduced in #576)  to manage device memory (e.g. to ensure that `cudaMalloc` is always paired with `cudaFree`, even in the presence of exceptions):

- [Potential::execute_host_du_dx](https://github.com/proteneer/timemachine/blob/6593d22a23709b52117b3ff99f78d032c216ccb9/timemachine/cpp/src/potential.cu#L87)
- [BoundPotential::execute_host](https://github.com/proteneer/timemachine/blob/10f8d2b6ddd6d89807436f24a329774e9c99bf98/timemachine/cpp/src/bound_potential.cu#L20)
- [Context::multiple_steps](https://github.com/proteneer/timemachine/blob/10f8d2b6ddd6d89807436f24a329774e9c99bf98/timemachine/cpp/src/context.cu#L56)
- [Context::multiple_steps_U](https://github.com/proteneer/timemachine/blob/10f8d2b6ddd6d89807436f24a329774e9c99bf98/timemachine/cpp/src/context.cu#L131)
- probably others too

PR #578, opened 2022-01-24
Add NonbondedInteractionGroup potential, rename existing nonbonded potentials

mcwitt:
- [x] Add `NonbondedInteractionGroup`, allowing computation of nonbonded interactions between two sets of atoms (useful for efficient computation of host-ligand interactions; this uses the neighborlist updates from #545)
- [x] Rename existing nonbonded potentials for clarity:
  - `NonbondedDense` -> `NonbondedAllPairs` (given a set of atoms, computes the sum of all pairwise interactions)
  - `NonbondedPairs` -> `NonbondedPairList` (given a list of atom pairs, computes the sum of the specified pair interactions)

The implementation strategy for the interaction groups potential is to compose an extra permutation with the existing hilbert-sorting permutation, resulting in a permutation that
1. sorts atoms into contiguous blocks (column atoms followed by row atoms), and
2. (if enabled) hilbert-sorts each block independently

The idea for the initial correctness test is due to @maxentile . This uses the identity
```
U = U_A + U_B + U_{AB}
```
where `U_A` (`U_B`) is the all-pairs potential for the first (second) group, and `U_{AB}` is the "interaction group" potential.

#### Notes
- This involves copying the existing nonbonded potential to a new file and making modifications, and as such is probably easiest to review commit-by-commit. I've squashed the original branch down to a smaller set of changes that should follow logically. The main implementation of the interaction group potential is in 945e62a, and the correctness test in 980185e
- Currently it's only supported to specify one set of atom indices (e.g. for the ligand), and the remaining indices are assumed to be the other interacting group. This was to simplify the initial implementation, but we might need to generalize this to make it useful


timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -280,35 +292,43 @@ void NonbondedInteractionGroup<RealType, Interpolated>::execute_device(
...
     // (ytz) see if we need to rebuild the neighborlist.
     // (ytz + jfass): note that this logic needs to change if we use NPT later on since a resize in the box
     // can introduce new interactions.
-    k_check_rebuild_coords_and_box<RealType>
-        <<<B, tpb, 0, stream>>>(N, d_x, d_nblist_x_, d_box, d_nblist_box_, nblist_padding_, d_rebuild_nblist_);
+    k_check_rebuild_coords_and_box_gather<RealType><<<B1, tpb, 0, stream>>>(
+        N1_, d_atom_idxs_1_, d_x, d_nblist_x_, d_box, d_nblist_box_, nblist_padding_, d_rebuild_nblist_);
+
+    k_check_rebuild_coords_and_box_gather<RealType><<<B2, tpb, 0, stream>>>(
+        N2_, d_atom_idxs_2_, d_x, d_nblist_x_, d_box, d_nblist_box_, nblist_padding_, d_rebuild_nblist_);
```
mcwitt, line 305: todo: should be able to do this with a single kernel call by concatenating `d_atom_idxs_*` first

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -280,35 +292,43 @@ void NonbondedInteractionGroup<RealType, Interpolated>::execute_device(
...
+    dim3 dimGrid(BI, 3, 1);
+
     if (p_rebuild_nblist_[0] > 0) {
 
         // (ytz): update the permutation index before building neighborlist, as the neighborlist is tied
         // to a particular sort order
         if (!disable_hilbert_) {
-            this->hilbert_sort(d_x, d_box, stream);
+            this->hilbert_sort(N1_, d_atom_idxs_1_, d_x, d_box, d_perm_, stream);
+            this->hilbert_sort(N2_, d_atom_idxs_2_, d_x, d_box, d_perm_ + N1_, stream);
```
mcwitt, line 323: the idea is to create a permutation that
1. partitions atoms into 2 contiguous groups
2. hilbert sorts within each group

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -280,35 +292,43 @@ void NonbondedInteractionGroup<RealType, Interpolated>::execute_device(
...
         } else {
-            k_arange<<<B, tpb, 0, stream>>>(N, d_perm_);
+            // just partition the atoms into groups without hilbert sorting
+            k_concatenate<<<BI, tpb, 0, stream>>>(NI_, N1_, d_atom_idxs_1_, d_atom_idxs_2_, d_perm_);
             gpuErrchk(cudaPeekAtLastError());
         }
 
         // compute new coordinates, new lambda_idxs, new_plane_idxs
-        k_permute<<<dimGrid, tpb, 0, stream>>>(N, d_perm_, d_x, d_sorted_x_);
+        k_permute<<<dimGrid, tpb, 0, stream>>>(NI_, d_perm_, d_x, d_sorted_x_);
```
mcwitt, line 331: todo: think of better name for `NI_` (number of interacting atoms, sum of the group sizes). Looks too much like `N1_`

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -529,14 +588,15 @@ void __global__ k_nonbonded_unified(
...
+    const unsigned int row_atom_offset, // index of the first atom in the first row block
     unsigned long long *__restrict__ du_dx,
     unsigned long long *__restrict__ du_dp,
     unsigned long long *__restrict__ du_dl_buffer,
     unsigned long long *__restrict__ u_buffer) {
 
     int tile_idx = blockIdx.x;
     int row_block_idx = ixn_tiles[tile_idx];
-    int atom_i_idx = row_block_idx * 32 + threadIdx.x;
+    int atom_i_idx = row_block_idx * 32 + threadIdx.x + row_atom_offset;
```
mcwitt, line 599: still a bit unsure about this. The idea is to make `atom_i_idx` index into the full permuted coordinate array of size `N1_ + N2_`

tests/test_nonbonded_interaction_group.py
```diff
@@ -0,0 +1,129 @@
...
+    _, _, box = test_system
+    return np.asarray(box / box.unit)
+
+
+@pytest.mark.parametrize("beta", [2.0])
+@pytest.mark.parametrize("cutoff", [1.1])
+@pytest.mark.parametrize("precision,rtol,atol", [(np.float64, 1e-8, 1e-8), (np.float32, 1e-4, 5e-4)])
+@pytest.mark.parametrize("num_row_atoms", [1, 15])
+@pytest.mark.parametrize("num_atoms", [33, 231, 1050])
+def test_nonbonded_interaction_group_correctness(
```
maxentile, line 64: A couple other possible self-consistency tests:

* Modify coords in a way that influences U_AB but not U_A or U_B, something like:
```python
# ... 
conf_prime = np.array(conf)
conf_prime[ligand_idxs] += 0.01 * np.random.randn(len(ligand_idxs), 3)  # assuming U_A is shift-invariant

ref_delta_U = U_ref(conf_prime) - U_ref(conf)
test_delta_U = U_ixn(conf_prime) - U_ixn(conf)
# ...
# assert ref_delta_U close to test_delta_U
```

* Modify lam in a way that influences U_AB but not U_A or U_B.
(For example, construct with lambda planes / offsets all one value for atoms in A, and a different value for all atoms in B.)

tests/test_nonbonded_interaction_group.py
```diff
@@ -0,0 +1,129 @@
...
+
+    ref_allpairs_minus_ixngroups = make_reference_nonbonded(
+        np.concatenate((np.zeros(num_row_atoms, dtype=np.int32), np.ones(num_col_atoms, dtype=np.int32)))
+    )
+
+    def ref_ixngroups(*args):
+        return ref_allpairs(*args) - ref_allpairs_minus_ixngroups(*args)
+
+    test_ixngroups = NonbondedInteractionGroup(
+        np.arange(0, num_row_atoms, dtype=np.int32),
```
maxentile, line 95: Possibility to strengthen test: Non-contiguous `row_atom_idxs_i`?

tests/test_nonbonded_interaction_group.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+@pytest.fixture(autouse=True)
+def set_random_seed():
+    np.random.seed(2021)
+    yield
+
+
+@pytest.fixture
+def test_system():
```
maxentile, line 23: Pytest aside: I didn't realize it was okay for fixture names to start with `test_`!

I don't see a prohibition in pytest docs, and confirmed `test_system`, `test_conf`, `test_box` aren't spuriously collected by default...

But it might still be safer if these are renamed to not start with `test_`?

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -0,0 +1,482 @@
...
+    gpuErrchk(cudaFree(d_sorted_p_));
+    gpuErrchk(cudaFree(d_unsorted_dp_dl_));
+    gpuErrchk(cudaFree(d_sorted_dp_dl_));
+    gpuErrchk(cudaFree(d_sorted_du_dx_));
+    gpuErrchk(cudaFree(d_sorted_du_dp_));
+
+    gpuErrchk(cudaFree(d_sort_keys_in_));
+    gpuErrchk(cudaFree(d_sort_keys_out_));
+    gpuErrchk(cudaFree(d_sort_vals_in_));
+    gpuErrchk(cudaFree(d_sort_storage_));
```
maxentile, line 202: Possible refinement: Could some of these repetitions be avoided using the `DeviceBuffer` added in https://github.com/proteneer/timemachine/pull/576?

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -0,0 +1,482 @@
...
+    }
+
+    // identify which tiles contain interpolated parameters
+
+    const int tpb = 32;
+    const int B = ceil_divide(N_, tpb);
+
+    // (ytz) see if we need to rebuild the neighborlist.
+    // (ytz + jfass): note that this logic needs to change if we use NPT later on since a resize in the box
+    // can introduce new interactions.
```
maxentile, line 304: Outdated comment (present also in previous)

maxentile (commented):
Looks good -- minor comments so far

tests/test_nonbonded_interaction_group.py
mcwitt: Agreed! Fixed in ecf76d8

timemachine/cpp/src/nonbonded_interaction_group.cu
mcwitt: ah, should this comment be removed both here and in `nonbonded_all_pairs.cu`? It wasn't clear to me that it's outdated

timemachine/cpp/src/nonbonded_interaction_group.cu
mcwitt: Agree, but was thinking to refactor this in a separate PR (I initially started to refactor this, but it involved more substantial changes than I had anticipated and this PR was getting big as-is). Tagging #577 to track.

tests/test_nonbonded_interaction_group.py
mcwitt: Good call, implemented in fcd4199

timemachine/cpp/src/nonbonded_interaction_group.cu
maxentile: I think the comment can be removed both places!

Was addressed last year by triggering nblist rebuild whenever box changes at all https://github.com/proteneer/timemachine/blob/5d8f3099aafa1cf7def0585525aaf4de73c23343/timemachine/cpp/src/kernels/k_nonbonded_dense.cuh#L48-L82

mcwitt: makes sense! removed the obsolete comment in both places in 4f4bc51

tests/test_nonbonded_interaction_group.py
badisa: I find modification of lambda values to test this quite confusing. It seems like we would want a generic jax reference implementation of this feature to test it?

timemachine/cpp/src/nonbonded_interaction_group.hpp
```diff
@@ -0,0 +1,125 @@
...
+#include "neighborlist.hpp"
+#include "potential.hpp"
+#include "vendored/jitify.hpp"
+#include <array>
+#include <set>
+#include <vector>
+
+namespace timemachine {
+
+typedef void (*k_nonbonded_fn)(
```
badisa, line 12: Do we need multiple copies of this typdef?

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -316,6 +298,12 @@ void __device__ v_nonbonded_unified(
     // int lambda_offset_i = atom_i_idx < N ? lambda_offset_idxs[atom_i_idx] : 0;
     // int lambda_plane_i = atom_i_idx < N ? lambda_plane_idxs[atom_i_idx] : 0;
 
+    int N = NC + NR;
```
badisa, line 301: nit: const

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -0,0 +1,480 @@
...
+    std::iota(all_atom_idxs.begin(), all_atom_idxs.end(), 0);
+    std::set<int> col_atom_idxs;
+    std::set_difference(
+        all_atom_idxs.begin(),
+        all_atom_idxs.end(),
+        row_atom_idxs.begin(),
+        row_atom_idxs.end(),
+        std::inserter(col_atom_idxs, col_atom_idxs.end()));
+
+    if (lambda_offset_idxs.size() != N_) {
```
badisa, line 84: nit: Move lambda offset and plane checks before construction of column atoms

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -0,0 +1,480 @@
...
+    // b. else:
+    //     - permute new coords
+    // c. permute parameters
+    // d. compute the nonbonded interactions using the neighborlist
+    // e. inverse permute the forces, du/dps into the original index.
+    // f. u and du/dl is buffered into a per-particle array, and then reduced.
+    // g. note that du/dl is not an exact per-particle du/dl - it is only used for reduction purposes.
+
+    if (N != N_) {
+        std::cout << N << " " << N_ << std::endl;
```
badisa, line 286: Remove? Also another print statement on 293

tests/test_nonbonded_interaction_group.py
maxentile: >I find modification of lambda values to test this quite confusing. 

In absolute free energy "decoupling" stages, do we not already define the lambda planes / offsets in such a way that `lam` only controls ligand-environment interactions (but does not influence ligand-ligand or environment-environment interactions, `U(x, lam) = U_A(x) + U_B(x) + U_AB(x, lam)`)?

If so, this would be another possible route to a self-consistency test for a new way to compute `U_AB`.

>It seems like we would want a generic jax reference implementation of this feature to test it?

Also a good idea to compare to jax reference implementations.

A helpful starting point may be this timing script using `timemachine.potentials.jax_utils` https://github.com/proteneer/timemachine/pull/545#issuecomment-991190069

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -0,0 +1,480 @@
...
+    gpuErrchk(cudaPeekAtLastError());
+
+    // look up which kernel we need for this computation
+    int kernel_idx = 0;
+    kernel_idx |= d_du_dp ? 1 << 0 : 0;
+    kernel_idx |= d_du_dl ? 1 << 1 : 0;
+    kernel_idx |= d_du_dx ? 1 << 2 : 0;
+    kernel_idx |= d_u ? 1 << 3 : 0;
+
+    if (p_ixn_count_[0] > 0) {
```
badisa, line 407: If this is the case, can we not just return from the function immediately as nothing will change?

timemachine/cpp/src/nonbonded_interaction_group.hpp
```diff
@@ -0,0 +1,125 @@
...
+    double *d_nblist_box_;  // box which was used to rebuild the nblist
+    int *d_rebuild_nblist_; // whether or not we have to rebuild the nblist
+    int *p_rebuild_nblist_; // pinned
+
+    unsigned int *d_perm_; // hilbert curve permutation
+
+    double *d_w_;     //
+    double *d_dw_dl_; //
+
+    double *d_sorted_x_;     //
```
badisa, line 61: Are these empty comments suppose to be filled out?

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -0,0 +1,480 @@
...
+
+template <typename RealType, bool Interpolated>
+NonbondedInteractionGroup<RealType, Interpolated>::~NonbondedInteractionGroup() {
+    gpuErrchk(cudaFree(d_col_atom_idxs_));
+    gpuErrchk(cudaFree(d_row_atom_idxs_));
+
+    gpuErrchk(cudaFree(d_lambda_plane_idxs_));
+    gpuErrchk(cudaFree(d_lambda_offset_idxs_));
+    gpuErrchk(cudaFree(d_du_dp_buffer_));
+    gpuErrchk(cudaFree(d_perm_)); // nullptr if we never built nblist
```
badisa, line 183: I believe this comment is no longer true. I also think if it were true this would segfault?

timemachine/cpp/src/nonbonded_interaction_group.cu
mcwitt: > Also how often is this check false?

Probably not very often -- I hit it initially when there was a bug in the neighborlist invocation resulting in some test cases having zero interactions. I added this guard to protect against a runtime error in this case (the kernel call fails if passed 0 for number of blocks), but I think you're right that it makes sense to just return early if we hit this case.

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -0,0 +1,480 @@
...
+        if (result != 0) {
+            throw std::runtime_error("Driver call to k_permute_interpolated failed");
+        }
+    } else {
+        k_permute<<<dimGrid, tpb, 0, stream>>>(N, d_perm_, d_p, d_sorted_p_);
+        gpuErrchk(cudaPeekAtLastError());
+        gpuErrchk(cudaMemsetAsync(d_sorted_dp_dl_, 0, N * 3 * sizeof(*d_sorted_dp_dl_), stream))
+    }
+
+    // this stream needs to be synchronized so we can be sure that p_ixn_count_ is properly set.
```
badisa, line 378: This comment seems to be no longer needed as there isn't a synchronization right here

timemachine/cpp/src/nonbonded_interaction_group.hpp
mcwitt: These are carried over from `nonbonded_all_pairs.cu`. We should probably clean these up at some point; I was thinking to do some of this in a separate PR that also DRYs some of the common bits.

mcwitt: Added to #639

timemachine/cpp/src/nonbonded_interaction_group.hpp
mcwitt: Added to #639

tests/test_nonbonded_interaction_group.py
```diff
@@ -0,0 +1,144 @@
...
+        lambda_offset_idxs,
+        beta,
+        cutoff,
+    )
+
+    GradientTest().compare_forces(
+        coords,
+        params,
+        example_box,
+        lamb=0.1,
```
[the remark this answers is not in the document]
mcwitt: hmm, since we're testing with a non-interpolated potential and `lambda_offset_idxs` are all zero, would the varying the lambda value make a difference here? It does seem like we should add a test covering the interpolated case, similar to what's in `test_parameter_interpolation.py`?

timemachine/cpp/src/kernels/k_nonbonded.cuh
mcwitt: thanks, fixed in 6185c4b

timemachine/cpp/src/nonbonded_interaction_group.cu
mcwitt: good call, moved in efa4828

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -406,7 +394,8 @@ void __device__ v_nonbonded_unified(
         // (ytz): note that d2ij must be *strictly* less than cutoff_squared. This is because we set the
         // non-interacting atoms to exactly real_cutoff*real_cutoff. This ensures that atoms who's 4th dimension
         // is set to cutoff are non-interacting.
-        if (d2ij < cutoff_squared && atom_j_idx > atom_i_idx && atom_j_idx < N && atom_i_idx < N) {
+        if (d2ij < cutoff_squared && atom_i_idx < N &&
+            (NR == 0 && atom_j_idx > atom_i_idx && atom_j_idx < N || NR != 0 && atom_j_idx < NC)) {
```
proteneer, line 398: ugh

timemachine/cpp/src/nonbonded_interaction_group.cu
mcwitt: Added a test for the case of no interactions in d0f154a, and early return in 0cd9168

timemachine/cpp/src/kernels/k_nonbonded.cuh
mcwitt: agreed this is bad -- attempted to clarify a bit in 05b1b35

timemachine/cpp/src/nonbonded_interaction_group.cu
mcwitt: I moved the comment to where I think it should be, and also moved the call to `cudaStreamSynchronize` to right after the relevant copy in 43fd522. Can you see any downside to moving it above the box copy?

timemachine/cpp/src/nonbonded_interaction_group.cu
mcwitt: Replaced the print statements with more informative errors both here and in `nonbonded_all_pairs.cu` in 653fe4d

timemachine/cpp/src/nonbonded_interaction_group.cu
mcwitt: Good catch, `d_perm_` should always be allocated (both here and in `nonbonded_all_pairs.cu`). I removed the comments in 690a57c and 1097ca4

tests/test_nonbonded_interaction_group.py
mcwitt: I added an additional consistency check comparing with [`nonbonded_v3_on_specific_pairs`](https://github.com/proteneer/timemachine/blob/5d8f3099aafa1cf7def0585525aaf4de73c23343/timemachine/potentials/nonbonded.py#L222) in eb07db3. This is hopefully more straightforward to follow.

Interestingly, I am getting a failure for this new check for some sizes, and seeing that only a small number of du_dps are significantly off. I will follow up on this.

mcwitt: > A couple other possible self-consistency tests:

Nice, thanks for the suggestions! I'll look into adding these next.

tests/test_nonbonded_interaction_group.py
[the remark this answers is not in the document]
badisa: It shouldn't make a difference, but no little cost to testing it

mcwitt:
> nit: Can also mean the identity permutation if hilbert sorting is diabled
thanks, addressed in 3e36b51

tests/test_nonbonded_interaction_group.py
mcwitt: > Interestingly, I am getting a failure for this new check for some sizes

I think I've tracked down the cause of this -- the nonbonded kernel [returns du/dp=0 for Lennard-Jones parameters at zero](https://github.com/proteneer/timemachine/blob/39907d6cc2bfe9ba36d4fffdecdeb93889507103/timemachine/cpp/src/kernels/k_nonbonded.cuh#L420), but the reference jax implementation returns nonzero values. Worked around this in 39907d6.

tests/test_nonbonded_interaction_group.py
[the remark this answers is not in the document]
badisa: We already do here: https://github.com/proteneer/timemachine/blob/master/tests/test_nonbonded.py#L471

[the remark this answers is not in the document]
mcwitt: ah, I see, I had missed that case (we don't do it in `test_correctness`, which is the test I had originally adapted from)

tests/test_nonbonded_interaction_group.py
maxentile: >I think I've tracked down the cause of this -- the nonbonded kernel [returns du/dp=0 for Lennard-Jones parameters at zero](https://github.com/proteneer/timemachine/blob/39907d6cc2bfe9ba36d4fffdecdeb93889507103/timemachine/cpp/src/kernels/k_nonbonded.cuh#L420), but the reference jax implementation returns nonzero values. Worked around this in https://github.com/proteneer/timemachine/commit/39907d6cc2bfe9ba36d4fffdecdeb93889507103.

Ahh -- I had overlooked this...

Tangents:
1. The reference jax implementation should probably be modified to stop gradient when `eps_ij == 0` https://github.com/proteneer/timemachine/blob/39907d6cc2bfe9ba36d4fffdecdeb93889507103/timemachine/potentials/nonbonded.py#L177  --> maybe by adding `eps_ij != 0` condition to `keep_mask` https://github.com/proteneer/timemachine/blob/39907d6cc2bfe9ba36d4fffdecdeb93889507103/timemachine/potentials/nonbonded.py#L208 
2. The special case of `eps_i == eps_j == 0` should be exercised in the default nonbonded correctness test, if it's not already

maxentile: Ahh, actually this line can just be deleted https://github.com/proteneer/timemachine/blob/39907d6cc2bfe9ba36d4fffdecdeb93889507103/timemachine/potentials/nonbonded.py#L208 (it overwrites a `keep_mask` that already reflects the `eps_ij != 0` condition)

tests/test_nonbonded_interaction_group.py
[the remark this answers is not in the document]
mcwitt: Added multiple lambda values to both the correctness and consistency tests in 
dd543ea.

timemachine/cpp/src/nonbonded_all_pairs.cu
```diff
@@ -295,12 +287,14 @@ void NonbondedDense<RealType, Interpolated>::execute_device(
...
 
+        // this stream needs to be synchronized so we can be sure that p_ixn_count_ is properly set.
+        cudaStreamSynchronize(stream);
+
         std::vector<double> h_box(9);
         gpuErrchk(cudaMemcpyAsync(&h_box[0], d_box, 3 * 3 * sizeof(*d_box), cudaMemcpyDeviceToHost, stream));
         // Verify that the cutoff and box size are valid together. If cutoff is greater than half the box
         // then a particle can interact with multiple periodic copies.
         const double db_cutoff = (cutoff_ + nblist_padding_) * 2;
-        cudaStreamSynchronize(stream);
```
badisa, line 303: This sync has to be here, else the box may not be copied from the device

mcwitt: I moved `cudaStreamSynchronize` back to the original location with respect to the memcpy call in 25fd4f6 (although [docs](https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html) seem to indicate that this should only matter for copies to non-pageable host memory)

tests/test_nonbonded_interaction_group.py
mcwitt: > Ahh, actually this line can just be deleted

This should be addressed by 860dc9d (we don't use `nonbonded_v3` for the test in question, but a similar change was needed in `nonbonded_v3_specific_pairs`)

maxentile: Ahh, sorry for these bunch of red herrings -- the value of `keep_mask` assigned at L182 applies to vdw, the value of `keep_mask` assigned at L208 applies to coulomb... (And as @mcwitt pointed out offline: this whole function isn't involved in the relevant test!)

maxentile: (Marking resolved)

maxentile (commented):
These changes are looking pretty good!

(Minor comments in new testing code)

timemachine/potentials/nonbonded.py
```diff
@@ -251,6 +253,17 @@ def apply_cutoff(x):
     return vdW, electrostatics
 
 
+def nonbonded_v3_interaction_groups(conf, params, box, inds_l, inds_r, beta: float, cutoff: Optional[float] = None):
```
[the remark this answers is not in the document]
maxentile: The more I think about this, the more I dislike the cryptic "`zip(inds_l, inds_r)`" idiom I introduced here, esp. when it gets in the way of more intuitive "`product(group_a_inds, group_b_inds)`" usage you're going for here.

There's no reason I can think of to prefer two (n_pairs,) arrays over a single (n_pairs, 2) array for pair lists, and I think I should remove the `(inds_l, inds_r)` / `(inds_i, inds_j)` pairs idiom throughout `jax_utils` in a follow-up PR...

timemachine/potentials/nonbonded.py
```diff
@@ -251,6 +253,17 @@ def apply_cutoff(x):
...
+    """Nonbonded interactions between all pairs of atoms $(i, j)$
+    where $i$ is in the first set and $j$ in the second.
+
+    See nonbonded_v3 docstring for more details
+    """
+    pairs = np.array(list(itertools.product(inds_l, inds_r)))
+    vdW, electrostatics = nonbonded_v3_on_specific_pairs(conf, params, box, pairs[:, 0], pairs[:, 1], beta, cutoff)
```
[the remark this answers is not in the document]
mcwitt: Wow, good catch! Switched to using `np.meshgrid` in 91d1722 (seems slightly more helpful than `np.indices` in this case, since we're passing in arrays of indices) and also saw a ~5000x speedup.

proteneer (changes requested):
This is looking fantastic! I added some comments in-line re: expanding coverage for varying lambda.

Given the critical nature of this code, I'm requesting an additional consistency tests as a sanity check that we didn't "break" anything. Can we check bitwise consistency of the original C++ `Nonbonded` code with `NonbondedAllPairs U_A` + `NonbondedAllPairs U_B` + `NonbondedInteractionGroup U_AB`, and let's vary lambda for this, and test it with and without hilbert curve sorting.

mcwitt:
> Can we check bitwise consistency of the original C++ Nonbonded code with NonbondedAllPairs U_A + NonbondedAllPairs U_B + NonbondedInteractionGroup U_AB

@proteneer  If I understand correctly I think this would require

1. exposing `NonbondedAllPairs` in Python, and
2. adding an option to `NonbondedAllPairs` to use a subset of atoms.

(1) may not be strictly necessary since we can just pass an empty list of exclusions to the full `Nonbonded` potential)

(2) is needed because `SummedPotential` passes the same coordinates and parameters to each potential in the sum (and I think we'd probably need to use this to check bitwise equality, versus doing the sum after the conversion to floating-point in Python).

Would it make sense to add this check once `NonbondedAllPairs` is exposed? Can make an issue to track.

PR #579, opened 2022-01-26
Disambiguate magic numbers in neighborlist

mcwitt:
There are 3 constants defined in the neighborlist code that typically take the value 32, but have different interpretations:
- `WARPSIZE`: the number of threads in a warp, fixed to be 32 in current device architectures
- `TILESIZE`: the number of atoms in a neighborlist row (column) block
- `tpb`: the number of threads per thread block. Usually a multiple of warp size; we currently set it equal to warp size for simplicity

This PR aims to replace all uses of the literal `32` referring to one of the above concepts with the appropriate named constant.

### DIFF
--- a/timemachine/cpp/src/kernels/k_neighborlist.cuh
+++ b/timemachine/cpp/src/kernels/k_neighborlist.cuh
@@ -178,8 +178,8 @@ void __global__ k_find_blocks_with_ixns(
     const int indexInWarp = threadIdx.x % warp_size;
     const int warpMask = (1 << indexInWarp) - 1;
 
-    __shared__ int ixn_j_buffer
-        [64]; // we can probably get away with using only 32 if we do some fancier remainder tricks, but this isn't a huge save
+    // we can probably get away with using only 32 if we do some fancier remainder tricks, but this isn't a huge save
+    __shared__ int ixn_j_buffer[2 * warp_size];
 
     // initialize
     ixn_j_buffer[threadIdx.x] = NC;
--- a/timemachine/cpp/src/kernels/k_neighborlist.cuh
+++ b/timemachine/cpp/src/kernels/k_neighborlist.cuh
@@ -146,8 +146,8 @@ This is launched with a threadblock size of 32, (i.e. one warp).
 
 Each block proceeds as follows:
 
-1. Loads its own row block (32 atoms).
-2. Compare the row block against 32 other column blocks via bounding box tests.
+1. Loads its own row block (tile_size = 32 atoms).
+2. Compare the row block against tile_size = 32 other column blocks via bounding box tests.
 3. Determine which blocks potentially interact using warp-level programming.
 4. Loop over each interacting block j, and see which row block atoms may interact with j's bbox.
 5. For atoms that interact, do a fine-grained comparison of each row block again against each col block atom.
--- a/timemachine/cpp/src/kernels/k_neighborlist.cuh
+++ b/timemachine/cpp/src/kernels/k_neighborlist.cuh
@@ -173,7 +173,7 @@ void __global__ k_find_blocks_with_ixns(
     unsigned int *__restrict__ trim_atoms,       // the left-over trims that will later be compacted
     const double cutoff) {
 
-    static_assert(tile_size == warp_size, "must have tile_size == warp_size");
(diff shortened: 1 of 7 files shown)


timemachine/cpp/src/neighborlist.hpp
```diff
@@ -68,16 +68,17 @@ template <typename RealType> class Neighborlist {
     unsigned int *get_ixn_count() { return d_ixn_count_; }
 
     // get max number of row blocks
-    int B() const { return ceil_divide(this->compute_full_matrix() ? NR_ : NC_, 32); }
+    int B() const;
```
mcwitt, line 71: I moved the header implementations into the `cu` file so they could access `warp_size` and `tile_size` without needing to add more includes here

timemachine/cpp/src/kernels/kernel_utils.cuh
```diff
@@ -32,6 +31,8 @@
 
 #define ONE_4PI_EPS0 138.935456
 
+static const int warp_size = 32;
```
mcwitt, line 34: I learned that CUDA also defines a run-time constant `warpSize`, which is automatically correct for the device architecture (but this is always 32 on existing architectures). There's some conflicting advice [here](https://stackoverflow.com/questions/36047035/when-should-i-use-cudas-built-in-warpsize-as-opposed-to-my-own-proper-constant), but it seems like our approach of defining this as a compile-time constant might be better for our use case, since it allows for compiler optimization.

badisa (approved):

timemachine/cpp/src/neighborlist.hpp
mcwitt: Actually, would it make sense for `B` and `compute_full_matrix`, `column_blocks`, etc. to be member constants rather than member functions? (since they are deterministic functions of other constants)

timemachine/cpp/src/kernels/k_neighborlist.cuh
```diff
@@ -167,19 +169,19 @@ void __global__ k_find_blocks_with_ixns(
     const double *__restrict__ box,
     unsigned int *__restrict__ interactionCount, // number of tiles that have interactions
     int *__restrict__ interactingTiles,          // the row block idx of the tile that is interacting
-    unsigned int *__restrict__ interactingAtoms, // the col block of the atoms that are interacting
+    unsigned int *__restrict__ interactingAtoms, // [NR * warp_size] atom indices interacting with each row block
```
mcwitt, line 172: Is `warp_size` correct here? Initially I thought this should be `tile_size`, but the code [here](https://github.com/proteneer/timemachine/blob/a6a9339d68578ca6fa20e46f421b794a5d709037/timemachine/cpp/src/kernels/k_neighborlist.cuh#L399) indicates the former is correct. Is this a case of accidental substitution?

badisa: It is a little ambiguous right now. I think tile_size is actually the correct variable. Which would mean it should use the tile_size in the code you referenced.

maxentile (approved):
Looks good! Generally in favor of naming constants like this.

Out of curiosity, which of these can be safely changed to values other than 32?

>This PR aims to replace all uses of the literal 32 referring to one of the above concepts with the appropriate named constant.

Might also expand on this condition while fresh in mind: https://github.com/proteneer/timemachine/blob/1ac56298ca05f64b612acae0379bae6797ac7a81/timemachine/cpp/src/wrap_kernels.cpp#L53-L56

mcwitt:
> Out of curiosity, which of these can be safely changed to values other than 32?

Good question. Currently, changing `tile_size` to 16 results in test failures, so it seems like this may not really be independent of `warp_size`.

mcwitt:
> Might also expand on this condition while fresh in mind:

Added a comment in 27d8827.

timemachine/cpp/src/kernels/k_neighborlist.cuh
mcwitt: I didn't change the referenced code for now as there are similar cases nearby where I'm unsure of the correct constant to use. I'll come back to this when I have a better understanding of the implementation in `k_neighborlist.cuh`.

maxentile (approved):

PR #580, opened 2022-01-26
Update absolute hydration example

maxentile:
Resolves broken references, largely rewrites script to more closely resemble protocols used in `fe` module.

### DIFF
--- a/examples/ahfe.py
+++ b/examples/ahfe.py
@@ -93,9 +93,7 @@ dt = 1.5e-3
 
 seed = 2022
 
-
-def construct_npt_move(lam):
-    return NPTMove(final_potentials, lam, masses, temperature, pressure, n_steps_per_move, seed)
+npt = NPTMove(final_potentials, 1.0, masses, temperature, pressure, n_steps_per_move, seed)
 
 
 # collect samples from p(x | lam) for lam in lambda_schedule
@@ -105,11 +103,10 @@ trajs = []
 
 for lam in lambda_schedule:
     print("lam = ", lam)
+    npt.lamb = lam
 
     x_equil = CoordsVelBox(coords, np.zeros_like(coords), box)
 
-    npt = construct_npt_move(lam)
-    assert npt.lamb == lam
     trange = tqdm(range(n_equil_moves), desc="equilibration")
     for _ in trange:
         x_equil = npt.move(x_equil)
--- a/examples/ahfe.py
+++ b/examples/ahfe.py
@@ -1,13 +1,14 @@
 # absolute hydration free energy
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


examples/ahfe.py
```diff
@@ -4,142 +4,112 @@
...
 
-# 1. build water box
-# 2. build ligand
-# 3. convert to recipe
+from pathlib import Path
 
-# construct an RDKit molecule of aspirin
-# note: not using OpenFF Molecule because want to avoid the dependency (YTZ?)
-romol = Chem.AddHs(Chem.MolFromSmiles("CC(=O)OC1=CC=CC=C1C(=O)O"))
+from tqdm import tqdm
```
badisa: We do not currently have tqdm as a requirement in the repo, should we add it?

badisa:
Naive question, how does this script differ from using `fe/model_rabfe.py::AbsoluteHydrationModel`?

examples/ahfe.py
maxentile: Ahh, I misremembered! (I remembered [pinning](https://github.com/proteneer/timemachine/pull/442#discussion_r645881764) the version in an earlier PR but it was unmerged.)

Can add tqdm as a dependency, remove the progress bar (just use prints), or use a different progress bar

maxentile:
>Naive question, how does this script differ from using fe/model_rabfe.py::AbsoluteHydrationModel?

Good question -- I need to look more carefully if parts of this should be replaced with fe/model_rabfe.py::AbsoluteHydrationModel or fe/model_rabfe.py::AbsoluteStandardHydrationModel .

Initially, I was trying to make the minimal set of changes to get `examples/ahfe.py` to run.

badisa:
> > Naive question, how does this script differ from using fe/model_rabfe.py::AbsoluteHydrationModel?
> 
> Good question -- I need to look more carefully if parts of this should be replaced with fe/model_rabfe.py::AbsoluteHydrationModel or fe/model_rabfe.py::AbsoluteStandardHydrationModel .
> 
> Initially, I was trying to make the minimal set of changes to get `examples/ahfe.py` to run.

The changes as they are look fine, and I only wondered about the differing styles. Not sure this example is worth too much more attention, and we can decide at a. later date which style is preferred.

examples/ahfe.py
maxentile: Removed in https://github.com/proteneer/timemachine/pull/580/commits/297fdec7509bd14e4175a211a14b6ca3be6964ec

maxentile:
Closing, probably better to re-write using `fe/model_rabfe.py::AbsoluteStandardHydrationModel`

PR #581, opened 2022-01-27
Add timeout to CI pipeline

mcwitt:
Adds 3-hour timout to the CI pipeline (meaning builds will automatically fail if they take longer than 3 hours).

The Jenkins config option `timeout` is documented [here](https://www.jenkins.io/doc/book/pipeline/syntax/#:~:text=set%20a%20timeout%20period%20for%20the%20pipeline%20run%2C%20after%20which%20jenkins%20should%20abort%20the%20pipeline).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,7 +6,7 @@ pipeline {
   stages {
     stage('Execute CI') {
       options {
-        timeout(time: 3, unit: 'HOURS')
+        timeout(time: 1, unit: 'MINUTES')
       }
       steps {
         sh '/usr/local/bin/fab -f /opt/jenkins/jenkins_scripts/timemachine_ci_fab.py ci:'+env.WORKSPACE+','+env.GIT_COMMIT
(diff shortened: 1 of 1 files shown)


badisa:
Looks like the 1 minute timeout worked!

mcwitt:
> Looks like the 1 minute timeout worked!

ah, so it did!

Jenkinsfile
```diff
@@ -5,6 +5,9 @@ pipeline {
   agent any
   stages {
     stage('Execute CI') {
+      options {
+        timeout(time: 3, unit: 'HOURS')
```
mcwitt, line 9: 3 hours seems like a safe choice for now, given that the nominal time seems to be ~35 minutes recently. We could potentially set this much lower, but my thinking is that we mainly want this to prevent zombie runs from clogging the pipeline, so a longer timeout should be adequate.

badisa (approved):

maxentile (approved):

Jenkinsfile
proteneer: now I want to know what are all the units...

proteneer (approved):

PR #588, opened 2022-01-31
Move `training/` to `attic/training` and `timemachine/training/`

maxentile:
Hasn't been under testing for a while (with the exception of `training/dataset.py`, which this PR moves to `timemachine/training/dataset.py`).

Did a quick read over all moved files and added a readme to summarize contents.


proteneer (approved):
Byebye thermodynamic derivatives!

attic/training/abfe.py
```diff
@@ -29,7 +29,7 @@
 from rdkit.Chem import AllChem
 from rdkit.Chem import rdFMCS
 
-from fe import dataset
+from timemachine.training import dataset
```
mcwitt, line 32: this change looks a little suspicious, just wanted to verify it's correct? (moving from `fe`, would've expected from `training`)

mcwitt (approved):

attic/training/abfe.py
maxentile: This was broken in the original script, fixed it and updated it in one step. This scripts does not currently run for earlier reasons (broken reference on L16 for example), but figured I'd fix any references I touched during the move.

PR #590, opened 2022-01-31
Move `fe/` to `timemachine/fe/`

maxentile:
Continuing the great namespace migration


examples/estimator_variance.py
```diff
@@ -27,8 +27,8 @@
 import numpy as np
 from rdkit import Chem
 
-from fe import free_energy, topology
-from fe.free_energy import construct_lambda_schedule
+from timemachine.fe import topology, free_energy
+from timemachine.fe.free_energy import construct_lambda_schedule
```
mcwitt, line 31: Maybe something to save for a future cleanup, but `fe.free_energy` feels reduntant to me. Would it make sense to move some of these defs into the `fe` namespace (i.e. move them to `__init__.py`)?

mcwitt (approved):

examples/estimator_variance.py
maxentile: Definitely -- in a more involved refactoring step, I'd very much like to avoid import phrases like these.

The aim of these PRs is just to move things into `timemachine` namespace and fix any references that would be immediately broken in the process.

>Would it make sense to move some of these defs into the fe namespace (i.e. move them to __init__.py)?

In general this requires some additional thought because of dual definitions between `fe.free_energy` and `fe.free_energy_rabfe` https://github.com/proteneer/timemachine/issues/565 . (Although `construct_lambda_schedule` should be fine since the `_rabfe` version does not contain this name, I'd rather wait to do a deeper reorganization in this module.)

PR #591, opened 2022-01-31
Move `ff/` to `timemachine/ff/`

maxentile:
Continuing the great namespace migration


mcwitt (approved):

.pre-commit-config.yaml
```diff
@@ -19,4 +19,4 @@ repos:
         types_or: [c, c++, cuda]
         files: '^timemachine/cpp/src/'
 
-exclude: '\.pdb$|\.sdf$|\.proto$|\.xml$|/vendored/|^ff/params/|^parallel/grpc/'
+exclude: '\.pdb$|\.sdf$|\.proto$|\.xml$|/vendored/|^timemachine.ff/params/|^parallel/grpc/'
```
mcwitt, line 22: nit: should this be
```suggestion
exclude: '\.pdb$|\.sdf$|\.proto$|\.xml$|/vendored/|^timemachine/ff/params/|^parallel/grpc/'
```
(I think the `.` is matching any character)

maxentile: Doh -- thanks for catching!

badisa:
With things moving from `ff/params/*.py` to `timemachine/ff/params/*.py` when need to change the references of
`"ff/params/smirnoff_1_1_0_ccc.py"` that we have in a lot of scripts to account for the move. Looks like there are some tests as well.

maxentile:
Doh-- good catch, will update these later this afternoon.

maxentile:
Thanks for catching these issues -- I think I've now fixed all affected path references. (Only one I couldn't update in exactly this way is in the readme -- changed to permalink to current version: https://github.com/proteneer/timemachine/pull/591/commits/42a74598606c90263eafadda0392073769d6440d )

Currently build is failing with the flaky curand error.

maxentile:
Can this get a quick re-review?

(Of the moves I was assigned for spring cleaning week, this one seems the most error-prone)

proteneer (approved):
Nice! Imports look a lot cleaner now. We may want to add a less obtuse way of referencing the default forcefields than `open("timemachine/ff/params/...py` but that can be done in a later PR.

badisa (approved):

PR #595, opened 2022-02-01
Move miscellaneous files to attic 

maxentile:
Miscellaneous things that looked out of place to me during day 1 of spring cleaning week:

~100% sure:
* Move ​`tests/dual_topology.py` to attic -- not currently under testing, not referenced, and has unresolved imports
* Move `langevin.cu`, `langevin.hpp` to attic -- all lines have been commented out since Feb 2020

~90% sure:
* Move `optimizer.cu`, `optimizer.hpp` to attic -- only referenced from `langevin.hpp`, which is commented out
* `cpp/src/generate_gradients.py` is not currently under testing and not easily discoverable -- added to a new `attic/codegen` folder. --> Should I open an issue about possibly adding a `codegen` module for functions like this? Seems useful to keep functionality like this around to document how initial versions of other source files were created / how to regenerate them with different parameters.

~80% sure:
* Move `gbsa.cu`, `gbsa.hpp`, `gbsa.py` to attic -- These implementations may be of independent interest, and may be of future interest in timemachine, but are not currently tested. These are currently only referenced from commented-out lines (or in one case, from an unused import), and they in turn currently reference files that are no longer in the repository (`#include "gbsa_jvp.cuh"`,  `#include "k_gbsa_jvp.cuh"`). Some Python GBSA code is still referenced in tests. (Note: GBSA parameter handler test is [skipped](https://github.com/proteneer/timemachine/blob/277cc7463d22af28e5d80478ed3628d0070ffd10/tests/test_handlers.py#L473-L520), but GBSA deserialization handler test is [active](https://github.com/proteneer/timemachine/blob/277cc7463d22af28e5d80478ed3628d0070ffd10/tests/test_serialization.py#L236-L277) ). --> @proteneer : (1) Do we want to move the GBSA parameter handler tests to `attic` as well? (2) Do we want to retrieve the latest versions of gbsa_jvp.cuh,  k_gbsa_jvp.cuh from an earlier point in stored history, and add them to the `attic` for completeness?


attic/readme.md
```diff
@@ -1,7 +1,11 @@
 Code, documentation, experiments we want to retain for reference, but that we're not currently maintaining.
 
 ### Contents
+* `codegen/` -- sympy script for generating CUDA code to compute bond, angle, and torsion gradients
```
mcwitt, line 4: nit: maybe
```suggestion
* `codegen/generate_gradients.py` -- sympy script for generating CUDA code to compute bond, angle, and torsion gradients
```

mcwitt (approved):
I'm not familiar enough to comment much on the deprecations, but otherwise lgtm, save for minor comment

proteneer:
@maxentile  do you mind if I take care of the C++ side of things? Will make merge conflicts etc easier to deal with. I'll incorporate some of your changes.

I'd be happy to just delete some of the C++ files entirely at this point.

maxentile:
Sounds good! Will make an issue to pick up the non-C++ files from this PR

maxentile:
(Superseded by https://github.com/proteneer/timemachine/pull/603 )

PR #596, opened 2022-02-01
Add flake8

mcwitt:
- Adds the [flake8](https://flake8.pycqa.org/en/latest/) linter to pre-commit configuration
- Configures initial set of [linting rules](https://github.com/proteneer/timemachine/pull/596/files#diff-fa602a8a75dc9dcc92261bac5f533c2a85e34fcceaff63b3a3a81d9acde2fc52)
- Fixes linting errors

#### Linting checks
The linting checks addressed here are:
- undefined names ([F821](https://www.flake8rules.com/rules/F821.html), commit f6d7fff)
- redefinitions (F811, commit 6fec4c2)
- f-string missing parameters (F541, commit aa7877b)
- invalid escape sequence (W605, commit c210a54)
- comparison to True/False (E712, commit 53ce581)
- unused local variables (F841, commit f1b2d80)

I'm fine with dropping any of these if we decide it's not worth the potential disruption or would cause undue burden to maintain.

Of these, I think f1b2d80 is potentially the most disruptive and deserves the most review attention.

#### Unused variables
While working on cleaning unused variables (in [f1b2d80](https://github.com/proteneer/timemachine/pull/596/commits/f1b2d80b58b77c701110ddae1b3fc679e1a58c4b)), I noticed cases where definitions were unused, but
- referenced in commented-out code
- accompanied by a useful comment
- nontrivial, and seemed possibly useful to keep around

In these cases I tried to comment rather than delete the unused definition.

Another case that came up a few times was that the right-hand side of unused assignment was clearly side-effecting (e.g. writing a file). In these cases I tried to just remove the LHS of the assignment and keep the RHS.


proteneer:
Does black enforce flake8? I'm on VSCode and I'm not sure if that's always supported

mcwitt:
> Does black enforce flake8?

With `extend-ignore = E203`, the output of black satisfies the formatting rules enforced by flake8.

But flake8 can do many other linting checks (unused imports, dead code, etc.), so it adds value independent of black.

timemachine/fe/frames.py
```diff
@@ -1,14 +1,16 @@
-from typing import List, Tuple, Generator
+from timemachine.fe import estimator, estimator_abfe
+from typing import List, Tuple, Generator, Union
 
 
+SimulationResult = Union[estimator.SimulationResult, estimator_abfe.SimulationResult]
```
maxentile, line 5: Nice -- `"SimulationResult"` was used instead of `SimulationResult` with [exactly this intent](https://github.com/proteneer/timemachine/pull/528#discussion_r746874882)!

(nit: only remaining line to update looks like [L6](https://github.com/proteneer/timemachine/blob/6fec4c23000a0f4170fee038cbcb2a1d385ba06c/timemachine/fe/frames.py#L6))

timemachine/ff/handlers/__init__.py
```diff
@@ -5,3 +5,12 @@
 
 from timemachine.ff.handlers.nonbonded import AM1CCCHandler
 from timemachine.ff.handlers.nonbonded import LennardJonesHandler
+
+__all__ = [
```
maxentile, line 9: 👍

.pre-commit-config.yaml
```diff
@@ -12,11 +12,25 @@ repos:
...
 
-exclude: '\.pdb$|\.sdf$|\.proto$|\.xml$|/vendored/|^timemachine/ff/params/|^parallel/grpc/'
+exclude: >
+  (?x)
+    \.pdb$
+  | \.sdf$
+  | \.proto$
+  | \.xml$
+  | /vendored/
+  | ^attic/
```
maxentile, line 34: 👍

timemachine/fe/pdb_writer.py
```diff
@@ -2,7 +2,8 @@
 
 import tempfile
 from simtk.openmm import app
-from simtk.openmm.app import PDBFile, Topology
+from simtk.openmm.app import PDBFile
+from fe.topology import SingleTopology
```
maxentile, line 6: Does this import still resolve?

timemachine/fe/frames.py
mcwitt: Ah, thanks! Fixed in [f15f731](https://github.com/proteneer/timemachine/pull/596/commits/f15f7313271b400ee416ed4889e33561d12a2038).

timemachine/fe/pdb_writer.py
mcwitt: Good catch, I think it should be either `from timemachine.fe.topology` or just `from .` (using a [relative import](https://docs.python.org/3/reference/import.html#package-relative-imports))

mcwitt: Fixed in [0537775](https://github.com/proteneer/timemachine/pull/596/commits/0537775d6dd1c0024c90a27954f602633770723d), using an absolute import to be consistent with existing convention

timemachine/fe/frames.py
mcwitt: Actually thinking through this more, it seems like it'd be more correct to use `Any` here, since the implementation doesn't assume anything about `SimulationResult`. I changed it in [f7a4e32](https://github.com/proteneer/timemachine/pull/596/commits/f7a4e328660aac1bbea7e2f952d0fe4b51b244ba), but willing to be convinced otherwise

badisa (approved):

maxentile:
>Of these, I think https://github.com/proteneer/timemachine/commit/f1b2d80b58b77c701110ddae1b3fc679e1a58c4b is potentially the most disruptive and deserves the most review attention.

This looks fine to me.

Nit: some cases looked slightly lossy to remove -- unused variables might serve as a reminder of array dimensions `N, D = x.shape` etc. -- these should maybe be comments instead? (Also, even if `N, D` are unused, this pattern does have the side effect of asserting `len(x.shape) == 2` before proceeding ...)

maxentile (approved):

tests/test_nonbonded.py
```diff
@@ -515,9 +506,6 @@ def run_nonbonded(precision, potential, x, box, params, lamb, steps=100):
             x = (x.astype(np.float32)).astype(np.float64)
             params = (params.astype(np.float32)).astype(np.float64)
 
-            N = x.shape[0]
-            D = x.shape[1]
```
maxentile, line 519: https://github.com/proteneer/timemachine/pull/596#issuecomment-1030172008

Err, `len(x.shape) >= 2`

mcwitt: Agreed -- I restored the definitions of array shape constants as comments and added shape assertions in a635a76

maxentile: Nice! This should preserve the previous shape asserting side effect without requiring exceptions to F841.

(Sorry for my confusing comment though: although the code _currently has the effect_ of asserting weaker condition `len(x.shape) >= 2`, the _intent_ is to assert stronger condition `len(x.shape) == 2`)

maxentile: (But it also makes sense to leave as `>=` to minimize any change in behavior during this PR -- no further change requested! 👍 )

mcwitt: oops, saw the above too late :) Went ahead and made the assertions more precise in c39b11e, but can revert if we'd rather not change the behavior here

maxentile: Hah, sorry for the flip-flopping -- I prefer the more precise assertions you've now applied!

PR #597, opened 2022-02-02
Add setuptools config, set up versioneer

mcwitt:
- [x] Adds project `setup.py`, adapted from https://github.com/pypa/sampleproject/blob/main/setup.py
- [x] Adds configuration for [versioneer](https://github.com/python-versioneer/python-versioneer) to handle versioning based on git tags (set up following instructions [here](https://github.com/python-versioneer/python-versioneer/blob/master/INSTALL.md))
- [x] Builds C++ lib with setuptools. Notes:
  - current method using `make build` still works
  - default `CUDA_ARCH` is now defined in CMakeLists.txt (as opposed to Makefile)
- [x] Changes the default `CUDA_ARCH` to `75` (this is what we're using in ansible builds)
- [x] Updates setup instructions

#### Versioning
Currently we get
```python
>>> import timemachine as tm
>>> tm.__version__
'0+untagged.642.gda016b2'
```
After tagging, e.g.
```shell
git tag -a 0.1.0
```
we'll get
```python
>>> tm.__version__
'0.1.0'
```
After the next (untagged) commit, it'll look like
```python
>>> tm.__version__
'0.1.0+1.g7ac99fd'
```


setup.py
```diff
@@ -0,0 +1,62 @@
...
+
+# Always prefer setuptools over distutils
+from setuptools import setup, find_packages
+import pathlib
+import versioneer
+
+here = pathlib.Path(__file__).parent.resolve()
+
+# Get the long description from the README file
+long_description = (here / "README.md").read_text(encoding="utf-8")
```
maxentile, line 14: That's neat

requirements.txt
```diff
@@ -8,9 +8,3 @@ pymbar==3.0.5
 pyyaml==5.4.1
 hilbertcurve==1.0.5
 networkx==2.5
-
-# Development/Testing Reqs
-black
-pytest
-pytest-cov
-pre-commit
```
mcwitt, line 16: Dev and testing dependencies can be installed via `pip install .[dev]` or `pip install .[test]` respectively.

setup.py
```diff
@@ -0,0 +1,62 @@
...
+        "scipy",
+    ],
+    extras_require={  # Optional
+        "dev": ["black==21.10b0", "pre-commit"],
+        "test": ["pytest", "pytest-cov"],
+    },
+    # package_data={
+    #     "sample": ["package_data.dat"],
+    # },
+    # entry_points={
```
mcwitt, line 53: I wasn't aware of any existing entry points. Are there any I missed? (i.e. scripts that we might want to expose in the path)

setup.py
```diff
@@ -0,0 +1,62 @@
...
+        "numpy",
+        "pymbar",
+        "pyyaml",
+        "scipy",
+    ],
+    extras_require={  # Optional
+        "dev": ["black==21.10b0", "pre-commit"],
+        "test": ["pytest", "pytest-cov"],
+    },
+    # package_data={
```
mcwitt, line 50: This could be a good way to include data files (non-python files, or python files that aren't part of a module)

setup.py
mcwitt: Yep! I hadn't used `pathlib` either (this is coming from the pypa template linked in the description)

mcwitt:
Force-pushed to remove some local configuration I had accidentally committed.

.pre-commit-config.yaml
```diff
@@ -10,7 +10,7 @@ repos:
     rev: 21.10b0
     hooks:
     -   id: black
-        args: [--line-length=120]
+        exclude: '^versioneer.py$|^timemachine/_version.py'
```
mcwitt, line 13: Unfortunately these exclusions need to be duplicated here from `pyproject.toml` (see https://github.com/psf/black/issues/438)

timemachine/cpp/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ if(NOT EXISTS ${EIGEN_SRC_DIR})
   execute_process(COMMAND git clone --branch 3.3.9 https://gitlab.com/libeigen/eigen.git ${EIGEN_SRC_DIR})
 endif()
 
-add_subdirectory(${CMAKE_CURRENT_BINARY_DIR}/${PYBIND_SRC_DIR})
+add_subdirectory(${CMAKE_CURRENT_BINARY_DIR}/${PYBIND_SRC_DIR} ${CMAKE_CURRENT_BINARY_DIR}/${PYBIND_SRC_DIR})
```
mcwitt, line 32: Without this I get
```
CMake Error at CMakeLists.txt:32 (add_subdirectory):
  add_subdirectory not given a binary directory but the given source
  directory
  ...
  When specifying an out-of-tree source a binary directory must be explicitly
  specified.
```

timemachine/cpp/CMakeLists.txt
```diff
@@ -65,7 +65,13 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR}/eigen)
 include_directories(${CMAKE_CURRENT_BINARY_DIR}/${CUB_SRC_DIR})
 
 set_property(TARGET ${LIBRARY_NAME} PROPERTY CUDA_STANDARD 14)
+
+if (NOT CUDA_ARCH)
+  set(CUDA_ARCH "75")
```
mcwitt, line 70: This is to allow `pip install .` to succeed without needing to first set e.g. `CMAKE_ARGS="-DCUDA_ARCH=75"`

setup.py
maxentile: I'm not sure -- two that come to mind are:
* the speed benchmarking script https://github.com/proteneer/timemachine/blob/1a721dd3f05d6011cf028b0588e066682d38ba59/slow_tests/test_benchmark.py
* RABFE script https://github.com/proteneer/timemachine/blob/1a721dd3f05d6011cf028b0588e066682d38ba59/examples/validate_relative_binding.py

setup.py
maxentile: `datasets/`, `testsystems/`, and forcefield files come to mind -- maybe others?

setup.py
badisa: Just be aware that pathlib creates objects and not strings. Lots of things will barf giving a pathlib path rather than a string path

setup.py
badisa: I don't think either of those are entry points, at least not ones we want to maintain.

pyproject.toml
```diff
@@ -0,0 +1,15 @@
...
+  | timemachine/_version.py$
+  | ff/params
+  | parallel/grpc
+  )
+'''
+
+[build-system]
+# These are the assumed default build requirements from pip:
+# https://pip.pypa.io/en/stable/reference/pip/#pep-517-and-518-support
+requires = ["setuptools>=43.0.0", "wheel", "cmake>=3.18", "versioneer-518"]
```
badisa, line 14: We should pin the cmake to >3.22 as that is what we specify in the docs

mcwitt:
Moving this to draft status as there are still a few issues to work out with the setuptools build.

Makefile
```diff
@@ -6,8 +6,6 @@ PYTEST_CI_ARGS := --cov=. --cov-report=term-missing
 
 NPROCS = `nproc`
 
-CUDA_ARCH := "70"
```
mcwitt, line 9: The default `CUDA_ARCH` setting was moved to CMakeLists.txt. This will allow `pip install -e .` to succeed without needing to set the CMake flags.

setup.py
mcwitt: Sounds like there aren't really clear CLI entry points at this point; I'll leave the comment in place in case it makes sense to add some in the future

setup.py
badisa: I don't think we want to package anything up beyond what is in `timemachine/`? Though really depends on what we want from the python package. Is it a testable python package on its own, or do we want to require the git repo along with the python package.

mcwitt: I'm leaning toward leaving these out of the package for now, since it seems like we might want to have a separate discussion about what's appropriate to include; I'll leave the commented-out lines in case we do want to add package data in the future.

setup.py
```diff
@@ -0,0 +1,110 @@
...
+    keywords="molecular dynamics",
+    ext_modules=[CMakeExtension("timemachine.lib.custom_ops", "timemachine/cpp")],
+    packages=find_packages(),
+    python_requires=">=3.7",
+    install_requires=[
+        "bootstrapped",
+        "grpcio",
+        "grpcio-tools",
+        "hilbertcurve",
+        "jax",
```
badisa, line 86: Should we specify >3.0.4 as it has a bug that will break the estimator by returning a dict?

versioneer.py
```diff
@@ -0,0 +1,2109 @@
+
+# Version: 0.21
+
+"""The Versioneer - like a rocketeer, but for versions.
```
badisa, line 4: Oh wow this is far less clever than I expected...

badisa (approved):

setup.py
mcwitt: this was for `pymbar`, correct?

badisa: Yes, sorry I selected the wrong row. It is `pymbar`

mcwitt: Added in e0097c5

versioneer.py
mcwitt: hopefully in a good way, haha

maxentile (approved):
These seem like great changes. Tested out the updated instructions for installing in develop mode and it all worked as expected. (Also quite nice to be able to access `timemachine.__version__` from within a script!)

PR #598, opened 2022-02-02
Move Python barostat to attic

maxentile:
Move Python barostat from https://github.com/proteneer/timemachine/pull/433 to attic. This implementation was immediately obviated by a CUDA implementation in https://github.com/proteneer/timemachine/pull/436, and since then has mostly served to slow down CI.

(Historical note about performance: Was [initially ~15x slower](https://gist.github.com/maxentile/c7a3624c3d9579e281f7787834c709ab) than it needed to be since it used timemachine float64 mode instead of float32 mode, and is still slower than it should be since it uses a high-overhead way of calling timemachine potentials / forces from Python. This general style of implementation of MD sampling loops might foreseeably be of interest in the future if there is a sufficiently low-overhead way to call timemachine potentials / forces from Python / Jax.)


slow_tests/test_barostat.py
```diff
@@ -414,13 +406,14 @@ def test_molecular_ideal_gas():
         v_0 = sample_velocities(masses * unit.amu, temperature)
 
         # rescale the box to be approximately the desired box volume already
-        rescaler = CentroidRescaler(group_indices)
+        # rescaler = CentroidRescaler(group_indices)
```
mcwitt, line 409: for intentionally commented-out lines, would it make sense to leave a line comment explaining when the code might be useful in the future and what will be the effect of uncommenting it?

maxentile: That was in error! Addressed in https://github.com/proteneer/timemachine/pull/598/commits/851bda71f746fd5a3de7702208fa5ad9c6b60005

(Had temporarily commented it out when checking if it was necessary for the test, forgot to delete after checking.)

maxentile: >for intentionally commented-out lines, would it make sense to leave a line comment explaining when the code might be useful in the future and what will be the effect of uncommenting it?

This sounds like a good practice!

maxentile:
Build is currently failing with flaky curand issue

badisa (approved):

proteneer:
Is the concern simply that it's slowing tests down? I don't think these tests are responsbile for any of the flakiness we've seen in CI. I actually still refer to the various classes in the Barostat periodically (esp when trying to craft new MC moves). I'd be inclined to keep them and figure out how to speed up the barostat tests later.

proteneer:
PS (one way we can speed this up is through the use of a DeviceBuffer wrapper that @mcwitt had introduced)

maxentile:
>Is the concern simply that it's slowing tests down? I don't think these tests are responsbile for any of the flakiness we've seen in CI. I actually still refer to the various classes in the Barostat periodically (esp when trying to craft new MC moves). I'd be inclined to keep them and figure out how to speed up the barostat tests later.

Main concerns: adds a slow test, is unused, and increases confusion about what to find in the `md` module.

It may be helpful to read this code for design ideas, and in the future we might want to use it, but for now I'd be inclined to move it out of the main `md` module.

Perhaps `attic` isn't right though -- would another category be more appropriate for code like this? (`experimental`? `prototype`? `incubator`?)

>PS (one way we can speed this up is through the use of a DeviceBuffer wrapper that @mcwitt had introduced)

Agreed!

proteneer:
> Main concerns: adds a slow test, is unused

Don't we use the test to verify the correctness of the C++ MonteCarloBarostat? This is pretty critical (especially if someone modifies the C++ implementation later on).

> It may be helpful to read this code for design ideas, and in the future we might want to use it, but for now I'd be inclined to move it out of the main md module. Perhaps attic isn't right though -- would another category be more appropriate for code like this? (experimental? prototype? incubator?)

Maybe we should properly come up with a `mc` module (or just rename `md` to `sampling` like you suggest earlier)?

> Perhaps attic isn't right though

I would not put this in the attic

maxentile:
>Don't we use the test to verify the correctness of the C++ MonteCarloBarostat? This is pretty critical (especially if someone modifies the C++ implementation later on).

I don't think so.

(`test_barostat.py` does not compare to the behavior of the production barostat to the Python barostat, and `test_barostat_water_density.py` does not use the C++/CUDA barostat.)

>Maybe we should properly come up with a mc module (or just rename md to sampling like you suggest earlier)?
>I would not put this in the attic

These might be better -- will withdraw this PR and open an issue about this. (May be better to further reorganize the current contents, which is a mix of at least two categories: simulation preparation tools (builders.py, fire.py, ...) and samplers + associated abstractions (moves.py, ensembles.py, ...))

Issue #549, opened 2021-12-13
Rename `fix_nvidia_fmad(a, b, c, d)` to `strict_mul(a, b) + strict_mul(c, d)`?


maxentile:
Migrating from https://github.com/proteneer/timemachine/pull/542#discussion_r767894876 .

Possible small readability improvement for the workaround introduced in https://github.com/proteneer/timemachine/pull/387 to address the issue described in https://github.com/proteneer/timemachine/issues/386

PR #599, opened 2022-02-02
Rename fix_nvidia_fmad

maxentile:
Possible small readability improvement related to the issue in https://github.com/proteneer/timemachine/issues/386 .

If accepted, closes https://github.com/proteneer/timemachine/issues/549 .

### DIFF
--- a/timemachine/cpp/src/kernels/nonbonded_common.cuh
+++ b/timemachine/cpp/src/kernels/nonbonded_common.cuh
@@ -97,10 +97,6 @@ k_add_ull_to_ull(const int N, const unsigned long long *__restrict__ src, unsign
 
 // These are two lines of code are to deal with the formation of a non-commutative fma.
 // For more information, see: https://github.com/proteneer/timemachine/issues/386
-float __device__ __forceinline__ fix_nvidia_fmad(float a, float b, float c, float d) {
-    return __fmul_rn(a, b) + __fmul_rn(c, d);
-}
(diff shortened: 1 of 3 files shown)


timemachine/cpp/src/kernels/nonbonded_common.cuh
```diff
@@ -97,10 +97,6 @@ k_add_ull_to_ull(const int N, const unsigned long long *__restrict__ src, unsign
 
 // These are two lines of code are to deal with the formation of a non-commutative fma.
 // For more information, see: https://github.com/proteneer/timemachine/issues/386
-float __device__ __forceinline__ fix_nvidia_fmad(float a, float b, float c, float d) {
-    return __fmul_rn(a, b) + __fmul_rn(c, d);
-}
+float __device__ __forceinline__ strict_mul(float a, float b) { return __fmul_rn(a, b); }
```
proteneer, line 100: I think this is less clear. The multiply operation *is* associative. It's the fused multiply add that introduces this issue. The old version is more explicit about dealing with *only* fma operations, where as this seems to obfuscate things more.

maxentile: Fair enough!

(Personally I found the phrase `fix_nvidia` slightly distracting and wanted to be able to read off:
(1) what mathematical operation is being performed on `a`, `b`, `c`, `d`,
(2) that it is important these multiplications are strictly translated into multiply instructions and not fused into FMA instructions.)

PR #604, opened 2022-02-02
Update jitify

proteneer:
This updates jitify to address a memory leak.


timemachine/cpp/src/vendored/jitify.hpp
```diff
@@ -2477,9 +2522,9 @@ inline void detect_and_add_cxx17_flag(std::vector<std::string>& options) {
       return;
     }
   }
-  // Jitify must be compiled with C++17 support, so we default to enabling it
```
badisa, line 2480: I believe we need to bump this back to C++17 unless something has changed since we last used this jitify script.

If it is still necessary we should put a comment at the top that the change from 11 to 17 is required

proteneer: in our CMakeLists.txt we already have:
```
set(CMAKE_CXX_STANDARD 17)
```

badisa: Looks like the build failed which I am pretty sure because it tries to use C++ 11 rather than 17. Not entirely clear how it is different from the CMakeLists.txt

proteneer: hm okay - I have no idea why the tests pass locally on my desktop, but will revert to c++17 before commenting further.

proteneer:
Thanks - upgrading to c++17 fixed it!

PR #608, opened 2022-02-03
Modifies Integrator's N_ to be an integer

badisa:
* Also makes it a const
* Adds call to set the stream of the RNG to ensure
  it is running in the same stream as specified by the API call


mcwitt:
Nice catch! I wonder if it'd make sense to turn on `-Werror=conversion` by default? I tried building with this locally and found a fair number (maybe ~10-100) cases, but it seemed like it might be viable to fix them? (not suggesting in this PR)

Issue #609, opened 2022-02-03
Speed up slow tests in test_md.py


maxentile:
From https://github.com/proteneer/timemachine/pull/607#issue-1123513657

```
319.21s call     tests/test_md.py::test_reference_langevin_integrator_with_custom_ops
120.34s call     tests/test_md.py::test_reference_langevin_integrator
```

~7 minutes in these two tests alone

maxentile:
Locally these appear much faster:

* [tests/test_md.py::test_reference_langevin_integrator_with_custom_ops](https://github.com/proteneer/timemachine/blob/e99b61b4b565cddb0870fb009f2ebb51336edcb8/tests/test_md.py#L306-L382) is ~ 8 seconds locally (rather than ~320 seconds). (And it only performs 3000 timesteps of MD...)
* [tests/test_md.py::test_reference_langevin_integrator](https://github.com/proteneer/timemachine/blob/e99b61b4b565cddb0870fb009f2ebb51336edcb8/tests/test_md.py#L264-L303) is ~40 seconds locally (rather than ~120 seconds) 

Propose to:
* move both of these to `slow_tests/` to avoid cuda-memcheck cost (since the tests are designed to check sampler fidelity and numerical stability, not memory safety)
* reduce the number of [setting combinations](https://github.com/proteneer/timemachine/blob/e99b61b4b565cddb0870fb009f2ebb51336edcb8/tests/test_md.py#L273-L277) checked from 16 to 2

proteneer:
sgtm - pr welcome!

maxentile:
After https://github.com/proteneer/timemachine/pull/610, the timings above ~10 seconds are now:

**tests/**
```
============================ slowest 100 durations =============================
207.71s call     tests/test_rabfe.py::TestRABFEModels::test_predict_complex_decouple
181.34s call     tests/test_rabfe.py::TestRABFEModels::test_predict_complex_conversion
117.47s call     tests/test_nonbonded.py::TestNonbonded::test_nonbonded
98.91s call     tests/test_nonbonded.py::TestNonbondedDHFR::test_correctness
84.95s call     tests/test_parameter_interpolation.py::TestInterpolatedPotential::test_nonbonded_advanced
78.59s call     tests/test_jit.py::TestJIT::test_random_directory
48.57s call     tests/test_nonbonded.py::TestNonbondedWater::test_nblist_box_resize
[...]

PR #610, opened 2022-02-04
Speed up slow tests in test_md.py

maxentile:
Addresses https://github.com/proteneer/timemachine/issues/609

------

Tangent: A more general task this suggests might be to find a more fine-grained way to apply cuda-memcheck in our test suite. (By default, we run all tests with cuda-memcheck unless moved to `slow_tests/`. This can be slow, especially for tests that involve iteration. Perhaps for some tests we would want to put the full version in `slow_tests/` and an iteration-reduced version (with asserts removed) in `tests/`? Not sure a clean way to do that...)

### DIFF
--- a/slow_tests/test_reference_langevin_integrator.py
+++ b/slow_tests/test_reference_langevin_integrator.py
@@ -1,0 +1,128 @@
+from itertools import product
+
+import numpy as np
+from jax import numpy as jnp, jit, grad
+
+from timemachine.constants import BOLTZ
+from timemachine.integrator import LangevinIntegrator
+
+
+def test_reference_langevin_integrator(threshold=1e-4):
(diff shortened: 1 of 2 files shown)


badisa (approved):

mcwitt (approved):
Nice! Tangential comment -- I wonder if it'd make sense in a future PR to move tests to a common directory and mark slow tests using [`@pytest.mark.slow`](https://docs.pytest.org/en/6.2.x/mark.html). I think this would allow a little more flexibility in how tests are run (e.g. could run both fast and slow versions of a test for a single feature with a command), and would make it possible to change whether tests are classified as "slow" in the future with smaller diffs.

maxentile:
Pytest markers sound useful here, both to avoid spurious large diffs, and to allow more fine-grained control.

Issue #600, opened 2022-02-02
Use jax.config.update("jax_enable_x64", True) more precisely


badisa:
## Goal
* Provide control over the precision of jax
~* Avoid flake8 linting issues regarding imports not at top of file (#596)~

## Issues
* Jax on the GPU has a hard time with 64 bit
* Want to be able to use 32 bit potentials

proteneer:
There are definitely use cases where we may want to use 32-bit versions of the potentials. It's pretty dangerous to forcefully set 64bit when we import the timemachine module. I think we need to be a bit more careful about how this is set (because it will be very hard to undo afterwards).

badisa:
Could you specify how we want this to function? Right now we have the call to configure it to use 64 bit in enough places that I would imagine almost all code uses the 64 bit version.

proteneer:
What about having this be set where appropriate, but more broadly.

For tests (where this 64bit critical), I think we can call `jax.config.update` during `setUpClass` of the TestCase base class, or have this be invoked by a pytest hook. For things like the endpoint correction in production (outside of tests), we probably still want to include this in the main runner script.

The problem is that jax has no way right now to specify the precision of a given operation (similar to how we can specify `.impl(precision)`). Given that we may at some point use jax on the GPU, which typically have crippled 64bit flops, we want to leave ourselves with an out.

badisa:
Done in https://github.com/proteneer/timemachine/pull/867

PR #613, opened 2022-02-04
Add isort

mcwitt:
Experiment with adding [`isort`](https://github.com/PyCQA/isort) to the pre-commit configuration.

`isort` automatically tidies imports, including
- removing duplicate imports
- categorizing in sections by stdlib, 3rd-party, 1st-party (configurable)
- sorting alphabetically

In this PR `isort` is configured with pre-commit, so it could be enabled as a git hook and would enforced in CI. There are also [editor plugins](https://github.com/pycqa/isort/wiki/isort-Plugins) available to sort imports on file save (looks like VS Code, Sublime, and Emacs are among the supported editors).

This PR configures `honor_noqa = true`, so tidying can be disabled with a `# noqa` comment on the import line

### DIFF
--- a/examples/hif2a/fit_to_multiple_rbfes.py
+++ b/examples/hif2a/fit_to_multiple_rbfes.py
@@ -1,47 +1,40 @@
 # Fit to multiple relative binding free energy edges
+import datetime
 import sys
 from argparse import ArgumentParser
+from collections import defaultdict, namedtuple
+from pathlib import Path
+from pickle import dump, load
+from time import time
+from typing import Any, Dict, List, Tuple, Union
+
 import jax
-from jax import numpy as jnp
 import numpy as np
-import datetime
+from jax import numpy as jnp
+
 import timemachine
-from timemachine.training.dataset import Dataset
+from timemachine.fe.estimator import SimulationResult
+
+# free energy classes
+from timemachine.fe.free_energy import RBFETransformIndex, RelativeFreeEnergy, construct_lambda_schedule
+from timemachine.fe.loss import pseudo_huber_loss  # , l1_loss, flat_bottom_loss
+from timemachine.fe.model import RBFEModel
 
 # forcefield handlers
 from timemachine.ff import Forcefield
-from timemachine.ff.handlers.serialize import serialize_handlers
 from timemachine.ff.handlers.deserialize import deserialize_handlers
 from timemachine.ff.handlers.nonbonded import AM1CCCHandler, LennardJonesHandler
-
-# free energy classes
-from timemachine.fe.free_energy import (
-    RelativeFreeEnergy,
-    construct_lambda_schedule,
-    RBFETransformIndex,
-)
-from timemachine.fe.estimator import SimulationResult
-from timemachine.fe.model import RBFEModel
-from timemachine.fe.loss import pseudo_huber_loss  # , l1_loss, flat_bottom_loss
+from timemachine.ff.handlers.serialize import serialize_handlers
 
 # MD initialization
 from timemachine.md import builders
+from timemachine.optimize.step import truncated_step
+from timemachine.optimize.utils import flatten_and_unflatten
 
 # parallelization across multiple GPUs
 from timemachine.parallel.client import CUDAPoolClient, GRPCClient
 from timemachine.parallel.utils import get_gpu_count
-
-from collections import namedtuple, defaultdict
-
-from pickle import load, dump
-
-from timemachine.optimize.step import truncated_step
-from timemachine.optimize.utils import flatten_and_unflatten
-
-from typing import Tuple, Dict, List, Union, Any
-
-from pathlib import Path
-from time import time
+from timemachine.training.dataset import Dataset
 
 array = Union[np.array, jnp.array]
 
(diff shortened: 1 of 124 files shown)


mcwitt:
Closing as there are currently cases where specifying the import ordering is important (e.g. importing `openmm` before `timemachine` breaks `cuda-memcheck`).

mcwitt:
> Closing as there are currently cases where specifying the import ordering is important (e.g. importing openmm before timemachine breaks cuda-memcheck).

The issue with importing `openmm` first breaking `cuda-memcheck` should have been fixed already upstream and/or by using OpenMM with CUDA disabled, so this may not be a blocker.

Reopening for discussion.

examples/hif2a/fit_to_multiple_rbfes.py
```diff
@@ -1,47 +1,44 @@
...
+
+# forcefield handlers
+from timemachine.ff import Forcefield
+from timemachine.ff.handlers.deserialize import deserialize_handlers
+from timemachine.ff.handlers.nonbonded import AM1CCCHandler, LennardJonesHandler
+from timemachine.ff.handlers.serialize import serialize_handlers
 
 # MD initialization
 from timemachine.md import builders
+from timemachine.optimize.step import truncated_step
```
mcwitt, line 35: unfortunately `isort` does seem to mangle annotated import groups. Interleaved comments keep the groups separate, so we might just manually fix these cases.

examples/validate_relative_binding.py
```diff
@@ -8,43 +8,40 @@
 # 4) Release the restraints attached to the blocker
 # Note that 2) and 4) are done directly via an endpoint correction.
 
+import argparse
```
proteneer, line 11: What's the reason for moving these 3 above `os` and `pickle`?

proteneer: Furthermore, this seems to break the comment into two sections as opposed to keeping them together

proteneer:
Indirectly related comment, now that we have black, flake8, and soon isort, is there a canonical ordering in which these tools should be ran?

maxentile:
I think we want to be able to configure so that:
* setting Jax precision mode to double happens as close to the top of each file as possible (since these configuration flags only work at start-up: https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#Double-(64bit)-precision ) However, this might be tricky to configure in isort since it's not an import? This appears in these two forms: https://github.com/proteneer/timemachine/blob/a1877dfa2337816c9e8b753f23d558cc05fbb6d9/tests/test_lambda_potential.py#L4 or https://github.com/proteneer/timemachine/blob/a1877dfa2337816c9e8b753f23d558cc05fbb6d9/tests/test_jax_utils.py#L3 
* optional dependencies are maintained. I think the only case where we do this is putting openeye imports inside functions, e.g. https://github.com/proteneer/timemachine/blob/a55a509a9db94b346b7ad75b02514ededd87e119/timemachine/ff/handlers/nonbonded.py#L56-L57
* I think there was a jax vs. rdkit import order issue but I don't remember the details (import jax before rdkit?)

--------

>The issue with importing openmm first breaking cuda-memcheck should have been fixed already upstream and/or by using OpenMM with CUDA disabled, so this may not be a blocker.

I didn't properly follow this issue, but this might be a bonus if specifying `timemachine` import group must proceed `openmm` import group provides a second way to resolve the issue short of building CUDA-disabled OpenMM from source.

badisa:
> I think we want to be able to configure so that:
> 
> * setting Jax precision mode to double happens as close to the top of each file as possible (since these configuration flags 

Based on discussion in #600 it would seem that this is something we need to be more careful about than trusting an automated tool to handle this?

maxentile:
Ahh, thanks -- I had missed that consideration. (Not aware of any rules we are manually applying that were more context-specific than "if `config.update("jax_enable_x64", True)` appears in a file, it should appear before any other use of jax".)

Anyway, this may be a non-issue, unless `isort` also changes the ordering of non-import statements like `config.update("jax_enable_x64", True)` (didn't see any examples of this when skimming over the PR changes)

examples/validate_relative_binding.py
mcwitt: Good catch, I was also puzzled by this one. I think the comment is considered to be attached to the following import line (in this case `import os`), and since `ensure_newline_before_comments` is enabled (by `profile = "black"`), isort inserts an extra newline before the comment.

mcwitt:
> setting Jax precision mode to double happens as close to the top of each file as possible

> Anyway, this may be a non-issue, unless isort also changes the ordering of non-import statements like config.update("jax_enable_x64", True) (didn't see any examples of this when skimming over the PR changes)

Agreed; I also don't see any cases where non-import statements (or even import statements) that were previously below `config.update("jax_enable_x64", True)` were moved above, so this seems safe to me.

proteneer (approved):
lgtm!

badisa (approved):

versioneer.py
```diff
@@ -1838,6 +1838,7 @@ def run(self):
 
     if "cx_Freeze" in sys.modules:  # cx_freeze enabled?
         from cx_Freeze.dist import build_exe as _build_exe
+
```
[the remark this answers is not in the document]
mcwitt: good catch, thanks! Fixed in c15ba82.

maxentile (approved):
Looks good! One more thing that can be handled automatically

mcwitt:
Rebased so that we can add 0fa1f8d to `.git-blame-ignore-revs` if desired

PR #614, opened 2022-02-04
Task/#612 move parallel to timemachine namespace

### LINKED ISSUE #612: Refactor `proto/` module and move `parallel` to `timemachine/parallel/`
## Goal
* Move under the `timemachine.parallel` namespace
* Add way to rebuild grpc code

## Proposed Solution

* Add make command to regenerate and place inside of `timemachine.parallel`

badisa:
Resolves #612


maxentile (approved):
Nice!

Possibly one more place to update:
* https://github.com/proteneer/timemachine/blob/82ad1a0f2f33b8be5b6606644d7300aeaa54a0ff/setup.cfg#L16

Quick question: where is this script executed? https://github.com/proteneer/timemachine/blob/82ad1a0f2f33b8be5b6606644d7300aeaa54a0ff/timemachine/parallel/launch_workers.sh (ran across it when searching for any straggler paths to update)

badisa:
> Nice!
> 
> Possibly one more place to update:
> 
> * https://github.com/proteneer/timemachine/blob/82ad1a0f2f33b8be5b6606644d7300aeaa54a0ff/setup.cfg#L16

Good catch!


> Quick question: where is this script executed? https://github.com/proteneer/timemachine/blob/82ad1a0f2f33b8be5b6606644d7300aeaa54a0ff/timemachine/parallel/launch_workers.sh (ran across it when searching for any straggler paths to update)

That script is run on our servers as a daemon. Not sure if we should keep it or not.

pyproject.toml
```diff
@@ -4,7 +4,7 @@ extend-exclude = '''
 ^/( versioneer.py$
   | timemachine/_version.py$
   | ff/params
```
mcwitt, line 6: not related to this PR, but should we also fix this path here?
```suggestion
  | timemachine/ff/params
```

Makefile
```diff
@@ -15,6 +15,10 @@ build:
 clean:
 	cd $(CPP_DIR) && rm -rf build/ | true
 
+.PHONY: grpc
+grpc:
+	python -m grpc_tools.protoc -I grpc/ --python_out=. --grpc_python_out=. grpc/timemachine/parallel/grpc/service.proto
```
mcwitt, line 20: nit: would it make sense to add an intermediate `%.proto` target to avoid rebuilding? (not sure there's much benefit)

badisa: I don't think so? And rebuilding produces the exact same code, so it shouldn't cause and trouble with committing modifications

mcwitt:
> Quick question: where is this script executed? https://github.com/proteneer/timemachine/blob/82ad1a0f2f33b8be5b6606644d7300aeaa54a0ff/timemachine/parallel/launch_workers.sh (ran across it when searching for any straggler paths to update)

tiny nit: would it make sense to update the script path in this file to `timemachine/parallel/worker.py`?

mcwitt (approved):

badisa:
> > Quick question: where is this script executed? https://github.com/proteneer/timemachine/blob/82ad1a0f2f33b8be5b6606644d7300aeaa54a0ff/timemachine/parallel/launch_workers.sh (ran across it when searching for any straggler paths to update)
> 
> tiny nit: would it make sense to update the script path in this file to `timemachine/parallel/worker.py`?

Removed it as I was incorrect in our usage of it. We use `worker.py` directly.

@proteneer do you still use that script?

proteneer:
I don't use it anymore - from way back. Feel free to remove.

Makefile
proteneer: Nice! Good idea putting it in the Makefile as opposed to some obscure comment

proteneer (approved):
lgtm!

Issue #340, opened 2021-02-05
Make testsystems file


proteneer:
Per #337 

we should find all instances where we load in the hif2a set and a pair of ligands and have them load from a common test system function.

files of interest

```
examples/rbfe_single.py
tests/test_free_energy.py
tests/test_topology.py
tests/benchmark.py
in a separate PR soon to come
tests/test_minimizer.py
```

maxentile:
The singleton design in https://github.com/proteneer/timemachine/blob/12bcc2e8d590ea6eded7124aa316ce14a6bfc703/testsystems/relative.py has caused problems, e.g. when an application mutates the test system. Modeling this instead on the testsystems modules in openmmtools, pymbar, or msmbuilder would be a good idea.

proteneer:
I think this is mostly done.

Issue #602, opened 2022-02-02
Standardize ff references


maxentile:
From: https://github.com/proteneer/timemachine/pull/591#pullrequestreview-870943891

>We may want to add a less obtuse way of referencing the default forcefields than open("timemachine/ff/params/...py") but that can be done in a later PR.

There are currently several ways we do this path search -- may be better to access via an import

badisa:
@maxentile Can this be closed with this (https://github.com/proteneer/timemachine/commit/46e4c351ff0de3d47d974a6d3d9ce80c35a51552) change?

maxentile:
Yes!

PR #615, opened 2022-02-05
Move `testsystems/` and `datasets/` to timemachine namespace

maxentile:
Last two moves of the great namespace migration -- did not seem error-prone

--------

Tangent: Reviewing `testsystems/` reminds me to update the singleton pattern in a follow-up: https://github.com/proteneer/timemachine/issues/340#issuecomment-799836192 -- for some reason that commit didn't make it into a PR earlier


proteneer:
This looks straightforward to me! and +1 on removing the singleton later on.

timemachine/testsystems/relative.py
```diff
@@ -3,13 +3,14 @@
...
 
+import timemachine
 from timemachine.fe import free_energy, topology
 from timemachine.ff import Forcefield
 from timemachine.ff.handlers.deserialize import deserialize_handlers
 
 from pathlib import Path
 
-root = Path(__file__).parent.parent
+root = Path(timemachine.__file__).parent.parent
```
proteneer, line 13: huh - didn't know you could do __file__ on a module... does this point to the __init__.py or something?

proteneer (approved):

timemachine/testsystems/relative.py
maxentile: Yep -- exactly!

This way is currently used in a couple other places ([tests/common.py](https://github.com/proteneer/timemachine/blob/a55a509a9db94b346b7ad75b02514ededd87e119/tests/common.py#L34), [RABFE script](https://github.com/proteneer/timemachine/blob/a55a509a9db94b346b7ad75b02514ededd87e119/examples/validate_relative_binding.py#L203)) -- avoiding unnecessary path search is [intended later](https://github.com/proteneer/timemachine/issues/602).

badisa (approved):

mcwitt (approved):

Issue #618, opened 2022-02-08
Assertion error in enhanced.py::identify_rotatable_bonds


maxentile:
Problem:
* Assertion error in identify_rotatable_bonds
https://github.com/proteneer/timemachine/blob/1e414a0e778948441cd3d01e51c62380230c96ea/timemachine/md/enhanced.py#L48-L52 encountered on 125 of 642 FreeSolv molecules

--------

Possible solutions:
* Remove assertion
* Modify assertion from `==` to `>=` (this modified assertion passes on all FreeSolv molecules)
* Better understand cause of mismatch. (At first glance, this does look consistent with https://github.com/rdkit/rdkit/blob/7e575db3c495d5546832be2317291861fa7b1030/rdkit/Chem/Lipinski.py#L37-L64  (up to a missing `uniquify=1`, but adding `uniquify=1` doesn't seem to make a difference...) )

PR #619, opened 2022-02-08
Fix assertion in identify_rotatable_bonds

maxentile:
Addressing https://github.com/proteneer/timemachine/issues/618

-------

Note:

Weakening assertion from `==` to `>=` causes the test to pass.

Adding `uniquify=1` has no apparent effect but looks more consistent with https://github.com/rdkit/rdkit/blob/7e575db3c495d5546832be2317291861fa7b1030/rdkit/Chem/Lipinski.py#L64 .

### DIFF
--- a/tests/test_enhanced.py
+++ b/tests/test_enhanced.py
@@ -1,0 +1,10 @@
+from timemachine.datasets import fetch_freesolv
+from timemachine.md.enhanced import identify_rotatable_bonds
+
+
+def test_identify_rotatable_bonds_runs_on_freesolv():
+    """pass if no runtime errors are encountered"""
+    mols = fetch_freesolv()
+
+    for mol in mols:
+        _ = identify_rotatable_bonds(mol)
--- a/timemachine/datasets/__init__.py
+++ b/timemachine/datasets/__init__.py
@@ -1,3 +1,3 @@
 from timemachine.datasets.utils import fetch_freesolv
 
-__all__ = [fetch_freesolv]
+__all__ = ["fetch_freesolv"]
--- a/timemachine/datasets/__init__.py
+++ b/timemachine/datasets/__init__.py
@@ -1,0 +1,3 @@
+from timemachine.datasets.utils import fetch_freesolv
+
+__all__ = [fetch_freesolv]
--- a/timemachine/datasets/utils.py
+++ b/timemachine/datasets/utils.py
@@ -1,13 +1,11 @@
-from pathlib import Path
+from importlib import resources
 from typing import List
 
 from rdkit import Chem
 
-import timemachine
-
-PATH_TO_FREESOLV = str(Path(timemachine.__file__).parent / "datasets/freesolv/freesolv.sdf")
-
 
 def fetch_freesolv() -> List[Chem.Mol]:
-    supplier = Chem.SDMolSupplier(PATH_TO_FREESOLV, removeHs=False)
-    return [mol for mol in supplier]
+    with resources.path("timemachine", "datasets") as datasets_path:
+        freesolv_path = str(datasets_path / "freesolv" / "freesolv.sdf")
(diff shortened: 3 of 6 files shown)


timemachine/datasets/utils.py
```diff
@@ -0,0 +1,13 @@
+from pathlib import Path
+from typing import List
+
+from rdkit import Chem
+
+import timemachine
+
+PATH_TO_FREESOLV = str(Path(timemachine.__file__).parent.parent / "timemachine/datasets/freesolv/freesolv.sdf")
```
badisa, line 8: Nit: Redundant to do `parent.parent` then reattach `timemachine`

badisa (approved):

timemachine/datasets/utils.py
maxentile: Was intentional, so that path strings are relative to project root (can be more easily found / changed later if we do another reorganization), but probably makes more sense for it to be relative to package root?

Applied in https://github.com/proteneer/timemachine/pull/619/commits/f2401a5b5e4e2a3bf790f0146e36c0f6de070322

tests/test_enhanced.py
```diff
@@ -0,0 +1,10 @@
+from timemachine.datasets import fetch_freesolv
+from timemachine.md.enhanced import identify_rotatable_bonds
+
+
+def test_identify_rotatable_bonds_runs_on_freesolv():
+    """pass if no runtime errors are encountered"""
+    mols = fetch_freesolv()
+
+    for mol in mols:
```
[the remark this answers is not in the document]
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/619/commits/5e22e8b787ab11cf2b03a110d4b7819672e4107d

timemachine/datasets/__init__.py
```diff
@@ -0,0 +1,3 @@
+from timemachine.datasets.utils import fetch_freesolv
+
+__all__ = [fetch_freesolv]
```
mcwitt, line 3: nit: I've found this a bit unclear in the past, but I think `__all__` is supposed to be a list of strings? ([reference](https://docs.python.org/3/reference/simple_stmts.html#:~:text=the%20public%20names%20defined%20by%20a%20module%20are%20determined%20by%20checking%20the%20module%E2%80%99s%20namespace%20for%20a%20variable%20named%20__all__%3B%20if%20defined%2C%20it%20must%20be%20a%20sequence%20of%20strings%20which%20are%20names%20defined%20or%20imported%20by%20that%20module.))

timemachine/datasets/utils.py
```diff
@@ -0,0 +1,13 @@
+from pathlib import Path
+from typing import List
+
+from rdkit import Chem
+
+import timemachine
+
+PATH_TO_FREESOLV = str(Path(timemachine.__file__).parent / "datasets/freesolv/freesolv.sdf")
```
mcwitt, line 8: tangential -- if we add `datasets` to `package_data` in `setup.py`, maybe this could eventually be something like
```python
with importlib.resources.path("timemachine", "datasets") as path:
    PATH_TO_FREESOLV = path / "freesolv" / "freesolv.sdf"
    # ...
```
https://docs.python.org/3.8/library/importlib.html?highlight=importlib#importlib.resources.path

mcwitt (approved):

timemachine/datasets/utils.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/619/commits/0c434fe4d71fec2f8226e45f22edac5cd413f853

timemachine/datasets/utils.py
maxentile: Superseded by https://github.com/proteneer/timemachine/pull/619/commits/0c434fe4d71fec2f8226e45f22edac5cd413f853

PR #620, opened 2022-02-08
Modifies Integrator's N_ to be an integer

badisa:
* Also makes it a const
* Adds call to set the stream of the RNG to ensure
  it is running in the same stream as specified by the API call

### DIFF
--- a/timemachine/cpp/src/integrator.cu
+++ b/timemachine/cpp/src/integrator.cu
@@ -37,9 +37,9 @@ __global__ void update_forward(
     const int N,
     const int D,
     const RealType ca,
-    const RealType *cbs, // N x 3, not P x N x 3, but we could just pass in the first index
-    const RealType *ccs, // N
-    const RealType *noise,
+    const RealType *cbs,   // N
+    const RealType *ccs,   // N
+    const RealType *noise, // N x 3
     RealType *x_t,
     RealType *v_t,
     const unsigned long long *du_dx,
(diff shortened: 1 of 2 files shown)


timemachine/cpp/src/integrator.cu
```diff
@@ -37,9 +37,9 @@ __global__ void update_forward(
     const int N,
     const int D,
     const RealType ca,
-    const RealType *cbs, // N x 3, not P x N x 3, but we could just pass in the first index
-    const RealType *ccs, // N
-    const RealType *noise,
+    const RealType *cbs,   // N x 3, not P x N x 3, but we could just pass in the first index
+    const RealType *ccs,   // N
+    const RealType *noise, // [N x 3]
```
maxentile, line 42: Tangent: Should these shape comments not all be N x 3 ?

maxentile (approved):
Looks good!

timemachine/cpp/src/integrator.cu
badisa: We only access `ccs` using the `atom_idx`. Which makes sense since it contains (based on the comment above its usage) `sqrt(1 - exp(-2 * friction * dt)) * sqrt(kT / mass)` which is independent of the dimensions

maxentile: Ahh, so should `cbs` also be N rather than N x 3? (Also only accessed using `atom_idx`)

mcwitt (approved):

proteneer (approved):
lgtm! at some point it might be useful to have `dt_` and `ca_` be adjustable on the fly (eg: for minimization), but we don't have setters for those anyways right now so I'm happy to have them be const.

PR #621, opened 2022-02-08
Include isort, flake8 in dev extras

mcwitt:
Small quality of life improvement: includes `isort==5.10.1` and `flake8==4.0.1` (the same versions used by pre-commit) in dev extras so that the proper versions can be installed locally as part of

```
pip install -e .[dev]
```

(this isn't necessary for the pre-commit hook, but is useful to use `isort` or `flake8` independently, e.g. for editor integration)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,7 +94,7 @@ setup(
         "scipy",
     ],
     extras_require={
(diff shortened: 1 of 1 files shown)


proteneer (approved):
lgtm!

Issue #623, opened 2022-02-09
IndexError when molecule contains no improper torsions


maxentile:
Parameterization assumes non-empty list of matches here.

Detailed error message:

<details>

```
File "/home/jfass/Documents/GitHub/timemachine/timemachine/fe/free_energy.py", line 119, in prepare_host_edge
    final_params, final_potentials = self._get_system_params_and_potentials(ff_params, hgt)
  File "/home/jfass/Documents/GitHub/timemachine/timemachine/fe/free_energy.py", line 42, in _get_system_params_and_potentials
    combined_params, combined_potential = fn(*params)
  File "/home/jfass/Documents/GitHub/timemachine/timemachine/fe/topology.py", line 196, in parameterize_periodic_torsion
    proper_params, improper_params
  File "/home/jfass/Documents/GitHub/timemachine/timemachine/fe/topology.py", line 322, in parameterize_periodic_torsion
    improper_params, improper_potential = self.parameterize_improper_torsion(improper_params)
  File "/home/jfass/Documents/GitHub/timemachine/timemachine/fe/topology.py", line 314, in parameterize_improper_torsion
    params, idxs = self.ff.it_handle.partial_parameterize(ff_params, self.mol)
  File "/home/jfass/Documents/GitHub/timemachine/timemachine/ff/handlers/bonded.py", line 174, in partial_parameterize
    return self.static_parameterize(params, self.smirks, mol)
  File "/home/jfass/Documents/GitHub/timemachine/timemachine/ff/handlers/bonded.py", line 216, in static_parameterize
    return params[param_idxs], np.array(improper_idxs, dtype=np.int32)
IndexError: arrays used as indices must be of integer (or boolean) type
```

</details>

------------

To reproduce:
```python
from timemachine.md import enhanced

ff = ... # e.g. openff 1.1.0 + am1ccc
mol = ... # something that won't be assigned impropers by chosen ff, e.g. mobley_1019269 in FreeSolv

_ = enhanced.get_solvent_phase_system(mol, ff) # error
```

-----------

Possible solutions:
* Add guard for empty `improper_params` in `HostGuestTopology::parameterize_periodic_torsion`
* Add guard for empty `improper_params` in `BaseTopology::parameterize_periodic_torsion`
* Add guard after `param_idxs` are computed from `mol` and `smirks` in `ImproperTorsionHandler::static_parameterize(params, smirks, mol)`, before `params[param_idxs]` is returned
[...]

PR #626, opened 2022-02-10
Handle case of no SMIRKS matches in {Proper/Improper}TorsionHandler

maxentile:
Resolves https://github.com/proteneer/timemachine/issues/623 , and applies the same patch for proper torsions while we're at it

### DIFF
--- a/timemachine/ff/handlers/bonded.py
+++ b/timemachine/ff/handlers/bonded.py
@@ -135,7 +135,15 @@ class ProperTorsionHandler:
 
         scatter_idxs = np.array(scatter_idxs)
 
-        return params[scatter_idxs], np.repeat(torsion_idxs, repeats, axis=0).astype(np.int32)
+        # if no matches found, return arrays that can still be concatenated as expected
+        if len(param_idxs) > 0:
+            assigned_params = params[param_idxs]
+            proper_idxs = np.repeat(torsion_idxs, repeats, axis=0).astype(np.int32)
+        else:
+            assigned_params = params[[False] * len(params)]  # empty slice with same dtype, other dimensions
+            proper_idxs = np.zeros((0, 4), dtype=np.int32)
+
+        return assigned_params, proper_idxs
 
     def serialize(self):
         list_params = []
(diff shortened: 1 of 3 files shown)


timemachine/ff/handlers/bonded.py
```diff
@@ -213,4 +217,12 @@ def make_key(idxs):
 
         param_idxs = np.array(param_idxs)
 
-        return params[param_idxs], np.array(improper_idxs, dtype=np.int32)
+        # if no matches found, return arrays that can still be concatenated as expected
```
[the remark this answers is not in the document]
maxentile: A simulation can be valid even if it contains no improper torsion terms. (E.g. simulation of a water box is valid, even though it contains no torsion terms.)

I think the desired behavior when the force field assigns no improper torsion terms to a mol is to generate potentials with no improper torsions (rather than raise an index error).

proteneer (approved):

timemachine/ff/handlers/bonded.py
```diff
@@ -139,7 +135,15 @@ def static_parameterize(params, smirks, counts, mol):
 
         scatter_idxs = np.array(scatter_idxs)
 
-        return params[scatter_idxs], np.repeat(torsion_idxs, repeats, axis=0).astype(np.int32)
+        # if no matches found, return arrays that can still be concatenated as expected
+        if len(param_idxs) > 0:
+            assigned_params = params[scatter_idxs]
+            proper_idxs = np.repeat(torsion_idxs, repeats, axis=0).astype(np.int32)
+        else:
+            assigned_params = params[[False] * len(params)]  # empty slice with same dtype, other dimensions
```
mcwitt, line 143: This confused me at first so I tried to come up with some alternatives: `params[slice(0)]`, `params[0:0]` (not sure if actually clearer but a bit more concise)

mcwitt (approved):

timemachine/ff/handlers/bonded.py
maxentile: Either of these alternatives would be an improvement, I think!

Switched to `params[:0]` and retained the comment in https://github.com/proteneer/timemachine/pull/626/commits/98c6a966c7e6e5e25369aeb98ee37ab78ba0052f

PR #629, opened 2022-02-12
Feature/#625 switch to gitlab

badisa:
A draft to get eyes on what I have so far. It seems to be working, though it's not the prettiest thing. 

Still needs to be hooked up to github.

## Notes
- The current gitlab mirroring, via pull by gitlab, takes a LONG time (seen up to 30 minutes) 
- Builds a base docker container, could be the full container as well, wasn't sure if we had a use for the extra layers


BaseDockerfile
```diff
@@ -0,0 +1,75 @@
...
+      -DOPENMM_BUILD_DRUDE_CUDA_LIB=OFF \
+      -DOPENMM_BUILD_DRUDE_OPENCL_LIB=OFF \
+      -DOPENMM_BUILD_EXAMPLES=OFF \
+      -DOPENMM_BUILD_OPENCL_LIB=OFF \
+      -DOPENMM_BUILD_RPMD_CUDA_LIB=OFF \
+      -DOPENMM_BUILD_RPMD_OPENCL_LIB=OFF \
+      -DCMAKE_INSTALL_PREFIX=../openmm_install \
+      ../ && \
+    make -j "$(nproc)" && \
+    make -j "$(nproc)" install && \
```
mcwitt, line 70: Is it possible to move the OpenMM build before the creation of the timemachine conda env? (would probably still need to have the base env activated..) The hope would be to avoid needing to rebuild OpenMM from scratch if e.g. there's only a change to `environment.yml`

mcwitt: Tangentially, I wonder if there's a way to move the OpenMM stuff into a separate Dockerfile using a [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds)? I haven't done this before, but it seems potentially nice for modularization (maybe at the cost of having more images to track)

(not intended to be a PR blocker)

.gitlab-ci.yml
```diff
@@ -0,0 +1,27 @@
...
+    - docker build -t $CI_REGISTRY_IMAGE:$CI_BUILD_REF -f BaseDockerfile .
+    - docker push $CI_REGISTRY_IMAGE:$CI_BUILD_REF
+
+unit-tests:
+  stage: test
+  image: $CI_REGISTRY_IMAGE:$CI_BUILD_REF
+  variables:
+    CMAKE_ARGS: -DCUDA_ARCH=75
+  script:
+    - CMAKE_BUILD_PARALLEL_LEVEL=$(nproc) pip install -e .[dev,test]
```
mcwitt, line 20: should we also `pip install -r requirements.txt` here?

.gitlab-ci.yml
```diff
@@ -0,0 +1,27 @@
...
+  - test
+
+
+docker_build:
+  stage: build-docker
+  image: docker:20.10
+  script:
+    - DOCKER_HOST=tcp://localhost:2375 docker login -u $CI_REGISTRY_USER -p $CI_JOB_TOKEN $CI_REGISTRY
+    - docker build -t $CI_REGISTRY_IMAGE:$CI_BUILD_REF -f BaseDockerfile .
+    - docker push $CI_REGISTRY_IMAGE:$CI_BUILD_REF
```
mcwitt, line 12: Does this step add a lot of time to CI runs? Is there a way to only rebuild less frequently than every commit, e.g. only if the CI Dockerfile has changed?

mcwitt (commented):
:tada:

mcwitt:
> The current gitlab mirroring, via pull by gitlab, takes a LONG time (seen up to 30 minutes)

Interesting, the default interval should be 5 minutes if things are working right. I wonder if we've configured it higher for some reason, or if things were just temporarily flaky.. https://docs.gitlab.com/ee/administration/instance_limits.html#pull-mirroring-interval

.gitlab-ci.yml
badisa: Since this dockerfile is only the base environment it adds no time if the machine has already has the container layers built, which that is correct.

.gitlab-ci.yml
badisa: The dockerfile built in the build stage takes care of that

BaseDockerfile
```diff
@@ -0,0 +1,75 @@
+FROM nvidia/cuda:11.6.0-devel-ubuntu20.04
```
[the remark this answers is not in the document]
badisa: Moving to staged builds which will change it to `Dockerfile` as it will build a complete image, but this will just stop at the base env stage

badisa:
Here is what the coverage looks like now: https://relaytx.gitlab.io/-/physics-platform/timemachine/-/jobs/2085430429/artifacts/coverage/index.html

You can find this by going to the artifact for the unit tests, browse them and look at `index.html`. There is some built in support for xml coverage, but seems to be built into the PRs/editor of Gitlab so it didn't seem as useful

.gitlab-ci.yml
mcwitt: Ah, got it; I guess we might want to revisit this if we move to autoscaling/ephemeral instances in the future, but makes sense for persistent instances

mcwitt: A couple other questions/comments here:
- would this push a new image to ECR on every commit to a feature branch? This seems like it could create a lot of clutter
- I couldn't find documentation for `CI_BUILD_REF` [here](https://docs.gitlab.com/ee/ci/variables/index.html#predefined-cicd-variables). It seems like this might be deprecated, at least according to this [issue](https://gitlab.gwdg.de/genepaint/gp3/-/issues/122)?

.gitlab-ci.yml
mcwitt: Ah, I see. But I notice the dockerfile also contains `pip install -e .[dev,test]` -- by this logic, do we need to repeat this here? If we rebuild the image on every CI run, this seems redundant?

.gitlab-ci.yml
badisa: This will only push layers that have changed, which since the base stage is cached based on the `environment.yml` and `requirements.txt` should change relatively rarely. It does create a lot of cluttered tags, which we can set to clean up on a regular basis. This seemed like a reasonable trade off to rebuilding everything from scratch on every build, but the build only adds 10 minutes

Good catch, will change to `CI_COMMIT_REF_NAME` which should only create one tag per branch.

.gitlab-ci.yml
badisa: We had discussed this offline, and for posterity: the docker build stage only targets the base environment. This way we have a fully fledged dockerfile in the repo, but only build and store what we need for CI.

badisa:
> > The current gitlab mirroring, via pull by gitlab, takes a LONG time (seen up to 30 minutes)
> 
> Interesting, the default interval should be 5 minutes if things are working right. I wonder if we've configured it higher for some reason, or if things were just temporarily flaky.. https://docs.gitlab.com/ee/administration/instance_limits.html#pull-mirroring-interval
> 
> I guess the docs do describe that option as the "minimum wait time between refreshes", but in practice I think I've rarely/never seen longer than that.

This has now been improved by a web hook. Does still not seem to trigger more often than every 5 minutes, but seems tolerable.

.gitlab-ci.yml
```diff
@@ -0,0 +1,68 @@
...
+  needs: ["docker_build"]
+  tags:
+    - gpu
+  rules:
+    - if: '$CI_EXTERNAL_PULL_REQUEST_IID != null || $CI_EXTERNAL_PULL_REQUEST_IID != ""'
+  variables:
+    CMAKE_ARGS: -DCUDA_ARCH=75
+  script:
+    - CMAKE_BUILD_PARALLEL_LEVEL=$(nproc) pip install -e .[dev,test]
+    - make verify tests
```
mcwitt, line 62: Would it make sense to run `verify` in a separate stage that runs first? I can think of 2 advantages:
- we'd get linting feedback right away (vs. after the docker build)
- wouldn't need to run `verify` in two places (`unit-tests` and `memory-tests`)

Makefile
```diff
@@ -19,8 +19,17 @@ clean:
...
+.PHONY: verify
+verify:
+	pre-commit run --all-files --show-diff-on-failure
+
+.PHONY: memcheck_tests
+memcheck_tests:
+	cuda-memcheck --leak-check full --error-exitcode 1 pytest $(PYTEST_CI_ARGS) tests/
+
+.PHONY: tests
+tests:
```
mcwitt, line 31: nit: slightly surprising to me that `tests` doesn't run all the tests. Would it make sense to rename this to `slow_tests` and add a separate target `tests` depending on both `memcheck_tests` and `slow_tests`?

.gitlab-ci.yml
```diff
@@ -0,0 +1,68 @@
...
+    - if: '$CI_EXTERNAL_PULL_REQUEST_IID != null || $CI_EXTERNAL_PULL_REQUEST_IID != ""'
+
+docker_build:
+  stage: build-docker
+  image: docker:20.10
+  tags:
+    - gpu
+    - docker
+  variables:
+    # For this to work, $CI_COMMIT_REF_SLUG == $CI_DEFAULT_BRANCH
```
mcwitt, line 18: I'm not sure I understand this comment. Is this saying we assume this equality, and something will fail if it's not satisfied?

.gitlab-ci.yml
```diff
@@ -0,0 +1,68 @@
...
+    DOCKER_TAG: $CI_REGISTRY_IMAGE:$CI_COMMIT_REF_SLUG
```
mcwitt, line 19: Now that the code and Dockerfile is contained in a single repo, would it make sense to tag the image with the versioneer version in addition to (or instead of) the commit slug?

mcwitt (approved):
Nice, this looks really good to me. Just had a question about the linting step and some minor comments re naming

.gitlab-ci.yml
badisa: Will clarify the comment, the idea is that the commit slug needs to be equivalent to the default branch. IE slugifying the default branch, in our case `master` must result in `master`. This wouldn't be true if we had something like `this/is-silly` would become `thisissilly`.

.gitlab-ci.yml
badisa: That would prevent the ability to get the last image built for caching, as we would have to determine the last version to pull. It could probably be done, but seems more complicated for no obvious utility.

mcwitt: Ah, sorry, I misunderstood what this is doing. I see now that `CI_COMMIT_REF_SLUG` is the abbreviated branch/tag name, and that we're only building the base image here.

proteneer (approved):
LGTM!

maxentile (approved):

PR #630, opened 2022-02-14
Fix np.testing.assert_almost_equal usage

mcwitt:
[np.testing.assert_almost_equal](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html) expects a number of decimals to check as its 3rd argument, but we instead pass `rtol` in tests for `du_dl` and `du_dp` in `GradientTest.compare_forces`.

It appears that `decimal` between 0 and 1 is interpreted as 0 rather than raising an error. For example, the following succeeds:
```python
np.testing.assert_almost_equal(0.1, 0.2, 1e-4)
```

#### To do
- [x] Investigate cases where we needed to modify tests to pass

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -432,6 +432,7 @@ class GradientTest(unittest.TestCase):
         ref_du_dx, ref_du_dp, ref_du_dl = grad_fn(x, params, box, lamb)
 
         for combo in itertools.product([False, True], repeat=4):
+
             (compute_du_dx, compute_du_dp, compute_du_dl, compute_u) = combo
 
             # do each computation twice to check determinism
--- a/tests/common.py
+++ b/tests/common.py
@@ -443,9 +443,9 @@ class GradientTest(unittest.TestCase):
             if compute_du_dx:
                 self.assert_equal_vectors(np.array(ref_du_dx), np.array(test_du_dx), rtol)
             if compute_du_dl:
-                np.testing.assert_almost_equal(ref_du_dl, test_du_dl, rtol)
+                np.testing.assert_allclose(ref_du_dl, test_du_dl, rtol=rtol)
             if compute_du_dp:
-                np.testing.assert_almost_equal(ref_du_dp, test_du_dp, rtol)
+                np.testing.assert_allclose(ref_du_dp, test_du_dp, rtol=rtol)
 
             test_du_dx_2, test_du_dp_2, test_du_dl_2, test_u_2 = test_impl.execute_selective(
                 x, params, box, lamb, compute_du_dx, compute_du_dp, compute_du_dl, compute_u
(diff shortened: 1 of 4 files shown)


tests/common.py
```diff
@@ -443,9 +443,9 @@ def compare_forces(
             if compute_du_dx:
                 self.assert_equal_vectors(np.array(ref_du_dx), np.array(test_du_dx), rtol)
             if compute_du_dl:
-                np.testing.assert_almost_equal(ref_du_dl, test_du_dl, rtol)
+                np.testing.assert_allclose(ref_du_dl, test_du_dl, rtol=rtol)
```
proteneer, line 446: should we add an `atol` too?

badisa: Do we want a `rtol` at all? Based on what is here: https://numpy.org/doc/stable/reference/generated/numpy.allclose.html it seems like `rtol` might be more flexible than we want? IE if the values are say 1,000,000 a relative tolerance of 1e-4 is 100, when we probably want to say that the values are always different by 0.001?

tests/test_nonbonded.py
```diff
@@ -218,7 +218,7 @@ def test_correctness(self):
                 self.cutoff,
             )
 
-            for precision, rtol in [(np.float64, 1e-8), (np.float32, 1e-4)]:
+            for precision, rtol in [(np.float64, 1e-7), (np.float32, 1e-2)]:
```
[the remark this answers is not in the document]
proteneer: I think 1e-2 is also way too loose here for single precision. Even 1e-7 in double precision seems egregiously loose. We may want to play around with the fixed point exponent a bit (and probably have different ones for force constants vs bond lengths etc.)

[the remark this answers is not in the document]
mcwitt: Agreed, my goal here is just to document what is required to get tests passing again (don't think we should merge these higher tolerances without investigation)

tests/common.py
mcwitt: > Do we want a rtol at all?

Using `atol` instead makes the most sense to me for the fixed-fixed comparisons, but also seems appropriate for fixed-floating like we have here. Maybe requiring both would make sense for fixed-floating?

tests/common.py
```diff
@@ -443,9 +443,9 @@ def compare_forces(
...
             if compute_du_dp:
-                np.testing.assert_almost_equal(ref_du_dp, test_du_dp, rtol)
+                np.testing.assert_allclose(ref_du_dp, test_du_dp, rtol=rtol, atol=atol)
```
maxentile, line 448: nit: `isclose(a, b, rtol=rtol) != isclose(b, a, rtol=rtol)` in general -- the order of `ref_` and `test_` [should probably](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html) be flipped in a few places

proteneer (approved):
LGTM!

badisa (approved):

tests/common.py
mcwitt: Since this occurs in other files, and to not block the bug fix, I will address these all together in an upcoming PR

PR #631, opened 2022-02-14
Fix importlib resources usage

mcwitt:
`importlib.resources.path` is not intended to work with directories; this works coincidentally in Python 3.7 but will break in future versions. See https://bugs.python.org/issue44162


maxentile (approved):

mcwitt:
Closing for now as this fix is not compatible with Python 3.7.

PR #633, opened 2022-02-15
Factorizable Dummy Bonds

proteneer:
This PR lays out much of the ground work for generating factorizable dummy bonds. Given a fixed partitioning of the molecule into core and dummy regions, this code identifies which bonded terms (`1-2, 1-3, 1-4`) can be safely left on without affecting the factorizability of the partition function (such that complex-solvent, or solvent-vacuum edges can cancel out). In particular, the function `flag_bonds(mol, core, bond_idxs)` will return a boolean array of `len(bond_idxs)` indicating which of the terms must be turned off and which of the terms must be turned on. The implementation is technically correct for any choice of core atoms (including those involved in possible ring/opening closing). While the efficiency of the transformation is not guaranteed, an heuristic is made in the form of minimizing the number of interactions that need to be turned off. 

The logic of the code is as follows:
1) Partition dummy atoms into disjoint groups using a heuristic
2) Identify optimal anchoring core atoms for each dummy group. Note that there are many possible choices, and we only apply a heuristic in picking the optimal one.
3) Loop over bond_idxs to see which ones can be safely turned off. 

The code is a little bit terse as-is, and example test cases are welcomed. There are further generalizations one can make later on to support virtual sites etc. 

Glossary of Terms:
-------------------------
dummy atom - an R-group atom that is inserted or deleted in an alchemical transformation
dummy group - a collection of dummy atoms (eg. multiple dummy hydrogen atoms on CH3 can belong to the same dummy group, with C being a core atom).
core atom - not a dummy atom
anchor/core anchor - a core atom that is allowed to interact with the atoms in a given dummy group
anchor group - a set of ordered anchor atoms (up to 3) that can be used to define bond, angle, torsion terms in specialized ways
root-anchor - first atom in an anchor group, also the anchor atom that has direct 1-2 bonds to dummy atoms.
partition - only applies to dummy groups, as we require that dummy groups disjointly partition dummy atoms.

Correctness vs Efficiency vs Numerical Stability 
---------------------------------------------------------------
`enumerate_dummy_ixns` generates interactions that are formally correct, but may be numerically unstable or inefficient.
`flag_bonds` uses a heuristic for efficiency, by first maximizing the number of 1-2 interactions, then the number of 1-3 interactions, and finally the number of 1-4 interactions. 
`verify_compatibility` checks that dummy atom interactions at the end-states are numerically stable. One example we want to avoid is if we allowed a dummy-anchor torsion into a nitrile group (which would result in collinear bonds)

The PR also adds some annotation/drawing code to render the allowed interactions, where the core is defined in (dark/light red), and dummy atoms are everything else. The highlighted bonds show the *allowed* interactions. 

toluene_0
met_biphenyl_0
aspirin_1
aspirin_0

Note that some of them are very tricky, remember that highlighted bonded interactions are the *only* ones left on between dummy-anchor ixns. Below, while 1-2 terms are annihlated between atoms 0 and 7, the 1-3, 1-4 terms can still span 0-7 without affecting the correctness (though they may affect numerical stability in MD if angle terms don't concomitantly have all the bond terms turned on), but this is very tricky for improper terms!

ring_open_0

We can remove these extra interactions (and impropers) by turning on strict mode:
ring_open_0

### DIFF
--- a/tests/test_dummy.py
+++ b/tests/test_dummy.py
@@ -104,8 +104,8 @@ def test_identify_dummy_groups():
 
     {D0, D1, D2}
 
-        D0-D1  D2
-       /       /  -> 1 dummy-anchor bond:
+        D0-D1--D2
+       /          -> 1 dummy-anchor bond:
       0-------1
 
     """
--- a/tests/test_dummy.py
+++ b/tests/test_dummy.py
@@ -1,0 +1,358 @@
+import numpy as np
+from rdkit import Chem
+
+from timemachine.fe import dummy, dummy_draw
+from timemachine.fe.dummy import (
+    enumerate_anchor_groups,
+    enumerate_dummy_ixns,
+    flag_bonds,
+    identify_anchor_groups,
+    identify_dummy_groups,
+    identify_root_anchors,
+)
+from timemachine.ff import Forcefield
+from timemachine.ff.handlers.deserialize import deserialize_handlers
+
+
+def test_identify_root_anchors():
+    mol = Chem.MolFromSmiles("C1CCC1N")
+    core = [0, 1, 2, 3]
+    anchors = identify_root_anchors(mol, core, dummy_atom=4)
+    assert set(anchors) == set([3])
+
+    mol = Chem.MolFromSmiles("C1CC2NC2C1")
+    core = [0, 1, 2, 4, 5]
+    anchors = identify_root_anchors(mol, core, dummy_atom=3)
+    assert set(anchors) == set([2, 4])
+
+    mol = Chem.MolFromSmiles("C1OCC11CCCCC1")
+    core = [3, 4, 5, 6, 7, 8]
+    anchors = identify_root_anchors(mol, core, dummy_atom=0)
+    assert set(anchors) == set([3])
+    anchors = identify_root_anchors(mol, core, dummy_atom=1)
+    assert set(anchors) == set([3])
+    anchors = identify_root_anchors(mol, core, dummy_atom=2)
+    assert set(anchors) == set([3])
+
+    mol = Chem.MolFromSmiles("C1CC1.C1CCCCC1")
+    core = [3, 4, 5, 6, 7, 8]
+    anchors = identify_root_anchors(mol, core, dummy_atom=0)
+    assert set(anchors) == set()
+    anchors = identify_root_anchors(mol, core, dummy_atom=1)
+    assert set(anchors) == set()
+    anchors = identify_root_anchors(mol, core, dummy_atom=2)
+    assert set(anchors) == set()
+
+    mol = Chem.MolFromSmiles("C1CC2NC2C1")
+    core = [0, 1, 2, 5]
+    anchors = identify_root_anchors(mol, core, dummy_atom=3)
+    assert set(anchors) == set([2, 5])
+    anchors = identify_root_anchors(mol, core, dummy_atom=4)
+    assert set(anchors) == set([2, 5])
+
+    # cyclohexane with a nitrogen inside
+    mol = Chem.MolFromSmiles("C1C2CC3CC1N23")
+    core = [0, 1, 2, 3, 4, 5]
+    anchors = identify_root_anchors(mol, core, dummy_atom=6)
+    assert set(anchors) == set([1, 3, 5])
+
+
+def assert_set_equality(a_sets, b_sets):
+    frozen_a = [frozenset(a) for a in a_sets]
+    frozen_b = [frozenset(b) for b in b_sets]
+    assert frozenset(frozen_a) == frozenset(frozen_b)
+
+
+def test_identify_dummy_groups():
+
+    mol = Chem.MolFromSmiles("FC1CC1(F)N")
+    core = [1, 2, 3]
+    dg = identify_dummy_groups(mol, core)
+    assert_set_equality(dg, [{0}, {4, 5}])
+
+    mol = Chem.MolFromSmiles("FC1CC1(F)NN")
+    core = [1, 2, 3]
+    dg = identify_dummy_groups(mol, core)
+    assert_set_equality(dg, [{0}, {4, 5, 6}])
+
+    mol = Chem.MolFromSmiles("C1CC11OO1")
+    core = [0, 1, 2]
+    dg = identify_dummy_groups(mol, core)
+    assert_set_equality(dg, [{3, 4}])
+
+    mol = Chem.MolFromSmiles("C1CC2OOC12")
+    core = [0, 1, 2, 5]
+    dg = identify_dummy_groups(mol, core)
+    assert_set_equality(dg, [{3}, {4}])
+
+
+def assert_anchor_group_equality(a_groups, b_groups):
+    frozen_a = [tuple(a) for a in a_groups]
+    frozen_b = [tuple(b) for b in b_groups]
+    assert frozenset(frozen_a) == frozenset(frozen_b)
+
+
+def test_identify_anchor_groups():
+
+    mol = Chem.MolFromSmiles("C1CC1F")
+    core = [0, 1, 2]
+    groups_of_1, groups_of_2, groups_of_3 = identify_anchor_groups(mol, core, 0)
+    assert_anchor_group_equality(groups_of_1, [[0]])
+    assert_anchor_group_equality(groups_of_2, [[0, 1], [0, 2]])
+    assert_anchor_group_equality(groups_of_3, [[0, 1, 2], [0, 2, 1]])
+
+    groups_of_1, groups_of_2, groups_of_3 = identify_anchor_groups(mol, core, 1)
+    assert_anchor_group_equality(groups_of_1, [[1]])
+    assert_anchor_group_equality(groups_of_2, [[1, 0], [1, 2]])
+    assert_anchor_group_equality(groups_of_3, [[1, 0, 2], [1, 2, 0]])
+
+    groups_of_1, groups_of_2, groups_of_3 = identify_anchor_groups(mol, core, 2)
+    assert_anchor_group_equality(groups_of_1, [[2]])
+    assert_anchor_group_equality(groups_of_2, [[2, 1], [2, 0]])
+    assert_anchor_group_equality(groups_of_3, [[2, 1, 0], [2, 0, 1]])
+
+    mol = Chem.MolFromSmiles("NCCF")
+    core = [1, 2]
+    groups_of_1, groups_of_2, groups_of_3 = identify_anchor_groups(mol, core, 1)
+    assert_anchor_group_equality(groups_of_1, [[1]])
+    assert_anchor_group_equality(groups_of_2, [[1, 2]])
+    assert_anchor_group_equality(groups_of_3, [])
+
+    mol = Chem.MolFromSmiles("C(C)(C)(C)C")
+    core = [0]
+    groups_of_1, groups_of_2, groups_of_3 = identify_anchor_groups(mol, core, 0)
+    assert_anchor_group_equality(groups_of_1, [[0]])
+    assert_anchor_group_equality(groups_of_2, [])
+    assert_anchor_group_equality(groups_of_3, [])
+
+    mol = Chem.MolFromSmiles("C(C)(F)CC")
+    core = [1, 0, 4, 3]
+    groups_of_1, groups_of_2, groups_of_3 = identify_anchor_groups(mol, core, 0)
+    assert_anchor_group_equality(groups_of_1, [[0]])
+    assert_anchor_group_equality(groups_of_2, [[0, 1], [0, 3]])
+    assert_anchor_group_equality(groups_of_3, [[0, 3, 4]])
+
+
+def test_enumerate_anchor_groups():
+    mol = Chem.MolFromSmiles("C1CC2OOC12")
+    core = [0, 1, 2, 5]
+    groups_of_1, groups_of_2, groups_of_3 = enumerate_anchor_groups(mol, core, [3, 4])
+    assert_anchor_group_equality(groups_of_1, [[2], [5]])
+    assert_anchor_group_equality(groups_of_2, [[2, 1], [2, 5], [5, 0], [5, 2]])
+    assert_anchor_group_equality(groups_of_3, [[2, 1, 0], [2, 5, 0], [5, 0, 1], [5, 2, 1]])
+
+    # aspirin
+    mol = Chem.MolFromSmiles("CC(=O)OC1=CC=CC=C1C(=O)O")
+    core = [3, 4, 5, 6, 7, 8, 9, 10]
+    groups_of_1, groups_of_2, groups_of_3 = enumerate_anchor_groups(mol, core, [0, 1, 2, 11, 12])
+    assert_anchor_group_equality(groups_of_1, [[3], [10]])
+    assert_anchor_group_equality(groups_of_2, [[3, 4], [10, 9]])
+    assert_anchor_group_equality(groups_of_3, [[3, 4, 5], [3, 4, 9], [10, 9, 8], [10, 9, 4]])
+    groups_of_1, groups_of_2, groups_of_3 = enumerate_anchor_groups(mol, core, [0, 1, 2])
+    assert_anchor_group_equality(groups_of_1, [[3]])
+    assert_anchor_group_equality(groups_of_2, [[3, 4]])
+    assert_anchor_group_equality(groups_of_3, [[3, 4, 5], [3, 4, 9]])
+    groups_of_1, groups_of_2, groups_of_3 = enumerate_anchor_groups(mol, core, [11, 12])
+    assert_anchor_group_equality(groups_of_1, [[10]])
+    assert_anchor_group_equality(groups_of_2, [[10, 9]])
+    assert_anchor_group_equality(groups_of_3, [[10, 9, 8], [10, 9, 4]])
+
+
+def test_enumerate_allowed_dummy_ixns_3_anchors():
+    dummy_group = {0, 1, 5, 8}
+    anchor_group = [2, 3, 4]
+
+    allowed_ixns = enumerate_dummy_ixns(dummy_group, anchor_group)
+
+    for ixn in allowed_ixns:
+        assert len(ixn) == len(set(ixn))
+
+    assert (0, 1, 2, 3) in allowed_ixns
+    assert (0, 1, 2) in allowed_ixns
+    assert (1, 2) in allowed_ixns
+    assert (0, 2, 3) in allowed_ixns
+    assert (1, 0, 2, 3) in allowed_ixns
+    assert (3, 2, 1, 5) in allowed_ixns
+    assert (1, 5, 2) in allowed_ixns
+    assert (2, 0, 5) in allowed_ixns
+    assert (1, 5, 8) in allowed_ixns
+    assert (1, 5, 8) in allowed_ixns
+    assert (0, 8, 5, 1) in allowed_ixns
+    assert (5, 1, 0, 8) in allowed_ixns
+    assert (3, 2, 8, 5) in allowed_ixns
+    assert (1, 8, 2, 3) in allowed_ixns
+
+    assert (0, 1, 3) not in allowed_ixns
+    assert (0, 1, 4) not in allowed_ixns
+    assert (0, 2, 4) not in allowed_ixns
+    assert (0, 3) not in allowed_ixns
+    assert (0, 4) not in allowed_ixns
+    assert (5, 3, 4) not in allowed_ixns
+    assert (1, 5, 4) not in allowed_ixns
+    assert (5, 8, 2, 4) not in allowed_ixns
+    assert (1, 8, 3, 4) not in allowed_ixns
+
+
+def test_enumerate_allowed_dummy_ixns_2_anchors():
+    dummy_group = {0, 1}
+    anchor_group = [4, 5]
+    allowed_ixns = enumerate_dummy_ixns(dummy_group, anchor_group)
+
+    for ixn in allowed_ixns:
+        assert len(ixn) == len(set(ixn))
+    assert len(allowed_ixns) == 16
+
+    assert (0, 5) not in allowed_ixns
+    assert (1, 5) not in allowed_ixns
+
+    # 3 ixns
+    assert (0, 1) in allowed_ixns
+    assert (0, 4) in allowed_ixns
+    assert (1, 4) in allowed_ixns
+
+    # 7 ixns
+    assert (0, 1, 4) in allowed_ixns
+    assert (1, 0, 4) in allowed_ixns
+    assert (0, 4, 1) in allowed_ixns
+    assert (0, 4, 5) in allowed_ixns
+    assert (1, 4, 5) in allowed_ixns
+    assert (0, 5, 4) in allowed_ixns
+    assert (1, 5, 4) in allowed_ixns
+
+    # 6 ixns
+    assert (0, 1, 4, 5) in allowed_ixns
+    assert (1, 0, 4, 5) in allowed_ixns
+    assert (1, 0, 5, 4) in allowed_ixns
+    assert (0, 4, 5, 1) in allowed_ixns
+    assert (0, 5, 4, 1) in allowed_ixns
+    assert (0, 1, 5, 4) in allowed_ixns
+
+
+def test_enumerate_allowed_dummy_ixns_1_anchor():
+    dummy_group = {0, 1}
+    anchor_group = [4]
+    allowed_ixns = enumerate_dummy_ixns(dummy_group, anchor_group)
+    assert len(allowed_ixns) == 6
+    assert (0, 1) in allowed_ixns
+    assert (0, 4) in allowed_ixns
+    assert (1, 4) in allowed_ixns
+    assert (0, 1, 4) in allowed_ixns
+    assert (1, 0, 4) in allowed_ixns
+    assert (0, 4, 1) in allowed_ixns
+
+
+def test_flag_bonds():
+    mol = Chem.MolFromSmiles("BrOC1=CC(F)=CC=N1")
+    core = [2, 3, 4, 6, 7, 8]
+    bond_pairs = [
+        (1, [0, 1, 2]),  # H-O-C
+        (1, [1, 2, 3]),  # O-C-C
+        (1, [1, 2, 3, 4]),  # O-C-C-C
+        (0, [1, 2, 7]),  # O-C-N
+        (1, [2, 3, 4]),  # C-C-C
+        (1, [5, 4, 3, 2]),  # F-C-C-C
+        (1, [5, 4, 3]),  # F-C-C
+        (0, [5, 4, 6]),  # F-C-C
+    ]
+
+    expected_flags = [x[0] for x in bond_pairs]
+    bond_idxs = [x[1] for x in bond_pairs]
+
+    keep_flags = flag_bonds(mol, core, bond_idxs)
+    assert tuple(keep_flags) == tuple(expected_flags)
+
+    # flipping the ordering should give us identical results
+    bond_idxs = [x[::-1] for x in bond_idxs]
+    keep_flags = flag_bonds(mol, core, bond_idxs)
+    assert tuple(keep_flags) == tuple(expected_flags)
+
+
+def test_flag_bonds_core_hop():
+
+    mol = Chem.MolFromSmiles("FC1CO1")
+
+    #    F0       F
+    #    |        .
+    #   _C1  ->  .C
+    # 3O |      O |
+    #   \C2       C
+    core = [1, 2]
+    bond_pairs = [
+        (1, [0, 1]),
+        (1, [1, 2]),
+        (0, [2, 3]),
+        (1, [3, 1]),
+        (1, [0, 1, 2]),
+        (1, [0, 1, 3]),
+        (1, [3, 1, 2]),
+        (0, [1, 3, 2]),
+        (1, [1, 2, 3]),  # this last one is technically allowed
+    ]
+
+    # TBD: we can prune this to only terms where the bonds actually exist between ijk, ijkl, etc.
+
+    expected_flags = [x[0] for x in bond_pairs]
+    bond_idxs = [x[1] for x in bond_pairs]
+
+    keep_flags = flag_bonds(mol, core, bond_idxs)
+
+    assert tuple(keep_flags) == tuple(expected_flags)
+
+
+def _draw_impl(mol, core, ff, fname):
+
+    hb_params, hb_idxs = ff.hb_handle.parameterize(mol)
+    ha_params, ha_idxs = ff.ha_handle.parameterize(mol)
+    pt_params, pt_idxs = ff.pt_handle.parameterize(mol)
+    it_params, it_idxs = ff.it_handle.parameterize(mol)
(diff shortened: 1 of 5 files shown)


timemachine/fe/dummy.py
```diff
@@ -0,0 +1,396 @@
+import numpy as np
+import networkx as nx
+import copy
+import itertools
+
+
+def _add_successors(mol, core, groups):
```
maxentile, line 7: Observation: The definitions in this file seem purely graph-based, but the signatures are in terms of full-fledged RDKit mols.

Question: Would it be preferable if these accepted networkx graphs instead?

------

(Primary consideration: Unsure if there are important logical differences in the graph traversal procedures here vs. [in networkx](https://networkx.org/documentation/stable/reference/algorithms/traversal.html#module-networkx.algorithms.traversal.breadth_first_search), aside from the adaptations to use RDKit molecules in the signatures.

Secondary consideration: Implementations might be marginally cleaner if iteration over nodes, edges, neighborhoods, etc. don't require `GetAtomWithIdx()`, `GetIdx()`, `GetBeginAtomIdx()` etc.)

timemachine/fe/dummy.py
```diff
@@ -0,0 +1,396 @@
...
+        groups will be [[anchor,x,y,], ...]
+
+    """
+
+    assert anchor in core
+
+    # We perform a bfs of depth 3 starting from the anchor.
+    layer_1_groups = [[anchor]]
+    layer_2_groups = _add_successors(mol, core, layer_1_groups)
+    layer_3_groups = _add_successors(mol, core, layer_2_groups)
```
maxentile, line 56: Clarification: The reason this stops at layer_3 is because each term in our valence forcefield only "talks to" a group of atoms up to 3 bonds away, correct?

------

Condition where this could be invalidated in the future: if the force field ever defines improper torsions on sets of 4 atoms (a, b, c, d) where shortest_path_length(a, d) > 3.

Defensive request: If the depth=3 here relies on this sort of assumption, could we add a function to check that a bond table + set of PeriodicTorsion indices are consistent with the assumption?

maxentile (commented):
Mostly superficial comments so far -- more substantive review hopefully tomorrow!

Question: Are intramolecular nonbonded interactions relevant for this PR?

proteneer:
> Question: Are intramolecular nonbonded interactions relevant for this PR?

Intramolecular short-range nonbonded interactions (and their exclusions/exceptions) can be treated as any other 1-2 interaction as a function of distance, so as long as they're properly turned off at the end-states they should work as intendend. However, extreme care must be taken at the intermediate lambda windows (eg. with softcores) to avoid singularities.

tests/test_dummy.py
```diff
@@ -0,0 +1,240 @@
...
+    assert (0, 1) in allowed_ixns
+    assert (1, 0) in allowed_ixns
+    assert (0, 4) in allowed_ixns
+    assert (1, 4) in allowed_ixns
+    assert (0, 1, 4) in allowed_ixns
+    assert (1, 0, 4) in allowed_ixns
+    assert (0, 4, 5) in allowed_ixns
+    assert (1, 4, 5) in allowed_ixns
+    assert (0, 1, 4, 5) in allowed_ixns
+    assert (1, 0, 4, 5) in allowed_ixns
```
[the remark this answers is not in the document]
proteneer: > Style question: Is there a reason to prefer this way of specifying a set rather than something like

Mainly for consistency with the `not in` statements with the 3 anchor test - which would have to resort to something like `ref.intersection(b) == set()` or something. I can change this though if its too hard to read.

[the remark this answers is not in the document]
proteneer: > Documentation request: At minimum, a brief comment at the top of this test file about how the reference values were produced. Ideally, also a one-sentence description in each test of what is being asserted.
Will add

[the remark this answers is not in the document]
proteneer: > (Some of these catch my eye because only one of the patterns is included both forward and reverse ((0,1) and (1,0)), not sure if that's noteworthy...)

I should probably uniquify the bonds in the set based on our typical lexical ordering of the atom idxs

timemachine/fe/dummy.py
proteneer: I debated about this too - I wrote this initially in rdkit simply because

1) I'm extremely familiar with the traversal/query logic, though there's no real barrier to adapting this to networkx

2) There was another version of the code from before that predicated on atom/bond properties ("eg. heavy atom or not" , "is SP3 hybridized", "single or double bond" etc.). Although the code currently doesn't predicate on more tricky logic, there may be instances where we need to avoid certain types of torsions defined over core anchors such as nitriles (where the collinear bonds would vanish the cross product).

timemachine/fe/dummy.py
proteneer: > Clarification: The reason this stops at layer_3 is because each term in our valence forcefield only "talks to" a group of atoms up to 3 bonds away, correct?

Pretty much - with the extra condition that 1-2 interactions must be defined over bonds, 1-3 over angles, and 1-4 over torsions. 

> Condition where this could be invalidated in the future: if the force field ever defines improper torsions on sets of 4 atoms (a, b, c, d) where shortest_path_length(a, d) > 3.

Impropers are technically fine from a correctness point of view. Underneath the hood they're just implemented as proper torsions of 4 indices. The difference lies in which bonds/angles are involved. A typical ProperTorsion has 1 torsion, 2 angle, and 3 bonded terms. An Improper Torsion has 1 torsion, 1 angle, and 2 bonded terms. We may run the risk of including a torsion without all the relevant bond/angle terms, but from a technically correct standpoint you can probably still factorize the resulting partition function (eg. in 2D you can define only an angle restraint without adding the bond restraints, the partition function still factorizes, but it just may be very hard to converge since the dummy atom will be drifting away). 
 
One good check we could assert (not here but at a later stage) is that every 1-3 term has both 1-2 terms turned on, and every 1-4 term has both 1-3 terms and all three 1-2 terms turned on.

timemachine/fe/dummy.py
```diff
@@ -0,0 +1,396 @@
...
+    return anchors
+
+
+def convert_to_nx(mol):
+    # assume mol is fully connected.
+    g = nx.Graph()
+    for bond in mol.GetBonds():
+        src, dst = bond.GetBeginAtomIdx(), bond.GetEndAtomIdx()
+        g.add_edge(src, dst)
+    return g
```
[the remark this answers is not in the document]
proteneer: willfix!

timemachine/fe/dummy.py
mcwitt: Just wanted to +1 the idea of separating the graph traversal logic from the details of the RDKit data structure and potentially using `bfs_successors`

mcwitt (commented):
Left some minor comments; planning to take another pass tomorrow!

maxentile (commented):
Thanks for the clarifications!

A few more comments, but I'll need to study `enumerate_dummy_ixns` and `flag_bonds` a bit more carefully still.

timemachine/fe/dummy.py
```diff
@@ -0,0 +1,618 @@
...
+            if nbr in core and nbr not in new_group:
+                new_group.append(nbr)
+                next_groups.append(new_group)
+
+    return next_groups
+
+
+def identify_anchor_groups(mol, core, anchor):
+    """
+    Generate putative choices for valid anchor groups. An anchor group
```
badisa, line 29: Do you mean that the choices are putatively valid? Wasn't familiar with the word, but `putative choices` doesn't quite make sense to me

pitmanme:
In fe/dummy.py

def flag_bonds(mol, core, bond_idxs)

Why are you maximizing allowed dummy to anchor interactions? I will read over this function more later, but I think you would instead what to maximize the number of dummy to anchor interaction you turn off.

proteneer:
@pitmanme 

You generally want to minimize the number of interactions that need to be turned off. One can trivially turn off every dummy-anchor interaction except a 1-2 bond interaction, and it would be correct. But you enable the dummy region to sample larger regions of phase space as a result. I've also attached two pictures detailing which types of dummy-anchor interactions are allowed, and which ones are not.


Untitled Notebook (1)-8 2
Untitled Notebook (1)-7 2

timemachine/fe/dummy.py
```diff
@@ -0,0 +1,396 @@
...
+        dense_graph[j, i] = 1
+
+    # sparsify to simplify and speed up traversal code
+    sparse_graph = []
+    for row in dense_graph:
+        nbs = []
+        for col_idx, col in enumerate(row):
+            if col == 1:
+                nbs.append(col_idx)
+        sparse_graph.append(nbs)
```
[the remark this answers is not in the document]
proteneer: > could we avoid constructing the dense matrix here using something like the following?

I didn't want to worry about iteration order within dictionaries later on during traversal. So just went with the simplest solution here.

> Side note: networkx graphs support conversion to adjacency matrix

Also really didn't have time to read into the bazillion formats out there for sparse matrices (COO, CSR, CRC etc have very different storage layout and performance behaviors, not that performance really matters here).

tests/test_dummy.py
```diff
@@ -0,0 +1,309 @@
...
+    assert set(anchors) == set()
+    anchors = identify_anchors(mol, core, dummy_atom=2)
+    assert set(anchors) == set()
+
+    mol = Chem.MolFromSmiles("C1CC2NC2C1")
+    core = [0, 1, 2, 5]
+    anchors = identify_anchors(mol, core, dummy_atom=3)
+    assert set(anchors) == set([2, 5])
+    anchors = identify_anchors(mol, core, dummy_atom=4)
+    assert set(anchors) == set([2, 5])
```
badisa, line 49: Should this test have a case where the anchor is made up of 3 atoms?

tests/test_dummy.py
```diff
@@ -0,0 +1,309 @@
...
+    groups_of_1, groups_of_2, groups_of_3 = enumerate_anchor_groups(mol, core, [3, 4])
+    assert_anchor_group_equality(groups_of_1, [[2], [5]])
+    assert_anchor_group_equality(groups_of_2, [[2, 1], [2, 5], [5, 0], [5, 2]])
+    assert_anchor_group_equality(groups_of_3, [[2, 1, 0], [2, 5, 0], [5, 0, 1], [5, 2, 1]])
+
+    # aspirin
+    mol = Chem.MolFromSmiles("CC(=O)OC1=CC=CC=C1C(=O)O")
+    core = [3, 4, 5, 6, 7, 8, 9, 10]
+    groups_of_1, groups_of_2, groups_of_3 = enumerate_anchor_groups(mol, core, [0, 1, 2, 11, 12])
+    assert_anchor_group_equality(groups_of_1, [[3], [10]])
```
badisa, line 140: Should this validate that groups_of_2 and groups_of_3 are empty sets?

tests/test_dummy.py
```diff
@@ -0,0 +1,309 @@
...
+def test_enumerate_allowed_dummy_ixns_3_anchors():
+    dummy_group = {0, 1, 5, 8}
+    anchor_group = [2, 3, 4]
+
+    allowed_ixns = enumerate_dummy_ixns(dummy_group, anchor_group)
+
+    for ixn in allowed_ixns:
+        assert len(ixn) == len(set(ixn))
+
+    assert (0, 1, 2, 3) in allowed_ixns
```
badisa, line 160: Is this the complete set of ixns and if so would be good to assert of the number of allowed ixns?

proteneer:
Some additional features, the test case `test_parameterize_and_draw_interactions` shows how it all ties together.

res_1
res_0

proteneer:
Improved support for ring opening closing:

Before:
ring_open_0

New:
ring_open_1
ring_open_0

timemachine/fe/dummy.py
proteneer: This turned out to be doable, but more complicated than I thought, mainly because of the need to deal with avoiding back-edges (since edges are undirected, a successor could actually point back to a predecessor if I'm not super careful). I'd still be happy to convert this to networkx but I'd rather save that for another PR (since there are probably other parts of this code that could benefit from the switch to networkx).

timemachine/fe/dummy.py
proteneer: `strict` mode will now check that that every 1-3 term has both 1-2 terms turned on, and every 1-4 term has both 1-3 terms and all three 1-2 terms turned on.

timemachine/fe/dummy.py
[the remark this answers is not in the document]
proteneer: Deferring this to another PR by cleaning up with networkx - for now, list of lists is much easier for me to reasonabout than iterating though kvs

timemachine/fe/dummy.py
```diff
@@ -0,0 +1,396 @@
...
+
+    def dfs(i, visited):
+        if i in visited:
+            return
+        else:
+            visited.add(i)
+            if i not in core:
```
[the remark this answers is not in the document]
proteneer: probably - but then again, dfs_traversal in networkx isn't super straightforward to adapt either. I'd rather do all of these conversions in a separate PR.

timemachine/fe/dummy.py
```diff
@@ -0,0 +1,618 @@
...
+                    allowed_ixns.add((i, j, a, b))
+                    # this looks weird at first, but is actually valid
+                    # consider:
+                    # i j
+                    # |/
+                    # a-b
+                    # we can freely rotate i-j independently of a-b
+                    # in both of the following cases.
+                    allowed_ixns.add((i, a, b, j))
+                    allowed_ixns.add((a, b, i, j))
```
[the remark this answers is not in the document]
proteneer: lmk if this is still not clear - trying to resolve all outstanding issues

tests/test_dummy.py
proteneer: This was confusing to the reader, I've renamed `identify_anchors` to `identify_root_anchors`, defined as the "starting" anchor in an anchor group. A case of a dummy atom connected 3 rooted anchors is extremely rare, but I'll add it for shits an giggles.

proteneer: added this:

```
    mol = Chem.MolFromSmiles("C1C2CC3CC1N23")
    core = [0, 1, 2, 3, 4, 5]
    anchors = identify_root_anchors(mol, core, dummy_atom=6)
    assert set(anchors) == set([1, 3, 5])
```

tests/test_dummy.py
proteneer: They are not empty sets. I'll add the members more explicitly

tests/test_dummy.py
proteneer: This is not the complete set of ixns, just a spot check. Enumerating the full set of `allowed_ixns` requires more skills in combinatorics than I currently possess.

proteneer:
I'm going to simplify the PR and remove the parts that deal with numerical stability (and save that for a separate PR).

proteneer:
Rebased onto master!

timemachine/fe/dummy.py
```diff
@@ -0,0 +1,631 @@
...
+    def dfs(i, visited):
+        if i in visited:
+            return
+        else:
+            visited.add(i)
+            if i not in core:
+                for nb in sparse_graph[i]:
+                    dfs(nb, visited)
+            else:
+                return
```
mcwitt, line 128: nit: no-op

timemachine/fe/dummy.py
```diff
@@ -0,0 +1,631 @@
...
+
+    visited = set()
+
+    dfs(dummy_atom, visited)
+    anchors = []
+
+    for a_idx in visited:
+        if a_idx in core:
+            anchors.append(a_idx)
```
mcwitt, line 137: nit: cleaner as list comprehension:
```suggestion
    anchors = [a_idx for a_idx in visited if a_idx in core]
```

timemachine/fe/dummy.py
```diff
@@ -0,0 +1,631 @@
...
+
+    return anchors
+
+
+def enumerate_anchor_groups(mol, core, dummy_group):
+    """
+    An anchor group is a set of core atoms that are allowed to interact with the dummy
+    atoms in a dummy group in a way that allows the partition function to be separated.
+    Unlike dummy groups, anchor groups do not partition the non-core atoms. They're allowed
+    to overlap core atom indices. In other words, anchor groups for any particular dummy
```
mcwitt, line 147: Finding this comment a little bit unclear on the first pass -- it sounds like anchor groups can include core atoms? and also, anchor groups for different dummy groups can overlap with one another?

timemachine/fe/dummy.py
```diff
@@ -0,0 +1,631 @@
...
+
+    While the choices of dummy groups is arbitrary, this function partitions the dummy
+    atoms into multiple dummy groups using a heuristic:
+
+    1) Identify all 1-2, 1-3 edges between dummy atoms.
+    2) Generate an induced graph using edges from above set.
+    3) Disconnected components of the resulting induced graph are our dummy groups.
+
+    We limit ourselves to only 1-2 and 1-3 interactions in 1) because we're typically
+    guaranteed that there exists a direct bond or that they're at most one bond away.
```
mcwitt, line 213: It's not clear to me when this guarantee wouldn't apply. Worth adding a comment if there are notable exceptions?

timemachine/fe/dummy.py
```diff
@@ -0,0 +1,631 @@
...
+
+    For example:
+
+    ```
+    mol = Chem.MolFromSmiles("FC1CC1(F)N")
+    core = [1, 2, 3]
+    dg = identify_dummy_groups(mol, core)
+    assert_set_equality(dg, [{0}, {4, 5}])
+    ```
+
```
mcwitt, line 223: Should parameters be documented here?

timemachine/fe/dummy.py
```diff
@@ -0,0 +1,631 @@
...
+            # get distance to anchor
+            dists = []
+            for a in anchors:
+                dists.append(nx.shortest_path_length(g, source=dummy, target=a))
+            anchor_membership.append(np.argmin(dists))
+
+        anchor_kv = {}
+        for idx, anchor in enumerate(anchor_membership):
+            if anchor not in anchor_kv:
+                anchor_kv[anchor] = set()
```
mcwitt, line 277: nit: simplify with `defaultdict(set)`?

badisa (approved):

timemachine/fe/dummy.py
```diff
@@ -0,0 +1,631 @@
...
+        return tuple(ixn[::-1])
+    else:
+        return tuple(ixn)
+
+
+def make_bond_set(old_set):
+    """
+    Bond set has the requirement that indices are symmetrically
+    reversible. i.e. ij = ji, ijk = kji, ijkl = lkji
+    """
```
mcwitt, line 438: nit: add summary of what this does? (I think something like "standardize atom ordering in bonds"?)

timemachine/fe/dummy.py
```diff
@@ -0,0 +1,631 @@
...
+    for agcs, agis in zip(all_agcs, all_agis):
+
+        counts = []
+        for mutual_bonds in agis:
+            bond_count = np.sum([len(idxs) == 2 for idxs in mutual_bonds])
+            angle_count = np.sum([len(idxs) == 3 for idxs in mutual_bonds])
+            torsion_count = np.sum([len(idxs) == 4 for idxs in mutual_bonds])
+            counts.append((bond_count, angle_count, torsion_count))
+
+        best_idx = sorted(zip(range(len(counts)), counts), reverse=True, key=lambda x: x[1])[0][0]
```
mcwitt, line 627: nit: could use `min` here?
```suggestion
        best_idx = min(zip(range(len(counts)), counts), reverse=True, key=lambda x: x[1])[0]
```

tests/test_dummy.py
```diff
@@ -0,0 +1,543 @@
...
+
+    assert (0, 1, 3) not in allowed_ixns
+    assert (0, 1, 4) not in allowed_ixns
+    assert (0, 2, 4) not in allowed_ixns
+    assert (0, 3) not in allowed_ixns
+    assert (0, 4) not in allowed_ixns
+    assert (5, 3, 4) not in allowed_ixns
+    assert (1, 5, 4) not in allowed_ixns
+    assert (5, 8, 2, 4) not in allowed_ixns
+    assert (1, 8, 3, 4) not in allowed_ixns
```
mcwitt, line 194: is it worth leaving a line comment for some of the trickier cases explaining why they're disallowed? A lot of these aren't obvious to me at this point

timemachine/fe/dummy.py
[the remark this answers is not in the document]
mcwitt: > To express this in code rather than comments, we might need to add more geometric / symmetry-flavored types than tuples

Just wanted to +1 this idea (probably for future PR) -- seems like it could help with readability to give names to the different interaction tuples and encode the equivalence logic close to the type definition

timemachine/fe/dummy.py
```diff
@@ -0,0 +1,631 @@
...
+
+    core: list of int
+        idxs of core atoms
+
+    bond_idxs: list of list of int
+        list of 2-tuple, 3-tuple, 4-tuple
+
+    Returns
+    -------
+        boolean flags, 1: keep, 0: remove
```
mcwitt, line 463: nit: is this always a list of the same length as `bond_idxs`?

mcwitt (approved):
Looks good -- I left a few requests for documentation, and some minor stylistic nits (I'm fine with deferring the latter to the upcoming cleanup PR)

maxentile (commented):
Still reviewing! The new plotting utilities, examples, and glossary are very helpful.

------

Testing comment: There are over a hundred assertions in `test_dummy.py`, and it is not immediately obvious to me where all of these come from. Compressing some of these assertions would improve their readability and maintainability:

* For example, if the lists of `(i, j, k, l) not in allowed_ixns` assertions are obtained by repeated application of rules that determine "this ixn is definitely not allowed", is it feasible to add functions to compute these rules? (It is not immediately obvious if the `not in` assertions might come from applying a single rule, several rules, or something else.)
* Some of the tests exactly characterize the output (e.g. [`test_enumerate_allowed_dummy_ixns_2_anchors`](https://github.com/proteneer/timemachine/blob/6c457adb354d00d12026df1c67418e6334cb443e/tests/test_dummy.py#L197-L229) checks that the output set exactly matches a reference set). If these reference sets are produced by a reference implementation, could we also include the reference implementation?

In general, greater compression here would make it easier to answer questions like: "If these functions are ever refactored, what properties should we intend to check in the refactored tests?"

Related: The tests also contain many neat test molecule definitions -- however, I do not know what corner cases are intended to be exercised by each molecule. (For some the intent looks obvious, e.g. https://github.com/proteneer/timemachine/blob/6c457adb354d00d12026df1c67418e6334cb443e/tests/test_dummy.py#L38 exercises the case of disconnected compounds, but in many cases I cannot easily infer what behavior the molecule is designed to test, that is not exercised by other molecules.) It would be ideal to document their intent while fresh in mind.

--------

Optional request: A usage example where a pair of molecules is input, and an alchemical topology is output.

Currently the definitions are in terms of a single molecule + core + ..., but the intended use case would involve pairs of molecules. It's not immediately clear if the conditions checked here are sufficient also in the case of molecule pairs, or if more conditions are needed.

timemachine/fe/dummy.py
```diff
@@ -0,0 +1,631 @@
...
+    is an ordered sequence of core atoms starting with the anchor (a-b-c) that
+    are connected by bonds. In other words, an anchor group is a rooted subtree
+    with three nodes spanning from the anchor.
+
+    Parameters
+    ----------
+    mol: Chem.Mol
+        rdkit molecule
+
+    core: list or set or iterable
```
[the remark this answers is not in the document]
proteneer: of int, added.

timemachine/fe/dummy.py
proteneer: d'oh thanks.

timemachine/fe/dummy.py
proteneer: thanks, i still suck at list comprehensions

proteneer: I'm going to just add these in manually as opposed to accepting the commit suggestion to avoid having to deal with conflicts since the code has updated locally on my end.

timemachine/fe/dummy.py
proteneer: 1) An anchor group is comprised exclusively of core atoms.
2) Yes, anchor groups for different dummy groups can overlap with one another.

proteneer: Made it slightly more clear in comments.

timemachine/fe/dummy.py
proteneer: I can't think of any counter examples actually.

proteneer: I think the message I was trying to convey was that 1-4 interactions don't have this property - if we allowed 1-4s as well we may miss bonds.

eg., let D be dummy, C be core:

```
D0     D1
  \   /
   C-C
```
Allowing `D0, D1` to be in the same group would be bad, since only one dummy-bond would be allowed (either D0-C or D1-C)

proteneer: Though on second thought, the extra refining steps should break these up nevertheless.

proteneer: I'm just going to remove this comment entirely since it seems to create more confusion

timemachine/fe/dummy.py
proteneer: Trying to not get too verbose for code examples - the documentation should be pretty clear here.

timemachine/fe/dummy.py
proteneer: will add

timemachine/fe/dummy.py
proteneer: will fix

timemachine/fe/dummy.py
proteneer: I think this should be `max` since `min` doesn't support `reverse=True` - will replace though.

tests/test_dummy.py
proteneer: I need to document these one by one... will get back to them

timemachine/fe/dummy.py
proteneer: We now only support strict mode - so the first comment is less relevant.

proteneer:
> Optional request: A usage example where a pair of molecules is input, and an alchemical topology is output.

This will be taken care of in a separate PR - there are a couple of other corner cases to address (esp when core-core interactions may be different at the end-states).

proteneer:
> For example, if the lists of (i, j, k, l) not in allowed_ixns assertions are obtained by repeated application of rules that determine "this ixn is definitely not allowed", is it feasible to add functions to compute these rules? (It is not immediately obvious if the not in assertions might come from applying a single rule, several rules, or something else.)

Not so simple - the code works by enumerating *all allowed* interactions, the set of *all unallowed interactions* is probably too big to enumerate. 

> Some of the tests exactly characterize the output (e.g. [test_enumerate_allowed_dummy_ixns_2_anchors](https://github.com/proteneer/timemachine/blob/6c457adb354d00d12026df1c67418e6334cb443e/tests/test_dummy.py#L197-L229) checks that the output set exactly matches a reference set). If these reference sets are produced by a reference implementation, could we also include the reference implementation?

These are generated by hand. I specifically didn't want to introduce another program to generate these because of possibility of bugs in writing to generate the disallowed parts.

tests/test_dummy.py
```diff
@@ -0,0 +1,543 @@
...
+        [0, 1],
+        [1, 2],
+        [1, 3],
+        [0, 2],
+        [0, 1, 2],
+        [0, 1, 3],
+        [2, 1, 3],
+        [0, 2, 1],
+        [0, 2, 1, 3],
+    ]
```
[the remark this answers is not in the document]
proteneer: think of these as the concatenation of:

`harmonic_bond.get_idxs()` + `harmonic_angle.get_idxs()` + `periodic_torsion.get_idxs()`.

timemachine/fe/dummy.py
proteneer: After discussing with @mcwitt we think its best to defer the nx conversion stuff to another PR. It will likely just trade one set of readability concerns for another unless we do it carefully.

timemachine/fe/dummy.py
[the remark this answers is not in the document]
proteneer: This has been simplified a bit since we no longer allow non-strict stuff. The important thing to remember is 1-2, 1-3, 1-4 take in ordered indices but I should be clearer about that in the comments.

tests/test_dummy.py
```diff
@@ -0,0 +1,358 @@
...
+    flag_bonds,
+    identify_anchor_groups,
+    identify_dummy_groups,
+    identify_root_anchors,
+)
+from timemachine.ff import Forcefield
+from timemachine.ff.handlers.deserialize import deserialize_handlers
+
+
+def test_identify_root_anchors():
```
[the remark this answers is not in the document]
proteneer: Added

maxentile:
>>For example, if the lists of (i, j, k, l) not in allowed_ixns assertions are obtained by repeated application of rules that determine "this ixn is definitely not allowed", is it feasible to add functions to compute these rules? (It is not immediately obvious if the not in assertions might come from applying a single rule, several rules, or something else.)

>Not so simple - the code works by enumerating all allowed interactions, the set of all unallowed interactions is probably too big to enumerate.

Sorry, didn't mean to suggest _enumerating_ all disallowed interactions. However, assertions like https://github.com/proteneer/timemachine/blob/337d8d1b145b23432093f04db8bd63fbcb551e66/tests/test_dummy.py#L250-L251 currently do not indicate what property would have been invalidated if `(0,3)` or `(0,4)` were found in `allowed_ixns`.

One way to improve this would be to keep the tests written as-is, and just add annotations to indicate why a block of interactions is disallowed. Preferably though, the test size and documentation burden might be reduced if some of the literal assertions could be replaced with generic computable conditions. For example, in testing `identify_dummy_groups`, the constraints listed in this docstring seem like they should correspond to unambiguous, computable functions, `is_constraint_1_satisfied(mol, core, dummy_groups)`, etc...

https://github.com/proteneer/timemachine/blob/337d8d1b145b23432093f04db8bd63fbcb551e66/timemachine/fe/dummy.py#L195-L199  

In a follow-up PR, we might want to generalize https://github.com/proteneer/timemachine/blob/337d8d1b145b23432093f04db8bd63fbcb551e66/tests/test_dummy.py#L80-L136 to assert these 3 constraints are satisfied, which should still be valid to assert even if the details of the heuristic change.

(It's less clear to me what properties can be asserted about the output of `enumerate_dummy_ixns`.)

maxentile:
Ahh, my concerns in https://github.com/proteneer/timemachine/pull/633#issuecomment-1048335806 were mostly addressed in https://github.com/proteneer/timemachine/pull/633/commits/0893b9345b0de5ad7b99a1402145440ebf597a38 , was commenting on an outdated version.

maxentile (approved):

proteneer:
Thanks for the massive amount of comments and feedback. Merging this in and preparing for the next set of changes!

proteneer:
(And special shoutout to @maxentile for helping me identify the incorrectness of non-strictness, allowing me to simplify the PR a great deal)

PR #638, opened 2022-02-15
Task/#636 add seeds to rabfe models

### LINKED ISSUE #636: Allow providing Seeds to RABFE Models
Need a means in which to pass seeds to RABFE models to get reproducible simulations.

badisa:
* Resolves #636 
* Adds functionality for loading forcefields with less code, because I got annoyed with copy pasting
* Speeds up the slow RABFE tests by making them solvent simulations'

Recommended to read by commit since the refactor code is super repetitive. Though might be desirable to remove the changes to some of the examples, as only tested the validate_relative_binding.py example


maxentile (approved):
Looks good!

slow_tests/test_rbfe.py
```diff
@@ -53,7 +51,6 @@ def test_predict(self):
 
         ddg, results = model.predict(ordered_params, mol_a, mol_b, core, seed=2022)
         self.assertEqual(len(results), 2)
-        np.testing.assert_equal(ddg, 12.32189679129123)
```
badisa: This value is significantly different on CI, about -0.5. Not sure about leaving it out, but wanted to avoid more if statements regarding os for these sorts of checks

maxentile: Interesting that this is an issue for test_rbfe but not test_rabfe... Taking a look.

Do we initially have a suggestion whether this is due to a random state issue specific to rbfe / this test, or something more generic different about CI environment (package version differences, OS differences)?

maxentile: Double-checking: equil_cache is freshly generated using the same random seed in both in local testing and CI?

https://github.com/proteneer/timemachine/blob/01b9cf14fc8f3d40777c3f6a99df3fd1d8721b8a/timemachine/fe/model.py#L152-L156

badisa: For this test there is no equil cache being generated and the other test that runs stores the cache in a temporary directory to avoid impacting other tests. 

My initial thought is that the difference comes from OS differences, but I had thought that was only for the AM1 charges, which this shouldn't be impacted by using the simple charge forcefield. Will see if I can't determine whats happening here, as I would prefer this to have deterministic value/assertion

maxentile: >For this test there is no equil cache being generated

Ahh, right! I had misremembered the default is  https://github.com/proteneer/timemachine/blob/01b9cf14fc8f3d40777c3f6a99df3fd1d8721b8a/timemachine/fe/model.py#L35

>My initial thought is that the difference comes from OS differences, but I had thought that was only for the AM1 charges, which this shouldn't be impacted by using the simple charge forcefield. 

Agreed.

>Will see if I can't determine whats happening here, as I would prefer this to have deterministic value/assertion

Perhaps a fallback: `assert ddg in allowed_values`, `allowed_values = [12.32189679129123, 11.78817841291265]`, with a note about the unexplained local / CI difference.

(So it still passes locally and in CI, but would still fail if `ddg(seed, os_details)` is not a deterministic function or if relevant `os_details` change...)

badisa: I decided to remove the changes specific to RBFE as it isn't immediately important to seed and I would like to dig into the source of this difference.

> Perhaps a fallback: assert ddg in allowed_values, allowed_values = [12.32189679129123, 11.78817841291265], with a note about the unexplained local / CI difference.

I do like this idea, and can put the changes back if you think it is important to change RABFE and RBFE models at the same time.

maxentile: >I do like this idea, and can put the changes back if you think it is important to change RABFE and RBFE models at the same time.

I don't see a reason to change the RABFE and RBFE models at the same time.

timemachine/ff/__init__.py
```diff
@@ -1,11 +1,44 @@
...
+        Forcefield
+            Return a ForceField object constructed from parameters file
+        """
+        original_path = str(path)
+        path = Path(path)  # Safe to construct a Path object from another Path object
+        if not path.is_file():
+            # Search built in params for the forcefields
+            path = PARAMS_DIR / original_path
+        if not path.is_file():
+            raise ValueError(f"Unable to find {original_path} in file system or built in forcefields")
```
mcwitt, line 37: nit: from an outside perspective this behavior (trying the file system first, then falling back to the built-in path) seems potentially a little surprising. E.g. is it possible to have a local file shadow a built-in, so it becomes tricky to figure out which version is being used?

Not a blocker, but wondering if this could be a little more explicit (e.g. having both `from_file` and `from_builtin` methods that share most of their implementation)

mcwitt (approved):
looks good, love the deduplication of the ff loading code! Just had 2 minor non-blocking suggestions.

examples/validate_relative_binding.py
```diff
@@ -195,14 +194,16 @@ def fn(*args, **kwargs):
         client = GRPCClient(hosts=cmd_args.hosts)
     client.verify()
 
+    seed = cmd_args.seed
+    if seed == 0:
```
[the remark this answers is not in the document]
badisa: This seems like a pretty common convention in our code and I feel I have seen it other places. But can make it None

timemachine/ff/__init__.py
badisa: Modified it in [edb97f9](https://github.com/proteneer/timemachine/pull/638/commits/edb97f939a10688fb3c4e2e2dc45fa2512ff200a) to print a warning and to fall back to the built in, as that is more likely to be trustworthy. Could also raise an exception, but in the moment I felt like just having a warning

examples/validate_relative_binding.py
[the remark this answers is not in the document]
maxentile: If we're changing this convention I prefer the type-based "`seed: Optional[int]`" over the flag-based "`seed: int`, see docstring for flag value".

The convention here may be intentional. (The C++ / CUDA implementations use `seed == 0` as a flag to randomize the seed -- it may be valuable to have reminders of this also in the Python layer.)

Anyway, may be best to defer to another Issue / PR.

>0 for the seed will result in a random seed (admittedly that's probably not a common value to pass)

Hah, personally, I had used `np.random.seed(0)` more commonly than any other value, until switching to the `np.random.seed(2022)` convention.

proteneer (approved):
nice! like the qol improvement in ff loading

PR #640, opened 2022-02-16
Cache all Openeye results for AM1 Charging

badisa:
* Helpful if you don't have an oe license in the simulation environment

### DIFF
--- a/tests/test_handlers.py
+++ b/tests/test_handlers.py
@@ -586,7 +586,7 @@ def test_compute_or_load_am1_charges():
     returned on repeated calls"""
 
     # get some molecules
-    cache_key = "AM1Cache"
+    cache_key = nonbonded.AM1_CHARGE_CACHE
     suppl = Chem.SDMolSupplier("tests/data/ligands_40.sdf", removeHs=False)
     all_mols = [mol for mol in suppl]
     all_mols = all_mols[:5]  # truncate so that whole test is ~ 10 seconds
@@ -608,11 +608,15 @@ def test_compute_or_load_am1_charges():
         np.testing.assert_array_equal(fresh, cached)
 
 
-def test_bond_smirks_matches():
+def test_compute_or_load_bond_smirks_matches():
     """Loop over test ligands, asserting that
+    * verify no cache key
     * returned indices are in bounds
-    * returned bonds are present in the mol"""
+    * returned bonds are present in the mol
+    * verify a cache key
+    * verify new values match initial matches"""
     # get some molecules
+    match_cache_key = nonbonded.BOND_SMIRK_MATCH_CACHE
     suppl = Chem.SDMolSupplier("tests/data/ligands_40.sdf", removeHs=False)
     all_mols = [mol for mol in suppl]
 
@@ -620,8 +624,12 @@ def test_bond_smirks_matches():
     smirks_list = [smirks for (smirks, param) in AM1CCC_CHARGES["patterns"]]
 
     for mol in all_mols:
-        bond_idxs, type_idxs = nonbonded.bond_smirks_matches(mol, smirks_list)
+        assert not mol.HasProp(match_cache_key)
 
+    fresh_matches = []
+    for mol in all_mols:
+        bond_idxs, type_idxs = nonbonded.compute_or_load_bond_smirks_matches(mol, smirks_list)
+        fresh_matches.append((bond_idxs, type_idxs))
         # assert indices in bounds
         assert (bond_idxs.min() >= 0) and (bond_idxs.max() < mol.GetNumAtoms())
         assert (type_idxs.min() >= 0) and (type_idxs.max() < len(smirks_list))
@@ -634,6 +642,12 @@ def test_bond_smirks_matches():
             bonds.add((b, a))
         for bond in bond_idxs:
             assert tuple(bond) in bonds
+    for mol in all_mols:
+        assert mol.HasProp(match_cache_key)
+    cached_matches = [nonbonded.compute_or_load_bond_smirks_matches(mol, smirks_list) for mol in all_mols]
+    for (fresh_bonds, fresh_types), (cached_bonds, cached_types) in zip(fresh_matches, cached_matches):
+        np.testing.assert_array_equal(fresh_bonds, cached_bonds)
+        np.testing.assert_array_equal(fresh_types, cached_types)
 
 
 def test_apply_bond_charge_corrections():
--- a/timemachine/ff/handlers/bcc_aromaticity.py
+++ b/timemachine/ff/handlers/bcc_aromaticity.py
@@ -6,9 +6,8 @@ import logging
 import re
 from typing import Any, Callable, Dict, List, Type, TypeVar
 
-from openeye import oechem
-
 T = TypeVar("T")
+OEMol = Any
 
 
 class RechargeException(BaseException):
@@ -51,6 +50,7 @@ def call_openeye(
     exception_kwargs
         The keyword arguments to pass to the exception.
     """
+    from openeye import oechem
 
     if exception_kwargs is None:
         exception_kwargs = {}
@@ -79,7 +79,7 @@ def call_openeye(
         logging.debug(output_string)
 
 
-def match_smirks(smirks: str, oe_molecule: oechem.OEMol, unique: bool = False) -> List[Dict[int, int]]:
+def match_smirks(smirks: str, oe_molecule: OEMol, unique: bool = False) -> List[Dict[int, int]]:
     """Attempt to find the indices (optionally unique) of all atoms which
     match a particular SMIRKS pattern.
     Parameters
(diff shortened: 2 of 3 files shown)


tests/test_handlers.py
```diff
@@ -581,12 +581,12 @@ def test_am1_differences():
             assert 0
 
 
-def test_compute_or_load_am1_charges():
+def test_compute_or_load_am1_charges_and_matches():
```
maxentile, line 584: Nit: naming -- test renamed but definition doesn't seem to reflect this change

timemachine/ff/handlers/nonbonded.py
```diff
@@ -15,6 +15,9 @@
 from timemachine.ff.handlers.utils import match_smirks as rd_match_smirks
 from timemachine.ff.handlers.utils import sort_tuple
 
+AM1_CHARGE_CACHE = "AM1Cache"
+AM1_SMIRK_MATCH_CACHE = "AM1BondMatchCache"
```
maxentile, line 19: Nit: naming -- these smirks matches are not strictly AM1-derived. Perhaps `BCC_SMIRK_MATCH_CACHE` or similar.

(In general, the list of smirks matches will depend on which force field is loaded -- not sure if this design needs to be defensive against populating the cache with one force field, then using the cache with a different force field. Tangent: An analogous consideration also applies to how the charge cache is implemented: https://github.com/proteneer/timemachine/pull/571#discussion_r788307734 )

timemachine/ff/handlers/nonbonded.py
```diff
@@ -159,23 +162,22 @@ def compute_or_load_am1_charges(mol):
...
 
     else:
-        am1_charges = pickle.loads(base64.b64decode(mol.GetProp(cache_key)))
+        am1_charges = pickle.loads(base64.b64decode(mol.GetProp(AM1_CHARGE_CACHE)))
 
     return np.array(am1_charges)
 
 
-def bond_smirks_matches(mol, smirks_list):
+def compute_or_load_bond_smirks_matches(mol, smirks_list):
```
maxentile, line 180: Nit: updates behavior but does not update docstring

timemachine/ff/handlers/bcc_aromaticity.py
```diff
@@ -77,7 +78,7 @@ def call_openeye(
         logging.debug(output_string)
 
 
-def match_smirks(smirks: str, oe_molecule: "OEMol", unique: bool = False) -> List[Dict[int, int]]:
```
badisa: flake8 complains about these being undefined variables, looks like it was decided that this was invalid style: https://github.com/PyCQA/pyflakes/issues/340

Even adding `oechem.OEMol` fails.

timemachine/ff/handlers/nonbonded.py
```diff
@@ -187,26 +189,29 @@ def bond_smirks_matches(mol, smirks_list):
...
-                bond_idxs.append(forward_matched_bond)
-                type_idxs.append(type_idx)
+                already_assigned = forward_matched_bond in bond_idxs or reverse_matched_bond in bond_idxs
 
+                if not already_assigned:
+                    bond_idxs.append(forward_matched_bond)
+                    type_idxs.append(type_idx)
+        mol.SetProp(AM1_SMIRK_MATCH_CACHE, base64.b64encode(pickle.dumps((bond_idxs, type_idxs))))
+    else:
+        bond_idxs, type_idxs = pickle.loads(base64.b64decode(mol.GetProp(AM1_SMIRK_MATCH_CACHE)))
```
maxentile, line 214: Nit: Now that multiple properties are stored as pickle files inside of each rdkit mol, we may want to consider factoring out a function / decorator for managing caches, so that each function body for computing cache-able properties doesn't need to be indented another level / modified so much if the caching policy changes...

timemachine/ff/handlers/nonbonded.py
```diff
@@ -187,26 +189,29 @@ def bond_smirks_matches(mol, smirks_list):
     * Order within each smirks pattern matters
         For example, "[#6:1]~[#1:2]" and "[#1:1]~[#6:2]" will match atom pairs in the opposite order
     """
-    oemol = convert_to_oe(mol)
-    AromaticityModel.assign(oemol)
+    if not mol.HasProp(AM1_SMIRK_MATCH_CACHE):
+        oemol = convert_to_oe(mol)
+        AromaticityModel.assign(oemol)
```
maxentile, line 194: Another possibility to remove the OpenEye dependence of this function might be to compute the list of bond smirks matches using RDKit -- but I don't immediately remember if there are important RDKit / OpenEye differences in how these SMIRKS are interpreted.

(In some corner cases, identical SMIRKS strings can have different meanings between RDKit and OpenEye.)

timemachine/ff/handlers/bcc_aromaticity.py
mcwitt: > Even adding oechem.OEMol fails.

hmm, did you import `oechem`? This seems to work on my end

maxentile (approved):
Looks good! Minor comments

timemachine/ff/handlers/bcc_aromaticity.py
mcwitt: one workaround could be adding `OEMol = Any` at top level, so we can continue to use the annotations for documentation purposes

timemachine/ff/handlers/nonbonded.py
badisa: In this context, it looks like we are using vanilla `AM1` not `AM1BCC` or `AM1BCCELF10`. Should we be using the BCC charges?

Thats a good question, will need to think about that.

timemachine/ff/handlers/bcc_aromaticity.py
mcwitt: nit: probably fancier than we need, but this looks like a potential solution: https://github.com/PyCQA/pyflakes/issues/340#issuecomment-513821884

timemachine/ff/handlers/nonbonded.py
badisa: I had considered this, though since we use OpenEye already for Quacpac it seemed natural to keep this. Plus exactly what you said about corner cases.

timemachine/ff/handlers/nonbonded.py
maxentile: I may have misunderstood -- I thought the bond smirks matches were defined by a file like https://github.com/proteneer/timemachine/blob/5d8f3099aafa1cf7def0585525aaf4de73c23343/timemachine/ff/params/smirnoff_1_1_0_ccc.py#L343 , not by AM1

badisa: You are right they are, sorry had thought we were using the AM1 charges in the function but we are not. Perhaps just `BOND_SMIRK_MATCH_CACHE`, as really the function doesn't care about the origins of the smirks?

maxentile: (For example, if we parameterize mol using the list of patterns in https://github.com/proteneer/timemachine/blob/5d8f3099aafa1cf7def0585525aaf4de73c23343/tests/test_handlers.py#L298 , we will populate this cache in an arbitrary, not AM1-derived way. If we then try to parameterize the same mol object with another force field with a different list of patterns, we will attempt to load from the cache, but the cache will be stale.)

maxentile: >Perhaps just BOND_SMIRK_MATCH_CACHE, as really the function doesn't care about the origins of the smirks?

That seems a better name.

(There's a separate issue of when the new cache might become stale, but this doesn't have to be addressed here -- similar to the current level of safety.)

timemachine/ff/handlers/nonbonded.py
badisa: Fixed in [21039c3](https://github.com/proteneer/timemachine/pull/640/commits/21039c3556e0ba96419309922e6b5c41f8411330)

timemachine/ff/handlers/nonbonded.py
badisa: Renamed in [21039c3](https://github.com/proteneer/timemachine/pull/640/commits/21039c3556e0ba96419309922e6b5c41f8411330)

tests/test_handlers.py
badisa: Fixed in [21039c3](https://github.com/proteneer/timemachine/pull/640/commits/21039c3556e0ba96419309922e6b5c41f8411330)

timemachine/ff/handlers/nonbonded.py
```diff
@@ -187,26 +189,29 @@ def bond_smirks_matches(mol, smirks_list):
...
-            already_assigned = forward_matched_bond in bond_idxs or reverse_matched_bond in bond_idxs
-
-            if not already_assigned:
-                bond_idxs.append(forward_matched_bond)
-                type_idxs.append(type_idx)
+                already_assigned = forward_matched_bond in bond_idxs or reverse_matched_bond in bond_idxs
 
+                if not already_assigned:
+                    bond_idxs.append(forward_matched_bond)
+                    type_idxs.append(type_idx)
```
mcwitt, line 211: nit: might be nice to split this into a separate function `compute_bond_smirks_matches` to avoid a large indented block

mcwitt (approved):
lgtm!

badisa:
Discussed with @proteneer offline

PR #642, opened 2022-02-22
Hack/gitlab testing

badisa:
Another attempt


badisa:
Converting to 'real' PR to see if that makes a difference

Issue #639, opened 2022-02-16
Clean and DRY nonbonded potential code


mcwitt:
Miscellaneous items applying to code that currently exists in multiple nonbonded potentials:

- https://github.com/proteneer/timemachine/pull/578#discussion_r808169237
- https://github.com/proteneer/timemachine/pull/578#discussion_r808165543
- https://github.com/proteneer/timemachine/pull/578#discussion_r808155816
- https://github.com/proteneer/timemachine/pull/578#discussion_r808131513

PR #643, opened 2022-02-22
Wrap NonbondedAllPairs, NonbondedPairList

mcwitt:
- Exposes `NonbondedAllPairs` and `NonbondedPairList` in the Python API
- Adds tests for both potentials
- Reorganizes new nonbonded tests under `tests/nonbonded` to allow sharing fixtures
- Improves some error messages

Note: this does not yet add the consistency check

```
U = U_A + U_B + U_AB
```

since this requires updating `NonbondedAllPairs` to accept a subset of atom indices. I'm planning to add this and the consistency check in an upcoming PR.

### DIFF
--- a/tests/nonbonded/test_nonbonded_all_pairs.py
+++ b/tests/nonbonded/test_nonbonded_all_pairs.py
@@ -1,0 +1,126 @@
+import functools
+
+import numpy as np
+import pytest
+from common import GradientTest
+
+from timemachine.lib.potentials import NonbondedAllPairs, NonbondedAllPairsInterpolated
+from timemachine.potentials import nonbonded
+
+
+def test_nonbonded_all_pairs_invalid_planes_offsets():
+    with pytest.raises(RuntimeError) as e:
+        NonbondedAllPairs([0], [0, 0], 2.0, 1.1).unbound_impl(np.float32)
+
+    assert "lambda offset idxs and plane idxs need to be equivalent" in str(e)
+
+
+def test_nonbonded_all_pairs_invalid_num_atoms():
+    potential = NonbondedAllPairs([0], [0], 2.0, 1.1).unbound_impl(np.float32)
+    with pytest.raises(RuntimeError) as e:
+        potential.execute(np.zeros((2, 3)), np.zeros((1, 3)), np.eye(3), 0)
+
+    assert "NonbondedAllPairs::execute_device(): expected N == N_, got N=2, N_=1" in str(e)
+
+
+def test_nonbonded_all_pairs_invalid_num_params():
+    potential = NonbondedAllPairs([0], [0], 2.0, 1.1).unbound_impl(np.float32)
+    with pytest.raises(RuntimeError) as e:
+        potential.execute(np.zeros((1, 3)), np.zeros((2, 3)), np.eye(3), 0)
+
+    assert "NonbondedAllPairs::execute_device(): expected P == M*N_*3, got P=6, M*N_*3=3" in str(e)
+
+    potential_interp = NonbondedAllPairsInterpolated([0], [0], 2.0, 1.1).unbound_impl(np.float32)
+    with pytest.raises(RuntimeError) as e:
+        potential_interp.execute(np.zeros((1, 3)), np.zeros((1, 3)), np.eye(3), 0)
+
+    assert "NonbondedAllPairs::execute_device(): expected P == M*N_*3, got P=3, M*N_*3=6" in str(e)
+
+
+def make_ref_potential(lambda_plane_idxs, lambda_offset_idxs, beta, cutoff):
+    @functools.wraps(nonbonded.nonbonded_v3)
+    def wrapped(conf, params, box, lamb):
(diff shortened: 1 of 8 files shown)


tests/nonbonded/test_nonbonded_all_pairs.py
```diff
@@ -0,0 +1,128 @@
...
+    example_nonbonded_params,
+    example_conf,
+    example_box,
+    rng: np.random.Generator,
+):
+    "Compares with jax reference implementation, with parameter interpolation."
+
+    conf = example_conf[:num_atoms]
+    params_initial = example_nonbonded_params[:num_atoms, :]
+    params_final = params_initial + np.where(params_initial, rng.normal(0, 0.01, size=params_initial.shape), 0)
```
mcwitt, line 117: Conditioning on the initial params being nonzero was necessary to prevent fixed-point overflow for sizes greater than ~2500. I'm not sure I fully understand this, but it seems like this might be the correct thing to do anyway (since LJ param == 0 <==> LJ disabled) and we might want to update the other interpolation tests with this condition

maxentile: >Conditioning on the initial params being nonzero was necessary to prevent fixed-point overflow for sizes greater than ~2500, presumably due to large LJ energies when we turn on LJ interactions that were initially disabled?

That sounds plausible.

--------

Tangent: The workaround `rng.normal(0, 0.01, ...)` can assign negative values to epsilon and sigma -- charges can be negative, but epsilon and sigma should always be non-negative.

Question: Do we have (or want to add) a warning if epsilon or sigma < 0?

tests/nonbonded/test_nonbonded_pair_list.py
```diff
@@ -0,0 +1,166 @@
...
+
+    # randomly select 2 interaction groups and construct all pairwise interactions
+    atom_idxs = rng.choice(
+        num_atoms,
+        size=(
+            2,
+            ixn_group_size,
+        ),
+        replace=False,
+    ).astype(np.int32)
```
maxentile, line 139: Question: does this generate inputs where `len(set(atom_idxs[0]).intersection(atom_idxs[1])) > 0`?

If so, it's good that this case is handled... Actually, does the code skip or include spurious self-interactions `(i, i)`? If these are included, they could possibly cause overflows or other issues.

In any case, it may be better to test this case separately from the standard case where `atom_idxs[0]` and `atom_idxs[1]` are disjoint.

maxentile (commented):
Looks good!

Just a couple small questions about the tests

badisa (approved):

tests/nonbonded/test_nonbonded_pair_list.py
mcwitt: > Question: does this generate inputs where len(set(atom_idxs[0]).intersection(atom_idxs[1])) > 0?

No, the intersection should be empty (`replace=False` ensures this).

> Actually, does the code skip or include spurious self-interactions (i, i)?

The `NonbondedPairList` constructor will [throw a runtime error](https://github.com/proteneer/timemachine/blob/a4a06d93df52536c115017f9b6bc2c3c329c3935/timemachine/cpp/src/nonbonded_pair_list.cu#L30-L37) if there are self interactions (tested [here](https://github.com/proteneer/timemachine/blob/a4a06d93df52536c115017f9b6bc2c3c329c3935/tests/nonbonded/test_nonbonded_pair_list.py#L21-L24)).

maxentile: Thanks -- that makes sense!

maxentile (approved):

tests/nonbonded/test_nonbonded_all_pairs.py
mcwitt: > The workaround rng.normal(0, 0.01, ...) can assign negative values to epsilon and sigma -- charges can be negative, but epsilon and sigma should always be non-negative.

Makes sense; I hadn't considered this. In d7c4095 I changed the generator to avoid negative values for sigma and epsilon.

> Question: Do we have (or want to add) a warning if epsilon or sigma < 0?

We do not have this currently; it sounds like a good idea to add a check. Unfortunately, adding this didn't seem totally trivial, since it currently requires duplicating the check code in each of the 3 nonbonded potentials. Is it OK with you if I add this as a subtask to #639 ?

maxentile: >In https://github.com/proteneer/timemachine/commit/d7c4095b6da7abfad6abf91416b26c86be8d740f I changed the generator to avoid negative values for sigma and epsilon.

Looks good!

>Unfortunately, adding this didn't seem totally trivial, since it currently requires duplicating the check code in each of the 3 nonbonded potentials. Is it OK with you if I add this as a subtask to https://github.com/proteneer/timemachine/issues/639 ?

Fine by me!

(Non-blocking for the current PR, and probably also not worth doing in a way that introduces much duplication)

PR #649, opened 2022-02-25
Switches OR to AND for check if the branch is in a PR

badisa:
I was dumb and put in a check that is always true, which is why all branches/commits were running. Verified that if you try to run the pipeline without a PR open, it won't trigger

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -14,7 +14,7 @@ workflow:
     # In the case of the default branch, we want to build the image so that
     # other branches have a base image to cache from when possible.
     - if: '$CI_COMMIT_BRANCH == $CI_DEFAULT_BRANCH'
-    - if: '$CI_EXTERNAL_PULL_REQUEST_IID != null || $CI_EXTERNAL_PULL_REQUEST_IID != ""'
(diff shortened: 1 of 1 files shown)


maxentile (approved):

.gitlab-ci.yml
```diff
@@ -14,7 +14,7 @@ workflow:
     # In the case of the default branch, we want to build the image so that
     # other branches have a base image to cache from when possible.
     - if: '$CI_COMMIT_BRANCH == $CI_DEFAULT_BRANCH'
-    - if: '$CI_EXTERNAL_PULL_REQUEST_IID != null || $CI_EXTERNAL_PULL_REQUEST_IID != ""'
+    - if: '$CI_EXTERNAL_PULL_REQUEST_IID != null && $CI_EXTERNAL_PULL_REQUEST_IID != ""'
```
mcwitt, line 17: Just curious, do we have evidence that both conditions are really needed here? Seeing a lot of [examples](https://github.com/search?q=ci_external_pull_request_iid+%21%3D&type=code) that just have
```yaml
- if: '$CI_EXTERNAL_PULL_REQUEST_IID 
```
which seems like it might cover all cases?

badisa:
There is a problem with github triggering builds on pull requests, won't be merged until that is resolved.

badisa:
With the help of @proteneer the mirroring/CI is working correctly. Commits in a PR will be triggered.

PR #652, opened 2022-02-25
Update README.md again

fehomi:


badisa:
This wasn't triggered in CI, which is what we had hoped for.

PR #653, opened 2022-02-25
Adds reproduction of failure

badisa:


.gitlab-ci.yml
```diff
@@ -51,6 +51,8 @@ memory-tests:
     - gpu
   variables:
     CMAKE_ARGS: -DCUDA_ARCH=75
+  rules:
+    - if: $CI_EXTERNAL_PULL_REQUEST_IID
```
badisa, line 55: If this stage wants to run but the stage that it needs doesn't run the yml is invalid. This only showed up on master where the stage was skipped.

Look at the previous commits and the corresponding gitlab statuses to see how this manifests

mcwitt: Ah, so is idea is to run only the docker build in the default branch but skip the other stages? And before, we were only explicitly skipping the lint stage, but the test stages depend on that and weren't able to start? If so, I think this fix makes sense.

mcwitt (approved):

.gitlab-ci.yml
badisa: Correct. We want to build the image for the default branch for caching reasons. The rest of the stages should be canceled, and I had incorrectly assumed that cancelling up an upstream dependency would cancel everything downstream.

PR #654, opened 2022-02-26
Reweighting utilities

maxentile:
Adds functions for two kinds of differentiable reweighting (`construct_endpoint_reweighting_estimator`, `construct_mixture_reweighting_estimator`).

These are tested for correctness on a 1D system (where comparisons to exact free energies and gradients are possible).

Less stringent tests confirm that the same implementation is compatible with custom_ops in an absolute hydration free energy test system.

--------

Notes to reviewers:
* Length -- Although the raw number of lines added is pretty large, some consolation is that **there are only ~40-50 lines of Python statements in `reweighting.py`** (most of the total line count is in tests, and most of the lines in `reweighting.py` are documentation / white space)
* Comparison to https://github.com/proteneer/timemachine/pull/404 -- Structured differently (free functions rather than classses), omits some features (no effective sample size diagnostic), is less jax-specific (assumes access to a `batched_u_fxn`, rather than a `u_fxn` which can be transformed by `jax.vmap`), and has a more generic interface.
  * The more generic interface attempts to clarify relation to MBAR:
     * Both of the reweighting estimators in this PR are compatible with inputs that are **not** necessarily processed by MBAR.
     * The `log_weights` required by one of these approaches **can** be computed from a collection of sampled states using MBAR via `interpret_as_mixture_potential`.
     * The tests assert the approach is compatible with other estimators in principle (although if we have the full `u_kn` matrix in hand, there would be no reason to prefer, say, TI's estimate of `f_k` over `mbar.f_k`).
     * A reference is included about other ways to get (higher-variance estimates) of these `log_weights`, that would not require the full `u_kn` matrix.
     * In the future, if we sampled from a mixture distribution with exactly known logpdf (rather than inferring the mixture logpdf from multiple states via MBAR), that would also supply us with these `log_weights`.
* Test structure -- there's some avoidable repetition in the tests, happy to re-organize this
* Type annotations -- added in this commit: https://github.com/proteneer/timemachine/commit/bff4ae53f1d31fa102d0b8b381ba047e45eb68a6 -- not sure if these clarify or just add noise, happy to pare down
* Docstrings -- didn't fully expand parameter descriptions, ASCII-art diagrams are currently a bit ugly


timemachine/fe/reweighting.py
```diff
@@ -0,0 +1,192 @@
...
+    assert mixture_u_n.shape == (n_samples,)
+
+    return mixture_u_n
+
+
+def construct_endpoint_reweighting_estimator(
+    samples_0: Samples,
+    samples_1: Samples,
+    batched_u_0_fxn: BatchedReducedPotentialFxn,
+    batched_u_1_fxn: BatchedReducedPotentialFxn,
```
maxentile, line 77: Naming -- I need a good variable name (and type annotation) for the user-provided function that computes a batch of energies.

Initially I had named this `vec_u_fxn`, but in discussion with @mcwitt noticed this is unwieldy (suggests vectorization, but this should be agnostic to how the batch is computed).

Currently using `batched_u_fxn: BatchedReducedPotentialFxn` (to be agnostic to implementation, but hint that we'll apply batch-specific performance optimizations). Not settled on this name, and the same concept will be used in a few other upcoming features.

@mcwitt has also suggested `map_` prefix.

timemachine/fe/reweighting.py
```diff
@@ -0,0 +1,192 @@
...
+        ref_delta_f ~= f(ref_params, 1) - f(ref_params, 0)
+
+    construct an estimator of f(params, 1) - f(params, 0)
+    """
+    ref_u_0 = batched_u_0_fxn(samples_0, ref_params)
+    ref_u_1 = batched_u_1_fxn(samples_1, ref_params)
+
+    def endpoint_correction_0(params) -> float:
+        """estimate f(ref, 0) -> f(params, 0) by reweighting"""
+        delta_us = batched_u_0_fxn(samples_0, params) - ref_u_0
```
maxentile, line 93: Interface -- Currently the signature is `BatchedReducedPotentialFxn = Callable[[Samples, Params], Energies]`. It may be more general to switch to `BatchedReducedPotentialFxn = Callable[[Params], Energies]`.

Since `samples` will be held constant, a summary of the `samples` can be precomputed once and incorporated into the definition of the batched potential function in some performance optimizations, as in https://github.com/proteneer/timemachine/blob/bb5c7288054ace37804d6c1e3c5cce24ad821d99/timemachine/potentials/nonbonded.py#L360

tests/test_reweighting.py
```diff
@@ -0,0 +1,392 @@
...
+    # u_mix_smc = ...
+
+    u_mixes = dict(exact_f_k=u_mix_exact, mbar_f_k=u_mix_mbar, ti_f_k=u_mix_ti)
+    analytical_delta_f = lambda params: reduced_free_energy(1.0, params) - reduced_free_energy(0.0, params)
+
+    for condition in u_mixes:
+        u_mix = u_mixes[condition]
+
+        # TODO [sign convention]: change signature to accept u_mix rather than log_weights?
+        estimate_delta_f = jit(construct_mixture_reweighting_estimator(xs, -u_mix, vec_u_0_fxn, vec_u_1_fxn))
```
maxentile, line 175: Inconsistent sign convention -- waffling between "reduced potential"s and "reduced_free_energies" everywhere vs. "log_pdf"s and "log_normalizing_constant"s everywhere...

timemachine/fe/reweighting.py
```diff
@@ -0,0 +1,192 @@
...
+    References
+    ----------
+    [1] [Shirts, 2017] Reweighting from the mixture distribution as a better way to describe
+        the Multistate Bennett Acceptance Ratio
+        https://arxiv.org/abs/1704.00891
+    [2] [Shirts, Chodera, 2008] Statistically optimal analysis of samples from multiple equilibrium states.
+        J. Chem. Phys. 129:124105, 2008.
+        http://dx.doi.org/10.1063/1.2978177
+    [3] [Elvira+, 2019] Generalized multiple importance sampling
+        https://arxiv.org/abs/1511.03095
```
maxentile, line 59: Figure 3 of ref [3] is pretty neat

timemachine/fe/reweighting.py
```diff
@@ -0,0 +1,286 @@
...
+    "interpret_as_mixture_potential",
+]
+
+from typing import Callable, Collection
+
+from jax import numpy as np
+from jax.scipy.special import logsumexp
+
+Samples = Params = Collection
+Energies = Array = np.array
```
mcwitt, line 14: I'm not sure why, but this style seems to break mypy checks for me (many seemingly-unrelated errors are generated). The problem disappears when I change this to
``` python
Samples = Collection
Params = Collection
Array = np.ndarray
Energies = Array
```

mcwitt, line 14:
```suggestion
Energies = Array = np.ndarray
```
(without this I get `error: Function "jax._src.numpy.lax_numpy.array" is not valid as a type`)

maxentile: Ahh! Thanks, I hadn't been running these using mypy so far, meant these mostly as human-readable annotations.

This file introduces a bunch of mypy errors:

<details>

```text
mypy timemachine/fe/reweighting.py 
timemachine/fe/reweighting.py:15: error: Variable "timemachine.fe.reweighting.Samples" is not valid as a type
timemachine/fe/reweighting.py:15: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
timemachine/fe/reweighting.py:15: error: Variable "timemachine.fe.reweighting.Params" is not valid as a type
timemachine/fe/reweighting.py:15: error: Variable "timemachine.fe.reweighting.Energies" is not valid as a type
timemachine/fe/reweighting.py:18: error: Variable "timemachine.fe.reweighting.Array" is not valid as a type
timemachine/fe/reweighting.py:18: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
timemachine/fe/reweighting.py:28: error: Variable "timemachine.fe.reweighting.Array" is not valid as a type
timemachine/fe/reweighting.py:28: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
timemachine/fe/reweighting.py:33: error: Variable "timemachine.fe.reweighting.Array" is not valid as a type
timemachine/fe/reweighting.py:33: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
timemachine/fe/reweighting.py:43: error: Unsupported operand type for unary - (Array?)
timemachine/fe/reweighting.py:46: error: Variable "timemachine.fe.reweighting.Array" is not valid as a type
timemachine/fe/reweighting.py:46: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
timemachine/fe/reweighting.py:101: error: Array? has no attribute "shape"
timemachine/fe/reweighting.py:103: error: Array? has no attribute "shape"
timemachine/fe/reweighting.py:106: error: Array? has no attribute "T"
timemachine/fe/reweighting.py:114: error: Variable "timemachine.fe.reweighting.Samples" is not valid as a type
timemachine/fe/reweighting.py:114: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
timemachine/fe/reweighting.py:115: error: Variable "timemachine.fe.reweighting.Samples" is not valid as a type
timemachine/fe/reweighting.py:115: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
timemachine/fe/reweighting.py:118: error: Variable "timemachine.fe.reweighting.Params" is not valid as a type
timemachine/fe/reweighting.py:118: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
timemachine/fe/reweighting.py:120: error: Variable "timemachine.fe.reweighting.Params" is not valid as a type
timemachine/fe/reweighting.py:120: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
timemachine/fe/reweighting.py:162: error: Unsupported left operand type for - (Energies?)
timemachine/fe/reweighting.py:167: error: Unsupported left operand type for - (Energies?)
timemachine/fe/reweighting.py:170: error: Variable "timemachine.fe.reweighting.Params" is not valid as a type
timemachine/fe/reweighting.py:170: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
timemachine/fe/reweighting.py:196: error: Variable "timemachine.fe.reweighting.Samples" is not valid as a type
timemachine/fe/reweighting.py:196: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
timemachine/fe/reweighting.py:197: error: Variable "timemachine.fe.reweighting.Array" is not valid as a type
timemachine/fe/reweighting.py:197: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
timemachine/fe/reweighting.py:200: error: Variable "timemachine.fe.reweighting.Params" is not valid as a type
timemachine/fe/reweighting.py:200: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
timemachine/fe/reweighting.py:263: error: Unsupported operand type for unary - (Energies?)
Found 22 errors in 1 file (checked 1 source file)
```

</details>

Will see if I can fix these up

timemachine/fe/reweighting.py
```diff
@@ -0,0 +1,289 @@
...
+    ----------
+    [1] pymbar implementation
+        https://github.com/choderalab/pymbar/blob/15f932a271343e611ed4be2d468c77b1d11cf01f/pymbar/exp.py#L54
+    """
+    # delta_us = -log_importance_weights
+    # delta_f  = -log_z_ratio
+    return -estimate_log_z_ratio(-delta_us)
+
+
+def interpret_as_mixture_potential(u_kn: Array, f_k: Array, N_k: Array) -> Array:
```
mcwitt, line 49: nit: could this be relaxed to
```suggestion
def interpret_as_mixture_potential(u_kn: Array, f_k: Array, N_k: list) -> Array:
```

tests/test_reweighting.py
```diff
@@ -0,0 +1,392 @@
...
+    vec_u = vmap(u_fxn, in_axes=(0, None, None))
+    vec_u_0_fxn = lambda xs, params: vec_u(xs, 0, params)
+    vec_u_1_fxn = lambda xs, params: vec_u(xs, 1, params)
+
+    estimate_delta_f = construct_endpoint_reweighting_estimator(
+        samples_0, samples_1, vec_u_0_fxn, vec_u_1_fxn, ref_params, ref_delta_f
+    )
+    analytical_delta_f = lambda params: reduced_free_energy(1.0, params) - reduced_free_energy(0.0, params)
+
+    assert_estimator_accurate(jit(estimate_delta_f), analytical_delta_f, ref_params, n_random_trials=10, atol=5e-3)
```
mcwitt, line 110: Out of curiosity, what's the reasoning for `jit`ing the estimator here? (i.e. is to make the test run faster, make the test closer common real-world usage, or something else?)

maxentile: Mostly to make the test run fast!

timemachine/fe/reweighting.py
mcwitt: (this is an extreme nit and the suggestion probably not what you intended -- just noticed we call `np.array` on the argument below, and pass in lists in the test code)

maxentile: Waffling between relaxing the signature to something more accurate like `Sequence` or Union[List, Array]` or similar vs. restricting the usages to `Array`...

Just to keep the signature simple / uniform, I'd prefer to just restrict the usages in the tests/ to `Array` (but keep the defensive array conversion in case someone accidentally passes in a list instead of an array)?

timemachine/fe/reweighting.py
maxentile: Nice -- applying the two small suggested changes https://github.com/proteneer/timemachine/pull/654/commits/f1ad7ab50a455f89be15722645a25a2578405562 and https://github.com/proteneer/timemachine/pull/654/commits/d0b565995c76f030899c18b4b70086f0e96818d5, whittles this down to only one mypy error:

```
timemachine/fe/reweighting.py:109: error: "ndarray" has no attribute "T"
Found 1 error in 1 file (checked 1 source file)
```
on this line
 
https://github.com/proteneer/timemachine/blob/d0b565995c76f030899c18b4b70086f0e96818d5/timemachine/fe/reweighting.py#L109

which I could avoid by using `np.transpose(x)` instead of `x.T`...

err, might be a good idea to expand this out to be more self-documenting / explicit anyway (avoiding the implicit vector minus matrix = matrix rank-promotion...)

tests/test_reweighting.py
```diff
@@ -0,0 +1,392 @@
...
+
+    def reduced_free_energy(lam, params):
+        mean, stddev = annealed_gaussian_def(lam, params)
+        log_z = np.log(stddev * np.sqrt(2 * np.pi))
+        return -log_z
+
+    return u_fxn, normalized_u_fxn, sample, reduced_free_energy
+
+
+def assert_estimator_accurate(estimate_delta_f, analytical_delta_f, ref_params, n_random_trials, atol=5e-3):
```
mcwitt, line 62: nit: I wonder if the default value for `atol` could lead to confusion if we write a new test and forget to tune the value. Would it make sense to omit the default?

timemachine/fe/reweighting.py
maxentile: I expanded this a bit in https://github.com/proteneer/timemachine/pull/654/commits/4dd40082698c18c34242313a30c1a0e7b6e46b22 to improve general readability

mcwitt: Nice, these seem like good changes! I'm also realizing that typing the `vmap`ped code might be more trouble than it's worth -- it looks like the [jax annotation](https://github.com/google/jax/blob/c7508d1f2daa4fc6f7d74464d9d840a274c46fec/jax/_src/api.py#L1384) isn't precise enough to infer the return type of the generated function, so it seems like typing code that uses vmap might require a lot of manual annotation

timemachine/fe/reweighting.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/654/commits/59d8a6356159fa3fecfbbdd7e65db7ca14ff1132 !

tests/test_reweighting.py
maxentile: Default omitted + parameter documentation added in https://github.com/proteneer/timemachine/pull/654/commits/dcfbe1d161930b8f0a765deb0aaac754ae5c779a

timemachine/fe/reweighting.py
maxentile: Documented these steps a little more, and retained the equivalent one-liner in a comment in https://github.com/proteneer/timemachine/pull/654/commits/9f9971ae8248f16b0b946a425935b70167db5fbb

maxentile: >I'm also realizing that typing the vmapped code might be more trouble than it's worth -- it looks like the [jax annotation](https://github.com/google/jax/blob/c7508d1f2daa4fc6f7d74464d9d840a274c46fec/jax/_src/api.py#L1384) isn't precise enough to infer the return type of the generated function, so it seems like typing code that uses vmap might require a lot of manual annotation

Hmm, yeah, the effect of jax transformations is not precisely annotated in the types (grad: Callable -> Callable, vmap: Callable -> Callable, ...).

I think it might still make sense to annotate types for (1) a function that processes "a single element" e.g. `f: Item -> float`, (2) interfaces that assume batches e.g. `batch_f: Items -> Array` -- even if static analysis might not be able to verify that applying vmap to (1) will produce a function that matches the type annotation of (2).

badisa (approved):

tests/test_reweighting.py
```diff
@@ -0,0 +1,407 @@
...
+    """normalized/unnormalized 1D Gaussian with a dependence on lambda and params"""
+
+    def annealed_gaussian_def(lam, params):
+        initial_mean, initial_log_sigma = 0.0, 0.0
+        target_mean, target_log_sigma = params
+
+        # lam = 0 -> (mean = 0, stddev = 1)
+        # lam = 1 -> (mean = target_mean, stddev = target_sigma)
+        mean = lam * target_mean - (1 - lam) * initial_mean
+        stddev = np.exp(lam * target_log_sigma + (1 - lam) * initial_log_sigma)
```
proteneer, line 34: Mostly for my own curiosity: any reason in particular to weight stddev in logspace?

tests/test_reweighting.py
```diff
@@ -0,0 +1,407 @@
...
+    ref_params: [2,] array
+        generate random parameters near ref_params
+    n_random_trials: int
+    atol: float
+        absolute tolerance
+    """
+
+    def sample_random_params():
+        mean = ref_params[0] + onp.random.rand()
+        log_sigma = ref_params[1] - onp.random.rand()
```
proteneer, line 83: Is the idea for log_sigma that we can avoid dealing with negative values of sigma by doing all of this in log space? (Since `exp(log_sigma)` is strictly positive for all `-inf < log sigma < inf`)

tests/test_reweighting.py
```diff
@@ -0,0 +1,407 @@
...
+
+    onp.testing.assert_allclose(f_hat, f_ref, atol=atol)
+    onp.testing.assert_allclose(g_hat, g_ref, atol=atol)
+
+    for _ in range(n_random_trials):
+        trial_params = sample_random_params()
+        f_hat, g_hat = value_and_grad(estimate_delta_f)(trial_params)
+        f_ref, g_ref = value_and_grad(analytical_delta_f)(trial_params)
+
+        onp.testing.assert_allclose(f_hat, f_ref, atol=atol)
```
proteneer, line 97: atol makes sense here, since we're typically happy with a precision of say around 0.1kJ/mol

tests/test_reweighting.py
```diff
@@ -0,0 +1,407 @@
...
+        onp.testing.assert_allclose(g_hat, g_ref, atol=atol)
```
proteneer, line 98: How should we think about precision of `dDeltaF/dParams` both for the gaussian case, and in general?  For the gaussian case, I'd imagine that `dDeltaF/dlog_sigma` might require more precision than `dDeltaF/dmean`. Presumably, an acceptable level of error is proportional to the internal learning rate one might take when optimizing the parameters, so maybe `rtol` is more approriate here?

tests/test_reweighting.py
```diff
@@ -0,0 +1,407 @@
...
+
+
+def test_endpoint_reweighting_1d():
+    """assert that endpoint reweighting estimator for delta_f(params), grad(delta_f)(params) is accurate
+    on tractable 1D system"""
+    onp.random.seed(2022)
+
+    u_fxn, _, sample, reduced_free_energy = make_gaussian_testsystem()
+
+    ref_params = np.zeros(2)
```
proteneer, line 108: I kept on forgetting that this is target `(mean, log_stddev)` - possible to add a brief comment here?

tests/test_reweighting.py
```diff
@@ -0,0 +1,407 @@
...
+        f_hat, g_hat = value_and_grad(estimate_delta_f)(trial_params)
+        f_ref, g_ref = value_and_grad(analytical_delta_f)(trial_params)
+
+        onp.testing.assert_allclose(f_hat, f_ref, atol=atol)
+        onp.testing.assert_allclose(g_hat, g_ref, atol=atol)
+
+
+def test_endpoint_reweighting_1d():
+    """assert that endpoint reweighting estimator for delta_f(params), grad(delta_f)(params) is accurate
+    on tractable 1D system"""
```
proteneer, line 103: Took me a bit of digging to realize this is testing reweighting from:

`u_0, stddev_0 = (0, 1)` to `u_1, stddev_1 = (1, 1)`

Can you add this to the documentation?

tests/test_reweighting.py
```diff
@@ -0,0 +1,407 @@
...
+    onp.random.seed(2022)
+
+    u_fxn, _, sample, reduced_free_energy = make_gaussian_testsystem()
+
+    ref_params = np.zeros(2)
+    ref_delta_f = reduced_free_energy(1.0, ref_params) - reduced_free_energy(0.0, ref_params)
+
+    n_samples = int(1e6)
```
proteneer, line 111: Might be useful to note that `n_samples` and `ref_params` is what determines `atol` in `assert_estimator_accurate` below - i.e. gaussians that are further apart need more samples or something.

tests/test_reweighting.py
```diff
@@ -0,0 +1,407 @@
...
+    on tractable 1D system"""
+    onp.random.seed(2022)
+
+    u_fxn, normalized_u_fxn, sample, reduced_free_energy = make_gaussian_testsystem()
+
+    ref_params = np.ones(2)
+    n_windows = 10
+    lambdas = np.linspace(0, 1, n_windows)
+
+    n_samples_per_window = int(1e5)
```
proteneer, line 140: Clarification - compared to the `endpoint_reweighting_test` above, we now have 10 windows that have 10x fewer steps, so the total number of samples is the same. However, the `atol=1e-2` is actually higher than the above test! This is counter-intuitive to me as I would have thought that use of intermediate windows should improve the efficiency of the estimator.

tests/test_reweighting.py
```diff
@@ -0,0 +1,407 @@
...
+    params_prime = ref_params + 0.01 * onp.random.randn(*ref_params.shape)
+    v_prime, g_prime = value_and_grad(estimate_delta_f)(params_prime)
+    assert v_prime != v
+    assert (g_prime != g).any()
+    assert np.isfinite(v_prime)
+    assert np.isfinite(g_prime).all()
+
+
+def test_one_sided_exp():
+    """assert consistency with pymbar.EXP on random instances"""
```
proteneer, line 321: Might be useful to add a test with `inf` values as well, I don't recall if pymbar does something clever to mask them out.

timemachine/fe/reweighting.py
```diff
@@ -0,0 +1,309 @@
...
+def one_sided_exp(delta_us: Array) -> float:
+    """exponential averaging
+
+    References
+    ----------
+    [1] pymbar implementation
+        https://github.com/choderalab/pymbar/blob/15f932a271343e611ed4be2d468c77b1d11cf01f/pymbar/exp.py#L54
+    """
+    # delta_us = -log_importance_weights
+    # delta_f  = -log_z_ratio
```
proteneer, line 45: extraneous?

timemachine/fe/reweighting.py
```diff
@@ -0,0 +1,309 @@
...
+    [1] [Shirts, 2017] Reweighting from the mixture distribution as a better way to describe
+        the Multistate Bennett Acceptance Ratio
+        https://arxiv.org/abs/1704.00891
+    [2] [Shirts, Chodera, 2008] Statistically optimal analysis of samples from multiple equilibrium states.
+        J. Chem. Phys. 129:124105, 2008.
+        http://dx.doi.org/10.1063/1.2978177
+    [3] [Elvira+, 2019] Generalized multiple importance sampling
+        https://arxiv.org/abs/1511.03095
+    """
+    # one-liner: mixture_u_n = -logsumexp(f_k - u_kn.T, b=N_k, axis=1)
```
proteneer, line 105: why not use one-liner?

proteneer (commented):
Did a first pass of the tests - will resume adding more comments on the actual implementation after dinner.

timemachine/fe/reweighting.py
proteneer: we use `batched_` pretty extensively in other parts of the code too

tests/test_reweighting.py
maxentile: No reason, just wrote down an arbitrary 1D test system with the desired behavior at lam = 0 and lam = 1.

---
Aside: in the special case of annealing between two 1D gaussians with the same mean but different sigma, this is an optimal protocol, as mentioned in eqs. 70-71 of [[Gelmen, Meng, 1998]](http://www.stat.columbia.edu/~gelman/research/published/path2.pdf).

(However, it doesn't matter whether this test uses an efficient protocol, the test doesn't use the appropriate generalization for 1D gaussians with different mean, and the differences in sigma are small -- log space isn't required for stability.)

tests/test_reweighting.py
maxentile: Will need to think about this.

tests/test_reweighting.py
maxentile: documented in the 1D system definition, will mention here too (or name the system) to reduce digging

timemachine/fe/reweighting.py
proteneer: We'd probably need to pin samples to a `DeviceBuffer` or something to truly see any meaningful performance gains here.

tests/test_reweighting.py
maxentile: willfix

timemachine/fe/reweighting.py
```diff
@@ -0,0 +1,309 @@
...
+        using this thermodynamic cycle:
+
+        f(params, 0)  --->  f(params, 1)
+
+             ^                   ^
+             |                   |
+             |                   |
+             |                   |
+
+        f(ref, 0)     --->  f(ref, 1)
```
proteneer, line 205: nice!

timemachine/fe/reweighting.py
```diff
@@ -0,0 +1,309 @@
...
+        J. Chem. Theory Comput. 2018, 14, 6, 3144–3162 https://doi.org/10.1021/acs.jctc.8b00223
+    [3] Wieder et al. PyTorch implementation of differentiable reweighting in neutromeratio
+        https://github.com/choderalab/neutromeratio/blob/2abf29f03e5175a988503b5d6ceeee8ce5bfd4ad/neutromeratio/parameter_gradients.py#L246-L267
+    """
+    assert len(samples) == len(log_weights)
+
+    def f_0(params):
+        """estimate f(params, 0) - f(ref) by reweighting"""
+        log_numerator = -batched_u_0_fxn(samples, params)
+        log_importance_weights = log_numerator - log_weights
```
proteneer, line 281: I probably need to offline this with you - but I'm slightly confused as to why `log_weights` is not dependent on `params`.

tests/test_reweighting.py
maxentile: >so the total number of samples is the same

Off by 2x. (endpoint test: 2 * 1e6, mixture test: 10 * 1e5)

>However, the atol=1e-2 is actually higher than the above test! This is counter-intuitive to me as I would have thought that use of intermediate windows should improve the efficiency of the estimator.

Initially I don't think this is surprising.

In the `endpoint_reweighting` case, we have `1e6` samples in lambda=0 state, and `1e6` samples in lambda=1 state.
In mixture case, we have at least `1e5` samples directly from lambda=0 state, but maybe we can borrow some effective samples from neighboring states. The total number of samples from all states is `1e6`. (So, we would only reach the same number of effective samples if all the other states were exactly equal to lambda=0 state... If the other states are different, we'll have somewhere in between `1e5` and `1e6` effective samples at lambda=0 state.)

If I instead fix the number of samples _per window_, then mixture reweighting should outperform endpoint reweighting. (Since the number of effective samples for estimating expectations at lambda=0 or lambda=1 would exceed the number of direct samples collected at lambda=0 or lambda=1.)

tests/test_reweighting.py
maxentile: will add a test of this

(logsumexp handles -inf gracefully, so this should be fine without additional trickery)

timemachine/fe/reweighting.py
maxentile: Previous feedback was to document sign flips / logs when switching between energies etc. <-> log probabilities etc. -- will remove the extraneous comments

timemachine/fe/reweighting.py
maxentile: Unsure which style is more readable.

Included both extremes (one-liner, vs. expanding all steps) -- maybe something in between is preferable, which only expands a couple steps.

tests/test_reweighting.py
proteneer: ah good to know!

timemachine/fe/reweighting.py
proteneer: Oops I see! Totally fine to leave it here then!

timemachine/fe/reweighting.py
proteneer: makes sense! Thanks

timemachine/fe/reweighting.py
maxentile: Should probably rename these arguments to be less confusing, perhaps to (`samples_ref_n`, `u_ref_n`), or similar?

https://github.com/proteneer/timemachine/blob/9f7065909e633ce8a25319d178cf7168de8d64ac/timemachine/fe/reweighting.py#L234-L237

It is an important feature of this approach that the reference state does not depend on params. We do not want to differentiate through whatever process is used to generate samples from the reference state.

When reweighting to new params, we will instead compute `w(x, params) = p_trial(x, params) / p_ref(x)`. The numerator depends on `params`, but the denominator does not.

maxentile: Name collision addressed in https://github.com/proteneer/timemachine/pull/654/commits/297f06100319b4749016aed21b459eb2024dcaa7 -- I think this is an improvement

tests/test_reweighting.py
maxentile: >For the gaussian case, I'd imagine that dDeltaF/dlog_sigma might require more precision than dDeltaF/dmean. 

I'd imagine so.

In a [documentation notebook](https://github.com/proteneer/timemachine/blob/c39f6dc765ffc8e7dfcdc1920daff02ca2df962e/examples/free_energy_gradients_comparison/comparison_on_1d_system.ipynb) of #404 , cell [46], note that estimates of the different gradient components have different variance.

maxentile: >How should we think about precision of dDeltaF/dParams both for the gaussian case, and in general?

Preferred interpretation: Exact gradient of a randomized approximation of `DeltaF(params)`.

`delta_f_approx_fxn = construct_approximation(ref_params, random_seed)`

>Presumably, an acceptable level of error is proportional to the internal learning rate one might take when optimizing the parameters

In the case that you created a fresh random approximation every optimization step

```python
delta_f_approx_fxn_0 = construct_approximation(params_0, seed_0)

params_1 = params_0 + stepsize * grad(delta_f_approx_fxn_0)(params_0)

delta_f_approx_fxn_1 = construct_approximation(params_1, seed_1)

params_2 = params_1 + stepsize * grad(delta_f_approx_fxn_1)(params_1)
...
```

then appropriate step size would be a function of both the geometry of `delta_f_exact_fxn` and the precision of the gradient estimate.

Instead, I think we would want to perform iterates like

```python
delta_f_approx_fxn_0 = construct_approximation(params_0, seed_0)

params_1 = local_optimize(fun=delta_f_approx_fxn_0, x0=params_0, ...)

delta_f_approx_fxn_1 = construct_approximation(params_1, seed_1)

params_2 = local_optimize(fun=delta_f_approx_fxn_1, x0=params_1, ...)
...
```

where `local_optimize` returns the local optimum (+ some regularization + optional restriction to some trust region, glossing over how that trust region is defined.)

Then whatever step sizes etc. are used inside `local_optimize` do NOT depend on the precision of the estimate, only on "the geometry of the problem."

The amount of progress you make in each outer loop step (construct_approximation, call local_optimize) would be influenced by how that trust region is defined.

(There's a similar way to view gradient descent: each step of SGD is exactly optimizing a random local approximation of f + some regularization.)

tests/test_reweighting.py
maxentile: Added in https://github.com/proteneer/timemachine/pull/654/commits/475a3996c0090ae9064436dde9ed8840d1816ecc

tests/test_reweighting.py
maxentile: Added in https://github.com/proteneer/timemachine/pull/654/commits/475a3996c0090ae9064436dde9ed8840d1816ecc

tests/test_reweighting.py
maxentile: Added in https://github.com/proteneer/timemachine/pull/654/commits/7c8b0f0925fc1ce45ac14bba130fc5f01fcf72c1

tests/test_reweighting.py
maxentile: Added in https://github.com/proteneer/timemachine/pull/654/commits/722a6d8eac7dfbe126b1f6be05e0015a13200cb7

tests/test_reweighting.py
maxentile: Sorry, missed replying to this part:

>so maybe rtol is more approriate here?

I don't think so.

`isclose(estimated_gradient, exact_gradient, rtol=...)` might be appropriate if the noise magnitude of each component `estimated_gradient[i]` was proportional to the magnitude of the corresponding gradient component `abs(exact_gradient[i])`.

(Counterexample: [1D Lennard-Jones testsystem](https://github.com/proteneer/timemachine/blob/c39f6dc765ffc8e7dfcdc1920daff02ca2df962e/examples/free_energy_gradients_comparison/comparison_on_1d_system.ipynb): exact gradient w.r.t. sigma, epsilon parameters ~= [-0.15, -0.3] (cell [43]), but estimates of these gradient components have marginal variance ~= [0.1, 0.01] (cell [41]).)

proteneer (approved):
This PR lgtm - especially after some in person discussions last week.

tests/test_reweighting.py
proteneer: I'm going to leave this "unresolved" so I can find the comments above easier. But this mostly addresses my concerns.

maxentile: Opened https://github.com/proteneer/timemachine/issues/667 for continued discussion after this PR

PR #655, opened 2022-02-26
Refactor memory management in nonbonded potentials

mcwitt:
- refactors memory management in nonbonded potentials using RAII style enabled by `DeviceBuffer`
- adds member functions to `DeviceBuffer` for async copies
- deletes `DeviceBuffer` copy constructor to prevent a potential pitfall of passing by value (demonstrated in e4c250d)

### DIFF
--- a/timemachine/cpp/src/device_buffer.cu
+++ b/timemachine/cpp/src/device_buffer.cu
@@ -1,17 +1,17 @@
 #include "device_buffer.hpp"
 #include "gpu_utils.cuh"
-#include <cstddef>
 
 namespace timemachine {
 
-template <typename T> T *allocate(const std::size_t length) {
-    T *buffer;
-    gpuErrchk(cudaMalloc(&buffer, length * sizeof(T)));
+void *allocate(const std::size_t size) {
+    void *buffer;
+    gpuErrchk(cudaMalloc(&buffer, size));
     return buffer;
 }
 
 template <typename T>
-DeviceBuffer<T>::DeviceBuffer(const std::size_t length) : size(length * sizeof(T)), data(allocate<T>(length)) {}
+DeviceBuffer<T>::DeviceBuffer(const std::size_t length)
+    : size(length * sizeof(T)), data(static_cast<T *>(allocate(size))) {}
 
 template <typename T> DeviceBuffer<T>::~DeviceBuffer() {
     // TODO: the file/line context reported by gpuErrchk on failure is
--- a/timemachine/cpp/src/device_buffer.cu
+++ b/timemachine/cpp/src/device_buffer.cu
@@ -20,14 +20,41 @@ template <typename T> DeviceBuffer<T>::~DeviceBuffer() {
     gpuErrchk(cudaFree(data));
 }
 
-template <typename T> void DeviceBuffer<T>::copy_from(const T *host_buffer) const {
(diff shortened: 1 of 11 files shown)


proteneer:
To clarify something in case you may have been wondering:

`gpuErrchk` is not a recoverable event. It's not something where we throw and try to recover, it will simply straight-up kill the program even though it technically leaks. This is why we are allowed to call `gpuErrchk` in the constructors (and in general) because RAII doesn't matter. If we threw an exception instead, we'd need to worry about callers catching the exception and not cleaning up allocated memory.

proteneer:
I like the idea of `DeviceBuffer`, but I'd maybe constrain its functionality a bit (these classes are extremely hard to refactor later on so usually we want to keep their methods as minimal as possible). The main benefits of the `DeviceBuffer` I see is simplifying the tedium of pairing `cudaMalloc` with `cudaMemcpy` and `cudaFree`, and being able to introduce `const` in most places. A common pattern in this PR is:

1) Initialize `DeviceBuffer` in the initializer list of some class
2) Set the contents of the `DeviceBuffer` later on. 

However, this runs into the issue where we can accidentally declare buffers with the wrong size  `K` and `memcpy` a different size `L` (and `gpuErrchk` certainly won't complain if `L` <= `K`, and may or may not complain if L < K).  To minimize the risk of this, you could add overloaded constructors to `DeviceBuffer` do both 1) and 2) at the same (especially with things like `lambda_plane_idxs`):


`template<T> DeviceBuffer(const std::vector<T> &host_data)` - this will invoke `cudaMalloc` with `cudaMemcpyHostToDevice`

 It would also let you declare `const` on members like `const DeviceBuffer<int> lambda_plane_idxs_` , which we couldn't do before since they were mutated in the `Nonbonded` constructors (so long as you add const and non-const getter pairs, eg. (`const T* DeviceBuffer::get_data() const). This would have lots of safety benefits later on and would improve the code quality significantly.

Though I'd only overload with parameters from C++ STL containers (eg `vector`/`set`/etc), and *not* raw pointers. Unfortunately with a signature like `template<T> DeviceBuffer(const T * ptr)`, we have no way of identifying if `ptr` points to host or device memory (and if it should invoke `cudaMemcpyHostToDevice` or `cudaMemcpyDeviceToDevice`).

mcwitt:
> If we threw an exception instead, we'd need to worry about callers catching the exception and not cleaning up allocated memory.

That makes sense, thanks for clarifying. So it seems like RAII is not helpful in the case of unrecoverable GPU errors, but still useful at least for the following?
1. avoiding the tedium and room for error in manually pairing allocations and deallocations
2. preventing leaks in cases where we do throw exceptions from functions that malloc

proteneer:
> That makes sense, thanks for clarifying. So it seems like RAII is not helpful in the case of unrecoverable GPU errors, but still useful at least for the following?

Yep!

proteneer:
>  The overloaded constructor approach you mentioned seems like it might be helpful for the latter. If we go with the latter, do you think it'd be reasonable to (eventually) change the Potential::execute_host interface to accept DeviceBuffers (or at least sized STL containers) instead of raw pointers?

`execute_host` can totally change - that's mostly a convenience function. The signature of `execute_device` probably can't be easily changed though.

mcwitt:
Closing for now since the PR has become stale and this isn't a current priority.

PR #656, opened 2022-02-28
Improve dummy atom and add detection of unstable core-dummy interactions.

proteneer:
This PR makes the following two major changes:

1) Remove the need to pass in an rdkit `mol` to most of the dummy atom code. Instead, only a bond list comprised of 2-tuples is needed. I'd debated about changing the API to instead take in a `networkx` graph but decided against it because the container for nodes, edges is extremely generic, and I wanted to make it clear to the caller what the minimal required inputs are (and it doesn't get simpler than a list of `int` pairs). This has the nice side benefit of removing redundant parameters and avoid inconsistencies in several `dg_ag` functions that required both a `mol` and `valence_idxs`. Instead, we now simply extract the `2-tuples` from `valence_idxs` that were already present (i.e. we can induce a `mol` from `valence_idxs`)

2) Adds a `flag_stable_dummy_ixns` function and associated tests to try and detect numerical instabilities that may arise in valence interactions that simultaneously involve both dummy atoms and core atoms. In particular, we require that all angles `i,j,k` have bond terms `i,j` and `j,k` defined, with non-zero equilibrium lengths and non-zero force constants. Similarly, all torsions `i,j,k,l` require `i,j,k` and `j,k,l` angles be defined, in addition to bond terms `i,j`, `j,k`, and `k,l` bond terms defined with similar considerations. Note that this function does not care about interactions that involve *only* dummy atoms, or interactions that *only* involve core atoms. 

This PR also adds two minor changes:

1) Improve coloring of bonds in drawing code
2) We now require that every atom has at least one bond.

### DIFF
--- a/tests/test_dummy.py
+++ b/tests/test_dummy.py
@@ -1,11 +1,13 @@
 import numpy as np
+import pytest
 from rdkit import Chem
 
 from timemachine.fe import dummy, dummy_draw
 from timemachine.fe.dummy import (
     enumerate_anchor_groups,
     enumerate_dummy_ixns,
-    flag_bonds,
+    flag_factorizable_bonds,
+    flag_stable_dummy_ixns,
     generate_optimal_dg_ag_pairs,
     identify_anchor_groups,
     identify_dummy_groups,
@@ -29,6 +31,14 @@ from timemachine.ff.handlers.deserialize import deserialize_handlers
 # partition - only applies to dummy groups, as we require that dummy groups disjointly partition dummy atoms.
 
 
+def get_bond_idxs(mol):
+    # not necessarily canonicalized!
+    idxs = []
+    for bond in mol.GetBonds():
+        idxs.append((bond.GetBeginAtomIdx(), bond.GetEndAtomIdx()))
+    return idxs
+
+
 def test_identify_root_anchors():
     """
     Test the identification of root anchors given a dummy atom.
@@ -42,17 +52,17 @@ def test_identify_root_anchors():
     its root anchors are the set of atoms {0, 1}
 
     """
-    mol = Chem.MolFromSmiles("C1CCC1N")
+    mol = get_bond_idxs(Chem.MolFromSmiles("C1CCC1N"))
     core = [0, 1, 2, 3]
     anchors = identify_root_anchors(mol, core, dummy_atom=4)
     assert set(anchors) == set([3])
 
-    mol = Chem.MolFromSmiles("C1CC2NC2C1")
+    mol = get_bond_idxs(Chem.MolFromSmiles("C1CC2NC2C1"))
     core = [0, 1, 2, 4, 5]
     anchors = identify_root_anchors(mol, core, dummy_atom=3)
     assert set(anchors) == set([2, 4])
 
-    mol = Chem.MolFromSmiles("C1OCC11CCCCC1")
+    mol = get_bond_idxs(Chem.MolFromSmiles("C1OCC11CCCCC1"))
     core = [3, 4, 5, 6, 7, 8]
     anchors = identify_root_anchors(mol, core, dummy_atom=0)
     assert set(anchors) == set([3])
@@ -61,7 +71,7 @@ def test_identify_root_anchors():
     anchors = identify_root_anchors(mol, core, dummy_atom=2)
     assert set(anchors) == set([3])
 
-    mol = Chem.MolFromSmiles("C1CC1.C1CCCCC1")
+    mol = get_bond_idxs(Chem.MolFromSmiles("C1CC1.C1CCCCC1"))
     core = [3, 4, 5, 6, 7, 8]
     anchors = identify_root_anchors(mol, core, dummy_atom=0)
     assert set(anchors) == set()
@@ -70,7 +80,7 @@ def test_identify_root_anchors():
     anchors = identify_root_anchors(mol, core, dummy_atom=2)
     assert set(anchors) == set()
 
-    mol = Chem.MolFromSmiles("C1CC2NC2C1")
+    mol = get_bond_idxs(Chem.MolFromSmiles("C1CC2NC2C1"))
     core = [0, 1, 2, 5]
     anchors = identify_root_anchors(mol, core, dummy_atom=3)
     assert set(anchors) == set([2, 5])
@@ -78,7 +88,7 @@ def test_identify_root_anchors():
     assert set(anchors) == set([2, 5])
 
     # cyclohexane with a nitrogen inside
-    mol = Chem.MolFromSmiles("C1C2CC3CC1N23")
+    mol = get_bond_idxs(Chem.MolFromSmiles("C1C2CC3CC1N23"))
     core = [0, 1, 2, 3, 4, 5]
     anchors = identify_root_anchors(mol, core, dummy_atom=6)
     assert set(anchors) == set([1, 3, 5])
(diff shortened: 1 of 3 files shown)


proteneer:
Taggin @wiederm and @pitmanme who may be interested in this.

timemachine/fe/dummy.py
```diff
@@ -91,19 +102,18 @@ def identify_root_anchors(mol, core, dummy_atom):
...
     core = set(core)
-
     assert dummy_atom not in core
 
-    # first convert to a dense graph
-    N = mol.GetNumAtoms()
+    # first convert to a dense graph, and assume that bond_idxs start from zero
+    flat_idxs = np.array(bond_idxs).flatten()
+    N = np.amax(flat_idxs) + 1
+    assert N == len(set(flat_idxs))
```
maxentile, line 113: nit: could be strengthened to 
```suggestion
    assert set(flat_idxs) == set(range(N))
```
or similar

(to rule out unexpected `-1 in flat_idxs` etc.)

timemachine/fe/dummy.py
```diff
@@ -581,3 +612,171 @@ def generate_optimal_dg_ag_pairs(mol, core, bond_idxs):
...
+                    b0_l > min_bond_l,
+                    b1_l > min_bond_l,
+                    b2_l > min_bond_l,
+                ]
+            ):
+                bonds_okay = True
+            else:
+                bonds_okay = False
+        else:
+            bonds_okay = False
```
maxentile, line 753: style preference -- compression

(e.g. `def bond_okay(b): return (b.k > min_bond_k) and (b.l > min_bond_l)`, `def bonds_okay(bonds): return np.all([bond_okay(b) for b in bonds])`)

maxentile (approved):
Good enhancements!

(Minor non-blocking comments)

timemachine/fe/dummy.py
mcwitt: nit: using Python builtin `all` might be preferable: `all(bond_okay(b) for b in bonds)`

proteneer: I'll clean this up.

timemachine/fe/dummy.py
```diff
@@ -441,45 +457,58 @@ def flag_bonds(mol, core, bond_idxs):
...
+    for b_idx, atom_idxs in enumerate(valence_idxs):
         if tuple(ordered_tuple(atom_idxs)) in allowed_ixns:
             keep_flags[b_idx] = 1
 
     return keep_flags
 
 
-def generate_dg_ag_pairs(mol, core, bond_idxs):
+class MissingBondError(Exception):
+    pass
```
mcwitt, line 475: The usage of this makes it seem like a more appropriate name might be something like `MissingAtomError` or `InvalidBondError`? If it doesn't make sense to include extra attributes (e.g. the invalid `i` or `j` value), would it make sense to just use `ValueError` with a message? (Could also see adding `i` and `j` fields, which could justify a custom error type)

timemachine/fe/dummy.py
```diff
@@ -581,3 +612,171 @@ def generate_optimal_dg_ag_pairs(mol, core, bond_idxs):
...
+    assert len(torsion_idxs) == len(torsion_params)
+
+    bond_kv = {}
+    for ij, params in zip(bond_idxs, bond_params):
+        bond_kv[ordered_tuple(ij)] = params
+
+    keep_angle_flags = []
+
+    def ixn_is_all_core_or_all_dummy(atom_idxs):
+        all_core = np.all([a in core for a in atom_idxs])
```
mcwitt, line 699: nit: would it make sense to use builtin `all` here with generator comprehension? E.g.
```suggestion
        all_core = all(a in core for a in atom_idxs)
```

timemachine/fe/dummy.py
```diff
@@ -581,3 +612,171 @@ def generate_optimal_dg_ag_pairs(mol, core, bond_idxs):
...
+        all_dummy = np.all([a not in core for a in atom_idxs])
+
+        return all_core or all_dummy
+
+    for (i, j, k), params in zip(angle_idxs, angle_params):
+        b0 = ordered_tuple((i, j))
```
mcwitt, line 705: The name `ordered_tuple` is a bit confusing to me, given what this function does (i.e. it sounds like it might sort the tuple, but it actually only looks at the first 2 elements and reverses the order if the first is larger than the second?). Would something like `canonicalize_ixn_tuple` or similar make sense?

mcwitt (approved):
Nice improvements! The transition to passing `bond_idx` instead of `Mol` objects makes it clearer to me what's important for the implementation. Had a few (mostly minor, stylistic) comments

timemachine/fe/dummy.py
maxentile: +1 for renaming to `canonicalize_` or similar

------

Tangent: Can we confirm that the implementation of `canonicalize_` should be the same for proper torsions and improper torsions?

In the force field handlers, `ReversibleBondHandler` applies a similar 2-fold symmetry to all valence terms except for impropers, which instead use this 3-fold symmetry:

https://github.com/proteneer/timemachine/blob/7c682225fd72deba70ad8003820061ca19dce63c/timemachine/ff/handlers/bonded.py#L211-L214

(If `canonicalize` should mirror this distinction, `canonicalize_` could be implemented as `min(permutations)` where `reversible_permutations(atom_idxs) = [atom_idxs, atom_idxs[::-1]]  # 2 permutations`, `improper_permutations(atom_idxs) = ...  # 3 permutations`. If `canonicalize` does not need to mirror this distinction, an inline comment is warranted.)

proteneer: 1.  Will rename to canonicalize - agreed that its better
2. Both propers and impropers get converted in to periodic torsions - which are then canonicalized. The requirement for canonicalization is that U(i,j,k,l)  = U(l,k,j,i). The 3-fold symmetry here is only for the generation of the 4-component terms. If anything, the improper hander should be sorting the generated trefoil terms with `sort_tuple`.

timemachine/fe/dummy.py
proteneer: I honestly haven't used python's `all` enough - I'm more familiar with `np.all` (esp when I need to specify axes etc, though it doesn't matter here.

tests/test_dummy.py
```diff
@@ -379,21 +390,21 @@ def test_flag_bonds():
     expected_flags = [x[0] for x in bond_pairs]
     bond_idxs = [x[1] for x in bond_pairs]
 
-    keep_flags = flag_bonds(mol, core, bond_idxs)
+    keep_flags = flag_factorizable_bonds(core, bond_idxs)
```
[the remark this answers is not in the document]
proteneer: Conciseness for the most part. It was much easier for me to write tests with `[0,1,0,0,0,0]` than `[False, True, False, False, False, False]`. I'd actually written this initially with True False values but it made the tests a lot harder to read.

timemachine/fe/dummy.py
```diff
@@ -91,19 +102,18 @@ def identify_root_anchors(mol, core, dummy_atom):
...
     assert len(set(core)) == len(core)
 
     core = set(core)
-
     assert dummy_atom not in core
 
-    # first convert to a dense graph
-    N = mol.GetNumAtoms()
+    # first convert to a dense graph, and assume that bond_idxs start from zero
+    flat_idxs = np.array(bond_idxs).flatten()
```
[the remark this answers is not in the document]
proteneer: The issue is in the `np.amax` right below this:

>>> np.amax([0,1,2])
2
>>> np.amax(set([0,1,2]))
{0, 1, 2}

(I have no idea why this is the case)

timemachine/fe/dummy.py
proteneer: Hm the name is correct, its indicating that a bond is missing (i.e. we have a disconnected atom). It's not clear *which* bond is missing unfortunately (since an arbitrary number of bonds may be possibe here).

proteneer: Also: `ValueError` is harder to test (since we'd need to match the resulting error string some how)

timemachine/fe/dummy.py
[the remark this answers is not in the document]
mcwitt: oh, that's bizarre -- could we just use the builtin `max` here instead since we're not dealing with numpy arrays?

timemachine/fe/dummy.py
```diff
@@ -91,19 +102,18 @@ def identify_root_anchors(mol, core, dummy_atom):
...
+    N = np.amax(flat_idxs) + 1
+    assert set(flat_idxs) == set(range(N))
     dense_graph = np.zeros((N, N), dtype=np.int32)
 
-    for bond in mol.GetBonds():
-        i, j = bond.GetBeginAtomIdx(), bond.GetEndAtomIdx()
+    for i, j in bond_idxs:
```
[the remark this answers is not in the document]
proteneer: tbh I don't understand the APIs on the page `https://networkx.org/documentation/stable/reference/algorithms/traversal.html` well enough right now to try and convert this, some of them many of the functionalities are for direct graphs. I'm trying to avoid too much logic change right now (and just make this a clean-up PR). Let me read up on this a bit more to understand how to use it?

timemachine/fe/dummy.py
mcwitt: > Hm the name is correct, its indicating that a bond is missing

Ah, sorry, I misunderstood. Looking again at `generate_dg_ag_ixns`, this makes sense.

Would it make sense to include a message and/or the missing indices as attributes on `MissingBondError`?

timemachine/fe/dummy.py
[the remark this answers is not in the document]
mcwitt: oh, sorry, I wasn't necessarily suggesting to switch to networkx for the implementation immediately; just that refactoring this slightly might make for a clearer separation between the bespoke parts of the algorithm vs. the more standard graph traversal implementation (which could be swapped out later)

timemachine/fe/dummy.py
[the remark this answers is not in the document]
proteneer: probably, but after some thought I personally find `flat_idxs = set(i for pair in bond_idxs for i in pair)` to be harder to read, and it embarrassingly took me a while to wrap my head around. nested list comprehensions are something I'm going to draw the line on.

timemachine/fe/dummy.py
proteneer: I think that's totally reasonable!

timemachine/fe/dummy.py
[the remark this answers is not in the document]
proteneer: After this some more thought (and attempting to refactor this into network graphs and using `nx.dfs_tree`), this is not doable as a standard `dfs` algorithm. The issue here is that if we induce a graph that's comprised of dummy atoms (and no core atoms), we have no way to actually "appending a core atom" to the anchor list, because it simply doesn't exist! No matter what dfs we do on the induced graph, it will only ever contain dummy atoms. There is no way to "terminate on first visit of a core atom".

timemachine/fe/dummy.py
proteneer: (adding this inline directly since I rebased)

timemachine/fe/dummy.py
proteneer: This has been modified to:

``` python
class MissingBondError(Exception):
    def __init__(self, bad_atom):
        message = "atom " + str(bad_atom) + " is disconnected."
        super().__init__(message)
```

timemachine/fe/dummy.py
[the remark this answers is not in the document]
proteneer: Resolved after offline discussion.

timemachine/fe/dummy.py
proteneer: #658 address this more broadly.

proteneer:
Going to merge this and deal with #658 later.

PR #658, opened 2022-02-28
Improve bond canonicalization consistency in Improper Torsions

proteneer:
Canonicalize improper torsions (after a discussion with @maxentile). Also rename `sort_tuple` to `canonicalize_bond`. Depending if this or #656 gets merged first, canonicalize bonds will be moved to a common_utils.py file.

### DIFF
--- a/timemachine/ff/handlers/utils.py
+++ b/timemachine/ff/handlers/utils.py
@@ -1,8 +1,23 @@
 from rdkit import Chem
 
 
-def sort_tuple(arr):
+def canonicalize_bond(arr):
+    """
+    Canonicalize a bonded interaction. If arr[0] < arr[-1] then arr is
+    returned, else if arr[0] > arr[-1], then arr[::-1] is returned. If
+    arr[0] == arr[-1] then an exception is thrown.
 
+    Parameters
+    ----------
+    arr: list of int
+        Bond indices.
+
+    Returns
+    -------
+    arr: list of int
+        Canonicalized bond indices.
+
+    """
     container_type = type(arr)
 
     if len(arr) == 0:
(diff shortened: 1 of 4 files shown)


maxentile (approved):

timemachine/ff/handlers/utils.py
```diff
@@ -11,6 +26,8 @@ def sort_tuple(arr):
         return arr
     elif arr[0] > arr[-1]:
         return container_type(reversed(arr))
+    elif arr[0] == arr[-1]:
+        raise Exception("arr[0] == [-1]")
```
mcwitt, line 30: nit: maybe something like
```suggestion
        raise ValueError("Invalid bond with first and last indices equal")
```

mcwitt (approved):

timemachine/ff/handlers/utils.py
proteneer: committed inline - thanks

timemachine/cpp/src/kernels/k_periodic_torsion.cuh
```diff
@@ -2,16 +2,16 @@
...
 
 template <typename RealType>
 inline __device__ void cross_product(const RealType a[3], const RealType b[3], RealType c[3]) {
-    // fix this one indexed garbage later
-    c[1 - 1] = a[2 - 1] * b[3 - 1] - a[3 - 1] * b[2 - 1];
-    c[2 - 1] = a[3 - 1] * b[1 - 1] - a[1 - 1] * b[3 - 1];
-    c[3 - 1] = a[1 - 1] * b[2 - 1] - a[2 - 1] * b[1 - 1];
+    // these extra __dmul_rn calls are needed to preserve bitwise anticommutativity
+    // i.e. cross(a,b) is bitwise identical to -cross(b,a) except in the sign-bit
+    c[0] = __dmul_rn(a[1], b[2]) - __dmul_rn(a[2], b[1]);
```
maxentile, line 12: Should this always be the double-precision intrinsic `__dmul_rn` regardless of `RealType`?

proteneer: great catch! d'oh.

PR #659, opened 2022-03-01
Make memory tests more explicit using pytest marks

badisa:


badisa:
This reduces the time it takes for the memory tests by 10 minutes and increases the time of the unittests by 4 minutes. 

I have tried to mark the tests that cover our C++/Cuda code, barring some of the more expensive ones (such as this https://github.com/proteneer/timemachine/blob/master/slow_tests/test_barostat.py#L243).

tests/conftest.py
```diff
@@ -9,13 +9,16 @@
...
-def reset_cuda_device_after_test():
+def reset_cuda_device_after_test(request):
     """Calls cudaDeviceReset() after each test.
 
     This is needed for 'cuda-memcheck --leak-check full' to catch leaks"""
 
     yield
 
+    # If the test is not marked for memory tests, no need to reset device
+    if request.node.get_closest_marker("memcheck") is None:
```
badisa, line 20: This saved about 30 seconds for `make unit_tests` locally

tests/nonbonded/test_nonbonded_all_pairs.py
```diff
@@ -8,6 +8,8 @@
 from timemachine.lib.potentials import NonbondedAllPairs, NonbondedAllPairsInterpolated
 from timemachine.potentials import nonbonded
 
+pytestmark = [pytest.mark.memcheck]
```
badisa, line 11: The syntax for marking a test module

pytest.ini
```diff
@@ -2,4 +2,5 @@
 addopts = -svx
 testpaths =
     tests/
-    slow_tests/
+markers =
+    memcheck: marks tests to be run with cuda memory checks, triggers cuda device reset at end of marked tests (deselect with '-m "not memcheck"')
```
mcwitt, line 6: not directly related to the PR, and not sure if really desirable, but just noting that it looks like pytest config can be [put in pyproject.toml](https://docs.pytest.org/en/6.2.x/customize.html#pyproject-toml)

Makefile
```diff
@@ -25,11 +27,11 @@ verify:
 
 .PHONY: memcheck_tests
 memcheck_tests:
-	cuda-memcheck --leak-check full --error-exitcode 1 pytest $(PYTEST_CI_ARGS) tests/
+	compute-sanitizer --leak-check full --error-exitcode 1 pytest -m $(MEMCHECK_MARKER) $(PYTEST_CI_ARGS)
```
mcwitt, line 30: would it be worth introducing and reverting a memory leak in this branch to make sure we're still catching them?

tests/conftest.py
mcwitt: nit: [this method](https://stackoverflow.com/a/38763328/1427284) looks a little cleaner to my eye (assuming it's actually equivalent)

mcwitt (approved):
Looks good!

maxentile (approved):
Nice -- this should simplify things.

Reviewed the list of test files, wanted to double-check if any of the following should also have the mark of the memcheck:
* `test_lambda_potential.py`
* `test_rmsd_align.py`
* `test_shape.py`

(since they were previously run with memcheck, and they exercise custom ops without running any MD)

Makefile
badisa: Good catch, looks like this isn't working in gitlab at the moment

A bit more information here:
https://github.com/NVIDIA/nvidia-docker/issues/454

badisa:
> Nice -- this should simplify things.
> 
> Reviewed the list of test files, wanted to double-check if any of the following should also have the mark of the memcheck:
> 
> * `test_lambda_potential.py`
> * `test_rmsd_align.py`
> * `test_shape.py`
> 
> (since they were previously run with memcheck, and they exercise custom ops without running any MD)

`test_lambda_potential.py` currently only has one test that is skipped. 
`test_rmsd_align.py` tests pure C++ code, the memory leaks we are checking for are only for Cuda so from my understanding it won't catch memory leaks/accesses there
`test_shape.py` currently only tests numpy code, the custom op test is disabled (The Cuda Shape was removed here: https://github.com/proteneer/timemachine/commit/87ddca0eb9352c640842bd6929b88d8397b6651b)

maxentile:
Thanks -- I had overlooked all of that!

badisa:
I had introduced a bug in [d9d618f](https://github.com/proteneer/timemachine/pull/659/commits/d9d618ffd4656c6d130a141b89c7614235696117) which was the cause of the CI issues. 

You can see the failures under cuda memcheck and compute sanitizer in the respective commits: 
[8d14e8f](https://github.com/proteneer/timemachine/pull/659/commits/8d14e8fc19e87ae7116ae99b0a632be03a6a7e0a)
[df70756](https://github.com/proteneer/timemachine/pull/659/commits/df70756fee5fc8ec4a0dc8c0a20ebe995be2d418)

mcwitt (approved):

tests/conftest.py
```diff
@@ -9,13 +9,16 @@
...
+def reset_cuda_device_after_test(request):
     """Calls cudaDeviceReset() after each test.
 
     This is needed for 'cuda-memcheck --leak-check full' to catch leaks"""
 
     yield
 
+    # If the test is not marked for memory tests, no need to reset device
+    if "memcheck" not in request.keywords:
+        return
```
maxentile, line 21: Could avoid double negative

```python
# If the test is not marked for memory tests, no need to reset device
if "memcheck" not in request.keywords:
    return
# ensure that destructors are called before cudaDeviceReset()
gc.collect()

custom_ops.cuda_device_reset()
```

-->

```python
# Device reset only needed during memory tests
if "memcheck" in request.keywords: 
    # ensure that destructors are called before cudaDeviceReset()
    gc.collect()
    custom_ops.cuda_device_reset()
```

or similar

maxentile (approved):
Ahh, sorry, I had missed the `not in` -> `in` issue in first review!

Looks good. Simplifies coverage reports etc. to have all the tests in the same directory, removes a source of name confusion, and allows more fine-grained selection in the future. 👍

proteneer (approved):
nice addition.

Issue #472, opened 2021-07-20
Implement ability to run Potentials over subset of System


badisa:
## Intent

Right now to perform MBAR we have to compute energies over the entire system when the only portion of the system that actually changes is the lambda. This means building the full neighborlist and performing more work. We would like to compute only the energy related to the ligand (or any arbitrary subset) atoms. This will allow for a much faster evaluation of energies and make MBAR (as well as other tasks) cheaper.

## Implementation

Modify the neighborlist code to be non-sequential, creating tiles as before but using a mapping array. This will allow us to reuse the non-bonded code otherwise as normally used, and still get performance improvements.

maxentile:
A bit more context for this (but since this comment expands scope, perhaps I should move it to a new issue?).

I think we would like to profile and improve performance for a few common workloads:

Loops
1. `u_k = [u(x, lam, params) for lam in lams]`, where `lams` are a collection of ~10-100 lambda values
2. `u_n = [u(x, lam, params) for x in xs]` , where `xs` are a collection of ~100-10000 snapshots
3. `u_n = [u(x, lam, params) for x in few_particle_diff_xs]` , where `few_particle_diff_xs` are a collection of ~100-10000 snapshots that are identical except for the positions of a small number of particles

Nested loops
4. `u_nk = [[u(x, lam, params) for lam in lams] for x in xs]`

Currently, performing this sort of loop using Python calls to timemachine custom ops takes about ~10x longer than we might expect (taking ~2 milliseconds per iteration, when we might expect "light speed" to be closer to ~200 microseconds per iteration).

We also expect to be able to go much faster than "light speed" in cases where we only need to compute energy differences `delta_u = u(x, lam, params) - u(x0, lam0, params0)`, in cases where only a very small fraction of the pairwise interactions contribute to the difference. In some cases, Jax's vmap transformation can generate reasonably fast batched code for `delta_u` and its derivatives, even from a pair-list based numpy implementation like the one added in https://github.com/proteneer/timemachine/pull/453 .

@proteneer points out that the fast CUDA code in timemachine to compute exclusions accepts a pair list in essentially the same format: https://github.com/proteneer/timemachine/blob/e9c528baac80150b70e2a3240d9ea8e1f41a335d/timemachine/cpp/src/kernels/k_nonbonded.cuh#L813-L822 

It may be possible to go even faster than `k_nonbonded_exclusions` using an approach like the one @badisa described above.

[...]

PR #660, opened 2022-03-02
Allow computing all-pairs potential on subset of atoms, add consistency check

mcwitt:
Related: #472
- Adds a constructor argument `atom_idxs` to `NonbondedAllPairs`; this is used to select a subset of atoms for computing the all-pairs potential
- Adds consistency check comparing the result of the full `Nonbonded(exclusions, scales)` potential with the sum of
    - `NonbondedAllPairs(host)`
    - `NonbondedAllPairs(ligand)`
    - `NonbondedInteractionGroup(host, ligand)`
    - `NonbondedPairList(exclusions, scales)`
---
#### Notes for review
- Commits starting with `Rename` are purely string substitutions and should contain no other changes -- these were mainly to reflect that we're no longer just doing permutations, since the potential can be computed on a subset of atoms
- Most of the implementation changes to allow computing on a subset have been squashed into 71cc0e5

### DIFF
--- a/tests/nonbonded/test_consistency.py
+++ b/tests/nonbonded/test_consistency.py
@@ -1,0 +1,144 @@
+from typing import Iterable, Tuple
+
+import numpy as np
+import pytest
+from parameter_interpolation import gen_params
+
+from timemachine.lib.potentials import (
+    FanoutSummedPotential,
+    Nonbonded,
+    NonbondedAllPairs,
+    NonbondedAllPairsInterpolated,
+    NonbondedInteractionGroup,
+    NonbondedInteractionGroupInterpolated,
+    NonbondedInterpolated,
+    NonbondedPairListNegated,
+    NonbondedPairListNegatedInterpolated,
+)
+
+
+def filter_valid_exclusions(
+    num_atoms: int, exclusions: Iterable[Tuple[int, int]], scales: Iterable[Tuple[float, float]]
+) -> Tuple[np.ndarray, np.ndarray]:
+    filtered_pairs = (((i, j), scales) for (i, j), scales in zip(exclusions, scales) if i < num_atoms and j < num_atoms)
+    idxs, scales = zip(*filtered_pairs)
+    return np.array(idxs, dtype=np.int32), np.array(scales)
+
+
+@pytest.mark.parametrize("lamb", [0.0, 0.1])
+@pytest.mark.parametrize("beta", [2.0])
+@pytest.mark.parametrize("cutoff", [1.1])
+@pytest.mark.parametrize("precision", [np.float64, np.float32])
+@pytest.mark.parametrize("num_atoms,num_atoms_ligand", [(33, 1), (4080, 1050)])
+@pytest.mark.parametrize("interpolated", [False, True])
+@pytest.mark.parametrize("disable_hilbert_sort", [False, True])
+def test_nonbonded_consistency(
+    disable_hilbert_sort,
+    interpolated,
+    num_atoms_ligand,
+    num_atoms,
+    precision,
+    cutoff,
+    beta,
+    lamb,
+    example_nonbonded_params,
+    example_nonbonded_exclusion_idxs,
+    example_nonbonded_exclusion_scales,
+    example_conf,
+    example_box,
+    rng: np.random.Generator,
+):
+    conf = example_conf[:num_atoms, :]
+    params_initial = example_nonbonded_params[:num_atoms, :]
+    params = gen_params(params_initial, rng) if interpolated else params_initial
+
+    exclusion_idxs, exclusion_scales = filter_valid_exclusions(
+        num_atoms, example_nonbonded_exclusion_idxs, example_nonbonded_exclusion_scales
+    )
+
+    lambda_plane_idxs = rng.integers(-2, 3, size=(num_atoms,), dtype=np.int32)
+    lambda_offset_idxs = rng.integers(-2, 3, size=(num_atoms,), dtype=np.int32)
+
+    ligand_idxs = rng.choice(num_atoms, size=(num_atoms_ligand,), replace=False).astype(np.int32)
+    host_idxs = np.setdiff1d(np.arange(num_atoms), ligand_idxs).astype(np.int32)
+
+    make_ref_potential = NonbondedInterpolated if interpolated else Nonbonded
+    ref_impl = make_ref_potential(
+        exclusion_idxs, exclusion_scales, lambda_plane_idxs, lambda_offset_idxs, beta, cutoff
+    ).unbound_impl(precision)
+
+    if disable_hilbert_sort:
+        ref_impl.disable_hilbert_sort()
+
+    def make_allpairs_potential(atom_idxs):
+        make_potential = NonbondedAllPairsInterpolated if interpolated else NonbondedAllPairs
+        return make_potential(
+            lambda_plane_idxs,
+            lambda_offset_idxs,
+            beta,
+            cutoff,
+            atom_idxs,
+            disable_hilbert_sort=disable_hilbert_sort,
+        )
+
+    def make_ixngroup_potential(ligand_idxs):
+        make_potential = NonbondedInteractionGroupInterpolated if interpolated else NonbondedInteractionGroup
+        return make_potential(
+            ligand_idxs,
+            lambda_plane_idxs,
+            lambda_offset_idxs,
+            beta,
+            cutoff,
+            disable_hilbert_sort=disable_hilbert_sort,
+        )
+
+    def make_pairlist_potential(exclusion_idxs, exclusion_scales):
+        make_potential = NonbondedPairListNegatedInterpolated if interpolated else NonbondedPairListNegated
+        return make_potential(exclusion_idxs, exclusion_scales, lambda_plane_idxs, lambda_offset_idxs, beta, cutoff)
+
+    test_impl = FanoutSummedPotential(
+        [
+            make_allpairs_potential(host_idxs),
+            make_allpairs_potential(ligand_idxs),
+            make_ixngroup_potential(ligand_idxs),
+            make_pairlist_potential(exclusion_idxs, exclusion_scales),
+        ]
+    ).unbound_impl(precision)
+
+    du_dx_ref, du_dp_ref, du_dl_ref, u_ref = ref_impl.execute(conf, params, example_box, lamb)
+    du_dx_test, du_dp_test, du_dl_test, u_test = test_impl.execute(conf, params, example_box, lamb)
+
+    np.testing.assert_array_equal(du_dx_test, du_dx_ref)
+
+    if interpolated:
+        # NOTE: bitwise equivalence is not currently possible for the
+        # interpolated case. To see this, note that the interpolated
+        # energy is given by
+        #
+        #   u(p0, p1) = (1 - lam) * F(p0) + lam * F(p1)
+        #
+        # In particular,
+        #
+        #   du_dp1 = lam * f(p1)
+        #
+        # where f(p) = F'(p). The reference potential effectively sums
+        # over interactions before multiplication by \lambda
+        #
+        #   du_dp1_ref = lam * fixed_sum(f_host(p1), f_ligand(p1), f_host_ligand(p1))
+        #
+        # while the test potential (because it's implemented using
+        # SummedPotential), effectively distributes multiplication by
+        # \lambda into the sum
+        #
+        #   du_dp1_test = fixed_sum(lam * f_host(p1), lam * f_ligand(p1), lam * f_host_ligand(p1))
+        #
+        # Since `c * fixed_sum(x, y)` is not bitwise equivalent to
+        # `fixed_sum(c * x, c * y)` in general, the reference and test
+        # du_dps are not guaranteed to be bitwise equivalent in the
+        # interpolated case.
+        np.testing.assert_allclose(du_dp_test, du_dp_ref, rtol=1e-8, atol=1e-8)
+    else:
+        np.testing.assert_array_equal(du_dp_test, du_dp_ref)
+
+    np.testing.assert_array_equal(du_dl_test, du_dl_ref)
+    assert u_test == u_ref
--- a/timemachine/cpp/src/kernels/k_lambda_transformer_jit.cuh
+++ b/timemachine/cpp/src/kernels/k_lambda_transformer_jit.cuh
@@ -58,7 +58,8 @@ void __global__ k_gather_interpolated(
     const double lambda,
     const int N,
     const unsigned int *__restrict__ idxs,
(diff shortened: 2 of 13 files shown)


timemachine/cpp/src/nonbonded_all_pairs.cu
```diff
@@ -182,9 +201,15 @@ void NonbondedAllPairs<RealType, Interpolated>::hilbert_sort(
...
+    const int B = (K_ + tpb - 1) / tpb;
 
-    k_coords_to_kv<<<B, tpb, 0, stream>>>(N_, d_coords, d_box, d_bin_to_idx_, d_sort_keys_in_, d_sort_vals_in_);
+    if (d_atom_idxs_) {
+        k_coords_to_kv_gather<<<B, tpb, 0, stream>>>(
+            K_, d_atom_idxs_, d_coords, d_box, d_bin_to_idx_, d_sort_keys_in_, d_sort_vals_in_);
+    } else {
+        // N_ == K_
+        k_coords_to_kv<<<B, tpb, 0, stream>>>(K_, d_coords, d_box, d_bin_to_idx_, d_sort_keys_in_, d_sort_vals_in_);
+    }
```
mcwitt, line 212: We might be able to eliminate the conditional branch here by setting `d_atom_idxs_` to the range `1..N` and always using `k_coords_to_kv_gather` (will want to benchmark this)

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -85,6 +85,53 @@ void __global__ k_check_rebuild_coords_and_box(
     }
 }
 
+// TODO: DRY with k_check_rebuild_coords_and_box
```
mcwitt, line 88: TODO: fix or add to #639

timemachine/cpp/src/nonbonded_all_pairs.cu
```diff
@@ -21,45 +23,57 @@
...
+                                             &k_nonbonded_unified<RealType, 0, 1, 1, 0>,
+                                             &k_nonbonded_unified<RealType, 0, 1, 1, 1>,
+                                             &k_nonbonded_unified<RealType, 1, 0, 0, 0>,
+                                             &k_nonbonded_unified<RealType, 1, 0, 0, 1>,
+                                             &k_nonbonded_unified<RealType, 1, 0, 1, 0>,
+                                             &k_nonbonded_unified<RealType, 1, 0, 1, 1>,
+                                             &k_nonbonded_unified<RealType, 1, 1, 0, 0>,
+                                             &k_nonbonded_unified<RealType, 1, 1, 0, 1>,
+                                             &k_nonbonded_unified<RealType, 1, 1, 1, 0>,
+                                             &k_nonbonded_unified<RealType, 1, 1, 1, 1>}),
```
mcwitt, line 73: side note: these huge diffs from minor changes to the initializer list are pretty annoying; I'm looking forward to eventually upgrading to `clang-format` 14 which has the [PackConstructorInitializers](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#:~:text=define%20BAR%0A%23endif-,packconstructorinitializers,-([PackConstructorInitializers](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#:~:text=define%20BAR%0A%23endif-,packconstructorinitializers,-(PackConstructorInitializersStyle)%20clang-format)Style)%20clang-format) setting to better control this

timemachine/cpp/src/nonbonded_all_pairs.hpp
```diff
@@ -31,42 +33,41 @@ template <typename RealType, bool Interpolated> class NonbondedAllPairs : public
...
-    double *d_sorted_w_; // sorted 4D coordinates
-    double *d_sorted_dw_dl_;
-    double *d_sorted_p_;   // sorted parameters
-    double *d_unsorted_p_; // unsorted parameters
-    double *d_sorted_dp_dl_;
-    double *d_unsorted_dp_dl_;
-    unsigned long long *d_sorted_du_dx_;
-    unsigned long long *d_sorted_du_dp_;
+    unsigned int *d_sorted_atom_idxs_; // [K_] indices of interacting atoms, sorted by hilbert curve index
+    double *d_gathered_x_;             // sorted coordinates
```
mcwitt, line 64: TODO: update some comments referring to "sorted" arrays

mcwitt: done in 518c76a

tests/nonbonded/test_consistency.py
```diff
@@ -0,0 +1,121 @@
...
+    ).unbound_impl(precision)
+
+    du_dx_ref, du_dp_ref, du_dl_ref, u_ref = ref_impl.execute(conf, params, example_box, lamb)
+    du_dx_test, du_dp_test, du_dl_test, u_test = test_impl.execute(conf, params, example_box, lamb)
+
+    np.testing.assert_array_equal(du_dx_test, du_dx_ref)
+
+    if interpolated:
+        # TODO: why does interpolation break bitwise equivalence?
+        np.testing.assert_allclose(du_dp_test, du_dp_ref, rtol=1e-10, atol=1e-10)
```
mcwitt, line 116: Not completely sure why this is necessary yet. At least all the other outputs seem bitwise equivalent (and `du_dp` for the non-interpolated case)

timemachine/cpp/src/nonbonded_all_pairs.cu
```diff
@@ -270,13 +272,15 @@ void NonbondedAllPairs<RealType, Interpolated>::execute_device(
     // identify which tiles contain interpolated parameters
 
     const int tpb = 32;
-    const int B = (N + tpb - 1) / tpb;
-
-    dim3 dimGrid(B, 3, 1);
```
mcwitt, line 256: Since there are now 2 different shapes used here (e.g. `dim3(ceil_divide(N, tpb), 3, 1)` vs. `dim3(ceil_divide(K_, tpb), 3,  1`, I opted to just define the shape parameter directly at the call site (I wrote a couple frustrating bugs along the way do to confusing the 2 shapes)

timemachine/cpp/src/nonbonded_all_pairs.cu
```diff
@@ -182,9 +197,15 @@ void NonbondedAllPairs<RealType, Interpolated>::hilbert_sort(
     const double *d_coords, const double *d_box, cudaStream_t stream) {
 
     const int tpb = 32;
-    const int B = (N_ + tpb - 1) / tpb;
+    const int B = (K_ + tpb - 1) / tpb;
```
badisa, line 200: Use `ceil_divide` here for consistency

timemachine/cpp/src/nonbonded_all_pairs.hpp
```diff
@@ -29,44 +31,44 @@ typedef void (*k_nonbonded_fn)(
 template <typename RealType, bool Interpolated> class NonbondedAllPairs : public Potential {
 
 private:
-    std::array<k_nonbonded_fn, 16> kernel_ptrs_;
+    const int N_;
```
badisa, line 34: Would be good to have a comment here to distinguish it from `K_`

timemachine/cpp/src/kernels/k_lambda_transformer_jit.cuh
```diff
@@ -54,13 +54,14 @@ void __global__ k_compute_w_coords(
...
-void __global__ k_permute_interpolated(
+void __global__ k_gather_interpolated(
     const double lambda,
     const int N,
-    const unsigned int *__restrict__ perm,
-    const double *__restrict__ d_p,
-    double *__restrict__ d_sorted_p,
-    double *__restrict__ d_sorted_dp_dl) {
+    const unsigned int *__restrict__ idxs,
+    const double *__restrict__ d_p0, // initial parameters
```
badisa, line 61: Would be good to indicate that d_p0 and d_p1 must have the same shape

timemachine/cpp/src/nonbonded_all_pairs.cu
badisa: It seems like a docstring explaining the difference between K and N would be handy for returning to this code later.

timemachine/cpp/src/nonbonded_all_pairs.cu
```diff
@@ -376,26 +407,31 @@ void NonbondedAllPairs<RealType, Interpolated>::execute_device(
 
     // coords are N,3
     if (d_du_dx) {
-        k_inv_permute_accum<<<dimGrid, tpb, 0, stream>>>(N, d_perm_, d_sorted_du_dx_, d_du_dx);
+        k_scatter_accum<<<dim3(ceil_divide(K_, tpb), 3, 1), tpb, 0, stream>>>(
```
badisa, line 410: If I understand the `scatter gather` paradigm, scatter and accumulate are antonyms?

mcwitt: Oh, maybe this is confusing.. my intent was to generalize the naming when going from the existing K=N case to the more general K<=N as:
- permute -> scatter
- inverse permute -> gather

This is intended to be independent of whether we accumulate or assign to the result array. Does that make sense?

badisa: I am more familiar with the `fork and join` paradigm and it seemed like fork and scatter are equivalent and gather and join are equivalent. In which case this seemed backwards? I might just not understand the paradigm, was unexpected to see a scatter (or what I interpreted as a broadcast/fork) to reduce.  

Doesn't matter as long as it is a consistent convention

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -164,14 +211,14 @@ void __global__ k_inv_permute_accum(
         return;
     }
 
-    array[perm[idx] * stride + stride_idx] += sorted_array[idx * stride + stride_idx];
+    array[idxs[idx] * stride + stride_idx] += gathered_array[idx * stride + stride_idx];
```
maxentile, line 214: A user of this kernel could accidentally cause a race condition by including repeated elements in `idxs` (previously `idxs` was a permutation, but now `idxs` could be more general).

Could we add a line comment that `idxs` is assumed to contain unique indices, and explicitly assert uniqueness near where `idxs` is defined?

timemachine/cpp/src/nonbonded_all_pairs.cu
mcwitt: Changed in af3b039

timemachine/cpp/src/nonbonded_all_pairs.cu
maxentile: >The picture I had for "scatter" was https://pytorch-scatter.readthedocs.io/en/latest/functions/scatter.html .

Err, posted that before I saw @mcwitt 's link https://github.com/proteneer/timemachine/pull/660#discussion_r826247997 , didn't mean to override that. (I think the only difference is the wiki definition and the current function assume no repeated scatter idxs, while the pytorch_scatter reduces over any repeated scatter idxs. For the current use, only unique idxs make sense.)

> I might just not understand the paradigm, was unexpected to see a scatter (or what I interpreted as a broadcast/fork) to reduce.

I had misread the `y[idxs[i]] += x[i]` as a reduction, but @mcwitt clarified offline that it's always a single addition to whatever was in the output array before.

timemachine/cpp/src/nonbonded_all_pairs.hpp
mcwitt: Good call, added in db146c6

timemachine/cpp/src/kernels/k_nonbonded.cuh
mcwitt: @maxentile Good call. I renamed `idxs` -> `unique_idxs` and added line comments in 77508be.

> explicitly assert uniqueness near where idxs is defined?

Hmm, I think it is a bit tricky to do this right at the call site, since `d_atom_idxs` is already in device memory. We do already require `atom_idxs` to be a set in the [constructor argument](https://github.com/proteneer/timemachine/blob/77508bef811eb25a529a758b16d3a1f005b09411/timemachine/cpp/src/nonbonded_all_pairs.cu#L32), so this may be sufficient? What do you think?

timemachine/cpp/src/kernels/k_lambda_transformer_jit.cuh
mcwitt: Thanks, added in f882a97

timemachine/cpp/src/kernels/k_nonbonded.cuh
maxentile: Looks good!

>Hmm, I think it is a bit tricky to do this right at the call site, since d_atom_idxs is already in device memory.

Ahh, good point

>We do already require atom_idxs to be a set in the [constructor argument](https://github.com/proteneer/timemachine/blob/77508bef811eb25a529a758b16d3a1f005b09411/timemachine/cpp/src/nonbonded_all_pairs.cu#L32), so this may be sufficient? What do you think?

I think that is sufficient!

timemachine/cpp/src/nonbonded_all_pairs.cu
mcwitt: > It seems like a docstring explaining the difference between K and N

@badisa I think this is addressed by db146c6 (but let me know if you had something else in mind)

timemachine/cpp/src/kernels/k_lambda_transformer_jit.cuh
badisa: Its a bit confusing since P is not defined, but does the job

timemachine/cpp/src/nonbonded_all_pairs.cu
badisa: Looks good

tests/nonbonded/test_consistency.py
mcwitt: Resolved offline in conversation with @proteneer. Bitwise equivalence is not possible with the current structure of the code, ultimately due to the fact that the distributive property of multiplication breaks bitwise equivalence (even when summation is done in fixed point), i.e. `c * fixed_sum(a, b) != fixed_sum(c * a, c * b)` (see example in [gist](https://gist.github.com/mcwitt/904adca81a9a67dd707c941fdbf6a1af)).

Added a note in 15a17b7.

timemachine/cpp/src/nonbonded_all_pairs.cu
mcwitt: Thinking to leave this as-is for now to avoid any performance regression in the K=N case. Added a note to #639 to revisit this as part of clean / DRY tasks.

maxentile (approved):
Looks good to me!

proteneer (approved):
lgtm!

PR #662, opened 2022-03-03
Enable colorized pytest output in CI

mcwitt:


proteneer:
this looks like an easy one - out of curiosity does it jumble the copy and paste output?

mcwitt:
> out of curiosity does it jumble the copy and paste output?

I don't see any issues (e.g. color codes) when copy-pasting the output on my end

badisa (approved):

maxentile (approved):
🎨  Easier to skim now!

Issue #663, opened 2022-03-03
Unexpected neighborlist behaviors when (n_atoms - 1) % 32 == 0


maxentile:
Recently noticed two small oddities when interacting with the neighborlist from Python. Cause currently unclear.

Possibly related to size-dependent test failures @mcwitt has recently observed.

-----------

1. On some inputs, the neighborlist uses block bounds where one bounding box has extent == (0,0,0).

Randomly sampling (conf, box) input pairs with random # atoms suffices to find a few examples of this behavior:

```python
from timemachine.lib import custom_ops
import numpy as np

def num_zero_extent_blocks(conf, box):
    nblist = custom_ops.Neighborlist_f64(len(conf))
    bb_centers, bb_extents = nblist.compute_block_bounds(conf, box, 32)
    return sum((bb_extents == 0).all(1))

np.random.seed(0)

inputs = []
results = []

for _ in range(1000):
    n_atoms = np.random.randint(100, 3000)
    scale = np.random.rand() + 3
    
    conf = scale * np.random.randn(n_atoms, 3)
    box = scale * np.eye(3)
    
    n = num_zero_extent_blocks(conf, box)
    if n > 0:
        print(f'found an instance that produces zero-extent blocks!\n(# atoms = {n_atoms}, # zero-extent blocks = {n})')
        inputs.append((conf, box))
        results.append(n)

print(len(results))  # 42
```

Looking at the number of atoms in each of these, I noticed that they all have in common `(n_atoms - 1) % 32 == 0`.

-----------

2. Segfault

Modifying the loop to _only_ try inputs where `(n_atoms - 1) % 32 == 0` reliably triggers segfaults after a random number of iterations.

(Note: Segfaults are _not_ triggered by this loop if `n_atoms = 32 * x + 1` is changed to `n_atoms = 32 * x + 2`, for example.)

```python
from timemachine.lib import custom_ops
import numpy as np

def num_zero_extent_blocks(conf, box):
    nblist = custom_ops.Neighborlist_f64(len(conf))
    bb_centers, bb_extents = nblist.compute_block_bounds(conf, box, 32)
    return sum((bb_extents == 0).all(1))

np.random.seed(0)

inputs = []
results = []

for t in range(1000):
    x = np.random.randint(10, 50)
    n_atoms = 32 * x + 1
    scale = np.random.rand() + 3
    
    conf = scale * np.random.randn(n_atoms, 3)
    box = scale * np.eye(3)

[...]

PR #664, opened 2022-03-04
Fix incorrect argument to `gpuErrchkCudaMallocAndCopy`

mcwitt:
Partially addresses #663.

Fixes incorrect arguments passed to `gpuErrchkCudaMallocAndCopy` in neighborlist code, which was causing nondeterministic segfaults ([example reproduction](https://gist.github.com/mcwitt/9c62a38bff4ad373ca7044cd18661941)).

The implementation of `gpuErrchkCudaMallocAndCopy` already multiplies the argument by `sizeof(*buffer)`, so we were previously multiplying by the element size twice in the buffer size computation.

### DIFF
--- a/timemachine/cpp/src/neighborlist.cu
+++ b/timemachine/cpp/src/neighborlist.cu
@@ -66,8 +66,8 @@ void Neighborlist<RealType>::compute_block_bounds_host(
     double *h_bb_ctrs,
     double *h_bb_exts) {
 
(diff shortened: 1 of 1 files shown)


proteneer (approved):
lgtm! This was probably my sloppiness to begin with. Thanks for the PR.

proteneer:
Note that this has slightly difference signature from cudaMalloc and cudaMemcpy, with expects `count * sizeof(*ptr)`

maxentile (approved):
Fix looks good -- nice minimal reproduction, too!

PR #665, opened 2022-03-04
Optimize CI linting stage

mcwitt:
Makes a couple of optimizations that speed up the linting stage from ~2 minutes to ~30 seconds:

- Removes `pip install -e .[dev]` from the `lint` stage. Dependencies installed here are not actually used by pre-commit, since it manages its own dependencies
- Installs pre-commit and hook dependencies in the `tm_base_env` build. These should be cached and only rebuilt if `.pre-commit-config.yaml` changes


Dockerfile
```diff
@@ -70,6 +70,11 @@ RUN git clone https://github.com/openmm/openmm.git --branch "${OPENMM_VERSION}"
     make -j "$(nproc)" install && \
     make PythonInstall
 
+# Install pre-commit and cache hooks
+RUN pip install --no-cache-dir pre-commit
```
badisa, line 74: This potentially can hide cases where pre-commit releases a new version that no longer works with our configuration. If we want to pin this here as well in the setup.py that would be a potential solution?

Dockerfile
```diff
@@ -70,6 +70,11 @@ RUN git clone https://github.com/openmm/openmm.git --branch "${OPENMM_VERSION}"
...
+COPY .pre-commit-config.yaml /code/timemachine/
+RUN cd /code/timemachine && git init . && pre-commit install-hooks
```
badisa, line 76: Do we want to use `git init --bare .` here, since we won't actually be using the working tree?

mcwitt: Makes sense, changed in 4cee42a

Dockerfile
mcwitt: Pinned pre-commit in both places in c82f820

badisa (approved):

Dockerfile
mcwitt: Hmm, seems like pre-commit doesn't like working with a bare repo (see [log](https://gitlab.com/relaytx/physics-platform/timemachine/-/pipelines/484885708)). I reverted the change in 23e3a49.

(note this is just a hack to make pre-commit happy; it shouldn't need the git repo for anything when installing hooks, but apparently does some startup checks that require the presence of a git repo)

PR #668, opened 2022-03-16
Increase launch timeout

badisa:
What I had thought was an issue with the Gitlab runner configuration turned out to be a timing issue with compute-sanitizer. If compute-sanitizer fails to attach to the process the tests hang indefinitely, so this PR increases the timeout from the default of 10 seconds to 2 minutes. This did reveal a worth while change to make to the gitlab runner though, so expect a PR for that as well. 

Examples:
Run using 14.8.2 gitlab runner that succeeded
https://gitlab.com/relaytx/physics-platform/timemachine/-/jobs/2200729882

Failure with 14.8.2 gitlab runner that logs issue about timing out, which does seem more common with new agent, but no differences in docker, nvidia configuration *container toolkit, drivers, etc).
https://gitlab.com/relaytx/physics-platform/timemachine/-/jobs/2211911855

Run from commit, cancelled with no progress after 15 minutes[ b707265](https://github.com/proteneer/timemachine/pull/668/commits/b70726534638206a6b2c2d864e585d155b2d84c5)
https://gitlab.com/relaytx/physics-platform/timemachine/-/jobs/2213005592

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ verify:
 
 .PHONY: memcheck_tests
 memcheck_tests:
-	compute-sanitizer --leak-check full --error-exitcode 1 pytest -m $(MEMCHECK_MARKER) $(PYTEST_CI_ARGS)
+	compute-sanitizer --launch-timeout 1 --leak-check full --error-exitcode 1 pytest -m $(MEMCHECK_MARKER) $(PYTEST_CI_ARGS)
 
 .PHONY: unit_tests
 unit_tests:
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ verify:
 
 .PHONY: memcheck_tests
 memcheck_tests:
-	compute-sanitizer --launch-timeout 1 --leak-check full --error-exitcode 1 pytest -m $(MEMCHECK_MARKER) $(PYTEST_CI_ARGS)
+	compute-sanitizer --launch-timeout 120 --leak-check full --error-exitcode 1 pytest -m $(MEMCHECK_MARKER) $(PYTEST_CI_ARGS)
 
 .PHONY: unit_tests
 unit_tests:


Makefile
```diff
@@ -27,7 +27,7 @@ verify:
 
 .PHONY: memcheck_tests
 memcheck_tests:
-	compute-sanitizer --leak-check full --error-exitcode 1 pytest -m $(MEMCHECK_MARKER) $(PYTEST_CI_ARGS)
+	compute-sanitizer --launch-timeout 120 --leak-check full --error-exitcode 1 pytest -m $(MEMCHECK_MARKER) $(PYTEST_CI_ARGS)
```
mcwitt, line 30: Could `--launch-timeout 0` make sense here? Apparently this makes compute-sanitizer wait indefinitely for an attachable process:

> A value of zero forces compute-sanitizer to wait infinitely.

badisa: `--launch-timeout 0` would work as well. I chose 2 minutes as I would like to know if it fails to attach after some period, would seem like something is very off in that case. 

Since we only run this against tests that use the `memcheck` mark, we should be able avoid the case of running non-cuda code as this won't catch any memory issues anyways.

badisa:
> The issue with compute-sanitizer not attaching to the process if the timeout is too short makes sense to me, but I'm not sure I understand why this leads to tests hanging indefinitely? When I run `compute-sanitizer --launch-timeout 1 pytest` locally, I see the `Error: No attachable process found` message, but the job terminates as usual when the tests finish.

I am not clear exactly why this is happening. I am able to replicate it locally by modifying the timeout to 1 and running `make memcheck_tests`, tried copying the make command into my shell and it behaves the same way. Looking at `nvidia-smi` it says that there is nothing running on the GPU.

Makefile
mcwitt: > Is there a way to make the tests fail if compute-sanitizer fails to attach?

Ah, never mind. I guess the current behavior of compute-sanitizer in CI is to hang indefinitely if it fails to attach, so we won't miss any failures.

Interestingly, when I test locally, compute-sanitizer does terminate when the tests finish even if it fails to attach (but does return an error code, so I think we're ok even if the hanging behavior goes away with an upgrade eventually)

mcwitt (approved):

proteneer (approved):
lgtm!

PR #670, opened 2022-03-17
Fix error when importing timemachine.lib.potentials without CUDA extension

mcwitt:
#661 broke the ability to import `timemachine.lib.potentials` when the extension module was not present (in particular on all non-linux machines, where the extension can't be built).

The cause was the addition of the `NonbondedImplWrapper` class, which allows the `disable_hilbert()` and `set_nblist_padding()` methods to work on custom nonbonded potentials built using `FanoutSummedPotential`. The wrapper class inherits from `custom_ops.FanoutSummedPotential`, so the declaration will fail if the `custom_ops` module is not present.

This PR:
- adds an empty stub for `FanoutSummedPotential` to `custom_ops.py`. This allows the top-level declaration of `NonbondedImplWrapper` to succeed even if the extension is not present
- adds a test that will fail if `import timemachine.lib.potentials` fails without the extension present

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -51,9 +51,7 @@ nogpu-tests:
   rules:
     - if: $CI_EXTERNAL_PULL_REQUEST_IID
   script:
-    - CMAKE_BUILD_PARALLEL_LEVEL=$(nproc) pip install -e .[test]
-    # HACK: remove C++/CUDA library (is there a way to tell pip not to build it?)
-    - rm timemachine/lib/custom_ops.*.so
+    - SKIP_CUSTOM_OPS=1 pip install -e .[test]
     - make nogpu_tests
   artifacts:
     name: "$CI_JOB_NAME-$CI_COMMIT_REF_NAME-nogpu-tests"
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -42,6 +42,26 @@ lint:
   script:
     - make verify
 
+nogpu-tests:
+  stage: test
+  image: $DOCKER_TAG
+  needs: ["lint"]
+  tags:
+    - timemachine
+  rules:
+    - if: $CI_EXTERNAL_PULL_REQUEST_IID
+  script:
+    - CMAKE_BUILD_PARALLEL_LEVEL=$(nproc) pip install -e .[test]
+    # HACK: remove C++/CUDA library (is there a way to tell pip not to build it?)
+    - rm timemachine/lib/custom_ops.*.so
+    - make nogpu_tests
+  artifacts:
+    name: "$CI_JOB_NAME-$CI_COMMIT_REF_NAME-nogpu-tests"
+    paths:
+        - coverage/
+    when: on_success
+    expire_in: 1 week
+
 memory-tests:
   stage: test
   image: $DOCKER_TAG
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ memcheck_tests:
 
 .PHONY: unit_tests
 unit_tests:
-	pytest -m 'not $(MEMCHECK_MARKER)' $(PYTEST_CI_ARGS)
+	pytest -m 'not $(NOGPU_MARKER) and not $(MEMCHECK_MARKER)' $(PYTEST_CI_ARGS)
 
 .PHONY: ci
 ci: verify memcheck_tests unit_tests
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ CPP_DIR := $(MKFILE_DIR)timemachine/cpp/
 INSTALL_PREFIX := $(MKFILE_DIR)timemachine/
 PYTEST_CI_ARGS := --color=yes --cov=. --cov-report=html:coverage/ --cov-append --durations=100
 
+NOGPU_MARKER := nogpu
 MEMCHECK_MARKER := memcheck
 
 NPROCS = `nproc`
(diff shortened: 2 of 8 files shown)


.gitlab-ci.yml
```diff
@@ -42,6 +42,26 @@ lint:
...
+  image: $DOCKER_TAG
+  needs: ["lint"]
+  tags:
+    - timemachine
+  rules:
+    - if: $CI_EXTERNAL_PULL_REQUEST_IID
+  script:
+    - CMAKE_BUILD_PARALLEL_LEVEL=$(nproc) pip install -e .[test]
+    # HACK: remove C++/CUDA library (is there a way to tell pip not to build it?)
+    - rm timemachine/lib/custom_ops.*.so
```
mcwitt, line 56: This is unfortunate but works. I couldn't find a way to install the package and test dependencies but skip building the extension with pip

tests/import_without_cuda.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+
+pytestmark = [pytest.mark.nogpu]
+
+
+def test_import_potentials_without_cuda():
+    import timemachine.lib.potentials  # noqa: F401
```
mcwitt, line 7: `noqa` to prevent flake8 yelling about unused import

timemachine/lib/potentials.py
```diff
@@ -217,17 +217,27 @@ def get_b_idxs(self):
...
 
 
-class NonbondedImplWrapper(custom_ops.FanoutSummedPotential):
+class NonbondedImplWrapper:
+    # NOTE: not implemented as a subclass of
+    # custom_ops.FanoutSummedPotential to allow import of
+    # timemachine.lib.potentials to succeed when the CUDA extension is
+    # not present
+    def __init__(self, *args):
+        self._impl = custom_ops.FanoutSummedPotential(*args)
```
mcwitt, line 226: Again, a bit of a hack. An alternative might be to define

```python
class FanoutSummedPotential:
    pass
```
in `custom_ops.py`, but that seemed like it might lead to more confusing errors if the user tries to use without the CUDA extension.

Makefile
```diff
@@ -25,6 +26,10 @@ grpc:
 verify:
 	pre-commit run --all-files --show-diff-on-failure --color=always
 
+.PHONY: nogpu_tests
+nogpu_tests:
+	pytest -m $(NOGPU_MARKER) $(PYTEST_CI_ARGS)
```
badisa, line 31: Should add `not $(NOGPU_MARKER)` to the unit_tests command, otherwise we run the test twice.

timemachine/lib/custom_ops.py
```diff
@@ -1,6 +1,14 @@
...
+# (e.g. modules containing subclasses of classes defined in the C++
+# code).
+#
+# If the extension module .so file is present, the definitions
+# in it will take precedence over the stubs defined here.
 
-# on the worker side, a custom .so file for the custom_op and will
-# have higher import priority than the .py file
+
+class FanoutSummedPotential:
```
badisa, line 13: What is the reason to inherit from this and not construct it in the same way that we do other potentials? I find inheriting from C++ objects sketchy, as I don't understand the mechanisms underlying it.

Makefile
mcwitt: Thanks, fixed in a4cb39b.

timemachine/lib/custom_ops.py
mcwitt: My reasoning is that `NonbondedImplWrapper` should have the same interface as the C++ `Potential` class (because it's returned by `Nonbonded.unbound_impl()`; i.e. it's not a `CustomOpWrapper`). The motivation for this was to allow `Nonbonded` as implemented as a `FanoutSummedPotential` to maintain the same interface with `disable_hilbert_sort()` and `set_nblist_padding()` -- the wrapper class adds these methods to the underlying `FanoutSummedPotential` class.

I agree that this is a bit hacky and confusing. One alternative could be to implement something like `NonbondedImplWrapper` on the C++ side. I preferred the current approach because it doesn't require an ad-hoc C++ class for each variant of the nonbonded potential. Longer term, it might be cleaner to change the API so that e.g. `disable_hilbert_sort()` isn't called directly on the full nonbonded potential (which is typically a sum of parts), but on the underlying all pairs or interaction group potential.

maxentile (approved):

.gitlab-ci.yml
jkausrelay: Can you just modify setup.py to skip the ext_module if an env var is set?

```
kwargs = {}
if not os.environ.get('SKIP_CUSTOM_OPS'):
    kwargs['ext_modules'] = [CMakeExtension("timemachine.lib.custom_ops", "timemachine/cpp")]
```

badisa (approved):

.gitlab-ci.yml
mcwitt: > but could maybe use the optional argument in Extension?

Nice, this seems better than what I had. It doesn't seem quite right to make the extension optional though, since much of the API will not function without the extension and we haven't really implemented proper error messages for this case.

> Can you just modify setup.py to skip the ext_module if an env var is set?

Thanks, I like this approach! Makes it a deliberate action to disable building the extension, so it seems less potentially confusing to users than making the extension optional. Changed in afd2eed.

timemachine/lib/custom_ops.py
jkausrelay: Add this as a comment to `NonbondedImplWrapper`.

jkausrelay (approved):

PR #671, opened 2022-03-18
Modifies Barostat tests to not rely on AM1 charges

badisa:
* Reduces complexity and platform dependency of tests


jkausrelay:
LGTM. Can confirm these tests pass on my machine with the change.

proteneer:
nice qol change. No real reason to use AM1BCC here for barostat tests.

PR #676, opened 2022-03-21
Fix package_data, remove references to local paths

### LINKED ISSUE #674: custom_ops binary references local source paths
The usage of `__FILE__` to construct paths to the kernel source directory at build time in [wrap_kernels.cpp](https://github.com/proteneer/timemachine/blob/2a971ee27cb3b8fc457fee1439af2f8dbd375b72/timemachine/cpp/src/wrap_kernels.cpp#L690) causes absolute paths to be compiled into the custom_ops extension module.

There are several potential issues with this:

- The compiled package is not portable (might not be a big issue right now)
- Separate instances of the timemachine package installed in different virtualenvs can reference the same source tree, when one might expect them to be independent. This can cause surprises and breaks the intent of reproducibility when maintaining separate virtualenvs for different experiments.

mcwitt:
Closes #674 

Fixes several packaging issues that prevented isolated installations with `pip install`. This should allow working locally with multiple timemachine versions in separate virtualenvs.

- Removes the `-e` option from the `pip install` commands in CI. This will cause failures where package code depends on data that's not included in the package (c4f3b6a)
- Includes `tests/data/ligands_40.sdf` in `package_data` so it can be accessed as a resource by the `testsystems` module (f82e116)
- Fixes test imports that required an editable installation or `PYTHONPATH` to be set (e2bfedb)
- Includes freesolv dataset in `package_data` to fix `test_fetch_freesolv` (f002baa)
- Includes the JIT kernel source (`k_lambda_transformer.cuh`) in custom_ops at compile time, eliminating run-time dependency on kernel sources (see below for more explanation of the method) (1f05ce4)

---

#### Explanation of changes to JIT kernel loading (1f05ce4)

Previously, the local absolute path of the kernel sources was hardcoded into the binary at compile time, which could lead to surprising behavior (see #674).

~This PR takes the approach of defining a constant string in the C++ code containing the JIT kernel source, eliminating runtime dependence on the source tree. This is done automatically at build time by additions to the CMake build:~
~1. `k_lambda_transformer_jit.cuh` is macro-expanded into `k_lambda_transformer_jit.cui`, eliminating `#include`s to generate a self-contained source file ([here](https://github.com/proteneer/timemachine/pull/676/commits/1f05ce4aaf233088926880ca99be11e59f34b9ec#diff-91d026c1fcb8424ad567b2343d63101fc584eb17536930067d7241f94ba7ae7eR39))~
~2. `k_lambda_transformer_jit.cui` is embedded into a `const char k_lambda_transformer_jit_cui[]` in a generated file `k_lambda_transformer_jit_cui.cpp` using the `xxd` utility ([here](https://github.com/proteneer/timemachine/pull/676/commits/1f05ce4aaf233088926880ca99be11e59f34b9ec#diff-91d026c1fcb8424ad567b2343d63101fc584eb17536930067d7241f94ba7ae7eR49), following the approach [here](https://stackoverflow.com/a/411000/1427284)).~

EDIT: the approach has been simplified so that we now pass the kernel source directory in from Python, where it can be calculated relative to the `timemachine` module path. This avoids the need for the preprocessing step, which had some surprising behavior of its own (e.g. eliminating `#define`s which we might expect to be able to reference from custom expressions). The above approach was reverted in commits 67ce2ca through d935cc0, and the new approach implemented in ac6407f.


jkausrelay:
> Includes tests/data/ligands_40.sdf in package_data so it can be accessed as a resource by the testsystems module (https://github.com/proteneer/timemachine/commit/f82e116516df3084ca1449955574d7d85f3170af)

Is this just for convenience (i.e why this file and not the other files in the tests/data folder?)

mcwitt:
> Is this just for convenience (i.e why this file and not the other files in the tests/data folder?)

@jkausrelay this was just a minimal change to get tests passing after `pip install .` -- `ligands_40.sdf` is the only file from `tests/data` that is referenced in the timemachine package (in `timemachine/datasets/utils.py`). I think it would make sense to eventually move more test system data into `testsystems/data`, but for now I opted to keep this PR focused.

tests/test_parameter_interpolation.py
```diff
@@ -132,9 +132,10 @@ def u_reference(x, params, box, lamb):
                 print("lambda", lamb, "cutoff", cutoff, "precision", precision, "xshape", coords.shape)
 
                 args = copy.deepcopy(test_potential.args)
+                pi = f"{np.pi:.14f}"
```
badisa, line 135: This goes from 36 decimal places to 14 decimal places, what is the intention with this change?

tests/test_vacuum_importance_sampling.py
```diff
@@ -12,8 +12,8 @@
 import jax
 import numpy as np
 import pytest
+import test_ligands
```
badisa, line 15: Why did this change?

tests/test_parameter_interpolation.py
mcwitt: The reason for removing the `#define` is that this is expanded by the preprocessor invocation when we generate the `k_lambda_transformer_jit.cui`, so isn't available for use in custom expressions (there might be some workarounds for this, but passing in the value from Python seemed the simplest solution to me).

14 decimal places is roughly the precision of IEEE 754 doubles, so I don't think we lose anything by truncating (`np.pi` is a double, so the decimals after the 15th deviate from the true value of pi anyway).

tests/test_vacuum_importance_sampling.py
mcwitt: The current `from tests import test_ligands` fails when the package is installed without `-e`:

```
ModuleNotFoundError: No module named 'tests'
```

To be honest I'm not quite sure how this worked before, but this change seems correct to me; `tests` is not a module with an `__init__.py` file, so the absolute import doesn't make sense to me.

tests/test_parameter_interpolation.py
mcwitt: Added a note in a2c3325d192c6962c20e70b1b641ac39c5d6d1a7

mcwitt:
Force-pushed to squash a confusing commit/revert/commit sequence

tests/test_vacuum_importance_sampling.py
mcwitt: Good call, I went ahead and converted `tests` to a module in b86e0f7. This allows using relative import syntax to more clearly distinguish local from package imports.

maxentile: >To be honest I'm not quite sure how this worked before

I think this had assumed path/to/timemachine was in python path -- good to fix this!

maxentile:
Related clean-up question: Is there a reason to retain `tests/data/ligands_40.sdf` in addition to `timemachine/datasets/fep-benchmark/hif2a/ligands.sdf`?

I think the `ligands_40.sdf` that lives in `tests/` is just a subset of the mols in the `ligands.sdf` that lives in `datasets/` (omitting two mols: the ones with ID '336' or '7b').

If we can remove the `ligands_40.sdf` that lives in `tests/`, that could further simplify things.

mcwitt:
> Is there a reason to retain tests/data/ligands_40.sdf in addition to timemachine/datasets/fep-benchmark/hif2a/ligands.sdf?

@maxentile hmm, I don't think `tests/data/ligands_40.sdf` exists anymore? (should have been moved in f82e116516df3084ca1449955574d7d85f3170af)

tests/test_vacuum_importance_sampling.py
mcwitt: Strange, adding `__init__.py` to `tests` in b86e0f7768a463dfb3ac480884f0d989ceb95073 seems to have broken the build. I'm investigating why; seems good to understand rather than just reverting the change.

maxentile:
>@maxentile hmm, I don't think tests/data/ligands_40.sdf exists anymore? (should have been moved in https://github.com/proteneer/timemachine/commit/f82e116516df3084ca1449955574d7d85f3170af)

Ahh, right -- missed this!

>Sounds like it could make sense to replace the usage of ligands_40.sdf with timemachine/datasets/fep-benchmark/hif2a/ligands.sdf?

I believe so! (Clearer provenance, etc.)

But may be better to leave to different PR. (With the current packaging improvements, this change would be small, if warranted.)

tests/test_vacuum_importance_sampling.py
mcwitt: > Strange, adding __init__.py to tests in b86e0f7 seems to have broken the build.

I think this is due to the ambiguity around whether to import from the installed package or from the source directory when we run tests from the repo root directory. It seems like the structure we currently have, with tests next to the module root directory, is discouraged for this reason (e.g. [here](https://stackoverflow.com/a/50796779/1427284)).

The recommendation seems to be to move the module root under `src` or similar ([example](https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure)). Maybe we could consider something like this for "deboggling" week? For now, I propose to revert the changes making `tests` a module.

mcwitt: Created #679 to track this.

jkausrelay (approved):

maxentile (approved):

mcwitt:
> I have a general question, what's the use case for the custom kernel here?

@jkausrelay most of our GPU kernel code is compiled at build time, but we have one kernel ([`k_lambda_transformer_jit.cuh`](https://github.com/proteneer/timemachine/blob/03759d3cfdf0422dca25dd124a6925bd61e6aaf3/timemachine/cpp/src/kernels/k_lambda_transformer_jit.cuh#L1)) that is JIT-compiled at runtime.

@proteneer might have more context here, but the short answer is that it's sometimes convenient to pass in custom expressions to control the lambda-dependence of various parameters at runtime. You can see some examples of this in the tests [here](https://github.com/proteneer/timemachine/blob/18d58ebcad08f28e27dbef1c6a26ac164dd9b6c9/tests/test_parameter_interpolation.py#L134-L142). These expression strings get substituted into the kernel at runtime, and the kernel is then JIT-compiled.

timemachine/lib/potentials.py
```diff
@@ -374,10 +391,10 @@ def unbound_impl(self, precision):
 
         custom_ctor = getattr(custom_ops, cls_name_base)
 
-        return custom_ctor(*self.args)
+        return custom_ctor(_kernel_source_dir, *self.args)
```
maxentile, line 394: Nice! So the signature of the Python constructors doesn't need to change to accommodate the new `kernel_dir` argument 👍

maxentile (approved):
Updated approach looks good to me!

timemachine/lib/potentials.py
```diff
@@ -1,13 +1,17 @@
...
 import numpy as np
 from numpy.typing import NDArray
 
+import timemachine
 from timemachine.lib import custom_ops
 
 # (ytz): classes in this class wrap custom_ops but have the added benefit
 # of being pickleable.
 
+_kernel_source_dir = str(pathlib.Path(timemachine.__file__).parent / "cpp" / "src" / "kernels")
```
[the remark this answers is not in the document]
mcwitt: Oh, that's a good catch. Should probably be consistent and use `importlib` here too.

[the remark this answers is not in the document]
jkausrelay: On the other hand, importlib would require cpp to be a python package? That's going to look odd...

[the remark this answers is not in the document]
mcwitt: Here's what it'd look like with `importlib`. Unfortunately `importlib` in Python 3.7 doesn't include the `files` helper, and the recommendation seems to be to pull in the backported `importlib_resources` package.

> On the other hand, importlib would require cpp to be a python package? That's going to look odd...

Agreed about this; however I'm currently leaning toward this method given that it appears to be the recommended approach, and the feature of being format-agnostic seems like it might prevent headaches down the road if we ever choose to distribute timemachine on PyPI, for example. What do you think?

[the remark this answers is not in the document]
jkausrelay: Yeah, I agree, importlib seems to be the way to go. Maybe add a comment in the the cpp __init__ (this file is present for JIT compilation of certain kernels)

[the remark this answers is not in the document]
mcwitt: Good call, added in 6a37006

jkausrelay (approved):

PR #677, opened 2022-03-21
Adds documentation of AM1 Handlers

badisa:
* Difference between the AM1, AM1BCC and AM1CCC handlers are significant


timemachine/ff/handlers/nonbonded.py
```diff
@@ -363,6 +375,18 @@ def static_parameterize(mol):
...
 
 class AM1BCCHandler(SerializableMixIn):
+    """The AM1BCCHandler generates charges for molecules using OpenEye's AM1BCCELF10[1] protocol.
+
+    Charges are conformer and platform dependent as of OpenEye Toolkits 2020.2.0 [2].
+
+    References
+    ----------
+    [1] AM1BCCELF10 Theory
+        https://docs.eyesopen.com/toolkits/python/quacpactk/molchargetheory.html#elf-conformer-selection
```
maxentile, line 385: Could we elaborate slightly on nature of conformer dependence for AM1BCC ELF10?

(I was under the impression the dependence on initial conformer was greatly reduced, or eliminated, by using ELF10.)

timemachine/ff/handlers/nonbonded.py
```diff
@@ -391,6 +415,22 @@ def static_parameterize(mol):
 
 
 class AM1CCCHandler(SerializableMixIn):
+    """The AM1CCCHandler stands for AM1 Correctable Charge Correction (CCC) which uses OpenEye's AM1 charges[1]
+    along with corrections provided by the Forcefield definition in the form of smirks and charge deltas.
+
+    This handler supports back propagation with respect to ligand parameters, which is what the "Correctable" refers
```
proteneer, line 421: I'd simplify this to simply "This handler supports jax.grad with respect to the forcefield parameters".

proteneer (approved):

timemachine/ff/handlers/nonbonded.py
```diff
@@ -391,6 +415,22 @@ def static_parameterize(mol):
 
 
 class AM1CCCHandler(SerializableMixIn):
+    """The AM1CCCHandler stands for AM1 Correctable Charge Correction (CCC) which uses OpenEye's AM1 charges[1]
+    along with corrections provided by the Forcefield definition in the form of smirks and charge deltas.
```
maxentile, line 419: Could we add a clarification whether the SMIRKS patterns use [Daylight SMARTS syntax](https://www.daylight.com/dayhtml/doc/theory/theory.smarts.html), the [OE SMARTS extension](https://docs.eyesopen.com/toolkits/cpp/oechemtk/SMARTS.html), or the common subset?

(I think we use [OE for matching bond smirks](https://github.com/proteneer/timemachine/blob/2a971ee27cb3b8fc457fee1439af2f8dbd375b72/timemachine/ff/handlers/nonbonded.py#L186) in AM1CCCHandler but [RDKit for matching atom smirks](https://github.com/proteneer/timemachine/blob/2a971ee27cb3b8fc457fee1439af2f8dbd375b72/timemachine/ff/handlers/nonbonded.py#L140) to assign LJ parameters. Not sure if this is just a historical quirk or a necessary aspect of the AM1CCC definition.)

timemachine/ff/handlers/nonbonded.py
maxentile: nit: wording -- differentiation w.r.t. ligand-specific system parameters is supported regardless of handler definition. (Want to emphasize difference between ligand-specific system parameters and forcefield-wide shared parameters.)

```suggestion
    This handler supports back propagation with respect to the charge deltas, which is what the "Correctable" refers
```

maxentile (approved):
Looks good! Documents important differences and provides helpful references

timemachine/ff/handlers/nonbonded.py
maxentile: Oh! Already raised in https://github.com/proteneer/timemachine/pull/677#discussion_r831371002 and addressed in https://github.com/proteneer/timemachine/pull/677/commits/98055a8f4c8e7e8f1453cfa2f5b561cf6e6fea10

timemachine/ff/handlers/nonbonded.py
badisa: > Could we add a clarification whether the SMIRKS patterns use [Daylight SMARTS syntax](https://www.daylight.com/dayhtml/doc/theory/theory.smarts.html), the [OE SMARTS extension](https://docs.eyesopen.com/toolkits/cpp/oechemtk/SMARTS.html), or the common subset?

By this do you mean how we interpret them or how they were defined? 

> (I think we use [OE for matching bond smirks](https://github.com/proteneer/timemachine/blob/2a971ee27cb3b8fc457fee1439af2f8dbd375b72/timemachine/ff/handlers/nonbonded.py#L186) in AM1CCCHandler but [RDKit for matching atom smirks](https://github.com/proteneer/timemachine/blob/2a971ee27cb3b8fc457fee1439af2f8dbd375b72/timemachine/ff/handlers/nonbonded.py#L140) to assign LJ parameters. Not sure if this is just a historical quirk or a necessary aspect of the AM1CCC definition.)

@proteneer can you comment on this? I would be for unifying to a single smirks matching if there is not a good reason for the split.

maxentile: >By this do you mean how we interpret them or how they were defined?

How we want to interpret them.

-----

(Some relevant differences encountered by OpenFF in this context:
(1) for some atomic primitives (`R4`, `R5`, `R6`, ...),  valid symbols in both flavors have intentionally different meanings,
(2) for malformed patterns (patterns containing nonsensical atomic primitives like `r0`, `r1`, `r2` , patterns containing ambiguous trailing bond symbols like `[*:1]~`, etc.), RDKit and OpenEye may have different behavior,
(3) for malformed molecules, RDKit and OpenEye have different aromaticity perception + sanitization behavior + ..., which can cause even simple connectivity primitives like `X3` to have different behavior,
(4) the hybridization primitive `^` is defined in OE flavor but not in Daylight flavor (although note that `^` is defined in RDKit extension of Daylight https://www.rdkit.org/docs/RDKit_Book.html#smarts-support-and-extensions ).

I don't think any of these issues impact the current `_ccc.py` files, but still seems good to be more explicit about which flavor this class assumes, since the constructor can accept an arbitrary smirks list.)

timemachine/ff/handlers/nonbonded.py
badisa: Reading more about ELF10, I am not sure we really want this as it expects an ensemble of conformers. Which in most cases we don't provide. The way I interpret the documentation (https://docs.eyesopen.com/toolkits/python/quacpactk/molchargetheory.html#elf-conformer-selection) and this issue (https://github.com/openforcefield/openff-toolkit/issues/1170) is that when you have an ensemble of conformers ELF10 will provide a more 'accurate' set of partial charges. 

In general we only provide a single conformer and even if we provide a few I don't know that we "start with enough conformers so that it can find a population of conformers that do not have strongly-interacting functional groups". While it is not explicit in the docs, I assume it is falling back to AM1BCC when there is only a single conformer though I should check the behavior. 

> I was under the impression the dependence on initial conformer was greatly reduced, or eliminated, by using ELF10.

Will play around with this. I might just need to clarify that this statement means that the charges are not bitwise deterministic across conformer/platform, as it does seem that this would improve the charges accuracy, but use only 2% of the actual conformers.

timemachine/ff/handlers/nonbonded.py
badisa: > (1) for some atomic primitives (R4, R5, R6, ...), valid symbols in both flavors have intentionally different meanings

Based on reading this (https://docs.eyesopen.com/toolkits/cpp/oechemtk/SMARTS.html), it seems like we would probably prefer the OE toolkits in this case (as it is deterministic), though I don't see any references to `R<n>` in our forcefields.

> (3) for malformed molecules, RDKit and OpenEye have different aromaticity perception + sanitization behavior + ..., which can cause even simple connectivity primitives like X3 to have different behavior

🤯 

It seems to me that it would be best to declare this class as using OE Toolkits SMARTS extension in this PR. And in a separate PR do an investigation of RDKIT and see if they differ.

timemachine/ff/handlers/nonbonded.py
maxentile: Thanks for clarifying this. I had misunderstood. (For some reason thought it was automatically generating more than 10 conformers to select from, if only 1 was provided initially...)

>While it is not explicit in the docs, I assume it is falling back to AM1BCC when there is only a single conformer though I should check the behavior.

This does appear to be the case! `oe_assign_charges(mol, 'AM1BCCELF10') == oe_assign_charges(mol, 'AM1BCC')` when `mol.GetNumConformers() == 1`, on some examples...

<details>
Script that might be helpful for playing around with examples:

```python
import numpy as np

from timemachine.ff.handlers.nonbonded import oe_assign_charges
from timemachine.datasets import fetch_freesolv

from rdkit import Chem

def load_freesolv_mol_with_multiple_conformers(i, n_conformers=1, keep_initial_conformer=True):
    mol = fetch_freesolv()[i]
    
    assert mol.GetNumConformers() == 1
    
    initial_conformer = mol.GetConformer()
    AllChem.EmbedMultipleConfs(mol, n_conformers, clearConfs=not keep_initial_conformer)
    
    return mol

mol_i = 0

# single conformer case, dataset conformer vs. rdkit-generated conformer
mols = dict()
mols['a'] = load_freesolv_mol_with_multiple_conformers(mol_i, 1, False)
mols['b'] = load_freesolv_mol_with_multiple_conformers(mol_i, 0, True)

keys = ['a', 'b']
methods = ['AM1BCC', 'AM1BCCELF10']

for key in keys:
    assert mols[key].GetNumConformers() == 1

charges = dict()
for key in keys:
    for method in methods:
        charges[(key, method)] = oe_assign_charges(mols[key], method)
        
# AM1BCCELF10 == AM1BCC in single-conformer case
for key in keys:
    print(charges[(key, 'AM1BCC')] - charges[(key, 'AM1BCCELF10')])
    
# the two mols -- which differ only in conformer -- get assigned different charges
for method in methods:
    print(charges[('a', method)] - charges[('b', method)])

```
</details>

Will leave this "unresolved" for any further discussion, but this fully addresses my question.

timemachine/ff/handlers/nonbonded.py
maxentile: >Based on reading this (https://docs.eyesopen.com/toolkits/cpp/oechemtk/SMARTS.html), it seems like we would probably prefer the OE toolkits in this case (as it is deterministic), though I don't see any references to R<n> in our forcefields.

Agreed about preferring OE behavior in this case, and correct -- no references to R<n> in OpenFF or in our `*_ccc.py` files. (Just this and other handlers could become ambiguous if constructed with an arbitrary `smirks` list in the future.)

>It seems to me that it would be best to declare this class as using OE Toolkits SMARTS extension in this PR. And in a separate PR do an investigation of RDKIT and see if they differ.

That sounds best!

jkausrelay (approved):

PR #680, opened 2022-03-26
Fixes bug where AM1CCCHandler failed on charge molecules

badisa:
* Bug introduced in 1ac56298ca05f64b612acae0379bae6797ac7a81

### DIFF
--- a/timemachine/ff/handlers/nonbonded.py
+++ b/timemachine/ff/handlers/nonbonded.py
@@ -57,6 +57,11 @@ def oe_assign_charges(mol, charge_model="AM1BCCELF10"):
 
     partial_charges = np.array([atom.GetPartialCharge() for atom in oemol.GetAtoms()])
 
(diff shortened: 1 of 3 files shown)


timemachine/ff/handlers/nonbonded.py
```diff
@@ -226,12 +231,9 @@ def apply_bond_charge_corrections(initial_charges, bond_idxs, deltas):
     assert len(deltas) == len(bond_idxs)
 
     net_charge = jnp.sum(initial_charges)
-    net_charge_is_integral = jnp.isclose(net_charge, jnp.round(net_charge), atol=1e-5)
```
badisa, line 229: These are not integral when the compound is charged as we multiply by sqrt(ONE_4PI_ESP0). I moved the integral check to the charging stage as that seemed conceptually cleaner.

maxentile (approved):

proteneer (approved):

PR #681, opened 2022-03-28
Modify Charge Conversion to be more consistent with Separated Topologies

proteneer:
This PR makes two important changes:

1) Remove the ff conversion stage, our new re-weighting protocol no longer relies on this.
2) Better support for charged compounds (note: changing charges is still not supported) using a thermodynamic cycle more consistent with separated topologies. In particular, we now do a charge conversion stage in between the decoupling stages where the charges are swapped.

We need to do a little bit more validation on some test systems before this can be merged. I will remove the `standard_qlj_typer` in a separate PR during deboggling week.

PR #684, opened 2022-03-31
Update free_energy_rabfe.py

proteneer:
This PR fixes a sign mismatch introduced by #681


maxentile (approved):
Ahh, shoot -- had missed this in review

proteneer:
Looking offline with @badisa it turns out the dynamic range of our predictions in the decoupling stages is about 2x higher than the old protocol. We strongly suspect this is because we did not port over the epsilon halving - a future possible optimization would be to also halve the sigma parameter as well.

maxentile (approved):
Looks great!

PR #686, opened 2022-04-01
Throw runtime error if mol contains elements unsupported by AM1CCCHandler

maxentile:
For now set of supported elements is hard-coded (notable omission: phosphorus)


proteneer (approved):
lgtm!

PR #689, opened 2022-04-05
Use optimized solvent decoupling protocol

maxentile:
Modifies RABFE script and relative hydration script to use pre-optimized protocol from https://github.com/proteneer/timemachine/pull/538 .

In a later PR, we will also replace instances of `construct_absolute_lambda_schedule_complex`


badisa (approved):

proteneer:
nice! We may want to sanity check this more by running some benchmarks first

maxentile:
>nice! We may want to sanity check this more by running some benchmarks first

Which benchmarks do we want to see?

PR #690, opened 2022-04-05
Adds -Wall and -Extra compiler flag and converts Warnings to Errors

badisa:


timemachine/cpp/CMakeLists.txt
```diff
@@ -59,10 +75,10 @@ pybind11_add_module(${LIBRARY_NAME} SHARED NO_EXTRAS
   src/kernels/nonbonded_common.cu
 )
 
-include_directories(${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES})
 include_directories(src/kernels)
-include_directories(${CMAKE_CURRENT_BINARY_DIR}/eigen)
-include_directories(${CMAKE_CURRENT_BINARY_DIR}/${CUB_SRC_DIR})
+include_directories(SYSTEM ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES})
```
badisa, line 79: Making these systems libraries suppresses any warnings within them, which is fine since we have no control over the code.

timemachine/cpp/CMakeLists.txt
```diff
@@ -9,6 +9,8 @@ project(timemachine LANGUAGES CXX CUDA)
 find_package(PythonInterp 3.7 REQUIRED)
 find_package(PythonLibs 3.7 REQUIRED)
 
+set(CMAKE_CXX_FLAGS "-Werror -Wall -Wextra")
```
mcwitt, line 12: Should we append to `CMAKE_CXX_FLAGS` instead of setting? (similar to the handling of `CMAKE_CUDA_FLAGS` below)

timemachine/cpp/CMakeLists.txt
```diff
@@ -32,6 +34,20 @@ endif()
...
+file(
+  GLOB
+  VENDORED_FILES
+  "src/vendored/*.c*"
+  "src/vendored/*.h*"
+)
+set_source_files_properties(
+  ${VENDORED_FILES}
+  PROPERTIES
+  COMPILE_FLAGS "-w"
```
mcwitt, line 47: Note: the [CMake docs for `COMPILE_FLAGS`](https://cmake.org/cmake/help/latest/prop_sf/COMPILE_FLAGS.html) mention

> Note This property has been superseded by the [COMPILE_OPTIONS](https://cmake.org/cmake/help/latest/prop_sf/COMPILE_OPTIONS.html#prop_sf:COMPILE_OPTIONS) property.

mcwitt: Assuming this applies to both C++ and CUDA?

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -279,7 +277,7 @@ void declare_context(py::module &m) {
             std::vector<unsigned long long> du_dx(N * D);
             ctxt.get_du_dx_t_minus_1(&du_dx[0]);
             py::array_t<double, py::array::c_style> py_du_dx({N, D});
-            for (int i = 0; i < du_dx.size(); i++) {
+            for (unsigned int i = 0; i < du_dx.size(); i++) {
```
mcwitt, line 280: I guess we could also use `auto` here?

mcwitt (approved):
Looks good to me! Will be comforting to have these new warnings, and enforce with -Werror

timemachine/cpp/src/wrap_kernels.cpp
maxentile: Can also remove this function entirely soon, based on this comment: https://github.com/proteneer/timemachine/blob/06299a7f60106eb1b04de625408ec9f30929b960/timemachine/cpp/src/wrap_kernels.cpp#L273 (but would need to edit also [test_md.py](https://github.com/proteneer/timemachine/blob/06299a7f60106eb1b04de625408ec9f30929b960/tests/test_md.py#L176-L179))

timemachine/cpp/CMakeLists.txt
badisa: Having added a small test file in the vendored with the extension `.cu`, it appears to correctly suppress warnings.

proteneer (approved):
lgtm!

timemachine/cpp/src/wrap_kernels.cpp
proteneer: `auto` is fine here as well.

proteneer: `size_t` is also fine.

maxentile (approved):
Looks good!

Issue #682, opened 2022-03-29
Constructing potential energy functions for many systems on same device


maxentile:
**Context**
We would like to define a Jax function referencing the potential energy functions of many systems at once. (E.g. to define a differentiable loss function in terms of a batch of molecules.)

**Problem**
The way the [Jax wrappers are currently implemented requires to construct separate unbound impls](https://github.com/proteneer/timemachine/blob/928c0c4739b586c45d899f37506dae644acbe798/timemachine/fe/functional.py#L55-L73) for each system, which means (1) some start-up time required to instantiate all impls, and (2) a limit on the number of systems, imposed by the available GPU memory.

**Reproduction**
The following loop to construct a potential energy function for each molecule in FreeSolv will take a couple seconds per iteration, then crash when out of GPU memory.

```python
import timemachine
print(timemachine.__version__)
# toggle-energy-minimization branch @ march 29, 2022
# https://github.com/proteneer/timemachine/tree/a2037e14ccefcdad2ac7465a139412893db27cf8
# (so that loop over mols doesn't have to call minimize_host_4d just to construct the potentials)

from timemachine.datasets import fetch_freesolv
from timemachine.md import enhanced
from timemachine.fe.functional import construct_differentiable_interface_fast
from timemachine.ff import Forcefield

ff = Forcefield.load_from_file("smirnoff_1_1_0_ccc.py")
mols = fetch_freesolv()


def prepare_energy_fxn(mol):
    ubps, params, masses, coords, box = enhanced.get_solvent_phase_system(mol, ff, minimize_energy=False)
    U_fxn = construct_differentiable_interface_fast(ubps, params)
    
    return U_fxn, params


# crashes after a few dozen iterations / several minutes
energy_fxns = []
for mol in mols:
    energy_fxns.append(prepare_energy_fxn(mol))
    print(len(energy_fxns))

# ...
# def loss_fxn(ff_params):
#    # ...
#    # something that requires U_fxn(...) for (U_fxn, _) in energy_fxns
#    # ...
#    return loss
# ...
# _ = grad(loss_fxn)(ff_params)
# ...
```

(And a slightly loggier version of this loop that also queries GPU memory each iteration)
<details>

```python
import subprocess as sp
import os

def get_gpu_memory(device=0):
    """adapted from https://stackoverflow.com/a/59571639"""
    command = "nvidia-smi --query-gpu=memory.free --format=csv"
[...]

PR #692, opened 2022-04-05
Reduce hilbert index memory footprint

maxentile:
Implements the shortest term patch suggested in https://github.com/proteneer/timemachine/issues/682#issuecomment-1083333968 to reduce the memory footprint.

(Modify implementation to allow N_BINS <= 256, then reduce N_BINS to 128 or so.)

Performance impact of varying N_BINS (locally re-compiling and running [`tests/test_benchmark.py`](https://github.com/proteneer/timemachine/blob/331edf4bbadfd28f235e8bcb2b877ecb00991467/tests/test_benchmark.py) on an RTX 3080):

```
percent change ns/day from
N_BINS=256 -> 
N_BINS=[128, 64, 32, 16]

[[-0.11 -0.19 -0.59 -1.79]
 [ 0.15  0.1  -0.32 -1.39]
 [ 0.18  0.07 -0.19 -1.46]
 [ 0.46  0.45  0.42 -0.22]
 [ 1.73  0.71  0.42  1.25]
 [ 0.28  0.13  0.36  0.42]
 [ 0.23  0.22  0.34  0.45]
 [-0.23 -0.41 -0.17  0.03]
 [ 0.28  0.08  0.31  0.46]
 [-0.   -1.45 -0.71 -1.06]
 [ 0.41 -0.25  0.82  0.58]
 [-1.27 -1.39 -0.99 -1.1 ]
 [-1.24 -1.33 -0.94 -1.07]
 [-0.58 -0.43 -0.05 -0.36]
 [-0.6  -0.61 -0.14 -0.44]]

across 15 benchmark conditions
```

Surprisingly:
* all of the performance changes appear to be within +/- 2%, even for N_BINS as low as 16
* the performance changes appear positive for some of the benchmark conditions (`hif2a-apo-barostat-interval-25`, for example -- the 5th listed in the table)

Since this may have uneven performance impact across the benchmark conditions, opening this as a draft: Should the implementation be modified to **expose N_BINS as a constructor parameter?**


proteneer:
What happens if `N_BINS = 1`? Its possible the initial state is already "sorted" (the protein certainly is!) and takes time to decorrelate?

maxentile:
Haven't checked N_BINS = 1 yet, but If N_BINS=2, the performance drops ~50% across the board

maxentile:
N_BINS = 1 looks about the same as N_BINS = 2 (both have significantly worse performance than larger N_BINS)

maxentile:
(Build currently failing due to an unrelated Python linting error on CI)

maxentile:
(Nice -- looks like CI linting stage here was indeed fixed by https://github.com/proteneer/timemachine/pull/695 )

timemachine/cpp/src/kernels/k_nonbonded.cu
```diff
@@ -9,6 +9,7 @@ void __global__ k_coords_to_kv(
     unsigned int *vals) {
 
     const int atom_idx = blockIdx.x * blockDim.x + threadIdx.x;
+    const int N_BINS = 128;
```
badisa, line 12: nit: Would be good to have a single definition of this variable rather than three. That said, not obvious where that should be.

badisa (approved):

timemachine/cpp/src/kernels/k_nonbonded.cu
maxentile: Agreed! To define this in a single place, cleanest approach I could think of was to expose N_BINS as a parameter in a few places, and then set the default value for that parameter once. However, I initially wanted to avoid changing any signatures.

mcwitt: Did you consider putting it in `k_nonbonded.cuh`? (I don't immediately see an issue with this)

maxentile: > nit: `N_BINS` seems slightly misleading to me, since it seems like there are really `N_BINS * N_BINS` "bins". Would something like `hilbert_grid_size` be any clearer?

Agreed -- it's really `N_BINS**3`... Grid size or similar would be an improvement! Will fix...

timemachine/cpp/src/nonbonded_all_pairs.cu
```diff
@@ -118,25 +118,36 @@ NonbondedAllPairs<RealType, Interpolated>::NonbondedAllPairs(
     gpuErrchk(cudaMalloc(&d_sort_vals_in_, K_ * sizeof(d_sort_vals_in_)));
 
     // initialize hilbert curve
-    std::vector<unsigned int> bin_to_idx(256 * 256 * 256);
-    for (int i = 0; i < 256; i++) {
-        for (int j = 0; j < 256; j++) {
-            for (int k = 0; k < 256; k++) {
+    const int N_BINS = 128;
+    const int N_BITS = 8;
+    const int capacity = 1 << N_BITS;
```
mcwitt, line 123: nit: do we have a case convention for constants? E.g. `N_BITS` vs. `capacity`.

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -132,25 +132,36 @@ NonbondedInteractionGroup<RealType, Interpolated>::NonbondedInteractionGroup(
     gpuErrchk(cudaMalloc(&d_sort_vals_in_, N_ * sizeof(d_sort_vals_in_)));
 
     // initialize hilbert curve
-    std::vector<unsigned int> bin_to_idx(256 * 256 * 256);
-    for (int i = 0; i < 256; i++) {
-        for (int j = 0; j < 256; j++) {
-            for (int k = 0; k < 256; k++) {
+    const int N_BINS = 128;
+    const int N_BITS = 8;
+    const int capacity = 1 << N_BITS;
```
mcwitt, line 137: if moving `N_BINS` to `k_nonbonded.cuh` is an option, seems like these constants could be moved there too?

mcwitt (approved):
Nice, this is made less intimidating by removing the magic numbers!

proteneer (approved):
lgtm!

jkausrelay (approved):

timemachine/cpp/src/kernels/k_nonbonded.cu
maxentile: Thanks for these suggestions! Consolidated the definitions into `k_nonbonded.cuh` as suggested: https://github.com/proteneer/timemachine/pull/692/commits/388e71ed81be993239bf2581ebc4d22067233776 . Also renamed from misleading "N_BINS" to "GRID_SIZE"

timemachine/cpp/src/nonbonded_all_pairs.cu
maxentile: You're right that was inconsistent -- modified in https://github.com/proteneer/timemachine/pull/692/commits/388e71ed81be993239bf2581ebc4d22067233776 to keep all-caps for the "global" constant defined in `.cuh`, but switched `N_BITS` to `nBits` for the local constant, also to match the parameter name for easier searching later https://github.com/proteneer/timemachine/blob/8b0e4d2281c69b6aba2f62794349fe1e2f7950bf/timemachine/cpp/src/vendored/hilbert.h#L50

timemachine/cpp/src/nonbonded_interaction_group.cu
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/692/commits/afc84f08cbe01169d63f257de12730843800a33a ! Also converted the checks from runtime errors to static_asserts, which seems more appropriate

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -5,6 +5,15 @@
 #include "nonbonded_common.cuh"
 #include "surreal.cuh"
 
+// Divide [0,1]^3 box into HILBERT_GRID_DIM^3 voxels for Hilbert sort
+static const int HILBERT_GRID_DIM = 128;
+
+// Encode grid index along each dimension using HILBERT_N_BITS
+static const int HILBERT_N_BITS = 8;
```
mcwitt, line 12: Good call changing to static assertions! Small nit: would it make sense for the assertions to live beside the implementation, rather than in the header file? To me this seems more appropriate, since the assertions are specific to the implementation. Admittedly it's also nice to have the assertions near the definition..

mcwitt (approved):

timemachine/cpp/src/kernels/k_nonbonded.cuh
maxentile: I think I'd prefer to keep these in the header -- to be right next to the definition as you mention, and to avoid 2x duplication if these were in the implementations!

PR #696, opened 2022-04-05
Extract reusable test systems

mcwitt:
This is pretty minimal so far and only moves the definitions in `tests/test_ligands.py` to the `testsystems` submodule.

It wasn't clear to me what else (if anything) makes sense to move from `tests/data/` into `testsystems` (note: data that we put in `testsystems` should probably be things we intend to be useful to a library user and that we want to package, vs. data that is only used in tests).

Any other ideas for things to move would be appreciated!


maxentile:
> This is pretty minimal so far and only moves the definitions in `tests/test_ligands.py` to the `testsystems` submodule.
> 
> It wasn't clear to me what else (if anything) makes sense to move from `tests/data/` into `testsystems` (note: data that we put in `testsystems` should probably be things we intend to be useful to a library user and that we want to package, vs. data that is only used in tests).
> 
> Any other ideas for things to move would be appreciated!

A few other possibilities from skimming through the tests directory: (erred on the side of making a large number of suggestions, but some of these may not be useful!)
* DHFR, maybe others from test_benchmark.py ? (seems useful)
* BenzenePhenolSparseTest in test_topology.py ? (unclear)
* make_gaussian_testsystem in test_reweighting.py ? (maybe useful for testing protocol optimization)
* some of the paired smiles in test_rabfe_topology.py ? (maybe useful for stress-testing single-topology methods)
* the log weights functions etc. in test_mtm.py::test_optimized_MTM (the batching and other transformations may be useful to extract to testsystems (or examples or md.enhanced))
* get_cdk8_coords etc in test_bonded.py ? (unclear)
* there are also many molecules defined inline in test_dummy.py, but I don't immediately know which exercise corner cases that could be relevant elsewhere (e.g. if this contains molecules that exercise the "collinear angles" corner case, maybe these are useful also in testing the HarmonicAngle implementation?)

badisa (approved):

jkausrelay (approved):
lgtm!

maxentile (approved):
Nice! The increased variety of convenient testsystems will be a boon for quick testing and benchmarking

PR #697, opened 2022-04-06
Configure warnings for nvcc

mcwitt:
Enables the following warnings for compilation of CUDA code with `nvcc`:

- `-Wall`
  - disable `-Wsign-compare`; fixing this would be a big project, probably requiring either changing the `Potential` interface to use unsigned ints for `N` and `P`, or introducing lots of `static_cast`s (which wouldn't help with safety)

- `-Wextra`
  - disable `-Wunused-parameter`; there are lots of false positives e.g. in `execute_device` subclass implementations that don't use all parameters; "fixing" these by removing the parameter names [harms readability](https://github.com/proteneer/timemachine/pull/697#discussion_r844199375)

- `-Wreorder`: ensures correct ordering of initializers to prevent undefined behavior
  - note: for plain C++ code this is already enabled by `-Wall`; however, it seems necessary to add this flag separately so it's [picked up by `nvcc`](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#generic-tool-options-Wreorder)

Also removes some unused definitions that were exposed by the (now disabled) warning `-Wunused-parameter`, but do seem correct to remove (ebb86dc)

### DIFF
--- a/timemachine/cpp/CMakeLists.txt
+++ b/timemachine/cpp/CMakeLists.txt
@@ -16,6 +16,8 @@ string(APPEND CMAKE_CUDA_FLAGS
   " -Werror all-warnings"
   " -Wall"
   " -Wno-sign-compare"
+  " -Wextra"
+  " -Wno-unused-parameter"
 )
 message(${CMAKE_CUDA_FLAGS})
 
--- a/timemachine/cpp/CMakeLists.txt
+++ b/timemachine/cpp/CMakeLists.txt
@@ -12,6 +12,11 @@ find_package(PythonLibs 3.7 REQUIRED)
 string(APPEND CMAKE_CXX_FLAGS "-Werror -Wall -Wextra")
 
 string(APPEND CMAKE_CUDA_FLAGS "-O3 -lineinfo")
+string(APPEND CMAKE_CUDA_FLAGS
+  " -Werror all-warnings"
+  " -Wall"
+  " -Wno-sign-compare"
+)
 message(${CMAKE_CUDA_FLAGS})
 
 if (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
--- a/timemachine/cpp/CMakeLists.txt
+++ b/timemachine/cpp/CMakeLists.txt
@@ -18,6 +18,7 @@ string(APPEND CMAKE_CUDA_FLAGS
   " -Wno-sign-compare"
   " -Wextra"
   " -Wno-unused-parameter"
+  " -Wreorder"
 )
 message(${CMAKE_CUDA_FLAGS})
 
--- a/timemachine/cpp/src/barostat.cu
+++ b/timemachine/cpp/src/barostat.cu
@@ -21,8 +21,8 @@ MonteCarloBarostat::MonteCarloBarostat(
     const int interval,
     const std::vector<BoundPotential *> bps,
     const int seed)
-    : N_(N), pressure_(pressure), temperature_(temperature), interval_(interval), bps_(bps), group_idxs_(group_idxs),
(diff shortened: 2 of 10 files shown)


timemachine/cpp/CMakeLists.txt
```diff
@@ -9,7 +9,13 @@ project(timemachine LANGUAGES CXX CUDA)
 find_package(PythonInterp 3.7 REQUIRED)
 find_package(PythonLibs 3.7 REQUIRED)
 
-string(APPEND CMAKE_CUDA_FLAGS "-O3 -lineinfo")
+string(APPEND CMAKE_CUDA_FLAGS "-O3 -lineinfo -Werror all-warnings")
+string(APPEND CMAKE_CUDA_FLAGS
+  " -Wall"
+  " -Wno-sign-compare"
```
mcwitt, line 15: I opted to disable `-Wsign-compare` for now, since it seems tricky to fix properly. It seems like we'd either need to transition to declaring `N`, `P`, `D`, etc. as `unsigned int`, or cast to unsigned wherever they are compared with `size_t`. The former would require changing the `Potential` interface.

timemachine/cpp/src/harmonic_bond.cu
```diff
@@ -62,11 +62,11 @@ template <typename RealType> HarmonicBond<RealType>::~HarmonicBond() {
 
 template <typename RealType>
 void HarmonicBond<RealType>::execute_device(
-    const int N,
```
proteneer, line 65: Why are some names of these removed?

timemachine/cpp/src/nonbonded_all_pairs.cu
```diff
@@ -439,7 +439,7 @@ void NonbondedAllPairs<RealType, Interpolated>::execute_device(
 
 template <typename RealType, bool Interpolated>
 void NonbondedAllPairs<RealType, Interpolated>::du_dp_fixed_to_float(
-    const int N, const int P, const unsigned long long *du_dp, double *du_dp_float) {
+    const int N, const int, const unsigned long long *du_dp, double *du_dp_float) {
```
proteneer, line 442: not sure why some of these are removed...

timemachine/cpp/src/harmonic_bond.cu
mcwitt: This was to satisfy `Wunused-parameter` (which is enabled as part of `-Wextra`). Agreed that this seems suboptimal. The warning seems useful to have in some cases, but maybe not worth this.

proteneer: Oof - this does remove useful self-documentation information. Hm... I personally like the old style better.

timemachine/cpp/CMakeLists.txt
proteneer: Yeah this is a nuisance - `static_casting` everywhere is sort of nuts (and just as unsafe).

timemachine/cpp/src/harmonic_bond.cu
mcwitt: I added `-Wno-unused-parameter` and reverted these changes in 36549c3

timemachine/cpp/src/nonbonded_all_pairs.cu
mcwitt: Addressed by https://github.com/proteneer/timemachine/pull/697#discussion_r845461217

proteneer:
nice! this seems like a good balanced set of changes that are not too intrusive and still quite useful! (especially on initializer lists).

proteneer (approved):

badisa (approved):

PR #698, opened 2022-04-06
Deboggle: Merge free_energy and free_energy_rabfe modules

jkausrelay:
- Merge RelativeFreeEnergy/AbsoluteFreeEnergy classes in free_energy.py from free_energy_rabfe.py
- Make interface consistent
    - prepare_combined_coords - separate method to merge the coords
    - prepare_vacuum_edge (used in the tests)
- Move the lambda schedule functions to timemachine.fe.lambda_schedule
- Move the restraint functions to timemachine.fe.restraints
- Modify the scripts/tests to be consistent with above changes

- Note: validate_relative_binding will conflict with #689 due to the lambda schedule import, will update once that is merged
- Note: rhfe_single.py/rbfe_single.py/estimator_variance.py don't appear functional so I moved them to the attic
- Note: I modified model.py so @maxentile @proteneer be aware of this when you make your changes

PR #700, opened 2022-04-06
Merges estimator_abfe.py and estimator.py

badisa:
* Prefers estimator_abfe.py logic

Blocked by #698


tests/test_free_energy.py
```diff
@@ -20,6 +21,7 @@
 
 
 def test_absolute_free_energy():
+    np.random.seed(2022)
```
badisa, line 24: This is redundant, can remove

tests/test_free_energy.py
```diff
@@ -42,7 +44,7 @@ def test_absolute_free_energy():
 
     lambda_schedule = np.linspace(0, 1.0, 4)
     equil_steps = 1000
-    prod_steps = 1000
+    prod_steps = 5000
```
badisa, line 47: These values increased as we were previously getting more samples for TI than we do for BAR. Could change the estimator code to take in a flag to determine how many frames to save.

timemachine/fe/estimator.py
```diff
@@ -130,52 +229,31 @@ def simulate(
...
-        "lambda_schedule",
-        "equil_steps",
-        "prod_steps",
-        "barostat",
-    ],
-)
-
-
-def deltaG(model, sys_params) -> Tuple[float, List]:
+def deltaG_from_results(model, results, sys_params) -> Tuple[float, float, List]:
```
jkausrelay, line 257: Add docs? It might be worth mentioning model.endpoint_correct in particular, since this is specific to a system set up in a certain way.

tests/test_free_energy.py
jkausrelay: Does this impact the time to run the test?

badisa: It does, increases both tests that had this change by about 20 seconds.

tests/test_free_energy.py
```diff
@@ -217,25 +239,30 @@ def binding_model(ff_params):
...
                 x0,
                 v0,
                 integrator,
+                barostat,
                 lambda_schedule,
                 equil_steps,
                 prod_steps,
-                barostat,
+                beta,
+                "prefix",
```
mcwitt, line 262: nit: this seems like a lot of positional args to keep track of. Consider using kwargs?

timemachine/fe/estimator.py
```diff
@@ -1,19 +1,45 @@
...
+        "v0",
+        "integrator",
+        "barostat",
+        "lambda_schedule",
+        "equil_steps",
+        "prod_steps",
+        "beta",
+        "prefix",
+    ],
+)
```
mcwitt, line 42: nit: how would you feel about using the style:

```python
class FreeEnergy(NamedTuple):
    unbound_potentials: List[Potential]
    # ...
```
(main benefit being that it allows typing the elements)

timemachine/fe/estimator.py
```diff
@@ -50,26 +76,92 @@ def equilibrate(integrator, barostat, potentials, coords, box, lamb, equil_steps
     return CoordsVelBox(coords=ctxt.get_x_t(), velocities=ctxt.get_v_t(), box=ctxt.get_box())
 
 
+def run_model_simulations(model, sys_params):
```
mcwitt, line 79: Docs and/or parameter types seem useful here

timemachine/fe/estimator.py
```diff
@@ -184,35 +262,115 @@ def deltaG(model, sys_params) -> Tuple[float, List]:
...
+        N_k.append(len(result.lambda_us))  # number of frames
+
+    U_knk = np.array(U_knk)
+
+    bar_dG = 0
+    bar_dG_err = 0
+
+    delta_Us = extract_delta_Us_from_U_knk(U_knk)
+
+    for lambda_idx in range(len(model.lambda_schedule) - 1):
```
mcwitt, line 283: nit: style preference for
```suggestion
    for lambda_idx, (lamb_start, lamb_end) in enumerate(zip(model.lambda_schedule, model.lambda_schedule[1:])):
```

mcwitt (commented):
Nice, merging these seems like a great simplification!

proteneer (approved):
nice!

timemachine/fe/estimator.py
```diff
@@ -50,26 +76,92 @@ def equilibrate(integrator, barostat, potentials, coords, box, lamb, equil_steps
...
+
+    bound_potentials = []
+    for params, unbound_pot in zip(sys_params, model.unbound_potentials):
+        bp = unbound_pot.bind(np.asarray(params))
+        bound_potentials.append(bp)
+
+    all_args = []
+    for lamb_idx, lamb in enumerate(model.lambda_schedule):
+
+        subsample_interval = 1000
```
maxentile: nit: can move this out of loop to quiet IDE's "might be referenced before assignment" warning at https://github.com/proteneer/timemachine/blob/20c83a8c5e6b334bdd50ba5100a41e1b1623fd6a/timemachine/fe/estimator.py#L124-L125

timemachine/fe/estimator.py
```diff
@@ -130,52 +229,31 @@ def simulate(
...
+    ctxt = custom_ops.Context(x0, v0, box, intg_impl, all_impls, barostat_impl)
 
     # equilibration
     equil_schedule = np.ones(equil_steps) * lamb
     ctxt.multiple_steps(equil_schedule)
 
-    baro_impl.set_interval(base_interval)
+    # (ytz): intentionally hard-coded, I'd rather the end-user *not*
+    # muck with this unless they have a good reason to.
+    barostat_impl.set_interval(25)
```
maxentile, line 271: Clarification: does this change the barostat frequency for production and/or equilibration?

(Possibly a more explicit / visible way to hard code this might be to put this in signatures of `simulate(..., barostat_interval=25)` and `equilibrate(..., barostat_interval=25)`, and raise value error if != 25... Side note: function `equilibrate` is used in RBFEModel.equilibrate_edges, but maybe also should be used at https://github.com/proteneer/timemachine/blob/20c83a8c5e6b334bdd50ba5100a41e1b1623fd6a/timemachine/fe/estimator.py#L237-L239?)

maxentile (approved):
Great simplifications!

timemachine/fe/estimator.py
badisa: Switched this to a data class in [8dae286](https://github.com/proteneer/timemachine/pull/700/commits/8dae286198c36e65ec9a0afac2ec4cfe5821a774)

timemachine/fe/estimator.py
badisa: Resolved and parameterized in [e19ead8](https://github.com/proteneer/timemachine/pull/700/commits/e19ead869f7eb48f47401759356432de3b5485c0)

tests/test_free_energy.py
badisa: Fixed this by parameterizing it in [e19ead8](https://github.com/proteneer/timemachine/pull/700/commits/e19ead869f7eb48f47401759356432de3b5485c0)

jkausrelay (approved):
Looks great!

timemachine/fe/estimator.py
badisa: Currently everywhere that is using 25 steps. 

In all likelihood that `equilibrate` function should disappear. I implemented the behavior in `equilibrate_edges` incorrectly and should probably be replaced with `minimizer.equilibrate_host`. I imagine this will also change when we start using Single Topology again and didn't want to muddy this PR with changes outside the scope of merging the two files.

I hesitate to raise a value error in this code as it runs on the servers. If someone changed the value and ran it in parallel could end up wasting a lot of resources to raise an exception that would occur for all simulations.

PR #702, opened 2022-04-07
Consider replacing Hilbert sort with Morton code sort

maxentile:
After observing in https://github.com/proteneer/timemachine/pull/692 that performance is fairly insensitive to the grid dimensions used for particle sorting, I was curious if performance is also insensitive to the choice of space filling curve used for particle sorting.

If so, then we might be able to remove ~2000 lines of vendored code (and a python dependency `hilbertcurve==1.0.5`), since Morton encoding is simpler. (I borrowed this implementation from a code snippet in a 2012 NVIDIA blog post https://developer.nvidia.com/blog/thinking-parallel-part-iii-tree-construction-gpu/ .)

Also, since Morton *decoding* can be done using cheap bit magic, we could later investigate performance impact of using that rather than a LUT for decoding. For now though, I just swapped out `hilbert_encode(grid_points)` -> `morton_encode(grid_points)` and retained the LUT based on `grid_points`.

Initial local tests show no performance difference between Morton and Hilbert ordering for grid dimensions 256 or 128 on `python tests/test_benchmark.py`. Further testing is necessary to ensure this would cause no other performance regressions, and to check if there may be other reasons to prefer the Hilbert over alternatives.

If this change would be something to consider further, it would require a few more changes not yet applied:
* name changes throughout ("hilbert_sort" -> "sort_for_locality" or "morton_sort" or something)
* updating tests


proteneer:
Nice! I remember looking into this (and other types of space filling curves a decade ago) with @peastman almost a decade ago. IIRC Hilbert curve was slightly better on bigger systems, but it wasn't very significant.

maxentile:
Closing for now since https://github.com/proteneer/timemachine/pull/692 unblocks https://github.com/proteneer/timemachine/issues/682 in the immediate term -- may revisit at a later time!

PR #704, opened 2022-04-07
Deboggle: Allow num_batches%num_workers!=0 for generate_log_weighted_samples

jkausrelay:
- Pads the computation to generate enough samples, then truncate at the end
- After discussion w/Yutong, decided to leave the XLA_FLAGS as is for now, we can revisit if pmap becomes more commonly used


proteneer (approved):
nice! the only pathology I can see is if `num_batches` << `num_workers`, in which case this would just tie up for the CPU for no good reason. But in most use cases I think `num_batches` >> `num_workers` so the overhead is trivial.

proteneer (approved):

maxentile (approved):
Nice --  looks like a neat change to decouple `num_batches` from `num_workers`.

In a future elaboration, we might see about factoring out the details of the `pmap` execution policy, `_wrap_simulate`, etc. from the definition of the MTM enhanced vacuum state. Seems like the trick could be reusable in other circumstances (e.g. running short trajectories in parallel for SMC), and is not inherent to the MTM enhanced vacuum state (depending on system size, maybe `vmap` on a single GPU device is preferable -- not sure if we have benchmarked this).

jkausrelay:
> Nice -- looks like a neat change to decouple `num_batches` from `num_workers`.
> 
> In a future elaboration, we might see about factoring out the details of the `pmap` execution policy, `_wrap_simulate`, etc. from the definition of the MTM enhanced vacuum state. Seems like the trick could be reusable in other circumstances (e.g. running short trajectories in parallel for SMC), and is not inherent to the MTM enhanced vacuum state (depending on system size, maybe `vmap` on a single GPU device is preferable -- not sure if we have benchmarked this).

Yes, I agree.

PR #705, opened 2022-04-08
Cleans up images older than 2 days

badisa:
* This can fail if we ever run more than one job on an instance

This will make CI intermittently more expensive by requiring downloading images from the registry every couple of days. As long as the base image (`nvidia/cuda:11.6.0-devel-ubuntu20.04`) doesn't change too often or requirements change. It should avoid the disk space issues we ran into this week.


.gitlab-ci.yml
```diff
@@ -24,6 +24,8 @@ docker_build:
     - docker
   script:
     - DOCKER_HOST=tcp://localhost:2375 docker login -u $CI_REGISTRY_USER -p $CI_JOB_TOKEN $CI_REGISTRY
+    # Prune images older than 2 days
+    - docker system prune --filter "until=48h" -f -a
```
mcwitt, line 28: 2 days seems pretty aggressive given that I don't think we were pruning at all before and didn't run into disk space issues for ~weeks? (I might be mistaken about this) Would it make sense to bump this up a bit to reduce the number of unnecessary downloads?

badisa: This is all dependent on how often pip/conda requirements change and if the base image changes. Could increase it to something like a week, but not sure that this wouldn't bite us in a week like these where we have lots of changes in our images. We could also increase the disk size as right now there is only about 80gb free at any time.

mcwitt (approved):

jkausrelay (approved):

maxentile (approved):

proteneer (approved):

badisa:
Going to merge this as is, and will be watch our CI builds in case of negative impacts

Issue #706, opened 2022-04-08
Spurious calls to execute_selective in Jax wrapper


maxentile:
Due to inefficient choices I made way back in https://github.com/proteneer/timemachine/pull/429 
https://github.com/proteneer/timemachine/blob/836aeac2aa1a3eac244682cb603cdc7f38fd66a3/timemachine/fe/functional.py#L15-L43

1. the Jax wrappers make multiple calls to `impl.execute_selective` whenever at least one derivative is needed
2. when multiple derivatives are needed, these are not collapsed into a single call to `impl.execute_selective`

See this notebook for details: https://gist.github.com/maxentile/0759c025f9d141060741c7520c54d960

```python
>>> # fine: 1 call to execute selective, as expected
>>> _ = U_fast(coords, sys_params, box, lam)
calling execute_selective(..., u)
```

```python
>>> # problem: spurious calls to request u, dudp, dudl
>>> _ = grad(U_fast, argnums=0)(coords, sys_params, box, lam)
calling execute_selective(..., u)
calling execute_selective(..., dudx)
calling execute_selective(..., dudp)
calling execute_selective(..., dudl)
```

```python
>>> # problem: not collapsed into execute_selective(..., u, dudx, dudp, dudl)
>>> _ = value_and_grad(U_fast, argnums=(0, 1, 3))(coords, sys_params, box, lam)
calling execute_selective(..., u)
calling execute_selective(..., dudx)
calling execute_selective(..., dudp)
calling execute_selective(..., dudl)
```

-------

Note: this is a source of inefficiency in both the Python loop version `construct_differentiable_interface` and the `SummedPotential` version `construct_differentiable_interface_fast`. (In the Python loop version, the total number of calls is multiplied by `len(unbound_potentials)`.)

maxentile:
WIP fix in https://github.com/proteneer/timemachine/pull/707 avoids the spurious calls, and collapses multiple calls: https://gist.github.com/maxentile/1463655fea7900317e7c0bc62c154a46 . (However, there's a remaining issue with param packing / unpacking that affects the SummedPotential version when du_dp is requested...)

PR #707, opened 2022-04-08
Avoid spurious calls to execute_selective in Jax wrapper

maxentile:
Mostly resolves https://github.com/proteneer/timemachine/issues/706 by
* switching from 3 independent functions in `.defjvps` to a single function in `.defjvp`
* detecting which derivatives are being requested, so these can be collapsed into a single call to `impl.execute_selective`

------

Note: I initially went on a bit of a goose-chase trying to switch from `defjvp` to `core.Primitive` based on a mistaken guess of what was causing these spurious calls. However, I ran into various snags, and this is a more direct solution to the immediate performance issue.

In the future, there will likely be additional value from going in the direction of https://github.com/dfm/extending-jax , but that will be much more involved!


timemachine/fe/functional.py
```diff
@@ -13,38 +16,61 @@ def _make_selection_mask(compute_du_dx=False, compute_du_dp=False, compute_du_dl
 
 
 def wrap_impl(impl, pack=lambda x: x):
```
maxentile, line 18: Before this is ready for review, I need to resolve an issue with how I'm handling `pack(params)`

maxentile: Resolved in https://github.com/proteneer/timemachine/pull/707/commits/09c0406635c6ceab0a740dbee938ad7441f253fa

timemachine/fe/functional.py
```diff
@@ -13,38 +16,64 @@ def _make_selection_mask(compute_du_dx=False, compute_du_dp=False, compute_du_dl
...
+
+        # unpack result tuple
+        primal_out = result_tuple[3]
+        coords_g, params_g, lam_g = result_tuple[:3]
+
+        # compute tangent_out by dotting tangents with jacobians
+        # TODO: make this a zippy comprehension?
+        tangent_out = 0.0
+        if derivative_requested(coords_t):
+            tangent_out += np.sum(coords_t * coords_g)
```
proteneer, line 64: nit: this should implement a jvp (matrix-vector product). As written, this looks like a `vjp`. 

The `J` here should be `coords_g`, which has shape `(1,3N)`, and `v=coords_t` has shape `(3N,)`. So I think you can simply rewrite this as `tangent_out += coords_g @ coords_t`

tests/test_free_energy.py
```diff
@@ -358,6 +360,10 @@ def low_dim_f(perturb: float) -> float:
                 abs_err = check_grad(low_dim_f, grad(low_dim_f), perturb, epsilon=1e-4)
                 assert abs_err < 1e-3
 
+        # grad w.r.t. box shouldn't be allowed
+        with pytest.raises(RuntimeError):
+            _ = grad(U, argnums=2)(coords, sys_params, box, lam)
```
mcwitt, line 365: nit: worth checking the error message to make sure we're getting  the expected `RuntimeError`?

timemachine/fe/functional.py
```diff
@@ -13,38 +16,64 @@ def _make_selection_mask(compute_du_dx=False, compute_du_dp=False, compute_du_dl
...
+        # call custom op once
+        result_tuple = impl.execute_selective(coords, params, box, lam, *selection)
+
+        # unpack result tuple
+        primal_out = result_tuple[3]
+        coords_g, params_g, lam_g = result_tuple[:3]
+
+        # compute tangent_out by dotting tangents with jacobians
+        # TODO: make this a zippy comprehension?
+        tangent_out = 0.0
```
mcwitt, line 62: not sure I'm getting this right, but would `out_t` be more consistent with the naming convention of `_t` for tangents?

mcwitt (approved):

timemachine/fe/functional.py
maxentile: Will look into this -- I assumed writing as `np.sum(coords_g * coords_t)` was fine since `U` is scalar-valued.

>So I think you can simply rewrite this as tangent_out += coords_g @ coords_t

Will look into this further -- replacing with `coords_g @ coords_t` locally introduces errors like `TypeError: dot_general requires contracting dimensions to have the same shape, got (3,) and (35,).`

timemachine/fe/functional.py
```diff
@@ -13,38 +16,64 @@ def _make_selection_mask(compute_du_dx=False, compute_du_dp=False, compute_du_dl
...
-    def U_jvp_lam(lam_dot, _, coords, params, box, lam):
-        selection = _make_selection_mask(compute_du_dl=True)
-        result_tuple = impl.execute_selective(coords, pack(params), box, lam, *selection)
-        return np.sum(lam_dot * result_tuple[2])
+        # handle case where _params is a list of (maybe traced) arrays
+        params, params_t = pack(_params), pack(_params_t)
+
+        # inspect tangent types to determine which derivatives are being requested
+        def derivative_requested(array_t):
+            return isinstance(array_t, Tracer)
```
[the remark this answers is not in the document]
maxentile: >Out of curiosity, what's the instance type if the derivative is not being requested? None?

When derivative is not requested, `DeviceArray`

When derivative is requested usually `Traced<ShapedArray>` (specifically something like `jax.interpreters.partial_eval.JaxprTracer`), specific instance of `jax.interpreters.partial_eval.Tracer`

>Are there other operations aside from jax.grad that might introduce Tracers? eg. vmap/etc.

Good call -- not sure. Unsupported `vmap(U)(...)` already fails with an error as expected as of #429 , and just confirmed that `vmap(grad)(U)(...)` also fails.

timemachine/fe/functional.py
maxentile: Will clarify!

tests/test_free_energy.py
maxentile: Will fix!

timemachine/fe/functional.py
maxentile: Note that `coords_g` and `coords_t` currently have the same shape `(n_atoms, 3)`.

I tried a few ways to expand `coords_g` (n_atoms, 3) to `coords_J` of shape `(1, n_atoms, 3)` and replace elementwise multiplication with matrix multiplication, but didn't view any of these as a safety or readability improvement yet...

Unless you have a strong preference here, I'd prefer to keep the implementation as-is, but just clarify naming and add a note that this is specialized to the case of `U` being scalar-valued, as in https://github.com/proteneer/timemachine/pull/707/commits/7bc5f6b709bd2a44e0ff988ef2eb39610145d53b ?

tests/test_free_energy.py
maxentile: Added in https://github.com/proteneer/timemachine/pull/707/commits/048e5d7fea60578063a1217686ba1493dae8aae2 !

timemachine/fe/functional.py
maxentile: Clarified in https://github.com/proteneer/timemachine/pull/707/commits/7bc5f6b709bd2a44e0ff988ef2eb39610145d53b ?

timemachine/fe/functional.py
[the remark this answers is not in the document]
proteneer: Thanks for checking!

proteneer (approved):
briefly discussed offline re: making the jvp calculation more explicit for pedagogical purposes (and noting that in general jvp != vjp).

timemachine/fe/functional.py
maxentile: Clarified in https://github.com/proteneer/timemachine/pull/707/commits/fb584904771cd22b2282bc5ed7e817d58319573b !

PR #711, opened 2022-04-08
Configure mypy and typing stubs for custom_ops, fix errors

mcwitt:
- Adds [mypy](https://mypy.readthedocs.io/en/stable/) check to pre-commit. This adds linting based on type annotations, e.g. ensuring annotations are correct and catching logic errors exposed by annotations. Initially configured with the following options:
  - [`ignore_missing_imports = True`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-ignore_missing_imports) -- suppresses errors due to 3rd-party imports that don't have type annotations (e.g. `rdkit`)
  - [`check_untyped_defs = False`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-check_untyped_defs) -- do not run checks inside of functions that don't have type annotations (useful for code that relies on dynamic typing)
- Fixes errors exposed by the mypy check, e.g.
  - rename `fep-benchmark` -> `fep_benchmark` (the former is invalid as a module name, so it would not have worked with importlib)
  - use correct type for `SimulationResult` (now possible since #700)
  - `DualTopology` extends `BaseTopology`
  - Fixes for [DeterministicMTMMove](https://github.com/proteneer/timemachine/pull/711/commits/37631dd4fa2a6edc8ebcce50e6dc99d6d0ace0f8#diff-ac7814ab4c10b35ed0db95923931a3fbb2dd37189d0fb42f22fcad92432d1ba3R123-R130) 
- Adds auto-generated stub file for the custom_ops extension module and updates the CMake config to generate the stub file using [mypy-stubgen](https://mypy.readthedocs.io/en/stable/stubgen.html)
  - the stub file contains signatures extracted from the compiled `.so` file, and enables typechecking and auto-completion for `timemachine.lib.custom_ops`
  - Note that the docstrings defined in `wrap_kernels.cpp` are currently not included. There is an alternative to mypy-stubgen, [pybind11-stubgen](https://github.com/sizmailov/pybind11-stubgen), that does include docstrings in the stubs -- however, there were other issues with this (e.g. needing to replace `s/timemachine:://g` for syntactically-valid stubs), and it seemed better to stick with the official mypy-stubgen
  - Adds pre-commit check that the stub file is up-to-date ([example failure output](https://gitlab.com/relaytx/physics-platform/timemachine/-/jobs/2328432212)) ~and a step in the CI unit-tests stage to verify this Note: the pre-commit check for stubs is skipped in CI, because it requires the .so file to be built, and this is not feasible in the linting stage. Instead, this adds a check to unit-tests, after the module has been built~ building the extension module as part of a pre-commit check does work and doesn't seem as unreasonable as I'd originally thought -- this is the method the PR is currently using

## Review notes

- Most of additions are from the auto-generated `custom_ops.pyi` file. This probably doesn't need careful review, as it's only used for typechecking and completions

- Even if we choose not to adopt mypy as a pre-commit check, we might consider merging the generated stub file and the annotation fixes. The mypy check should be fully contained in 157cd7f.

### Some common warnings and fixes
<details>

#### Using the same name for variables of different types
See [docs](https://mypy.readthedocs.io/en/stable/common_issues.html#redefinitions-with-incompatible-types)
```python
a = [1, 2, 3]
a = np.array(a)
```
#### Fix
```python
a_ = [1, 2, 3]
a = np.array(a_)
```
#### Initializing empty container and appending
See [docs](https://mypy.readthedocs.io/en/stable/common_issues.html#types-of-empty-collections)
```python
a = []
a.append(1)
```
#### Fix
```python
a: List[int] = []
a.append(1)
```
or
```python
a: List[Any] = []
a.append(1)
```
or (not recommended)
```python
a: Any = []
a.append(1)
```
</details>


timemachine/potentials/jax_utils.py
```diff
@@ -3,10 +3,8 @@
 from jax import vmap
 from numpy.typing import NDArray
 
-Array = NDArray
 
-
-def get_all_pairs_indices(n: int) -> Array:
+def get_all_pairs_indices(n: int) -> NDArray:
```
maxentile, line 7: Question: Does mypy not allow type aliases like this? (Or perhaps there's a preferred way to declare type aliases?)

mcwitt: I was surprised by this too -- there doesn't seem to be a problem with type aliases in general, but a naive `Array = NDArray` does not work; this results in `Variable "timemachine.potentials.jax_utils.Array" is not valid as a type` wherever the alias is used. Still trying to understand the reason for this and whether there's a solution that still allows aliasing..

mcwitt: The issue seems to be that mypy doesn't correctly identify this as a type alias (relevant [mypy issue](https://github.com/python/mypy/issues/7866)).

[PEP 613](https://peps.python.org/pep-0613/) seems to fix this by allowing explicit specification of type aliases (via the `TypeAlias` type), and an implementation is available for 3.7 in the `typing-extensions` backport package. I added `typing-extensions` and reverted the removal of the `Array` alias in 37d5f5e.

generate_stubs.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+pybind11-stubgen --no-setup-py --root-module-suffix='' --bare-numpy-ndarray --output-dir='./stubs' timemachine.lib.custom_ops
+sed -i 's/timemachine:://g' stubs/timemachine/lib/custom_ops/__init__.pyi
+cp stubs/timemachine/lib/custom_ops/__init__.pyi timemachine/lib/custom_ops.pyi
```
mcwitt, line 5: TODO: could this be added to pre-commit?

timemachine/md/minimizer.py
```diff
@@ -113,6 +113,7 @@ def minimize_host_4d(mols, host_system, host_coords, ff, box, mol_coords=None) -
 
     num_host_atoms = host_coords.shape[0]
 
+    top: Union[topology.BaseTopology, topology.DualTopology]
```
mcwitt, line 116: Should `DualTopology` extend `BaseTopology`?

timemachine/potentials/jax_utils.py
maxentile: Nice! Thanks for looking into that -- the  `MyT = T` -> `MyT: TypeAlias = T` fix looks easy (and imo improves clarity)

generate_stubs.sh
mcwitt: Done in 683080b

timemachine/md/minimizer.py
mcwitt: Done in f0622c3

.gitlab-ci.yml
```diff
@@ -44,6 +44,7 @@ lint:
     - if: $CI_EXTERNAL_PULL_REQUEST_IID
     - if: $NIGHTLY_TESTS
   script:
+    - pip install -r ci/requirements.txt
```
mcwitt, line 47: This seems unfortunate, since we already specify the mypy dependency in 2 places (dev extras in setup.py and build system depends in pyproject.toml), but neither of these can be used here because I think we want to avoid building the package (this would add significant time to the linting stage)

pyproject.toml
```diff
@@ -15,6 +15,21 @@ skip_gitignore = true
...
+
+[[tool.mypy.overrides]]
+module = "timemachine._vendored.fire"
+follow_imports = "silent"
+
 [build-system]
-requires = ["setuptools>=43.0.0", "wheel", "cmake==3.22.1", "versioneer-518"]
+requires = [
+  "cmake==3.22.1",
+  "mypy==0.942",
```
mcwitt, line 30: Needed to add mypy here because we now use `stubgen` (part of mypy) in the CMake build to generate the stub file

.pre-commit-config.yaml
```diff
@@ -24,13 +24,35 @@ repos:
     -   id: flake8
         exclude: '^timemachine/cpp'
 
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.942
+    hooks:
+    -   id: mypy
+        additional_dependencies:
+          - "numpy==1.21.4"
```
mcwitt, line 32: Note: required for [numpy plugin](https://github.com/proteneer/timemachine/pull/711/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R19)

setup.py
```diff
@@ -104,6 +104,7 @@ def build_extension(self, ext):
             "isort==5.10.1",
             "mypy==0.942",
             "pre-commit==2.17.0",
+            "pybind11-stubgen==0.10.5",
```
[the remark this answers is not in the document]
mcwitt: I ended up switching to use `stubgen` bundled with mypy; unfortunately this doesn't include docstrings defined in the pybind11 wrapping code, but seemed more reliable in my testing (for example, pybind11-stubgen generated syntactically-invalid `.pyi` files with `timemachine::` in types)

timemachine/fe/estimator.py
```diff
@@ -321,13 +321,13 @@ def deltaG_from_results(
     else:
         sim_results = results
 
-    U_knk = []
+    U_knk_ = []
```
badisa, line 324: Does mypy enforce snake case?

timemachine/fe/model.py
```diff
@@ -54,9 +55,9 @@ def __init__(
         self.barostat_interval = barostat_interval
         self.pre_equilibrate = pre_equilibrate
         self.hmr = hmr
-        self._equil_cache = {}
+        self._equil_cache: Dict[str, Any] = {}
```
badisa, line 58: Is this a required change? Not clear why the others aren't typed but this is

Dockerfile
```diff
@@ -75,6 +75,10 @@ RUN pip install --no-cache-dir pre-commit==2.17.0
 COPY .pre-commit-config.yaml /code/timemachine/
 RUN cd /code/timemachine && git init . && pre-commit install-hooks
 
+# Install CI requirements
+COPY ci/requirements.txt /code/timemachine/ci/requirements.txt
```
badisa, line 79: If we are going to expand this dockerfile to support more CI logic, we should create a separate layer just for CI. Otherwise right now these ci requirements will get baked into production containers

timemachine/datasets/readme.md
```diff
@@ -3,5 +3,5 @@
 
 (Note: `freesolv.sdf` reflects a fix described in https://github.com/MobleyLab/FreeSolv/issues/48 )
 
-## `fep-benchmark`
-Datasets in the `fep-benchmark` folder are imported from v1.0 of https://github.com/MCompChem/fep-benchmark/releases/tag/v1.0 (distributed under MIT license), created by Christina Schindler and Daniel Kuhn, Merck KGaA, Darmstadt, Germany, December 2018 , accompanying https://chemrxiv.org/articles/preprint/Large-Scale_Assessment_of_Binding_Free_Energy_Calculations_in_Active_Drug_Discovery_Projects/11364884 (Schindler et al., 2019)
+## `fep_benchmark`
```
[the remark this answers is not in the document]
mcwitt: This was caught by mypy, but is a more general issue: `fep-benchmark` is invalid as a module name. I think this would have prevented us from accessing the data using importlib.

timemachine/lib/custom_ops.pyi
```diff
@@ -0,0 +1,150 @@
+from typing import List, Optional
```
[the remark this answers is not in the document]
mcwitt: > It seems like CMAKE will create this file on the fly?

Correct; the original reason to check it in was so it can be used by the mypy check during linting. But since we're now building the extension during linting anyway to verify that it's up to date, seems like we might be able to avoid checking it in. I'll look into this.

setup.py
```diff
@@ -94,15 +94,17 @@ def build_extension(self, ext):
         "pymbar>3.0.4",
         "pyyaml",
         "scipy",
+        "typing-extensions",
```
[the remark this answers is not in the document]
mcwitt: Done in df88b84

timemachine/lib/custom_ops.pyi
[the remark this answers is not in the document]
mcwitt: Another benefit of checking it in is to get typechecking and completions to work when developing the Python code on non-Linux machines. Not sure how common of a use case that is, though.. It seems like most of us are either developing locally on Linux, or running a remote IDE server on a Linux box?

@maxentile @proteneer @jkausrelay 
 do you know if having the stub file locally be useful for your dev setups?

Dockerfile
badisa: Should be relatively straight forward, just replicate the 'final' docker stage and run the two CI specific commands within it and change the docker target in CI.

Seems best to do it in this PR, unless it isn't as trivial as I am imagining

timemachine/md/barostat/utils.py
```diff
@@ -34,18 +35,18 @@ def get_bond_list(harmonic_bond_potential: HarmonicBond) -> List[Tuple[int, int]
         or if there are valence bonds not represented as harmonic bonds (e.g. as length constraints)
     """
 
-    bond_list = list(map(tuple, harmonic_bond_potential.get_idxs()))
+    bond_list = [(i, j) for i, j in harmonic_bond_potential.get_idxs()]
```
badisa, line 38: Is this a mypy change?

mcwitt: Yes, mypy isn't able to infer the type of the tuples. The error was:

> error: Incompatible return value type (got "List[Tuple[_T_co, ...]]", expected "List[Tuple[int, int]]")

This is fixed by making the tuple construction explicit.

.gitlab-ci.yml
badisa: Looks like this is no longer required, since it is in the docker container?

Dockerfile
mcwitt: Good call, done in 5d36eb4 (and fixed ordering in ba2d125)

.gitlab-ci.yml
mcwitt: Good catch; removed in 829e253

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -158,15 +158,6 @@ void declare_context(py::module &m) {
             py::arg("store_du_dl_interval") = 0,
             py::arg("store_x_interval") = 0)
         // .def("multiple_steps", &timemachine::Context::multiple_steps)
-        .def(
-            "get_x_t",
```
maxentile, line 162: Nice -- didn't realize this was duplicated!

timemachine/fe/model_rabfe.py
```diff
@@ -17,17 +18,17 @@
 class AbsoluteModel(ABC):
     def __init__(
         self,
-        client: AbstractClient or None,
+        client: Optional[AbstractClient],
         ff: Forcefield,
         host_system: openmm.System,
-        host_schedule: np.ndarray,
+        host_schedule: List[float],
```
maxentile, line 24: Mild surprise: I thought this was often an array

timemachine/fe/utils.py
```diff
@@ -233,7 +233,7 @@ def get_connected_components(nodes, relative_inds, absolute_inds) -> List[np.arr
         return filtered_components
 
 
-def validate_map(n_nodes: int, relative_inds: np.array, absolute_inds: np.array) -> bool:
+def validate_map(n_nodes: int, relative_inds: List[Tuple[int, int]], absolute_inds: List[int]) -> bool:
```
maxentile, line 236: Tangent / curiosity: It's often convenient to use array containers and python containers interchangeably in this sort of code.

Is there a more generic type for this situation? (should accept: int array of shape (n_pairs, 2), a list of 2-tuples, etc... Initial thought is something like `Union[Array, List[Tuple[int, int]]]`, but that seems counterproductive since more verbose, and since array typing is less precise)

maxentile (approved):
The changes this suggests mostly look like improvements to me!

timemachine/fe/model_rabfe.py
mcwitt: Good call; I don't think we have any cases of passing an array in the timemachine code, but there are plenty of examples in the tests (we don't currently run mypy on tests; this seemed like maybe too much overhead).

I think a good general approach might be to use the most general type that has the capabilities exercised by the implementation. In this case, we use `__getitem__` and `__len__`, which suggests [`Sequence`](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes) might be appropriate. `numpy.ndarray` does not implement one of the required methods of Sequence (`.index()`), but we don't use that anyway. So, an appropriate type here might be something like `Union[Sequence[float], NDArray]`.

mcwitt: Changed this in ff5c69b

badisa (approved):

timemachine/fe/utils.py
maxentile: Nice: `Iterable[Tuple[int, int]]` seems both generic and precise

>(I think this sort of generalization might be best done as an iterative process; trying to come up with the most general types initially is a fair amount of work and error-prone.

Agreed!

PR #712, opened 2022-04-09
Task/nightly tests

badisa:
This might be a bit late in deboggling week to consider, but putting it up as its pretty straight forward.


badisa:
Triggered the nightly pipeline here: https://gitlab.com/relaytx/physics-platform/timemachine/-/pipelines/512836136

Will need to modify the schedule to point at master after this is merged, for now it is pointing at this branch

tests/test_vacuum_importance_sampling.py
```diff
@@ -87,7 +87,7 @@ def get_torsion(x_l):
     vanilla_samples = enhanced.sample_from_log_weights(weighted_samples, log_weights, 100000)
 
     vanilla_torsions = batch_torsion_fn(vanilla_samples).reshape(-1)
-    vanilla_samples_lhs, _ = np.histogram(vanilla_torsions, bins=50, range=(-np.pi, 0), density=True)
+    vanilla_samples_lhs, _ = np.histogram(vanilla_torsions, bins=50, range=(0, np.pi), density=True)
```
maxentile, line 90: nit: naming: does _lhs / _rhs here mean [-pi, 0] / [0, +pi] ?

tests/test_mtm_condensed.py
```diff
@@ -121,10 +120,11 @@ def get_torsion(x_l):
         ubps, params, masses, coords, box, temperature, pressure, num_equil_steps, seed
     )
 
+    lamb = 0.0
     # test with both frozen masses and free masses
     for test_masses in [frozen_masses, masses]:
 
-        # (ytz): emprically scanning over multiple Ks seem to suggest 50 is a sweet spot
+        # (ytz): emprically scanning over multiple Ks seem to suggest 100 is a sweet spot
```
maxentile, line 127: Curiosity: Is this from a using a different scan approach?

tests/test_examples.py
```diff
@@ -0,0 +1,102 @@
...
+
+def test_relative_binding():
+    """
+    Test validate_relative_binding.py to ensure that it functions correctly
+    """
+    protein_path = str(Path(__file__).resolve().parent / "data" / "hif2a_nowater_min.pdb")
+    temp_ligands = get_hif2a_ligands_as_sdf_file(1)
+
+    # Any fewer windows and the lambda schedule validation will fail
+    windows = str(5)
```
maxentile, line 66: Surprising...

Should we modify the `> 0` to ` >= 0`? https://github.com/proteneer/timemachine/blob/56d9933d208b07bfa0524376086f27a26a468f82/timemachine/fe/lambda_schedule.py#L32-L37

tests/test_vacuum_importance_sampling.py
proteneer: yes

tests/test_examples.py
badisa: If I remember correctly it was the linear interpolation that allows the endpoints to not be 0.0 and 1.0

tests/test_mtm_condensed.py
badisa: This was from talking to @proteneer about the inconsistency between the comment and the code. So no change besides correcting the comment

examples/readme.md
```diff
@@ -1,25 +1,4 @@
 ### Examples
 
-#### MD
```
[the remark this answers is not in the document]
badisa: Woops, I thought I had, but looks like I lost those instead

tests/test_examples.py
maxentile: >If I remember correctly it was the linear interpolation that allows the endpoints to not be 0.0 and 1.0

Should I move this to a new issue?

Locally it looks like this is a problem with the methods that use `linspace` and `concatenate`:
* construct_absolute_lambda_schedule_solvent
* construct_absolute_lambda_schedule_complex
* construct_relative_lambda_schedule

but not with the method that uses `interp`
* construct_pre_optimized_absolute_lambda_schedule_solvent

badisa: Ah that is the same for me, was wrong about which schedule it was.

A new issue seems fine, though I don't know we have much utility for tiny numbers of windows.

maxentile: Will open a new issue in a second!

maxentile: Moved to https://github.com/proteneer/timemachine/issues/714

proteneer (approved):
nice!

PR #718, opened 2022-04-11
SMC for absolute hydration

maxentile:
Still requires some work before ready for review -- will tag reviewers in the next few days after I

- [x] add SMC tests
   - [x] unit tests: assert accuracy on tractable test system, test utility functions
   - [x] nightlies: run complete calculations with custom_ops systems
- [x] double-check manual resolutions of a merge conflict
- [x] add documentation throughout
- [x] add two new example scripts to nightly tests


timemachine/fe/absolute_hydration.py
```diff
@@ -0,0 +1,115 @@
...
+    return all_xvbs
+
+
+def get_solvent_phase_system(mol, ff):
+    water_system, water_coords, water_box, water_topology = builders.build_water_system(3.0)
+    water_box = water_box + np.eye(3) * 0.5  # add a small margin around the box for stability
+    afe = free_energy.AbsoluteFreeEnergy(mol, ff)
+    ff_params = ff.get_ordered_params()
+    ubps, params, masses, coords = afe.prepare_host_edge(ff_params, water_system, water_coords)
+    return ubps, params, masses, coords, water_box
```
maxentile, line 46: Can probably discard this one -- superseded by implementation in md/enhanced.py -- thanks to @mcwitt for taking a look!

maxentile: Applied in https://github.com/proteneer/timemachine/pull/718/commits/65dbef2dc35caac29dc2be3d2993d97dd36d6fd6

timemachine/md/enhanced.py
```diff
@@ -422,41 +435,50 @@ def jax_sample_from_log_weights(weighted_samples, log_weights, size, key):
...
-
-    new_water_coords = minimizer.minimize_host_4d([mol], water_system, water_coords, ff, water_box)
-    coords = np.concatenate([new_water_coords, ligand_coords])
-
-    return ubps, params, masses, coords, water_box
+    potentials, params, masses = afe.prepare_host_edge(ff_params, water_system)
 
+    # concatenate (optionally minimized) water_coords and ligand_coords
+    ligand_coords = get_romol_conf(mol)
+    if minimize_energy:
```
jkausrelay, line 471: Nice, I did the same thing in my fork.

timemachine/md/enhanced.py
```diff
@@ -553,3 +575,90 @@ def aligned_batch_propose(xvb, K, key, vacuum_samples, vacuum_log_weights):
...
+    mol_hash = hash_mol(mol)
+    # TODO: do I want this to be sensitive to other mol properties? conformer?
+
+    # hash all of the other parameters
+    # ff_hash = hash(tuple(np.array(ff.get_ordered_params()).flatten()))
+    # arg_hash = hash((seed, n_solvent_samples, n_ligand_batches, temperature, pressure))
+    # TODO: do I want this to be sensitive to ff's smirks strings too?
+    # TODO: store and check ff_hash, arg_hash match the one stored...
+
+    cache_path = PATH_TO_SAMPLE_CACHES / f"mol_{mol_hash}_x_solvent_samples.pkl"
```
jkausrelay, line 607: Why is a cache needed here? In my version of run_smc_on_freesolv, I have it skip the generation of save_smc_result pkl file is present, making it restartable.

In general, I think it's better to prevent the library functions from side effects (writing files). If there needs to be a cache it's better if it's managed on the application level.

timemachine/md/enhanced.py
```diff
@@ -553,3 +575,90 @@ def aligned_batch_propose(xvb, K, key, vacuum_samples, vacuum_log_weights):
...
+    n_samples,
+    num_equil_steps=50000,
+    md_steps_per_move=1000,
+):
+    """Discard num_equil_steps of MD, then return n_samples each separated by md_steps_per_move"""
+    xvb0 = equilibrate_solvent_phase(
+        potentials, params, masses, coords, box, temperature, pressure, num_equil_steps, seed
+    )
+
+    lamb = 1.0  # non-interacting state
```
jkausrelay, line 646: If the state is non-interacting why is the ligand needed? Couldn't the same solvent box be re-used for all of the simulations?

timemachine/md/smc.py
```diff
@@ -0,0 +1,159 @@
...
+        resampled = [sample_traj[-1][i] for i in indices]
+
+        # propagate
+        samples = propagate(resampled, lam_target)
+
+        # log
+        accumulate_results(samples, indices, log_weights, incremental_log_weights)
+
+    # final result: a collection of samples, with associated log weights
+    incremental_log_weights = log_prob(samples, lambdas[-1]) - log_prob(samples, lambdas[-2])
```
jkausrelay, line 103: samples here are collected from lambda[-2]. Should propagate be run at lambdas[-1] to add to sample_traj? (Depends on answer to above question)

timemachine/md/smc.py
```diff
@@ -0,0 +1,159 @@
...
+    def accumulate_results(samples, indices, log_weights, incremental_log_weights):
+        sample_traj.append(samples)
+        ancestry_traj.append(indices)
+        log_weights_traj.append(np.array(log_weights))
+        incremental_log_weights_traj.append(np.array(incremental_log_weights))
+        running_estimate = -logsumexp(log_weights - np.log(len(log_weights)))
+        trange.set_postfix(EXP=running_estimate)
+
+    # main loop
+    for (lam_initial, lam_target) in zip(trange, lambdas[1:-1]):
```
jkausrelay, line 87: Should this be zip(lambdas[:-1], lambdas[1:])?

examples/run_smc_on_biphenyl.py
```diff
@@ -0,0 +1,59 @@
...
+
+    if cmd_args.debug_mode:
+        n_frames = cmd_args.n_walkers * cmd_args.n_windows
+        print(f"Warning! This will take a lot of disk space! ({n_frames} simulation frames)")
+
+    return cmd_args
+
+
+def save_smc_result(smc_result, save_full_trajectories=False):
+    uid = f"{datetime.now()}"
```
mcwitt, line 28: nit: if this is to be used in a filename, a format without spaces might be easier to deal with, e.g.
```python
datetime.now().isoformat(timespec="seconds")
```
or
```python
datetime.now().strftime("%Y%m%dT%H%M%S"))
```

tests/test_mtm.py
```diff
@@ -46,9 +46,10 @@ def test_optimized_MTM():
     proposal_U = state.U_decharged
     num_batches = 485
 
-    vacuum_samples, vacuum_log_weights = enhanced.generate_log_weighted_samples(
+    _vacuum_xv_samples, vacuum_log_weights = enhanced.generate_log_weighted_samples(
         mol, temperature, state.U_easy, proposal_U, seed, num_batches=num_batches
     )
+    vacuum_samples = _vacuum_xv_samples[:, 0, :]
```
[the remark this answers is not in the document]
maxentile: Added in https://github.com/proteneer/timemachine/pull/718/commits/8a715b740ff382a7f14c48fcfc81394ac30604b4

timemachine/fe/absolute_hydration.py
```diff
@@ -0,0 +1,143 @@
...
+    solvent_idxs = np.arange(0, num_solvent_atoms)
+
+    # sample according to log weights
+    ligand_xvs = enhanced.sample_from_log_weights(ligand_samples, ligand_log_weights, size=num_samples)
+
+    # sample uniformly with replacement
+    solvent_choice_idxs = np.random.choice(np.arange(len(solvent_samples)), size=num_samples)
+
+    all_xvbs = []
+    for i, choice_idx in enumerate(tqdm(solvent_choice_idxs, desc="generating endstate samples")):
```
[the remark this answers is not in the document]
maxentile: Removed in https://github.com/proteneer/timemachine/pull/718/commits/cf9d7654295faf0b3763c0946188bbfe30a45ca1

timemachine/md/enhanced.py
maxentile: The same collection of solvent box samples can be reused for all of the simulations -- that would be cleaner

timemachine/md/smc.py
maxentile: They're proposed according to `p(x | lambda[-2])`, but weighted for the difference between `p(x | lambda[-1])` and `p(x | lambda[-2])`

timemachine/md/smc.py
maxentile: Could be shifted.

Right now it loops from 0 to -2, and then has a block outside the loop to compute importance weights from -2 to -1.

Could rearrange so that the loop goes from 1 to -1, and put a similar block before the loop.

examples/run_smc_on_biphenyl.py
mcwitt: This seems almost identical to `save_smc_result` in `run_smc_on_freesolv.py`. Could this take `uid` as an argument for consistency, and would it make sense to then merge?

maxentile: Applied in https://github.com/proteneer/timemachine/pull/718#discussion_r854309159

mcwitt: Another thought -- not necessary for this PR, but it seems like it might be nice for reproducibility to compute the output UID from the timemachine version and a hash of the inputs, rather than a timestamp. Would something like this make sense here?

timemachine/md/smc.py
proteneer: This doesn't seem to do a final propagation step though - so the final, end-states samples we gathered are for `lambda[-2]` as opposed to `lambda[-1]`. (Which is problematic for re-weighting since we need equilibrium samples for `lambda[-1]`)

maxentile: >This doesn't seem to do a final propagation step though - so the final, end-states samples we gathered are for lambda[-2] as opposed to lambda[-1].

The final propagation step here targets p(x | lambda[-2]), and then the final weighting step weights for the difference between p(x | lambda[-1]) and p(x | lambda[-2]).

The output of SMC is a weighted sample from p(x | lambda[-1]).

---------

>(Which is problematic for re-weighting since we need equilibrium samples for lambda[-1])

Summarizing from offline discussion: When we want to convert this into a collection of _unweighted_ samples from p(x | lambdas[-1]), we should: (1) resample (e.g. `multinomial_resample`, `stratiified_resample`, `systematic_resample`, ..., to go from weighted to unweighted at the cost of introducing degeneracy), (2) run short propagation targeting p(x | lambdas[-1]) (to mitigate the degeneracy introduced by the last step).

Notes:
* Appending `propagate(xs, lambdas[-1])` to the end of this function won't contribute to the estimate returned by SMC
* Simply appending `propagate(xs, lambdas[-1])` is insufficient to discard the weights
* (subtle point) Appending `resample` and `propagate` is insufficient to discard the weights, since the `resample` function used in the main loop could be optional. (E.g. if the user supplied `null_resample` or `conditional_multinomial_resample`)
* It may be best for this to be handled in application code, or perhaps to add a utility `get_endstate_samples_from_smc_result` to this PR

maxentile: Utility added in https://github.com/proteneer/timemachine/pull/718/commits/ee0616f895e788c06bfcb6b102be059ed79c9df5

maxentile: Utility tested in https://github.com/proteneer/timemachine/pull/718/commits/4b914d94f04a3f1cfabb8da6ceaca8497c59e635

timemachine/fe/absolute_hydration.py
```diff
@@ -0,0 +1,142 @@
...
+from timemachine.constants import BOLTZ
+from timemachine.fe.lambda_schedule import construct_pre_optimized_absolute_lambda_schedule_solvent
+from timemachine.ff import Forcefield
+from timemachine.md import enhanced, moves
+from timemachine.md.smc import conditional_multinomial_resample
+from timemachine.md.states import CoordsVelBox
+from timemachine.utils import bind_potentials, construct_potential
+
+
+def generate_endstate_samples(num_samples, solvent_samples, ligand_samples, ligand_log_weights, num_ligand_atoms):
```
mcwitt, line 16: nit: worth adding types here? e.g. is `solvent_samples: Sequence[CoordsVelBox]`?

timemachine/fe/absolute_hydration.py
```diff
@@ -0,0 +1,142 @@
...
+
+    Notes
+    -----
+    * TODO[generality]: refactor to accept two streams of unweighted samples, concatenate them
+        (rather than requiring and discarding ligand component in solvent_samples, performing multinomial resampling)
+    """
+
+    # assume this layout
+    num_total_atoms = len(solvent_samples[0].coords)
+    num_solvent_atoms = num_total_atoms - num_ligand_atoms
```
mcwitt, line 48: Would it be useful to `assert num_total_atoms > num_ligand_atoms`

timemachine/fe/absolute_hydration.py
```diff
@@ -0,0 +1,142 @@
...
+    solvent_idxs = np.arange(0, num_solvent_atoms)
+
+    # sample according to log weights
+    ligand_xvs = enhanced.sample_from_log_weights(ligand_samples, ligand_log_weights, size=num_samples)
+
+    # sample uniformly with replacement
+    solvent_choice_idxs = np.random.choice(np.arange(len(solvent_samples)), size=num_samples)
```
mcwitt, line 55: nit: simplify to
```suggestion
    solvent_choice_idxs = np.random.choice(len(solvent_samples), size=num_samples)
```

timemachine/fe/absolute_hydration.py
```diff
@@ -0,0 +1,142 @@
...
+    potential_fxn = construct_potential(potentials, params)
+
+    kBT = BOLTZ * temperature
+
+    def reduced_potential_fxn(xvb, lam):
+        return potential_fxn(xvb, lam) / kBT
+
+    bind_potentials(potentials, params)
+
+    # set up npt mover
```
mcwitt, line 104: nit: redundant comment

timemachine/fe/absolute_hydration.py
```diff
@@ -0,0 +1,142 @@
...
+def set_up_ahfe_system_for_smc(mol, n_walkers, n_windows, n_md_steps, resample_thresh, seed=2022):
+    """define initial samples, lambdas schedule, propagate fxn, log_prob fxn, resample fxn"""
+    reduced_potential, mover, initial_samples = setup_absolute_hydration_with_endpoint_samples(
+        mol, n_steps=n_md_steps, seed=seed
+    )
+
+    sample_inds = np.random.choice(np.arange(len(initial_samples)), size=n_walkers)
+    samples = [initial_samples[i] for i in sample_inds]
+
+    lambdas = construct_pre_optimized_absolute_lambda_schedule_solvent(n_windows)[::-1]
```
mcwitt, line 129: worth commenting on why this is reversed?

timemachine/integrator.py
```diff
@@ -161,11 +161,13 @@ def multiple_steps(x0, v0, key0):
...
 
     for batch_step in range(num_batches):
         #                                             [B,N,3][B,N,3][B,2]
         xs_t, vs_t, keys_t = batched_multiple_steps_fn(xs_t, vs_t, keys_t)
         all_xs.append(xs_t)
+        all_vs.append(vs_t)
 
     # result has shape [num_workers, num_batches, num_atoms, num_dimensions]
-    return np.transpose(np.array(all_xs), axes=[1, 0, 2, 3])
+    return np.transpose(np.array(all_xs), axes=[1, 0, 2, 3]), np.transpose(np.array(all_vs), axes=[1, 0, 2, 3])
```
mcwitt, line 173: nit: `swapaxes` seems slightly simpler:
```suggestion
    return np.swapaxes(np.array(all_xs), 0, 1), np.swapaxes(np.array(all_vs), 0, 1)
```

timemachine/md/enhanced.py
```diff
@@ -553,3 +574,90 @@ def aligned_batch_propose(xvb, K, key, vacuum_samples, vacuum_log_weights):
...
+    mol_hash = hash_mol(mol)
+    # TODO: do I want this to be sensitive to other mol properties? conformer?
+
+    # hash all of the other parameters
+    # ff_hash = hash(tuple(np.array(ff.get_ordered_params()).flatten()))
+    # arg_hash = hash((seed, n_solvent_samples, n_ligand_batches, temperature, pressure))
+    # TODO: do I want this to be sensitive to ff's smirks strings too?
+    # TODO: store and check ff_hash, arg_hash match the one stored...
+
+    cache_path = PATH_TO_SAMPLE_CACHES / f"mol_{mol_hash}_x_solvent_samples.pkl"
```
mcwitt, line 606: Is it possible to generate a hash of the force field to include in the cache key?

timemachine/md/enhanced.py
mcwitt: > In general, I think it's better to prevent the library functions from side effects

:+1: this in general (though don't have a completely clear understanding of the use case here, yet..)

timemachine/md/enhanced.py
```diff
@@ -553,3 +574,90 @@ def aligned_batch_propose(xvb, K, key, vacuum_samples, vacuum_log_weights):
...
+    for _ in range(n_samples):
+        xvbs.append(npt_mover.move(xvbs[-1]))
+    return xvbs
+
+
+def generate_ligand_samples(num_batches, mol, ff, temperature, seed):
+    """Generate (weighted) samples of the ligand in vacuum, by importance sampling from a less-hindered state where
+    torsions and intramolecular nonbonded terms are disabled"""
+    state = VacuumState(mol, ff)
+    proposal_U = state.U_full
```
mcwitt, line 658: Isn't `U_full` the target distribution (not proposal)?

timemachine/md/smc.py
```diff
@@ -0,0 +1,179 @@
...
+from scipy.special import logsumexp
+from typing_extensions import TypeAlias
+
+# type annotations
+Sample: TypeAlias = Any
+Samples: TypeAlias = List[Sample]  # e.g. List[CoordsVelBox]
+
+Lambda: TypeAlias = float
+LogWeight: TypeAlias = float
+Array: TypeAlias = np.ndarray
```
mcwitt, line 15: nit: should we prefer `numpy.typing.NDArray`?

timemachine/md/smc.py
```diff
@@ -0,0 +1,179 @@
...
+IndexArray: TypeAlias = Array
+LogWeights: TypeAlias = Array
+
+BatchPropagator = Callable[[Samples, Lambda], Samples]
```
mcwitt, line 19: nit: why use `TypeAlias` annotations for some aliases and not others?

timemachine/md/smc.py
```diff
@@ -0,0 +1,179 @@
...
+
+from typing import Any, Callable, Dict, List, Tuple
+
+import numpy as np
+from scipy.special import logsumexp
+from typing_extensions import TypeAlias
+
+# type annotations
+Sample: TypeAlias = Any
+Samples: TypeAlias = List[Sample]  # e.g. List[CoordsVelBox]
```
mcwitt, line 11: Curious to understand the motivation for this alias, vs. just using `List[Samples]` directly. Is the idea to abstract over the container type?

timemachine/md/smc.py
```diff
@@ -0,0 +1,179 @@
...
+    incremental_log_weights_traj = []  # note: redundant but convenient
+
+    def accumulate_results(samples, indices, log_weights, incremental_log_weights):
+        sample_traj.append(samples)
+        ancestry_traj.append(indices)
+        log_weights_traj.append(np.array(log_weights))
+        incremental_log_weights_traj.append(np.array(incremental_log_weights))
+
+    # main loop
+    for (lam_initial, lam_target) in zip(lambdas[:-2], lambdas[1:-1]):
```
mcwitt, line 82: This is a bit unintuitive to me. Why don't we target `lambdas[-1]`?

timemachine/md/smc.py
```diff
@@ -0,0 +1,179 @@
...
+    log_weights_traj.append(np.array(log_weights + incremental_log_weights))
+
+    # cast everything (except samples list) to arrays
+    trajs_dict = dict(
+        traj=sample_traj,
+        log_weights_traj=np.array(log_weights_traj),
+        ancestry_traj=np.array(ancestry_traj),
+        incremental_log_weights_traj=np.array(incremental_log_weights_traj),
+    )
+    return trajs_dict
```
mcwitt, line 109: nit: should we prefer returning results as `NamedTuple`s vs `dict`s in general? (this would allow for more static checking)

timemachine/md/smc.py
```diff
@@ -0,0 +1,179 @@
...
+
+def conditional_multinomial_resample(log_weights, thresh=0.5):
+    """if fractional_effective_sample_size(log_weights) < thresh, then multinomial_resample"""
+    if fractional_effective_sample_size(log_weights) < thresh:
+        return multinomial_resample(log_weights)
+    else:
+        return null_resample(log_weights)
+
+
+def refine_samples(samples, log_weights, propagate, lam):
```
mcwitt, line 157: nit: annotation for `propagate` seems useful here (and in general for arguments of function type)

timemachine/utils.py
```diff
@@ -0,0 +1,28 @@
...
+from timemachine.fe import functional
+
+
+def bind_potentials(potentials, params):
+    """modifies potentials in-place"""
+    for U, p in zip(potentials, params):
+        U.bind(p)
+
+
+def construct_potential(potentials, params):
```
mcwitt, line 14: nit: type annotation (and maybe a more descriptive name) seems particularly helpful here

timemachine/md/enhanced.py
maxentile: >Why is a cache needed here?

Cost. For re-running multiple times with the same mol (but different SMC settings etc.), the cost of generating ligand vacuum samples was comparable to or exceeded the cost of the hydration free energy calculation (~3 minutes).

>In general, I think it's better to prevent the library functions from side effects (writing files). If there needs to be a cache it's better if it's managed on the application level.

Good point -- will remove

maxentile: Removed in https://github.com/proteneer/timemachine/pull/718/commits/fea2af2b93d4b18cb6d53be11d7562dd7139d72e , can be handled better elsewhere

timemachine/fe/absolute_hydration.py
maxentile: Removed in https://github.com/proteneer/timemachine/pull/718/commits/3a99dc756c6701dc9e685c6d123c3f378c7e7e3e

timemachine/md/enhanced.py
maxentile: Removed for now in https://github.com/proteneer/timemachine/pull/718/commits/fea2af2b93d4b18cb6d53be11d7562dd7139d72e , but I think that would be a good idea

timemachine/md/smc.py
maxentile: Only added where needed for mypy. The aliases need this annotation, but the `Callable` type declaration does not, I think

timemachine/md/smc.py
maxentile: I preferred to use a dict (with simple key and value types) over a namedtuple here for pickle convenience, esp. while the code is rapidly changing.

Can pickle with this version of timemachine, and if I change my mind about the field names or add fields or whatever in a later PR, I may still be able to load pickle files written with older versions (or different branches) of timemachine. However, if I make similar changes to the namedtuple definition between versions, reading / writing pickles between versions would be less convenient.

timemachine/md/smc.py
maxentile: Mostly to emphasize that the signatures here are all in terms of collections rather than individual samples.

In the future, we may want to use a specialized container for a collection of samples. (Something like an mdtraj Trajectory object, or for toy systems a raw numpy array, or in local MD maybe a new container specialized to the case where most atoms are held constant and only a small number of atoms are mobile.) There may be optimizations that apply to propagating (or computing the potential energy of) a batch of samples, that are not applicable to individual samples.

timemachine/md/smc.py
maxentile: Oh! I was confused about whether this was necessary. To fix the mypy error I had done two things at once: put declarations on separate lines, and annotate with :TypeAlias https://github.com/proteneer/timemachine/pull/718/commits/78e8edabfba371c27b41d305822895d9548e9bcf . However, it looks like all of the :TypeAlias annotations may have been unnecessary, and I've now removed them in: https://github.com/proteneer/timemachine/pull/718/commits/5baafa71d9e482035fc7388da3d6063d741615fd

timemachine/md/smc.py
maxentile: Clarifications from offline discussion:
* changing the _class name_ of the namedtuple between versions versions used for pickle.dump and pickle.load would cause pickle.load to fail
* changing the _number_ of fields in the namedtuple between versions would cause pickle.load to fail
* if only the _names_ of fields changed, pickle.load would not have any errors
* if only the _order_ of fields changed, pickle.load would not have any errors, but would invoke the namedtuple constructor with arguments in the wrong order

`Dict[str, array]`, while less amenable to verification, seems more convenient for allowing pickle to "just work" between timemachine versions that are likely to add/remove/reorder fields.

In offline discussion, @mcwitt suggested possibly using pydantic or another tool for schema versioning, which I think would be a safer and better long-term solution for situations like this.

timemachine/md/smc.py
maxentile: Agreed -- annotations added in https://github.com/proteneer/timemachine/pull/718/commits/5baafa71d9e482035fc7388da3d6063d741615fd

timemachine/utils.py
maxentile: On second thought, seems better to just inline this in https://github.com/proteneer/timemachine/pull/718/commits/07e6c98310e6bb413670543b3b2063a633d8c543 , since only used once

timemachine/md/enhanced.py
maxentile: Doh, fixed in https://github.com/proteneer/timemachine/pull/718/commits/45ca591324edc1c481fb0a0b7881cbd324af11a9

timemachine/fe/absolute_hydration.py
maxentile: comment added in https://github.com/proteneer/timemachine/pull/718/commits/18ccc6997b339fed02220b4194623ca68b090b64

timemachine/md/smc.py
jkausrelay: Can you add a comment here, in case I forget our discussion?

jkausrelay: Can you add a comment here, in case I forget our discussion?

jkausrelay (approved):
Minor comment, LGTM!

timemachine/md/smc.py
maxentile: Good call -- added comment and cross-references in https://github.com/proteneer/timemachine/pull/718/commits/272241d15cd37214d6cd9d662bdcd5492e067416

tests/test_examples.py
```diff
@@ -100,3 +106,102 @@ def test_relative_binding():
...
+        _ = run_example("run_smc_on_biphenyl.py", cli_args, cwd=temp_dir)
+        smc_result_fnames = glob(output_path)
+        assert len(smc_result_fnames) == 1
+
+        # load result
+        with open(smc_result_fnames[0], "rb") as f:
+            smc_result = load(f)
+
+        # expect no NaNs in accumulated log weights
+        log_weights_traj = smc_result["log_weights_traj"]
```
jkausrelay, line 133: incremental_log_weights_traj, is log_weights_traj still being stored?

tests/test_examples.py
```diff
@@ -100,3 +106,102 @@ def test_relative_binding():
...
+
+    experimental_dGs = {mol.GetProp("_Name"): float(mol.GetProp("dG")) for mol in fetch_freesolv()}
+
+    def get_predicted_dG(fname):
+        """in kcal/mol"""
+        with open(fname, "rb") as f:
+            smc_result = load(f)
+
+        # expect no NaNs in accumulated log weights
+        log_weights_traj = smc_result["log_weights_traj"]
```
jkausrelay, line 161: same here, double check that this corresponds to what you are saving in the pkl

tests/test_examples.py
maxentile: I think this is correct as-is.

Confirmed `log_weights_traj` is being stored https://github.com/proteneer/timemachine/blob/4b914d94f04a3f1cfabb8da6ceaca8497c59e635/timemachine/md/smc.py#L105 , and I think it's slightly more stringent to check for nans in the accumulated log weights rather than the incremental log weights.

maxentile: Thanks for catching this -- fixed in https://github.com/proteneer/timemachine/pull/718/commits/7879e6768798dc8fab6f8dc0bb44bac35725e776

tests/test_examples.py
```diff
@@ -100,3 +106,103 @@ def test_relative_binding():
...
+
+        # expect no NaNs in incremental log weights
+        incremental_log_weights_traj = smc_result["incremental_log_weights_traj"]
+        assert np.isfinite(incremental_log_weights_traj).all()
+
+        # expect delta_f in ballpark of 0
+        final_log_weights = smc_result["final_log_weights"]
+        final_weights = np.exp(final_log_weights)
+        delta_f = -np.log(np.mean(final_weights))
+        assert (delta_f >= -10) and (delta_f <= +10)
```
mcwitt, line 140: nit: could this be
```suggestion
        assert np.abs(delta_f) <= 10

```

examples/run_smc_on_biphenyl.py
```diff
@@ -0,0 +1,59 @@
...
+from timemachine.testsystems.ligands import get_biphenyl
+
+
+def parse_options():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--n_walkers", type=int, help="number of walkers", default=100)
+    parser.add_argument("--n_windows", type=int, help="number of lambda windows", default=100)
+    parser.add_argument("--n_md_steps", type=int, help="number of MD steps per move", default=100)
+    parser.add_argument("--resample_thresh", type=float, help="resample when fractional ESS < thresh", default=0.6)
+    parser.add_argument("--debug_mode", type=bool, help="save full trajectories", default=False)
```
mcwitt, line 16: should this accept a random seed argument? This seems at least useful for making `test_examples.py` deterministic.

examples/run_smc_on_freesolv.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def parse_options():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--n_walkers", type=int, help="number of walkers", default=100)
+    parser.add_argument("--n_windows", type=int, help="number of lambda windows", default=100)
+    parser.add_argument("--n_md_steps", type=int, help="number of MD steps per move", default=100)
+    parser.add_argument("--resample_thresh", type=float, help="resample when fractional ESS < thresh", default=0.6)
+    parser.add_argument("--debug_mode", type=bool, help="save full trajectories", default=False)
+    parser.add_argument("--n_mols", type=int, help="how many freesolv molecules to run on", default=10)
```
mcwitt, line 23: same here, should we allow setting the seed?

tests/test_examples.py
```diff
@@ -1,13 +1,19 @@
 import os
 import subprocess
 import sys
+from glob import glob
 from pathlib import Path
+from pickle import load
```
mcwitt, line 6: nit: slight preference for qualified import here (`import pickle`), since `load` is a generic-sounding name

tests/test_smc.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # random weights
+    n_instances = 100
+    for _ in range(n_instances):
+        n_particles = np.random.randint(1, 100)
+        log_weights = generate_log_weights(n_particles)
+        ess = effective_sample_size(log_weights)
+        fractional_ess = fractional_effective_sample_size(log_weights)
+
+        np.testing.assert_almost_equal(fractional_ess, ess / n_particles)
```
mcwitt, line 129: nit: I'm not sure this test is really adding much, since it seems like it's just reimplementing the logic in `fractional_effective_sample_size` (dividing by `n`)?

tests/test_examples.py
maxentile: Agreed -- changed in new files in https://github.com/proteneer/timemachine/pull/718/commits/35bb6b37b036c166881ed38ee0596c04c550b5e0.

Note 1 other usage (in [`fe/model.py`](https://github.com/proteneer/timemachine/blob/c1a64ddfce22e90424586b0b804d8733ae285f70/timemachine/fe/model.py#L2) ) we might want to change in a future PR

tests/test_smc.py
maxentile: Didn't have a good idea what else to test about this function... Maybe also just assert in bounds [0, 1] or something

tests/test_smc.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+@pytest.mark.parametrize("resampling_fxn", [null_resample, multinomial_resample, conditional_multinomial_resample])
+def test_sequential_monte_carlo(resampling_fxn: Resampler):
+    """Run SMC with the desired resampling_fxn on a Gaussian 1D test system, and assert that
+    * running estimates of the free energy as a fxn of lambda match analytical free energies
+    * endstate samples have expected mean and stddev
```
mcwitt, line 136: sentence fragment -- missing "and"?

timemachine/fe/absolute_hydration.py
```diff
@@ -0,0 +1,153 @@
...
+    num_total_atoms = len(solvent_samples[0].coords)
+    num_solvent_atoms = num_total_atoms - num_ligand_atoms
+    assert num_solvent_atoms > 0, "Oops, did you really mean num_ligand_atoms >= num_total_atoms?"
+    solvent_idxs = np.arange(0, num_solvent_atoms)
+
+    # sample according to log weights
+    ligand_xvs = enhanced.sample_from_log_weights(ligand_samples, ligand_log_weights, size=num_samples)
+
+    # sample uniformly with replacement
+    solvent_choice_idxs = np.random.choice(len(solvent_samples), size=num_samples)
```
mcwitt, line 64: nit: might be nice to pass `replace=True` here (even though it's apparently the default) for a more self-documenting style

examples/run_smc_on_biphenyl.py
maxentile: Would be useful -- added in https://github.com/proteneer/timemachine/pull/718/commits/ec5cf5c69cc393497b4a2e0c72fec0a69f4e8c1e

tests/test_examples.py
```diff
@@ -100,3 +106,103 @@ def test_relative_binding():
...
             assert isinstance(result.dG_bind_err, float)
+
+
+def test_smc_biphenyl():
+    """run_smc_on_biphenyl.py with reasonable settings, and expect
+    * output in summary_smc_result_*.pkl
+    * no NaNs in accumulated log weights
+    * delta_f in ballpark of 0
+    """
+    config = dict(n_walkers=100, n_windows=100, n_md_steps=100)
```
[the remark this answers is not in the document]
maxentile: I don't think so? The example script now sets the seed by default: https://github.com/proteneer/timemachine/blob/ec5cf5c69cc393497b4a2e0c72fec0a69f4e8c1e/examples/run_smc_on_biphenyl.py#L17

timemachine/fe/absolute_hydration.py
maxentile: Good call -- added here and at one other site https://github.com/proteneer/timemachine/pull/718/commits/63fb45f3814f9bdf91579c63200bef1060dc643c

tests/test_smc.py
mcwitt: Got it. I wonder if this function could be removed? To me, reading `effective_sample_size(log_weights) / len(log_weights)` seems clearer than `fractional_effective_sample_size(log_weights)`, but maybe "fractional effective sample size" is a well-understood term?

> Maybe also just assert in bounds [0, 1] or something

This also feels a bit redundant since we check that ess is between 0 and `n` above

timemachine/md/smc.py
```diff
@@ -0,0 +1,190 @@
...
+    trajs_dict = dict(
+        traj=sample_traj,
+        log_weights_traj=np.array(log_weights_traj),
+        ancestry_traj=np.array(ancestry_traj),
+        incremental_log_weights_traj=np.array(incremental_log_weights_traj),
+    )
+    return trajs_dict
+
+
+def null_resample(log_weights):
```
mcwitt, line 121: nit: something like `identity_resample` makes slightly more sense to me here

timemachine/md/smc.py
```diff
@@ -0,0 +1,190 @@
...
+def effective_sample_size(log_weights):
+    r"""Effective sample size, taking values in interval [1, len(log_weights)]
+
+    Notes
+    -----
+    * This uses the conventional definition ESS(w) = 1 / \sum_i w_i^2, which has some important known limitations!
+    * See [Elvira, Martino, Robert, 2018] "Rethinking the effective sample size" https://arxiv.org/abs/1809.04129
+        and references therein for some insightful discussion of limitations and possible improvements
+    """
+    weights = np.exp(log_weights - logsumexp(log_weights))
```
mcwitt, line 148: nit: would it make sense to call this something like `probs` or `normed_weights`?

mcwitt (approved):
Looks great! Mostly only minor nitpicks on the second pass. I'm excited to see this added to the tool box!

tests/test_smc.py
maxentile: Removed in https://github.com/proteneer/timemachine/pull/718/commits/787a6eb6a47f0fb8c5b84789e893bd2ed782fd65

tests/test_smc.py
maxentile: Replaced

```
assert that
* item1
* item2
* item3
```

with
```
assert that:
* item1,
* item2, and
* item3
```

in this file in https://github.com/proteneer/timemachine/pull/718/commits/af746deab1995f233fd247207eb046e7d23f5c3c (but may have missed other files)

tests/test_examples.py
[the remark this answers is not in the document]
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/718/commits/4281443093a06c8949826130af9584f8bdf94a4f

timemachine/md/smc.py
maxentile: Applied in https://github.com/proteneer/timemachine/pull/718/commits/36eb30d93f0d70df8daa23e33b129d69956c1304

timemachine/md/smc.py
maxentile: Applied in https://github.com/proteneer/timemachine/pull/718/commits/5ba052456ac800a0d6bd132f8e880f21cd982b94

tests/test_smc.py
```diff
@@ -0,0 +1,197 @@
...
+
+        proposal_scale = np.exp(lam * target_log_sigma)  # to match annealed_gaussian_def  # TODO: extract
+        proposals = xs + proposal_scale * np.random.randn(n)
+        assert proposals.shape == xs.shape
+
+        # metropolis
+        log_probs_before = log_prob(xs, lam)
+        log_probs_after = log_prob(proposals, lam)
+        log_accept_probs = np.minimum(log_probs_after - log_probs_before, 0)
+        accept_probs = np.exp(log_accept_probs)
```
proteneer, line 158: nice! `exp(min(x,0)) = min(exp(x),exp(0)) = min(exp(x),1)` was used to seeing the right hand side equation in wiki and had to pause for a second

timemachine/integrator.py
proteneer: how do you guys even discover these functions?

tests/test_examples.py
```diff
@@ -100,3 +106,103 @@ def test_relative_binding():
...
+        dG_expts = np.array(dG_expts)
+
+        # compute error summaries
+        mean_abs_err_kcalmol = np.mean(np.abs(dG_preds - dG_expts))
+
+        # expect small error
+        # * MAE of ~1.5 kcal/mol: run with these settings on 10 molecules from FreeSolv
+        # * MAE of ~1.5 kcal/mol: run with these settings on ~500 molecules from FreeSolv
+        # * MAE of ~1.1 kcal/mol: FreeSolv reference calculations
+        #   https://www.biorxiv.org/content/10.1101/104281v1.full
```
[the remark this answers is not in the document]
proteneer: ps: we should also revisit on how changing `config = dict(n_walkers=100, n_windows=100, n_md_steps=100, n_mols=10, seed=2022)` affects the result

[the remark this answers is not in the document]
maxentile: >Not a blocker for merging, but I wonder what the reference calculation's MAE is on the same ~500 molecules.

Haven't checked -- imagine close to ~1.5

>ps: we should also revisit on how changing config = dict(n_walkers=100, n_windows=100, n_md_steps=100, n_mols=10, seed=2022) affects the result

Agreed, esp. increasing `n_walkers`

maxentile:
Thanks for the reviews!

PR #719, opened 2022-04-12
Fix for empty bonded terms

jkausrelay:
- With enhanced sampling, some torsions are disabled
- For certain molecules (butane) all torsions are disabled leading to an error
- Do not initialize empty arrays on the gpu
- Skip reference potential term if no terms given


jkausrelay:
Confirmed this also fixes molecules that don't have torsions:
image

tests/test_bonded.py
```diff
@@ -178,7 +178,8 @@ def test_harmonic_bond(self, n_particles=64, n_bonds=35, dim=3):
 
             # test bitwise commutativity
             test_potential = potentials.HarmonicBond(bond_idxs, lamb_mult, lamb_offset)
-            test_potential_rev = potentials.HarmonicBond(bond_idxs[:, ::-1], lamb_mult, lamb_offset)
+            bond_idxs_rev = bond_idxs[:, ::-1] if n_bonds else bond_idxs
```
maxentile, line 181: Surprised the new conditional is needed, since

```python
>>> import numpy as np
>>> x = np.zeros(shape=(0, 2))
>>> (x == x[:, ::-1]).all()
True
```

Clarification: Is `bond_idxs.shape` not `(0, 2)` in the empty bonded terms case?

timemachine/potentials/bonded.py
```diff
@@ -115,6 +115,9 @@ def harmonic_bond(conf, params, box, lamb, bond_idxs, lamb_mult=None, lamb_offse
     * lamb argument is unused
 
     """
+    if bond_idxs.shape[0] == 0:
+        return 0.0
+
     assert params.shape == bond_idxs.shape
```
maxentile, line 121: Suggestion: Is it preferable to move this one to be right after the shape assertion?

proteneer (approved):
nice!

maxentile (approved):
Looks great!

Possible testing enhancement: The modified tests randomly generate bond_idxs etc., but don't exercise any of the other machinery for constructing potentials given a molecule. Should we add one of the molecules listed in https://github.com/proteneer/timemachine/pull/719#issuecomment-1097086143 to other parts of the test suite, to make sure the other components gracefully handle case of molecules containing no torsions? Specifically thinking of the ff handlers (should handle case of no torsion matches after https://github.com/proteneer/timemachine/pull/626 , but this could be tested better, and I don't remember if this already handles the case of no angle matches).

Possible further testing enhancements:
* Possibly also add a molecule to test the case of no angles, such as `smi='O=O'`.
* Possibly also add a molecule to test the case of no bonds, such as `smi=[Na+]`.

tests/test_bonded.py
jkausrelay: It was, I updated in [c6ea7e9](https://github.com/proteneer/timemachine/pull/719/commits/c6ea7e9b6810816ba18f44a9b4b1f5f8eb95822b)

timemachine/potentials/bonded.py
jkausrelay: Fixed in [c6ea7e9](https://github.com/proteneer/timemachine/pull/719/commits/c6ea7e9b6810816ba18f44a9b4b1f5f8eb95822b)

jkausrelay:
> Looks great!
> 
> Possible testing enhancement: The modified tests randomly generate bond_idxs etc., but don't exercise any of the other machinery for constructing potentials given a molecule. Should we add one of the molecules listed in [#719 (comment)](https://github.com/proteneer/timemachine/pull/719#issuecomment-1097086143) to other parts of the test suite, to make sure the other components gracefully handle case of molecules containing no torsions? Specifically thinking of the ff handlers (should handle case of no torsion matches after #626 , but this could be tested better, and I don't remember if this already handles the case of no angle matches).
> 
> Possible further testing enhancements:
> 
> * Possibly also add a molecule to test the case of no angles, such as `smi="O=O"`.
> * Possibly also add a molecule to test the case of no bonds, such as `smi="[Na+]"`.

Added bond/angle handler test and updated the handlers to return correct shape for empty idxs

maxentile:
Nice -- these additions greatly improve the thoroughness of the tests!

maxentile (approved):

Issue #717, opened 2022-04-11
Modify some uses of jax numpy -> original numpy


maxentile:
Esp. for index arrays -- see examples identified by @proteneer  in https://github.com/proteneer/timemachine/pull/716#discussion_r847727443

maxentile:
A large fraction of these are index arrays -- may want to standardize to always use original numpy for integer arrays (also potentially in files that didn't appear in the diff accompanying #716 )

maxentile:
I've started categorizing the list of 40 requested changes to clarify intent, with the aim of applying this intent across other files / lines that do not appear in the diff accompanying #716 :

Location:
* tests (33/40)
    * 27/40 in test_reweighting.py
    * 6/40 in test_jax_utils.py
    * 2/40 in test_jax_nonbonded.py
* attic (3/40)
    * 3/40 in attic'd potentials (2 in evp.py, 1 in shape.py)
* timemachine (3/40)
    * 2/40 in fe/reweighting.py
    * 1/40 in optimize/protocol.py
    * 1/40 in potentials/nonbonded.py

-------

Usage type:
* integer index arrays:
    * jnp.triu_indices
    * jnp.*, dtype=int
    * jnp.arange
    * N_k = jnp.array
* reshaping inside of Jax JIT functions: jnp.newaxis
* float array construction: jnp.linspace, jnp.ones, jnp.zeros, jnp.diag, jnp.eye
* int math: jnp.sum( int_array )
* float math: jnp.sum, jnp.mean, jnp.trapz, jnp.exp, jnp.matmul
* array concatenation: jnp.vstack, jnp.array, jnp.stack
* assertions: jnp.isfinite, jnp.isclose
* constants: jnp.inf
* printing: jnp.set_print_options

--------

@proteneer: Could you comment briefly on the general motivation for these changes?

The motivation for integer index arrays is the only one that's currently transparent to me. (Since it can cause errors with slicing.) In other cases, I've defaulted / preferred to use the jax numpy versions and not noticed a drawback yet.

proteneer:
Safety and Consistency:

Safety: By default I'd prefer numpy over jax.numpy, since jax arrays aren't as strict as numpy arrays when it comes to `dtype`, i.e. declaring a `jax.ndarray(..., dtype=np.float64)` may not actually be 64bit unless the correct XLA flags are turned on. This may make tests comparing against jax arrays to have weird behavior.

[...]

PR #716, opened 2022-04-11
Normalize imports: numpy as np, jax.numpy as jnp

maxentile:
Small cleanup: Applies convention suggested by @mcwitt and discussed last week:
* original numpy should always be `np`
* jax.numpy should always be  `jnp`

-----

Notes:
* This DOES update references within code but does NOT update within strings or comments
* This does NOT change the couple places where `jax.numpy.eye` or `jax.numpy.sum` were used
* This should later be added to .git-blame-ignore-revs
* To scan for any unwanted changes, reviewing commit by commit may be easiest
* This does not add a linting stage to enforce this convention, but that may be something to consider!

PR #720, opened 2022-04-12
[WIP] Modify some uses of jax numpy to original numpy

maxentile:
Addressing the changes suggested where they became visible in the diff of https://github.com/proteneer/timemachine/pull/716 , and categorized in https://github.com/proteneer/timemachine/issues/717 

-------

Notes:
* Currently complete for integer index arrays -- other cases started. Will tag for review this afternoon when all cases are handled

### DIFF
--- a/tests/test_reweighting.py
+++ b/tests/test_reweighting.py
@@ -106,7 +106,7 @@ def test_mixture_reweighting_1d():
     n_samples_per_window = int(1e5)
     atol = 1e-2
 
-    N_k = jnp.array([n_samples_per_window] * n_windows)
+    N_k = np.array([n_samples_per_window] * n_windows)
     n_samples_total = sum(N_k)
 
     trajs = [sample(lam, ref_params, n_samples_per_window) for lam in lambdas]
--- a/tests/test_reweighting.py
+++ b/tests/test_reweighting.py
@@ -237,9 +237,9 @@ def test_endpoint_reweighting_ahfe():
 
     v, g = value_and_grad(estimate_delta_f)(ref_params)
 
-    assert jnp.isfinite(v)
+    assert np.isfinite(v)
     assert v == ref_delta_f
-    assert jnp.isfinite(g).all()
+    assert np.isfinite(g).all()
     assert (g != 0).any()
     assert g.shape == ref_params.shape
     # assert anything_about_direction_of_g  # not expected because the "sample" arrays are made up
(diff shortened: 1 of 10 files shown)


maxentile:
Applies most suggested changes from https://github.com/proteneer/timemachine/pull/716 and https://github.com/proteneer/timemachine/issues/717 .

A couple deliberately not applied (e.g. jnp.newaxis -> np.newaxis : 4 instances (3 in attic, 1 in tests), all immediately precede other jax usages. also `jnp.newaxis == np.newaxis`)

tests/test_reweighting.py
```diff
@@ -137,7 +137,7 @@ def test_mixture_reweighting_1d():
     # TI
     vec_du_dl = vmap(grad(u_fxn, 1), (0, None, None))
     mean_du_dls = jnp.array([jnp.mean(vec_du_dl(traj, lam, ref_params)) for (traj, lam) in zip(trajs, lambdas)])
```
proteneer, line 139: did we want these to be numpy as well?

timemachine/optimize/protocol.py
```diff
@@ -115,7 +116,7 @@ def log_weights_from_mixture(u_kn: Array, f_k: Array, N_k: Array) -> Array:
     mixture of states p(x) = (1 / K) \sum_k e^-u_k / Z_k
     """
     log_q_k = f_k - u_kn.T
-    N_k = jnp.array(N_k, dtype=jnp.float64)  # may be ints, or in a list...
+    N_k = np.array(N_k, dtype=np.float64)  # may be ints, or in a list...
```
proteneer, line 119: actually why are these `np.float64` dtypes?

maxentile: not a strong reason, but didn't modify this instance since passing to `jax.scipy.logsumexp(..., b=N_k)` in the next line

there's not an error if these are ints rather than floats (I think in earlier version there was), and it doesn't matter if these are np arrays or jnp arrays, but it may be marginally clearer / more general to emphasize that the mixture weights can be arbitrary floats (that in this case just happen to be proportional to ints)

tests/test_reweighting.py
maxentile: sure -- the inputs are device arrays from jax.vmap (and the usages on the subsequent lines are jax), but this can be converted to original numpy

proteneer (approved):
lgtm!

timemachine/optimize/protocol.py
proteneer: thanks! forgot that weights in `logsumexp` must be floats!

mcwitt (approved):
Looks great! Curious what criteria you used for deciding what should remain a jax array? Is it just the cases where we're currently making use of a jax transformation?

PR #722, opened 2022-04-13
Local Geometry Helper Functions

proteneer:
This PR is part of a much more involved branch dealing with restraints in single topology methods. In particular, this PR introduces a heuristic for identifying the local geometry of a given atom. This will be used later on to setup specialized, factorizable, restraints for each local geometry combination.

The local geometry classification is as follows:

```
class LocalGeometry(Enum):
    G1_TERMINAL = 0  # R-X
    G2_KINK = 1  # R-X-H
    G2_LINEAR = 2  # R-X#N
    G3_PLANAR = 3  # R-X(=O)O
    G3_PYRAMIDAL = 4  # R-X(-H)H
    G4_TETRAHEDRAL = 5  # R-X(-H)(-H)H
```

A function `assign_atom_geometry` is introduced that uses simple hybridization and connectivity rules to determine the local geometry of each atom within a molecule. It does not care for rings, nor does it currently require structural information! The tests cover three categories:

1) drug-like molecules (aspirin, etc.)
2) common neutral heterocycles from Baran I
3) some examples of 2) that have reasonable titratable sites (though I'm looking for more examples of this in general).

### DIFF
--- a/tests/test_geometry.py
+++ b/tests/test_geometry.py
@@ -1,0 +1,534 @@
+# This file tests geometry classification for a wide variety of molecules,
+# selected from common drugs and baran's heterocyclic chemistry
+
+from rdkit import Chem
+
+from timemachine.fe import geometry
+from timemachine.fe.geometry import LocalGeometry as LG
+
+
+def test_assign_aspirin():
+    mol = Chem.AddHs(Chem.MolFromSmiles("CC(=O)OC1=CC=CC=C1C(=O)O"))
+    atom_types = geometry.classify_geometry(mol)
+
+    expected_types = [
+        LG.G4_TETRAHEDRAL,
+        LG.G3_PLANAR,
+        LG.G1_TERMINAL,
+        LG.G2_KINK,
+        LG.G3_PLANAR,
+        LG.G3_PLANAR,
+        LG.G3_PLANAR,
+        LG.G3_PLANAR,
+        LG.G3_PLANAR,
+        LG.G3_PLANAR,
+        LG.G3_PLANAR,
+        LG.G1_TERMINAL,
+        LG.G2_KINK,
+        LG.G1_TERMINAL,
+        LG.G1_TERMINAL,
+        LG.G1_TERMINAL,
+        LG.G1_TERMINAL,
+        LG.G1_TERMINAL,
+        LG.G1_TERMINAL,
+        LG.G1_TERMINAL,
+        LG.G1_TERMINAL,
+    ]
+
+    assert atom_types == expected_types
+
+
+def test_assign_nitrogens():
+    """Test assignment with weird SP and SP2 nitrogens."""
+
+    mol = Chem.AddHs(Chem.MolFromSmiles("CC1C(N)C2=C(C=NC(=C2)C#N)C1=O"))
+    atom_types = geometry.classify_geometry(mol)
+
+    expected_types = [
+        LG.G4_TETRAHEDRAL,
+        LG.G4_TETRAHEDRAL,
+        LG.G4_TETRAHEDRAL,
+        LG.G3_PYRAMIDAL,
+        LG.G3_PLANAR,
+        LG.G3_PLANAR,
+        LG.G3_PLANAR,
+        LG.G2_KINK,
+        LG.G3_PLANAR,
+        LG.G3_PLANAR,
+        LG.G2_LINEAR,
+        LG.G1_TERMINAL,
+        LG.G3_PLANAR,
+        LG.G1_TERMINAL,
+        LG.G1_TERMINAL,
+        LG.G1_TERMINAL,
+        LG.G1_TERMINAL,
+        LG.G1_TERMINAL,
+        LG.G1_TERMINAL,
+        LG.G1_TERMINAL,
+        LG.G1_TERMINAL,
+        LG.G1_TERMINAL,
+        LG.G1_TERMINAL,
+    ]
+
+    assert atom_types == expected_types
+
+
+def test_assign_truncated_sildenafil():
+    # useful for having a mixture of sulfur groups, and common heterocycles
+    mol = Chem.AddHs(Chem.MolFromSmiles("CCOC1=C(C=C(C=C1)S(=O)(=O)N1CCN(C)CC1)C1=NC=CC(=O)N1"))
+    atom_types = geometry.classify_geometry(mol)
+    expected_types = [
+        LG.G4_TETRAHEDRAL,
+        LG.G4_TETRAHEDRAL,
+        LG.G2_KINK,
+        LG.G3_PLANAR,
+        LG.G3_PLANAR,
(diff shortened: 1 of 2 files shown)


tests/test_geometry.py
```diff
@@ -0,0 +1,534 @@
...
+def test_baran_pyridine():
+    mol = Chem.AddHs(Chem.MolFromSmiles("c1ccccn1"))
+    atom_types = geometry.classify_geometry(mol)
+    expected_types = [
+        LG.G3_PLANAR,
+        LG.G3_PLANAR,
+        LG.G3_PLANAR,
+        LG.G3_PLANAR,
+        LG.G3_PLANAR,
+        LG.G2_KINK,
```
jkausrelay, line 306: Why is it Kink and not Planar for the N in pyridine?

jkausrelay, line 306: Why is it Kink and not Planar for the N in pyridine?

proteneer: the N in Pyridine has two neighbors: C=N-C, which means it can only be assigned to "Kink" or "Linear", and this certainly isn't linear.

jkausrelay (approved):
lgtm

timemachine/fe/geometry.py
```diff
@@ -0,0 +1,67 @@
...
+    Heuristic using hybridization information to assign local description
+    of geometry.
+    """
+    nbrs = list(atom.GetNeighbors())
+    if len(nbrs) == 0:
+        assert 0, "Ion not supported"
+    elif len(nbrs) == 1:
+        return LocalGeometry.G1_TERMINAL
+    elif len(nbrs) == 2:
+        if atom.GetHybridization() == Chem.HybridizationType.SP3:
```
maxentile, line 28: Readability preference: move RDKit API syntax (`atom.GetHybridization()`) out of conditionals.

(General preference: introduce aliases where needed, so that lines with key logic are as close as possible to a direct statement of intent:
`if atom.GetHybridization() == Chem.HybridizationType.SP3`
-->
`if hybridization == SP3`
or similar.)

timemachine/fe/geometry.py
```diff
@@ -0,0 +1,67 @@
...
+            return LocalGeometry.G3_PYRAMIDAL
+        elif atom.GetHybridization() == Chem.HybridizationType.SP2:
+            return LocalGeometry.G3_PLANAR
+        else:
+            assert 0, "Unknown 3-nbr geometry"
+    elif len(nbrs) == 4:
+        if atom.GetHybridization() == Chem.HybridizationType.SP3:
+            return LocalGeometry.G4_TETRAHEDRAL
+    else:
+        assert 0, "Too many neighbors"
```
maxentile, line 47: Nit: the `len(nbrs) == 4` branch is missing a final `else: assert 0, "unknown 4-nbr geometry"`, will instead print `"Too many neighbors"`

maxentile (approved):
The direction of applying new restraints per atom is appealing, and classifying atoms into these 6 categories based on `(num_neighbors, hybridization_state)` seems natural.

However, I am currently missing some information required to assess suitability of this function for its intended application -- may be best to defer discussion of that until a later PR?

timemachine/fe/geometry.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def classify_geometry(mol):
+    """
+    Identify the local geometry of the molecule. This current uses a heuristic but we
+    should really be generating this from gas-phase simulations of the real forcefield.
```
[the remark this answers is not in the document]
proteneer: > Are there any implications for the correctness of the intended restraints if this heuristic and the forcefield disagree? (e.g. if hybridization says an atom should be pyramidal, but for some reason the forcefield makes it planar)

This is an excellent point, and one where I'm not sure how to fully resolve yet. The tests were written against molecules whose geometries are "fairly well-understood", and the current typing system is consistent with them. If we change this code to use forcefield parameters, and the tests fail, this indicates that our forcefield has issues and they should be fixed. While the restraints themselves depend on a correct description of the local geometry (and should be forcefield independent), if say right now, we misparameterized an improper torsion for sp2 atoms, then we'd have blow-ups in our simulations.

> If these should be in agreement, could you comment on how to compare these classes with the applied force field parameters? (Not immediately obvious to me, since the forcefield can express geometry preference that don't neatly fall into these classes (what if it's "partially planar" "partially pyramidal" etc.))

We'd need to run vacuum simulations and compare the local geometries computed by looking at the coordinates. The results don't have to fall "neatly", but there are certain singularities to look out for. This can be problematic where an exceeding small fraction of the time you adopt a numerically unstable geometry for the constraint (eg. H-O-H 99% of the time is in the kink formation, but it by pure luck adopted the linear formation, then any cross-product based restraints defined over H-O and O-H would blow up. And this is the most important point, so what matters is not that the angle is 90, 120, 145, etc. but rather that it's *not* 180.

timemachine/fe/geometry.py
```diff
@@ -0,0 +1,67 @@
...
+
+    Currently, 3D coordinates are not required, but this may change in the future.
+
+    Parameters
+    ----------
+    mol: Chem.Mol
+        Input molecule.
+    """
```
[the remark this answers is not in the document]
proteneer: willfix

timemachine/fe/geometry.py
proteneer: will fix

timemachine/fe/geometry.py
proteneer: great catch. will fix

timemachine/fe/geometry.py
proteneer: For the `HybridizationType`, unfortunately I can't just import `from rdkit.Chem.HybridizationType import SP3` since it's not a module but an enum Class

timemachine/fe/geometry.py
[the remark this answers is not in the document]
maxentile: >The tests were written against molecules whose geometries are "fairly well-understood", and the current typing system is consistent with them.

Makes sense!

>If we change this code to use forcefield parameters, and the tests fail, this indicates that our forcefield has issues and they should be fixed. 

May be independently useful to check applied parameters against these classes!

(Perhaps at runtime, or perhaps on a challenging set like [vehicle](https://pubs.acs.org/doi/abs/10.1021/jm801513z).)

>The results don't have to fall "neatly", but there are certain singularities to look out for. This can be problematic where an exceeding small fraction of the time you adopt a numerically unstable geometry for the constraint (eg. H-O-H 99% of the time is in the kink formation, but it by pure luck adopted the linear formation, then any cross-product based restraints defined over H-O and O-H would blow up. And this is the most important point, so what matters is not that the angle is 90, 120, 145, etc. but rather that it's not 180.

Thanks -- this clarifies!

proteneer:
rotm (rebase onto master) - will merge when tests pass!

PR #724, opened 2022-04-15
Change convention of multiple_steps store variables

badisa:
The original behavior was to store elements (energies, du_dls, coords/boxes) with the inputs stored and evaluated. It also had the behavior of not storing frames as you might expect. Previously if you ran something like `context.multiple_steps(np.zeros(100), 200, 200)` it would store du_dls and coordinates/box at step 0 (the original input) and step 10, this results in 99 steps that are performed but unused.

Other minor changes
* Used the outputs of multiple_steps where reasonable
* Removed a useless test since we no longer have observables


timemachine/fe/model_rabfe.py
```diff
@@ -97,11 +97,12 @@ def simulate_futures(
             bp = unbound_pot.bind(np.asarray(params))
             bound_potentials.append(bp)
 
+        # Always get at least one sample
+        subsample_interval = min(self.prod_steps, 1000)
```
badisa, line 101: These changes are gross, but seemed like the simplest way to ensure that you get at least one frame back

badisa: Another alternative is that if the interval is greater than the number of steps, reduce the interval to the number of steps

timemachine/cpp/src/context.cu
```diff
@@ -63,8 +63,11 @@ Context::multiple_steps(const std::vector<double> &lambda_schedule, int store_du
     if (store_x_interval <= 0) {
         throw std::runtime_error("store_x_interval <= 0");
     }
-    int du_dl_buffer_size = (lambda_schedule.size() + store_du_dl_interval - 1) / store_du_dl_interval;
-    int x_buffer_size = (lambda_schedule.size() + store_x_interval - 1) / store_x_interval;
+    if (lambda_schedule.size() % store_x_interval != 0) {
+        std::cout << "warning:: length of lambda_schedule modulo store_x_interval does not equal zero" << std::endl;
```
mcwitt, line 67: I'm having trouble thinking of cases where this shouldn't just be an error. It seems like this would result in extra effort that's unobservable from the calling code?

timemachine/cpp/src/context.cu
```diff
@@ -147,54 +149,44 @@ std::array<std::vector<double>, 3> Context::multiple_steps_U(
...
 
+    if (n_steps % store_x_interval != 0) {
+        std::cout << "warning:: n_steps modulo store_x_interval does not equal zero" << std::endl;
+    }
+
     int n_windows = lambda_windows.size();
-    int u_traj_size = ((n_steps + store_u_interval - 1) / store_u_interval) * n_windows;
-    int x_traj_size = (n_steps + store_x_interval - 1) / store_x_interval;
+    int u_traj_size = (n_steps / store_u_interval) * n_windows;
+    int x_traj_size = (n_steps / store_x_interval);
```
mcwitt, line 158: nit: redundant parens

timemachine/md/moves.py
```diff
@@ -73,16 +74,16 @@ def n_proposed(self):
...
-        lamb,
-        masses,
-        temperature,
-        pressure,
-        n_steps,
-        seed,
-        dt=1.5e-3,
-        friction=1.0,
-        barostat_interval=5,
+        ubps: List,
```
mcwitt, line 77: nit: use more specific `List[CustomOpWrapper]`?

Not the fault of this PR, but I think `ubps` is also a misnomer?

mcwitt: note: I think we'd have to add `assert isinstance(ubps[0], HarmonicBond)` before `bond_list = get_bond_list(ubps[0])` below for this to work.

badisa: `ubps` is definitely unclear, not sure what to call it though and we use `ubps` for these binded (having called `.bind()`)  potentials in a lot of places. 

Adding in `List[CustomOpWrapper]` causes ln 93 to fail as it expects a `HarmonicBond` type. Will see if there is a way around this

mcwitt: > Will see if there is a way around this

I think my comment above should work

> add assert isinstance(ubps[0], HarmonicBond) before bond_list = get_bond_list(ubps[0])

mcwitt: > ubps is definitely unclear, not sure what to call it though and we use ubps for these binded

Ah, that's true; I think my comment about it being a misnomer was a bit confused; this does make sense to me now to contrast with bound potentials

timemachine/cpp/src/context.cu
```diff
@@ -147,54 +149,44 @@ std::array<std::vector<double>, 3> Context::multiple_steps_U(
...
-                    d_box_t_,
-                    lambda,
-                    d_du_dx_t_, // we only need the forces
-                    nullptr,
-                    nullptr,
-                    nullptr,
-                    stream);
-            }
-
             // we need to compute aggregate energies on this step
```
mcwitt, line 187: nit: is this comment still relevant since the first step was removed?

badisa: I think so, the step that it does it changes but it is still aggregating all of the energies. Not the most informative comment though, so can remove if it seems confusing

timemachine/md/moves.py
badisa: Resolved in [a1d2bbf](https://github.com/proteneer/timemachine/pull/724/commits/a1d2bbf1be0d77f4282149c66ff3a743e193fd25)

timemachine/cpp/src/context.cu
```diff
@@ -147,54 +149,44 @@ std::array<std::vector<double>, 3> Context::multiple_steps_U(
...
                     cudaMemcpyDeviceToHost));
                 gpuErrchk(cudaMemcpy(
-                    &d_box_traj[0] + (step / store_x_interval) * 3 * 3,
+                    &d_box_traj[0] + ((step / store_x_interval) - 1) * 3 * 3,
                     d_box_t_,
                     3 * 3 * sizeof(*d_box_traj),
                     cudaMemcpyDeviceToDevice));
             }
 
-            cudaStream_t stream = static_cast<cudaStream_t>(0);
```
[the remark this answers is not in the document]
mcwitt: Just so I understand, was this simplification enabled by not needing to compute the energy before the first step?

timemachine/cpp/src/context.cu
mcwitt: Maybe just reword to avoid "on this step"? It's a bit more confusing now because it seems like "this step" could refer to `step`

timemachine/md/moves.py
mcwitt: Ah, nice, I hadn't used `cast` before. That seems nice for some situations in that it's apparently completely ignored at runtime. In this case though, I'm on the fence -- it seems like the assertion might be nice to have anyway so there's a clear error message if the user passes in a list of potentials where `ubp[0]` is not a `HarmonicBond`?

badisa: Will add in the assertion

timemachine/cpp/src/context.cu
maxentile: >I'm having trouble thinking of cases where this shouldn't just be an error. It seems like this would result in extra effort that's unobservable from the calling code?

Oh -- not always an error, still observable from calling code!

Consider an idiom [like](https://github.com/proteneer/timemachine/blob/ba7114f6147b55d6ff972ffc901dc32eb39dda9d/timemachine/md/moves.py#L110-L111)

```python
x_before = ...
_ = ctxt.multiple_steps(x_before, ...)
x_after = ctxt.get_x_t()
```

(where you discard the result of multiple_steps, and access the updated state of the context). The value of `store_x_interval` will be irrelevant to the behavior of this code.

The change in this PR should mostly obviate that idiom though, and make it more natural to use the return value of `multiple_steps`, rather than accessing the updated state of `ctxt`

mcwitt: Ah, makes sense. Thanks for the clarification!

timemachine/cpp/src/context.cu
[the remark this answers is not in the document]
badisa: Exactly

timemachine/md/moves.py
badisa: Resolved in [8c394ab](https://github.com/proteneer/timemachine/pull/724/commits/8c394ab189e4d3cfdddf1420555d8242bc24d609)

timemachine/cpp/src/context.cu
badisa: Resolved in [e7ab5f5](https://github.com/proteneer/timemachine/pull/724/commits/e7ab5f54c1e718de58d841e800e2228272127329)

timemachine/cpp/src/context.cu
badisa: Resolved in [dfb5198](https://github.com/proteneer/timemachine/pull/724/commits/dfb5198720efe6e9aed1a8c200f32d1926825f5a)

mcwitt (approved):
Looks good! The new behavior seems less surprising, and a nice simplification

proteneer (commented):
we should add/update a docstring somewhere that details exactly the intervals work (i.e. they've been offset by 1).

tests/test_md.py
```diff
@@ -183,15 +325,15 @@ def integrate_once_through(x_t, v_t, box, params):
 
         lambda_schedule = np.ones(num_steps) * lamb
 
-        du_dl_interval = 3
-        x_interval = 2
+        du_dl_interval = 2
```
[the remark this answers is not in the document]
badisa: Nope, can revert these changes

proteneer:
> > we should add/update a docstring somewhere that details exactly the intervals work (i.e. they've been offset by 1).
> 
> Would in the docstring of `multiple_steps` and `multiple_steps_U` work?

Yep!

PR #725, opened 2022-04-15
Trigger verify-typing-stubs hook only on changes to wrapper code

mcwitt:
This avoids unnecessary compilation at commit time, e.g. for Python-only changes.

Verified that the hook runs as expected locally, and continues to work in CI (example failure in 82b78a3)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -57,6 +57,7 @@ repos:
 
         language: system
         pass_filenames: false
+        files: '^timemachine/cpp/src/'
 
 exclude: >
   (?x)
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -57,6 +57,7 @@ repos:
 
         language: system
         pass_filenames: false
+        files: '^timemachine/cpp/src/'
```
badisa, line 60: Actually is there a reason this doesn't only look at the `wrap_kernels.cpp`? Presumably the typing stubs are only for pybind?

PR #726, opened 2022-04-19
Clean up neighborlist code

badisa:
These are changes that I made in my quest to modify the neighborlist for Local MD. To avoid dropping a massive PR at the end, wanted to put these out earlier.


timemachine/cpp/src/neighborlist.cu
```diff
@@ -99,13 +100,21 @@ std::vector<std::vector<int>> Neighborlist<RealType>::get_nblist_host(
...
         throw std::runtime_error("No row coords provided, but NR != 0");
     }
-    double *d_col_coords = gpuErrchkCudaMallocAndCopy(h_column_coords, NC * 3);
 
-    double *d_row_coords = this->compute_full_matrix() ? gpuErrchkCudaMallocAndCopy(h_row_coords, NR * 3) : nullptr;
+    bool store_row = this->compute_full_matrix();
+
+    DeviceBuffer<double> d_col_coords(NC * 3);
+    DeviceBuffer<double> d_box(3 * 3);
+    DeviceBuffer<double> d_row_coords(store_row ? NR * 3 : 0);
```
mcwitt, line 108: nit: this changes the behavior to always allocate. It probably won't make much difference, but I think we can use `std::unique_ptr<DeviceBuffer<double>>` to recover the old behavior (e.g. like we do in [potential.cu](https://github.com/proteneer/timemachine/blob/776873a013e65f8d825570312d0fa87d3a438522/timemachine/cpp/src/potential.cu#L36-L57))

tests/test_nblist.py
```diff
@@ -59,59 +62,91 @@ def get_water_coords(D, sort=False):
...
-    water_coords = get_water_coords(3, sort=False)
-    for size in [35, 64, 129, 1025, 1259, 2029]:
+    block_size = 32
 
-        for nblist in (custom_ops.Neighborlist_f32(size), custom_ops.Neighborlist_f64(size)):
+    num_blocks = (N + block_size - 1) // block_size
+    col_coords = np.expand_dims(coords, axis=0)
 
-            for _ in range(2):
+    box_diag = np.diag(box)
```
mcwitt, line 75: nit: prefer `diagonal` to be explicit

tests/test_nblist.py
```diff
@@ -59,59 +62,91 @@ def get_water_coords(D, sort=False):
...
+    for rbidx in range(num_blocks):
+        row_start = rbidx * block_size
+        row_end = min((rbidx + 1) * block_size, N)
+        row_coords = coords[row_start:row_end]
+        row_coords = np.expand_dims(row_coords, axis=1)
 
-                print("testing size:", size)
+        deltas = row_coords - col_coords
+        deltas -= box_diag * np.floor(deltas / box_diag + 0.5)
```
mcwitt, line 83: nit: could this be a utility function? Maybe something like

```python
def nearest_image(x: NDArray, box_diag: NDArray) -> NDArray:
    return x - box_diag * np.floor(x / box_diag + 0.5)
```

tests/test_nblist.py
```diff
@@ -21,7 +24,7 @@ def test_block_bounds():
 
         coords = np.random.randn(N, D)
         box_diag = np.random.rand(3) + 1
-        box = np.diag(box_diag)
+        box = np.eye(3) * box_diag
```
mcwitt, line 27: Note: I think `np.diagflat(box_diag)` would also work

mcwitt (approved):

tests/test_nblist.py
badisa: Resolved in https://github.com/proteneer/timemachine/pull/726/commits/c42141260bc069e76d470677efef14c91524bb3a

tests/test_nblist.py
badisa: Resolved in https://github.com/proteneer/timemachine/pull/726/commits/c42141260bc069e76d470677efef14c91524bb3a

timemachine/cpp/src/neighborlist.cu
proteneer: I would've thought that a malloc/free of size 0 should be a no-op. However, there may be an issue with the resulting `data` ptr inside a `DeviceBuffer`. Semantically, `execute_device` code expects the device points to be a `nullptr` to indicate that an operation want to be skipped. The current `allocate` function of the `DeviceBuffer` class does:

```
template <typename T> T *allocate(const std::size_t length) {
    T *buffer;
    gpuErrchk(cudaMalloc(&buffer, length * sizeof(T)));
    return buffer;
}
```

So if `length` is zero, I don't think this returns a `nullptr` (rather it just returns a random pointer that points to a zero-sized chunk). Though I'm not sure what the correct behavior is:

1) return a `nullptr` if `length` is zero.
2) return a non-nullptr

What do you guys think?

timemachine/cpp/src/neighborlist.hpp
```diff
@@ -62,7 +62,7 @@ template <typename RealType> class Neighborlist {
     unsigned int *get_ixn_count() { return d_ixn_count_; }
 
     // get max number of row blocks
-    int B() const;
+    int row_blocks() const;
```
proteneer, line 65: can we rename this to `num_row_blocks()`?

timemachine/cpp/src/neighborlist.cu
badisa: I would say it should be `nullptr` as that is our convention in the rest of the code for "don't compute anything for this"

badisa: It is actually already a `nullptr`, looks like cudaMalloc is a noop if the size is 0. Will make it more explicit to be safe

timemachine/cpp/src/neighborlist.hpp
badisa: Should we also rename `column_blocks` to `num_column_blocks`?

timemachine/cpp/src/neighborlist.cu
proteneer: technically if `cudaMalloc` sets `buffer` to a `nullptr` then it's not quite a `no-op` since the pointer will have been changed.

timemachine/cpp/src/neighborlist.hpp
proteneer: yes!

timemachine/cpp/src/neighborlist.cu
badisa: I find it convenient and succinct to be able to say `DeviceBuffer(0)` rather than messing with a unique pointer. And since we own the code, and are explicitly not allocating I don't see it as a code smell.

badisa: Thinking about it more, going to follow the unique ptr style and as add a runtime warning to device buffer if you set the size to <= 0

proteneer (approved):
lgtm!

PR #727, opened 2022-04-20
Fixes Docker build and image size

badisa:
- The version of git that was specified no longer was available, updated to now build
- Removes the openmm source to remove about .4GB from the image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,11 +64,13 @@ RUN git clone https://github.com/openmm/openmm.git --branch "${OPENMM_VERSION}" 
       -DOPENMM_BUILD_OPENCL_LIB=OFF \
       -DOPENMM_BUILD_RPMD_CUDA_LIB=OFF \
       -DOPENMM_BUILD_RPMD_OPENCL_LIB=OFF \
-      -DCMAKE_INSTALL_PREFIX=../openmm_install \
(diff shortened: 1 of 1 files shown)


jkausrelay (approved):
lgtm

Dockerfile
```diff
@@ -64,11 +64,13 @@ RUN git clone https://github.com/openmm/openmm.git --branch "${OPENMM_VERSION}"
...
       -DOPENMM_BUILD_RPMD_OPENCL_LIB=OFF \
-      -DCMAKE_INSTALL_PREFIX=../openmm_install \
+      -DCMAKE_INSTALL_PREFIX=/opt/openmm_install \
       ../ && \
     make -j "$(nproc)" && \
     make -j "$(nproc)" install && \
-    make PythonInstall
+    make PythonInstall && \
+    cd /code/ && \
+    rm -rf openmm/
```
mcwitt, line 73: nit: any reason not to combine these into
```suggestion
    rm -rf /code/openmm
```
?

mcwitt (approved):
LGTM

Dockerfile
badisa: Was mostly to avoid removing the directory that the command is run from, though don't think it would make a difference

mcwitt: I see. Makes sense. No need to change, was mostly just curious.

PR #730, opened 2022-04-28
Parallel version of free solv SMC

jkausrelay:
- Draft until #718 is merged
- Modifies the run_smc_on_freesolv to work in parallel
- Ex on lambda cluster is 
- `python examples/run_smc_on_freesolv.py --n_gpus 10 --n_cpus 4  --result_path $PWD/cache --ff fit_ff.py --filter_mols mobley_8117218 mobley_9460824 ...`
- Offline discussion with Josh - move biphenyl test system to attic


examples/run_smc_on_freesolv.py
```diff
@@ -32,32 +55,90 @@ def parse_options():
...
+    Return the path to the smc results for a particular molecule.
+    """
+    return path / f"summary_smc_result_{mol_id}.pkl"
+
+
+def get_full_traj_path(path: Path, mol_id: str) -> Path:
+    return path / f"full_smc_traj_{mol_id}.pkl"
+
+
+def save_smc_result(path: Path, mol, smc_result: Dict, cmd_args: argparse.Namespace, save_full_trajectories=False):
```
maxentile, line 76: nit: annotation for mol?

examples/run_smc_on_freesolv.py
```diff
@@ -1,17 +1,28 @@
 import argparse
```
maxentile, line 1: Based on offline discussion, makes sense to move the other smc example to the attic (either in this PR or I can do so in a later PR)

examples/run_smc_on_freesolv.py
```diff
@@ -32,32 +55,90 @@ def parse_options():
...
+
+    mol: ROMol
+        Molecule sampled using smc
+
+    smc_result:
+        'initial_samples_refined, 'final_samples_refined' correspond to
+        the refined samples from the end states.
+        'initial_log_weights', 'final_log_weights' correspond to the
+        log weights from both states.
+        'traj' is a list if trajectories, one for each lambda.
```
maxentile, line 93: nit: if -> of typo, I think the description is transposed? (there's a list of samples for each* lambda, but each of those lists is not time ordered. (*all but one lambda))

examples/run_smc_on_freesolv.py
```diff
@@ -69,16 +150,53 @@ def run_on_freesolv_mol(mol):
...
+    for i in range(len(mols)):
+        batch_mols[i % num_gpus].append(mols[i])
+
+    futures = []
+    for mol_subset in batch_mols.values():
+        futures.append(client.submit(run_on_mols, mol_subset, cmd_args))
+
+    # Wait for jobs to complete
+    _ = [fut.result() for fut in futures]
+
```
maxentile, line 199: Some of this logic looks generic enough to consider moving into the parallel module at some point

timemachine/datasets/utils.py
```diff
@@ -1,10 +1,26 @@
...
-from typing import List
+from typing import List, Optional, Set
 
 from rdkit import Chem
 
+from timemachine.fe.utils import get_mol_name
 
-def fetch_freesolv() -> List[Chem.Mol]:
+
+def fetch_freesolv(n_mols: Optional[int] = None, filter_mols: Optional[Set[str]] = None) -> List[Chem.Mol]:
```
maxentile, line 9: Naming preference: `filter_mols` -> `exclude_mols` or similar

("filter" could be misinterpreted as only _keeping_ select mols, rather than only excluding select mols, e.g. based on how python built-in `filter` works)

maxentile (approved):
Looks good!

Left some very minor comments inline

tests/test_datasets.py
```diff
@@ -16,3 +16,14 @@ def test_fetch_freesolv():
     # unique names
     names = [mol.GetProp("_Name") for mol in mols]
     assert len(set(names)) == len(names)
+
+    # trunacated list
+    mols = fetch_freesolv(n_mols=5)
+    assert len(mols) == 5
+
+    # excludeed mol
```
badisa, line 24:
```suggestion
    # excluded mol
```

tests/test_datasets.py
```diff
@@ -16,3 +16,14 @@ def test_fetch_freesolv():
     # unique names
     names = [mol.GetProp("_Name") for mol in mols]
     assert len(set(names)) == len(names)
+
+    # trunacated list
```
badisa, line 20: nit
```suggestion
    # truncated list
```

tests/test_examples.py
```diff
@@ -108,38 +109,6 @@ def test_relative_binding():
             assert isinstance(result.dG_bind_err, float)
 
 
-def test_smc_biphenyl():
```
badisa, line 111: Should we replace this test with a test for the `run_smc_on_freesolv.py`?

jkausrelay: This is tested below (test_smc_freesolv)

badisa (approved):

PR #731, opened 2022-04-29
Modifies Neighborlist to take row indices rather than partitioning 

badisa:
- Makes the neighborlist more flexible by setting row indices, at the cost of `N * sizeof(uint32)` memory. This will be important for local md to avoid having to construct multiple neighborlists
- Replaces a non-async call with an async call
- Unifies a bit of the code in the nonbonded potentials
- No performance difference (within 2ns/day in most cases using benchmark script) on T4 (cuda arch 7.5) and A4000 (cuda arch 8.6) cards


badisa:
Closing until @mcwitt is back

PR #733, opened 2022-05-02
Update git in Dockerfile

jkausrelay:
- E: Version '1:2.25.1-1ubuntu3.3' for 'git' was not found
- Update to newer version

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM nvidia/cuda:11.6.0-devel-ubuntu20.04 AS tm_base_env
 # Copied out of anaconda's dockerfile
 ARG MINICONDA_VERSION=4.6.14
 ARG MAKE_VERSION=4.2.1-1.2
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM nvidia/cuda:11.6.0-devel-ubuntu20.04 AS tm_base_env
 # Copied out of anaconda's dockerfile
 ARG MINICONDA_VERSION=4.6.14
 ARG MAKE_VERSION=4.2.1-1.2
-ARG GIT_VERSION=1:2.25.1-1ubuntu3.3
+ARG GIT_VERSION=1:2.25.1-1ubuntu3.4
```
badisa, line 6: What about doing something like `1:2.25.1-*` that way we get the same version of git, but don't have to worry about the specific apt versions disappearing (which they seem to do)?

badisa (approved):

PR #735, opened 2022-05-04
Bugfix/fix cpu only tests

badisa:


timemachine/parallel/utils.py
```diff
@@ -8,7 +8,10 @@
 
 def get_gpu_count() -> int:
     # Expected to return a line delimited summary of each GPU
-    output = check_output(["nvidia-smi", "-L"])
+    try:
+        output = check_output(["nvidia-smi", "-L"])
+    except FileNotFoundError:
```
badisa, line 13: Perhaps a log statement after this? Didn't want it to be noisy, but can see an argument for one.

.gitlab-ci.yml
```diff
@@ -52,6 +52,7 @@ nogpu-tests:
   needs: ["lint"]
   tags:
     - timemachine
+    - cpu
```
badisa, line 55: Truly run on a machine without a GPU, and without any of cuda drivers/nvidia-smi/etc

proteneer (approved):
lgtm!

timemachine/parallel/utils.py
badisa: Driver errors will still raise an exception. This only returns 0 if there is no binary for `nvidia-smi`

PR #737, opened 2022-05-07
Detect non-trainable parameters in AM1CCCHandler

maxentile:
Problem:

I observed that it is currently possible for AM1CCCHandler to assign distinct parameters to indistinguishable atoms.

Based on discussion with @proteneer , we found that this behavior can be triggered when a symmetric bond pattern in AM1CCCHandler such as 
https://github.com/proteneer/timemachine/blob/fd14908113315ca07c8983e7ecd4dd92178d03a8/timemachine/ff/params/smirnoff_1_1_0_ccc.py#L363
is assigned a non-zero parameter. When a BCC like `('[#6X4:1]-[#6X4:2]', +1.0)` can match a pair of atoms in either direction, an arbitrary direction is selected, leading to undefined behavior in general. When applied to a symmetric molecule like cyclohexane, this can even result in different charges being applied to indistinguishable atoms.

In the initial, untrained force field file, BCCs defined by symmetric bond SMARTS all have parameter values of 0.0, so the undefined match direction has no effect on the assigned charges. However, gradients of realistic training objectives w.r.t. these parameters can be non-zero, so numerical optimization can readily exploit this undefined behavior, which is undesirable.

------------
Possible workarounds:

* Manually identify symmetric patterns, and set their parameter values to nan in the forcefield .py file, so that gradients w.r.t. these parameters would be nan-poisoned. (However, this would also have the effect of nan-poisoning the assigned charges themselves...)
* Detect symmetric SMARTS patterns in the forcefield by string manipulation
* Detect symmetric SMARTS patterns in the force field by inspection of the resulting query mols
* Detect symmetric patterns at runtime by seeing if they match in both directions on a given molecule
* Detect where initial parameter == 0.0. (However, there are some safely asymmetric patterns, such as https://github.com/proteneer/timemachine/blob/fd14908113315ca07c8983e7ecd4dd92178d03a8/timemachine/ff/params/smirnoff_1_1_0_ccc.py#L430 , where the initial parameter is 0.0.)
* Modify definition of AM1CCCHandler so that it's safe for arbitrary SMARTS patterns to have arbitrary parameter values (e.g. by matching both directions rather than an arbitrary direction)

--------

Current PR:

This draft PR adds two heuristic ways to detect parameter indices that currently should not be trained in AM1CCCHandler:

* `check_bond_smarts_symmetric` -- string manipulation heuristic -- regex match `"[<atom1>:1]*[<atom2>:2]"` and return whether atom1 and atom2 are identical strings... If applied to https://github.com/proteneer/timemachine/blob/fd14908113315ca07c8983e7ecd4dd92178d03a8/timemachine/ff/params/smirnoff_1_1_0_ccc.py#L343 , will flag 25 patterns as symmetric.

* `get_spurious_param_idxs` -- intended to be similar to the use case of gradient-based force field optimization (defines an objective function that can only be increased if spuriously distinct parameters are assigned to indistinguishable atoms, takes gradient w.r.t. params, reports indices of any non-zero components of this gradient). This was how I initially detected the problem, and in principle it could be useful for validating any model that assigns parameters to atoms.

------

TBD:

Not sure exactly where best to use this information yet.
* Add a "trainable" vs. "non-trainable" mask to the `AM1CCCHandler` object? (This seems most natural to me, since training script already makes a selection of which parameters to train -- want to use this information to narrow the selection of trainable parameters.)
* Throw an exception due to undefined behavior in `AM1CCCHandler.static_parameterize` when parameters associated with symmetric bond SMARTS are != 0?

### DIFF
--- a/tests/test_handlers.py
+++ b/tests/test_handlers.py
@@ -946,14 +946,20 @@ def test_spurious_param_idxs():
     """Assert that get_spurious_param_idxs
     * returns arrays with length sometimes > 0
     * only ever returns indices of symmetric bond types"""
-    mols = fetch_freesolv()[50:55]  # expect three of these to trigger
+
+    mols = fetch_freesolv()
+    mol_dict = {mol.GetProp("_Name"): mol for mol in mols}
+
+    # expect three of these to trigger
+    names = ["mobley_1743409", "mobley_1755375", "mobley_1760914", "mobley_1770205", "mobley_1781152"]
+    select_mols = [mol_dict[name] for name in names]
 
     ff = Forcefield.load_from_file(DEFAULT_FF)
     q_handle = ff.q_handle
     sym_idxs = set(np.where(q_handle.symmetric_pattern_mask)[0])
 
     n_params_detected_per_mol = []
-    for mol in mols:
+    for mol in select_mols:
         spurious_idxs = get_spurious_param_idxs(mol, q_handle)
         assert set(spurious_idxs).issubset(sym_idxs)
 
--- a/tests/test_handlers.py
+++ b/tests/test_handlers.py
@@ -10,7 +10,8 @@ import pytest
 from rdkit import Chem
 from rdkit.Chem import AllChem, rdmolops
 
-from timemachine.constants import ONE_4PI_EPS0
+from timemachine.constants import DEFAULT_FF, ONE_4PI_EPS0
+from timemachine.datasets import fetch_freesolv
 from timemachine.ff import Forcefield
 from timemachine.ff.charges import AM1CCC_CHARGES
 from timemachine.ff.handlers import bonded, nonbonded
@@ -567,6 +568,37 @@ def test_am1ccc_throws_error_on_phosphorus():
     with pytest.raises(RuntimeError) as e:
         _ = ff.q_handle.parameterize(mol)
     assert "unsupported element" in str(e)
+
+
+def test_am1ccc_symmetric_patterns():
+    """Assert that AM1CCCHandler instances:
+    * have a symmetric_pattern_mask property with the expected shape and contents
+    * can be used with jax.grad
+    * raise ValueError when parameters associated with symmetric patterns != 0"""
+    ff = Forcefield.load_from_file(DEFAULT_FF)
+    q_handle = ff.q_handle
+    smirks, params = q_handle.smirks, q_handle.params
+    mol = fetch_freesolv()[123]
+
+    sym_pattern_mask = q_handle.symmetric_pattern_mask
+    assert len(sym_pattern_mask) == len(q_handle.smirks)
+    assert sum(sym_pattern_mask) == 25
+
+    def f(params):
+        return np.sum(q_handle.static_parameterize(params, smirks, mol) ** 2)
+
+    _ = f(params)
+    _ = jax.grad(f)(params)
+
+    # expect problems when parameters associated with symmetric patterns are 1.0
+    bad_params = np.array(params)
+    bad_params[sym_pattern_mask] = 1.0
+
+    with pytest.raises(ValueError):
+        _ = f(bad_params)
+
+    with pytest.raises(ValueError):
+        _ = jax.grad(f)(bad_params)
 
 
 def test_am1_differences():
--- a/tests/test_handlers.py
+++ b/tests/test_handlers.py
@@ -16,6 +16,7 @@ from timemachine.ff import Forcefield
 from timemachine.ff.charges import AM1CCC_CHARGES
 from timemachine.ff.handlers import bonded, nonbonded
 from timemachine.ff.handlers.deserialize import deserialize_handlers
+from timemachine.ff.handlers.utils import get_spurious_param_idxs, get_symmetry_classes
 
 
 def test_harmonic_bond():
@@ -931,3 +932,32 @@ def test_lennard_jones_handler():
     # if a parameter is > 99 then its adjoint should be zero (converse isn't necessarily true since)
     mask = np.argwhere(params > 90)
     assert np.all(adjoints[mask] == 0.0)
+
+
+def test_symmetry_classes():
+    """Assert get_symmetry_classes returns arrays of expected length num_atoms"""
+    mols = fetch_freesolv()[:10]
+    for mol in mols:
+        sym_classes = get_symmetry_classes(mol)
+        assert len(sym_classes) == mol.GetNumAtoms()
+
+
+def test_spurious_param_idxs():
+    """Assert that get_spurious_param_idxs
+    * returns arrays with length sometimes > 0
+    * only ever returns indices of symmetric bond types"""
+    mols = fetch_freesolv()[50:55]  # expect three of these to trigger
+
+    ff = Forcefield.load_from_file(DEFAULT_FF)
+    q_handle = ff.q_handle
+    sym_idxs = set(np.where(q_handle.symmetric_pattern_mask)[0])
+
+    n_params_detected_per_mol = []
+    for mol in mols:
+        spurious_idxs = get_spurious_param_idxs(mol, q_handle)
+        assert set(spurious_idxs).issubset(sym_idxs)
+
+        n_params_detected_per_mol.append(len(spurious_idxs))
+
+    n_mols_with_any_detected_params = (np.array(n_params_detected_per_mol) > 0).sum()
+    assert n_mols_with_any_detected_params == 3
--- a/timemachine/ff/handlers/nonbonded.py
+++ b/timemachine/ff/handlers/nonbonded.py
@@ -506,7 +506,7 @@ class AM1CCCHandler(SerializableMixIn):
         return self.partial_parameterize(self.params, mol)
 
     @staticmethod
-    def static_parameterize(params, smirks, mol):
+    def static_parameterize(params, smirks, mol, validate=True):
         """
         Parameters
         ----------
@@ -516,9 +516,12 @@ class AM1CCCHandler(SerializableMixIn):
             SMIRKS patterns matching bonds, to be parsed using OpenEye Toolkits
         mol: Chem.ROMol
             molecule to be parameterized.
-
+        validate: bool
+            check params, smirks
         """
-        AM1CCCHandler.static_validate(params, smirks)
+        if validate:
+            AM1CCCHandler.static_validate(params, smirks)
+
         am1_charges = compute_or_load_am1_charges(mol)
         bond_idxs, type_idxs = compute_or_load_bond_smirks_matches(mol, smirks)
 
--- a/timemachine/ff/handlers/nonbonded.py
+++ b/timemachine/ff/handlers/nonbonded.py
@@ -2,19 +2,16 @@ import base64
 import pickle
 from collections import Counter
 
-import jax
 import jax.numpy as jnp
 import networkx as nx
 import numpy as np
-from jax import grad
-from numpy.typing import NDArray
 from rdkit import Chem
 
 from timemachine import constants
 from timemachine.ff.handlers.bcc_aromaticity import AromaticityModel
 from timemachine.ff.handlers.bcc_aromaticity import match_smirks as oe_match_smirks
 from timemachine.ff.handlers.serialize import SerializableMixIn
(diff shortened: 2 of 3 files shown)


tests/test_handlers.py
```diff
@@ -899,3 +932,32 @@ def test_lennard_jones_handler():
...
+    for mol in mols:
+        sym_classes = get_symmetry_classes(mol)
+        assert len(sym_classes) == mol.GetNumAtoms()
+
+
+def test_spurious_param_idxs():
+    """Assert that get_spurious_param_idxs
+    * returns arrays with length sometimes > 0
+    * only ever returns indices of symmetric bond types"""
+    mols = fetch_freesolv()[50:55]  # expect three of these to trigger
```
jkausrelay, line 949: Specific idx into the results of fetch here seems kind of fragile, should these be retrieved by mol name or split off into a file specific for this test?

maxentile: Good call! Retrieved by name in https://github.com/proteneer/timemachine/pull/737/commits/0ab67964a592d8df1cec6eaf3be994b6d98f2b73

timemachine/ff/handlers/nonbonded.py
```diff
@@ -534,8 +516,12 @@ def static_parameterize(params, smirks, mol):
             SMIRKS patterns matching bonds, to be parsed using OpenEye Toolkits
         mol: Chem.ROMol
             molecule to be parameterized.
-
+        validate: bool
+            check params, smirks
         """
+        if validate:
```
jkausrelay, line 522: Is there a reason this is optional (i.e. isn't the ff invalid if the check fails)?

jkausrelay: nvm, I see the use case below

timemachine/ff/handlers/utils.py
```diff
@@ -67,3 +91,99 @@ def match_smirks(mol, smirks):
...
+        expected_constant_within_class = class_means[symmetry_classes]
+        assert expected_constant_within_class.shape == sys_params.shape
+        deviation_from_class_means = sys_params - expected_constant_within_class
+        spuriosity = jnp.sum(deviation_from_class_means ** 2)
+
+        return spuriosity
+
+    # TODO: may also want to try several points in the parameter space,
+    #   randomly or systematically flipping signs...
+    trial_params = np.ones(len(handle.params))  # TODO: generalize
```
jkausrelay, line 181: handle.params + 1.0 ? So it will always be a perturbation to the original params.

timemachine/ff/handlers/nonbonded.py
```diff
@@ -534,8 +516,12 @@ def static_parameterize(params, smirks, mol):
...
+            AM1CCCHandler.static_validate(params, smirks)
+
         am1_charges = compute_or_load_am1_charges(mol)
         bond_idxs, type_idxs = compute_or_load_bond_smirks_matches(mol, smirks)
```
jkausrelay, line 526: Could `compute_or_load_bond_smirks_matches` return the list of symmetric bonds and then do the check for delta[symmetric_bond_idxs] != 0 here?

Basically check the result of the match instead of the string.

timemachine/ff/handlers/utils.py
```diff
@@ -67,3 +91,99 @@ def match_smirks(mol, smirks):
...
+        will not be matched, will return False by default.
+        However, for the bond smarts subset used by the AM1CCC model, this covers most cases
+    """
+
+    pattern = re.compile(r"\[(?P<atom1>.*)\:1\].\[(?P<atom2>.*)\:2\]")
+    match = pattern.match(bond_smarts)
+
+    if type(match) is re.Match:
+        complete = match.span() == (0, len(bond_smarts))
+        symmetric = match.group("atom1") == match.group("atom2")
```
badisa, line 122: Would it be advantageous to do the following?

```suggestion
        symmetric = list(sorted(match.group("atom1").split(","))) == list(sorted(match.group("atom2").split(",")))
```

Seems to work in the case of `[#6,#7:1]~[#7,#6:2]`

That said, nothing in our AM1BCC charges seems to have this lack of ordering, so probably unimportant

timemachine/ff/handlers/utils.py
```diff
@@ -67,3 +91,99 @@ def match_smirks(mol, smirks):
...
+    -----
+    * The AM1CCC model contains symmetric patterns that must be assigned 0 parameters
+        (Otherwise, undefined behavior when symmetric bond matches in an arbitrary direction)
+    * Only checks string equivalence!
+        for example
+        check_bond_smarts_symmetric("[#6,#7:1]~[#7,#6:2]")
+        will be a false negative
+    * Does not handle all possible bond smarts
+        for example
+        "[#6:1]~[#6:2]~[#1]"
```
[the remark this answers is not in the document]
maxentile: Already a problem: there are two patterns in the current model that are not parsed correctly

`[#6X3:1](~[#8X1,#16X1])(~[#16X1:2])`
and
`[#6X3:1](~[#8X1,#16X1])(~[#8X1:2])`

Since these don't match the regex, they return `False` by default.

Don't have a sense of what SMARTS subset may be exercised by future edits to this force field.

timemachine/ff/handlers/utils.py
```diff
@@ -67,3 +91,99 @@ def match_smirks(mol, smirks):
...
+    # imported here for optional dependency
+    from openeye import oechem
+
+    oemol = convert_to_oe(rdmol)
+    oechem.OEPerceiveSymmetry(oemol)
+    symmetry_classes = np.array([atom.GetSymmetryClass() for atom in oemol.GetAtoms()])
+    n_classes = len(set(symmetry_classes))
+
+    # make indexy / contiguous from 0 to n_classes
+    idx_map = {old_idx: new_idx for (new_idx, old_idx) in enumerate(set(symmetry_classes))}
```
badisa, line 143: nit: don't need the parens around (new_idx, old_idx).

tests/test_handlers.py
```diff
@@ -899,3 +932,38 @@ def test_lennard_jones_handler():
...
+
+def test_spurious_param_idxs():
+    """Assert that get_spurious_param_idxs
+    * returns arrays with length sometimes > 0
+    * only ever returns indices of symmetric bond types"""
+
+    mols = fetch_freesolv()
+    mol_dict = {mol.GetProp("_Name"): mol for mol in mols}
+
+    # expect three of these to trigger
```
badisa, line 953: nit: Might be nice to separate out the 3 that trigger, making it easier to reason about the test at a later date

timemachine/ff/handlers/utils.py
```diff
@@ -67,3 +91,99 @@ def match_smirks(mol, smirks):
...
+    symmetry_classes = np.array([idx_map[old_idx] for old_idx in symmetry_classes])
+    assert set(symmetry_classes) == set(range(n_classes))
+
+    return symmetry_classes
+
+
+def get_spurious_param_idxs(mol, handle) -> NDArray:
```
badisa, line 150: nit: Is `spurious` the right term? This seems more specific to indistinguishable param idxs?

timemachine/ff/handlers/utils.py
maxentile: Can double-check this -- one issue with `.split(",")` is that comma-separated lists can appear inside of nested structures. For example, applying `match.group("atom1").split(",")`  to this pattern https://github.com/proteneer/timemachine/blob/fd14908113315ca07c8983e7ecd4dd92178d03a8/timemachine/ff/params/smirnoff_1_1_0_ccc.py#L346

returns `['#6X3$(*=[#7', '#15])']` which looks undesirable -- would probably want to sort recursively

However, there are other ways you could produce syntactically distinct SMARTS with identical meaning. Not sure if there's a robust "canonicalize SMARTS" function...

timemachine/ff/handlers/utils.py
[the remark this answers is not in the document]
badisa: What about using the following pattern?

```
pattern = re.compile(r"(\[(?P<atom1>.+)\:1\].+\[(?P<atom2>.+)\:2\]`)
...
complete = match.span() == (0, bond_smarts.index(":2]") + 3)
```

Would cover those two cases. Unless you also need to match the `16X1` to the `16X1` in the middle? Which I guess you could do the hacky thing of having a third group in the middle and check if the atom1 and atom2 are represented somewhere in the middle....

timemachine/ff/handlers/utils.py
badisa: Increasingly disliking the regex approach

timemachine/ff/handlers/nonbonded.py
badisa, line 522: After talking offline about how this is intended to verify the FF and not the molecule's parameterization. What about moving this validation to the Forcefield class? It seems like longer term we will want to do more validation of the forcefield parameters as we train them.

timemachine/ff/handlers/utils.py
maxentile: Readability preference for including them, unless they get in the way of something

tests/test_handlers.py
```diff
@@ -899,3 +932,38 @@ def test_lennard_jones_handler():
     # if a parameter is > 99 then its adjoint should be zero (converse isn't necessarily true since)
     mask = np.argwhere(params > 90)
     assert np.all(adjoints[mask] == 0.0)
+
+
+def test_symmetry_classes():
+    """Assert get_symmetry_classes returns arrays of expected length num_atoms"""
+    mols = fetch_freesolv()[:10]
```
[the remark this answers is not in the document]
maxentile: I don't think so -- the numbering won't be comparable across compounds

tests/test_handlers.py
maxentile: Good call -- will fix!

timemachine/ff/handlers/nonbonded.py
maxentile: Hmm, it might be more appropriate for this to be in the Forcefield constructor at this branch: https://github.com/proteneer/timemachine/blob/776873a013e65f8d825570312d0fa87d3a438522/timemachine/ff/__init__.py#L71-L73

Based on discussion with @jkausrelay yesterday, the initial thought was to put this check in the core function that gets called by everything else (`static_parameterize`, which is called by `partial_parameterize`, which is called by `parameterize`, ...), so it would cover the most use cases (such as training, where we don't necessarily create an instance of the Forcefield class for every optimizer step).

I'm agnostic about where the check should go, as long as it surfaces an error/warning before applying invalid parameters to a molecule.

timemachine/ff/handlers/utils.py
maxentile: "Indistinguishable" sounds like an improvement. 

(Although still not perfect, since I don't want to detect "redundant / indistinguishable parameters", but instead I want to detect "parameters whose variation can introduce spurious differences between atoms that should be indistinguishable")

maxentile:
Should clarify: the reason the `get_spurious_idxs` function is more complicated than just "check if any charges in a symmetry class are different" is that AM1CCC already assigns _slightly_ different partial charges to atoms that should be indistinguishable (because the AM1 charges are currently not symmetrized). 

For example, the carbons in benzene are assigned charges that differ by <0.1%, but are not exactly the same.

```python
import numpy as np
from rdkit import Chem
from timemachine.constants import DEFAULT_FF

# molecule with 2 atomic symmetry classes
benzene = 'C1=CC=CC=C1'
mol = Chem.MolFromSmiles(benzene)
mol = Chem.AddHs(mol)

# expect 2 distinct charges, {+q, -q}
ff = Forcefield.load_from_file(DEFAULT_FF)
assigned_charges = np.array(ff.q_handle.parameterize(mol))

# observe 3 distinct charges
print(set(assigned_charges))  # {1.5335002, -1.5333823, -1.5340897}
```

That implementation is designed to ignore that asymmetry, and instead check whether you can make asymmetry _worse_ by locally varying the adjustable parameters.

maxentile:
Thanks for the discussion of this approach.

Closing since superseded by https://github.com/proteneer/timemachine/pull/739 and https://github.com/proteneer/timemachine/pull/738

PR #738, opened 2022-05-10
Make charge symmetrization stereo-aware

maxentile:
* Disables passing `symmetrize=True` to the AM1 routine (since this appears to average over stereo-oblivious atom classes, and occasionally introduces very small asymmetries in cases like benzene)
* Symmetrizes by averaging over RDKit stereo-sensitive atom classes
* Tests on benzene (where 2 distinct charges are expected, but 3 were produced)
* Tests on a case where symmetry is broken by stereo configuration (where `n_atoms` distinct charges are expected, but ~`n_atoms/2` were produced)

------

Initial PR description:

`symmetrize=True` flag was already being passed to toolkit functions but there were still small asymmetries in some cases (see e.g. https://github.com/proteneer/timemachine/pull/737#issuecomment-1122880117 )

Issue magnitude: very very small. (In 82 of 642 FreeSolv molecules, the partial charges within a symmetry class varied up to ~0.0002 e. However, it's still nice for `len(set(assigned_charges))` for benzene to be 2 rather than 3.)


timemachine/ff/handlers/utils.py
```diff
@@ -67,3 +73,57 @@ def match_smirks(mol, smirks):
...
+    # make contiguous from 0 to n_classes
+    idx_map = {old_idx: new_idx for (new_idx, old_idx) in enumerate(set(symmetry_classes))}
+    symmetry_classes = np.array([idx_map[old_idx] for old_idx in symmetry_classes])
+    assert set(symmetry_classes) == set(range(n_classes))
+
+    return symmetry_classes
+
+
+def symmetrize(per_particle_params, mol):
+    """Replace param[atom] with mean([param[a] for a in symmetry_class(atom)])"""
```
jkausrelay, line 118: nit: update to "Replace per_particle_params[atom] ..."

jkausrelay (approved):

timemachine/ff/handlers/utils.py
```diff
@@ -67,3 +73,57 @@ def match_smirks(mol, smirks):
...
+    """Convenient wrapper for [atom.GetSymmetryClass() for atom in oemol]
+
+    (Accepts rdmol instead of oemol, and renumbers classes to be contiguous from 0 to n_classes.)
+    """
+
+    # imported here for optional dependency
+    from openeye import oechem
+
+    oemol = convert_to_oe(rdmol)
+    oechem.OEPerceiveSymmetry(oemol)
```
proteneer, line 105: 1) is this stereo-aware?
2) bool OEPerceiveSymmetry(OEMolBase &, bool includeH=true, bool automorph=false), what does automorph do here? API docs unfortunately are quite sparse. @bp-kelley @badisa might know...

proteneer:
To clarify on symmetry perception under stereochemistry, consider this molecule:

F\C=C\C=C\C=C/F
Screen Shot 2022-05-11 at 4 46 47 PM

proteneer:
The two fluorines should **not** be in the same symmetry class (for the stereoisomer depicted in the diagram). Considering all conformations that satisfy the depicted stereo, running AM1 calculations should've resulted in *different* charges and should not be symmetrized. Although the diff is very small [-2.082 vs -2.081] shown below, they should be different! 

[-2.082, 0.794, -2.22, -1.398, -1.398, -2.22, 0.794, -2.081, 1.686, 1.69, 1.53, 1.53, 1.69, 1.686]

I suspect it may be because AM1 is not sensitive enough on the raw charges for the example I provided, may try this one, where the environmental differences around the Fs are more drastic:

O/C(=C(/O)\C(\Cl)=C(\F)Cl)/C(/Cl)=C(\F)Cl
Screen Shot 2022-05-13 at 11 27 42 AM

maxentile:
Ahh, thanks for the more dramatic example!

Enumerating all configurations around the double-bonds in the molecule `O/C(=C(/O)\C(\Cl)=C(\F)Cl)/C(/Cl)=C(\F)Cl`, again the picture is similar:
image

* Proposed symmetry classes are stereo invariant (same assignment of 8 classes to 16 atoms for all stereoisomers)
* Currently assigned charges are stereo sensitive (differences up to ~0.03 e across stereoisomers)
* Currently assigned charges within each symmetry class within each stereoisomer are exactly or nearly constant (exactly constant for 7 classes, exception: difference of ~0.00003 e for the class containing oxygen)

If we want the charges to be stereo asymmetric, I wonder if we should modify these lines https://github.com/proteneer/timemachine/blob/451803e01afe6231147a0e6a3ca019d4aa5069d8/timemachine/ff/handlers/nonbonded.py#L89-L91

maxentile:
Toggling `symmetrize=True` to `symmetrize=False` in AM1ELF10 https://github.com/proteneer/timemachine/blob/451803e01afe6231147a0e6a3ca019d4aa5069d8/timemachine/ff/handlers/nonbonded.py#L89-L91 causes large differences of up to 0.1 e within each graph symmetry class within each stereoisomer of `O/C(=C(/O)\C(\Cl)=C(\F)Cl)/C(/Cl)=C(\F)Cl` (as should be expected).

maxentile:
Toggling `symmetrize=True` to `symmetrize=False` in AM1ELF10 https://github.com/proteneer/timemachine/blob/451803e01afe6231147a0e6a3ca019d4aa5069d8/timemachine/ff/handlers/nonbonded.py#L89-L91 causes large differences of up to 0.1 e within each graph symmetry class within each stereoisomer of `O/C(=C(/O)\C(\Cl)=C(\F)Cl)/C(/Cl)=C(\F)Cl` (as should be expected).

maxentile:
Based on offline discussion with @proteneer -- looked into using the RDKit alternative https://www.rdkit.org/docs/source/rdkit.Chem.rdmolfiles.html#rdkit.Chem.rdmolfiles.CanonicalRankAtoms . This appears to have the desired behavior!

The PR now:
* Disables passing `symmetrize=True` to the AM1 routine (since this appears to average over stereo-oblivious atom classes)
* Symmetrizes the result by averaging over RDKit stereo-sensitive atom classes

The number of distinct partial charges is **reduced** on benzene (from 3 on master to 2 on [a220996](https://github.com/proteneer/timemachine/pull/738/commits/a2209969b180dbd4674697c422ace01585eeea52)), and **increased** on `O/C(=C(/O)\C(\Cl)=C(\F)Cl)/C(/Cl)=C(\F)Cl`  (from 9 on master to 16 on [a220996](https://github.com/proteneer/timemachine/pull/738/commits/a2209969b180dbd4674697c422ace01585eeea52))

timemachine/ff/handlers/utils.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/738/commits/511bc2b7b3c0b6cef6cc8d415a118614cfd2b58a

tests/test_handlers.py
```diff
@@ -691,14 +691,14 @@ def mol_with_precomputed_charges():
...
-        2.86426252, 2.86426252])
+        -3.38603383, 0.26304593, 2.30806183, 1.11718269, -2.11466618,
+        0.97255779, -9.18759228, 0.04222331, -2.49355872, 1.87712273,
+        -1.64885893, 0.57591133, -1.68888899, 1.19419578, -1.04842043,
+        -1.52418996, 0.3067501, -1.2601894, -1.32341895, 33.79966262,
+        -10.86363258, -10.86363258, -1.72379533, -0.10198921, -1.47957496,
+        -1.72379533, -11.32267114, 2.74244351, 1.64587022, 1.54997692,
+        1.69815056, 1.83481513, 2.11438216, 1.99952399, 1.98443746,
+        2.86429782, 2.86429782])
     precomputed_charges['ubuntu-20.04.2'] = np.array([
```
maxentile, line 702: Haven't updated the ref values for `ubuntu-20.04.2` yet. If someone has these easily accessible, let me know, otherwise I'm tempted to remove this test altogether.

jkausrelay: These are the values I get:
[-3.3878352035023025, 0.26727080686425936, 2.3026757907523128, 1.1166965996609888, -2.107513358517747, 0.9612083769496058, -9.154301226767831, 0.03288091151318266, -2.4851042282886255, 1.8723247315824088, -1.6547659251138624, 0.5817522843277193, -1.6897608745990291, 1.191577230801982, -1.0478643525397062, -1.5200599330326403, 0.3044723600341638, -1.2594928059462964, -1.326439016721788, 33.755887188134906, -10.851852320219985, -10.851852320219985, -1.724017252391697, -0.10285399008405359, -1.4786711130785013, -1.724017252391697, -11.319224351341875, 2.7444770823135793, 1.6425628940842003, 1.5516172308731973, 1.6987242017361495, 1.8342000296623708, 2.114672322516452, 1.9941244248440126, 1.9856416523206082, 2.866430198256134, 2.866430198256134]

maxentile: Thanks!

proteneer (approved):
nice! would be good sanity check that this doesn't drastically affect FreeSolv results but LGTM!

maxentile:
>would be good sanity check that this doesn't drastically affect FreeSolv results

Checking effect on computed charges for FreeSolv molecules (ref: master (451803e) -> proposed: most recent PR commit (9ce6cb9))

Histogram of signed charge changes per atom (note log y scale)

image

For ~99.2% of atoms, the effect on charge is < 0.001 e, and ~99.7% of atoms, the effect on charge is < 0.01 e.

Histogram of unsigned charge changes per molecule (max over atoms within each molecule, note log x scale)
image

Displaying the molecules where the difference exceeds arbitrary threshold of 0.01 e 
image

Interesting that these all have one or more nitrate groups...

Haven't yet checked the effect of the proposed change on computed hydration free energies, and I'd be curious also to compare the assigned symmetry classes, esp. for the plotted molecules.

proteneer:
Can this be merged?

proteneer:
Can you try this (thanks @bp-kelley for the suggestion on ResonanceMolSupplier!):

``` python
from rdkit import Chem
import numpy as np

def resonance_aware_ranking(mol):

    flags = Chem.ResonanceFlags() | Chem.ResonanceFlags.ALLOW_CHARGE_SEPARATION
    permutations = []

    for m in Chem.ResonanceMolSupplier(mol, flags):
        res = Chem.CanonicalRankAtoms(m, breakTies=False, includeChirality=True)
        permutations.append(list(res))

    permutations = np.array(permutations).T

    # identify which group each orbit belongs to
    iota = 0
    kv = dict()
    classes = []
    for a_idx, orbit in enumerate(permutations):
        canon_orbit = tuple(sorted(orbit))
        if canon_orbit not in kv:
            kv[canon_orbit] = iota
            iota += 1

        classes.append(kv[canon_orbit])

    return classes
        
print(resonance_aware_rank(Chem.AddHs(Chem.MolFromSmiles("[O-]C(=O)C"))))
print(resonance_aware_rank(Chem.AddHs(Chem.MolFromSmiles("CO[N+]([O-])=O"))))
```

maxentile:
Thanks @proteneer and @bp-kelley -- will try this out!

(Just want to strengthen the test to assert that the difference between old and new symmetrization behavior is indeed limited to stereo-awareness.)

PR #739, opened 2022-05-10
Apply bccs both ways

maxentile:
Another possible workaround for the issue in https://github.com/proteneer/timemachine/pull/737 , suggested in https://github.com/proteneer/timemachine/pull/737#issue-1228477107 :
> Modify definition of AM1CCCHandler so that it's safe for arbitrary SMARTS patterns to have arbitrary parameter values (e.g. by matching both directions rather than an arbitrary direction)

This is simpler than adding special case code to detect symmetric bond smarts and avoid applying the correction in an arbitrary direction.

In the symmetric case where a smarts pattern can match the same bond both directions, charges will be incremented by `(+param) + (-param)`, effectively ignoring `param`

### DIFF
--- a/tests/test_handlers.py
+++ b/tests/test_handlers.py
@@ -899,3 +899,25 @@ def test_lennard_jones_handler():
     # if a parameter is > 99 then its adjoint should be zero (converse isn't necessarily true since)
     mask = np.argwhere(params > 90)
     assert np.all(adjoints[mask] == 0.0)
+
+
+def test_symmetric_am1ccc():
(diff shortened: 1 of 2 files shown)


proteneer:
Out of curosity, with this patch, do the gradients (when backprop'd to the symmetric patterns), also cancel out to zero?

suppose f(x) = x - x then f'(x)=0 everywhere I'd hope.

jkausrelay:
@maxentile  Nice, I like the simplicity of this approach.

proteneer:
perfect!

jkausrelay (approved):

badisa (approved):

PR #741, opened 2022-05-12
Adds `execute_coords_batch` and `execute_params_batch`

badisa:
- Adds methods to batch over coords and params. Made separate to avoid nuances in determining which coords/params pair the returned values come from.
- Adds benchmarks for all of the potentials


timemachine/cpp/src/potential.cu
```diff
@@ -11,28 +11,42 @@ namespace timemachine {
...
 
-void Potential::execute_coords_batch_host(
-    const int K,
-    const int N,
-    const int P,
-    const double *h_x,           // [K, N,3]
-    const double *h_p,           // [P,]
-    const double *h_box,         // [K, 3, 3]
+// COORDS_BATCH flag of true indicates to batch over coords, false indicates to batch over parameters
+template <bool COORDS_BATCH>
```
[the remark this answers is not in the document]
maxentile: Enum seems preferable, to handle case of batching over lambda.

Another batch style we may want to distinguish here:
* `[NonbondedInteractionGroup(x_lig, x_env) for x_lig in lig_coords_batch]` (with fixed `x_env`) vs.
* `[NonbondedInteractionGroup(x[lig], x[env]) for x in full_coords_batch]`

badisa:
> * Should there also be batched modes for `execute_selective`? (e.g. if we just want to compute the energies on a batch of coords, without returning a large batch of dUdx, dUdparams)

Would be easy to add.

timemachine/cpp/src/potential.cu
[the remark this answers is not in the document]
badisa: Regarding the `NonbondedInteractionGroup` methods, do you mean that `x_lig` is different coords idxs or just where the ligand atoms are all that change? In the latter case this implementation should perform the same. The NBlist will be rebuilt as long as a single atom has moved enough, whether that is a host or ligand atom.

tests/test_benchmark.py
```diff
@@ -5,30 +5,152 @@
...
 from timemachine.lib import LangevinIntegrator, MonteCarloBarostat, custom_ops
+from timemachine.lib.potentials import Nonbonded, NonbondedInteractionGroup, NonbondedInteractionGroupInterpolated
 from timemachine.md import builders, minimizer
 from timemachine.md.barostat.utils import get_bond_list, get_group_indices
 from timemachine.testsystems.dhfr import setup_dhfr
+from timemachine.testsystems.relative import setup_hif2a_ligand_pair
 
 
-def recenter(conf, box):
+@pytest.fixture(autouse=True, scope="module")
```
mcwitt, line 22: I'm not sure I understand why `autouse=True` is needed here?

timemachine/cpp/src/potential.cu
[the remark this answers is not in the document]
maxentile: > Regarding the NonbondedInteractionGroup methods, do you mean that x_lig is different coords idxs or just where the ligand atoms are all that change?

Second case

>In the latter case this implementation should perform the same. The NBlist will be rebuilt as long as a single atom has moved enough, whether that is a host or ligand atom.

Even if nblist rebuild condition is identical, I guessed one might still be cheaper than the other since there's much less data to transfer...

Point isn't to suggest applying optimizations for this case, but to confirm there are likely to be >2 relevant batch modes -- enum seems preferable to bool.

tests/test_benchmark.py
```diff
@@ -5,30 +5,152 @@
...
+    boxes,
+    lamb,
+    verbose=True,
+    num_batches=5,
+):
+    unbound = potential.unbound_impl(precision=precision)
+    start = time.time()
+    batch_times = []
+    frames_per_batch = coords.shape[0]
+    for _ in range(num_batches):
```
mcwitt, line 99: Is it worth extracting the timing logic into a helper function, since it's replicated in each benchmark?

[timeit](https://docs.python.org/3/library/timeit.html) in the standard library might be useful here. E.g. something like
```python
    def run_batch():
        unbound.execute_coords_batch(
            coords,
            params,
            boxes,
            lamb,
        )

    mean_time = timeit.timeit("run_batch", number=num_batches)
```
(note this also disables garbage collection in the test loop)

tests/test_benchmark.py
```diff
@@ -280,7 +400,199 @@ def test_hif2a():
...
+def test_nonbonded_potential(hi2fa_test_frames):
+    pots, sys_params, frames, boxes, _ = hi2fa_test_frames
+
+    nonbonded_interp = pots[3]
+    lamb = 0.0
+    nonbonded_params = sys_params[3]
+
+    precisions = [np.float32, np.float64]
+
+    potential = Nonbonded(
```
mcwitt, line 491: Should we also test the interpolated variant?

tests/test_benchmark.py
```diff
@@ -280,7 +400,199 @@ def test_hif2a():
...
+        np.zeros(num_atoms, dtype=np.int32),
+        beta,
+        cutoff,
+    )
+    class_name = potential.__class__.__name__
+    for precision in precisions:
+        if precision == np.float32:
+            label = class_name + "_f32"
+        else:
+            label = class_name + "_f64"
```
mcwitt, line 425: Could this logic be moved into the `benchmark_potential_over_*` functions to reduce duplication?

mcwitt (commented):
Nice! This makes sense to me as a pragmatic performance improvement.

Longer-term, I'm wondering how far we want to go down the path of moving simulation / batch loops into the C++ code, versus working to solve the Python/C++ interface inefficiency (e.g. by introducing some Python handle to device memory like jax's `DeviceArray`, or by interfacing directly with XLA as @maxentile has suggested). The latter options seem preferable for flexibility and development speed, though a bit more of an unknown (for me at least) in terms of difficulty

tests/test_benchmark.py
badisa: It tests the interpolated version here, it isn't constructed since it is already constructed: https://github.com/proteneer/timemachine/blob/feature/add-batch-execute-to-potentials/tests/test_benchmark.py#L527-L551

tests/test_benchmark.py
```diff
@@ -280,7 +400,199 @@ def test_hif2a():
...
+            boxes[0],
+            lamb,
+            verbose=False,
+        )
+
+
+def test_nonbonded_potential(hi2fa_test_frames):
+    pots, sys_params, frames, boxes, _ = hi2fa_test_frames
+
+    nonbonded_interp = pots[3]
```
mcwitt, line 485: Would it be worth adding `assert isinstance(nonbonded_interp, NonbondedInterpolated)` to prevent breakage if the order changes?

tests/test_benchmark.py
badisa: Resolved in [70a31c1](https://github.com/proteneer/timemachine/pull/741/commits/70a31c1d95c6cbeb51e57a6ac0d944e9c06295e5)

tests/test_benchmark.py
badisa: The replication is reduced now. Seems cleaner to benchmark this in the same sort of context that we would run it, rather than using `timeit`.

tests/test_benchmark.py
badisa: Resolved in [b4d7ac3](https://github.com/proteneer/timemachine/pull/741/commits/b4d7ac3b6554de025049e14ad10623b08d071fb5)

tests/test_benchmark.py
badisa: Resolved in [3bf17ef](https://github.com/proteneer/timemachine/pull/741/commits/3bf17ef6f23380471ea241680180119ddbcc988e)

mcwitt (approved):

timemachine/cpp/src/potential.cu
```diff
@@ -1,16 +1,112 @@
...
+    const int P,                 // Number of parameters
+    const int lambda_batches,    // Number of lambda values
+    const double *h_x,           // [coord_batches, N, 3]
+    const double *h_p,           // [param_batches, P]
+    const double *h_box,         // [coord_batches, 3, 3]
+    const double *lambdas,       // [lambda_batches]
+    unsigned long long *h_du_dx, // [coord_batches, param_batches, lambda_batches, N, 3]
+    unsigned long long *h_du_dp, // [coord_batches, param_batches, lambda_batches, P]
+    unsigned long long *h_du_dl, // [coord_batches, param_batches, lambda_batches, N]
+    unsigned long long *h_u) {   // [coord_batches, param_batches, lambda_batches, N]
```
proteneer, line 28: nice! This looks like it implements the "outer/cartesian product" of the batches. Which I think covers most use cases (Eg. computing a frame at different values of lambda). Maybe later on there will be a use case for the "inner-product" but probably lower priority.

timemachine/cpp/src/potential.cu
```diff
@@ -1,16 +1,112 @@
...
+
+    if (h_u) {
+        d_u_buffer.reset(new DeviceBuffer<unsigned long long>(K * N));
+        gpuErrchk(cudaMemsetAsync(d_u_buffer->data, 0, d_u_buffer->size, stream));
+    }
+
+    for (unsigned int i = 0; i < coord_batches; i++) {
+        for (unsigned int j = 0; j < param_batches; j++) {
+            for (unsigned int k = 0; k < lambda_batches; k++) {
+                unsigned int offset_factor = (i * param_batches * lambda_batches) + (j * lambda_batches) + k;
```
proteneer, line 74: nice! had to double check this but looks correct to me!

timemachine/cpp/src/potential.cu
proteneer: Actually we already implicitly inner-product the coords and box - (i.e. they're always tied together, we don't evaluate a single conformation over multiple boxes)

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -381,6 +381,174 @@ void declare_potential(py::module &m) {
...
+            A multi dimensional array containing a batch of parameters. First dimension
+            determines the batch size, the rest of the array is passed to the potential as the
+            parameters.
+
+        boxes: NDArray
+            A three dimensional array containing a batch of boxes.
+
+        lambdas: NDArray
+            An array of lambda values.
+
```
proteneer, line 525: nit: extra line here

proteneer (approved):
minor comments only - but this looks great!

tests/test_benchmark.py
```diff
@@ -5,30 +5,136 @@
...
+    coords,
+    boxes,
+    lambdas,
+    verbose=True,
+    num_batches=5,
+    compute_du_dx=True,
+    compute_du_dp=True,
+    compute_du_dl=True,
+    compute_u=True,
+):
```
maxentile, line 103: Documentation request: add one-line description

tests/test_benchmark.py
```diff
@@ -5,30 +5,136 @@
...
+        batch_end = time.time()
+        delta = batch_end - batch_start
+
+        batch_times.append(delta)
+        runs_per_second = runs_per_batch / np.mean(batch_times)
+
+        if verbose:
+            print(f"executions per second: {runs_per_second:.3f}")
+    print(
+        f"{label}: N={coords.shape[1]} Frames={frames} Params={param_batches} Lambdas={num_lambs} speed: {runs_per_second:.2f} executions/seconds (ran {runs_per_batch * num_batches} potentials in {(time.time() - start):.2f}s)"
```
maxentile, line 136: Nit: preference between `sum(batch_times)` and `time.time() - start`? (to exclude some setup and printing)

timemachine/cpp/src/potential.cu
maxentile, line 28: Nit: naming, would `batch_size` be more accurate than `batches`?

(Related curiosity: if these imply a result that is too large to be stored on the device at once, I wonder if this (or perhaps better, the `execute_batch_host` function) should split the job into multiple smaller batches...)

timemachine/cpp/src/potential.cu
```diff
@@ -1,16 +1,112 @@
...
+    }
+
+    if (h_u) {
+        d_u_buffer.reset(new DeviceBuffer<unsigned long long>(K * N));
+        gpuErrchk(cudaMemsetAsync(d_u_buffer->data, 0, d_u_buffer->size, stream));
+    }
+
+    for (unsigned int i = 0; i < coord_batches; i++) {
+        for (unsigned int j = 0; j < param_batches; j++) {
+            for (unsigned int k = 0; k < lambda_batches; k++) {
```
maxentile, line 73: I think this is a good default loop ordering, and if we want to reorder these for any reason from calling code, can reduce this to nested loops over one argument at a time (`coord_batches=1, param_batches=B, lambda_batches=1`) cases.

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -381,6 +381,173 @@ void declare_potential(py::module &m) {
...
+            py::arg("coords"),
+            py::arg("params"),
+            py::arg("boxes"),
+            py::arg("lambs"),
+            py::arg("compute_du_dx"),
+            py::arg("compute_du_dp"),
+            py::arg("compute_du_dl"),
+            py::arg("compute_u"),
+            R"pbdoc(
+        Execute the potential over a batch of coords, parameters and lambdas.
```
maxentile, line 508: Documentation: Would emphasize earlier that this function does `coords_batches * param_batches * lambda_batches` work.

(The return type documentation clarifies this, but a possible user error could still be to assume that it does `batch_size` work, with `coords_batches == param_batches == lambda_batches == batch_size`.)

timemachine/cpp/src/potential.cu
```diff
@@ -1,16 +1,112 @@
...
+
+    DeviceBuffer<double> d_x_buffer(coord_batches * N * D);
+    d_x_buffer.copy_from(h_x);
+
+    std::unique_ptr<DeviceBuffer<unsigned long long>> d_du_dx_buffer(nullptr);
+    std::unique_ptr<DeviceBuffer<unsigned long long>> d_du_dp_buffer(nullptr);
+    std::unique_ptr<DeviceBuffer<unsigned long long>> d_du_dl_buffer(nullptr);
+    std::unique_ptr<DeviceBuffer<unsigned long long>> d_u_buffer(nullptr);
+
+    const int K = coord_batches * param_batches * lambda_batches;
```
maxentile, line 46: Nit: possible confusion with later loop nest, where lower-case `k` indexes the innermost loop, rather than overall progress within the entire loop nest.

(Possibly better to use a different name for total number of iterations than upper case version of `i`, `j`, or `k`.)

timemachine/cpp/src/potential.cu
```diff
@@ -1,16 +1,112 @@
...
+    if (h_u) {
+        d_u_buffer.reset(new DeviceBuffer<unsigned long long>(K * N));
+        gpuErrchk(cudaMemsetAsync(d_u_buffer->data, 0, d_u_buffer->size, stream));
+    }
+
+    for (unsigned int i = 0; i < coord_batches; i++) {
+        for (unsigned int j = 0; j < param_batches; j++) {
+            for (unsigned int k = 0; k < lambda_batches; k++) {
+                unsigned int offset_factor = (i * param_batches * lambda_batches) + (j * lambda_batches) + k;
+                this->execute_device(
```
maxentile, line 75: Curiosity: Does this "enqueue" all work items from the batch for parallel execution, or are there sync barriers between loop iterations?

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -381,6 +381,173 @@ void declare_potential(py::module &m) {
...
+                }
+                if (params.ndim() < 2) {
+                    throw std::runtime_error("parameters must have at least 2 dimensions");
+                }
+                if (lambdas.size() < 1) {
+                    throw std::runtime_error("lambdas must have at least 1 value");
+                }
+                const long unsigned int coord_batches = coords.shape()[0];
+                const long unsigned int N = coords.shape()[1];
+                const long unsigned int D = coords.shape()[2];
```
maxentile, line 409: Would it be useful to assert D is 3 or 4? (In case user-provided array has erroneous shape.)

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -381,6 +381,173 @@ void declare_potential(py::module &m) {
...
+
+                const long unsigned int param_batches = params.shape()[0];
+                const long unsigned int P = params.size() / param_batches;
+
+                const long unsigned int lambda_batches = lambdas.size();
+
+                const long unsigned int K = coord_batches * param_batches * lambda_batches;
+
+                // initialize with fixed garbage values for debugging convenience (these should be overwritten by `execute_host`)
```
maxentile, line 418: nit
```suggestion
                // initialize with fixed garbage values for debugging convenience (these should be overwritten by `execute_batch_host`)
```

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -381,6 +381,173 @@ void declare_potential(py::module &m) {
...
+                    du_dp.resize(K * P, 9999);
+                }
+                std::vector<unsigned long long> du_dl;
+                if (compute_du_dl) {
+                    du_dl.resize(K * N, 9999);
+                }
+                std::vector<unsigned long long> u;
+                if (compute_u) {
+                    u.resize(K * N, 9999);
+                }
```
maxentile, line 435: A little tripped up why these two are `K * N` rather than just `K`, but I think this is consistent the logic already present in the `execute` wrapper. May still be worth a line comment?

maxentile (approved):
Looks great to me! Should provide an immediate speed-up for several use cases, and establish a baseline for any further performance improvements.

(Left a few minor non-blocking comments, mostly about naming and documentation.)

timemachine/cpp/src/potential.cu
badisa: Will rename to `*_batch_size`

Right now if you set a result that is too large, this will crash. We could try and track memory usage in the future, though that is usually a fraught task (at least with CPU memory).

timemachine/cpp/src/wrap_kernels.cpp
badisa: I wanted to leave that sort of logic to the potential, as we have checks at that level. Wanted to avoid too much repetition of checks and leave the most stringent checks to the derived potentials.

timemachine/cpp/src/wrap_kernels.cpp
badisa: Will add a comment

timemachine/cpp/src/potential.cu
badisa: Correct, this places all of the executions in the same stream and then synchronizes afterwards.

tests/test_benchmark.py
badisa: I prefer the latter as it is consistent with the other benchmarking. We may want to revisit how we do benchmarking in the next deboggling week

PR #742, opened 2022-05-16
Validate harmonic bonds

maxentile:
Raise ValueError in `HarmonicBondHandler.static_parameterize(... mol)` if the ~number~ set of harmonic bonds is not equal to ~`mol.GetNumBonds()`~ the set contained in `mol.GetBonds()`.

Unclear how widespread the issue might be, but it seems that:
* when there are no smirks matches for a bond, the current harmonic bond handler will silently fail to assign parameters to the bond
* this occurs at least for [oxygen] ~ [halogen] bonds (e.g. the molecule `smi="O(F)F"` is assigned 0 harmonic bonds)
* the proposed change does not raise any new errors on FreeSolv

### DIFF
--- a/tests/test_handlers.py
+++ b/tests/test_handlers.py
@@ -10,7 +10,7 @@ import pytest
 from rdkit import Chem
 from rdkit.Chem import AllChem, rdmolops
 
-from timemachine.constants import ONE_4PI_EPS0
+from timemachine.constants import DEFAULT_FF, ONE_4PI_EPS0
 from timemachine.ff import Forcefield
 from timemachine.ff.charges import AM1CCC_CHARGES
 from timemachine.ff.handlers import bonded, nonbonded
@@ -919,3 +919,25 @@ def test_symmetric_am1ccc():
     # set(ref_charges) == {-1.8165082, -1.8158009, 0.90795946}
     # set(test_charges) == {-3.815801, -1.8158009, 0.18349183, 0.90795946}
     np.testing.assert_array_equal(test_charges, ref_charges)
+
+
+def test_harmonic_bonds_complete():
+    """On a test molecule containing [oxygen] ~ [halogen] bonds,
+    assert that either: (1) a ValueError is raised or (2) the number of
+    parameterized bonds equals the number of input bonds"""
+
+    mol = Chem.MolFromSmiles("O(F)F")  # 0 smirks matches using current (2022-05-16) handler
+
+    ff = Forcefield.load_from_file(DEFAULT_FF)
+    parameterize = ff.hb_handle.parameterize
+
+    try:
+        with pytest.raises(ValueError) as e:
+            _, _ = parameterize(mol)
+
+            assert "number of bonds" in e
+    except Exception:
+        expected_num_bonds = mol.GetNumBonds()
+        _, idxs = parameterize(mol)
+
+        assert len(idxs) == expected_num_bonds
--- a/tests/test_handlers.py
+++ b/tests/test_handlers.py
@@ -932,4 +932,4 @@ def test_harmonic_bonds_complete():
 
     with pytest.raises(ValueError) as e:
         _, _ = parameterize(mol)
-    assert "number of bonds" in str(e)
+    assert "missing bonds" in str(e)
--- a/tests/test_handlers.py
+++ b/tests/test_handlers.py
@@ -923,21 +923,14 @@ def test_symmetric_am1ccc():
 
 def test_harmonic_bonds_complete():
     """On a test molecule containing [oxygen] ~ [halogen] bonds,
-    assert that either: (1) a ValueError is raised or (2) the number of
-    parameterized bonds equals the number of input bonds"""
+    assert that a ValueError is raised."""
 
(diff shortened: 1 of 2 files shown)


proteneer:
Should we do something similar for angles?

maxentile:
>Should we do something similar for angles?

Possibly -- would that be an error or a warning?

For bonds it's clear this should be an error, since
(1) bonds are a key aspect of the input that should be preserved,
(2) if harmonic bonds are missing the molecule will "dissociate" during simulation.

It's less clear to me for other terms since:
(1) the input `mol` doesn't contain "angles" or "torsions",
(2) if higher-order terms terms are missing, the molecule will just be more "floppy," but won't dissociate,
(3) the AM1CCCHandler doesn't currently match all bonds (e.g. on 14 molecules in FreeSolv)

tests/test_handlers.py
```diff
@@ -919,3 +919,25 @@ def test_symmetric_am1ccc():
...
+    """On a test molecule containing [oxygen] ~ [halogen] bonds,
+    assert that either: (1) a ValueError is raised or (2) the number of
+    parameterized bonds equals the number of input bonds"""
+
+    mol = Chem.MolFromSmiles("O(F)F")  # 0 smirks matches using current (2022-05-16) handler
+
+    ff = Forcefield.load_from_file(DEFAULT_FF)
+    parameterize = ff.hb_handle.parameterize
+
+    try:
```
badisa, line 934: I find this test confusing. What lines actually run and which raise exceptions?

timemachine/ff/handlers/bonded.py
```diff
@@ -76,6 +76,14 @@ def static_parameterize(params, smirks, mol):
         if len(mol_params) == 0:
             mol_params = params[:0]  # empty slice with same dtype, other dimensions
             bond_idxs = np.zeros((0, 2), dtype=np.int32)
+
+            expected_num_bonds = mol.GetNumBonds()
+            assigned_num_bonds = len(bond_idxs)
```
badisa, line 81: Nit: its a bit confusing that this is dynamic but is always zero per line 78

badisa: Is this suppose to be unindented??

maxentile: Oof, good catch -- this shouldn't have been inside the `if len(mol_params) == 0` branch! Reordered and fixed in https://github.com/proteneer/timemachine/pull/742/commits/727449df41ff10eac22849c29b9ca3ffba548b80

tests/test_handlers.py
maxentile: Currently, just one branch runs. (The one where pytest expects a ValueError.)

I initially included a branch in case the handler is modified to another correct behavior in the future (return 2 bonds: pass, raise ValueError: pass, return 0 bonds: fail)

Simplified in https://github.com/proteneer/timemachine/pull/742/commits/7608223abf871d472fb2340b8e98c985877c5835 to (raise ValueError: pass, else: fail)

tests/test_handlers.py
```diff
@@ -923,21 +923,14 @@ def test_symmetric_am1ccc():
...
+    with pytest.raises(ValueError) as e:
+        _, _ = parameterize(mol)
 
-            assert "number of bonds" in e
-    except Exception:
-        expected_num_bonds = mol.GetNumBonds()
-        _, idxs = parameterize(mol)
-
-        assert len(idxs) == expected_num_bonds
+        assert "number of bonds" in e
```
badisa, line 936: I believe this needs to be outside of the context manager for it to run, else the value error out of parameterize will skip it.

badisa (approved):
Good catch!

timemachine/ff/handlers/bonded.py
```diff
@@ -73,9 +73,20 @@ class HarmonicBondHandler(ReversibleBondHandler):
     @staticmethod
     def static_parameterize(params, smirks, mol):
         mol_params, bond_idxs = super(HarmonicBondHandler, HarmonicBondHandler).static_parameterize(params, smirks, mol)
+
+        # validate expected number of bonds
+        expected_num_bonds = mol.GetNumBonds()
+        assigned_num_bonds = len(bond_idxs)
+        if assigned_num_bonds != expected_num_bonds:
```
proteneer, line 80: strictly speaking, we probably don't want to just assert that the # of bonds are the same, but that their connectivities are the same

```
rd_bonds = set()
for b in mol.GetBonds():
    rd_bonds.add(tuple(sorted([b.GetBeginAtomIdx(), b.GetEndAtomIdx()])))

ff_bonds = set()
for i,j in bond_idxs:
        rd_bonds.add(tuple(sorted([i,j])))

assert rd_bonds == ff_bonds
```

proteneer (approved):
Minor comment, but this PR is still an improvement nevertheless.

timemachine/ff/handlers/bonded.py
maxentile: Good call! Applied in https://github.com/proteneer/timemachine/pull/742/commits/10153916345f8342e202f9bc2ad16a8e2b87df86

PR #744, opened 2022-05-17
Validate harmonic angle idxs, refactor idx canonicalization

maxentile:
Based on suggestion https://github.com/proteneer/timemachine/pull/742#issuecomment-1128356628 , adds a check for completeness of the `angle_idxs` returned by `HarmonicAngleHandler`. This raises an error unless angle terms are present for all canonical angle tuples `(a, b, c)`.

Side quest: Also attempted to improve naming of index canonicalization function (`canonicalize_bond((a, b, c))` -> `canonicalize_angle((a, b, c))` or `canonicalize_bonded_ixn((a, b, c))`).

Side-side quest: Ran into a hiccup while adding `canonicalize_improper_torsion` -- will need to look more carefully at the logic of the two sites I'm extracting the definition from, or skip adding this for now.

### DIFF
--- a/timemachine/ff/handlers/utils.py
+++ b/timemachine/ff/handlers/utils.py
@@ -59,26 +59,42 @@ def canonicalize_proper_torsion(torsion):
 
 @wraps(canonicalize_bonded_ixn)
 def canonicalize_improper_torsion(torsion):
-    """
-    WIP: extract from two definitions:
-    1. https://github.com/proteneer/timemachine/blob/8d6bd25a143aa81e8b3b8c6a33e6e03afe272c56/timemachine/ff/handlers/bonded.py#L226-L229
-    2. https://github.com/proteneer/timemachine/blob/8d6bd25a143aa81e8b3b8c6a33e6e03afe272c56/timemachine/ff/handlers/bonded.py#L206-L213
+    """(a, b, c, d) -> (a', b, c', d')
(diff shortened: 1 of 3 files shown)


timemachine/ff/handlers/utils.py
```diff
@@ -32,6 +39,68 @@ def canonicalize_bond(arr):
...
+    a, c, d = sorted([_a, _c, _d])
+    canonicalized_1 = container_type([a, b, c, d])
+
+    # approach 2: take min((b, a, c, d), (d, c, a, b)) (used for applying trefoil convention)
+    a, b, c, d = torsion
+    canonicalized_2 = canonicalize_bonded_ixn([b, a, c, d])
+
+    # TODO: which one?
+    # return canonicalized
+    raise NotImplementedError(f"can't decide between {canonicalized_1} and {canonicalized_2}!")
```
maxentile, line 81: I think it makes sense to go with approach 1 here, and possibly add a separate function to enumerate the trefoil permutations

maxentile: Have now split this into a couple functions, and added documentation on points that tripped me up

maxentile:
Superseded by #762

proteneer:
Note that #762 operates at the level of the "combined single topology" molecule, not on the handler level.

PR #745, opened 2022-05-18
Parallel SMC updates

jkausrelay:
- Use a FileClient to manage file I/O (which may be replaced by a different client)
- Make sure the ff is passed by content instead of file name
- Pass back a list of files stored
- Copy the files if run on a remote server
- Add dill

### DIFF
--- a/examples/run_smc_on_freesolv.py
+++ b/examples/run_smc_on_freesolv.py
@@ -125,7 +125,7 @@ def save_smc_result(
             pickle.dump(smc_result, f)
 
 
-def run_on_freesolv_mol(mol: Chem.rdchem.Mol, cmd_args: argparse.Namespace):
+def run_on_freesolv_mol(mol: Chem.rdchem.Mol, ff: Forcefield, cmd_args: argparse.Namespace):
     name = get_mol_name(mol)
     props = mol.GetPropsAsDict()
     result_path = Path(cmd_args.result_path)
@@ -139,7 +139,7 @@ def run_on_freesolv_mol(mol: Chem.rdchem.Mol, cmd_args: argparse.Namespace):
         cmd_args.n_md_steps,
         cmd_args.resample_thresh,
         seed=cmd_args.seed,
-        ff=get_ff(cmd_args.ff),
+        ff=ff,
         num_workers=cmd_args.n_cpus,
     )
     # run simulation
@@ -162,9 +162,9 @@ def run_on_freesolv_mol(mol: Chem.rdchem.Mol, cmd_args: argparse.Namespace):
     save_smc_result(result_path, mol, smc_result, cmd_args, save_full_trajectories=cmd_args.debug_mode)
 
 
-def run_on_mols(mols: List[Chem.rdchem.Mol], cmd_args: argparse.Namespace):
+def run_on_mols(mols: List[Chem.rdchem.Mol], ff: Forcefield, cmd_args: argparse.Namespace):
     for mol in mols:
-        run_on_freesolv_mol(mol, cmd_args)
+        run_on_freesolv_mol(mol, ff, cmd_args)
 
 
 def main():
@@ -182,10 +182,11 @@ def main():
     print(f"using {num_gpus} gpus with {cmd_args.n_cpus or 'default'} cpus per gpu")
 
     # Batch mols
+    ff = get_ff(cmd_args.ff)
     batch_mols = batch_list(mols, num_gpus)
     futures = []
     for mol_subset in batch_mols:
-        futures.append(client.submit(run_on_mols, mol_subset, cmd_args))
+        futures.append(client.submit(run_on_mols, mol_subset, ff, cmd_args))
 
     # Wait for jobs to complete
     _ = [fut.result() for fut in futures]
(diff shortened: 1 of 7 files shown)


examples/run_smc_on_freesolv.py
```diff
@@ -159,12 +178,16 @@ def run_on_freesolv_mol(mol: Chem.rdchem.Mol, cmd_args: argparse.Namespace):
     smc_result["final_samples_refined"] = final_samples_refined
 
     # save summary
-    save_smc_result(result_path, mol, smc_result, cmd_args, save_full_trajectories=cmd_args.debug_mode)
+    return save_smc_result(file_client, mol, smc_result, cmd_args, save_full_trajectories=cmd_args.debug_mode)
```
badisa, line 181: nit: Seems odd to run a function that saves a result

timemachine/parallel/utils.py
```diff
@@ -43,3 +43,7 @@ def batch_list(values: List, num_workers: int) -> List[List]:
     for i, value in enumerate(values):
         batched_values[i % num_workers].append(value)
     return list(batched_values.values())
+
+
+def flatten_list(results: List[List[str]]) -> List[str]:
```
badisa, line 48: The type should probably be Any

requirements.txt
```diff
@@ -7,3 +7,4 @@ pyyaml==5.4.1
 networkx==2.5
 importlib-resources==5.4.0
 typing-extensions==4.1.1
+dill==0.3.4
```
badisa, line 10: I don't see this used within the repo, do we need this here?

timemachine/parallel/client.py
```diff
@@ -238,3 +239,94 @@ def verify(self):
...
+class FileClient(AbstractFileClient):
+    def __init__(self, base: Path = None):
+        self.base = base or Path().cwd()
+
+    def store(self, path: str, data: bytes):
+        """
+        See abstract class for documentation.
+        """
+        full_path = Path(self.full_path(path))
+        full_path.write_bytes(data)
```
badisa, line 306: The base class defines this as returning a string, but seems less useful to return. Should unify the behavior one way or another.

requirements.txt
jkausrelay: removed

timemachine/parallel/client.py
jkausrelay: removed str return, this was from an earlier version

timemachine/parallel/utils.py
jkausrelay: updated

badisa (approved):

timemachine/parallel/client.py
```diff
@@ -238,3 +239,88 @@ def verify(self):
...
+
+
+class FileClient(AbstractFileClient):
+    def __init__(self, base: Path = None):
+        self.base = base or Path().cwd()
+
+    def store(self, path: str, data: bytes):
+        """
+        See abstract class for documentation.
+        """
```
maxentile, line 298: I think by default, if no docstring is provided here, `help(FileClient.store)` will show the docstring of `AbstractFileClient.store`! May be preferable to remove these?

```suggestion
```

timemachine/parallel/utils.py
```diff
@@ -43,3 +43,7 @@ def batch_list(values: List, num_workers: int) -> List[List]:
     for i, value in enumerate(values):
         batched_values[i % num_workers].append(value)
     return list(batched_values.values())
+
+
+def flatten_list(results: List[List[Any]]) -> List[Any]:
+    return [i for j in results for i in j]
```
maxentile, line 49: Trying to get a better sense for whether this belongs in parallel/utils.py or if it's specific to a structure that exists in examples/run_smc_on_freesolv.py ... 

Would we ever need to flatten a structure that's nested more than one level deep, or has a mix of lists, tuples, dicts, etc?

maxentile (approved):
Looks good to me!

timemachine/parallel/utils.py
```diff
@@ -43,3 +43,7 @@ def batch_list(values: List, num_workers: int) -> List[List]:
     for i, value in enumerate(values):
         batched_values[i % num_workers].append(value)
     return list(batched_values.values())
+
+
+def flatten_list(results: List[List[Any]]) -> List[Any]:
```
mcwitt, line 48: nit: could use a `TypeVar` here for a more precise type (otherwise e.g. `foo: List[str] = flatten_list([[1, 2], [3]])` would typecheck)
```suggestion
A = TypeVar("A")
def flatten_list(results: List[List[A]]) -> List[A]:
```

timemachine/parallel/utils.py
jkausrelay: inlined in run_smc

timemachine/parallel/utils.py
jkausrelay: inlined in run_smc

jkausrelay:
@mcwitt Let me know if you have any more comments, otherwise I'll merge.

examples/run_smc_on_freesolv.py
```diff
@@ -179,16 +196,23 @@ def main():
...
     batch_mols = batch_list(mols, num_gpus)
     futures = []
     for mol_subset in batch_mols:
-        futures.append(client.submit(run_on_mols, mol_subset, cmd_args))
+        futures.append(client.submit(run_on_mols, file_client, mol_subset, ff, cmd_args))
 
     # Wait for jobs to complete
-    _ = [fut.result() for fut in futures]
+    batched_results = [fut.result() for fut in futures]
+    results = [i for j in batched_results for i in j]
```
mcwitt, line 211: nit: naming might improve readability
```suggestion
    results = [result for batch in batched_results for result in batch]
```

mcwitt (approved):
LGTM!

PR #746, opened 2022-05-19
Single Topology Rewrite

proteneer:
This PR re-writes most of the single topology code to be significantly more robust. In particular, this PR defines the end-states (code for efficiently interpolating between end-states will come later).

1) The end-states are now comprised of dummy atoms that are provably factorizable.
2) Proper atom stereochemistry is correctly maintained, so that stereoisomers maintain their configuration at the endstates. Note that while we can handle atom chirality at the junction site, we still can't handle bond chirality (this is detected and an assertion is made).
3) Adding extensions to support ring-opening and closing, core-hopping, etc. should be relatively straight forward.
4) Dummy groups have modified parameters such that they are "enhanced" (i.e. nonbonded interactions are disabled within the group and other groups, torsions are softened, etc.)
5) A new SingleTopologyV2 class is introduced. The old code will removed during deboggling week.

Summary of changes:

- timemachine/fe/geometry.py | Modified so that geometry is read from the forcefield parameters as opposed to the rdkit Hybridization code.
- timemachine/fe/single_topology.py | Core functionality and logic of this PR
- timemachine/fe/utils.py | Minor utilities
- timemachine/potentials/bonded.py | Added functional form for symmetrized cross product and centroid based angle restraints.
- tests/test_geometry.py | One of the tests had to be modified since a discrepancy was found between rdkit and ff
- tests/test_single_topology_stereo.py | Tests that stereochemistry is properly maintained at the end-states when simulating. - 
- tests/test_single_topology_v2.py | Tests parameterization of interactions involving dummy atoms.


timemachine/fe/geometry.py
```diff
@@ -3,56 +3,106 @@
...
+                if j == atom_idx:
+                    if (i, k) == (ii, kk) or (i, k) == (kk, ii):
+                        if abs(angle - np.pi) < 0.05:
+                            local_geometry = LocalGeometry.G2_LINEAR
+                            break
+            atom_geometries.append(local_geometry)
+        elif len(atom_nbs) == 1:
+            atom_geometries.append(LocalGeometry.G1_TERMINAL)
+        elif len(atom_nbs) == 0:
+            atom_geometries.append(None)
```
jkausrelay, line 100: `None` -> `LocalGeometry.G0_ION`?

timemachine/fe/geometry.py
```diff
@@ -64,15 +114,69 @@ def classify_geometry(mol: Chem.Mol) -> List[LocalGeometry]:
...
     Returns
     -------
     List[LocalGeometry]
-        List of per atom geometries
+        List of per atom geometries. Dummy atoms are None
+
 
     """
+    if ff is None:
+        ff = Forcefield.load_from_file("smirnoff_1_1_0_sc.py")
```
jkausrelay, line 128: nit: Use DEFAULT_FF ? (https://github.com/proteneer/timemachine/blob/master/timemachine/constants.py#L14)

timemachine/fe/geometry.py
```diff
@@ -3,56 +3,106 @@
...
+        elif len(atom_nbs) == 0:
+            atom_geometries.append(None)
+
+    return atom_geometries
+
+
+def classify_geometry(mol: Chem.Mol, ff: Forcefield = None, core: List[int] = None) -> List[LocalGeometry]:
     """
     Identify the local geometry of the molecule. This current uses a heuristic but we
     should really be generating this from gas-phase simulations of the real forcefield.
```
jkausrelay, line 108: Is this comment still valid? I.e. we should to run gas-phase simulations then get the geometry?

tests/test_geometry.py
```diff
@@ -90,7 +90,7 @@ def test_assign_truncated_sildenafil():
         LG.G4_TETRAHEDRAL,  # S
         LG.G1_TERMINAL,  # O-S
         LG.G1_TERMINAL,  # O-S
-        LG.G3_PYRAMIDAL,  # N
+        LG.G3_PLANAR,  # N, note, rdkit hybridization thinks this is pyramidal
```
jkausrelay, line 93: So there is an improper which makes it planar? That seems surprising to me.

timemachine/fe/geometry.py
jkausrelay: Nvm: Add a comment # None here indicates a dummy

timemachine/potentials/bonded.py
```diff
@@ -147,6 +151,95 @@ def harmonic_bond(conf, params, box, lamb, bond_idxs, lamb_mult=None, lamb_offse
...
+        x_k = conf[k]
+        vik.append(x_k - x_j)
+
+    vij = jnp.array(vij)
+    vik = jnp.array(vik)
+
+    top = jnp.sum(jnp.multiply(vij, vik), -1)
+    bot = jnp.linalg.norm(vij, axis=-1) * jnp.linalg.norm(vik, axis=-1)
+
+    tb = top / bot
```
maxentile, line 194: Extract function to compute `cos_angle(x_a, x_b, x_c)`, reuse here and in `harmonic_angle`?

(So that responsibility of `get_centroid_cos_angles` can be simplified to just defining the centroid `x_i`)

tests/test_geometry.py
```diff
@@ -1,5 +1,5 @@
 # This file tests geometry classification for a wide variety of molecules,
-# selected from common drugs and baran's heterocyclic chemistry
+# selected from common drugs and baran's heterocyclic chemistry.
```
maxentile, line 2:
```suggestion
# selected from common drugs and Baran's heterocyclic chemistry.
```

maxentile (commented):
First round of comments: There are some examples I want to test on my own, and a few places where I requested documentation changes, but nothing dramatic so far!

tests/test_geometry.py
proteneer: Yes - the improper is probably this one:

```
['[*:1]~[#7X3$(*~[#6X3]):2](~[*:3])~[*:4]', 1.3946666666666667, 3.141592653589793, 2.0],
```

proteneer: Note that presence of an improper doesn't guarantee that it's planar - it depends on the strength of the force constants etc relative to other forces that may be at play. But the intention of impropers is to planarize.

timemachine/fe/geometry.py
proteneer: willfix

proteneer: (actually I wanted to use the _sc one to avoid the need to run AM1 calculations since nonbonded interactions are not considered here).

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,784 @@
...
+                if canon_jk not in planar_bonds:
+                    planar_bonds[canon_jk] = []
+                planar_bonds[canon_jk].append((i, j, k, l))
+
+    return planar_bonds
+
+
+# recursively replace items in a list of list
+def recursive_map(items, mapping):
+    """
```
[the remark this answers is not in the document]
proteneer: nice! I try to avoid using jax internals for stuff like this (and the recursion here is simple enough)

timemachine/fe/geometry.py
proteneer: willfix

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,784 @@
...
+    jkl = dummy.canonicalize_bond((j, k, l))
+
+    if jkl not in canonical_angles:
+        return False
+    elif canonical_angles[jkl][0] < 50.0:
+        return False
+    elif abs(canonical_angles[jkl][1] - 0) < 0.05:
+        return False
+    elif abs(canonical_angles[jkl][1] - 3.1415) < 0.05:
+        return False
```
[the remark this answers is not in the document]
proteneer: I had no idea you could mod over floats in python - let me read up on the behavior more carefully before acceping this commit.

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,784 @@
...
+
+    return True
+
+
+def check_bond_angle_stability(j, k, l, bond_idxs, bond_params, angle_idxs, angle_params):
+    jk_stable = check_bond_stability(j, k, bond_idxs, bond_params)
+    kl_stable = check_bond_stability(k, l, bond_idxs, bond_params)
+    jkl_stable = check_angle_stability(j, k, l, angle_idxs, angle_params)
+    return jk_stable and kl_stable and jkl_stable
```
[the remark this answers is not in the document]
proteneer: Practically, the only case where `j-l` would even be present are cyclopropanes being morphed at the junction site, which seem like an extremely rare case. But most cyclopropanes should be numerically stable under this check regardless. Angles `k-l-j` and `k-j-l` would only be present if `j-l` is present.

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,784 @@
...
+
+    return nbs_1, nbs_2
+
+
+def setup_orientational_restraints(ff, mol_a, mol_b, core, dg, anchor):
+    """
+    Add restraints between dummy atoms in a dummy_group and core atoms.
+
+    mol_a's parameters are mainly used to check for numerical stability.
+    """
```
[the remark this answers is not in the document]
proteneer: > I'm having a hard time reviewing this function due to length and complexity -- can it be broken up into smaller pieces? (Currently ~400 lines.)

I agree there's a lot of cruft (but necessary). But I don't think the verbosity is the difficult part, but rather the varying responsibilities of this function:

1) Setting up the end-state forcefield parameters (independent of dummy atoms)
2) Detecting the geometry of (core and full) and assigning class-specific restraints
3) Verifying numerical stability of the restraints

> can asserting bond_angle_stability be done in a loop all at once at the end, rather than separately for each (anchor_core_geometry, anchor_full_geometry) branch?

Unfortunately this is needed - and specifically for ring-opening and closing:

1) If core_geometry == G1_TERMINAL: there is no way to recover from an unstable core-anchor bond. It usually implies a molecule is being disconnected, which should just assert
2) If core_geometry == G2_KINK or G3_PYRAMIDAL: some choices of core-anchor maybe okay, some are not. This behavior is specific to the geometry class. G2_KINKs have 2 choices of bonds to build cross-product restraints, and G3_PYRAMIDAL have 3 choices of bonds. But not all classes need cross-product based restraints and some eg. (centroid restraints in core_geometry==G2_KINK, full_geometry==G3_PYRAMIDAL) need *both* bonds to be present.

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,784 @@
...
+    )
+    all_params = (
+        restraint_bond_params,
+        restraint_angle_params,
+        restraint_proper_params,
+        restraint_improper_params,
+        restraint_cross_angle_params,
+        restraint_centroid_angle_params,
+    )
+    return all_idxs, all_params
```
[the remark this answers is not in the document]
proteneer: probably! would a named tuple suffice?

timemachine/fe/geometry.py
proteneer: We should eventually classify geometries by running simulations. Right now we simply inspect the forcefield parameters pray to the stat mech gods.

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,784 @@
...
+    torsions = dict()
+    for idxs, params in zip(proper_idxs, proper_params):
+        idxs = tuple(idxs.tolist())
+        if idxs not in torsions:
+            torsions[idxs] = []
+        torsions[idxs].append(params)
+
+    for (i, j, k, l), all_params in torsions.items():
+        # planarizing torsions should be unique
+        if len(all_params) == 1:
```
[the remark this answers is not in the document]
proteneer: probably - but I'm not sure what to do with them since the default behavior here is to skip

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,784 @@
...
+def find_stereo_atoms(mol):
+    mol_geom = geometry.classify_geometry(mol)
+    stereo_atoms = set()
+    for a in mol.GetAtoms():
+        a_idx = a.GetIdx()
+        nbs = a.GetNeighbors()
+        if len(nbs) == 4:
+            stereo_atoms.add(a_idx)
+        elif len(nbs) == 3 and mol_geom[a_idx] == LocalGeometry.G3_PYRAMIDAL:
+            # if in ring, or is sulfur or phosphorus, this may not be guaranteed by
```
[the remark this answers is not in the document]
proteneer: willfix

timemachine/potentials/bonded.py
proteneer: Will refactor!

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,783 @@
...
+
+def identify_bonds_spanned_by_planar_torsions(proper_idxs, proper_params):
+    """
+    Identify bonds that are spanned by planar torsions and returns a dict of bonds
+    and associated torsions that span it.
+    """
+    planar_bonds = dict()
+
+    # collect unique idxs first
+    torsions = dict()
```
jkausrelay, line 21: nit: `torsions = defaultdict(list)`

then you can remove `if idxs not in torsions: ...`

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,783 @@
...
+            #           \
+            #            b
+            #
+            # we have some choices here:
+            # 1) if there is no ring-opening and closing, then we use a centroid angle restraint defined by [a,b,c],j,k
+            # 2) if there is ring-opening and closing, then we need to enumerate possible cross-product based restraints.
+            # we currently support only 1) right now. But it would not be difficult to implement 2).
+
+            # it turns out that 1) is quite difficult, since ammonium groups do not have reasonable centroids due to
+            # rapid conversion. So we should always do 2) as it's more robust and enables more kinds of transformations.
```
jkausrelay, line 525: update doc

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,783 @@
...
+        A tuple of bonded, angle, proper, c_angle, x_angle idxs and parameters.
+
+    """
+    mol_b_top = topology.BaseTopology(mol_b, ff)
+
+    _, mol_b_hb = mol_b_top.parameterize_harmonic_bond(ff.hb_handle.params)
+    mol_b_bond_idxs = mol_b_hb.get_idxs()
+
+    mol_b_core = core[:, 1]
+    mol_b_bond_idxs = [(b.GetBeginAtomIdx(), b.GetEndAtomIdx()) for b in mol_b.GetBonds()]
```
jkausrelay, line 614: use `get_romol_bonds`?

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,783 @@
...
+        assert core.shape[1] == 2
+
+        # map into idxs in the combined molecule
+        self.a_to_c = np.arange(mol_a.GetNumAtoms(), dtype=np.int32)  # identity
+        self.b_to_c = np.zeros(mol_b.GetNumAtoms(), dtype=np.int32) - 1
+
+        self.NC = mol_a.GetNumAtoms() + mol_b.GetNumAtoms() - len(core)
+
+        # mark membership in the combined molecule:
+        # 0: CORE
```
jkausrelay, line 735: Would it make sense to have an enum for this?

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,783 @@
...
+        (mol_c_bond_idxs, mol_c_bond_params),
+        (mol_c_angle_idxs, mol_c_angle_params),
+        (mol_c_proper_idxs, mol_c_proper_params),
+        (mol_c_improper_idxs, mol_c_improper_params),
+        (mol_c_x_angle_idxs, mol_c_x_angle_params),
+        (mol_c_c_angle_idxs, mol_c_c_angle_params),
+    )
+
+
+class SingleTopologyV2:
```
jkausrelay, line 697: Would this stay here or be moved to fe.topology?
Would the current SingleTopology be removed?

tests/test_geometry.py
```diff
@@ -493,6 +493,58 @@ def test_baran_tetrazole():
     assert atom_types == expected_types
 
 
+def test_aniline():
+    # although aniline is slightly pyramidal, we'd rather it
+    # be assigned planar for safety.
+    mol = Chem.AddHs(Chem.MolFromSmiles("c1ccccc1N"))
+    atom_types = geometry.classify_geometry(mol)
```
mcwitt, line 500: nit: should the naming be updated from `atom_types` -> `atom_geometries` as above?

tests/test_single_topology_stereo.py
```diff
@@ -0,0 +1,486 @@
...
+
+def minimize_bfgs(x0, U_fn):
+    N = x0.shape[0]
+
+    def u_bfgs(x_flat):
+        x_full = x_flat.reshape(N, 3)
+        return U_fn(x_full)
+
+    grad_bfgs_fn = jax.grad(u_bfgs)
+    res = scipy.optimize.minimize(u_bfgs, x0.reshape(-1), jac=grad_bfgs_fn)
```
mcwitt, line 27: nit: would it make sense to specify `method="L-BFGS-B"`, or update the name of the function to just something like `minimize_potential`?

tests/test_single_topology_stereo.py
```diff
@@ -0,0 +1,486 @@
...
+    x_b = utils.get_romol_conf(mol_b)
+    x0 = s_top.combine_confs(x_a, x_b)
+
+    vol_cl = measure_chiral_volume(x0[0], x0[1], x0[2], x0[3])
+    vol_br = measure_chiral_volume(x0[0], x0[1], x0[2], x0[4])
+
+    assert vol_cl > 0 and vol_br < 0
+
+    idxs_and_params = s_top.generate_end_state_mol_a()
+    frames = simulate_idxs_and_params(idxs_and_params, x0)
```
mcwitt, line 181: I don't see where we set a random seed here. Should we do that to ensure test determinism?

mcwitt (commented):
Left some comments from the initial pass!

tests/test_single_topology_stereo.py
```diff
@@ -0,0 +1,486 @@
...
+    num_vol_a_neg = 0
+
+    for f in frames:
+        vol_a = measure_chiral_volume(f[0], f[1], f[2], f[3])
+        vol_d = measure_chiral_volume(f[0], f[1], f[2], f[5])
+        if vol_a < 0:
+            num_vol_a_neg += 1
+        else:
+            num_vol_a_pos += 1
+        assert vol_d < 0
```
jkausrelay, line 463: turn this into a function, since it's repeated a few times

timemachine/fe/single_topology.py
proteneer: willfix

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,783 @@
...
+
+def enumerate_anchor_groups(anchor_idx, bond_idxs, core_idxs):
+
+    # enumerate all 1 and 2 neighbor anchor atoms to form valid anchor groups.
+
+    assert anchor_idx in core_idxs
+    assert anchor_idx in [x[0] for x in bond_idxs] or anchor_idx in [x[1] for x in bond_idxs]
+
+    nbs_1 = set()
+    nbs_2 = set()
```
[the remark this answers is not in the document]
proteneer: willfix

timemachine/fe/single_topology.py
proteneer: willfix

timemachine/fe/single_topology.py
proteneer: sure - i don't see why not

timemachine/fe/single_topology.py
proteneer: This will be removed - but a lot of files need to be cleaned up (a lot of logic on the MD/estimator needs will need to be modified). I will take care of it during deboggling week.

timemachine/fe/geometry.py
```diff
@@ -3,56 +3,106 @@
...
+def label_stereo(
+    num_atoms,
+    bond_idxs,
+    angle_idxs,
+    angle_params,
+    improper_idxs,
+    improper_params,
+):
+    # list of list representation
+    nblist = bond_idxs_to_nblist(num_atoms, bond_idxs)
```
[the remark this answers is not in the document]
proteneer: I'm okay with this change, to be clear:

neighborlist: spatial context under some cutoff 
adj_list: graph context under edge connectivity?

timemachine/fe/geometry.py
```diff
@@ -3,56 +3,106 @@
...
+
+    nblist = []
+    for i in range(num_atoms):
+        nbs = []
+        for j in range(num_atoms):
+            if cmat[i][j]:
+                nbs.append(j)
+        nblist.append(nbs)
+
+    return nblist
```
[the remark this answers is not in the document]
proteneer: nice! I'm okay with bringing in `networkx` but for internal algorithms for now. I do not want to networkx touching our APIs if possible (too many horror memories refactoring networkx nodes/edges objects that contain a billion keys).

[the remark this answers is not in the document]
proteneer: hm this almost works, but doesn't handle `num_atoms` - i.e. dummy atoms or disconnected atoms that are not connected.

timemachine/fe/geometry.py
```diff
@@ -3,56 +3,106 @@
...
+
+
+def label_stereo(
+    num_atoms,
+    bond_idxs,
+    angle_idxs,
+    angle_params,
+    improper_idxs,
+    improper_params,
+):
```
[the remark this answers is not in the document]
proteneer: willfix

timemachine/fe/geometry.py
```diff
@@ -3,56 +3,106 @@
...
+            atom_geometries.append(local_geometry)
+        elif len(atom_nbs) == 1:
+            atom_geometries.append(LocalGeometry.G1_TERMINAL)
+        elif len(atom_nbs) == 0:
+            atom_geometries.append(None)
+
+    return atom_geometries
+
+
+def classify_geometry(mol: Chem.Mol, ff: Forcefield = None, core: List[int] = None) -> List[LocalGeometry]:
```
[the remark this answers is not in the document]
proteneer: Dumb question - what's the rationale for adding `Optional` to something that already has a default parameter? isn't that a bit redundant?

timemachine/fe/geometry.py
```diff
@@ -3,56 +3,106 @@
...
     """
     Identify the local geometry of the molecule. This current uses a heuristic but we
```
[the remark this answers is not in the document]
proteneer: willfix

timemachine/fe/geometry.py
```diff
@@ -64,15 +114,69 @@ def classify_geometry(mol: Chem.Mol) -> List[LocalGeometry]:
...
+    proper_idxs = pt.get_idxs()
+    improper_idxs = it.get_idxs()
+
+    if core is None:
+        core = np.arange(mol.GetNumAtoms())
+
+    core_bond_idxs = []
+    core_bond_params = []
+    for ij, p in zip(bond_idxs, bond_params):
+        if np.all([a in core for a in ij]):
```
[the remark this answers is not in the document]
proteneer: agreed - willfix

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,783 @@
+from collections.abc import Iterable
+
+import numpy as np
+
+from timemachine.fe import dummy, geometry, topology
+from timemachine.fe.geometry import LocalGeometry
+
+
+def is_planarizing(force_constant, phase, period):
+    return period == 2 and (phase - np.pi) < 0.05 and force_constant > 10.0
```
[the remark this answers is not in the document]
proteneer: 1) yes - d'oh
2) 0.05 is in rads, and is pretty arbitrary (corresponds to about 3 degrees), currently our ff has these set to pi rads. Force constants are always in OpenMM units, here it's kJ/mol/rad^2
3) yes - but it's also a function of the environment as well (eg. 300K in vacuum is much easier to sample than 300K in solvent), so not sure how deep to go down this rabbit hole

timemachine/fe/geometry.py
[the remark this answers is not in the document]
maxentile: In cases where the default is `None`, the default doesn't type-check

timemachine/fe/single_topology.py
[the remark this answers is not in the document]
proteneer: PS what makes this even more difficult is that planarizing torsions can be redundant - so a lot of this is a placeholder heuristic that will be replaced later by something more robust.

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,783 @@
...
+                nbs_2.add((atom, dst))
+            elif dst == atom and src in core_idxs and src != anchor_idx:
+                nbs_2.add((atom, src))
+
+    return nbs_1, nbs_2
+
+
+def setup_orientational_restraints(ff, mol_a, mol_b, core, dg, anchor):
+    """
+    Add restraints between dummy atoms in a dummy_group and core atoms.
```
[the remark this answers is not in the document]
proteneer: willfix

timemachine/fe/geometry.py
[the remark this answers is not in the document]
mcwitt: >  what's the rationale for adding Optional to something that already has a default parameter? isn't that a bit redundant?

It's a bit confusing because "optional" has several meanings, e.g. an argument that need not be provided, or the type `Optional = Union[T, None]` (the latter is the case here).

Without `Optional` here, the default values don't have the correct type (`None` is not an allowed value of `List[int]`, for example) and mypy won't catch potential bugs in the implementation if e.g. we missed handling the case where `ff = None`.

I'm not sure why mypy with our configuration doesn't catch this, but it's flagged in my editor (using pyright):
image

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,783 @@
...
+
+    mol_a's parameters are mainly used to check for numerical stability.
+    """
+    core_b_to_a = dict()
+    for a, b in core:
+        core_b_to_a[b] = a
```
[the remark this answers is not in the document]
proteneer: willfix inline

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,783 @@
...
+        for src, dst in bond_idxs:
+            if src == atom and dst in core_idxs and dst != anchor_idx:
+                nbs_2.add((atom, dst))
+            elif dst == atom and src in core_idxs and src != anchor_idx:
+                nbs_2.add((atom, src))
+
+    return nbs_1, nbs_2
+
+
+def setup_orientational_restraints(ff, mol_a, mol_b, core, dg, anchor):
```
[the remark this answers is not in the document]
proteneer: sure - we use dg for something else anyways (though typically its capitalized dG)

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,783 @@
...
+
+    mol_a's parameters are mainly used to check for numerical stability.
+    """
+    core_b_to_a = dict()
+    for a, b in core:
+        core_b_to_a[b] = a
+
+    assert anchor in core_b_to_a
+    for d in dg:
+        assert d not in core_b_to_a
```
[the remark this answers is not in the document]
proteneer: will fix

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,783 @@
...
+    mol_b_bond_idxs = mol_b_hb.get_idxs()
+    mol_b_angle_idxs = mol_b_ha.get_idxs()
+    mol_b_proper_idxs = mol_b_pt.get_idxs()
+    mol_b_improper_idxs = mol_b_it.get_idxs()
+
+    mol_b_core = core_b_to_a.keys()
+    mol_b_full_geometry = geometry.classify_geometry(mol_b)
+    mol_b_core_geometry = geometry.classify_geometry(mol_b, core=mol_b_core)
+
+    dg = list(dg)
```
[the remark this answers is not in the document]
proteneer: hm sure - this is a set but no reason not to move it up

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,783 @@
...
+        self.core = core
+        self.ff = forcefield
+
+        assert core.shape[1] == 2
+
+        # map into idxs in the combined molecule
+        self.a_to_c = np.arange(mol_a.GetNumAtoms(), dtype=np.int32)  # identity
+        self.b_to_c = np.zeros(mol_b.GetNumAtoms(), dtype=np.int32) - 1
+
+        self.NC = mol_a.GetNumAtoms() + mol_b.GetNumAtoms() - len(core)
```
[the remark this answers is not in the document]
proteneer: number of atoms in the combined, alchemical hybrid.

timemachine/potentials/bonded.py
```diff
@@ -147,6 +148,95 @@ def harmonic_bond(conf, params, box, lamb, bond_idxs, lamb_mult=None, lamb_offse
...
+
+    return tb
+
+
+def harmonic_x_angle(conf, params, box, lamb, angle_idxs):
+    if len(angle_idxs) == 0:
+        return 0.0
```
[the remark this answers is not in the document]
proteneer: it's slightly cleaner this way than to have to calling code deal with empty lists - no big preference tho.

tests/test_geometry.py
proteneer: yes - i will add a note to redo this - but this will be a huge diff so I won't include it here.

tests/test_single_topology_stereo.py
```diff
@@ -0,0 +1,486 @@
...
+
+# test that we do not invert stereochemical barriers at the end-states for various susceptible transformations.
+# most of these tests proceed by measuring the chiral volume defined by 4 atoms and ensuring that they're consistent
+# at both end-states.
+
+
+def minimize_bfgs(x0, U_fn):
+    N = x0.shape[0]
+
+    def u_bfgs(x_flat):
```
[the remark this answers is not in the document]
proteneer: agreed - will fix, let's stick with the convention that U (kJ/mol), and u (kTs)

tests/test_single_topology_stereo.py
proteneer: i think scipy.minimize has become synonymous with BFGS at this point, but I agree that we should be consistent. I will rename to minimize_scipy just incase something better than BFGS comes along (I hope so!)

tests/test_single_topology_stereo.py
```diff
@@ -0,0 +1,486 @@
...
+
+    num_atoms = x0.shape[0]
+
+    x_min = minimize_bfgs(x0, U_fn)
+    num_workers = 1
+    num_batches = 2000
+    frames, _ = simulate(x_min, U_fn, 300.0, np.ones(num_atoms) * 4.0, 1000, num_batches, num_workers)
+    # (ytz): discard burn in later
+    # burn_in_batches = num_batches//10
+    burn_in_batches = 0
```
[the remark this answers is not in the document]
proteneer: in production code as part of a public API, yes, in testing code I'd rather just hard code. I toggle this between different values during debugging alot, but for testing purposes only the user really shouldn't have to be changing this.

tests/test_single_topology_stereo.py
proteneer: hm lemme dig a bit deeper

tests/test_single_topology_stereo.py
proteneer: hm I'm not sure the best way to functionalize this - the idxs that go into `measure_chiral_volume` can be 8 different idxs, not all tests permit pos and neg vols for a single atom. Was it the `if` statement itself that seemed verbose?

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,784 @@
...
+            res.append(recursive_map(item, mapping))
+        return tuple(res)
+    else:
+        return mapping[items]
+
+
+def find_stereo_bonds(ring_bonds, proper_idxs, proper_params, mol):
+    # a stereo bond is defined as a bond that
+    # 1) has a proper torsion term that has k > 30 kJ/mol, period=2, phase=3.1415
+    # 2) a bond that is not part of a ring system.
```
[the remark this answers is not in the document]
proteneer: willfix inline

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,784 @@
...
+            if src.GetDegree() == 4 or dst.GetDegree() == 4:
+                # if either atom is tetrahedral then it has no pi bonds to conjugate over
+                continue
+            else:
+                canonical_stereo_bonds.add(k)
+
+    return canonical_stereo_bonds
+
+
+def find_stereo_atoms(mol):
```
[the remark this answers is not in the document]
proteneer: modified to chirality, will add extra comment.

tests/test_single_topology_stereo.py
jkausrelay: just that the calculation of the fraction positive seemed a bit repetitive in the 3 examples here. Up to you if it doesn't seem easy to functionalize that's fine too.

timemachine/fe/single_topology.py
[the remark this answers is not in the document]
proteneer: Hm - I'm not a huge fan of modding over floats, simple things like `3.5 % 0.2 == 0.2` would return `False` due to round-off. I guess in this case since its a `<` sign it's safe but I'm not sure I'm comfortable in general with float mods.

timemachine/fe/single_topology.py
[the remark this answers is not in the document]
proteneer: willfix - but not sure on the best way to do it yet, may just stuff them onto lib.Potentials objects

timemachine/potentials/bonded.py
proteneer: refactored in https://github.com/proteneer/timemachine/pull/746/commits/07eeaa154e1bba54fc1c7a8512410470aef8709f

timemachine/fe/geometry.py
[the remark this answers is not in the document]
proteneer: will revisit this later

timemachine/fe/geometry.py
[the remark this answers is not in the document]
proteneer: fixed

timemachine/potentials/bonded.py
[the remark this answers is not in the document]
proteneer: (and technically, you can return any constant and still be correct here)

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,783 @@
...
+    # specialized restraints that are factorizable
+    restraint_cross_angle_idxs = []
+    restraint_cross_angle_params = []
+
+    restraint_centroid_angle_idxs = []
+    restraint_centroid_angle_params = []
+
+    nbs_1, _ = enumerate_anchor_groups(anchor, mol_b_bond_idxs, mol_b_core)
+
+    if anchor_core_geometry == LocalGeometry.G1_TERMINAL:
```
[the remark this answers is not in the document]
proteneer: I'm not sure how to restructure this without introducing more boilerplate, do I set up the interaction matrix explicitly? with nan or assertions if one of the invalid configurations is found? How do I carry over state and deal with list concatenation? I'd need *every* possible case to return *every* possible set of bond/angle/c_angle/x_angle/proper/improper lists to make it generic.

I'd originally tried to write it in a more modular way but it wasn't all that pretty - I'd be happy to attempt again.

timemachine/fe/single_topology.py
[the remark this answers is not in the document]
proteneer: fixed this inline with an assertion comment

timemachine/fe/geometry.py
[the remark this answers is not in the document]
proteneer: fixed

timemachine/potentials/bonded.py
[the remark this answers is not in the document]
mcwitt: ah ok -- I wasn't sure if this was an exceptional case that we'd want to know about (vs. silently returning an incorrect result)

timemachine/fe/geometry.py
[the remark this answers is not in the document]
proteneer: added

tests/test_single_topology_stereo.py
proteneer: fixed

timemachine/fe/single_topology.py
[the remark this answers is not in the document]
proteneer: Actually I'm going to just modify the `<0.05` case to throw an assertion. This should never happen.

timemachine/fe/single_topology.py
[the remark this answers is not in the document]
mcwitt: Got it; for a set it's probably fine as is (wasn't sure if this was here to force materialization of an iterator)

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,866 @@
...
+    for b in mol_b.GetBonds():
+        if b.IsInRing():
+            mol_b_ring_bonds.append((b.GetBeginAtomIdx(), b.GetEndAtomIdx()))
+
+    stereo_bonds = find_chiral_bonds(mol_b_ring_bonds, mol_b_proper_idxs, mol_b_proper_params, mol_b)
+    stereo_atoms = find_chiral_atoms(mol_b)
+
+    junction_bonds = find_junction_bonds(anchor, mol_b_bond_idxs)
+    # (ytz): the mapping code should hopefully be able to guarantee this
+    assert junction_bonds.intersection(stereo_bonds) == set()
```
mcwitt, line 315: nit:
```suggestion
    assert junction_bonds.isdisjoint(stereo_bonds)
```

timemachine/fe/single_topology.py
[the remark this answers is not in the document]
maxentile: >I'm not a huge fan of modding over floats, simple things like 3.5 % 0.2 == 0.2 would return False due to round-off. I guess in this case since its a < sign it's safe but I'm not sure I'm comfortable in general with float mods.

Huh! I didn't realize mod was an unsafe operation to use here.

If concerned about rounding errors in this comparison, may be preferable to use `np.pi` rather than the literal `3.1415`.

```suggestion
    eq_theta = canonical_angles[jkl][1]
    elif abs(eq_theta - 0) < 0.05:
        return False
    elif abs(eq_theta - np.pi) < 0.05:
        return False
```

>Actually I'm going to just modify the abs(canonical_angles[jkl][1] - 0) case to throw an assertion. This should never happen.

Good call.

(May also want a check that `eq_theta` is in interval [0, pi] -- should never happen, but just to rule out silently returning `True` for `3 * pi` but `False` for `1 * pi`...)

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,866 @@
...
+    mol_b_core = core_b_to_a.keys()
+    mol_b_full_geometry = geometry.classify_geometry(mol_b)
+    mol_b_core_geometry = geometry.classify_geometry(mol_b, core=mol_b_core)
+
+    # pick an arbitrary atom in the dummy_group and find the anchors, there may be
+    # multiple anchors, eg (d=dummy, c=core):
+    #   d...d
+    #   |   |
+    #   c---c
+    root_anchors = dummy.identify_root_anchors(mol_b_bond_idxs, mol_b_core, dummy_group[0])
```
mcwitt, line 300: should we `assert len(dummy_group) > 0`?

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,866 @@
...
+    assert anchor in root_anchors
+    # (ytz): we can relax this assertion later on.
+    assert len(root_anchors) == 1, "multiple root anchors found."
+
+    mol_b_ring_bonds = []
+    for b in mol_b.GetBonds():
+        if b.IsInRing():
+            mol_b_ring_bonds.append((b.GetBeginAtomIdx(), b.GetEndAtomIdx()))
```
mcwitt, line 308: could this reuse `get_romol_bonds`? I.e. something like
```suggestion
    mol_b_ring_bonds = [bond for bond in utils.get_romol_bonds(mol_b) if bond.IsInRing()]
```

timemachine/fe/single_topology.py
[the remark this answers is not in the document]
proteneer: this seems reasonable - will add.

timemachine/fe/single_topology.py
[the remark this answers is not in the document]
proteneer: Does something like the solution in https://github.com/proteneer/timemachine/pull/746/commits/6d68a980d21b1e191038cf7ea8142dc2455fe4fc work?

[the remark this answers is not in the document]
maxentile: Nice -- this looks like an improvement!

Mainly want to group together the idxs and params for each restraint type. Later it might be worthwhile to extract more of the validation logic into each restraint type's constructor / updater methods, or into the constructor of `VacuumSystem`, as appropriate.

tests/test_geometry.py
```diff
@@ -493,6 +493,58 @@ def test_baran_tetrazole():
     assert atom_types == expected_types
 
 
+def test_aniline():
+    # although aniline is slightly pyramidal, we'd rather it
+    # be assigned planar for safety.
```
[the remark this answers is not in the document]
proteneer: Centroid angles are generally only computed for a G2_KINK -> G3_PLANAR case. This comment is stale (and was for when we also used centroid angles for G3_PYRAMIDAL -> G4_TETRAHEDRAL, which is unstable since the NH3 can invert, and G3_PLANAR -> G4_TETRAHEDRAL used cross-product based restraints. Now, both (G3_PYRAMIDAL, G3_PLANAR) will always use a cross product restraint.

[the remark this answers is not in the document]
proteneer: (removing this comment)

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,954 @@
...
+    # near zero probability at bond lengths < 0.5 Angstroms
+    canonical_bonds = dict()
+    for idxs, params in zip(bond_idxs, bond_params):
+        canonical_bonds[dummy.canonicalize_bond(idxs)] = params
+
+    jk = dummy.canonicalize_bond((j, k))
+
+    if jk not in canonical_bonds:
+        return False
+    elif canonical_bonds[jk][0] < 50.0:
```
mcwitt, line 158: could we add a comment on where `50.0` comes from? (the 0.05 nm is mentioned in the comment above)

timemachine/fe/single_topology.py
```diff
@@ -1,13 +1,19 @@
...
-from timemachine.fe import dummy, geometry, topology
+from timemachine.fe import dummy, geometry, topology, utils
 from timemachine.fe.geometry import LocalGeometry
+from timemachine.lib import potentials
 
 
 def is_planarizing(force_constant, phase, period):
-    return period == 2 and (phase - np.pi) < 0.05 and force_constant > 10.0
+    """
+    Determine if
```
jkausrelay, line 14: update or remove

timemachine/fe/single_topology.py
```diff
@@ -18,11 +24,9 @@ def identify_bonds_spanned_by_planar_torsions(proper_idxs, proper_params):
     planar_bonds = dict()
```
jkausrelay, line 24: nit: `planar_bonds = defaultdict(list)`

and clean up below

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,954 @@
...
+            jbs.add(dummy.canonicalize_bond((i, j)))
+    return jbs
+
+
+def enumerate_anchor_groups(anchor_idx, bond_idxs, core_idxs):
+
+    # enumerate all 1 and 2 neighbor anchor atoms to form valid anchor groups.
+
+    assert anchor_idx in core_idxs
+    assert anchor_idx in [x[0] for x in bond_idxs] or anchor_idx in [x[1] for x in bond_idxs]
```
mcwitt, line 221: nit: maybe slightly cleaner
```suggestion
    assert any(anchor_idx == atom_idx for bond in bond_idxs for atom_idx in bond)
```

timemachine/fe/single_topology.py
```diff
@@ -55,12 +59,39 @@ def recursive_map(items, mapping):
...
+
+    """
+    # a chiral bond is defined as a bond that
     # 1) has a single proper torsion term with k > 10 kJ/mol, period=2, phase=3.1415
-    # 2) a bond that is not part of a ring system.
+    # 2) is not part of a ring system.
     # the reason why 2) is present is because the planar torsions spanning a
-    # ring system are not used to enforce stereochemistry, since if we simply
-    # disabled them, we would *still* get the correct stereochemistry due to steric effects.
+    # ring system are not used to enforce chiralchemistry, since if we simply
```
jkausrelay, line 93: nit: chiralchemistry -> chiral chemistry

timemachine/fe/single_topology.py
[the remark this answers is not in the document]
mcwitt: I think this didn't get resolved yet?

timemachine/fe/single_topology.py
proteneer: This came from an overly zealous stereo->chiral replacement. I'm trying to avoid the use of the words stereo/stereochemistry due to confusion with CIP, so I will rename this to "chirality".

timemachine/fe/single_topology.py
```diff
@@ -193,19 +238,41 @@ def enumerate_anchor_groups(anchor_idx, bond_idxs, core_idxs):
...
+    core: np.array(N,2)
+        Core atoms
+
+    dummy_group: set
+        A set of dummy atoms tethered to an anchor
+
+    Returns
+    -------
+    (idxs, params)
+        A set of bond, angle, proper, improper, c_angle, and x_angle restraints
```
jkausrelay, line 273: return is now `VacuumSystem`

timemachine/fe/single_topology.py
proteneer: added comment for both check_bond and  check_angle

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,962 @@
...
+    #   c---c
+    root_anchors = dummy.identify_root_anchors(mol_b_bond_idxs, mol_b_core, dummy_group[0])
+    assert anchor in root_anchors
+    # (ytz): we can relax this assertion later on.
+    assert len(root_anchors) == 1, "multiple root anchors found."
+
+    mol_b_ring_bonds = []
+    for b in mol_b.GetBonds():
+        if b.IsInRing():
+            mol_b_ring_bonds.append((b.GetBeginAtomIdx(), b.GetEndAtomIdx()))
```
mcwitt, line 322: nit: could be list comprehension?
```suggestion
    mol_b_ring_bonds = [(b.GetBeginAtomIdx(), b.GetEndAtomIdx()) for b in mol_b.GetBonds() if b.IsInRing()]
```

timemachine/fe/topology.py
```diff
@@ -292,6 +292,39 @@ def parameterize_periodic_torsion(self, proper_params, improper_params):
...
 
+    def parameterize_nonbonded_pairlist(self, ff_q_params, ff_lj_params):
+        # generate intramolecular nonbonded pairlist
+        # use same scale factors for electrostatics and vdWs
+        exclusion_idxs, scale_factors = nonbonded.generate_exclusion_idxs(
+            self.mol, scale12=_SCALE_12, scale13=_SCALE_13, scale14=_SCALE_14
+        )
+
+        # note: use same sf for electrostatics and vdw
+        # tpyically in protein ffs, gaff, the 1-4 ixns use different scale factors between vdw and electrostatics
```
jkausrelay, line 303: typo: typically

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,962 @@
...
+    # copy restraints that involve only anchor and dummy atoms. These restraints are always
+    # numerically stable and factorizable. In order to enhance sampling, dummy atoms are designed
+    # to sort of mimic the "enhanced" state. Where by:
+    # 1) bond, angle, and improper torsions are copied over.
+    # 2) chiral bonds are copied over, but generic rotatable bonds are not.
+    # 3) dummy-dummy nonbonded interactions are fully disabled.
+    # tbd: achiral dummy atoms that are G3_PYRAMIDAL should be planarized?
+    dga = dummy_group + [anchor]
+    for idxs, params in zip(mol_b_bond_idxs, mol_b_bond_params):
+        if np.all([a in dga for a in idxs]):
```
mcwitt, line 353: nit: stdlib `all`

timemachine/fe/single_topology.py
proteneer: willfix inline

timemachine/fe/single_topology.py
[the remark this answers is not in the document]
mcwitt: I'm ok with saving this for a later cleanup

timemachine/fe/single_topology.py
proteneer: willfix

timemachine/fe/single_topology.py
proteneer: maybe

mcwitt: minor, no need to change!

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,953 @@
...
+                restraint_cross_angle_idxs.append(((j, a), (j, b), (j, k)))
+                restraint_cross_angle_params.append(1000.0)
+            else:
+                # planarize so we can enhance sample both chiral states using a centroid
+                # type a
+                #       \
+                #      c.j.k <- angle (c,j,k) = 0.0
+                #       /
+                #      b
+                restraint_centroid_angle_idxs.append((tuple(sorted((a, b))), j, k))
```
mcwitt, line 517: could we use `canonicalize_bond` here?

proteneer: `(a,b)` aren't actually bonded to each other. Rather they're bonded to the junction atom `j`. The `sorted()` is purely a convenience for testing purposes. So I didn't want to confuse the reader into thinking they were actually bonds.

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,953 @@
...
+    mol_a: Chem.Mol
+        mol_a's parameters are mainly used to check for numerical stability.
+
+    mol_b: Chem.Mol
+        Source of the dummy atoms
+
+    core: np.array(N,2)
+        Core atoms
+
+    dummy_group: set
```
mcwitt, line 268: should this be a list instead? (`dummy_group[0]` below fails if this is a set)

timemachine/fe/single_topology.py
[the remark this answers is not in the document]
proteneer: resolved in https://github.com/proteneer/timemachine/pull/746/commits/db713f8f71ecc70ac14969e8c6221b0a7cd19b8f

tests/test_single_topology_v2.py
```diff
@@ -0,0 +1,314 @@
...
+import numpy as np
+import pytest
+from rdkit import Chem
+
+from timemachine.fe import single_topology, utils
+from timemachine.fe.single_topology import setup_dummy_interactions
+from timemachine.ff import Forcefield
+
+
+def test_find_chiral_atoms():
```
maxentile, line 12: Missing test: `test_find_chiral_bonds`

Would want to add:
1. minimal examples like "F/C=C/F" and "F\C=C/F" -- assert chiral bonds ==  `{(1, 2)}` (currently passing)

2. out-there examples like `"C([H])(C)=C=C([H])(C)"` -- an example of "chirality without stereocenters" from https://www.colby.edu/chemistry/CH241F/Chapter%204.pdf (currently, `single_topology.find_chiral_atoms` says the 2 methyls are chiral but not the other 3 double-bonded carbons, and `single_topology.find_chiral_bonds` returns the empty set, which seems incorrect? seems like an error should be raised on an input like this)

3. more cases of specifications on rings, e.g. cis vs. trans 1,2-dichlorocyclohexane ("C1CC[C@@H]([C@@H](C1)Cl)Cl" and "C1CC[C@H]([C@@H](C1)Cl)Cl"). find_chiral_bonds returns empty set for both, and find_chiral_atoms returns all ring atoms for both... This seemed counterintuitive at first, but on further thought is fine! (Atom-centered approach doesn't exactly specify the concepts "cis" vs. "trans" here, but specifies a specific cis configuration and a specific trans configuration in "absolute" coordinates...)

timemachine/potentials/bonded.py
```diff
@@ -147,6 +148,88 @@ def harmonic_bond(conf, params, box, lamb, bond_idxs, lamb_mult=None, lamb_offse
...
+
+    ci = jnp.array(ci)
+    cj = jnp.array(cj)
+    ck = jnp.array(ck)
+
+    # compute cosine angle of three points
+    vij = ci - cj
+    vjk = ck - cj
+    top = jnp.sum(jnp.multiply(vij, vjk), -1)
+    bot = jnp.linalg.norm(vij, axis=-1) * jnp.linalg.norm(vjk, axis=-1)
```
maxentile, line 161: Naming nit: subscript "ij" on L158 uses convention "first minus second", subscript "jk" on L159 uses convention "second minus first" -- prefer to standardize one way or the other within the definition

```suggestion
    # compute cosine angle of three points
    vij = ci - cj
    vkj = ck - cj
    top = jnp.sum(jnp.multiply(vij, vkj), -1)
    bot = jnp.linalg.norm(vij, axis=-1) * jnp.linalg.norm(vkj, axis=-1)
```

timemachine/potentials/bonded.py
```diff
@@ -147,6 +148,88 @@ def harmonic_bond(conf, params, box, lamb, bond_idxs, lamb_mult=None, lamb_offse
...
+
+    x_is = []
+    x_js = []
+    x_ks = []
+
+    for idxs, j, k in angle_idxs:
+        x_j = conf[j]
+        # compute v-site from unit vectors to avoid bond length
+        # change differences
+        a, b = idxs
```
maxentile, line 178: Makes it slightly easier to skim implementation to remind what format to provide `angle_idxs`

```suggestion
    for (a, b), j, k in angle_idxs:
        x_j = conf[j]
        # compute v-site from unit vectors to avoid bond length
        # change differences
```

timemachine/potentials/bonded.py
```diff
@@ -147,6 +148,88 @@ def harmonic_bond(conf, params, box, lamb, bond_idxs, lamb_mult=None, lamb_offse
...
+        v_ja = x_a - x_j
+        v_jb = x_b - x_j
+        x_is.append(x_j + jnp.cross(v_ja, v_jb))
+        x_js.append(x_j)
+        x_ks.append(x_c)
+
+    cos_angles = get_cos_angle(x_is, x_js, x_ks)
+    kas = params
+    cos_2_angles = 2 * cos_angles ** 2 - 1  # double angle - symmetrized to both ends
+    energies = kas / 2 * jnp.power(cos_2_angles - 1, 2)
```
maxentile, line 215: Style nit -- extra parentheses would be marginally nicer to read
```suggestion
    energies = (kas / 2) * jnp.power(cos_2_angles - 1, 2)
```
or
```suggestion
    energies = 0.5 * kas * jnp.power(cos_2_angles - 1, 2)
```
or similar...

Tangent: is there a reason to prefer `jnp.power(x, 2)` over `x**2` in this line?
```suggestion
    energies = 0.5 * kas * (cos_2_angles - 1)**2
```

timemachine/potentials/nonbonded.py
```diff
@@ -300,6 +305,13 @@ def apply_cutoff(x):
     qij = apply_cutoff(charges[inds_l] * charges[inds_r])
     electrostatics = direct_space_pme(dij, qij, beta)
 
+    if rescale_mask is not None:
+        assert len(rescale_mask) == len(pairs)
```
maxentile, line 309: nit: strengthen shape assertion
```suggestion
        assert rescale_mask.shape == (len(pairs), 2)
```

timemachine/potentials/nonbonded.py
```diff
@@ -274,6 +276,9 @@ def nonbonded_v3_on_specific_pairs(conf, params, box, pairs, beta: float, cutoff
         can be returned.
     """
 
+    if len(pairs) == 0:
+        return 0.0, 0.0
```
maxentile, line 280: nit: makes return type unstable (array, array) or (float, float) -- would it be better to return zero arrays in this case instead?
```suggestion
        return np.zeros(1), np.zeros(1)
```
Or refactor call site not to call this function when `len(pairs) == 0`?

tests/test_single_topology_stereo.py
```diff
@@ -0,0 +1,450 @@
...
+  3  2  0  0  0  0            999 V2000
+    0.5038    0.9940    0.3645 O   0  0  0  0  0  0  0  0  0  0  0  0
+    0.4342    0.1988    0.9333 H   0  0  0  0  0  0  0  0  0  0  0  0
+    0.5015    0.5419   -0.8126 H   0  0  0  0  0  0  0  0  0  0  0  0
+  1  3  1  0  0  0  0
+  1  2  1  0  0  0  0
+M  END
+$$$$""",
+        removeHs=False,
+    )
```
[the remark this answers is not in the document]
proteneer: > I guess explicitness about the conformer and atom indexing?

Is exactly why. The initial method used `EmbedMolecule()` but the atom-mapping became ambiguous when mapping with stereochemistry concerns. I generate the conformers here explicitly by starting with `mol_a`, then glueing atoms of `mol_b` unto it in a way that guarantees the initial chirality to be that of the intended sign.

tests/test_single_topology_stereo.py
```diff
@@ -0,0 +1,450 @@
...
+    # compute vectors
+    v0 = x1 - x0
+    v1 = x2 - x0
+    v2 = x3 - x0
+
+    v0 = v0 / np.linalg.norm(v0)
+    v1 = v1 / np.linalg.norm(v1)
+    v2 = v2 / np.linalg.norm(v2)
+
+    return np.dot(np.cross(v0, v1), v2)
```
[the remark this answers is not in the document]
proteneer: I'm a little reluctant to use a geometric expression here involving the determinant (as opposed to just a triple product, which is mathematically simpler). Does this generalize to higher dimensions?

tests/test_single_topology_stereo.py
```diff
@@ -0,0 +1,450 @@
...
+
+    """
+    # compute vectors
+    v0 = x1 - x0
+    v1 = x2 - x0
+    v2 = x3 - x0
+
+    v0 = v0 / np.linalg.norm(v0)
+    v1 = v1 / np.linalg.norm(v1)
+    v2 = v2 / np.linalg.norm(v2)
```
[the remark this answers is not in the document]
proteneer: willfix renaming - but probably just `x_0 -> x_c` for brevity.

this is a utility function initially - wasn't sure if I wanted to make this more generally available yet.

tests/test_single_topology_stereo.py
```diff
@@ -0,0 +1,450 @@
...
+        vol_br = measure_chiral_volume(f[0], f[1], f[2], f[4])
+        assert vol_cl > 0 and vol_br < 0
+
+    system = s_top.generate_end_state_mol_b()
+    frames = simulate_system(system, x0)
+
+    for f in frames:
+        vol_cl = measure_chiral_volume(f[0], f[1], f[2], f[3])
+        vol_br = measure_chiral_volume(f[0], f[1], f[2], f[4])
+        assert vol_cl > 0 and vol_br < 0
```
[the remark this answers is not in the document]
proteneer: ah okay - this is probably simpler.

tests/test_single_topology_v2.py
proteneer: 1) will take a look
2) it depends on if this is a forcefield issue or not - our FF currently reqiures torsions be spanned over contiguous atoms with bond/angles adjacent. I need to check and see if there's even a torsion here spanning either of the `C=C` bonds (which would be undefined anyways). One possible way to deal with this is to define an extended torsion skipping the middle "C"
3) I can add more examples for some heterocycles (aromatic and non-aromatic ones)

timemachine/potentials/bonded.py
proteneer: sgtm!

timemachine/potentials/nonbonded.py
proteneer: good call on the typing

timemachine/potentials/bonded.py
proteneer: I think `jnp.power(x,2) and x**2` are the same - I go back and forth between them. Do we have a preference?

timemachine/potentials/bonded.py
proteneer: I may modify this code a little bit so its more easily vectorizable in a future pass. `jax.jit` really doesn't like for loops.

timemachine/potentials/bonded.py
maxentile: No preference -- just wanted to double-check if these are interchangeable or if `jnp.power` was needed

tests/test_single_topology_v2.py
maxentile: 2.

>I need to check and see if there's even a torsion here spanning either of the C=C bonds (which would be undefined anyways).

There are torsion terms spanning these, but they're assigned force constants of 0 as you would expect!

image

(via the type with this pattern `'[*:1]~[*:2]=[#6,#7,#16,#15;X2:3]=[*:4]'`)

timemachine/potentials/bonded.py
```diff
@@ -147,6 +148,87 @@ def harmonic_bond(conf, params, box, lamb, bond_idxs, lamb_mult=None, lamb_offse
...
+
+def harmonic_x_angle(conf, params, box, lamb, angle_idxs):
+    if len(angle_idxs) == 0:
+        return 0.0
+
+    params = jnp.array(params)
+    x_is = []
+    x_js = []
+    x_ks = []
+    for (j, a), (j, b), (j, c) in angle_idxs:
```
badisa, line 200: Find it a bit odd that you reuse the variable `j` three times here. Presumably always the same value, but might warrant either an assertion or a comment.

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,954 @@
...
+    mapping = np.arange(100)[::-1]
+    items = [[0,2,3], [5,1,[2,5,6]], 3]
+    result = recursive_map(items, mapping)
+
+    """
+    if isinstance(items, Iterable):
+        res = []
+        for item in items:
+            res.append(recursive_map(item, mapping))
+        return tuple(res)
```
badisa, line 58: nit: looks like this is a list of tuples, not list of lists

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,954 @@
...
+    jkl = dummy.canonicalize_bond((j, k, l))
+
+    # similar to bonds, these hopefully will prevent angles from linearizing
+    # eg. in the case of nitrile groups or from spinning arbitrarily. these
+    # angle terms are typically used to define cross-products.
+    if jkl not in canonical_angles:
+        return False
+    elif canonical_angles[jkl][0] < 50.0:
+        return False
+    elif abs(canonical_angles[jkl][1] - 3.1415) < 0.05:
```
badisa, line 192: Should we use `np.pi` here, or is there a reason to prefer 3.1415?

timemachine/fe/geometry.py
```diff
@@ -1,78 +1,222 @@
...
+    G3_PYRAMIDAL = 5  # R-X(-H)H
+    G4_TETRAHEDRAL = 6  # R-X(-H)(-H)H
 
 
-def assign_atom_geometry(atom):
+def bond_idxs_to_adj_list(num_atoms, bond_idxs):
     """
-    Heuristic using hybridization information to assign local description
-    of geometry.
+    Convert tuples of bond idxs to a neighbor list encoded as
```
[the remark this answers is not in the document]
proteneer: Willfix

timemachine/fe/single_topology.py
proteneer: willfix to np.pi

tests/test_single_topology_stereo.py
```diff
@@ -0,0 +1,450 @@
...
+
+
+def test_ammonium_to_chloromethyl():
+    # NH3 easily interconverts between the two chiral states. In the event that we
+    # morph NH3 to something that is actually chiral, we should still be able to
+    # ensure enantiopurity of the end-states.
+    # we expect the a-state to be:
+    #   mixed stereo on vol_a, fixed stereo on vol_b
+    # we expect the b-state to be:
+    #   fixed stereo on vol_a, fixed stereo on vol_b
```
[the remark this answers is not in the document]
proteneer: turns out this is quite tricky for cases where we need to enforce stereochemistry in the absence of *any* dummy atoms. Eg, simply morphing Achiral `N(F)(Cl)Br` to a Chiral `S(F)(Cl)Br`. I need to extend some of these checks to the core.

tests/test_single_topology_stereo.py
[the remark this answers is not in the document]
proteneer: good call, willfix

timemachine/fe/single_topology.py
```diff
@@ -0,0 +1,954 @@
...
+        nbs = a.GetNeighbors()
+        if len(nbs) == 4:
+            chiral_atoms.add(a_idx)
+        elif len(nbs) == 3 and mol_geom[a_idx] == LocalGeometry.G3_PYRAMIDAL:
+            # Trivalent Sulfur/Phosphorus typically are not invertible, nor
+            # are they properly parameterized. So we assert if we detect this case.
+            if a.GetAtomicNum() == 16 or a.GetAtomicNum() == 15:
+                assert 0, "Pyramidal Phosphorous or Sulfur detected."
+        elif len(nbs) > 4:
+            assert 0, "More than four neighbors found"
```
maxentile, line 146: Aside from the assertions, this could also be implemented with an atomic primitive smarts pattern, if desired

```suggestion
    query_mol = Chem.MolFromSmarts('[X4:1]')
    chiral_atoms = set(match[0] for match in mol.GetSubstructMatches(query_mol))
```

May be more appropriate to move these assertions into one of the TorsionHandlers or similar, if it's mostly meant to catch chemistries that are not well modeled by the forcefield.

proteneer: agreed! I think SMARTS for these probably makes sense.

timemachine/fe/single_topology.py
proteneer: will fix the docstring

proteneer:
Closing this in favor of #756

PR #747, opened 2022-05-24
Adds gpuErrchk to syncs that were missing it

badisa:
Ran into a error at a latter point due to not checking the error, this fixes it in all the places that it was missing.


proteneer (approved):
lgtm!

PR #748, opened 2022-05-26
Add setters for beta and cutoff

maxentile:
Also reorder so that completeness is more obvious at a glance


maxentile:
(Closing for now since stale and can be already accomplished by `Nonbonded.args[4] = beta`)

Issue #622, opened 2022-02-08
Option to freeze `d_volume_scale` in barostat?


maxentile:
Possible problem:

The Monte Carlo barostat contains some history-dependent state, used for on-the-fly adaptation of the box-scaling proposal (see `num_attempted`, `num_accepted`, `d_volume_scale` in [production implementation](https://github.com/proteneer/timemachine/blob/61e068297ce26083b1168535ba908f9987696605/timemachine/cpp/src/barostat.cu#L246-L265). see also `n_proposed`, `acceptance_fraction`, `max_delta_volume` in [python implementation](https://github.com/proteneer/timemachine/blob/61e068297ce26083b1168535ba908f9987696605/timemachine/md/barostat/moves.py#L149-L180)).

This could potentially introduce a bias when used [as an approximate MCMC move](https://github.com/proteneer/timemachine/blob/61e068297ce26083b1168535ba908f9987696605/timemachine/md/moves.py#L72). For example, we might mistakenly assume that `xs_updated = [npt.move(x) for x in xs]` updates each of a list of states independently, but in fact `x[i]` influences which value of `d_volume_scale`, etc. will be used to update `x[i>1]`...

How much bias?
* Unknown -- would guess very small. (Initially considering this a low-priority "cleaning" item unless this starts biting us somewhere.)

-----------

Possible solutions:
* Reset `num_attempted=0`, `num_accepted=0`, `d_volume_scale=some_default` every time an NPT move is used. Possible complication: While this avoids interaction between different calls to npt.move,  each move is individually made up of more non-markovian steps (since more likely to adjust `d_volume_scale` on-the-fly during the move, since `num_accepted`, `num_attempted` counters are low and `d_volume_scale` is not pre-adapted).
* Add an `adapt_proposal_scale` boolean option to the barostat. (Set `true` to obtain a good system-dependent `d_volume_scale`, then set `false` to freeze the value of `d_volume_scale` before doing anything like ` [npt.move(x) for x in xs]`.) This can probably be emulated without touching the logic, by setting the counters `num_attempted = some_giant_number`, `num_accepted = some_giant_number / 2`, where `some_giant_number` is chosen large enough that `num_accepted / num_attempted` is guaranteed to be in the range [0.25, 0.75] during the lifetime of the move even if 0% or 100% of moves are accepted)

jchodera:
[...]

PR #750, opened 2022-06-01
Add methods to set RNG seeds in custom_ops

mcwitt:
Adds `set_seed(const int)` member functions to
- LangevinIntegrator
- MonteCarloBarostat

The motivation is to allow for finer-grained reproducibility, and to reduce the friction of making random state explicit as in the `jax.random` API.


mcwitt:
Closing this for now after realizing that `MonteCarloBarostat` maintains internal state in addition to the PRNG state (see #622), so the value of being able to reset the PRNG state is unclear at the moment.

PR #751, opened 2022-06-07
Add functions to use SMC to compute dG_parameters in solvent/vacuum legs

jkausrelay:
- Adds absolute_hydration.set_up_ahfe_system_for_smc_parameter_changes which
    can be used during refitting to generate new samples under a new set of
    charge parameters.

- Adds moves.NVTMove so NVT simulations can be run with SMC.

 - Adds topology.RelativeFreeEnergyForcefield which is used to represent
    a molecule under different parameters (controlled by lambda).

### DIFF
--- a/tests/test_absolute_hydration.py
+++ b/tests/test_absolute_hydration.py
@@ -1,0 +1,166 @@
+import functools
+from typing import Tuple
+
+import numpy as np
+import pymbar
+import pytest
+
+from timemachine import testsystems
+from timemachine.constants import BOLTZ, DEFAULT_FF
+from timemachine.fe import absolute_hydration
+from timemachine.fe.functional import construct_differentiable_interface_fast
+from timemachine.fe.reweighting import one_sided_exp
+from timemachine.ff import Forcefield
+from timemachine.md import enhanced, moves, smc
+from timemachine.md.states import CoordsVelBox
+
+
+def get_ff_am1ccc():
+    return Forcefield.load_from_file(DEFAULT_FF)
+
+
+def test_smc_parameter_change_vacuum():
+    """
+    Tests correctness of using SMC to switch between parameters.
+    The SMC predicted dg ff0 -> ff1 should match results from
+    BAR using independent sampling.
+
+    """
+    seed = 2022
+    np.random.seed(seed)
+
+    # Reduce these so the test is faster
+    num_batches = 10
+    n_md_steps = 10
+    temperature = 300.0
+    kBT = BOLTZ * temperature
+
+    mol, _ = testsystems.ligands.get_biphenyl()
+    ff0 = get_ff_am1ccc()
+
+    # shift the ff parameters
+    # should be small enough so that 1-EXP is still valid
+    ff1 = get_ff_am1ccc()
+    ff1.q_handle.params += 0.2
+    box = np.eye(3) * 1000.0
+
+    # generate vacuum samples for each ff
+    def generate_samples(mol, ff) -> Tuple[enhanced.VacuumState, smc.Samples]:
+        state = enhanced.VacuumState(mol, ff)
+        _vacuum_xv_samples, vacuum_log_weights = enhanced.generate_log_weighted_samples(
+            mol,
+            temperature,
+            state.U_easy,
+            state.U_full,
+            seed,
+            num_batches=num_batches,
+            num_workers=1,
+            steps_per_batch=n_md_steps,
+        )
+
+        # discard velocities: (x, v) -> x
+        vacuum_samples = _vacuum_xv_samples[:, 0, :]
+        vacuum_vels = _vacuum_xv_samples[:, 1, :]
+        idxs = smc.multinomial_resample(vacuum_log_weights)[0]
+        vacuum_samples = [CoordsVelBox(coords=vacuum_samples[i], velocities=vacuum_vels[i], box=box) for i in idxs]
+        return state, vacuum_samples
+
+    state0, vacuum_samples0 = generate_samples(mol, ff0)
+    state1, vacuum_samples1 = generate_samples(mol, ff1)
+
+    # set up a system for sampling the new parameters using SMC
+    samples, lambdas, propagate, log_prob, resample = absolute_hydration.set_up_ahfe_system_for_smc_parameter_changes(
+        mol,
+        n_walkers=num_batches,
+        n_md_steps=n_md_steps,
+        resample_thresh=0.6,
+        initial_samples=vacuum_samples0,
+        ff0=ff0,
+        ff1=ff1,
+        is_vacuum=True,
+        n_windows=10,
+    )
+
+    smc_result = smc.sequential_monte_carlo(samples, lambdas, propagate, log_prob, resample)
+    smc_dg = one_sided_exp(-smc_result["log_weights_traj"][-1])
+
+    # BAR from independent runs
+    def batched_u_fxn(samples, state):
+        # Only compute vacuum energies
+        return np.array([state.U_full(x.coords) for x in samples]) / kBT
+
+    u_fwd = batched_u_fxn(vacuum_samples0, state1) - batched_u_fxn(vacuum_samples0, state0)
+    u_rev = batched_u_fxn(vacuum_samples1, state0) - batched_u_fxn(vacuum_samples1, state1)
+    bar_dg = pymbar.BAR(u_fwd, u_rev, compute_uncertainty=False)
+    assert smc_dg == pytest.approx(bar_dg, abs=1e-1)
+
+
+def test_smc_parameter_change_solvent():
+    """
+    Tests correctness of using SMC to switch between parameters.
+    The SMC predicted dg ff0 -> ff1 should match results from
+    BAR using independent sampling.
+
+    """
+    seed = 2022
+    np.random.seed(seed)
+
+    # Reduce these so the test is faster
+    num_batches = 20
+    n_md_steps = 100
+    temperature = 300.0
+    pressure = 1.0
+    n_eq_steps = 1000
+    kBT = BOLTZ * temperature
+
+    mol, _ = testsystems.ligands.get_biphenyl()
+    ff0 = get_ff_am1ccc()
+
+    # shift the ff parameters
+    # should be small enough so that 1-EXP is still valid
+    ff1 = get_ff_am1ccc()
+    ff1.q_handle.params += 0.2
+
+    def generate_samples(mol, ff):  # -> Tuple[Ene_fxn, smc.Samples]:
+        ubps, params, masses, coords, box = enhanced.get_solvent_phase_system(mol, ff, minimize_energy=True)
+        xvb0 = enhanced.equilibrate_solvent_phase(
+            ubps, params, masses, coords, box, temperature, pressure, n_eq_steps, seed
+        )
+        lam = 0.0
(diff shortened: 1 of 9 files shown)


timemachine/md/moves.py
```diff
@@ -71,56 +71,77 @@ def n_proposed(self):
...
+        after_steps = CoordsVelBox(x_t, v_t, box)
 
         self.n_proposed += 1
         self.n_accepted += 1
 
-        return after_npt
+        return after_steps
+
+
+class NPTMove(NVTMove):
```
maxentile, line 114: Nit: This inheritance sounds odd (although it does avoid some duplication between these two moves)...

Would it make sense to add a third class that both NPTMove and NVTMove inherit from?

jkausrelay: Hmm, the base class would be trivial (just rename NVTMove -> EnsembleBaseMove). Not sure if that would help.

maxentile: Hmm, this is probably a non-issue then.

(Looked odd at first since "NPT" is not a special case of "NVT", but makes sense functionally.)

jkausrelay: Added comment:

    Functionally, NPT is implemented as NVTMove plus a MC Barostat.
    So inherit from NVTMove here.

tests/test_enhanced.py
```diff
@@ -8,3 +19,83 @@ def test_identify_rotatable_bonds_runs_on_freesolv():
...
+
+@pytest.fixture
+def ff_pair():
+    ff0 = Forcefield.load_from_file(DEFAULT_FF)
+    ff1 = Forcefield.load_from_file(DEFAULT_FF)
+
+    # Modify the charge parameters for ff1
+    for h, p in zip(ff1.get_ordered_handles(), ff1.get_ordered_params()):
+        if isinstance(h, AM1CCCHandler):
+            p += 1.0
```
maxentile, line 32: Nit: to avoid second-guessing whether modifying the arrays returned by `ff1.get_ordered_params()` in-place also modifies `ff1` in-place

```suggestion
    ff1.q_handle.params += 1.0
```

timemachine/fe/topology.py
```diff
@@ -319,6 +319,74 @@ def parameterize_nonbonded(self, ff_q_params, ff_lj_params):
...
+        src_qlj_params, nb_potential = super().parameterize_nonbonded(ff_q_params[0], ff_lj_params[0])
+        dst_qlj_params, _ = self.bt1.parameterize_nonbonded(ff_q_params[1], ff_lj_params[1])
+
+        combined_qlj_params = jnp.concatenate([src_qlj_params, dst_qlj_params])
+        lambda_plane_idxs = np.zeros(self.mol.GetNumAtoms(), dtype=np.int32)
+        lambda_offset_idxs = np.zeros(self.mol.GetNumAtoms(), dtype=np.int32)
+
+        interpolated_potential = nb_potential.interpolate()
+        interpolated_potential.set_lambda_plane_idxs(lambda_plane_idxs)
+        interpolated_potential.set_lambda_offset_idxs(lambda_offset_idxs)
```
maxentile, line 364: Not 100% sure about this step -- is there a situation where we would want to simultaneously control the decoupling _and_ the parameter interpolation?

`u(x=x, params=interpolate(params_0, params_1, 0.33), lam=0.25)`

(Maybe not: probably suffices to interpolate params at decoupling distance = 0, and construct two classes corresponding to coupled vs. decoupled...)

timemachine/ff/__init__.py
```diff
@@ -95,3 +95,7 @@ def get_ordered_handles(self):
             Return a flat, pre-determined ordering of the handlers
         """
         return [self.hb_handle, self.ha_handle, self.pt_handle, self.it_handle, self.q_handle, self.lj_handle]
+
+
+def combine_ordered_params(ff0: Forcefield, ff1: Forcefield):
+    return list(zip(ff0.get_ordered_params(), ff1.get_ordered_params()))
```
maxentile, line 101: Python clarification: would `combined = combine_ordered_params(ff0, ff1); do_some_modification_in_place(combined)` also modify `ff0` and `ff1` in-place?

timemachine/md/enhanced.py
```diff
@@ -470,6 +471,75 @@ def get_solvent_phase_system(mol, ff, box_width=3.0, margin=0.5, minimize_energy
...
+
+    # concatenate (optionally minimized) water_coords and ligand_coords
+    ligand_coords = get_romol_conf(mol)
+    if minimize_energy:
+        new_water_coords = minimizer.minimize_host_4d([mol], water_system, water_coords, ff0, water_box)
+        coords = np.concatenate([new_water_coords, ligand_coords])
+    else:
+        coords = np.concatenate([water_coords, ligand_coords])
+
+    return potentials, params, np.array(masses), coords, water_box
```
maxentile, line 517: Possible future to-do: Since multiple functions now return `(potentials, params, masses, coords, Optional[box])` tuples, may want to put these in some kind of `System` container...

timemachine/fe/topology.py
```diff
@@ -319,6 +319,74 @@ def parameterize_nonbonded(self, ff_q_params, ff_lj_params):
...
+    corresponding to forcefield0 and forcefield1.
+    """
+
+    def __init__(self, mol, forcefield0, forcefield1):
+        """
+        Utility for working with a single ligand.
+
+        Parameter
+        ---------
+        mol: ROMol
```
maxentile, line 337: Nit: Annotations are sometimes `mol: Chem.Mol` and sometimes `mol: ROMol` -- is there a distinction?

timemachine/md/enhanced.py
maxentile, line 517: Observation: Introduces near-duplication of https://github.com/proteneer/timemachine/blob/98b4d4b0112c7bf23bd5562da163567a2021373f/timemachine/md/enhanced.py#L432-L471 , except for `topology.BaseTopology(mol, ff)` -> `topology.RelativeFreeEnergyForcefield(mol, ff0, ff1)` and `ff.get_ordered_params()` -> `combine_ordered_params(ff0, ff1)`...

Would it be easy / useful to reduce duplication? (Perhaps by modifying the signature to accept a `top` and `ff_params`)

timemachine/md/enhanced.py
```diff
@@ -470,6 +471,75 @@ def get_solvent_phase_system(mol, ff, box_width=3.0, margin=0.5, minimize_energy
...
+
+    top = topology.RelativeFreeEnergyForcefield(mol, ff0, ff1)
+    afe = free_energy.AbsoluteFreeEnergy(mol, top)
+    combined_ff_params = combine_ordered_params(ff0, ff1)
+    potentials, params, masses = afe.prepare_host_edge(combined_ff_params, water_system)
+
+    # concatenate (optionally minimized) water_coords and ligand_coords
+    ligand_coords = get_romol_conf(mol)
+    if minimize_energy:
+        new_water_coords = minimizer.minimize_host_4d([mol], water_system, water_coords, ff0, water_box)
```
maxentile, line 512: Question: Would it ever be relevant that we use `ff0` rather than, say,`interpolate(ff0, ff1, 0.5)` for initial minimization?

(If so, should add a note under `minimize_energy` parameter in docstring)

timemachine/fe/absolute_hydration.py
```diff
@@ -156,3 +156,125 @@ def log_prob(xs, lam):
     resample = partial(conditional_multinomial_resample, thresh=resample_thresh)
 
     return samples, lambdas, propagate, log_prob, resample
+
+
+def set_up_ahfe_for_smc_parameter_changes(
```
maxentile, line 161: Naming: is `ahfe` accurate?

(Perhaps `set_up_smc_parameter_changes_at_endstates` or similar?)

timemachine/fe/absolute_hydration.py
```diff
@@ -156,3 +156,125 @@ def log_prob(xs, lam):
...
+    ----------
+    is_vacuum: bool
+        Set to True to set up the vacuum leg, using NVT.
+        Set to False to set up the solvent leg, using NPT.
+
+    Returns
+    -------
+    * reduced_potential_fxn
+    * mover at lam=0
+    * mover at lam=1
```
maxentile, line 185: Returns a single mover with lam=None -- should add comment about this.

(That intended usage is `mover.lamb = 0.0; x_next = mover.move(x)` or similar)

tests/test_absolute_hydration.py
```diff
@@ -0,0 +1,166 @@
...
+    u_rev = batched_u_fxn(vacuum_samples1, state0) - batched_u_fxn(vacuum_samples1, state1)
+    bar_dg = pymbar.BAR(u_fwd, u_rev, compute_uncertainty=False)
+    assert smc_dg == pytest.approx(bar_dg, abs=1e-1)
+
+
+def test_smc_parameter_change_solvent():
+    """
+    Tests correctness of using SMC to switch between parameters.
+    The SMC predicted dg ff0 -> ff1 should match results from
+    BAR using independent sampling.
```
maxentile, line 102: Another possible test is to predict in both directions `ff0 -> ff1` and `ff1 -> ff0`

tests/test_topology.py
```diff
@@ -367,3 +370,72 @@ def test_component_idxs():
...
+
+    def u_endstates(ubp, params) -> Tuple[float, float]:
+        u = ubp.bind(params).bound_impl(precision=np.float32)
+        _, _, u0_val = u.execute(coords, box, lam=0)
+        _, _, u1_val = u.execute(coords, box, lam=1)
+        return u0_val, u1_val
+
+    u0_combined, u1_combined = u_endstates(combined_ubp, combined_qlj_params)
+    u0, _ = u_endstates(ubp0, qlj0_params)
+    u1, _ = u_endstates(ubp1, qlj1_params)
```
maxentile, line 418: Not important since only executed once in test, but `ubp.bind(params).bound_impl(np.float32)` is an setup operation that only needs to be done once, so could consider 

`def get_impl(ubp, params): return ubp.bind(params).bound_impl(precision=np.float32); impl0 = get_impl(ubp0, qlj0_params); u0 = impl0.execute(coords, box, lam=0)` or similar. (In case the test is modified in the future to be called multiple times with different (coords, box, lam).)

maxentile (approved):
Looks good to me!

(Some minor, non-blocking comments)

timemachine/md/enhanced.py
jkausrelay: NTS: deboggle

timemachine/fe/topology.py
jkausrelay: Yeah, probably not. At least I can't think of a reason to do this right now.

timemachine/md/enhanced.py
jkausrelay: https://github.com/proteneer/timemachine/issues/757

timemachine/fe/topology.py
jkausrelay: No, there isn't. I'll switch to Chem.Mol.

timemachine/fe/absolute_hydration.py
jkausrelay: updated

timemachine/fe/absolute_hydration.py
jkausrelay: * mover with lamb=None. The mover.lamb attribute must be set before use.

timemachine/md/enhanced.py
jkausrelay: Could be for later

PR #752, opened 2022-06-08
Chiral restraints

proteneer:
This PR implements chiral restraints (which will be later used in SingleTopology, and probably in DualTopology as well) in a controlled way. Suppose the domain of the probability distribution is separated into R and S regions. Suppose we intend to efficiently sample only the R region, using a sampler like MD. These restraints have the following properties:

1) If we sample into the S region, the restraining term turns on, and tries to force the geometry back into R.
2) If we sample into the R region, the restraint is a "flat-bottom", and does not affect the probability distribution within R. 

The two supported types of restraints are chiral pyramidal restraints, and chiral torsion restraints. Both are defined over 4 atoms but with a different convention for computing the chiral volume. The pyramidal restraint is implemented as the triple product of 4 points, and the torsion restraint is implemented as dot product between two cross products. The restraint is also normalized such that its volume is strictly between -1 and 1 (so as to make the strength of the restraint independent of the bond lengths). Tetrahedral restraints are implemented as 4-fold pyramidal restraints arising from 4 choose 3 choices of neighbors.

These restraints will greatly simplify the single topology code, as they're fairly robust against multiple "normal" behaviors (implemented as tests):

1) Inversion of pyramidal centers
2) Inversion of tetrahedral centers
3) Inversion of cis/trans bonds
4) Inversion of even things like spiro centers.

Note that they are not bullet proof, and do require parameterization of a force constant. There are certain pathologies (eg. enforcing trans on a ring bond that should be cis, which would require stretching bonds). In addition, if there are large vdw barriers (eg. nonbonded clashes from either the environment or intramolecular clashes), these probably wouldn't work (but we can detect this cases easily by re-weighting into the +inf potential, or by simply inspecting the geometries themselves to see if we achieve the desired state).

In otherwise, the chiral restraints are strong enough to overcome most angle and torsion terms, but may not be enough to overcome bonded or lennard jones interactions. However, needing to overcome a bond terms is usually indicative of a ring problem, and most tetrahedral centers have 1-2, 1-3, and weaken'd 1-4 exclusions to allow for easier local conversion.

### DIFF
--- a/tests/test_chiral_restraints.py
+++ b/tests/test_chiral_restraints.py
@@ -1,0 +1,522 @@
+import multiprocessing
+import os
+
+os.environ["XLA_FLAGS"] = "--xla_force_host_platform_device_count=" + str(multiprocessing.cpu_count())
+
+import jax
+import jax.numpy as jnp
+import numpy as np
+import scipy
+from rdkit import Chem
+from rdkit.Chem import AllChem
+
+from timemachine.fe import chiral_utils, topology, utils
+from timemachine.ff import Forcefield
+from timemachine.integrator import simulate
+from timemachine.potentials.chiral_restraints import (
+    U_chiral_atom,
+    U_chiral_atom_batch,
+    U_chiral_bond,
+    U_chiral_bond_batch,
+    pyramidal_volume,
+    torsion_volume,
+)
+
+
+def minimize_scipy(x0, U_fn):
+    shape = x0.shape
+
+    def U_flat(x_flat):
+        x_full = x_flat.reshape(*shape)
+        return U_fn(x_full)
+
+    grad_bfgs_fn = jax.grad(U_flat)
+    res = scipy.optimize.minimize(U_flat, x0.reshape(-1), jac=grad_bfgs_fn)
+    xi = res.x.reshape(*shape)
+    return xi
+
+
+def simulate_system(U_fn, x0, num_samples=20000):
+    num_atoms = x0.shape[0]
+    x_min = minimize_scipy(x0, U_fn)
+    seed = 2023
+
+    num_workers = multiprocessing.cpu_count()
+    samples_per_worker = int(np.ceil(num_samples / num_workers))
+
+    # batches_per_worker = num_samples // num_workers
+    burn_in_batches = 2000
+    frames, _ = simulate(
+        x_min, U_fn, 300.0, np.ones(num_atoms) * 4.0, 500, samples_per_worker + burn_in_batches, num_workers, seed=seed
+    )
+    # (ytz): discard burn in batches
+    frames = frames[:, burn_in_batches:, :, :]
+    # collect over all workers
+    frames = frames.reshape(-1, num_atoms, 3)[:num_samples]
+    # sanity check that we didn't undersample
+    assert len(frames) == num_samples
+    return frames
+
+
+def test_setup_chiral_atom_restraints():
+    """On a methane conformer, assert that permuting coordinates or permuting restr_idxs
+    both independently toggle the chiral restraint"""
+    mol = Chem.MolFromMolBlock(
+        """
+  Mrv2202 06072215563D
+
+  5  4  0  0  0  0            999 V2000
+    0.0000    0.0000    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.3633   -0.5138    0.8900 H   0  0  0  0  0  0  0  0  0  0  0  0
+    1.0900    0.0000    0.0000 H   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.3633    1.0277    0.0000 H   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.3633   -0.5138   -0.8900 H   0  0  0  0  0  0  0  0  0  0  0  0
+  1  2  1  0  0  0  0
+  1  3  1  0  0  0  0
+  1  4  1  0  0  0  0
+  1  5  1  0  0  0  0
+M  END
+$$$$""",
+        removeHs=False,
+    )
+
+    # needs to be batched in order for jax to play nicely
+    x0 = utils.get_romol_conf(mol)
+    normal_restr_idxs = chiral_utils.setup_chiral_atom_restraints(mol, x0, 0)
+
+    x0_inverted = x0[[0, 2, 1, 3, 4]]  # swap two atoms
+    inverted_restr_idxs = chiral_utils.setup_chiral_atom_restraints(mol, x0_inverted, 0)
+
+    # check the sign of the resulting idxs
+    k = 1000.0
+    assert np.all(np.asarray(U_chiral_atom_batch(x0, normal_restr_idxs, k)) == 0)
+    assert np.all(np.asarray(U_chiral_atom_batch(x0, inverted_restr_idxs, k)) > 0)
+    assert np.all(np.asarray(U_chiral_atom_batch(x0_inverted, normal_restr_idxs, k)) > 0)
+    assert np.all(np.asarray(U_chiral_atom_batch(x0_inverted, inverted_restr_idxs, k)) == 0)
+
+
+def test_setup_chiral_bond_restraints():
+    """On a 'Cl/C(F)=N/F' conformer, assert that flipping a dihedral angle or permuting restr_idxs
+    both independently toggle the chiral bond restraint"""
+
+    mol_cis = Chem.MolFromSmiles(r"Cl\C(F)=N/F")
+    mol_trans = Chem.MolFromSmiles(r"Cl\C(F)=N\F")
+
+    AllChem.EmbedMolecule(mol_cis)
+    AllChem.EmbedMolecule(mol_trans)
+
+    # needs to be batched in order for jax to play nicely
+    x0_cis = utils.get_romol_conf(mol_cis)
+    x0_trans = utils.get_romol_conf(mol_trans)
+    src_atom = 1
+    dst_atom = 3
+    normal_restr_idxs, signs = chiral_utils.setup_chiral_bond_restraints(mol_cis, x0_cis, src_atom, dst_atom)
+
+    inverted_restr_idxs, inverted_signs = chiral_utils.setup_chiral_bond_restraints(
+        mol_trans, x0_trans, src_atom, dst_atom
+    )
+    k = 1000.0
+
+    assert np.all(np.asarray(U_chiral_bond_batch(x0_cis, normal_restr_idxs, k, signs)) == 0)
+    assert np.all(np.asarray(U_chiral_bond_batch(x0_cis, inverted_restr_idxs, k, inverted_signs)) > 0)
+    assert np.all(np.asarray(U_chiral_bond_batch(x0_trans, normal_restr_idxs, k, signs)) > 0)
+    assert np.all(np.asarray(U_chiral_bond_batch(x0_trans, inverted_restr_idxs, k, inverted_signs)) == 0)
+
+
+def test_chiral_restraints_pyramidal():
+    """For ammonium, assert that:
+    * without chiral restraints, up and down states are ~ equally sampled
+    * with chiral restraints, ~ only the specified state is sampled"""
+    mol = Chem.MolFromMolBlock(
+        """
+  Mrv2202 05192218063D
+
+  4  3  0  0  0  0            999 V2000
+   -0.0541    0.5427   -0.3433 N   0  0  0  0  0  0  0  0  0  0  0  0
+    0.4368    0.0213    0.3859 H   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.9636    0.0925   -0.4646 H   0  0  0  0  0  0  0  0  0  0  0  0
+    0.4652    0.3942   -1.2109 H   0  0  0  0  0  0  0  0  0  0  0  0
+  1  2  1  0  0  0  0
+  1  3  1  0  0  0  0
+  1  4  1  0  0  0  0
+M  END
+$$$$""",
+        removeHs=False,
+    )
+
+    ff = Forcefield.load_from_file("smirnoff_1_1_0_sc.py")
+    s_top = topology.BaseTopology(mol, ff)
+    x0 = utils.get_romol_conf(mol)
+
+    # check initial chirality
+    restr_idxs = np.array([0, 1, 2, 3])
+    vol = pyramidal_volume(*x0[restr_idxs])
+    assert vol > 0.5
+    system = s_top.setup_end_state()
+    U_fn = system.get_U_fn()
+    vols_orig = []
+    frames = simulate_system(U_fn, x0)
+    for f in frames:
+        vol = pyramidal_volume(*f[restr_idxs])
+        vols_orig.append(vol)
+
+    def U_total(x):
+        return U_fn(x) + U_chiral_atom(x, restr_idxs, 1000.0)
+
+    vols_chiral = []
+    frames = simulate_system(U_total, x0)
+    for f in frames:
+        vol = pyramidal_volume(*f[restr_idxs])
+        vols_chiral.append(vol)
+
+    vols_orig = np.array(vols_orig)
+    vols_chiral = np.array(vols_chiral)
+
+    # should be within 5% of 50/50 for the original distribution, the <0 case
+    # is implied
+    assert np.abs(np.mean(vols_orig < 0) - 0.5) < 0.05
+    assert np.abs(np.mean(vols_chiral < 0) - 0.95) < 0.05
+
+    ref_dist = [x for x in vols_orig if x < 0]
+    test_dist = [x for x in vols_chiral if x < 0]
+
+    # should be indistinguishable under KS-test
+    ks, pv = scipy.stats.ks_2samp(ref_dist, test_dist)
+    assert ks < 0.05
+    assert pv > 0.10
+
+    # # useful plotting diagnostics, do not remove.
+    # # compare original distributions
+    # plt.hist(vols_orig, bins=np.linspace(-1, 1, 80), alpha=0.5, label="no_chiral_restr", density=True)
+    # plt.hist(vols_chiral, bins=np.linspace(-1, 1, 80), alpha=0.5, label="with_chiral_restr", density=True)
+    # plt.legend()
+    # plt.title("pyramidal chirality")
+    # plt.xlabel("chiral volume")
+    # plt.ylabel("samples")
+    # plt.title("raw")
+    # plt.show()
+
+    # # time series
+    # plt.plot(vols_chiral, label="with_chiral_restraint")
+    # plt.show()
+
+    # # compare filtered distributions
+    # plt.hist([x for x in vols_orig if x < 0], bins=np.linspace(-1, 1, 80), alpha=0.5, label="no_chiral_restr", density=True)
+    # plt.hist([x for x in vols_chiral if x < 0], bins=np.linspace(-1, 1, 80), alpha=0.5, label="with_chiral_restr", density=True)
+    # plt.legend()
+    # plt.title("pyramidal chirality")
+    # plt.xlabel("chiral volume")
+    # plt.ylabel("samples")
+    # plt.title("normalized")
+    # plt.show()
+
+
+class BaseTopologyRescaledCharges(topology.BaseTopology):
+    def __init__(self, scale, *args, **kwargs):
+        self.scale = scale
+        super().__init__(*args, **kwargs)
+
+    def parameterize_nonbonded(self, ff_q_params, ff_lj_params):
+        params, nb = topology.BaseTopology.parameterize_nonbonded(self, ff_q_params, ff_lj_params)
+        charge_indices = jnp.index_exp[:, 0]
+        new_params = jnp.asarray(params).at[charge_indices].multiply(self.scale)
+        return new_params, nb
+
+    def parameterize_nonbonded_pairlist(self, ff_q_params, ff_lj_params):
+        params, nb = topology.BaseTopology.parameterize_nonbonded_pairlist(self, ff_q_params, ff_lj_params)
+        charge_indices = jnp.index_exp[:, 0]
+        new_params = jnp.asarray(params).at[charge_indices].multiply(self.scale)
+        return new_params, nb
+
+
+def test_chiral_restraints_torsion():
+    """For a charge-scaled version of hydrogen peroxide, assert that:
+    * without chiral bond restraints, cis/trans states are sampled ~ 30%/70%
+    * with chiral bond restraints, ~ only specified state is sampled"""
+    mol = Chem.MolFromMolBlock(
+        """
+  Mrv2202 06062222163D
+
+  4  3  0  0  0  0            999 V2000
+   -0.1757   -0.7570    0.3351 H   0  0  0  0  0  0  0  0  0  0  0  0
+    0.0031   -0.0645   -0.3351 O   0  0  0  0  0  0  0  0  0  0  0  0
+    1.3069    0.0645   -0.3351 O   0  0  0  0  0  0  0  0  0  0  0  0
+    1.4857    0.7570    0.3351 H   0  0  0  0  0  0  0  0  0  0  0  0
+  2  3  1  0  0  0  0
+  1  2  1  0  0  0  0
+  3  4  1  0  0  0  0
+M  END
+$$$$""",
+        removeHs=False,
+    )
+
+    ff = Forcefield.load_from_file("smirnoff_1_1_0_sc.py")
+    scale = 0.4
+    s_top = BaseTopologyRescaledCharges(scale, mol, ff)
+    x0 = utils.get_romol_conf(mol)
+
+    # check initial chirality
+    torsion_idxs = np.array([0, 1, 2, 3])
+    vol = torsion_volume(*x0[torsion_idxs])
+
+    assert abs(vol - 0.0) < 0.05
+    system = s_top.setup_end_state()
+    U_fn = system.get_U_fn()
+
+    vols_orig = []
+    frames = simulate_system(U_fn, x0)
+    for f in frames:
+        vols_orig.append(torsion_volume(*f[torsion_idxs]))
+    vols_orig = np.array(vols_orig)
+
+    # the 30/70 ratio is dependent on the scale defined above, which affects the repulsive
+    # strength of the hydrogens.
+    assert np.abs(np.mean(vols_orig > 0) - 0.3) < 0.05
+
+    all_signs = [1, -1]  # [trans, cis]
+    for sign in all_signs:
+
+        def U_total(x):
+            return U_fn(x) + U_chiral_bond(x, torsion_idxs, 1000.0, sign)
+
+        vols_chiral = []
+        frames = simulate_system(U_total, x0)
+        for f in frames:
+            vols_chiral.append(torsion_volume(*f[torsion_idxs]))
+        vols_chiral = np.array(vols_chiral)
+
+        # # useful for plotting
+        # import matplotlib.pyplot as plt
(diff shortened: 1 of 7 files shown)


timemachine/ff/handlers/nonbonded.py
```diff
@@ -349,7 +349,15 @@ def static_parameterize(params, smirks, mol):
...
-    pass
+    def __init__(self, *args, **kwargs):
+        self.scale = 1
+        super().__init__(*args, **kwargs)
+
+    def partial_parameterize(self, params, mol):
+        return self.static_parameterize(params, self.smirks, mol) * self.scale
+
+    def parameterize(self, mol):
+        return self.static_parameterize(self.params, self.smirks, mol) * self.scale
```
maxentile, line 360: Document intent of `self.scale`? Introduces possibility for `static_parameterize` to have inconsistent behavior from `partial_parameterize` and `parameterize` (if `scale != 1`)

timemachine/potentials/chiral_restraints.py
```diff
@@ -0,0 +1,131 @@
...
+    Returns
+    -------
+    float
+        A number between -1.0 < x < 1.0 denoting the normalized chirality
+
+    """
+    # compute vectors
+    v0 = x1 - x0
+    v1 = x2 - x0
+    v2 = x3 - x0
```
maxentile, line 33: naming comments from https://github.com/proteneer/timemachine/pull/746#discussion_r881070577

timemachine/potentials/chiral_restraints.py
```diff
@@ -0,0 +1,131 @@
...
+    rkj = rkj / jnp.linalg.norm(rkj)
+    rkl = rkl / jnp.linalg.norm(rkl)
+
+    n1 = jnp.cross(rij, rkj)
+    n2 = jnp.cross(rkj, rkl)
+
+    return jnp.dot(n1, n2)
+
+
+def U_chiral_atom(x, idxs, kc):
```
maxentile, line 81: docstring "kc * v**2 if v > 0 else 0"

timemachine/potentials/chiral_restraints.py
```diff
@@ -0,0 +1,131 @@
...
+
+def U_chiral_bond_batch(x, all_idxs, kc, all_s):
+
+    x = jnp.array(x)
+
+    nrgs = []
+    for idxs, s in zip(all_idxs, all_s):
+        nrgs.append(U_chiral_bond(x, idxs, kc, s))
+
+    return nrgs
```
maxentile: Is this because of the assertion commented out in L98?

(Might add a comment / TODO about when this function might be removed, e.g. if an assertion in U_chiral_bond is removed / re-written to be compatible with more Jax transformations)

maxentile: (Oh! Resolved in https://github.com/proteneer/timemachine/pull/752/commits/7dd4c25bd8c1a8af94ae1ea604121878a0d0800c , just hadn't refreshed)

timemachine/potentials/chiral_restraints.py
```diff
@@ -0,0 +1,131 @@
...
+    return jnp.dot(n1, n2)
+
+
+def U_chiral_atom(x, idxs, kc):
+    # guard against numpy/raw lists
+    x = jnp.array(x)
+    assert len(idxs) == 4
+    x0, x1, x2, x3 = x[idxs]
+    v = pyramidal_volume(x0, x1, x2, x3)
+    return jnp.where(v > 0, kc * v ** 2, 0.0)
```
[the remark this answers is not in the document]
jkausrelay: Yeah, since you do this below for the torsions.

timemachine/fe/chiral_utils.py
```diff
@@ -0,0 +1,55 @@
...
+    # build chiral restraints
+    restr_idxs = []
+    signs = []
+    for a in src_nbs:
+        for b in dst_nbs:
+            # set up torsion i,j,k,l
+            i, j, k, l = a, src_idx, dst_idx, b
+            vol = torsion_volume(conf[i], conf[j], conf[k], conf[l])
+            restr_idxs.append([i, j, k, l])
+            if vol < 0:
```
jkausrelay, line 50: If vol is negative, sign is positive?

Also nit: `signs.append(-np.sign(vol))` or `signs.append(np.sign(vol))` depending on answer to above.

tests/test_chiral_restraints.py
```diff
@@ -0,0 +1,506 @@
...
+    system = s_top.setup_end_state(ff)
+    U_fn = system.get_U_fn()
+
+    vols_orig = []
+    frames = simulate_system(U_fn, x0)
+    for f in frames:
+        vols_orig.append(torsion_volume(*f[torsion_idxs]))
+    vols_orig = np.array(vols_orig)
+
+    assert (np.sum(vols_orig > 0) / len(vols_orig) - 0.22) < 0.05
```
jkausrelay, line 253: Comment about 22/78%?

tests/test_chiral_restraints.py
```diff
@@ -0,0 +1,506 @@
...
+        test_dist = [x for x in vols_chiral if sign * x < 0]
+        # should be indistinguishable under KS-test
+        ks, pv = scipy.stats.ks_2samp(ref_dist, test_dist)
+        print("KS, PV", ks, pv)
+        assert ks < 0.05
+        assert pv > 0.10
+
+
+def test_chiral_restraints_tetrahedral():
+    # test that we can restrain the chirality of a tetrahedral molecule
```
jkausrelay, line 302: ... to the opposite of it's starting chirality

timemachine/ff/handlers/nonbonded.py
proteneer: will add documentation - note this is only defined on the `SimpleChargeHandler`, which we only ever use for production. This is to allow me to weaken the charges so that I can sample the "trans" conformation of `H-O-O-H` less frequently. I could just add another `smirnoff_1_1_0_rescale_sc.py` as an alternative?

timemachine/fe/chiral_utils.py
proteneer: will add comment

timemachine/ff/handlers/nonbonded.py
maxentile: >note this is only defined on the SimpleChargeHandler, which we only ever use for production. This is to allow me to weaken the charges so that I can sample the "trans" conformation of H-O-O-H less frequently.

Okay, maybe a non-issue if not used for production, and if it simplifies some use that's needed in this PR.

(Currently I just don't understand the intent of putting it inside the handler in a way that makes `static_parameterize` inconsistent with `partial_parameterize`, instead of, say, explicitly multiplying the assigned charges by `scale` when needed (similar to how we rescale charges, etc. after parameterization in https://github.com/proteneer/timemachine/blob/d7c8ee5c1ab421ef17922621089ec910467bb9b8/timemachine/md/enhanced.py#L120-L124 and https://github.com/proteneer/timemachine/blob/776873a013e65f8d825570312d0fa87d3a438522/timemachine/fe/topology.py#L296-L309). Also note `scale != 1` will not preserve net charge if `net_charge != 0`.)

> I could just add another smirnoff_1_1_0_rescale_sc.py as an alternative?

This seems unnecessary

proteneer: The latter approach is fine too! but I'd need to subclass the `BaseTopology` class, or modify it to include a scale term. I can just make a `BaseTopologyRescaledCharges` as well.

jkausrelay: Can you just do `ff.q_handle.params *= 0.5`?

proteneer: That modifies it in a non-differentiable way (since it's not traceable in the constructor)

tests/test_chiral_restraints.py
```diff
@@ -0,0 +1,506 @@
...
+
+    ff = Forcefield.load_from_file("smirnoff_1_1_0_sc.py")
+    s_top = topology.BaseTopology(mol, ff)
+    x0 = utils.get_romol_conf(mol)
+
+    # check initial chirality
+    restr_idxs = np.array([0, 1, 2, 3])
+    vol = pyramidal_volume(*x0[restr_idxs])
+    assert vol > 0.5
+    system = s_top.setup_end_state(ff)
```
mcwitt, line 157: I don't see `setup_end_state` defined?

tests/test_chiral_restraints.py
```diff
@@ -0,0 +1,506 @@
...
+    frames = simulate_system(U_total, x0)
+    for f in frames:
+        vol = pyramidal_volume(*f[restr_idxs])
+        vols_chiral.append(vol)
+
+    vols_orig = np.array(vols_orig)
+    vols_chiral = np.array(vols_chiral)
+
+    # should be within 5% of 50/50 for the original distribution, a bit redundant
+    assert (np.sum(vols_orig > 0) / len(vols_orig) - 0.5) < 0.05
```
mcwitt, line 178: Could this be combined with the following line if we use abs()? Nit: use `np.mean`
```suggestion
    assert np.abs(np.mean(vols_orig > 0) - 0.5) < 0.05
```

tests/test_chiral_restraints.py
mcwitt: Would it make sense to use abs() to combine the two conditions?

tests/test_chiral_restraints.py
```diff
@@ -0,0 +1,506 @@
...
+        print("SIGN", sign)
+
+        def U_total(x):
+            return U_fn(x) + U_chiral_bond(x, torsion_idxs, 1000.0, sign)
+
+        vols_chiral = []
+        frames = simulate_system(U_total, x0)
+        for f in frames:
+            vols_chiral.append(torsion_volume(*f[torsion_idxs]))
+        vols_chiral = np.array(vols_chiral)
```
mcwitt, line 286: nit: cleaner as list comprehension?
```suggestion
        frames = simulate_system(U_total, x0)
        vols_chiral = np.array([torsion_volume(*f[torsion_idxs]) for f in frames])
```

tests/test_chiral_restraints.py
```diff
@@ -0,0 +1,506 @@
...
+    system = s_top.setup_end_state(ff)
+    U_fn = system.get_U_fn()
+
+    vols_orig = []
+    frames = simulate_system(U_fn, x0)
+    for f in frames:
+        vol_list = []
+        for p in perms:
+            vol_list.append(pyramidal_volume(*f[p]))
+        vols_orig.append(vol_list)
```
mcwitt, line 367: nit: cleaner as list comprehension?

tests/test_chiral_restraints.py
```diff
@@ -0,0 +1,506 @@
...
+    vols_orig = np.array(vols_orig)
+    vols_chiral = np.array(vols_chiral)
+
+    # should be within 5% of 50/50 for the original distribution, a bit redundant
+    assert (np.sum(vols_orig > 0) / len(vols_orig) - 0.5) < 0.05
+    assert (np.sum(vols_orig < 0) / len(vols_orig) - 0.5) < 0.05
+    assert (np.sum(vols_chiral > 0) / len(vols_chiral) - 0.95) < 0.05
+
+    ref_dist = [x for x in vols_orig if x < 0]
+    test_dist = [x for x in vols_chiral if x < 0]
```
mcwitt, line 183: I might be missing something here, but don't we want to compare the distributions in the allowed region (`x > 0`)?

tests/test_chiral_restraints.py
proteneer: missed this in a bad cherry-pick from the single_topology PR,  just added the definitions now

tests/test_chiral_restraints.py
proteneer: If `x` is the chiral volume, the allowed region is `x < 0`, since the chiral penalty is turned on when `x > 0`.

tests/test_chiral_restraints.py
proteneer: willfix

tests/test_chiral_restraints.py
proteneer: willfix

tests/test_chiral_restraints.py
```diff
@@ -0,0 +1,506 @@
...
+        # plt.hist([x for x in vols_chiral if sign*x < 0], bins=np.linspace(-1, 1, 80), alpha=0.5, label="with_chiral_restr", density=True)
+        # plt.legend()
+        # plt.title("torsion chirality")
+        # plt.xlabel("chiral volume")
+        # plt.ylabel("samples")
+        # plt.title("normalized")
+        # plt.show()
+
+        # we should predominantly sample the correct chiral state.
+        assert (np.sum(vols_chiral > 0) / len(vols_chiral) - 0.95) < 0.05
```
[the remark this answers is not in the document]
proteneer: yeah this was just wrong - willfix. (the sign should've been < 0 as well I think)

tests/test_chiral_restraints.py
mcwitt: Got it -- sorry, had misunderstood the sign convention.

tests/test_chiral_restraints.py
```diff
@@ -0,0 +1,506 @@
...
+        for p in perms:
+            vol_list.append(pyramidal_volume(*f[p]))
+        vols_chiral.append(vol_list)
+
+    vols_orig = np.array(vols_orig).reshape(-1)
+    vols_chiral = np.array(vols_chiral).reshape(-1)
+
+    # this should generate a symmetric distribution
+    ref_dist = np.array([x for x in vols_orig if x > 0])
+    test_dist = np.array([x for x in vols_chiral if x < 0])
```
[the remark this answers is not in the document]
proteneer: The ref_dist samples predominantly positive chiral volumes, and the test_dist should pre-dominantly sample the negative chiral volumes, and the distribution is symmetric about vol=0. I'll add a comment here

timemachine/ff/handlers/nonbonded.py
proteneer: actually sorry - I misinterpreted this! I thought you meant in the constructor - but it sounds like you just mean in the handle itself. To be clear, `ff.q_handle.params` are BCC parameters, so scaling the raw parameters only adjusts the polarization across bonds. However, we want the actual point charges to be scaled here.

timemachine/potentials/chiral_restraints.py
proteneer: willfix

tests/test_chiral_restraints.py
```diff
@@ -0,0 +1,506 @@
...
+M  END
+$$$$""",
+        removeHs=False,
+    )
+
+    # needs to be batched in order for jax to play nicely
+    x0 = utils.get_romol_conf(mol)
+    normal_restr_idxs = chiral_utils.setup_chiral_atom_restraints(mol, x0, 0)
+
+    x0_inverted = [x0[0], x0[2], x0[1], x0[3], x0[4]]
```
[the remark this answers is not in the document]
mcwitt: > i think it's pissing the x0 in the above expression
whoops, fixed!

timemachine/fe/system.py
```diff
@@ -0,0 +1,61 @@
+import functools
+
+import jax.numpy as jnp
+import numpy as np
+
+from timemachine.potentials import bonded, nonbonded
+
+
+class VacuumSystem:
```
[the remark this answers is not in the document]
proteneer: Can you define methods on `dataclass/NamedTuples`?

timemachine/fe/system.py
```diff
@@ -0,0 +1,61 @@
...
+        self.angle = angle
+        self.torsion = torsion
+        self.nonbonded = nonbonded
+
+    def get_U_fn(self):
+        """
+        Return a jax function that evaluates the potential energy of a set of coordinates.
+        """
+
+        box = None
```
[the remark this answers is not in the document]
proteneer: Can you clarify by inline?

timemachine/fe/system.py
```diff
@@ -0,0 +1,61 @@
...
+            pairs=np.array(self.nonbonded.get_idxs()),
+            params=np.array(self.nonbonded.params),
+            box=box,
+            beta=self.nonbonded.get_beta(),
+            cutoff=self.nonbonded.get_cutoff(),
+            rescale_mask=np.array(self.nonbonded.get_rescale_mask()),
+        )
+
+        def U_fn(x):
+            vdw, q = nbpl_U(x)
```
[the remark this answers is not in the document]
proteneer: willfix

timemachine/fe/topology.py
```diff
@@ -245,6 +246,39 @@ def parameterize_nonbonded(self, ff_q_params, ff_lj_params):
...
         return params, nb
 
+    def parameterize_nonbonded_pairlist(self, ff_q_params, ff_lj_params):
+        # generate intramolecular nonbonded pairlist
+        # use same scale factors for electrostatics and vdWs
+        exclusion_idxs, scale_factors = nonbonded.generate_exclusion_idxs(
+            self.mol, scale12=_SCALE_12, scale13=_SCALE_13, scale14=_SCALE_14
+        )
+
+        # note: use same sf for electrostatics and vdw
```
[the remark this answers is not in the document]
proteneer: willfix

timemachine/fe/topology.py
```diff
@@ -245,6 +246,39 @@ def parameterize_nonbonded(self, ff_q_params, ff_lj_params):
 
         return params, nb
 
+    def parameterize_nonbonded_pairlist(self, ff_q_params, ff_lj_params):
+        # generate intramolecular nonbonded pairlist
```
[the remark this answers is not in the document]
proteneer: willfix

tests/test_chiral_restraints.py
proteneer: fixed inline!

timemachine/fe/system.py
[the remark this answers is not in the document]
mcwitt: Yes, both allow defining methods (for namedtuples you need to use the `class Foo(NamedTuple):` syntax). I find them convenient when the constructor doesn't need to do any work. Comparison of namedtuple and dataclass:
- dataclass has more features and can inherit from another class. Fields are mutable by default (but can pass `frozen=True` to disallow mutation)
- NamedTuples cannot inherit from another class. Fields are immutable

timemachine/ff/handlers/nonbonded.py
jkausrelay: Makes sense

mcwitt (approved):

timemachine/fe/system.py
```diff
@@ -0,0 +1,61 @@
...
+
+    # utility system container
+
+    def __init__(self, bond, angle, torsion, nonbonded):
+        self.bond = bond
+        self.angle = angle
+        self.torsion = torsion
+        self.nonbonded = nonbonded
+
+    def get_U_fn(self):
```
jkausrelay, line 19: (deboggle): Seems to overlap with enhanced.VacuumState: https://github.com/proteneer/timemachine/blob/master/timemachine/md/enhanced.py#L63

timemachine/fe/system.py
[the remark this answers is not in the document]
proteneer: I tried `namedtuple` and thought it was slightly awkward still:

```
class VacuumSystem(namedtuple("VacuumSystem", "bond angle torsion nonbonded"))
```

1) If I later on need to support constructors that do work the signature would have to change (not sure how the `__init__`) would work in relation to the required `super` calls.
2) I don't care much about (im)mutability at this point

I'd still prefer to `namedtuple` for things that are truly data containers - which this class isn't quite.

timemachine/fe/system.py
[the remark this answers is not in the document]
proteneer: ah sorry I see what you mean! Sorry this was just laziness on my part from previous code that set `box=box` when `box=[[x,0,0],[0,y,0],[0,0,z]]` or something. I can totally inline this. (willfix)

timemachine/fe/system.py
proteneer: ugh - good call. We should deboggle this for sure - but I didn't want to touch production code for this PR yet.

timemachine/fe/topology.py
```diff
@@ -292,6 +329,25 @@ def parameterize_periodic_torsion(self, proper_params, improper_params):
         combined_potential = potentials.PeriodicTorsion(combined_idxs, combined_lambda_mult, combined_lambda_offset)
         return combined_params, combined_potential
 
+    def setup_end_state(self, ff):
```
jkausrelay, line 332: Should this just be 
`setup_end_state(self)` and use `self.ff` here?

It seems odd to pass in a (possibly different) ff here.

proteneer: good call - goof'd on my end

tests/test_chiral_restraints.py
```diff
@@ -0,0 +1,526 @@
...
+    U_chiral_bond_batch,
+    pyramidal_volume,
+    torsion_volume,
+)
+
+
+def minimize_scipy(x0, U_fn):
+    N = x0.shape[0]
+
+    def U_bfgs(x_flat):
```
maxentile, line 28: Naming: `U_bfgs` -> `U_flat`?

(The energy function isn't modified by the optimize.)

jkausrelay (approved):

tests/test_chiral_restraints.py
```diff
@@ -0,0 +1,526 @@
...
+        x_full = x_flat.reshape(N, 3)
```
maxentile, line 29: Generality: `x_flat.reshape(*x0.shape)` rather than `x_flat.reshape(x0.shape[0], 3)`?

tests/test_chiral_restraints.py
```diff
@@ -0,0 +1,526 @@
...
+    # (ytz): discard burn in batches
+    frames = frames[:, burn_in_batches:, :, :]
+    # collect over all workers
+    frames = frames.reshape(-1, num_atoms, 3)[:num_samples]
+    # sanity check that we didn't undersample
+    assert len(frames) == num_samples
+    return frames
+
+
+def test_setup_chiral_atom_restraints():
```
maxentile, line 60: Add one-sentence description of what is being asserted
```suggestion
def test_setup_chiral_atom_restraints():
    """On a methane conformer, assert that permuting coordinates or permuting restr_idxs
    both independently toggle the chiral restraint"""
```

tests/test_chiral_restraints.py
```diff
@@ -0,0 +1,526 @@
...
+
+    # check the sign of the resulting idxs
+    k = 1000.0
+    assert np.all(np.asarray(U_chiral_atom_batch(x0, normal_restr_idxs, k)) == 0)
+    assert np.all(np.asarray(U_chiral_atom_batch(x0, inverted_restr_idxs, k)) > 0)
+    assert np.all(np.asarray(U_chiral_atom_batch(x0_inverted, normal_restr_idxs, k)) > 0)
+    assert np.all(np.asarray(U_chiral_atom_batch(x0_inverted, inverted_restr_idxs, k)) == 0)
+
+
+def test_setup_chiral_bond_restraints():
```
maxentile, line 95: Add one-sentence description of what is being asserted
```suggestion
def test_setup_chiral_bond_restraints():
    """On a 'Cl/C(F)=N/F' conformer, assert that flipping a dihedral angle or permuting restr_idxs
    both independently toggle the chiral bond restraint"""
```

tests/test_chiral_restraints.py
```diff
@@ -0,0 +1,526 @@
...
+        removeHs=False,
+    )
+
+    # needs to be batched in order for jax to play nicely
+    x0 = utils.get_romol_conf(mol)
+    src_atom = 1
+    dst_atom = 3
+    normal_restr_idxs, signs = chiral_utils.setup_chiral_bond_restraints(mol, x0, src_atom, dst_atom)
+
+    x0_inverted = [x0[0], x0[1], x0[2], x0[3], [0.07922, 0.06280, 0.0000]]
```
maxentile, line 122: Nit: this flips the conformer from
image
 to 
image
which does change the dihedral (0,1,3,4) dihedral angle from -pi to 0 (but also changes an angle)

Consider using

```python
from rdkit.Chem import rdMolTransforms

assert np.abs(rdMolTransforms.GetDihedralRad(mol.GetConformer(0), 0, 1, 3, 4)) == np.pi
rdMolTransforms.SetDihedralRad(mol.GetConformer(0), 0, 1, 3, 4, 0.0)
```
or similar instead?

tests/test_chiral_restraints.py
```diff
@@ -0,0 +1,526 @@
...
+    )
+    k = 1000.0
+
+    assert np.all(np.asarray(U_chiral_bond_batch(x0, normal_restr_idxs, k, signs)) == 0)
+    assert np.all(np.asarray(U_chiral_bond_batch(x0, inverted_restr_idxs, k, inverted_signs)) > 0)
+    assert np.all(np.asarray(U_chiral_bond_batch(x0_inverted, normal_restr_idxs, k, signs)) > 0)
+    assert np.all(np.asarray(U_chiral_bond_batch(x0_inverted, inverted_restr_idxs, k, inverted_signs)) == 0)
+
+
+def test_chiral_restraints_pyramidal():
```
maxentile, line 135: Add one-sentence description of what is being asserted

```suggestion
def test_chiral_restraints_pyramidal():
    """For ammonium, assert that:
    * without chiral restraints, up and down states are ~ equally sampled
    * with chiral restraints, ~ only the specified state is sampled"""
```

tests/test_chiral_restraints.py
```diff
@@ -0,0 +1,526 @@
...
+        return new_params, nb
+
+    def parameterize_nonbonded_pairlist(self, ff_q_params, ff_lj_params):
+        params, nb = topology.BaseTopology.parameterize_nonbonded_pairlist(self, ff_q_params, ff_lj_params)
+        charge_indices = jnp.index_exp[:, 0]
+        new_params = jnp.asarray(params).at[charge_indices].multiply(self.scale)
+        return new_params, nb
+
+
+def test_chiral_restraints_torsion():
```
maxentile, line 238: Add one-sentence description of what is being asserted
```suggestion
def test_chiral_restraints_torsion():
    """For a charge-scaled version of hydrogen peroxide, assert that:
    * without chiral bond restraints, cis/trans states are sampled ~ 30%/70%
    * with chiral bond restraints, ~ only specified state is sampled"""
```

tests/test_chiral_restraints.py
```diff
@@ -0,0 +1,526 @@
+import multiprocessing
+import os
+
+os.environ["XLA_FLAGS"] = "--xla_force_host_platform_device_count=" + str(multiprocessing.cpu_count())
```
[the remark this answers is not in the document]
proteneer: last time I checked, `jax.vmap` doesn't efficiently push through `jax.while_loop` (it will still run, but you won't get any speed-up, and the cpu usage is pegged to 1-core)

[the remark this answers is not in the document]
proteneer: Sorry didn't see "on the test-runner", I have no idea, but locally (on my mac and desktop), its a 10x speed-up.

tests/test_chiral_restraints.py
maxentile: (Alternately, could have RDKit generate conformers for 'Cl\C(F)=N/F'` vs. `'Cl/C(F)=N/F'`)

tests/test_chiral_restraints.py
proteneer: willfix

tests/test_chiral_restraints.py
proteneer: willfix

tests/test_chiral_restraints.py
proteneer: willfix

proteneer:
rebased and enabling auto-merge.

PR #753, opened 2022-06-09
Remove usage of matplotlib

mcwitt:
Matplotlib is imported in one place in timemachine: https://github.com/proteneer/timemachine/blob/d0e4787559382a804b07805dec254c8fa30c5ec6/timemachine/fe/atom_mapping.py#L3

We should ~either add it to `install_requires`, or~ remove the usage in this file if it's no longer relevant (there are other cases of commented-out usage in tests and `attic/`)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,5 +5,6 @@ scipy==1.6.0
 pymbar==3.0.5
 pyyaml==5.4.1
 networkx==2.5
+matplotlib==3.5.2
 importlib-resources==5.4.0
 typing-extensions==4.1.1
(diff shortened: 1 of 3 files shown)


maxentile:
>We should either add it to install_requires, or remove the usage in this file if it's no longer relevant (there are other cases of commented-out usage in tests and attic/)

I think this usage can be removed -- was added when I didn't have as clear a picture of what belongs in timemachine vs. elsewhere

mcwitt:
> I think this usage can be removed

Agreed, the usage actually looks pretty trivial (just a small wrapper around `plt.imshow`). Removed in d2341.

jkausrelay (approved):

badisa (approved):

PR #754, opened 2022-06-10
Utilities for find_chiral_atoms and find_chiral_bonds

proteneer:
This PR implements a SMARTS based way to identify chiral atoms and bonds, it's mostly a simplification of a small [comment](https://github.com/proteneer/timemachine/pull/746#discussion_r894019109) that @maxentile made yesterday. Note that chiral here simply means that there's an extremely high inversion barrier, so even a symmetric atom like CH4 would be considered "chiral", since one can't arbitrary swap the two hydrogens without overcoming an extremely high energy barrier (high enough to be deemed kinetically inaccessible in an experiment).

For any given pyramidal structure, eg. NH3, the following permutations would be okay, let (1,2,3) denote the hydrogen barriers:

(1,2,3)
(2,3,1)
(3,1,2)

But pair swaps would be illlegal:

(2,1,3)
(1,3,2)
(3,2,1)

Note that we can't simply inspect force constants, because most classical forcefields are generally not well parameterized with respect to the barrier heights (they only get the curvature of the minimas correct, not the height of the maximas). So we just use a couple of simple heuristics to catch cases like amides, ring bonds, sulfurs, phosphor, etc.

### DIFF
--- a/tests/test_chiral_utils.py
+++ b/tests/test_chiral_utils.py
@@ -1,0 +1,134 @@
+import numpy as np
+from rdkit import Chem
+from rdkit.Chem import AllChem
+
+from timemachine.fe import chiral_utils, utils
+from timemachine.potentials.chiral_restraints import U_chiral_atom_batch, U_chiral_bond_batch
+
+
+def test_setup_chiral_atom_restraints():
+    """On a methane conformer, assert that permuting coordinates or permuting restr_idxs
+    both independently toggle the chiral restraint"""
+    mol = Chem.MolFromMolBlock(
+        """
+  Mrv2202 06072215563D
+
+  5  4  0  0  0  0            999 V2000
+    0.0000    0.0000    0.0000 C   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.3633   -0.5138    0.8900 H   0  0  0  0  0  0  0  0  0  0  0  0
+    1.0900    0.0000    0.0000 H   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.3633    1.0277    0.0000 H   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.3633   -0.5138   -0.8900 H   0  0  0  0  0  0  0  0  0  0  0  0
+  1  2  1  0  0  0  0
+  1  3  1  0  0  0  0
+  1  4  1  0  0  0  0
+  1  5  1  0  0  0  0
+M  END
+$$$$""",
+        removeHs=False,
+    )
+
+    # needs to be batched in order for jax to play nicely
+    x0 = utils.get_romol_conf(mol)
+    normal_restr_idxs = chiral_utils.setup_chiral_atom_restraints(mol, x0, 0)
+
+    x0_inverted = x0[[0, 2, 1, 3, 4]]  # swap two atoms
+    inverted_restr_idxs = chiral_utils.setup_chiral_atom_restraints(mol, x0_inverted, 0)
+
+    # check the sign of the resulting idxs
+    k = 1000.0
+    assert np.all(np.asarray(U_chiral_atom_batch(x0, normal_restr_idxs, k)) == 0)
+    assert np.all(np.asarray(U_chiral_atom_batch(x0, inverted_restr_idxs, k)) > 0)
+    assert np.all(np.asarray(U_chiral_atom_batch(x0_inverted, normal_restr_idxs, k)) > 0)
+    assert np.all(np.asarray(U_chiral_atom_batch(x0_inverted, inverted_restr_idxs, k)) == 0)
+
+
+def test_setup_chiral_bond_restraints():
+    """On a 'Cl/C(F)=N/F' conformer, assert that flipping a dihedral angle or permuting restr_idxs
+    both independently toggle the chiral bond restraint"""
+
+    mol_cis = Chem.MolFromSmiles(r"Cl\C(F)=N/F")
+    mol_trans = Chem.MolFromSmiles(r"Cl\C(F)=N\F")
+
+    AllChem.EmbedMolecule(mol_cis)
+    AllChem.EmbedMolecule(mol_trans)
+
+    # needs to be batched in order for jax to play nicely
+    x0_cis = utils.get_romol_conf(mol_cis)
+    x0_trans = utils.get_romol_conf(mol_trans)
+    src_atom = 1
+    dst_atom = 3
+    normal_restr_idxs, signs = chiral_utils.setup_chiral_bond_restraints(mol_cis, x0_cis, src_atom, dst_atom)
+
+    inverted_restr_idxs, inverted_signs = chiral_utils.setup_chiral_bond_restraints(
+        mol_trans, x0_trans, src_atom, dst_atom
+    )
+    k = 1000.0
+
+    assert np.all(np.asarray(U_chiral_bond_batch(x0_cis, normal_restr_idxs, k, signs)) == 0)
+    assert np.all(np.asarray(U_chiral_bond_batch(x0_cis, inverted_restr_idxs, k, inverted_signs)) > 0)
+    assert np.all(np.asarray(U_chiral_bond_batch(x0_trans, normal_restr_idxs, k, signs)) > 0)
+    assert np.all(np.asarray(U_chiral_bond_batch(x0_trans, inverted_restr_idxs, k, inverted_signs)) == 0)
+
+
+def test_find_chiral_atoms():
+    # test that we can identify chiral atoms ub a couple of tetrahedral and pyramidal
+    # molecules.
+
+    mol = Chem.MolFromSmiles(r"FC(Cl)Br")
+    res = chiral_utils.find_chiral_atoms(mol)
+    assert res == set([1])
+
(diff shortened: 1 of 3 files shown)


jkausrelay (approved):

jkausrelay:
lgtm

tests/test_chiral_utils.py
```diff
@@ -0,0 +1,168 @@
...
+    )
+    k = 1000.0
+
+    assert np.all(np.asarray(U_chiral_bond_batch(x0_cis, normal_restr_idxs, k, signs)) == 0)
+    assert np.all(np.asarray(U_chiral_bond_batch(x0_cis, inverted_restr_idxs, k, inverted_signs)) > 0)
+    assert np.all(np.asarray(U_chiral_bond_batch(x0_trans, normal_restr_idxs, k, signs)) > 0)
+    assert np.all(np.asarray(U_chiral_bond_batch(x0_trans, inverted_restr_idxs, k, inverted_signs)) == 0)
+
+
+def test_find_chiral_atoms():
```
maxentile, line 74: In general, do we want to add one-sentence descriptions of what is being asserted?

tests/test_chiral_utils.py
```diff
@@ -0,0 +1,168 @@
...
+
+    mol = Chem.MolFromSmiles(r"FC(Cl)Br")
+    res = chiral_utils.find_chiral_atoms(mol)
+    assert res == set([1])
```
maxentile, line 78: This one relies on `mol.GetAtomWithIdx(1).GetNumImplicitHs() == 1`, would be slightly clearer if the hydrogen weren't implicit, since we generally will only be applying this method to rdmols with explicit hydrogens?

```suggestion
    mol = Chem.AddHs(Chem.MolFromSmiles(r"FC(Cl)Br"))
    res = chiral_utils.find_chiral_atoms(mol)
    assert res == set([1])
```

timemachine/fe/chiral_utils.py
```diff
@@ -55,3 +56,69 @@ def setup_chiral_bond_restraints(mol, conf, src_idx, dst_idx):
...
+    -------
+    set of 2-tuple
+        Chiral bonds
+
+    """
+
+    chiral_patterns = [
+        "[X2,X3:1]:[X2,X3:2]",  # any aromatic bond
+        "[X2,X3:1]=[X2,X3:2]",  # all double bonds with two or three neighbors,
+        "[NX3,NX2:1][CX3:2](=[OX1])",  # amide bond
```
maxentile, line 114: Curiosity [do not merge necessarily]: What amide bonds does the "or NX2" condition catch?

```suggestion
        "[NX3:1][CX3:2](=[OX1])",  # amide bond
```

timemachine/fe/chiral_utils.py
```diff
@@ -55,3 +56,69 @@ def setup_chiral_bond_restraints(mol, conf, src_idx, dst_idx):
...
+
+    Returns
+    -------
+    set of 2-tuple
+        Chiral bonds
+
+    """
+
+    chiral_patterns = [
+        "[X2,X3:1]:[X2,X3:2]",  # any aromatic bond
```
maxentile, line 112: Specification: I don't understand why this pattern should be considered chiral. (I don't understand the goal that's achieved by saying there are 6 chiral bonds in benzene.) Deserves more elaboration!

tests/test_chiral_utils.py
```diff
@@ -0,0 +1,168 @@
...
+            (10, 11),
+            (6, 11),
+            (13, 14),
+            (14, 15),
+            (15, 16),
+            (16, 17),
+            (17, 18),
+            (13, 18),
+        ]
+    )
```
maxentile, line 168: Specification: This seems surprising to me as desired behavior (catches only aromatic bonds) -- perhaps would be resolved by adding a definition in docstring `find_chiral_bonds`

maxentile (commented):
I have a blocking request for specification of `chiral_bonds` (and some small non-blocking implementation nits), but the chiral_atoms part looks good to me!

timemachine/fe/chiral_utils.py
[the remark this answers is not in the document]
proteneer: I think it should be safe, since I couldn't come up with any examples of X1 or X4 aromatic atoms. I didn't think too hard about this on the initial pass and just went with X2/X3

timemachine/fe/chiral_utils.py
```diff
@@ -55,3 +56,69 @@ def setup_chiral_bond_restraints(mol, conf, src_idx, dst_idx):
...
+        assert query_mol is not None
+        for match in mol.GetSubstructMatches(query_mol):
+            chiral_atoms.add(match[0])
+
+    return chiral_atoms
+
+
+def find_chiral_bonds(mol):
+    """
+    Find chiral bonds in a molecule:
```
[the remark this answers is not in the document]
proteneer: willfix

tests/test_chiral_utils.py
proteneer: yes! Sorry I keep forgetting to add these.

tests/test_chiral_utils.py
proteneer: Agreed - sorry I missed the implicit H here!

timemachine/fe/chiral_utils.py
proteneer: I was thinking about this from the geometric perspective, any planar bond group that has an extremely high rotational barrier (eg. amides). It seems like the aromatic ring bonds followed this rule of thumb as well. Unlike the chiral atom case, I guess the end-state for aromatic rings are probably well defined enough that even in the absence of chiral restraints they should adopt the correct geometry.

In any case, these restraints do not hurt (and are extremely fast to compute), and are similar to the butane case that it's probably fine to just leave them on?

maxentile: >I was thinking about this from the geometric perspective, any planar bond group that has an extremely high rotational barrier (eg. amides)

The motivation for amides makes perfect sense!

In general, if our MM energy function has multiple "wells", but we want to ensure we sample only one of them, this sort of restraint seems well-motivated.

(E.g. defining methane as "chiral" in your example -- hydrogen permutations would be iso-energetic, but (1) permutations are practically not sampled by MD for kinetic reasons, (2) we want to ensure we don't start sampling permutations unevenly if sampling becomes more thorough.)

>Unlike the chiral atom case, I guess the end-state for aromatic rings are probably well defined enough that even in the absence of chiral restraints they should adopt the correct geometry.

Yeah, for this reason the motivation for aromatic bonds is less obvious to me, since I don't see what the restraint is actually preventing. (H-C:C-H "inversions" in benzene would be extremely energetically unfavorable with or without the new restraints.)

>In any case, these restraints do not hurt (and are extremely fast to compute), and are similar to the butane case that it's probably fine to just leave them on?

Probably doesn't hurt! Would still be good to have a motivating example though, of some "inversion" that's energetically allowed in the absence of the aromatic bond restraint, that the restraint now rules out.

timemachine/fe/chiral_utils.py
proteneer: I might have been a bit too aggressive here, maybe something like `[H]\C(O)=N/C([H])=O`?

maxentile: Nice! Thanks for the example.

Seems reasonable to catch this one, but I don't know how edge-case-y this might be....

---------

fwiw, two observations about this molecule:

(1) rdkit doesn't draw this stereo-annotated molecule with wedge bonds,
(2) the default forcefield assigns ~2.5 + ~5 = ~7.5 kJ/mol torsion barrier to the C-N bond matched by this pattern. This is lower than the ~10 + ~3 = ~13kJ/mol barrier assigned to the C-O bond in this molecule (which is not marked as a chiral bond by this function).
image

proteneer: 1) i think wedge bond is only for tetrahedral geometries
2) the ff probably doesn't parameterize this correctly

tests/test_chiral_utils.py
proteneer: I'll remove this pattern for now - if we need to add it later on it's simple to do.

timemachine/fe/chiral_utils.py
proteneer: For the sake of simplicity - I'll remove this for now. We can add this later on if needed.

timemachine/fe/chiral_utils.py
maxentile: >i think wedge bond is only for tetrahedral geometries

Ahh, right!

maxentile (approved):
Looks good!

Some non-blocking comments outstanding (esp. adding a brief definition of "chiral bond" to the docstring of `find_chiral_bonds`), but removal of "aromatic bonds are chiral" from the implementation in https://github.com/proteneer/timemachine/pull/754/commits/72b4cd098f74cf31efca8f34545193d1cf993b8c sufficiently reduces surprise in this definition.

(In offline discussion, I don't think we've yet found any examples where treating aromatic bonds as "chirally restrained" would have had an effect. If we find a motivating example later, the definition can be expanded easily by adding `[*:1]:[*:2]` to the list of smarts patterns.)

PR #755, opened 2022-06-15
Replace print statement in U_easy with logging

mcwitt:
`VacuumState.U_easy` currently prints a list of disabled torsions every time it is called, which can lead to a lot of noise in scripts and notebooks.

There is a workaround when running from IPython (e.g. in a Jupyter notebook):

```python
with IPython.utils.io.capture_output(stdout=True, stderr=False):
    # code calling U_easy
```

This PR changes the `print` statement to a `DEBUG`-level log message, so that `turning off torsion` messages will not be printed to stdout unless the logging level is set to `DEBUG` or lower (e.g. with `logging.basicConfig(level=logging.DEBUG)` in application code).

### DIFF
--- a/timemachine/md/enhanced.py
+++ b/timemachine/md/enhanced.py
@@ -2,6 +2,7 @@
 
 # This file contains utility functions to generate samples in the gas-phase.
 
+import logging
 import multiprocessing
 import os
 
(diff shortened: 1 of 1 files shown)


proteneer:
totally reasonable

timemachine/md/enhanced.py
```diff
@@ -173,7 +176,7 @@ def U_easy(self, x):
         for idxs, params in zip(self.pt_potential.get_idxs(), self.proper_torsion_params):
             _, j, k, _ = idxs
             if (j, k) in rotatable_bonds:
-                print("turning off torsion", idxs)
+                logger.debug(f"turning off torsion %s", idxs)
```
proteneer, line 179: debug sounds like the right level to me as well

proteneer (approved):

PR #756, opened 2022-06-16
Simplified Single Topology

proteneer:
This PR is a slimmed down version of #746 since the chiral restraints PR #752 vastly simplifies the logic for chirality. We now only use bond and angle terms to setup orientational restraints for the dummy groups. In particular, dummy interactions now only are comprised of bond, angle, and improper terms. Propers and nonbonded interactions involving dummy atoms are fully disabled. 

This PR also handles many of the difficult ring bond breaking/forming cases in the `hif2a` set, though the intermediate stages will have to be optimized significantly. A set of nightly tests will be done in a separate PR later on for:

1) Ensuring end-state stability for many more transformations
2) Verifying factorizability of just the dummy-states
3) Testing different atom-mapping approaches


tests/test_chiral_utils.py
```diff
@@ -75,19 +75,19 @@ def test_find_chiral_atoms():
     # test that we can identify chiral atoms ub a couple of tetrahedral and pyramidal
     # molecules.
 
-    mol = Chem.MolFromSmiles(r"FC(Cl)Br")
+    mol = Chem.AddHs(Chem.MolFromSmiles(r"FC(Cl)Br"))
```
proteneer, line 78: I forgot these in a previous PR! These changes are clerical.

timemachine/fe/dummy.py
```diff
@@ -269,28 +268,7 @@ def identify_dummy_groups(bond_idxs, core):
...
-
-        anchor_kv = defaultdict(set)
-        for idx, anchor in enumerate(anchor_membership):
-            anchor_kv[anchor].add(dg[idx])
-
-        for v in anchor_kv.values():
-            dummy_groups.append(v)
-
-    return dummy_groups
+    return cc
```
proteneer, line 271: The code below was deleted because it's no longer relevant, and was just plain buggy.

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
...
+
+
+def get_romol_bonds(mol):
+    """
+    Return bond idxs given a mol. These are not canonicalized.
+    """
+    bond_list = []
+    for bond in mol.GetBonds():
+        bond_list.append([bond.GetBeginAtomIdx(), bond.GetEndAtomIdx()])
+    return bond_list
```
maxentile, line 30: Nit: PR adds two copies of this function -- prefer the one added to `fe/utils.py` https://github.com/proteneer/timemachine/blob/94c67b8cd679a347528cbef2fd03b2256cfeb7da/timemachine/fe/utils.py#L267-L274 ?

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
...
+    )
+
+
+def setup_end_state(ff, mol_a, mol_b, core, a_to_c, b_to_c):
+    """
+    Setup end-state for mol_a with dummy atoms of mol_b attached. The mapped indices will correspond
+    to the alchemical molecule with dummy atoms.
+
+    Parameters
+    ----------
```
maxentile, line 210: Nit: docstring -- missing parameter `ff`

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
...
+    core_a: list of int
+        Core atoms in mol_a
+
+    core_b: list of int
+        Core atoms in mol_b
+
+    Returns
+    -------
+    set, list
+        Returns a set of dummy_groups, and a listo 2-tuples (j,k). j is the root_anchor atom,
```
maxentile, line 331: nit: typo
```suggestion
        Returns a set of dummy_groups, and a list of 2-tuples (j,k). j is the root_anchor atom,
```

tests/test_single_topology_v3.py
```diff
@@ -0,0 +1,180 @@
...
+            pairs.append((mol_a, mol_b))
+
+    np.random.seed(2023)
+    np.random.shuffle(pairs)
+    ff = Forcefield.load_from_file("smirnoff_1_1_0_sc.py")
+
+    batch_distance_check = jax.vmap(get_max_distance)
+
+    # this has been tested for up to 50 random pairs
+    for mol_a, mol_b in pairs[:5]:
```
maxentile, line 157: Nice! This randomized stress-testing is good.

* Should we add a version of this test with a larger number of pairs to nightly tests?
* Is there a cheaper condition that can be tested for a large number of pairs? (E.g. omitting `simulate_system`, and just checking that system can be created & energy is finite on initial combined conf?)

tests/test_single_topology_v3.py
```diff
@@ -0,0 +1,180 @@
...
+
+        for system in systems:
+            U_fn = system.get_U_fn()
+            batch_U_fn = jax.vmap(U_fn)
+            frames = simulate_system(system.get_U_fn(), x0, num_samples=1000)
+            nrgs = batch_U_fn(frames)
+
+            assert np.all(np.isfinite(nrgs))
+            assert np.all(np.isfinite(frames))
+            distance_cutoff = 2.5  # in nanometers
```
maxentile, line 179: (Possible refinement for later: could tighten this `distance_cutoff` to `max([length(path) for path in shortest_paths(mol_graph)])` or similar)

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
...
+        res = []
+        for item in items:
+            res.append(recursive_map(item, mapping))
+        return tuple(res)
+    else:
+        return mapping[items]
+
+
+def find_nbs(mol, atom):
+    nbs = [a.GetIdx() for a in mol.GetAtomWithIdx(atom).GetNeighbors()]
```
maxentile, line 62: nit: annotation 

(return type differs from `get_neighbors`)

```suggestion
def find_nbs(mol, atom_idx: int) -> List[Int]:
    nbs = [a.GetIdx() for a in mol.GetAtomWithIdx(atom_idx).GetNeighbors()]
```

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
...
+        -------
+        Callable:
+            An energy function f: R^(Nx3) -> R^1
+
+        """
+        U0_fn = self.setup_end_state_src().get_U_fn()
+        U1_fn = self.setup_end_state_dst().get_U_fn()
+
+        def U_fn(x):
+            return (1 - lamb) * U0_fn(x) + lamb * U1_fn(x)
```
maxentile, line 537: Maybe better for a TODO note and later revisit: Wondering how best to refactor to expose more parameters of `U0_fn`, `U1_fn`...

timemachine/fe/dummy.py
jkausrelay: remove comment above too

timemachine/fe/system.py
```diff
@@ -54,6 +92,7 @@ def get_U_fn(self):
 
         def U_fn(x):
             Us_vdw, Us_coulomb = nbpl_U(x)
+            # (ytz): revert!
```
jkausrelay, line 95: revert!

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
...
+        Parameters
+        ----------
+        mol_a: ROMol
+            First ligand
+        mol_b: ROMol
+            Second ligand
+        core: np.array (C, 2)
+            Atom mapping from mol_a to to mol_b.
+        ff: ff.Forcefield
+            Forcefield to be used for parameterization.
```
maxentile, line 415: nit: parameter name in docstring != parameter name in signature

timemachine/fe/single_topology_v3.py
jkausrelay, line 30: remove duplicate

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
...
+        for a, b in core:
+            self.c_flags[a] = 0
+            self.b_to_c[b] = a
+
+        iota = self.mol_a.GetNumAtoms()
+        for b_idx, c_idx in enumerate(self.b_to_c):
+            if c_idx == -1:
+                self.b_to_c[b_idx] = iota
+                self.c_flags[iota] = 2
+                iota += 1
```
maxentile, line 448: Question: Should we warn or log if `c_idx != -1`?

maxentile (commented):
Looks good in initial exploration yesterday!

Mostly superficial comments in this review pass

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
...
+    nbor_anchor_atom: int
+        Another core atom connected to root_anchor_atom to build an angle restraint off of.
+
+    Returns
+    -------
+    (bonded_idxs, bonded_params)
+        Returns bonds, angles, and improper idxs and parameters.
+    """
+
+    # copy interactions that involve only root_anchor_atom
```
jkausrelay, line 152: nit: move comment to 165

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
...
+    bond_idxs,
+    bond_params,
+    angle_idxs,
+    angle_params,
+    improper_idxs,
+    improper_params,
+    dummy_group,
+    root_anchor_atom,
+    nbor_anchor_atom,
+):
```
jkausrelay, line 103: Is there a reason this is a separate function?

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
...
+            # This is a little arbitrary, but some choices are probably more efficient than others:
+            #
+            # hard     easy
+            # D0.D1    D0 D1
+            #    .     .  .
+            # C0-C1    C0-C1
+            #
+            # the lhs is more difficult because it has significantly more phase space that can be
+            # sampled than the fused case, but it's not super obvious how to best detect this.
+            # So instead, we will pick a random anchor atom. One possible solution later on is to
```
jkausrelay, line 368: nit: random -> arbitrary

also is it consistent between runs for the same pair of molecules?

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
...
+            # minimize the number of rotatable bonds?
+            warnings.warn(
+                f"Multiple root anchors {root_anchors} found for dummy group: {dg}, picking a random anchor.",
+                MultipleAnchorWarning,
+            )
+
+        jk = None
```
jkausrelay, line 375: nit: jk -> angle_jk to make it more clear what this is for

tests/test_single_topology_v3.py
```diff
@@ -0,0 +1,180 @@
...
+    dij = compute_distance_matrix(x0)
+    return jnp.amax(dij)
+
+
+def test_hif2a_end_state_stability():
+    """
+    Pick some random pairs from the hif2a set and ensure that they're numerically stable at the
+    end-states under some a standard distance based atom_mapping protocol.
+    """
+    suppl = Chem.SDMolSupplier("timemachine/testsystems/data/ligands_40.sdf", removeHs=False)
```
mcwitt, line 143: nit: better to use
```suggestion
    with resources.path("timemachine.testsystems.data", "ligands_40.sdf") as path_to_ligand:
        suppl = Chem.SDMolSupplier(str(path_to_ligand), removeHs=False)
```
to explicitly use the packaged test data ([example](https://github.com/mcwitt/timemachine/blob/45b973f312fedaa8fe4693b6c32bb481f8afbb40/tests/test_handlers.py#L685-L686))

tests/test_single_topology_v3.py
```diff
@@ -0,0 +1,180 @@
...
+    mols = [m for m in suppl]
+    pairs = []
+    for mol_a in mols:
+        for mol_b in mols:
+            pairs.append((mol_a, mol_b))
```
mcwitt, line 148: nit: slightly cleaner?
```suggestion
    pairs = [(mol_a, mol_b) for mol_a in mols for mol_b in mols]
```

tests/test_single_topology_v3.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def compute_distance_matrix(x0):
+    N = x0.shape[0]
+    D = x0.shape[1]
+    xi = x0.reshape((1, N, D))
+    xj = x0.reshape((N, 1, D))
+    rij = xi - xj
+    dij = jnp.linalg.norm(rij, axis=-1)
+    return dij
```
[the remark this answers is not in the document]
mcwitt: Could also use `cdist` from `scipy.spatial.distance` here (if the result doesn't need to be jax-transformable)

tests/test_single_topology_v3.py
```diff
@@ -0,0 +1,180 @@
...
+
+    # set [H] as the dummy group
+    all_idxs, _ = setup_dummy_interactions_from_ff(ff, mol, dummy_group=[12], root_anchor_atom=6, nbor_anchor_atom=None)
+    bond_idxs, angle_idxs, improper_idxs = all_idxs
+
+    assert set(bond_idxs) == set([(6, 12)])
+    assert set(angle_idxs) == set()
+    assert set(improper_idxs) == set()
+
+    # set [H] as the dummy group
```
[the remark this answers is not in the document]
proteneer: willfix inline

tests/test_single_topology_v3.py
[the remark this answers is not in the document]
proteneer: yes - will replace!

tests/test_single_topology_v3.py
proteneer: willfix inline

tests/test_single_topology_v3.py
proteneer: willfix

tests/test_single_topology_v3.py
proteneer: 1)  yes but there are other things i'd also like to add to the nightly
2) good idea, however, the MCS code is also surprisingly slow for some of these pairs, not just the MD. But the MCS could be cached, and we could probably verify end-states can have their energies be proper minimized for a larger set?

proteneer: 3. jit itself can be pretty slow (4-5 seconds per pair)

tests/test_single_topology_v3.py
proteneer: agreed - but saving this for later.

timemachine/fe/dummy.py
proteneer: willfix

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
+import warnings
```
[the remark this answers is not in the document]
proteneer: willfix  (in final pass)

timemachine/fe/single_topology_v3.py
proteneer: oops!

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
...
+    """
+    bond_list = []
+    for bond in mol.GetBonds():
+        bond_list.append([bond.GetBeginAtomIdx(), bond.GetEndAtomIdx()])
+    return bond_list
+
+
+def get_neighbors(bond_idxs, query):
+    """
+    Get neighbors of query atom query.
```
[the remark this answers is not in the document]
proteneer: willfix

timemachine/fe/single_topology_v3.py
proteneer: hm yeah - I don't think we have doubly connected atoms, so either should be fine.

timemachine/fe/single_topology_v3.py
proteneer: This may be useful later on if we want to `setup_dummy_interactions` with already parameterized terms - the `_ff` version is just a utility. It's also more clear to document the required terms here explicitly (eg. bond, angle, impropers only).

timemachine/fe/single_topology_v3.py
proteneer: sure

timemachine/fe/single_topology_v3.py
proteneer: pausing review here - will resume later tonight (note to self mostly)

tests/test_single_topology_v3.py
```diff
@@ -0,0 +1,180 @@
...
+    assert dgs == [{3}]
+    assert jks == [(2, 1)]
+
+    # angle should swap
+    core = np.array([[1, 2], [2, 0], [3, 1]])
+
+    with pytest.warns(CoreBondChangeWarning):
+        dgs, jks = single_topology_v3.find_dummy_groups_and_anchors(mol_a, mol_b, core[:, 0], core[:, 1])
+        assert dgs == [{3}]
+        assert jks == [(2, None)]
```
mcwitt, line 94: What does it mean when `k` is `None` here? (this case doesn't seem to be documented in `find_dummy_groups_and_anchors`)

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
...
+        mol_c_nbpl_idxs,
+        mol_a_nbpl.get_rescale_mask(),
+        mol_a_nbpl.get_beta(),
+        mol_a_nbpl.get_cutoff(),
+    ).bind(mol_c_nbpl_params)
+
+    return system.VacuumSystem(bond_potential, angle_potential, torsion_potential, nonbonded_potential)
+
+
+def find_dummy_groups_and_anchors(mol_a, mol_b, core_a, core_b):
```
mcwitt, line 309: Question: do `core_a` and `core_b` need to be the same length? If so, should we check for this explicitly?

tests/test_single_topology_v3.py
```diff
@@ -0,0 +1,180 @@
...
+    xi = x0.reshape((1, N, D))
+    xj = x0.reshape((N, 1, D))
+    rij = xi - xj
+    dij = jnp.linalg.norm(rij, axis=-1)
+    return dij
+
+
+def get_max_distance(x0):
+    dij = compute_distance_matrix(x0)
+    return jnp.amax(dij)
```
mcwitt, line 135: nit: maybe clearer to just inline this?

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
...
+
+    # dummy_group may be a set in certain cases, so sanity check.
+
+    assert len(dummy_group) == len(list(dummy_group))
+    dummy_group = list(dummy_group)
+    # dummy group and anchor
+
+    dga = dummy_group + [root_anchor_atom]
+    for idxs, params in zip(bond_idxs, bond_params):
+        if all([a in dga for a in idxs]):
```
mcwitt, line 168: nit: redundant brackets
```suggestion
        if all(a in dga for a in idxs):
```

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
...
+            dummy_bond_idxs.append(tuple([int(x) for x in idxs]))  # tuples are hashable etc.
```
mcwitt, line 169: nit: redundant brackets
```suggestion
            dummy_bond_idxs.append(tuple(int(x) for x in idxs))  # tuples are hashable etc.
```

timemachine/fe/single_topology_v3.py
proteneer: yes - will add an assertion, they also need to be unique as well

timemachine/fe/single_topology_v3.py
proteneer: willfix inline

timemachine/fe/single_topology_v3.py
proteneer: Hm this a tricky question to answer it turns out. There's no "random-ness" in the code itself, i.e. it will always break on the first valid restraint found. However, if a different atom ordering is provided (eg. input SDF has its nodes/edges permuted), then a different anchor may be found. So we guarantee that for a fixed `Chem.Mol` this will always return the same groups. But different permutations of the initial mol itself cannot be guaranteed (and this requires some kind of canonicalization to solve).

proteneer: I will add a test for this actually!

timemachine/fe/single_topology_v3.py
proteneer: willfix

timemachine/fe/single_topology_v3.py
proteneer: willfix

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
...
+    def get_num_atoms(self):
+        """
+        Get the total number of atoms in the alchemical hybrid.
+
+        Returns
+        -------
+        int
+            Total number of atoms.
+        """
+        return self.NC
```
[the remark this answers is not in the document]
proteneer: I'm just goin to remove this attribute directly and have it be computed on the fly

timemachine/fe/single_topology_v3.py
proteneer: willfix

timemachine/fe/system.py
```diff
@@ -1,11 +1,49 @@
...
+    xi = res.x.reshape(*shape)
+    return xi
+
+
+def simulate_system(U_fn, x0, num_samples=20000):
+    num_atoms = x0.shape[0]
+    x_min = minimize_scipy(x0, U_fn)
+    seed = 2023
+
+    num_workers = multiprocessing.cpu_count()
```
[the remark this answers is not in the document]
proteneer: sure

timemachine/fe/system.py
proteneer: lol - this was a spurious comment! willfix

timemachine/fe/single_topology_v3.py
proteneer: ah neat - didn't realize python `all` can do comprehensions magically (`np.all` can't!)

timemachine/fe/single_topology_v3.py
proteneer: edit - loop is still needed to setup `b_to_c`, we do not need to warn or log, it is expected that the `b_to_c` will already be mapped (namely, core atoms in `b`), as shown in the above loop. This code sets up the idxs in the combined molecule for the *dummy* atoms of `b`

tests/test_single_topology_v3.py
```diff
@@ -0,0 +1,180 @@
...
+    bond_idxs, angle_idxs, improper_idxs = all_idxs
+
+    assert set(bond_idxs) == set([(6, 12)])
+    assert set(angle_idxs) == set([(5, 6, 12)])
+    assert set(improper_idxs) == set()
+
+    with pytest.raises(single_topology_v3.MissingAngleError):
+        all_idxs, _ = setup_dummy_interactions_from_ff(
+            ff, mol, dummy_group=[12], root_anchor_atom=6, nbor_anchor_atom=4
+        )
```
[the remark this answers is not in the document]
proteneer: willfix

tests/test_single_topology_v3.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def test_find_dummy_groups_and_anchors():
+    """
+    Test that we can find the anchors and dummy groups when there's a single core anchor atom. When core bond
+    is broken, we should disable one of the angle atoms.
+    """
+    mol_a = Chem.MolFromSmiles("OCCC")
+    mol_b = Chem.MolFromSmiles("CCCF")
+    core = np.array([[1, 2], [2, 1], [3, 0]])
```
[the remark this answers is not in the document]
proteneer: (`map` reminds me of a `kv/dict`, which this isn't)

tests/test_single_topology_v3.py
proteneer: k being `None` means that there are no suitable angle terms we can add, this can arise when a core bond is broken.

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
...
+    """
+    Get neighbors of query atom query.
+    """
+    nbs = set()
+    for i, j in bond_idxs:
+        if i == query:
+            nbs.add(j)
+        if j == query:
+            nbs.add(i)
+    return nbs
```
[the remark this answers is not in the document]
proteneer: this adds the `other` atom not matching the `idx==query` - I can't tell if the comprehension is doing this?

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,539 @@
...
+            dummy_angle_idxs.append(tuple([int(x) for x in idxs]))
+            dummy_angle_params.append(params)
+    for idxs, params in zip(improper_idxs, improper_params):
+        if all([a in dga for a in idxs]):
+            dummy_improper_idxs.append(tuple([int(x) for x in idxs]))
+            dummy_improper_params.append(params)
+
+    # (ytz): copy interactions that involve nbor_anchor_atom, if not None
+    # this may be set to None
+    if nbor_anchor_atom is not None:
```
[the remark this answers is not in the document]
proteneer: goin to just gow ith `nbr` here. The canadian in me still thinks it should be `neighbour`

timemachine/fe/single_topology_v3.py
proteneer: fixed

timemachine/fe/single_topology_v3.py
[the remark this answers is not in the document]
mcwitt: Ah, sorry, I think I misread the code. Disregard my suggestion!

tests/test_single_topology_v3.py
proteneer: will inline

timemachine/fe/single_topology_v3.py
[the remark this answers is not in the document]
mcwitt: Actually, is `get_neighbors` unused? Was confused because the tests still passed with the suggested change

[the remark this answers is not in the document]
proteneer: hm good call - they were definitely unused!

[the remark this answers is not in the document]
proteneer: (will just delete these functions)

tests/test_single_topology_v3.py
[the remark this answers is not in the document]
mcwitt: I'd be happy with `core_pairs`

timemachine/fe/single_topology_v3.py
jkausrelay: @proteneer The explanation sounds good to me. Maybe add a summary to the comment (expected to give consistent results if the mol is exactly the same including atom ordering).

timemachine/fe/single_topology_v3.py
```diff
@@ -372,7 +348,8 @@ def find_dummy_groups_and_anchors(mol_a, mol_b, core_a, core_b):
                 MultipleAnchorWarning,
             )
 
-        jk = None
+        # (i,j,k) where is a dummy, j is anchor, and k is the angle anchor
```
jkausrelay, line 351: `where i is a dummy` ?

timemachine/fe/single_topology_v3.py
jkausrelay: sgtm

timemachine/fe/single_topology_v3.py
proteneer: willfix

proteneer:
Rebased from master after merging in #758 - now the end-states have chiral restraints. Anything else blocking approval?

jkausrelay (approved):

timemachine/fe/single_topology_v3.py
```diff
@@ -0,0 +1,533 @@
...
+        Parameters
+        ----------
+        mol_a: ROMol
+            First ligand
+
+        mol_b: ROMol
+            Second ligand
+
+        core: np.array (C, 2)
+            Atom mapping from mol_a to to mol_b.
```
maxentile, line 414: typo

tests/test_single_topology_v3.py
```diff
@@ -0,0 +1,210 @@
...
+
+def test_phenol():
+    """
+    Test that dummy interactions are setup correctly for a phenol. We want to check that bonds and angles
+    are present when either a single root anchor is provided, or when a root anchor and a neighbor anchor is provided.
+    """
+    mol = Chem.AddHs(Chem.MolFromSmiles("c1ccccc1O"))
+
+    # do not allow H to map to O
+    ff = Forcefield.load_from_file("smirnoff_1_1_0_sc.py")
```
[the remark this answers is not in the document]
proteneer: will fix - choice of ff is irrelvant

tests/test_single_topology_v3.py
```diff
@@ -0,0 +1,210 @@
...
+    dgs, jks = single_topology_v3.find_dummy_groups_and_anchors(mol_a, mol_b, core_pairs[:, 0], core_pairs[:, 1])
+    assert dgs == [{0}]
+    assert jks == [(1, 2)] or jks == [(1, 3)]
+
+    dgs_zero, jks_zero = single_topology_v3.find_dummy_groups_and_anchors(
+        mol_a, mol_b, core_pairs[:, 0], core_pairs[:, 1]
+    )
+    for idx in range(100):
+        np.random.seed(idx)
+        dgs, jks = single_topology_v3.find_dummy_groups_and_anchors(mol_a, mol_b, core_pairs[:, 0], core_pairs[:, 1])
```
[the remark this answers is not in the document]
proteneer: This was a sanity check that @jkausrelay raised, where since the choice of the root anchor and angle anchor can be arbitrary, we want to make sure we're not using a `np.random.choice()` to pick the atoms. The current behavior is non-random, it just breaks on the first valid anchor found.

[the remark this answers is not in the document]
proteneer: Adding an invariance check to core_pairs would be useful though! I can include that here.

tests/test_single_topology_v3.py
```diff
@@ -0,0 +1,210 @@
...
+    with pytest.warns(MultipleAnchorWarning):
+        dgs, jks = single_topology_v3.find_dummy_groups_and_anchors(mol_a, mol_b, core_a, core_b)
+        assert dgs == [{0}]
+        assert jks == [(1, 2)]
+
+
+def test_hif2a_end_state_stability():
+    """
+    Pick some random pairs from the hif2a set and ensure that they're numerically stable at the
+    end-states under some a standard distance based atom_mapping protocol.
```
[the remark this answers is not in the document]
proteneer: willfix

timemachine/lib/potentials.py
```diff
@@ -427,6 +427,20 @@ def get_cutoff(self):
         return self.args[3]
 
 
+class NonbondedPairListPrecomputed(NonbondedCustomOpWrapper):
```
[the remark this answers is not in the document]
proteneer: Sure. Note that this is placeholder still - the C++ equivalent for this is not done. I was going add this in a later pass. There's some nuance here I should probably document - we can't actually use this class for the "cancellation of NaNs", since the floating point operations will be different from NonbondedAllPairs

tests/test_single_topology_v3.py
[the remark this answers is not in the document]
proteneer: (note this comment can be just removed, there's just pair mapping consideration here, just a single molecule).

tests/test_single_topology_v3.py
```diff
@@ -0,0 +1,210 @@
...
+        assert dgs == [{0}]
+        assert jks == [(1, 2)] or jks == [(2, 1)]
+
+    # test determinism, should be robust against seeds
+    dgs_zero, jks_zero = single_topology_v3.find_dummy_groups_and_anchors(
+        mol_a, mol_b, core_pairs[:, 0], core_pairs[:, 1]
+    )
+    for idx in range(100):
+        np.random.seed(idx)
+        dgs, jks = single_topology_v3.find_dummy_groups_and_anchors(mol_a, mol_b, core_pairs[:, 0], core_pairs[:, 1])
```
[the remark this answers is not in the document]
proteneer: 2) is already satisified, but as mentioned above, I will add a test!

timemachine/fe/single_topology_v3.py
proteneer: willfix

timemachine/potentials/nonbonded.py
```diff
@@ -308,9 +308,40 @@ def apply_cutoff(x):
...
-        vdW = jnp.where(rescale_vdW != 0, rescale_vdW * vdW, 0)
+        vdW = jnp.where(rescale_vdW != 0, vdW * rescale_vdW, 0)
         rescale_electrostatics = rescale_mask[:, 0]
-        electrostatics = jnp.where(rescale_electrostatics != 0, rescale_electrostatics * electrostatics, 0)
+        electrostatics = jnp.where(rescale_electrostatics != 0, electrostatics * rescale_electrostatics, 0)
+
+    return vdW, electrostatics
+
+
+def nonbonded_v3_on_precomputed_pairs(conf, params, box, pairs, beta: float, cutoff: Optional[float] = None):
```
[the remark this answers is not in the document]
proteneer: willfix

tests/test_single_topology_v3.py
```diff
@@ -0,0 +1,217 @@
...
+    assert jks == [(1, 2)] or jks == [(1, 3)]
+
+    dgs_zero, jks_zero = single_topology_v3.find_dummy_groups_and_anchors(
+        mol_a, mol_b, core_pairs[:, 0], core_pairs[:, 1]
+    )
+
+    # this code should be invariant to different random seeds and different ordering of core pairs
+    for idx in range(100):
+        np.random.seed(idx)
+        core_pairs_shuffle = np.array(core_pairs)
```
maxentile, line 115: (Nit: don't forget to shuffle these if intended! Currently may just create a copy)

```suggestion
        np.random.seed(idx)
        core_pairs_shuffle = np.array(core_pairs)
        np.random.shuffle(core_pairs_shuffle)
```

mcwitt (approved):

PR #762, opened 2022-06-23
Canonicalize valence indices

proteneer:
This PR is a sanity PR to ensure that the combined alchemical molecule has valence idxs that are properly canonicalized. Bond, angle, torsion, nonbonded, chiral_bond, are straight forward. chiral_atom has a special symmetry check added.

### DIFF
--- a/timemachine/fe/single_topology_v3.py
+++ b/timemachine/fe/single_topology_v3.py
@@ -177,7 +177,7 @@ def setup_dummy_interactions(
     )
 
 
-def canonicalize_improper_idxs(idxs):
+def canonicalize_improper_idxs(idxs) -> Tuple[int, int, int, int]:
     """
     Canonicalize an improper_idx while being symmetry aware.
 
--- a/timemachine/fe/single_topology_v3.py
+++ b/timemachine/fe/single_topology_v3.py
@@ -177,6 +177,54 @@ def setup_dummy_interactions(
     )
 
 
+def canonicalize_improper_idxs(idxs):
+    """
+    Canonicalize an improper_idx while being symmetry aware.
+
+    Given idxs (i,j,k,l), where i is the center, and (j,k,l) are neighbors:
+
+    0) Canonicalize the (j,k,l) into (jj,kk,ll) by sorting
+    1) Generate clockwise rotations of (j,k,l)
+    2) Generate counter clockwise rotations of (j,k,l)
+    3) We now can sort 1) and 2) and assign a mapping
+
+    If the (j,k,l) is in the cw rotation ordered set, we're done. Otherwise it must
+    be in the ccw ordered set. We look up the corresponding idx in the cw set.
+
+    This does not do idxs[0] < idxs[-1] canonicalization.
+    """
+    i, j, k, l = idxs
+
+    # i is the center
+    # generate lexical order
+    key = (j, k, l)
+
+    jj, kk, ll = sorted(key)
+
+    # generate clockwise permutations
+    cw_jkl = (jj, kk, ll)
+    cw_klj = (kk, ll, jj)
+    cw_ljk = (ll, jj, kk)
+    cw_items = sorted([cw_jkl, cw_klj, cw_ljk])
+
+    if key in cw_items:
+        return (i, j, k, l)
+
+    # generate counter clockwise permutations
+    ccw_kjl = (kk, jj, ll)
+    ccw_jlk = (jj, ll, kk)
+    ccw_lkj = (ll, kk, jj)
+    ccw_items = sorted([ccw_kjl, ccw_jlk, ccw_lkj])
+
+    assert key in ccw_items
+
+    for idx, cw_item in enumerate(ccw_items):
+        if cw_item == key:
+            break
+
+    return (i, *cw_items[idx])
+
+
 def setup_end_state(ff, mol_a, mol_b, core, a_to_c, b_to_c):
     """
     Setup end-state for mol_a with dummy atoms of mol_b attached. The mapped indices will correspond
@@ -266,6 +314,17 @@ def setup_end_state(ff, mol_a, mol_b, core, a_to_c, b_to_c):
     mol_c_improper_idxs = mol_a_improper_idxs + all_dummy_improper_idxs
     mol_c_improper_params = mol_a_improper_params + all_dummy_improper_params
 
+    # canonicalize improper with cw/ccw check
+    mol_c_improper_idxs = tuple([canonicalize_improper_idxs(idxs) for idxs in mol_c_improper_idxs])
+
+    # check that the improper idxs are canonical
+    def assert_improper_idxs_are_canonical(all_idxs):
+        for _, j, k, l in all_idxs:
+            jj, kk, ll = sorted((j, k, l))
+            assert (jj, kk, ll) == (j, k, l) or (kk, ll, jj) == (j, k, l) or (ll, jj, kk) == (j, k, l)
+
+    assert_improper_idxs_are_canonical(mol_c_improper_idxs)
+
     # combine proper + improper
     mol_c_torsion_idxs = mol_c_proper_idxs + mol_c_improper_idxs
     mol_c_torsion_params = mol_c_proper_params + mol_c_improper_params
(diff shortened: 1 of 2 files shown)


proteneer:
Context: canonicalization is needed to ensure that we can "align" the indices when doing parameter interpolation (follow-up PR).

timemachine/fe/single_topology_v3.py
```diff
@@ -269,16 +270,32 @@ def setup_end_state(ff, mol_a, mol_b, core, a_to_c, b_to_c):
...
     nonbonded_potential = mol_a_nbpl.bind(mol_a_nbpl_params)
 
-    chiral_atom_idxs = np.array(mol_a_chiral_atom_idxs, dtype=np.int32).reshape((-1, 4))
-    chiral_bond_idxs = np.array(mol_a_chiral_bond_idxs, dtype=np.int32).reshape((-1, 4))
+    # chiral atoms need special code for canonicalization, since triple product is invariant
+    # under rotational symmetry (but not something like swap symmetry)
+    canon_chiral_atom_idxs = []
+    for i, j, k, l in mol_a_chiral_atom_idxs:
+        jj, kk, ll = min([(j, k, l), (l, j, k), (k, l, j)])
+        canon_chiral_atom_idxs.append((i, jj, kk, ll))
```
jkausrelay, line 294: Add as a function to dummy?

jkausrelay:
/@proteneer  Can you confirm this doesn't affect the parameter values (i.e. sign)?

timemachine/fe/single_topology_v3.py
proteneer: I can't do this on the dummy level - the canonicalization is with respect to the *mapped* idxs in the *combined* molecule, i.e. only after the `recursive_map` has been called

jkausrelay: Ah, makes sense

proteneer:
`sign` is specific to the chiral bond restraint - the equivalence classes for the symmetry w.r.t. the `sign` is as follows:

`{i,j,k,l,s}, {l,k,j,i,s}` belong to the same class
`{i,j,k,l,-s}, {l,k,j,i,-s}` belong to the same class

jkausrelay (approved):

timemachine/fe/single_topology_v3.py
```diff
@@ -269,16 +270,32 @@ def setup_end_state(ff, mol_a, mol_b, core, a_to_c, b_to_c):
     mol_c_torsion_idxs = mol_c_proper_idxs + mol_c_improper_idxs
     mol_c_torsion_params = mol_c_proper_params + mol_c_improper_params
 
-    bond_potential = potentials.HarmonicBond(mol_c_bond_idxs).bind(mol_c_bond_params)
-    angle_potential = potentials.HarmonicAngle(mol_c_angle_idxs).bind(mol_c_angle_params)
-    torsion_potential = potentials.PeriodicTorsion(mol_c_torsion_idxs).bind(mol_c_torsion_params)
+    # canonicalize bonds
```
maxentile, line 273:
```suggestion
```

(or add `# canonicalize angles`, `# canonicalize torsions` etc. comments to subsequent blocks)

timemachine/fe/single_topology_v3.py
```diff
@@ -269,16 +270,32 @@ def setup_end_state(ff, mol_a, mol_b, core, a_to_c, b_to_c):
...
+    mol_c_bond_idxs_canon = [canonicalize_bond(idxs) for idxs in mol_c_bond_idxs]
+    bond_potential = potentials.HarmonicBond(mol_c_bond_idxs_canon).bind(mol_c_bond_params)
+
+    mol_c_angle_idxs_canon = [canonicalize_bond(idxs) for idxs in mol_c_angle_idxs]
+    angle_potential = potentials.HarmonicAngle(mol_c_angle_idxs_canon).bind(mol_c_angle_params)
+
+    mol_c_torsion_idxs_canon = [canonicalize_bond(idxs) for idxs in mol_c_torsion_idxs]
+    torsion_potential = potentials.PeriodicTorsion(mol_c_torsion_idxs_canon).bind(mol_c_torsion_params)
```
maxentile, line 281: Point of uncertainty: Should proper and improper torsion idxs be canonicalized differently?

(Consider if mol_a has an improper (0, 1, 2, 3), and mol_b has an improper (3, 1, 2, 0) -- `canonicalize_bond` will map these to (0, 1, 2, 3) and (0, 2, 1, 3), i.e. not be able to recognize these as describing the same set of neighbors {0, 2, 3} around the same central atom 1...)

maxentile (approved):

timemachine/fe/single_topology_v3.py
proteneer: willfix

timemachine/fe/single_topology_v3.py
maxentile: ^ Based on offline discussion with @proteneer -- there is a fix to apply here, but it's more involved than I was originally thinking. Also note the example I provided contains an error (assumed central atom is in tuple[1], but it should be tuple[0]).

proteneer:
@maxentile I've added a new canonicalization scheme that operates on just a single `(i,j,k,l)` improper

timemachine/fe/single_topology_v3.py
```diff
@@ -177,10 +177,59 @@ def setup_dummy_interactions(
...
 
+def canonicalize_improper_idxs(idxs):
+    """
+    Canonicalize an improper_idx while being symmetry aware.
+
+    Given idxs (i,j,k,l), where i is the center, and (j,k,l) are neighbors:
+
+    0) Canonicalize the (j,k,l) into (jj,kk,ll) by sorting
+    1) Generate clockwise rotations of (j,k,l)
+    2) Generate counter clockwise rotations of (j,k,l)
```
maxentile, line 188:
```suggestion
    1) Generate clockwise rotations of (jj,kk,ll)
    2) Generate counter clockwise rotations of (jj,kk,ll)
```
(otherwise step (0) appears unreferenced)

timemachine/fe/single_topology_v3.py
```diff
@@ -177,10 +177,59 @@ def setup_dummy_interactions(
     )
 
 
+def canonicalize_improper_idxs(idxs):
```
maxentile, line 180: Nit: annotation
```suggestion
def canonicalize_improper_idxs(idxs) -> Tuple[int, int, int, int]:
```
(since other canonicalize function returns length-4 sequence of type `type(idxs)` and this discards `type(idxs)` and returns 4-tuple)

timemachine/fe/single_topology_v3.py
```diff
@@ -177,10 +177,59 @@ def setup_dummy_interactions(
...
+    cw_items = sorted([cw_jkl, cw_klj, cw_ljk])
+
+    if key in cw_items:
+        return (i, j, k, l)
+
+    # generate counter clockwise permutations
+    ccw_kjl = (kk, jj, ll)
+    ccw_jlk = (jj, ll, kk)
+    ccw_lkj = (ll, kk, jj)
+    ccw_items = sorted([ccw_kjl, ccw_jlk, ccw_lkj])
```
maxentile, line 217: Might be slightly cleaner to refactor this using two operations `swap(i, j, k) -> (j, i, k)`, `rotate(i, j, k) -> (j, k, i)`, rather than writing out all the permuted idxs?

```suggestion
    def rotate(i, j, k):
        return (j, k, i)

    def generate_rotations(i, j, k):
        orig = (i, j, k)
        rot1 = rotate(*orig)
        rot2 = rotate(*rot1)
        rotations = [orig, rot1, rot2]
        return sorted(rotations)

    cw_items = generate_rotations(i, j, k)

    if key in cw_items:
        return (i, j, k, l)

    ccw_items = generate_rotations(j, i, k)
```

(on the fence about this suggestion)

maxentile (approved):
The improved handling of improper idxs looks right, but is tricky!

timemachine/fe/single_topology_v3.py
proteneer: For a simple case like this I prefer explicitly listing out the elements of the cw/ccw idxs, since the suggested version requires me to actually mentally work out what the elements are (even though its more semantically precise). I'll add a comment detailling the steps though.

PR #763, opened 2022-06-24
Parameter Interpolatio

proteneer:
This PR sets up parameter interpolation for bond, angle, torsion, nonbonded, chiral_atom/bond terms. It roughly operates in two passes:

1) For a given potential, define its idxs and parameters at the end-states as `idxs0, p0` and `idxs1, p1`
2) `idxs0` and `idxs1` are aligned and deduplicated to generate a list of tuples `idxs_c, (p0, p1)`
3) A simple linear interpolation scheme `p_i(lambda)=(1-lambda)*p0 + lambda*p1` is implemented, but @mcwitt will be optimizing this extensively to be more efficient.

There were lots of tricky cases to cover:

1) for bonded terms that are being turned off,  equilibrium bond lengths, angles, and phases are *not* modified. only the force constants are scaled down.
2) for nonbonded terms that are being turned off, `q_ij`, `sig_ij`, `eps_ij` are *all* scaled down.
3) for chiral bond restraints, the deduplication uses the `sign` information in addition to the `idxs` 
4) for torsions, the deduplication uses the `period` information in addition to the `idxs`

Not done yet:

documentation on tests some of the functions - trying to figure out what level of detail I should add to them.

### DIFF
--- a/tests/test_interpolate.py
+++ b/tests/test_interpolate.py
@@ -1,0 +1,223 @@
+from jax.config import config
+
+config.update("jax_enable_x64", True)
+
+from importlib import resources
+
+import numpy as np
+from rdkit import Chem
+
+from timemachine.fe import atom_mapping, interpolate, single_topology_v3
+from timemachine.fe.utils import get_romol_conf
+from timemachine.ff import Forcefield
+
+
+def test_align_harmonic_bond():
+    """
+    Test that we can align idxs and parameters correctly for harmonic bonds.
+    We expect that decoupled terms have their force constants turned set to zero,
+    while maintaining the same equilibrium bond lengths.
+    """
+    a, b, c, d, e, f, g, h, i, j = np.random.rand(10)
+
+    # tbd: what do we do if there are repeats?
+    # merge repeats into a single term first?
+    src_idxs = [(4, 9), (3, 4)]
+    src_params = [(a, b), (c, d)]
+
+    dst_idxs = [(3, 4), (5, 9), (2, 3)]
+    dst_params = [(e, f), (g, h), (i, j)]
+
+    test_kv = interpolate.align_harmonic_bond_idxs_and_params(src_idxs, src_params, dst_idxs, dst_params)
+
+    ref_kv = {
+        ((4, 9), (a, b), (0, b)),
+        ((3, 4), (c, d), (e, f)),
+        ((5, 9), (0, h), (g, h)),
+        ((2, 3), (0, j), (i, j)),
+    }
+
+    assert test_kv == ref_kv
+
+
+def test_align_harmonic_angle():
+    """
+    Test that we can align idxs and parameters correctly for harmonic angles.
+    We expect that decoupled terms have their force constants turned set to zero,
+    while maintaining the same equilibrium angles.
+    """
+    a, b, c, d, e, f, g, h, i, j = np.random.rand(10)
+
+    # tbd: what do we do if there are repeats?
+    # merge repeats into a single term first?
+    src_idxs = [(4, 9, 5), (3, 4, 6)]
+    src_params = [(a, b), (c, d)]
+
+    dst_idxs = [(3, 4, 6), (4, 5, 9), (2, 3, 4)]
+    dst_params = [(e, f), (g, h), (i, j)]
+
+    test_kv = interpolate.align_harmonic_bond_idxs_and_params(src_idxs, src_params, dst_idxs, dst_params)
+
+    ref_kv = {
+        ((4, 9, 5), (a, b), (0, b)),
+        ((3, 4, 6), (c, d), (e, f)),
+        ((4, 5, 9), (0, h), (g, h)),
+        ((2, 3, 4), (0, j), (i, j)),
+    }
+
+    assert test_kv == ref_kv
+
+
+def test_align_torsion():
+    """
+    Test that we can align idxs and parameters correctly for periodic torsions.
+    Periodic torsions differ from bonds and angles in that their uniqueness is
+    also determined by the "period", which is currently encoded as one of the parameters.
+
+    We expect that decoupled terms have their force constants turned set to zero,
+    while maintaining the same equilibrium angles for the *same* period.
+    """
+    a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p = np.random.rand(16)
+
+    # tbd: what do we do if there are repeats?
+    # merge repeats into a single term first?
+    src_idxs = [(2, 3, 9, 4), (2, 1, 4, 3), (0, 1, 4, 2), (0, 1, 4, 2)]
+    src_params = [(a, b, 2), (c, d, 1), (e, f, 3), (g, h, 1)]
+
+    dst_idxs = [(2, 3, 9, 4), (2, 3, 9, 4), (0, 1, 4, 2), (3, 0, 2, 6)]
+    dst_params = [(i, j, 2), (k, l, 1), (m, n, 3), (o, p, 4)]
+
+    test_kv = interpolate.align_torsion_idxs_and_params(src_idxs, src_params, dst_idxs, dst_params)
+
+    ref_kv = {
+        ((2, 3, 9, 4), (a, b, 2), (i, j, 2)),
+        ((2, 1, 4, 3), (c, d, 1), (0, d, 1)),
+        ((0, 1, 4, 2), (e, f, 3), (m, n, 3)),
+        ((0, 1, 4, 2), (g, h, 1), (0, h, 1)),
+        ((2, 3, 9, 4), (0, l, 1), (k, l, 1)),
+        ((3, 0, 2, 6), (0, p, 4), (o, p, 4)),
+    }
+
+    assert test_kv == ref_kv
+
+
+def test_align_nonbonded():
+    """
+    Test that we can align idxs and parameters for nonbonded forces. For decoupled
+    nonbonded interactions, we should set q_ij, s_ij, and e_ij all to zero.
+    """
+    a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p, q, r = np.random.rand(18)
+
+    src_idxs = [(0, 3), (0, 2), (2, 3), (2, 4)]
+    src_params = [(a, b, c), (d, e, f), (g, h, i), (j, k, l)]
+    dst_idxs = [(0, 2), (4, 5)]
+    dst_params = [(m, n, o), (p, q, r)]
+
+    test_kv = interpolate.align_nonbonded_idxs_and_params(src_idxs, src_params, dst_idxs, dst_params)
+
+    ref_kv = {
+        ((0, 3), (a, b, c), (0, 0, 0)),
+        ((0, 2), (d, e, f), (m, n, o)),
+        ((2, 3), (g, h, i), (0, 0, 0)),
+        ((2, 4), (j, k, l), (0, 0, 0)),
+        ((4, 5), (0, 0, 0), (p, q, r)),
+    }
+
+    assert test_kv == ref_kv
+
+
+def test_align_chiral_atoms():
+    """
+    Test that we can align idxs and parameters for chiral atom restraints. We should
+    expect that force constants are set to zero.
+    """
+    a, b, c, d, e = np.random.rand(5)
+
+    src_idxs = [(0, 3, 4, 5), (0, 4, 3, 5), (4, 3, 5, 6)]
+    src_params = [a, b, c]
+    dst_idxs = [(0, 4, 3, 5), (4, 3, 5, 7)]
+    dst_params = [d, e]
+
+    test_kv = interpolate.align_chiral_atom_idxs_and_params(src_idxs, src_params, dst_idxs, dst_params)
+
+    ref_kv = {
+        ((0, 3, 4, 5), a, 0),
+        ((0, 4, 3, 5), b, d),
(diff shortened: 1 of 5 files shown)


timemachine/fe/single_topology_v3.py
```diff
@@ -612,3 +612,95 @@ def U_fn(x):
...
+
+        return potentials.ChiralBondRestraint(chiral_bond_idxs, chiral_bond_signs).bind(chiral_bond_params)
+
+    def setup_intermediate_state(self, lamb):
+        """
+        Setup intermediate states at some value of lambda.
+        """
+        src_system = self.setup_end_state_src()
+        dst_system = self.setup_end_state_dst()
+        interpolate_fn = interpolate.linear_interpolation
```
proteneer, line 676: @mcwitt this is probably where you will supply more efficient interpolation schemes later on.

timemachine/fe/single_topology_v3.py
```diff
@@ -612,3 +612,95 @@ def U_fn(x):
...
+        bond = self._setup_intermediate_bonded_term(
+            src_system.bond, dst_system.bond, lamb, interpolate.align_harmonic_bond_idxs_and_params, interpolate_fn
+        )
+        angle = self._setup_intermediate_bonded_term(
+            src_system.angle, dst_system.angle, lamb, interpolate.align_harmonic_angle_idxs_and_params, interpolate_fn
+        )
+        torsion = self._setup_intermediate_bonded_term(
+            src_system.torsion, dst_system.torsion, lamb, interpolate.align_torsion_idxs_and_params, interpolate_fn
+        )
+        nonbonded = self._setup_intermediate_bonded_term(
```
proteneer, line 688: @mcwitt you probably want to modify this to do some sort of 4D decoupling scheme (possibly in addition to some kind of parameter interpolation scheme).

tests/test_interpolate_fe.py
```diff
@@ -0,0 +1,104 @@
...
+            fwd_delta_u = beta * (cur_U_fn(prev_frames) - prev_U_fn(prev_frames))
+            rev_delta_u = beta * (prev_U_fn(cur_frames) - cur_U_fn(cur_frames))
+
+            dG_exact, exact_bar_err = pymbar.BAR(fwd_delta_u, rev_delta_u)
+            dG_exact /= beta
+            exact_bar_err /= beta
+
+            print(
+                f"BAR: lambda {lambda_schedule[lambda_idx-1]:.3f} -> {lambda_schedule[lambda_idx]:.3f} dG: {dG_exact:.3f} dG_err: {exact_bar_err:.3f}"
+            )
```
proteneer, line 104: note: this code doens't actually test anything - not sure what quantities would be good here... `exact_bar_err not None`?

tests/test_interpolate.py
```diff
@@ -0,0 +1,227 @@
...
+
+from timemachine.fe import atom_mapping, interpolate, single_topology_v3
+from timemachine.fe.utils import get_romol_conf
+from timemachine.ff import Forcefield
+
+
+def test_align_harmonic_bond():
+    """
+    Test that we can align idxs and parameters correctly for harmonic bonds.
+    We expect that decoupled terms have their force constants turned set to zero,
```
mcwitt, line 18: nit: typo
```suggestion
    We expect that decoupled terms have their force constants set to zero,
```

tests/test_interpolate.py
```diff
@@ -0,0 +1,227 @@
...
+    while maintaining the same equilibrium bond lengths.
+    """
+    a, b, c, d, e, f, g, h, i, j = np.random.rand(10)
+
+    # tbd: what do we do if there are repeats?
+    # merge repeats into a single term first?
```
mcwitt, line 24: For now, would it make sense to raise an error if there are repeats? (to prevent any potentially surprising behavior)

tests/test_interpolate.py
```diff
@@ -0,0 +1,227 @@
...
+        ((4, 3, 5, 7), 0, e),
+    }
+
+    assert test_kv == ref_kv
+
+
+def test_align_chiral_bonds():
+    """
+    Similar to the chiral_atoms test, except that we check to see if deduplication
+    is using the sign information correctly.
```
mcwitt, line 156: nit: might be explicit what "correctly" means here (e.g. as part of the alignment key)

tests/test_interpolate_fe.py
mcwitt: maybe `np.isfinite` on `dG_exact` and `exact_bar_err`?

timemachine/fe/interpolate.py
```diff
@@ -0,0 +1,217 @@
...
+    src_kv = _add_dst_to_src_bond_or_angle(src_idxs, src_params, dst_idxs, dst_params)
+    dst_kv = _add_dst_to_src_bond_or_angle(dst_idxs, dst_params, src_idxs, src_params)
+    assert src_kv.keys() == dst_kv.keys()
+
+    res = set()
+
+    for k in src_kv.keys():
+        res.add((tuple(k), src_kv[k], dst_kv[k]))
+
+    return res
```
mcwitt, line 57: nit: I think this could be simplified a bit (and remove the need for the helper function) as something like:

```python
def align_harmonic_bond_or_angle_idxs_and_params(src_idxs, src_params, dst_idxs, dst_params):

    src_kv = dict(zip(src_idxs, src_params))
    dst_kv = dict(zip(dst_idxs, dst_params))

    aligned = {
        (
            k,
            tuple(src_kv[k]) if k in src_kv else (0, dst_kv[k][1]),
            tuple(dst_kv[k]) if k in dst_kv else (0, src_kv[k][1]),
        )
        for k in set(src_kv.keys()).union(dst_kv)
    }

    return aligned
```

mcwitt: I think the various `align_*` functions defined here have a common pattern that might be useful to factor out.

E.g. with a common helper like
```python
def align_idxs_and_params(
    src_idxs,
    src_params,
    dst_idxs,
    dst_params,
    make_default,
    key=lambda idxs, _: idxs,
    get_idxs=lambda key: key,
):
    def to_hashable(x):
        return x if isinstance(x, Hashable) else tuple(x)

    src_kv = {to_hashable(key(idxs, params)): to_hashable(params) for idxs, params in zip(src_idxs, src_params)}
    dst_kv = {to_hashable(key(idxs, params)): to_hashable(params) for idxs, params in zip(dst_idxs, dst_params)}

    return {
        (
            get_idxs(k),
            src_kv[k] if k in src_kv else make_default(dst_kv[k]),
            dst_kv[k] if k in dst_kv else make_default(src_kv[k]),
        )
        for k in set(src_kv.keys()).union(dst_kv)
    }
```,

then the others can become one-liners, e.g.

```python
align_harmonic_bond_or_angle_idxs_and_params = partial(align_idxs_and_params, make_default=lambda p: (0, p[1]))
align_nonbonded_idxs_and_params = partial(align_idxs_and_params, make_default=lambda _: (0, 0, 0))
align_chiral_atom_idxs_and_params = partial(align_idxs_and_params, make_default=lambda _: 0)

align_torsion_idxs_and_params = partial(
    align_idxs_and_params,
    make_default=lambda p: (0, p[1], p[2]),
    key=lambda idxs, params: (idxs, params[2]),
    get_idxs=lambda key: key[0],
)
```

tests/test_interpolate.py
proteneer: It currently asserts

tests/test_interpolate_fe.py
proteneer: This is a pretty expensive test to run as well - seems like a bit of a waste just to assert that `dG_exact` is finite

tests/test_interpolate.py
mcwitt: > It currently asserts

nit: add test for this?

proteneer: willdo

proteneer:
After some more debugging, it looks like the nonbonded terms are already numerically unstable for any transformation that involves a dummy hydroxyl that might clash:

There are some absurdly small terms for intermediate windows:

`[i,j] [q_ij, s_ij, e_ij]`
`[ 5 33] [-5.46873960e+00  1.09283576e-01  4.94850147e-03]`

maxentile (approved):

timemachine/fe/single_topology_v3.py
```diff
@@ -612,3 +612,95 @@ def U_fn(x):
...
+        assert type(src_bond) == potentials.ChiralBondRestraint
+        assert type(dst_bond) == potentials.ChiralBondRestraint
+
+        idxs_and_params = interpolate.align_chiral_bond_idxs_and_params(
+            src_bond.get_idxs(),
+            src_bond.params,
+            src_bond.get_signs(),
+            dst_bond.get_idxs(),
+            dst_bond.params,
+            dst_bond.get_idxs(),
```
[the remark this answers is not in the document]
proteneer: good call

proteneer:
2 changes:

1) modified nonbonded_v3_pairlist_precomputed to take in an offset parameter to implement 4D decoupling
2) modified the interpolation function for nonbonded parameters to instead interpolate the offset from 0 to cutoff (or cutoff to 0).

timemachine/fe/single_topology_v3.py
```diff
@@ -612,3 +612,95 @@ def U_fn(x):
...
+        chiral_bond_params = []
+        chiral_bond_signs = []
+        for idxs, sign, src_k, dst_k in idxs_and_params:
+            chiral_bond_idxs.append(idxs)
+            new_params = interpolate_fn(src_k, dst_k, lamb)
+            chiral_bond_params.append(new_params)
+            chiral_bond_signs.append(sign)
+
+        # these should be properly sized
+        chiral_bond_idxs = np.array(chiral_bond_idxs, dtype=np.int32).reshape((-1, 4))
```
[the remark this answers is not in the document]
proteneer: the reshape is to deal with weird zero shaped things like an empty list `[]` which need to be re-interpreted as a `(0, 4)` shape in order for the jax functions downstream to be robust.

tests/test_interpolate.py
```diff
@@ -0,0 +1,227 @@
...
+        # since the permuting the order of idxs will affect
+        # the order of operations
+        np.testing.assert_almost_equal(U_ref(x0), U_test(x0))
+
+        system_dst = top.setup_end_state_dst()
+        system_lambda_1 = top.setup_intermediate_state(1)
+        U_ref = system_dst.get_U_fn()
+        U_test = system_lambda_1.get_U_fn()
+
+        np.testing.assert_almost_equal(U_ref(x0), U_test(x0))
```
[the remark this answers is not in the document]
maxentile: >How random are we talking about here? Sampled from MD?

I don't think it has to be sampled from MD, just it looked like the test could easily be made stronger if it looped over a few conformers with different values of the relevant internal d.o.f.s (distances referenced by harmonic bonds, angles referenced by harmonic angles, etc.).

Something like

```python
np.random.seed(2022)
xs = [x0]
for _ in range(10):
   xs.append(x0 + 0.01 * np.random.randn(*x0.shape))

for x in xs:
   np.testing.assert_almost_equal(U_ref(x), U_test(x))
```

would more than suffice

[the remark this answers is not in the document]
proteneer: ah good call! Thanks that's easy!

mcwitt (approved):

tests/test_interpolate.py
```diff
@@ -0,0 +1,227 @@
...
+from timemachine.ff import Forcefield
+
+
+def test_align_harmonic_bond():
+    """
+    Test that we can align idxs and parameters correctly for harmonic bonds.
+    We expect that decoupled terms have their force constants turned set to zero,
+    while maintaining the same equilibrium bond lengths.
+    """
+    a, b, c, d, e, f, g, h, i, j = np.random.rand(10)
```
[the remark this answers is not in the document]
proteneer: I don't think this really matters that much - this could be literally just `np.arange(10)` if that's more preferable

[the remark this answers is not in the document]
proteneer: This test is still deterministic in the output even if a random number is used in the input. 

Consider: you can have a function f(x)=0 have random inputs but the output will always be zero. This is a deterministic output.

Furthermore, you can have a function f(x)=x assert that output=input for any value of x (even random ones). The behavior is completely deterministic. This is what the current usage pattern is, ie we're just asserting that the broadcasting is done correctly on some arbitrary values. 

We require determinism tests for nonbonded tests and randomly generated inputs because there are two sources of non-determinism:

The test function f(x) may return different values if we do not use bitwise checks (for the same value of x). 

The test function f(x) may not agree with a reference function g(x) to a sufficient level of precision for different values of x.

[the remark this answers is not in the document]
mcwitt: I agree that this seems minor, but just wanted to clarify my reasoning for the initial comment.

Deterministic inputs seem preferable to me, even in cases where we are confident that the outcome is regardless deterministic, because
- We don't need to rely on (potentially flawed, especially in more complex cases) reasoning that the result is indeed deterministic. With nondeterministic inputs, the test might seem suspicious to future readers expecting deterministic tests until they've verified this reasoning for themselves
- There doesn't seem to be a benefit to nondeterministic inputs. If we want to test that the outcome is satisfied for varying inputs, we can still do this deterministically by generating a fixed number of random variates, as @maxentile mentioned

PR #764, opened 2022-06-27
Add flat-bottom bond potential

maxentile:
Intended usage:
* Generating samples where a bond length is nearly uniform between r_min, r_max, e.g. as a proposal for reweighting to a variety of equilibrium bond-lengths

Design notes:
* uses r^4 instead of r^2 so that the function is C^2 continuous

Implementation notes:
* Signature is simplified relative to harmonic bond (does not reference `lamb_mult`, `lamb_offset` vectors or expose a dependence on lambda)
* `test_flat_bottom_bond` introduces code duplication with `test_harmonic_bond` -- if we add more `*BondPotential` variants, may be worthwhile to factor out common parts of the test

### DIFF
--- a/tests/test_bonded.py
+++ b/tests/test_bonded.py
@@ -196,6 +196,60 @@ class TestBonded(GradientTest):
             np.testing.assert_array_equal(test_du_dp, test_du_dp_rev)
             np.testing.assert_array_equal(test_du_dl, test_du_dl_rev)
 
+    def test_flat_bottom_bond(self, n_particles=64, n_bonds=35, dim=3):
+        """Randomly connect pairs of particles, then validate the resulting FlatBottomBond force"""
+        np.random.seed(2022)
+
+        # TODO(deboggle) : reduce code duplication between HarmonicBond and FlatBottomBond
+        x = self.get_random_coords(n_particles, dim)
+
+        atom_idxs = np.arange(n_particles)
+
+        # (k, r_min, r_max)
+        params = np.random.rand(n_bonds, 3).astype(np.float64)
+        params[:,0] *= 1000           # make k large
+        params[:,-1] += params[:,-2]  # guarantee r_max >= r_min
+        
+        bond_idxs = []
+        for _ in range(n_bonds):
+            bond_idxs.append(np.random.choice(atom_idxs, size=2, replace=False))
+        bond_idxs = np.array(bond_idxs, dtype=np.int32) if n_bonds else np.zeros((0, 2), dtype=np.int32)
+
+        box = np.eye(3) * 100
+
+        relative_tolerance_at_precision = {np.float64: 1e-7, np.float32: 2e-5}
+        lamb = 0.0
+        
+        for precision, rtol in relative_tolerance_at_precision.items():
+            test_potential = potentials.FlatBottomBond(bond_idxs)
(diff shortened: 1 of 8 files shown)


maxentile:
Thanks for diagnosing and fixing these issues @badisa !

Summarizing 2 important differences relative to `HarmonicBond` implementation:
* Needed to use PBC-aware displacements
* Needed to compute displacements in double precision

tests/test_bonded.py
```diff
@@ -196,6 +196,59 @@ def test_harmonic_bond(self, n_particles=64, n_bonds=35, dim=3):
...
+        # TODO(deboggle) : reduce code duplication between HarmonicBond and FlatBottomBond
+        box = np.eye(3) * 100
+        x = self.get_random_coords(n_particles, dim)
+
+        atom_idxs = np.arange(n_particles)
+
+        # (k, r_min, r_max)
+        params = np.random.rand(n_bonds, 3).astype(np.float64)
+        params[:, 0] *= 1000  # make k large
+        params[:, -1] += params[:, -2]  # guarantee r_max >= r_min
```
mcwitt, line 212: Is this the wrong way around? (looks like it guarantees `r_min >= r_max`)

maxentile: I think this is correct (initialize `r_min, r_max` non-negative, update `r_max += r_min`)

```python
import numpy as np
n_bonds = 100

params = np.random.rand(n_bonds, 3).astype(np.float64)
params[:, 0] *= 1000  # make k large
params[:, -1] += params[:, -2]
k, r_min, r_max = params.T
assert (r_max >= r_min).all()
```


but I should make this more direct...

mcwitt: Ah, sorry -- I was reading `params[:, -1] += params[:, -2]` as ` params[:, 1] += params[:, 2]`

maxentile: Rephrased in https://github.com/proteneer/timemachine/pull/764/commits/e93b8b5df364716e4bfc42cce7e155ceaf6a129b

mcwitt (approved):
Nice, this looks great to me!

timemachine/cpp/src/kernels/k_flat_bottom_bond.cuh
```diff
@@ -0,0 +1,98 @@
...
+    }
+    RealType r = sqrt(r2);
+
+    // branches -> masks
+    RealType r_gt_rmax = static_cast<RealType>(r > rmax);
+    RealType r_lt_rmin = static_cast<RealType>(r < rmin);
+
+    if (u) {
+        // branchless implementation of piecewise function
+        RealType u_real = (k / 4) * (r_lt_rmin * (pow(r - rmin, 4)) + r_gt_rmax * (pow(r - rmax, 4)));
```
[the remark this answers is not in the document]
maxentile: Probably would be fine either way. May be a premature optimization (haven't checked if there's any difference in emitted device code if this were replaced with an if/else), but another motivation was to write in a way that ~the same line of code could be used in the Jax implementation https://github.com/proteneer/timemachine/blob/e93b8b5df364716e4bfc42cce7e155ceaf6a129b/timemachine/potentials/bonded.py#L349

badisa (approved):

PR #765, opened 2022-07-05
CUDA Nonbonded Precomputed

proteneer:
This is a pretty simple PR, and mirrors the nonbonded jax implementation identically.

Note: This class should be used for small-molecule intramolecular interactions only (ligands in particular). This class should not be used for the cancellation of nans since the interpolation of `q_ij, sig_ij, eps_ij` is done on python side and is *not* bit-wise cancellable with the exclusions.


timemachine/cpp/src/nonbonded_precomputed.cu
```diff
@@ -0,0 +1,94 @@
...
+    unsigned long long *d_du_dx,
+    unsigned long long *d_du_dp,
+    unsigned long long *d_du_dl,
+    unsigned long long *d_u,
+    cudaStream_t stream) {
+
+    if (P != 3 * B_) {
+        throw std::runtime_error(
+            "NonbondedPairListPrecomputed::execute_device(): expected P == 2*B, got P=" + std::to_string(P) +
+            ", 2*B=" + std::to_string(2 * B_));
```
badisa, line 61:
```suggestion
            "NonbondedPairListPrecomputed::execute_device(): expected P == 3*B, got P=" + std::to_string(P) +
            ", 3*B=" + std::to_string(3 * B_));
```

timemachine/potentials/nonbonded.py
```diff
@@ -326,6 +334,7 @@ def nonbonded_v3_on_precomputed_pairs(conf, params, box, pairs, offsets, beta: f
     conf: N,3
     params: P,3 (q_ij, s_ij, e_ij)
     pairs: P,2 (i,j)
+    offsets: P, 2 (w,) distance offsets for electrostatic and lj interactions
```
proteneer, line 337: note to self: fix

badisa (approved):

tests/nonbonded/test_nonbonded_precomputed.py
```diff
@@ -0,0 +1,95 @@
...
+    ixn_group_size,
+    precision,
+    rtol,
+    atol,
+    cutoff,
+    beta,
+    example_nonbonded_potential,
+    example_conf,
+    example_box,
+    rng: np.random.Generator,
```
mcwitt, line 50: nit: unused argument

mcwitt (approved):

tests/nonbonded/test_nonbonded_precomputed.py
proteneer: the conftest setup doesn't let me rename these - and I don't think I can just remove them.

proteneer: nevermind - I was just confused about how the conftest works. Thanks for resolving this offline.

PR #768, opened 2022-07-18
Support simulations without barostat and ensure same types for edge type

badisa:
- Previously we required a barostat, which will cause vacuum simulations to crash after awhile (box can always get smaller)
- `prepare_host_edge` and `prepare_vacuum_edge` didn't return the same type for masses, resulting in having to use them differently, it has been made so that both return masses as an `np.ndarray`


tests/test_free_energy.py
```diff
@@ -455,5 +455,27 @@ def test_absolute_vacuum():
     afe = free_energy.AbsoluteFreeEnergy(mol, bt)
 
     unbound_potentials, sys_params, masses = afe.prepare_vacuum_edge(ff_params)
-    assert masses == utils.get_mol_masses(mol)
+    assert np.all(masses == utils.get_mol_masses(mol))
     np.testing.assert_array_almost_equal(afe.prepare_combined_coords(), utils.get_romol_conf(mol))
+
+
+def test_vacuum_and_solvent_edge_types():
+    """Ensure that the values returned by the vacuum and solvent edges are all of the same time."""
```
maxentile, line 463:
```suggestion
    """Ensure that the values returned by the vacuum and solvent edges are all of the same type."""
```

maxentile (approved):

timemachine/fe/estimator.py
```diff
@@ -252,23 +253,27 @@ def simulate(
         integrator = copy.deepcopy(integrator)
         integrator.seed = np.random.randint(np.iinfo(np.int32).max)
 
-    if barostat.seed == 0:
-        barostat = copy.deepcopy(barostat)
-        barostat.seed = np.random.randint(np.iinfo(np.int32).max)
+    barostat_impl = None
+    if barostat is not None:
+        if barostat.seed == 0:
```
jkausrelay, line 258: nit: `if barostat is not None and barostat.seed == 0:`

jkausrelay (approved):
LGTM, minor comment

PR #769, opened 2022-07-18
Update to jax 0.3.14

jkausrelay:


environment.yml
```diff
@@ -4,7 +4,7 @@ channels:
   - openeye
 dependencies:
   - python=3.7
-  - jax=0.2.27
-  - jaxlib=0.1.75
```
proteneer, line 8: weird jaxlib got a huge bump it seems

jkausrelay: they decided to make them both use consistent version #s

proteneer: that just makes too much sense!

proteneer (approved):
lgtm

PR #772, opened 2022-07-20
Move parallel client executor so it can be pickled

jkausrelay:


timemachine/parallel/client.py
```diff
@@ -97,13 +97,17 @@ def __init__(self, max_workers):
...
         """
-        ctxt = multiprocessing.get_context("spawn")
-        # (ytz): on python <= 3.6 this will throw an exception since mp_context is
-        # not supported
-        self.executor = futures.ProcessPoolExecutor(max_workers=max_workers, mp_context=ctxt)
         self.max_workers = max_workers
         self._idx = 0
 
+    @property
+    def executor(self):
```
badisa, line 104: This will spawn a new executor for every time you call `self.executor` in `submit`. The logic will probably still work, but will accumulate workers and most likely cause ooms

I would do something like the following
```suggestion
    def executor(self):
        if self._executor is None:
            ctxt = multiprocessing.get_context("spawn")
            self._executor = futures.ProcessPoolExecutor(max_workers=self.max_workers, mp_context=ctxt)
        return self._executor
```

You could also use this: https://docs.python.org/3/library/functools.html#functools.cached_property though don't know if it is picklable

badisa (approved):

PR #775, opened 2022-07-25
Refactor RBFE code and fix various bugs

proteneer:
This PR

1) disables the chiral restraints until we've validated them some more (temporary)
2) cleans up the 20ns and 5ns edge lists in the example hif2a directory
3) refactors free energy setup code to `fe/rbfe.py` to be more modular
4) implements delta_U decomposition


timemachine/datasets/fep_benchmark/hif2a/results_edges_20ns.csv
```diff
@@ -1,93 +1,81 @@
...
-7a,23.0,2.93,0.35,0.06,0.57,0.39,N/A,N/A
-224.0,231.0,-0.36,0.06,0.1,0.09,0.29,N/A,N/A
-84.0,15.0,-3.47,-2.7,0.11,-2.49,1.08,N/A,N/A
-254.0,251.0,-0.69,0.5,0.02,0.62,1.21,N/A,N/A
-254.0,50.0,-0.15,1.56,0.08,1.37,0.58,N/A,N/A
-23.0,15.0,-3.98,-3.41,0.06,-3.29,0.98,N/A,N/A
-251.0,15.0,-0.8,-1.61,0.05,-1.68,1.21,N/A,N/A
-251.0,25.0,-0.04,1.04,0.08,0.44,1.31,N/A,N/A
-15.0,25.0,0.76,2.16,0.04,2.12,1.31,N/A,N/A
+Ligand1,Ligand2,Exp.,FEP,FEP Error,CCC,CCC Error
```
proteneer, line 1: this file has fewer rows than the lhs because the N/A rows have been removed

timemachine/fe/rbfe.py
```diff
@@ -0,0 +1,287 @@
...
+                dG, dG_err = pymbar.BAR(fwd_delta_u, rev_delta_u)
+                dG /= beta
+                dG_err /= beta
+
+                plot_BAR(dG, dG_err, fwd_delta_u, rev_delta_u, U_names[u_idx], get_plot_idx())
+
+                all_fwd_delta_us.append(fwd_delta_u)
+                all_rev_delta_us.append(rev_delta_u)
+
+            # sanity check - I don't think the dG calculation commutes with its components, so we have to re-estimate
```
proteneer, line 239: @maxentile this is where the delta_U decomposition is done, I've added a comment here for future readers

timemachine/fe/system.py
```diff
@@ -183,8 +183,8 @@ def get_U_fns(self):
             self.bond,
             self.angle,
             self.torsion,
-            self.chiral_atom,
-            self.chiral_bond,
+            # self.chiral_atom,
```
proteneer, line 186: # intentionally disabled until we fix MCS bits (and enforce chirality checking)

timemachine/fe/rbfe.py
```diff
@@ -0,0 +1,378 @@
...
+    temperature = 300.0
+    friction = 1.0
+
+    intg = LangevinIntegrator(temperature, dt, friction, masses, seed)
+    intg_impl = intg.impl()
+
+    baro = MonteCarloBarostat(len(masses), 1.0, temperature, group_idxs, 15, seed + 1)
+    baro_impl = baro.impl(bound_impls)
+
+    # tbd: add barostat
```
jkausrelay, line 74: stale comment?

timemachine/fe/rbfe.py
```diff
@@ -0,0 +1,378 @@
...
+        # minimize water box around the ligand by 4D-decoupling
+        mol_a_conf = get_romol_conf(st.mol_a)
+        mol_b_conf = get_romol_conf(st.mol_b)
+        ligand_conf = st.combine_confs(mol_a_conf, mol_b_conf)
+        combined_conf = np.concatenate([host_conf, ligand_conf])
+        x0 = combined_conf
+        v0 = np.zeros_like(x0)
+        box0 = host_config.box
+        group_idxs = get_group_indices(get_bond_list(hgs.bond))
+        seed = np.random.randint(0, 1000)
```
jkausrelay, line 180: is there a reason for this (vs just passing a fixed seed?)

timemachine/fe/rbfe.py
```diff
@@ -0,0 +1,378 @@
...
+
+    Return
+    ------
+    3-tuple of dGs, dG_errs, overlap_plot
+        All components have have shape (L-1,)
+
+    """
+    # assume pair-BAR format
+    # setup U_fns to operate in batch mode
+    kT = BOLTZ * 300.0
```
jkausrelay, line 235: `timemachine.constants.DEFAULT_KT`

timemachine/fe/rbfe.py
```diff
@@ -0,0 +1,378 @@
...
+        # convert from '<timemachine.lib.potentials.Nonbonded object at 0x7f7880b900b8>' -> Nonbonded
+        U_names.append(repr(U_fn).split(".")[-1].split()[0])
+
+    num_rows = len(lambda_schedule) - 1
+    num_cols = len(U_names) + 1
+
+    plot_iota = 1
+
+    def get_plot_idx():
+        nonlocal plot_iota  # what is this black magic? needed to deal with += increment
```
jkausrelay, line 258: 😮

timemachine/fe/rbfe.py
```diff
@@ -0,0 +1,378 @@
...
+            # loop over bond, angle, torsion, nonbonded terms etc.
+            all_fwd_delta_us = []
+            all_rev_delta_us = []
+
+            for u_idx, (prev_U_fn, cur_U_fn) in enumerate(zip(prev_U_fns, cur_U_fns)):
+                fwd_delta_u = beta * (cur_U_fn(prev_frames, prev_boxes) - prev_U_fn(prev_frames, prev_boxes))
+                rev_delta_u = beta * (prev_U_fn(cur_frames, cur_boxes) - cur_U_fn(cur_frames, cur_boxes))
+
+                dG, dG_err = pymbar.BAR(fwd_delta_u, rev_delta_u)
+                dG /= beta
```
jkausrelay, line 285: dg /= beta converts to kJ/mol (so update plot_BAR title units or convert after passing to plot_BAR)

timemachine/fe/rbfe.py
```diff
@@ -0,0 +1,378 @@
...
+        atom_mapping of atoms in mol_a into atoms in mol_b
+
+    ff: ff.Forcefield
+        Forcefield to be used for the system
+
+    host_config: HostConfig
+        Configuration for the host system.
+
+    n_frames: int
+        number of samples to generate for each lambda windows
```
jkausrelay, line 349: "generate for endstate lambda windows"

Since you filter them below

jkausrelay: ... where each sample is after 750 steps of MD.

timemachine/fe/rbfe.py
```diff
@@ -0,0 +1,378 @@
...
+    # burn-in
+    ctxt.multiple_steps_U(
+        lamb=lamb,
+        n_steps=burn_in,
+        lambda_windows=[lamb],
+        store_u_interval=0,
+        store_x_interval=0,
+    )
+
+    steps_per_frame = 750
```
jkausrelay, line 86: Short comment about the choice of 750 here

examples/rbfe_edge_list.py
```diff
@@ -0,0 +1,129 @@
...
+            return m
+
+    assert 0, "Mol not found"
+
+
+def run_pair(mol_a, mol_b, core, forcefield_path, protein_path, n_frames):
+
+    box_width = 4.0
+    solvent_sys, solvent_conf, solvent_box, solvent_top = builders.build_water_system(box_width)
+    solvent_box += np.diag([0.1, 0.1, 0.1])  # remove any possible clashes
```
jkausrelay, line 29: deboggle: seems like build_water_system should do this

examples/rbfe_edge_list.py
```diff
@@ -0,0 +1,129 @@
...
+            fut = cpc.submit(run_pair, mol_a, mol_b, core, args.forcefield, args.protein, args.n_frames)
+            futures.append(fut)
+
+            metadata.append(
+                (
+                    mol_a,
+                    mol_b,
+                    mcs_result.smartsString,
+                    core,
+                    float(exp_ddg) * 4.18,
```
jkausrelay, line 96: timemachine.constants.KCAL_TO_KJ

examples/rbfe_edge_list.py
```diff
@@ -0,0 +1,129 @@
...
+
+    for fut, meta in zip(futures, metadata):
+        solvent_res, solvent_top, complex_res, complex_top = fut.result()
+        solvent_ddg = np.sum(solvent_res.all_dGs)
+        solvent_ddg_err = np.linalg.norm(solvent_res.all_errs)
+        complex_ddg = np.sum(complex_res.all_dGs)
+        complex_ddg_err = np.linalg.norm(complex_res.all_errs)
+
+        mol_a, mol_b, _, _, exp_ddg, fep_ddg, fep_ddg_err, ccc_ddg, ccc_ddg_err = meta
+        mol_a_name = mol_a.GetProp("_Name")
```
jkausrelay, line 112: `timemachine.fe.utils.get_mol_name`

examples/relative_free_energy.py
```diff
@@ -1,194 +1,63 @@
...
+    solvent_res = estimate_relative_free_energy(
+        mol_a, mol_b, core, forcefield, solvent_host_config, n_frames=n_frames, prefix="solvent"
     )
 
-    assert all_coords.shape[0] == n_frames
-    assert all_boxes.shape[0] == n_frames
-
-    return all_coords, all_boxes
+    with open("hif2a_solvent_overlap.png", "wb") as fh:
+        fh.write(solvent_res.plot_png)
```
jkausrelay, line 41: remove or change name

examples/relative_free_energy.py
```diff
@@ -279,6 +93,7 @@ def read_from_args():
     parser.add_argument("--mol_b_name", type=str, help="name of the end molecule", required=True)
     parser.add_argument("--forcefield", type=str, help="location of the ligand forcefield", required=True)
     parser.add_argument("--protein", type=str, help="PDB of the protein complex", required=True)
```
jkausrelay, line 95: Is this meant to be specific to hif2a or general? args.protein seems unused here

examples/rbfe_edge_list.py
```diff
@@ -0,0 +1,129 @@
...
+
+    cpc = CUDAPoolClient(args.n_gpus)
+    cpc.verify()
+
+    with open(args.results_csv) as csvfile:
+        reader = csv.reader(csvfile, delimiter=",")
+        next(reader)
+        rows = [row for row in reader]
+        for row in rows:
+            mol_a_name, mol_b_name, exp_ddg, fep_ddg, fep_ddg_err, ccc_ddg, ccc_ddg_err = row
```
jkausrelay, line 78: Should the format be more flexible (allow exp_ddg, ... to be empty strings)?

examples/rbfe_edge_list.py
proteneer: agreed - I think a recent version of OpenMM may have fixed this: https://github.com/openmm/openmm/issues/3502 but we need to validate

examples/rbfe_edge_list.py
proteneer: probably - this file only works on the pre-generated edge list right now, but we can change it later.

examples/rbfe_edge_list.py
proteneer: willfix

examples/rbfe_edge_list.py
proteneer: willfix

examples/relative_free_energy.py
proteneer: should be in general, willfix

timemachine/fe/rbfe.py
proteneer: willfix

timemachine/fe/rbfe.py
proteneer: willfix

timemachine/fe/rbfe.py
proteneer: was a temporary solution - wasn't sure if i should have the user put in a single seed, (then increment it by # lambda windows along the way) or have them pass in an arrya of seeds.

timemachine/fe/rbfe.py
proteneer: I need to fix this to be self-consistent with the temperature of the thermostat.

timemachine/fe/rbfe.py
proteneer: willfix

timemachine/fe/rbfe.py
proteneer: willfix

timemachine/fe/rbfe.py
proteneer: I'm going to set this to 1000 (was what we had for RABFE)

timemachine/fe/rbfe.py
proteneer: fixed by propagating seeds all the out

timemachine/fe/rbfe.py
proteneer: to clarify, this is saying that each window runs for n_frames*1000 steps of MD. It's independent of how many frames we store etc. I'll add a note in the return type that only first and last windows are currently returned.

examples/rbfe_edge_list.py
```diff
@@ -0,0 +1,134 @@
...
+from timemachine.fe.rbfe import HostConfig, estimate_relative_free_energy
+from timemachine.fe.utils import get_mol_name
+from timemachine.ff import Forcefield
+from timemachine.md import builders
+from timemachine.parallel.client import CUDAPoolClient
+
+
+def get_mol_by_name(mols, name):
+    for m in mols:
+        if m.GetProp("_Name") == name:
```
mcwitt, line 21: nit: use `get_mol_name`?

mcwitt: Aside: a pattern I've found useful is something like
```python
mol_by_name = {get_mol_name(mol): mol for mol in mols}
# …
mol_a = mol_by_name[mol_a_name]
```

examples/relative_free_energy.py
```diff
@@ -1,194 +1,63 @@
...
-    baro_impl = baro.impl(bound_impls)
+    for window_idx, window_frames in enumerate(all_frames):
+        out_path = f"{prefix}_{window_idx}.pdb"
+        writer = pdb_writer.PDBWriter([host_topology, dummy_st.mol_a, dummy_st.mol_b], out_path)
+        for frame in window_frames:
+            host_frame = frame[: host_topology.getNumAtoms()]
+            ligand_frame = frame[host_topology.getNumAtoms() :]
+            mol_ab_frame = pdb_writer.convert_single_topology_mols(ligand_frame, dummy_st)
+            writer.write_frame(np.concatenate([host_frame, mol_ab_frame]) * 10)
+    writer.close()
```
mcwitt, line 27: Should this line be indented to be inside the outer loop? Otherwise it seems like this only closes the final `writer`.

examples/rbfe_edge_list.py
```diff
@@ -0,0 +1,134 @@
...
+        mol_a, mol_b, core, forcefield, complex_host_config, seed + 1, n_frames=n_frames, prefix="complex"
+    )
+
+    return solvent_res, solvent_top, complex_res, complex_top
+
+
+def read_from_args():
+
+    parser = argparse.ArgumentParser(
+        description="Estimate relative free energy difference between complex and solvent given two ligands mol_a and mol_b."
```
mcwitt, line 54: Update description (looks like this was copied from the `relative_free_energy.py` example?)

examples/relative_free_energy.py
```diff
@@ -301,20 +117,9 @@ def read_from_args():
...
     forcefield = Forcefield.load_from_file(args.forcefield)
 
-    estimate_relative_free_energy(mol_a, mol_b, core, forcefield, solvent_host_config, n_frames=100, prefix="solvent")
-
-    complex_sys, complex_conf, _, _, complex_box, complex_top = builders.build_protein_system(args.protein)
-    complex_box += np.diag([0.1, 0.1, 0.1])  # remove any possible clashes
-    complex_host_config = HostConfig(complex_sys, complex_top, complex_conf, complex_box)
-
-    estimate_relative_free_energy(mol_a, mol_b, core, forcefield, complex_host_config, n_frames=100, prefix="complex")
+    run_pair(mol_a, mol_b, core, forcefield, args.protein)
```
mcwitt, line 122: Missing `n_frames`?

examples/relative_free_energy.py
```diff
@@ -1,194 +1,63 @@
...
 
-    # burn-in
-    ctxt.multiple_steps_U(
-        lamb=lamb,
-        n_steps=burn_in,
-        lambda_windows=[lamb],
-        store_u_interval=0,
-        store_x_interval=0,
-    )
+def run_pair(mol_a, mol_b, core, forcefield, n_frames, protein_path):
```
mcwitt, line 30: Worth making `seed` and argument to be consistent with other examples?

timemachine/fe/rbfe.py
mcwitt: I think this could be avoided with something like
```python
_, axss = plt.subplots(num_rows, num_cols)
# …
for lamb, axs in zip(lambda_schedule, axss):
    # …
    for prev_U_fn, cur_U_fn, ax in zip(prev_U_fns, cur_U_fns, axs):
        # …
        plot_BAR(df, df_err, fwd_delta_u, rev_delta_u, U_names[u_idx], ax=ax)
```
(and update `plot_BAR` to take an `ax` argument instead of `plot_idx`)

timemachine/fe/system.py
mcwitt: Could we add this comment to the code?

timemachine/datasets/fep_benchmark/hif2a/results_edges_20ns.csv
maxentile: Hmm, are the rows where `"Exp."` field is a float but `"FEP"` field is a NaN still useful for anything?

timemachine/fe/rbfe.py
```diff
@@ -0,0 +1,423 @@
...
+from timemachine.fe.single_topology_v3 import SingleTopologyV3
+from timemachine.fe.system import convert_bps_into_system
+from timemachine.fe.utils import get_romol_conf
+from timemachine.ff.handlers import openmm_deserializer
+from timemachine.lib import LangevinIntegrator, MonteCarloBarostat, custom_ops
+from timemachine.md import minimizer
+from timemachine.md.barostat.utils import get_bond_list, get_group_indices
+
+
+def plot_atom_mapping_grid(mol_a, mol_b, core_smarts, core, show_idxs=False):
```
maxentile, line 22: Should this be moved into a plotting utilities module or similar?

(This file contains the only library code with a dependence on matlotlib. (Other usages appear in `attic`, `tests`, or commented-out code.) Implementation also seems equally applicable to other cases where atom-mapping is used, such as relative hydration.)

timemachine/fe/rbfe.py
```diff
@@ -0,0 +1,423 @@
...
+        molsPerRow=3,
+        highlightAtomLists=[list(range(mol_q_2d.GetNumAtoms())), core[:, 0].tolist(), core[:, 1].tolist()],
+        highlightAtomColors=[atom_colors_q, atom_colors_a, atom_colors_b],
+        subImgSize=(300, 300),
+        legends=["core", mol_a.GetProp("_Name"), mol_b.GetProp("_Name")],
+        useSVG=True,
+    )
+
+
+def sample(bound_impls, masses, temperature, x0, v0, box0, lamb, group_idxs, seed, n_frames=500, burn_in=10000):
```
maxentile, line 62: Is this function `rbfe`-specific? (On first read, this implementation does not have obvious RBFE-tailored logic)

timemachine/fe/rbfe.py
```diff
@@ -0,0 +1,423 @@
...
+        frames, boxes = sample(
+            bound_impls, combined_masses, temperature, x0, v0, box0, lamb, group_idxs, run_seed, n_frames=n_frames
+        )
+        all_frames.append(frames)
+        all_boxes.append(boxes)
+
+    return all_frames, all_boxes, all_U_fns
+
+
+def plot_BAR(df, df_err, fwd_delta_u, rev_delta_u, title, plot_idx):
```
maxentile, line 193: Should this be moved into a plotting utilities module or similar?

(Implementation does not seem RBFE-specific -- also seems equally applicable to other cases where forward and reverse works can be computed.)

timemachine/fe/rbfe.py
```diff
@@ -0,0 +1,423 @@
...
+    plt.title(f"{title}, dg: {df:.2f} +- {df_err:.2f} kTs")
+    plt.hist(fwd_delta_u, alpha=0.5, label="fwd", density=True, bins=20)
+    plt.hist(-rev_delta_u, alpha=0.5, label="-rev", density=True, bins=20)
+    plt.xlabel("work (kTs)")
+    plt.legend()
+
+
+SimulationResult = namedtuple(
+    "SimulationResult", ["all_dGs", "all_errs", "plot_png", "frames", "boxes", "U_fns", "lambda_schedule"]
+)
```
maxentile, line 228: Introduces possible name collision with `SimulationResult` from fe.estimator -- are these intended to be merged, or can one of them be renamed?

timemachine/fe/rbfe.py
```diff
@@ -0,0 +1,423 @@
...
+
+
+def estimate_free_energy_given_samples(all_frames, all_boxes, all_U_fns, temperature, lambda_schedule, prefix):
```
maxentile, line 231: 1. Is this implementation RBFE-specific, or can it be moved somewhere less specific, like `estimator.py` or similar?

(Implementation seems equally applicable to other cases where a free energy can be estimated from equilibrium samples.)

2. Multiple estimators are possible in this setting -- may want to mention in name or docstring which estimator this function implements.

timemachine/fe/rbfe.py
```diff
@@ -0,0 +1,423 @@
...
+    lambda_schedule: np.array of shape (L,)
+        Lambda schedule
+
+    prefix: str
+        A prefix to append to overlap figures
+
+    Return
+    ------
+    3-tuple of dGs, dG_errs, overlap_plot
+        All components have have shape (L-1,)
```
maxentile, line 260: nit: typo

tests/test_relative_free_energy.py
```diff
@@ -0,0 +1,78 @@
...
+
+    st = get_hif2a_ligand_pair_single_topology()
+    mol_a = st.mol_a
+    mol_b = st.mol_b
+    core = st.core
+    forcefield = st.ff
+    protein_path = "tests/data/hif2a_nowater_min.pdb"
+
+    # fast
+    run_pair(mol_a, mol_b, core, forcefield, n_frames=100, protein_path=protein_path)
```
maxentile, line 78: nit: test is annotated as both `fast` and `"Slow!"`

(If under a minute, may be useful run in standard test suite rather than nightly)

maxentile (approved):

examples/rbfe_edge_list.py
proteneer: willfix

examples/rbfe_edge_list.py
proteneer: willfix

examples/relative_free_energy.py
proteneer: sure

tests/test_relative_free_energy.py
proteneer: i'll remove this comment

timemachine/datasets/fep_benchmark/hif2a/results_edges_20ns.csv
proteneer: These are experimental ddGs - which can always be computed from the node dGs directly (present in a different file). So - I don't really think its useful to keep these

timemachine/fe/rbfe.py
proteneer: This plot is specific to a relative free energy calculation though - not sure what the best way to organize this is.

jkausrelay: Should matplotlib be added as a dependency then?

timemachine/fe/rbfe.py
proteneer: fe.estimator.SimulationResult will be completely removed during deboggling

timemachine/fe/rbfe.py
proteneer: this is specific to this script since it's calling subplots (as the outside code sets-up the figure sizes etc.)

timemachine/fe/rbfe.py
proteneer: willfix

timemachine/fe/system.py
proteneer: willfix

timemachine/fe/rbfe.py
proteneer: does that actually line-up? since `num_rows = len(lambda_schedule) - 1` I think this would result in an empty row?

proteneer: (also the main problem is due to a limitation in python not allowing us to modify closures iirc, and nonlocal is a patch, similar to global)

timemachine/fe/rbfe.py
proteneer: 1) the problem with `estimator.py` is that it's fairly tied to RABFE specific logic (like setting up the orientational restraints), didn't seem right to put it there
2) will add docstring

examples/relative_free_energy.py
proteneer: oops - good call (too many scripts to check... will revalidate all the examples once more before final merge)

timemachine/fe/rbfe.py
proteneer: yeah this probably shouldn't be here.

let's offline discuss the best way to organize free-energy/sampling modules going forward

timemachine/fe/rbfe.py
proteneer: per @mcwitt comment on using axes

(maybe if this took in arbitrary `plot` object (could be ax, or plot), as opposed to `plot_idx` object, this can be refactored out)

timemachine/fe/rbfe.py
mcwitt: > does that actually line-up?

Sorry, the example might need a little tweaking (maybe `prev_lambda, cur_lambda in zip(lambda_schedule, lambda)`), but my main point was that I think a nonlocal variable to keep track of which axis can be avoided by explicitly looping over the axes (e.g. by zipping with the other loop variables)

proteneer: this is a little annoying, a matplotlib `Axes` isn't the same as that of `title` in weird ways, for example:

`plt.title("my title")` is fine
`ax.title("my title")` breaks, and you'd have to use `ax.set_title("my title")` instead

proteneer: `plt.set_title` is also not a valid option either!

mcwitt: Yeah, it is a bit annoying. My rule of thumb is to prepend `set_` when working with axis objects 🤷‍♂️

maxentile: >this is a little annoying, a matplotlib axes isn't the same as that of plt in weird ways, for example:
>ax.title("my title") breaks, and you'd have to use ax.set_title("my title") instead
>plt.set_title is also not a valid option either!
>(similarly for xlabel vs set_xlabel)

This section in the matplotlib documentation may be helpful re: differences between "pyplot style" and "OO style" https://matplotlib.org/stable/tutorials/introductory/usage.html#the-object-oriented-and-the-pyplot-interfaces

(and following section I think describes the current use case https://matplotlib.org/stable/tutorials/introductory/usage.html#making-a-helper-functions )

timemachine/fe/rbfe.py
proteneer: good call - hm

proteneer: added matplotlib as a dependency. I'd like to remove this at a future deboggling though

requirements.txt
```diff
@@ -7,3 +7,4 @@ pyyaml==5.4.1
 networkx==2.5
 importlib-resources==5.4.0
 typing-extensions==4.1.1
+matplotlib==3.3.3
```
mcwitt, line 10: We should also add `matplotlib` to `install_requires` in setup.py

timemachine/fe/rbfe.py
proteneer: (let's leave this for deboggling - this isn't a public function that will be called. We can change this without breaking the API later)

timemachine/fe/rbfe.py
proteneer: will deal with 1) during deboggling

mcwitt (approved):

timemachine/fe/rbfe.py
proteneer: sorted this out - uses axes now

jkausrelay (approved):

jkausrelay:
lgtm!

PR #776, opened 2022-07-27
Jax reference VelocityVerletIntegrator

maxentile:
- [x] Add `VelocityVerletIntegrator` to `integrator.py`, with 3 methods of updating state (following what is present for `LangevinIntegrator`):
  - 2 "public" methods `.step` and `.multiple_steps` (only requiring a callable `force_fxn`), and
  - 1 "private" method `._update_via_fori_loop` (using accelerated jax.lax loop when `force_fxn` is jax-transformable)
- [x] Test reversibility up to numerical precision (key property for some applications, e.g. as a proposal in HMC)

### DIFF
--- a/tests/test_reference_velocity_verlet_integrator.py
+++ b/tests/test_reference_velocity_verlet_integrator.py
@@ -10,10 +10,11 @@ from timemachine.integrator import VelocityVerletIntegrator
 from timemachine.testsystems.relative import hif2a_ligand_pair
 
 
-def assert_reversible(x0, v0, update_fxn, atol=1e-8):
-    """assert that
-    * F(I(x0, v0))      != (x0, v0)
-    * F(I(F(I(x0, v0))) == (x0, v0)"""
+def assert_reversible(x0, v0, update_fxn, atol=1e-10):
+    """Assert that
+    * I(x0, v0) != (x0, v0)
+    * F(I(F(I(x0, v0)))) == (x0, v0)"""
+
     close = partial(np.allclose, atol=atol)
 
     # integrate, flip v, integrate, flip v
@@ -26,11 +27,11 @@ def assert_reversible(x0, v0, update_fxn, atol=1e-8):
     assert close(v0_, v0), f"max(abs(v0 - v0_)) = {np.max(np.abs(v0 - v0_))}"
 
     # also assert this is not a no-op
-    assert (not close(x1, x0)) and (not close(v1, v0)), "update_fxn didn't do anything!"
+    assert (not close(x1, x0)) and (not close(v1, v0))
 
 
-def assert_reversibility_using_step_implementations(intg, x0, v0, n_steps=1000, atol=1e-8):
-    """assert reversibility .step and .multiple_steps implementations"""
+def assert_reversibility_using_step_implementations(intg, x0, v0, n_steps=1000, atol=1e-10):
+    """Assert reversibility of .step and .multiple_steps implementations"""
 
     # check step implementation
     def step_update(x, v):
@@ -49,7 +50,7 @@ def assert_reversibility_using_step_implementations(intg, x0, v0, n_steps=1000, 
 
 
 def test_reversibility_with_jax_potentials():
-    """on a simple jax-transformable potential (quartic oscillators)
+    """On a simple jax-transformable potential (quartic oscillators)
     with randomized parameters and initial conditions
     (n oscillators, masses, dt, x0, v0)
     assert all 3 update functions
@@ -82,12 +83,13 @@ def test_reversibility_with_jax_potentials():
         def jax_update(x, v):
             return intg._update_via_fori_loop(x, v, n_steps)
 
-        assert_reversible(x0, v0, jax_update, atol=1e-8)
+        assert_reversible(x0, v0, jax_update, atol=1e-10)
 
 
 def test_reversibility_with_custom_ops_potentials():
     """Check reversibility of "public" .step and .multiple_steps implementations when `force_fxn`
     is a custom_op potential"""
+
     np.random.seed(2022)
 
     # define a Python force fxn that calls custom_ops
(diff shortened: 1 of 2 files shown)


tests/test_reference_velocity_verlet_integrator.py
```diff
@@ -0,0 +1,123 @@
...
+
+config.update("jax_enable_x64", True)
+
+from timemachine.integrator import VelocityVerletIntegrator
+from timemachine.testsystems.relative import hif2a_ligand_pair
+
+
+def assert_reversible(x0, v0, update_fxn, atol=1e-10):
+    """Assert that
+    * I(x0, v0) != (x0, v0)
```
badisa, line 15: `I` hear means integrate and `F` means flip?

maxentile: Correct

timemachine/integrator.py
```diff
@@ -88,6 +88,74 @@ def step(self, x, v):
...
+        # finalize traj
+        v = v + (0.5 * self.dt / self.masses) * self.force_fxn(x)
+
+        zs.append((x, v))
+
+        xs = np.array([x for (x, _) in zs])
+        vs = np.array([v for (_, v) in zs])
+        return xs, vs
+
+    def _update_via_fori_loop(self, x, v, n_steps=1000):
```
badisa, line 138: Do we need this? Seems to be only intended for testing and ends up duplicating some code

timemachine/integrator.py
```diff
@@ -88,6 +88,74 @@ def step(self, x, v):
         return new_x, new_v
 
 
+class VelocityVerletIntegrator(Integrator):
+    def __init__(self, force_fxn, masses, dt):
```
badisa, line 92: A docstring describing the preference of `multiple_steps` over `step` to avoid running 2x force computations in a tight loop would be nice.

timemachine/integrator.py
maxentile: Not just intended for testing -- also intended to be usable in situations where https://github.com/proteneer/timemachine/blob/2f06d9586693f97abba07452cb91b895265ec704/timemachine/integrator.py#L91 is usable

badisa (approved):

tests/test_reference_velocity_verlet_integrator.py
maxentile: Clarified this comment in https://github.com/proteneer/timemachine/pull/776/commits/5a2bdc9aa81837364aa18223d759b5fed01c7d82 and https://github.com/proteneer/timemachine/pull/776/commits/6eabb32f1fe60699d9a3a15e40ec33127262eaee

timemachine/integrator.py
maxentile: Added warnings in `__init__` and `step` docstrings about performance https://github.com/proteneer/timemachine/pull/776/commits/4e4a0e1fc20059ff6cc3261cc4bf04325923d1d6

tests/test_reference_velocity_verlet_integrator.py
```diff
@@ -0,0 +1,129 @@
...
+
+    x0 = np.array(coords)
+
+    for n_steps in [1, 10, 100, 500, 1000]:  # , 10000]:
+        v0 = np.random.randn(*coords.shape)
+
+        # check "public" .step and .multiple_steps implementations
+        assert_reversibility_using_step_implementations(intg, x0, v0, n_steps, atol=1e-10)
+
+    # TODO: possibly investigate why n_steps = 10000 fails
```
proteneer, line 125: after some offline discussion with @maxentile  - we can definitely make this bitwise reversible if we do accumulation in fixed point.

timemachine/integrator.py
```diff
@@ -88,6 +88,75 @@ def step(self, x, v):
...
+        zs.append((x, v))
+
+        xs = np.array([x for (x, _) in zs])
+        vs = np.array([v for (_, v) in zs])
+        return xs, vs
+
+    def _update_via_fori_loop(self, x, v, n_steps=1000):
+        # initialize
+        v = v + (0.5 * self.dt / self.masses) * self.force_fxn(x)
+        x = x + self.dt * v
```
proteneer, line 142: nice - this looks like its setting up the initial "half-step" kick?

timemachine/integrator.py
```diff
@@ -88,6 +88,75 @@ def step(self, x, v):
...
+
+        zs = [(x, v)]
+
+        # initialize traj
+        v = v + (0.5 * self.dt / self.masses) * self.force_fxn(x)
+        x = x + self.dt * v
+
+        # run n_steps-1 steps
+        for t in range(n_steps - 1):
+            v = v + (self.dt / self.masses) * self.force_fxn(x)
```
proteneer, line 125: Screen Shot 2022-07-28 at 3 03 24 PM

This looks identical to the OpenMM implementation - you mentioned an inconsistency but I'm not seeing it?

maxentile: Just wanted to draw your attention to the inclusion of initialize and finalize half-steps.

Note this introduces an inconsistency between `.step` and `.multiple_steps` 

https://github.com/proteneer/timemachine/blob/a3d1a49fcbb8191c3d8ee1f96723c487775224be/timemachine/integrator.py#L107-L113

* `.step` will update both x and v by a full timestep -- this is inefficient (requires 2 force evaluations) and is only intended for reference
* `.multiple_steps` will save x and v "staggered by a half-step" for all intermediates (as I think you would get if you queried [OpenMM context](http://docs.openmm.org/7.6.0/userguide/theory/04_integrators.html#verletintegrator) every step). Only the initial `(xs[0], vs[0])` and final `(xs[T], vs[T])` frames returned by this function will actually be "on-step." Could clean this up to return `(xs[t], vs[t]) = (x_t, v_t)` for all `t` to avoid headaches / be self-consistent, but initially I just care about the first and last frames.

proteneer: thanks! this makes sense to me - we can deal with the half-step issues later and just document it now.

proteneer (approved):

PR #780, opened 2022-07-28
Fixes #779: Specify jax cpu and update installation instructions

### LINKED ISSUE #779: update build instructions
```
conda env create -f environment.yml -n timemachine
conda install openmm=7.5.1 -c conda-forge # only if using openmm from conda
conda activate timemachine
```

should probably be:

```
conda env create -f environment.yml -n timemachine
conda activate timemachine
conda install openmm=7.5.1 -c conda-forge # only if using openmm from conda
```

else openmm gets installed into `base` env

and

we should probably add mypy to the `requirements.txt` even if we're not installing in developer mode since it provides the `stubgen` needed for `make` to succeed

badisa:
fixes #779

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ pip install -e .
 ```
 or
 ```shell
-make build
+make build  # Must have installed dev dependencies for this to work
 ```
 
 ### Running Tests
--- a/README.md
+++ b/README.md
@@ -39,8 +39,8 @@ If using conda the following can be used to configure your environment
 
 ```shell
 conda env create -f environment.yml -n timemachine
-conda install openmm=7.5.1 -c conda-forge # only if using openmm from conda
 conda activate timemachine
+conda install openmm=7.5.1 -c conda-forge # only if using openmm from conda
 ```
 
 ### Install Time Machine
--- a/environment.yml
+++ b/environment.yml
@@ -5,6 +5,6 @@ channels:
 dependencies:
   - python=3.7
   - jax=0.3.14
-  - jaxlib=0.3.14
+  - jaxlib=0.3.14=*cpu*
   - openeye-toolkits=2020.2.0
   - rdkit=2021.03.1
--- a/environment.yml
+++ b/environment.yml
@@ -5,6 +5,6 @@ channels:
 dependencies:
   - python=3.7
   - jax=0.3.14
-  - jaxlib=0.3.14=*cpu*
+  - jaxlib=0.3.14=cpu*
   - openeye-toolkits=2020.2.0
   - rdkit=2021.03.1
--- a/setup.py
+++ b/setup.py
@@ -92,7 +92,6 @@ setup(
         "pymbar>3.0.4",
         "pyyaml",
         "scipy",
-        "mypy",
         "typing-extensions",
     ],
     extras_require={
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -39,8 +39,8 @@ If using conda the following can be used to configure your environment
 
 ```shell
 conda env create -f environment.yml -n timemachine
-conda install openmm=7.5.1 -c conda-forge # only if using openmm from conda
 conda activate timemachine
+conda install openmm=7.5.1 -c conda-forge # only if using openmm from conda
```
mcwitt, line 43: ah, good catch!

setup.py
```diff
@@ -92,6 +92,7 @@ def build_extension(self, ext):
         "pymbar>3.0.4",
         "pyyaml",
         "scipy",
+        "mypy",
```
mcwitt, line 95: Could you comment on why this change was needed?

environment.yml
```diff
@@ -5,6 +5,6 @@ channels:
 dependencies:
   - python=3.7
   - jax=0.3.14
-  - jaxlib=0.3.14
+  - jaxlib=0.3.14=*cpu*
```
mcwitt, line 8: nit: maybe
```suggestion
  - jaxlib=0.3.14=cpu*
```
(since `cpu`/`cuda` is at the beginning of the build description string)

setup.py
badisa: Because of the change to require `stubgen` to build the package, you can't install this without mypy based on this comment (https://github.com/proteneer/timemachine/pull/711#discussion_r848956244)

We could go back to an explicit version of stubgen and keep mypy as a separate requirement. We probably should add some form of testing in CI that doing `pip install .` works out of the box, as right now it doesn't.

badisa: After digging in a bit more with @mcwitt it looks like we have a discontinuity between `make build` and `pip install`. `pip install` reaches into `pyproject.toml` and gets the build requirements there (that include mypy) and builds just fine, however `make build` will fail as `mypy` isn't in the env. 

Probably warrants an offline chat on how to proceed

mcwitt: Just a note that if we decide to leave things as-is, we should probably add a note to the readme that `make build` only works after installing the dev dependencies (`pip install .[dev]`)

proteneer: ah I see! I'm okay either way. (I do make build)

environment.yml
```diff
@@ -5,6 +5,6 @@ channels:
 dependencies:
   - python=3.7
   - jax=0.3.14
-  - jaxlib=0.3.14
+  - jaxlib=0.3.14=cpu*
```
proteneer, line 8: just checking that this should be `=cpu*` and not `=*cpu*`, if the former works that's probably preferable?

mcwitt: I suggested the change from `=*cpu*` to `=cpu*`, but only had a very minor preference (`cpu` is at the beginning of the string in the current schema; I think it comes down to whether we want things to break if they change the schema or continue to work (but slight chance of doing something unexpected))

jkausrelay (approved):

proteneer (approved):
lgtm

mcwitt (approved):
lgtm!

PR #781, opened 2022-07-29
Make intermediate potentials `jax.jit`table

mcwitt:
This should allow code like:

```python
st: SingleTopologyV3
# …

@jax.jit
def U(lam):
    return st.setup_intermediate_state(lam).get_U_fn()(conf)
```

Changes:
- Adds test for jit-tability of intermediate potentials
- Moves initialization of end states to `SingleTopologyV3` constructor. This avoids needing to JIT forcefield parameterization code
- Replaces some `np.array` conversions with `jnp.array` (done by following errors in tracing)


mcwitt:
Accidentally opened from the wrong branch.

PR #782, opened 2022-07-29
Make intermediate potentials `jax.jit`table

mcwitt:
This is to allow code like:

```python
@jax.jit # or @jax.vmap, @jax.pmap
def U(lam):
    return st.setup_intermediate_state(lam).get_U_fn()(conf) # st an instance of SingleTopologyV3
```

#### Changes
- Adds test for jit-tability of intermediate potentials
- Moves initialization of end states to `SingleTopologyV3.__init__` (avoids needing to jit forcefield parameterization code) and makes `setup_end_state_{src,dst}` methods private (since they should only be used by `__init__`)
- Replaces some `np.array` usages with `jnp.array` (done by following errors in tracing)

#### Todo
- [x] Update `SingleTopologyV3` to use end states stored as attributes (vs. recomputing)


proteneer (approved):
lgtm!

tests/test_single_topology_v3.py
```diff
@@ -273,3 +272,32 @@ def test_combine_masses():
...
+        st = SingleTopologyV3(mol_a, mol_b, core_pairs, ff)
+        conf = st.combine_confs(get_romol_conf(mol_a), get_romol_conf(mol_b))
+        return st, conf
+
+    st, conf = setup_arbitary_transformation()
+
+    def U(lam):
+        return st.setup_intermediate_state(lam).get_U_fn()(conf)
+
+    jax.jit(U)(0.1)
```
maxentile, line 303: Nit: might expand this test slightly to exercise fast map over lambdas and or confs

```python
def U(x, lam):
   ...

confs = np.array([conf for _ in range(100)])
lams = np.linspace(0, 1, 100)

vmapped_U = jax.vmap(U, in_axes=(0,0))
_ = vmapped_U(confs, lams)
_ = jax.jit(vmapped_U)(confs, lams)
```

or similar

maxentile (approved):

tests/test_single_topology_v3.py
mcwitt: Good call, added in 9018ca9.

PR #783, opened 2022-07-29
Improve self-consistency among Verlet stepper methods

maxentile:
Addresses https://github.com/proteneer/timemachine/pull/776#discussion_r932660565 by simplifying `.multiple_steps` to return both positions and velocities "on-step", so that it is now consistent with `.step`.

This may be helpful if we wish to use the trajectory returned by `multiple_steps` for an application that requires `U(xs[t]) + KE(vs[t]) == U(x_t) + KE(v_t)` for intermediate values of `t`. One such application is windowed acceptance HMC, described in sec. 5.4 of https://arxiv.org/abs/1206.1901 .

-----

(However, based on a discussion with @proteneer Tuesday, rearranging the loop in this way would introduce a different inconvenience if translated to a timemachine C++/CUDA integrator, as it would require evaluating `force(x)` once "within a timestep" rather than once "between timesteps" -- [`step_fwd` and `update_forward`](https://github.com/proteneer/timemachine/blob/8fc9b108435f313d0f604ede2500c23526975bd3/timemachine/cpp/src/integrator.cu#L71) reference a buffer of forces computed on previous `x`, rather than a force function. The previous implementation in https://github.com/proteneer/timemachine/pull/776 was intended to be consistent with the current signatures of `step_fwd` and `update_forward`.)

### DIFF
--- a/tests/test_reference_velocity_verlet_integrator.py
+++ b/tests/test_reference_velocity_verlet_integrator.py
@@ -127,3 +127,43 @@ def test_reversibility_with_custom_ops_potentials():
     # * also fails with reduced dt = 1.0e-3
     # * passes with greatly reduced dt = 1.0e-4
     # TODO: does replacing np.sum(du_dxs, 0) with SummedPotential address this?
+
+
+def test_trajectory_wise_self_consistency():
+    """Assert that trajectories produced by .step and .multiple_steps agree at all timesteps,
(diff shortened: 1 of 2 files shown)


maxentile:
Stale -- may be useful to revisit if on-step velocities are needed from `multiple_steps`

PR #784, opened 2022-08-01
Improve error handling and diagnostics in rbfe script

proteneer:
This is a minor PR that:

-Handles exceptions more gracefully in the `rbfe.py` code when an exception is thrown
-Adds a test for bitwise reproducibility using the `initial_state` and `protocol` object
-Removes an extra `print` statement related to plotting

### DIFF
--- a/examples/rbfe_edge_list.py
+++ b/examples/rbfe_edge_list.py
@@ -1,6 +1,7 @@
 import argparse
 import csv
 import pickle
+import traceback
 
 import numpy as np
 from rdkit import Chem
@@ -26,13 +27,12 @@ def get_mol_by_name(mols, name):
 
 def run_pair(mol_a, mol_b, core, forcefield_path, protein_path, n_frames, seed):
 
+    forcefield = Forcefield.load_from_file(forcefield_path)
+
     box_width = 4.0
     solvent_sys, solvent_conf, solvent_box, solvent_top = builders.build_water_system(box_width)
     solvent_box += np.diag([0.1, 0.1, 0.1])  # remove any possible clashes, deboggle later
     solvent_host_config = HostConfig(solvent_sys, solvent_conf, solvent_box)
-
-    forcefield = Forcefield.load_from_file(forcefield_path)
-
     solvent_res = estimate_relative_free_energy(
         mol_a, mol_b, core, forcefield, solvent_host_config, seed, n_frames=n_frames, prefix="solvent"
     )
@@ -40,7 +40,6 @@ def run_pair(mol_a, mol_b, core, forcefield_path, protein_path, n_frames, seed):
     complex_sys, complex_conf, _, _, complex_box, complex_top = builders.build_protein_system(protein_path)
     complex_box += np.diag([0.1, 0.1, 0.1])  # remove any possible clashes, deboggle later
     complex_host_config = HostConfig(complex_sys, complex_conf, complex_box)
-
     complex_res = estimate_relative_free_energy(
         mol_a, mol_b, core, forcefield, complex_host_config, seed + 1, n_frames=n_frames, prefix="complex"
     )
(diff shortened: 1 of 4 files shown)


timemachine/fe/rbfe.py
```diff
@@ -181,13 +174,28 @@ def generate_samples(st, host_config, temperature, lambda_schedule, n_frames, se
...
+
+
+def generate_samples_for_all_states(initial_states, protocol):
+    all_frames, all_boxes = [], []
+    for state_idx, state in enumerate(initial_states):
+        try:
+            frames, boxes = sample(state, protocol)
+        except Exception as exc:
+            # propagate the line information
+            raise ValueError(f"generate_samples_for_all_states failed on {state_idx}") from exc
```
proteneer, line 194: @mcwitt not sure if this is the most "pythonic" way of adding more information to an existing exception

examples/rbfe_edge_list.py
```diff
@@ -108,25 +107,44 @@ def read_from_args():
...
             pickle.dump(pkl_obj, fh)
 
-        tm_ddg = complex_ddg - solvent_ddg
-        tm_err = np.linalg.norm([complex_ddg_err, solvent_ddg_err])
+        if solvent_res.exception:
+            exception = "".join(
+                traceback.format_exception(
+                    type(solvent_res.exception), solvent_res.exception, solvent_res.exception.__traceback__
+                )
+            )
```
mcwitt, line 120: Idea to make this more "pythonic": instead of catching errors in `estimate_relative_free_energy` and passing them as values in the result object, could it work to catch and handle exceptions at the top layer (here)? E.g. thinking of something like this:
```suggestion
        try:
            solvent_res, solvent_top, complex_res, complex_top = fut.result()
        except FreeEnergyEstimationError: # custom exception class optional
            print(f"failed: {mol_a_name} -> {mol_b_name} solvent leg")
            traceback.print_exc()
```

timemachine/fe/rbfe.py
```diff
@@ -125,34 +85,70 @@ def __init__(self, omm_system, conf, box):
...
+        self.n_frames = n_frames
+        self.burn_in = burn_in
+        self.steps_per_frame = steps_per_frame
 
-    Returns
-    -------
-    3-tuple of all_frames, all_boxes, all_U_fns
-        all_frames has shape (L,T,N,3), all_boxes has shape (L,T,3,3), all_U_fns has shape (L,)
 
+class InitialState:
```
mcwitt, line 131: "State" seems like a pretty overloaded term. I wonder if there's something more specific we could use here? One thought is something like `TrajectorySeed` (I initially equated this with "thermodynamic state / potential" and was confused about where the rng seed was being stored)

mcwitt (approved):
Looks good, just had a few suggestions for naming and "pythonizing"

timemachine/fe/rbfe.py
```diff
@@ -125,34 +85,70 @@ def __init__(self, omm_system, conf, box):
...
+    """
+
+    def __init__(self, U_fns, integrator, barostat, x0, v0, box0, lamb):
+        self.U_fns = U_fns
+        self.integrator = integrator
+        self.barostat = barostat
+        self.x0 = x0
+        self.v0 = v0
+        self.box0 = box0
+        self.lamb = lamb
```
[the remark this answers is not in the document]
proteneer: willfix

examples/rbfe_edge_list.py
proteneer: When I wrote this i wasn't sure how exception propagation is handled through multiprocessing - so I wanted to store the exception objection explicitly. I'll play around and see

timemachine/fe/rbfe.py
proteneer: yeah I agree - I'll rename this

timemachine/fe/rbfe.py
proteneer: I'm not sure how to do this still, I want the subprocess behavior to be:

1) Return diagnostic information (InitialState) along with the original exception if the subprocess fails on the estimate (but not on the initial state generation)
2) Catch and store the initial state somewhere for this special type of exception

proteneer: (may best to offline this 2mrw)

examples/rbfe_edge_list.py
proteneer: I'm not sure how to do this properly still - let's offline this 2mrw

proteneer:
closing this - reopening a cleaned up one!

PR #785, opened 2022-08-01
Jax reference local resampling move

maxentile:
Minimal reference implementation of a `local_resampling_move`. At each iteration, this move defines a new sampling problem on a subset of particles in a system, where particles are probabilistically selected according to an arbitrary user-specified function.

(For example, particles may be selected based on distance to an interesting part of the system.)

This can be useful if running MCMC on a selected subset of the system is significantly more efficient than running MCMC on the whole system.

Theory notes:
* Note a pitfall (illustrated in the test) which can occur if the user-specified particle selection probabilities are not accounted for. (This differs from Gibbs sampling, Metropolis-within-Gibbs, etc., where the selected subset of problem variables is not conditioned on the current configuration `x`.)

Implementation notes:
* The interface of `local_resampling_move(x, target_logpdf_fxn, particle_selection_log_prob_fxn, mcmc_move)` is generic, but does not expose sufficient information for an efficient implementation. ( (1) Assumes target logpdf fxn is jax transformable if `mcmc_move` relies on gradients, (2) does not tell `mcmc_move` which particle positions are constant, (3) calls the full target logpdf fxn whenever `subproblem_logpdf` is evaluated, etc... Efficient implementations should apply optimizations to evaluate `subproblem_logpdf` at cost significantly lower than `target_logpdf_fxn`.)
* This implementation assumes the availability of valid MCMC moves `x_next, _ = mcmc_move(x, logpdf_fxn)`, but it will be important to investigate the practicality of unsound optimizations (such as using unadjusted Langevin as an approximate MCMC move)

### DIFF
--- a/tests/test_local_move.py
+++ b/tests/test_local_move.py
@@ -163,15 +163,18 @@ def test_ideal_gas():
         return np.mean(np.exp(central_particle_selection_log_prob_fxn(x)))
 
     def assert_correctness(local_move):
-        # expect no drift
+        # primary assertion: expect no drift in % of particles in resampled region
         traj, aux_traj = expect_no_drift(
             x0, local_move, observable_fxn=particle_frac_near_center, n_local_resampling_iterations=100
         )
 
-        # assert move was not trivial
+        # secondary assertion: confirm move was not trivial
         avg_accept_prob = np.mean([np.mean(np.exp(log_accept_probs)) for log_accept_probs in aux_traj])
-        assert avg_accept_prob > 0.1
-        assert np.max(np.abs(traj[0] - traj[-1])) > r0
+        assert avg_accept_prob > 0.1, "traj was probably trivial: didn't accept enough moves"
+        assert np.max(np.abs(traj[0] - traj[-1])) > r0, "traj was probably trivial: didn't move very far"
+        initially_within_region = vmap(lambda x_i: jnp.linalg.norm(delta_r(x_i, center, box)))(x0) < r0
+        all_moved = ((traj[-1] - traj[0])[initially_within_region] != 0).all()
+        assert all_moved, "traj was probably trivial: didn't update all particles in local region"
 
     # expect local move to be correct
     assert_correctness(correct_local_move)
--- a/tests/test_local_move.py
+++ b/tests/test_local_move.py
@@ -68,8 +68,8 @@ def expect_no_drift(x0, move_fxn, observable_fxn, n_local_resampling_iterations=
     deviated_by_50percent_or_more = (avg_at_start / avg_at_end) <= 0.5 or (avg_at_start / avg_at_end) >= 1.5
     if deviated_by_50percent_or_more:
         msg = f"""
-            observable avg over start frames = {100 * avg_at_start:.3f}%
-            observable avg over end frames = {100 * avg_at_end:.3f}%
+            observable avg over start frames = {avg_at_start:.3f}
+            observable avg over end frames = {avg_at_end:.3f}
             but averages of this (and all other observables) should be constant over time
         """
         raise RuntimeError(msg)
--- a/tests/test_local_move.py
+++ b/tests/test_local_move.py
@@ -1,0 +1,183 @@
+from jax import config, grad, jit
+from jax import numpy as jnp
+from jax import vmap
+
+config.update("jax_enable_x64", True)
+
+from functools import partial
+
+import numpy as np
+import pytest
+
+from timemachine.integrator import VelocityVerletIntegrator
+from timemachine.md.local_resampling import local_resampling_move
+from timemachine.potentials.jax_utils import delta_r
+
+
+def make_hmc_mover(x, logpdf_fxn, dt=0.1, n_steps=100):
+    masses = np.ones(len(x))
+
+    def force_fxn(x):
+        return -grad(logpdf_fxn)(x)
+
+    integrator = VelocityVerletIntegrator(force_fxn, masses=masses, dt=dt)
+
+    def augmented_logpdf(x, v):
+        return logpdf_fxn(x) - np.sum(0.5 * masses[:, np.newaxis] * v ** 2)
+
+    @jit
+    def augmented_proposal(x0, v0):
+        logp_before = augmented_logpdf(x0, v0)
+        x1, v1 = integrator._update_via_fori_loop(x0, v0, n_steps=n_steps)
+        logp_after = augmented_logpdf(x1, v1)
+
+        log_accept_prob = jnp.clip(jnp.nan_to_num(logp_after - logp_before, nan=-np.inf), a_max=0.0)
+
+        return (x1, v1), log_accept_prob
+
+    def hmc_move(x0):
+        v0 = np.random.randn(*x.shape)
+        (x1, v1), log_accept_prob = augmented_proposal(x0, v0)
+
+        if np.random.rand() < jnp.exp(log_accept_prob):
+            return x1, log_accept_prob
+        else:
+            return x0, log_accept_prob
+
+    return hmc_move
+
+
+def expect_no_drift(x0, move_fxn, observable_fxn, n_local_resampling_iterations=100):
+    traj = [jnp.array(x0)]
+    aux_traj = []
+
+    for _ in range(n_local_resampling_iterations):
+        updated, aux = move_fxn(traj[-1])
+
+        traj.append(updated)
+        aux_traj.append(aux)
+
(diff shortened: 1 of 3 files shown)


badisa (approved):

tests/test_local_move.py
```diff
@@ -0,0 +1,183 @@
...
+
+    for _ in range(n_local_resampling_iterations):
+        updated, aux = move_fxn(traj[-1])
+
+        traj.append(updated)
+        aux_traj.append(aux)
+
+    expected_selection_fraction_traj = np.array([observable_fxn(x) for x in traj])
+
+    # TODO: don't hard-code T, thresholds, etc.
```
mcwitt, line 62: reminder: TODO

tests/test_local_move.py
```diff
@@ -0,0 +1,183 @@
...
+        return (x1, v1), log_accept_prob
+
+    def hmc_move(x0):
+        v0 = np.random.randn(*x.shape)
+        (x1, v1), log_accept_prob = augmented_proposal(x0, v0)
+
+        if np.random.rand() < jnp.exp(log_accept_prob):
+            return x1, log_accept_prob
+        else:
+            return x0, log_accept_prob
```
mcwitt, line 45: super nit: maybe
```suggestion
        x_new = x1 if np.random.rand() < jnp.exp(log_accept_prob) else x0
        return x_new, log_accept_prob
```

tests/test_local_move.py
```diff
@@ -0,0 +1,183 @@
...
+
+    expected_selection_fraction_traj = np.array([observable_fxn(x) for x in traj])
+
+    # TODO: don't hard-code T, thresholds, etc.
+    T = 10
+    assert n_local_resampling_iterations > 2 * T
+    avg_at_start = np.mean(expected_selection_fraction_traj[:T])
+    avg_at_end = np.mean(expected_selection_fraction_traj[-T:])
+
+    deviated_by_50percent_or_more = (avg_at_start / avg_at_end) <= 0.5 or (avg_at_start / avg_at_end) >= 1.5
```
mcwitt, line 68: The ratios here seem flipped if we mean deviated from start to end by 50% or more?

tests/test_local_move.py
```diff
@@ -0,0 +1,183 @@
...
+    selection_probs = np.exp(particle_selection_log_prob_fxn(x))
+    assert np.min(selection_probs) >= 0 and np.max(selection_probs) <= 1, "selection_probs must be in [0,1]"
+    assert selection_probs.shape == (n_particles,), "must compute per-particle selection_probs"
+    selection_mask = np.random.rand(n_particles) < selection_probs  # TODO: factor out dependence on global numpy rng?
+
+    # NOTE: missing restraint! will result in incorrect sampling
+
+    # def restrained_logpdf_fxn(x) -> float:
+    #    log_p_i = particle_selection_log_prob_fxn(x)
+    #    return target_logpdf_fxn(x) + bernoulli_logpdf(log_p_i, selection_mask)
```
mcwitt, line 101: nit suggestion: could it make sense for `local_resampling_move` to take `restrained_logpdf_fxn` as an argument, to avoid this test needing to replicate it?

tests/test_local_move.py
```diff
@@ -0,0 +1,181 @@
...
+    def assert_correctness(local_move):
+        # expect no drift
+        traj, aux_traj = expect_no_drift(
+            x0, local_move, observable_fxn=particle_frac_near_center, n_local_resampling_iterations=100
+        )
+
+        # assert move was not trivial
+        avg_accept_prob = np.mean([np.mean(np.exp(log_accept_probs)) for log_accept_probs in aux_traj])
+        assert avg_accept_prob > 0.1
+        assert np.max(np.abs(traj[0] - traj[-1])) > r0
```
proteneer, line 174: what is the purpose of this test? this seems to check that the first last frame has at least one particle that moved more than `r0`?

proteneer: also, `np.abs(x0 - x1)` i think returns maximum displacement along any dimension `(x,y,z)` and not the actual distance.

tests/test_local_move.py
maxentile: I would prefer not to make changes of this flavor (at least in this implementation), as a main purpose of this reference implementation is to show a recipe for combining arbitrary user-defined functions to produce a valid move.

Currently, for arbitrary* user-specified functions, the reference implementation will still be a valid MCMC move.

However, if the signature were modified to permit independent specification of the restraint and `particle_selection_log_prob_fxn`, the resulting move could have incorrect behavior (such as the pitfall shown in the test when `restrained_logpdf_fxn(x) = target_logpdf_fxn(x)`, independent of `particle_selection_log_prob_fxn`).

The additional duplication in the test doesn't seem very costly or easily avoidable, imo.

(*Main unchecked assumption about the user input: `mcmc_move(x, log_target) -> x_next` is a valid MCMC move.)

tests/test_local_move.py
maxentile: To assert that the move was not trivial.

maxentile: Is there a more stringent thing that would be easy to assert?

maxentile: (Later assertions may pass trivially if `traj[t]` is constant as a function of `t`)

tests/test_local_move.py
mcwitt: That makes sense. I was thinking more along the lines of not modifying the public interface at all, rather just splitting the implementation into a composition of private functions. But I agree that it doesn't seem like there's much cost to the duplication.

maxentile: >not modifying the public interface at all, rather just splitting the implementation into a composition of private functions. 

Ahh! Will think if there's a cleaner way to do this without compromising generality / correctness.

If there's a good way to split it up, another possible benefit would be to exploit fast specialized implementations of the subproblem logpdfs...

tests/test_local_move.py
proteneer: hm not sure either, maybe check that at least `x` number of particles have moved outside of `r0` (as opposed to just 1 particle)

proteneer: Anyways - not going to be picky about this. Please don't let this block the merge

proteneer (approved):
lgtm

mcwitt (approved):

tests/test_local_move.py
maxentile: Sorry, missed replying in proper detail!

The primary purpose of the test is to assert that even though we are repeatedly introducing a "selection bias" (only moving particles that are in some specified region of the box), the average number of ideal gas particles in that region does not decrease over time. Naively, the average number of particles in the resampled region would eventually decrease to 0.

This secondary assertion is intended to enforce that the primary assertion is not passing trivially. (One way to trivially satisfy an assertion of "no drift in this specific observable" is to return a constant trajectory -- "no drift in any observable." Want the test to rule out this possibility, since you could accidentally be in this situation if, for example, the test accidentally used an overly large step size, so that 0% of proposed moves were being accepted.) 

Another, stronger indication that the primary assertion is not passing trivially is that the primary assertion fails if the same experiment is performed using all the same sampler settings etc., just commenting out the restraint (https://github.com/proteneer/timemachine/blob/b1deb66cc5ae436a6612c44d9fd5637bcf941471/tests/test_local_move.py#L179-L181 ).

I'll also make this secondary assertion stronger for completeness (assert that all particles in the resampled region have moved at least a little).

maxentile: Applied in https://github.com/proteneer/timemachine/pull/785/commits/8709b31af63867a49a3365e3a0570cb51f735ba6

maxentile:
Thanks for the reviews + discussions before the PR.

I'll go ahead and merge, but note there's a change I want to make to the signature, based on [this open thread](https://github.com/proteneer/timemachine/pull/785#discussion_r936015559) with @mcwitt , so that instead of accepting a `target_logpdf_fxn: x -> float`, it accepts a `target_logpdf_factory_fxn: (x0, selection_mask) -> (x_selected -> float)`.

(By default it would fall back to the unoptimized implementation `target_logpdf_fxn(x.at[selection_mask].set(x_selected))` hard-coded in the current PR, but it could also exploit optimized implementations if provided.)

tests/test_local_move.py
maxentile: Good call -- fixed in https://github.com/proteneer/timemachine/pull/785/commits/5aceb6884604415c9e88ef2d17e06aeda4f34e60

PR #786, opened 2022-08-01
Add deterministic interface to reference LangevinIntegrator, XLA-accelerated `multiple_steps`

mcwitt:
Adds a new method, `multiple_steps_deterministic`, to `LangevinIntegrator`.

The main API difference vs. the existing `multiple_steps` is that `multiple_steps_deterministic` also accepts a RNG key instance, allowing for determinism.

The main motivation for this is a ~60% speedup ([example](https://gitlab.com/relaytx/users/mwittmann/notebooks/-/blob/main/experiments/20220801-deterministic-langevin/readme.ipynb)) gained by implementing `multiple_steps_deterministic` in terms of `jax.lax.scan` (after applying `jax.jit`). The new implementation also works with `vmap` and `pmap`.

### DIFF
--- a/tests/test_reference_langevin_integrator.py
+++ b/tests/test_reference_langevin_integrator.py
@@ -58,28 +58,52 @@ def test_reference_langevin_integrator(threshold=1e-4):
 
 def test_reference_langevin_integrator_deterministic():
     """
-    Asserts that trajectories produced by `_deterministic` methods
-    are reproducible, and that the implementation in terms of the XLA
-    scan primitive gives a consistent result to that of the
-    implementation as a Python for-loop
+    Asserts that trajectories produced by `_deterministic` methods are deterministic
     """
     force_fxn = lambda x: -4 * x ** 3
     langevin = LangevinIntegrator(force_fxn, masses=1.0, temperature=300.0, dt=0.1, friction=1.0)
     x0, v0 = 0.1 * jax.random.uniform(jax.random.PRNGKey(1), shape=(2, 5))
+    key, key_ = jax.random.split(jax.random.PRNGKey(1))
 
-    xs1, vs1 = langevin.multiple_steps_deterministic(jax.random.PRNGKey(1), x0, v0)
+    def assert_deterministic(method):
+        xs1, vs1 = method(key, x0, v0)
 
-    # implementation based on jax.lax.scan should give identical result
-    xs2, vs2 = langevin.multiple_steps_deterministic_lax(jax.random.PRNGKey(1), x0, v0)
+        # same key should give same result
+        xs2, vs2 = method(key, x0, v0)
+        np.testing.assert_array_equal(xs1, xs2)
+        np.testing.assert_array_equal(vs1, vs2)
 
+        # different key should give different result
+        xs3, vs3 = method(key_, x0, v0)
+        assert not np.allclose(xs2, xs3)
+        assert not np.allclose(vs2, vs3)
+
+    assert_deterministic(langevin.multiple_steps_deterministic)
+    assert_deterministic(langevin.multiple_steps_deterministic_lax)
+
+
+def test_reference_langevin_integrator_consistent():
+    """
+    Asserts that the implementation in terms of jax.lax primitives
+    gives a consistent result to that of the implementation as a
+    Python for-loop
+    """
+    force_fxn = lambda x: -4 * x ** 3
+    langevin = LangevinIntegrator(force_fxn, masses=1.0, temperature=300.0, dt=0.1, friction=1.0)
+    x0, v0 = 0.1 * jax.random.uniform(jax.random.PRNGKey(1), shape=(2, 5))
+    key = jax.random.PRNGKey(1)
+
+    xs1, vs1 = langevin.multiple_steps_deterministic(key, x0, v0)
+
+    # should be consistent with the jax.lax implementation
+    xs2, vs2 = langevin.multiple_steps_deterministic_lax(key, x0, v0)
+
+    # NOTE: result of the jax.lax implementation is NOT bitwise
+    # equivalent to the pure Python implementation. This might be due
+    # to loop-unrolling and reassociation optimizations performed by
+    # XLA
     np.testing.assert_allclose(xs1, xs2)
     np.testing.assert_allclose(vs1, vs2)
-
-    # different seed; should give a different result
-    xs3, vs3 = langevin.multiple_steps_deterministic_lax(jax.random.PRNGKey(2), x0, v0)
-
-    assert not np.allclose(xs1, xs3)
-    assert not np.allclose(vs1, vs3)
 
 
 def test_reference_langevin_integrator_with_custom_ops():
--- a/tests/test_reference_langevin_integrator.py
+++ b/tests/test_reference_langevin_integrator.py
@@ -1,6 +1,9 @@
 from itertools import product
 
 import jax
+
+jax.config.update("jax_enable_x64", True)
+
 import numpy as np
 from jax import grad, jit
 from jax import numpy as jnp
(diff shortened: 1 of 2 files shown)


timemachine/integrator.py
```diff
@@ -78,15 +79,59 @@ def __init__(self, force_fxn, masses, temperature, dt, friction):
         # note: per-atom frictions allowed
         self.ca, self.cb, self.cc = np.expand_dims(ca, -1), np.expand_dims(cb, -1), np.expand_dims(cc, -1)
 
-    def step(self, x, v):
+    def _step(self, x, v, eta):
```
badisa, line 82: Perhaps noise would be a better name? eta doesn't mean anything to me

tests/test_reference_langevin_integrator.py
```diff
@@ -52,6 +53,32 @@ def test_reference_langevin_integrator(threshold=1e-4):
...
+    are reproducible, and that the implementation in terms of the XLA
+    scan primitive gives a consistent result to that of the
+    implementation as a Python for-loop
+    """
+    force_fxn = lambda x: -4 * x ** 3
+    langevin = LangevinIntegrator(force_fxn, masses=1.0, temperature=300.0, dt=0.1, friction=1.0)
+    x0, v0 = 0.1 * np.ones((2, 5))
+
+    # private "reference" implementation (not using XLA primitives)
+    xs1, vs1 = langevin._multiple_steps_deterministic(jax.random.PRNGKey(1), x0, v0)
```
badisa, line 68: Is there value having the non-deterministic implementation? Seems like it would be better to have determinism all the time than to have two implementations, especially if the deterministic one can be made faster than the non-deterministic one.

timemachine/integrator.py
```diff
@@ -78,15 +79,59 @@ def __init__(self, force_fxn, masses, temperature, dt, friction):
...
         return new_x, new_v
 
+    def step(self, x, v):
+        return self._step(x, v, np.random.randn(*x.shape))
+
+    def step_deterministic(self, key, x, v):
+        """Return copies x and v, updated by a single timestep. Accepts a PRNGKey for determinism."""
+        return self._step(x, v, jax.random.normal(key, x.shape))
+
+    def _multiple_steps_deterministic(self, key, x, v, n_steps=1000):
```
badisa, line 98: If this is only for testing, perhaps move to the test file?

tests/test_reference_langevin_integrator.py
```diff
@@ -52,6 +53,32 @@ def test_reference_langevin_integrator(threshold=1e-4):
...
+
+    # should give the same result using XLA primitives
+    xs2, vs2 = langevin.multiple_steps_deterministic(jax.random.PRNGKey(1), x0, v0)
+
+    # different seed; should give a different result
+    xs3, vs3 = langevin.multiple_steps_deterministic(jax.random.PRNGKey(2), x0, v0)
+
+    assert np.allclose(xs1, xs2)
```
badisa, line 76: Should this be `allclose` or `np.all(xs1 == xs2)`, is this not bitwise deterministic?

mcwitt: Good catch. The answer is that they are not currently bitwise equivalent, but I'm not sure why! Will investigate..

tests/test_reference_langevin_integrator.py
mcwitt: I like the idea of making everything deterministic, but we haven't fully bought in to the `jax.random` API, and as it is removing the non-deterministic version would break a fair amount of existing code.

timemachine/integrator.py
maxentile: Is it also useful in cases where `self.force_fxn` is not jax transformable, but we wish to use jax keyed rng instead of numpy global rng?

maxentile (approved):

timemachine/integrator.py
mcwitt: > Is it also useful in cases where self.force_fxn is not jax transformable

ah, that's a good point, I think so. With that in mind it might be best to make it "public" and update the docstring to mention this use case

timemachine/integrator.py
mcwitt: Thanks, done in 6a11e1e

tests/test_reference_langevin_integrator.py
badisa: Perhaps modify the constructor to take a seed and either use the jax random API or the numpy random depending on the implementation. That will make it inline with our C++ Langevin integrator and give us determinism.

timemachine/integrator.py
badisa: Would it be better than to have one implementation that you can manually jax transform if you need it, rather than having two identical implementations where one is transformed and the other isn't?

badisa (approved):

tests/test_reference_langevin_integrator.py
mcwitt: After looking into this a bit, I think we cannot expect bitwise equivalence between the pure-Python loop and `jax.lax.scan`, because the latter may perform some optimizations that break floating-point equivalence (e.g. loop unrolling + reassociation).

timemachine/integrator.py
mcwitt: Done in b1b50ce and f6e6c7e

tests/test_reference_langevin_integrator.py
mcwitt: > Perhaps modify the constructor to take a seed and either use the jax random API or the numpy random depending on the implementation

I like the idea of making the jax reference `LangevinIntegrator` more consistent with the C++ implementation, but I think I'd actually prefer modifying the signature of the C++ version to accept `seed` as an argument to `step_fwd` (and `Context::multiple_steps`), thus moving away from keeping RNGs as hidden internal state of objects. One case where this could be useful is sampling independent chains, where we could avoid initializing a new integrator object for each chain.

To make the jax and numpy APIs more consistent, I added a `StochasticIntegrator` base class in 86a5f18. This allows explicitly passing an RNG instance to `multiple_steps` for determinism when using the non-jax API.

mcwitt:
@maxentile @badisa since approval I've refactored this a bit to address comments about providing a deterministic interface for the non-jax version of `multiple_steps`, and avoiding near-identical versions of `multiple_steps` in the library code. Would appreciate another quick look when you have a chance.

badisa (approved):
lgtm

timemachine/integrator.py
```diff
@@ -66,7 +71,55 @@ def multiple_steps(self, x, v, n_steps=1000):
         return np.array(xs), np.array(vs)
 
 
-class LangevinIntegrator(Integrator):
+class StochasticIntegrator(ABC):
+    @abstractmethod
+    def step(self, x, v, rng) -> Tuple[Any, Any]:
```
maxentile, line 76: I think this looks good!

Comment (not a change request): Another alternative might be to have `rng` be an instance attribute referenced by `step` implementation (`self.rng.randn(...)`), rather than an explicit argument of the `step` function. (Cons: This wouldn't make `step` deterministic, and it wouldn't be consistent with the jax rng approach. Pros: It would avoid dependence on global rng state, and it would be [similar to the C++ interface](https://github.com/proteneer/timemachine/blob/5aa164035823e13d58aff2c01ad2d71223c25dd0/timemachine/cpp/src/integrator.hpp#L26-L29), where `rng` is a private class member and `seed` can only be set in constructor.)

maxentile (approved):
Sorry for my delay! These updates look good to me.

timemachine/integrator.py
mcwitt: > Another alternative might be to have rng be an instance attribute referenced by step implementation

I think @badisa suggested this too (https://github.com/proteneer/timemachine/pull/786#discussion_r935629249). My inclination is to avoid having the RNG as an instance variable, since it seems to reduce flexibility a bit (https://github.com/proteneer/timemachine/pull/786#discussion_r935944325) relative to allowing RNG instances to be passed to member functions. It does seem preferable to global RNG state though (but I'd be in favor of adopting something like the more explicit style encouraged by jax).

PR #788, opened 2022-08-04
Atom mapping improvements

proteneer:
This PR implements a slight improvement to the previous MCS protocol by removing the need for conformation alignment during the initial core smarts matching.  It also resolves a subtle issue where sometimes a single bond core is generated.

We still take the argmin of all the distance of all possible matches in a subsequent step pruning set. On the hif2a set, it reduces the average number of dummy atoms from about 14 to 7. This still doesn't check for chiral restraints (separate PR in the works). This still prevents bond breaking on the core atoms.


jkausrelay (approved):
LGTM!

PR #789, opened 2022-08-09
C++ Velocity Verlet

badisa:
Implements Velocity Verlet in C++

- Adds a `finalize` method to context for finalizing the xs and velocities
- Moves the responsibility of the forces to the integrator rather than the context (which seems cleaner and more inline with our reference integrators)
- Integrators handle du_dl computation to avoid performance hit, though likely desirable to remove if we remove du_dl.


timemachine/cpp/src/context.cu
```diff
@@ -76,7 +76,7 @@ Context::multiple_steps(const std::vector<double> &lambda_schedule, int store_du
     int box_buffer_size = x_buffer_size * 3 * 3;
 
     std::vector<double> h_x_buffer(x_buffer_size * N_ * 3);
-
+    cudaStream_t stream = static_cast<cudaStream_t>(0);
```
badisa, line 77: This commit could be dropped. Was from my Local MD PR and had thought it would be more useful. I left it in as I find it more flexible.

timemachine/cpp/src/integrator.cu
```diff
@@ -80,78 +104,9 @@ void LangevinIntegrator::step_fwd(
...
 
     update_forward<double>
-        <<<dimGrid_dx, tpb, 0, stream>>>(N_, D, ca_, d_cbs_, d_ccs_, d_noise_, d_x_t, d_v_t, d_du_dx_t, dt_);
+        <<<dimGrid_dx, tpb, 0, stream>>>(N_, D, ca_, d_cbs_, d_ccs_, d_noise_, d_x_t, d_v_t, d_du_dx_, dt_);
 
     gpuErrchk(cudaPeekAtLastError());
 }
 
 } // end namespace timemachine
-
```
badisa, line 89: Can add this back in, removed out of habit

timemachine/cpp/src/integrator.hpp
```diff
@@ -9,8 +11,14 @@ class Integrator {
 public:
     virtual ~Integrator(){};
 
-    virtual void
-    step_fwd(double *d_x_t, double *d_v_t, unsigned long long *d_du_dx_t, double *d_box_t_, cudaStream_t stream) = 0;
+    virtual void step_fwd(
+        std::vector<BoundPotential *> bps,
```
badisa, line 14: &bps

tests/test_velocity_verlet_integrator.py
```diff
@@ -0,0 +1,142 @@
...
+from timemachine.lib.potentials import SummedPotential
+from timemachine.testsystems.relative import hif2a_ligand_pair
+
+
+def setup_velocity_verlet(bps, x0, box, dt, masses):
+    integrator = VelocityVerletIntegrator(dt, masses)
+    # return integrator impl to avoid deallocating
+    intg = integrator.impl()
+    context = custom_ops.Context(x0, np.zeros_like(x0), box, intg, bps)
+    return intg, context
```
mcwitt, line 16: Would it be worth adding a wrapper for `Context` to the Python API to avoid needing this workaround? Something like

```python
class Context(custom_ops.Context):
    def __init__(self, x0, v0, box, integ, bps):
        self._integ = integ
        super().__init__(x0, v0, box, integ, bps)
```
and then
```suggestion
    return custom_ops.Context(x0, np.zeros_like(x0), box, integrator.impl(), bps)
```

timemachine/cpp/src/context.hpp
```diff
@@ -21,6 +21,7 @@ class Context {
     ~Context();
 
     void step(double lambda);
+    void finalize(double lambda);
```
mcwitt, line 24: I might be missing some subtlety, but is there a reason not to just call `finalize` automatically at the end of `step` and `multiple_steps`? It seems like this is what we do in the reference implementation (https://github.com/proteneer/timemachine/blob/b548e75f226d65c502b0a53d1a53918a051bc0c9/timemachine/integrator.py#L190)

badisa: It is called for `multiple_steps`, or really it calls the finalize on the integrator directly.

`step` is different in that we don't want to run 2 force calculation per step which is what would happen (actually three with initialization) if you called `finalize` every step. Would also require a different code path as its two half steps instead of a half step, full steps then half step. 

Basically the methods that take multiple steps should correctly finalize, but if you call step manually, to ensure correct coords/velocities you have to call finalize.

tests/test_velocity_verlet_integrator.py
badisa: Makes sense to me, will give that a shot.

timemachine/cpp/src/verlet_integrator.cu
```diff
@@ -0,0 +1,107 @@
+
```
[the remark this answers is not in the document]
badisa: I would be fine putting them in the same file or otherwise. @maxentile could probably comment on the desire for additional variants.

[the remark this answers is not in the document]
maxentile: I don't foresee a situation where it would be confusing to use terms "Verlet" and "velocity Verlet" interchangeably in this context.

Main variation I can currently foresee is to potentially simplify the staggering-by-half-a-timestep complication in the trajectory returned by multiple steps (e.g. https://github.com/proteneer/timemachine/pull/783), but that would still be velocity verlet, just with a slightly different interface.

tests/test_reference_velocity_verlet_integrator.py
```diff
@@ -116,14 +116,8 @@ def force(coords):
...
 
         # check "public" .step and .multiple_steps implementations
-        assert_reversibility_using_step_implementations(intg, x0, v0, n_steps, atol=1e-10)
-
-    # TODO: possibly investigate why n_steps = 10000 fails
-    # assert_reversibility_using_step_implementations(intg, x0, v0, n_steps=10000, atol=0.1)
-    # * also fails with reduced dt = 1.0e-3
-    # * passes with greatly reduced dt = 1.0e-4
-    # TODO: does replacing np.sum(du_dxs, 0) with SummedPotential address this?
+        assert_reversibility_using_step_implementations(intg, x0, v0, n_steps, atol=1e-11)
```
maxentile, line 123: Nice!

timemachine/cpp/src/kernels/k_integrator.cuh
```diff
@@ -0,0 +1,83 @@
...
+    RealType force = FIXED_TO_FLOAT<RealType>(du_dx[local_idx]);
+
+    v_t[local_idx] += (0.5 * cbs[atom_idx]) * force;
+    if (UPDATE_X) {
+        x_t[local_idx] += dt * v_t[local_idx];
+    }
+};
+
+template <typename RealType>
+__global__ void update_forward_verlocity_verlet(
```
maxentile, line 63: Nit: Typo

timemachine/lib/__init__.py
```diff
@@ -20,6 +20,18 @@ def impl(self):
...
+class VelocityVerletIntegrator:
+    def __init__(self, dt, masses):
+        self.dt = dt
+
+        cb = dt / masses
+        cb *= -1
+        self.cbs = cb
+
+    def impl(self):
+        return custom_ops.VelocityVerletIntegrator(self.dt, self.cbs)
```
maxentile, line 32: Not a huge fan of the `self.cb`, `self.cbs` naming convention / sign conventions here and in the other implementations -- seems like this is exposing an internal implementation detail that could be kept internal.

Would it still be convenient if these both had the same constructor arguments: `custom_ops.VelocityVerletIntegrator(dt, masses)`, rather than `custom_ops.VelocityVerletIntegrator(dt, -dt/masses)`?

tests/test_velocity_verlet_integrator.py
```diff
@@ -0,0 +1,142 @@
...
+
+    # Is not infinitely reversible, will fail after 5000 steps due to accumulation of coords/velos in float
+    for n_steps in [1, 10, 100, 500, 1000, 3000]:
+        v0 = np.random.randn(*coords.shape)
+        intg, ctxt = setup_velocity_verlet(bound_potentials, coords, box, dt, masses)  # noqa
+        ctxt.set_v_t(v0)
+
+        # check "public" .step and .multiple_steps implementations
+        assert_reversibility_using_step_implementations(ctxt, n_steps, atol=1e-10)
+
```
maxentile, line 101: Would there be utility in adding a test to exercise the case of `multiple_steps(lambdas)`, where `lambdas` is not `np.ones(n_steps) * lamb`?

(I think we should still be able to assert reversibility if `lambdas = np.linspace(0,1,n_steps)`, `lambdas_rev = lambdas[::-1]` or so, but it may require some care with `finalize` etc.)

tests/test_velocity_verlet_integrator.py
```diff
@@ -0,0 +1,142 @@
...
+    unbound_potentials, sys_params, masses = rfe.prepare_vacuum_edge(rfe.ff.get_ordered_params())
+    coords = rfe.prepare_combined_coords()
+    bound_potentials = [
+        ubp.bind(params).bound_impl(np.float32) for (ubp, params) in zip(unbound_potentials, sys_params)
+    ]
+    box = 100 * np.eye(3)
+
+    dt = 1.5e-3
+
+    # Is not infinitely reversible, will fail after 5000 steps due to accumulation of coords/velos in float
```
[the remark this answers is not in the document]
badisa: Correct, though to do that would require more extensive changes to the context/integrators and @proteneer said we should only do that if you thought that mattered, which offline you didn't seem to think was an issue.

timemachine/cpp/src/kernels/k_integrator.cuh
badisa: Ugh... if only I didn't want to type verlocity...

timemachine/lib/__init__.py
badisa: I was trying to stay consistent with the Langevin integrator. We could move the pre-processing of the reused parameters to the C++. And would you suggest something else for the name, I certainly don't find it very obvious what it is?

tests/test_velocity_verlet_integrator.py
badisa: Will switch out to a linspace steps and see. Seems like it should be reversible in that case, though perhaps there is nuance I am missing.

tests/test_velocity_verlet_integrator.py
```diff
@@ -0,0 +1,142 @@
...
+        context.set_v_t(v)
+        _, xs, _ = context.multiple_steps(np.ones(n_steps) * lamb)
+        v = context.get_v_t()
+        return xs[-1], v
+
+    assert_reversible(x0, v0, multiple_steps_update, atol=atol)
+
+
+def test_reversibility():
+    """Check reversibility of "public" .step and .multiple_steps implementations when constructing a Context with"""
```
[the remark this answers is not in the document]
badisa: Fixed in 0f10a29741a2bb095c59fd63a7ebb16bf619758c

tests/test_velocity_verlet_integrator.py
```diff
@@ -0,0 +1,142 @@
...
+        x_next, v_next = update_fxn(x, v)
+        return x_next, -v_next
+
+    close = partial(np.allclose, atol=atol)
+
+    # assert "self_inverse" is really its own inverse
+    x1, v1 = self_inverse(x0, v0)
+    x0_, v0_ = self_inverse(x1, v1)
+
+    assert close(x0_, x0), f"max(abs(x0 - x0_)) = {np.max(np.abs(x0 - x0_))}"
```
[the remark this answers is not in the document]
badisa: Switched to `assert_allclose` in 0f10a29741a2bb095c59fd63a7ebb16bf619758c

timemachine/cpp/src/verlet_integrator.cu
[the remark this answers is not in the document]
mcwitt: Makes sense. Is it worth then updating the filename to match the class it defines? (nit)

tests/test_velocity_verlet_integrator.py
badisa: Created an issue for this https://github.com/proteneer/timemachine/issues/791 as it seems out of scope for this PR (especially since this is a test function) and is a broader problem we should resolve.

timemachine/cpp/src/context.hpp
mcwitt: Ah, I see. I guess I'm still a little bit unclear on the use cases for `step`, since it seems like we would prefer `multiple_steps` in most cases vs. running `step` inside of a loop`? IIUC the reference integrator does finalize after each step (https://github.com/proteneer/timemachine/blob/b548e75f226d65c502b0a53d1a53918a051bc0c9/timemachine/integrator.py#L159)

timemachine/cpp/src/verlet_integrator.cu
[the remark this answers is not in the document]
badisa: Inclined to leave it, but only because the longer file name doesn't provide value when the class name is qualified and we refer to BAOAB as langevin even though there are many variants of langevin.

tests/test_velocity_verlet_integrator.py
badisa: Fixed in https://github.com/proteneer/timemachine/commit/0f10a29741a2bb095c59fd63a7ebb16bf619758c

Also added multiple_steps_U, which used a fixed lambda since it only runs equilibrium simulations

timemachine/cpp/src/context.hpp
badisa: Not sure we rely on `step` besides as a means of verification in some of our tests. I think there was also some use in the non-equilibrium stuff that is now attic'd.

The reference integrator step is different. It takes two half steps, while the C++ step is a half step, followed by a step, with finalize doing another half step.  We could have it so that on the first step of the integrator it only takes a half step, followed by full steps after that. Would make it more consistent. Will give that a shot

We could also store the last force computation like @maxentile did here: https://github.com/proteneer/timemachine/pull/783 though that doubles the memory requirement of the C++ implementation for no benefit besides .

To make `step` match for both the reference and C++ would be difficult as we would need two code paths which seems less desirable especially if `step` is not our preferred method.

timemachine/lib/__init__.py
maxentile: >We could move the pre-processing of the reused parameters to the C++.

Non-blocking of course, but that sounds preferable to me (perhaps for a later PR, since it might involve either introducing an inconsistency with Langevin or making more changes to Langevin).

(I also may be missing if there's a reason it's more convenient to do these computations in the Python layer.)

>And would you suggest something else for the name, I certainly don't find it very obvious what it is?

No preference. I hadn't ever thought to name `dt / m` (or `-dt / m`), and I assumed the names `ca`, `cb`, `cc` from `langevin_coefficients` were arbitrary.

timemachine/cpp/src/context.hpp
badisa: Doing a half step on the first call to `step` makes things non-reversible unless we want to add an `initialize` method as well. Which I had considered, and decided against to avoid the additional line required to make valid steps. Could easily add though.

I am comfortable with this as it seems correct and agrees with the reference, even if it is not identical in its implementation.

badisa: Thinking about it more, there will probably always be a discontinuity between the reference and the C++. The reference is both integrator and context.

In the reference it knows to do exactly N steps, which is made up of 2 half steps and N - 1 full steps. The context on the other hand doesn't know anything about the integrator and is going to run N steps regardless of integrator, which in the case of Velocity Verlet results in N steps plus 2 half steps.

Which leads me to think agreement between the two implementations is the crucial part to verify.

@maxentile Do you have thoughts on this?

timemachine/lib/__init__.py
badisa: Created issue https://github.com/proteneer/timemachine/issues/792

timemachine/cpp/src/context.hpp
maxentile: Mostly agreed with the comments above.

* I don't know of any use cases where Python `.step` is preferred over `.multiple_steps`, besides testing.
* Making the internal implementations of `.step` match does not seem necessary -- agreement of the final state (xT, vT) of the trajectory seems the most important part to verify.
* Would adding an `initialize` method to Integrator base class help avoid the discontinuity you mention in https://github.com/proteneer/timemachine/pull/789#discussion_r942656625 ? (Context could do `initialize; step in a loop; finalize;` -- where `initialize` and `finalize` would be empty for some integrators.)

badisa: > Would adding an initialize method to Integrator base class help avoid the discontinuity you mention in https://github.com/proteneer/timemachine/pull/789#discussion_r942656625 ? (Context could do initialize; step in a loop; finalize; -- where initialize and finalize would be empty for some integrators.)

I think so, will add an `initialize` method.

badisa: Fixed in a1abf0ba37b3a35947907e41dba129d48c2860c5

tests/test_velocity_verlet_integrator.py
```diff
@@ -0,0 +1,153 @@
...
+        (expected to be an "involution" i.e. its own inverse)"""
+        x_next, v_next = update_fxn(x, v, lamb_sched)
+        return x_next, -v_next, lamb_sched[::-1]
+
+    # assert "self_inverse" is really its own inverse
+    x1, v1, rev_lamb = self_inverse(x0, v0, lambdas)
+    x0_, v0_, _ = self_inverse(x1, v1, rev_lamb)
+
+    np.testing.assert_almost_equal(x0_, x0, decimal=10)
+    np.testing.assert_almost_equal(v0_, v0, decimal=10)
```
mcwitt, line 37: nit: from the [docs](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html):

> It is recommended to use one of [assert_allclose](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose), [assert_array_almost_equal_nulp](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_almost_equal_nulp.html#numpy.testing.assert_array_almost_equal_nulp) or [assert_array_max_ulp](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_max_ulp.html#numpy.testing.assert_array_max_ulp) instead of this function for more consistent floating point comparisons.

Should we standardize on `assert_allclose` since we're using it elsewhere?

tests/test_velocity_verlet_integrator.py
```diff
@@ -0,0 +1,153 @@
...
+    intg_impl, ctxt = setup_velocity_verlet([bound_summed], coords, box, dt, masses)  # noqa
+
+    ctxt.set_v_t(v0)
+
+    n_steps = 10
+
+    # Add one step as the C++ context does N steps + 2 half steps (initialize, finalize)
+    ref_xs, ref_vs = intg.multiple_steps(coords, v0, n_steps=n_steps + 1)
+    np.testing.assert_array_almost_equal(ref_xs[0], coords, decimal=10)
+    np.testing.assert_array_almost_equal(ref_xs[0], ctxt.get_x_t(), decimal=10)
```
mcwitt, line 146: nit: `assert_allclose` here?

tests/test_velocity_verlet_integrator.py
```diff
@@ -0,0 +1,153 @@
...
+
+    _, xs, _ = ctxt.multiple_steps(np.ones(n_steps) * lamb, 0, 1)
+    assert xs.shape[0] == n_steps
+    v1 = ctxt.get_v_t()
+    atol = 1e-5
+    np.testing.assert_allclose(ref_xs[-1], xs[-1], atol=atol)
```
mcwitt, line 152: Why not compare the whole trajectory?
```suggestion
    np.testing.assert_allclose(ref_xs, xs, atol=atol)
```

timemachine/cpp/src/context.hpp
mcwitt: This solution makes sense to me. Longer term, I wonder whether the C++ integrators could own `multiple_steps` (similarly to the reference) to avoid needing to expose `initialize` and `finalize`.

timemachine/cpp/src/verlet_integrator.cu
```diff
@@ -0,0 +1,119 @@
...
+            lamb,
+            d_du_dx_, // we only need the forces
+            nullptr,
+            nullptr,
+            nullptr,
+            stream);
+    }
+    half_step_velocity_verlet<double, true><<<dimGrid_dx, tpb, 0, stream>>>(N_, D, d_cbs_, d_x_t, d_v_t, d_du_dx_, dt_);
+    gpuErrchk(cudaPeekAtLastError());
+    initialized_ = true;
```
mcwitt, line 79: Could this actually just throw a runtime exception? Are there any cases where calling this twice before `finalize` isn't a bug?

timemachine/cpp/src/verlet_integrator.cu
```diff
@@ -0,0 +1,119 @@
...
+    std::vector<BoundPotential *> &bps,
+    double lamb,
+    double *d_x_t,
+    double *d_v_t,
+    double *d_box_t,
+    cudaStream_t stream) {
+
+    if (initialized_) {
+        // No op if this is initialized, will just mess up the velos
+        return;
```
mcwitt, line 56: Could this actually just throw a runtime exception? Are there any cases where calling this twice before finalize isn't a bug?

timemachine/cpp/src/verlet_integrator.cu
```diff
@@ -0,0 +1,119 @@
...
+            d_du_dx_, // we only need the forces
+            nullptr,
+            nullptr,
+            nullptr,
+            stream);
+    }
+    half_step_velocity_verlet<double, false>
+        <<<dimGrid_dx, tpb, 0, stream>>>(N_, D, d_cbs_, d_x_t, d_v_t, d_du_dx_, dt_);
+    gpuErrchk(cudaPeekAtLastError());
+    initialized_ = false;
```
mcwitt, line 116: Q: is it legal to call `finalize` immediately followed by `initialize` to resume a trajectory? Python pseudocode:
```python
integ = VelocityVerletIntegrator()
integ.initialize()
for _ in range(100):
    integ.step_fwd()
integ.finalize()
integ.initialize()
for _ in range(100):
    integ.step_fwd()
integ.finalize()
```
and is that the same as running 200 steps without finalize/initialize in between?

proteneer (approved):
lgtm!

mcwitt (approved):

timemachine/cpp/src/verlet_integrator.cu
maxentile: Should be perfectly fine.

The trajectory is defined by a for loop, where the loop body is three steps

____
V(dt/2)
R(dt)
V(dt/2)
____
V(dt/2)
R(dt)
V(dt/2)
____
V(dt/2)
R(dt)
V(dt/2)
____
...
____
V(dt/2)
R(dt)
V(dt/2)
____

However, for implementation reasons it is more convenient if the main loop body only requires the forces computed the previous step. We can accomplish this by manually applying a "loop shifting" optimization:
____
V(dt/2)
R(dt)
____
V(dt/2)
V(dt/2)
R(dt)
____
V(dt/2)
V(dt/2)
R(dt)
____
...
____
V(dt/2)
V(dt/2)
R(dt)
____
V(dt/2)
____

This just causes a small complication https://github.com/proteneer/timemachine/blob/53495d130c70d010ce1e04c96d72d90fa4441687/timemachine/integrator.py#L158-L191 where the state at any point between iterations of the main for-loop will have the positions and velocities staggered by half a timestep. Regardless, "initialize; main loop; finalize;" should implement the exact same sequence of updates as the defined trajectory, just grouping the updates differently for implementation convenience.

maxentile (approved):
Looks good to me!

tests/test_reference_velocity_verlet_integrator.py
```diff
@@ -23,20 +21,20 @@ def self_inverse(x, v):
         x_next, v_next = update_fxn(x, v)
         return x_next, -v_next
 
-    close = partial(np.allclose, atol=atol)
+    x0 = FIXED_TO_FLOAT(FLOAT_TO_FIXED(x0))
+    v0 = FIXED_TO_FLOAT(FLOAT_TO_FIXED(v0))
```
maxentile, line 25: Should capture comment from offline about the importance of this step!

(i.e. bitwise determinism applies for starting conditions where `FIXED_TO_FLOAT(FLOAT_TO_FIXED(x0)) == x0` ..., but that condition won't be true in general for all floating-point inputs `x0`, `v0`. However, arbitrary floating-point inputs can be made to satisfy condition by applying the round-trip conversion in these 2 lines.)

badisa: Added in [f575314](https://github.com/proteneer/timemachine/pull/789/commits/f5753143c43232dcf5476b96f128d6597d88eca1)

timemachine/cpp/src/langevin_integrator.cu
```diff
@@ -0,0 +1,93 @@
...
+LangevinIntegrator::~LangevinIntegrator() {
+    gpuErrchk(cudaFree(d_cbs_));
+    gpuErrchk(cudaFree(d_ccs_));
+    gpuErrchk(cudaFree(d_noise_));
+    gpuErrchk(cudaFree(d_du_dx_));
+    curandErrchk(curandDestroyGenerator(cr_rng_));
+}
+
+void LangevinIntegrator::step_fwd(
+    std::vector<BoundPotential *> &bps,
```
proteneer, line 40: can you remind me again why we're shifting the responsibility of the force calculation to the `step()` function as opopsed to the context? (i.e. why can't we just pass in a `double *d_du_dx_t` instead?

proteneer: (and similar for `initialize()` and `finalize()`)

proteneer: in the old behavior, the context sometimes will compute multiple values (eg. `U`, `dU_dx`, `dU_dp`, `dU_dl`) in a single pass, where as here, you will force the integrator to compute an exra `dU_dx` on every step.

timemachine/integrator.py
```diff
@@ -154,14 +164,17 @@ def __init__(self, force_fxn, masses, dt):
...
     def step(self, x, v):
         """WARNING: makes 2 calls to force_fxn per timestep -- prefer `.multiple_steps` in most cases"""
-        v_mid = v + (0.5 * self.dt / self.masses) * self.force_fxn(x)
-        x = x + self.dt * v_mid
-        v = v_mid + (0.5 * self.dt / self.masses) * self.force_fxn(x)
+        fixed_x = FLOAT_TO_FIXED(x)
+        fixed_v = FLOAT_TO_FIXED(v)
+        v_mid = fixed_v + FLOAT_TO_FIXED((0.5 * self.cb) * self.force_fxn(x))
+        fixed_x = fixed_x + FLOAT_TO_FIXED(self.dt * FIXED_TO_FLOAT(v_mid))
+        fixed_v = v_mid + FLOAT_TO_FIXED((0.5 * self.cb) * self.force_fxn(FIXED_TO_FLOAT(fixed_x)))
```
proteneer, line 175: would this be easier to read:
```
v_mid = FLOAT_TO_FIXED(v) + FLOAT_TO_FIXED((0.5 * self.cb) * self.force_fxn(x))
fixed_x = FLOAT_TO_FIXED(x) + FLOAT_TO_FIXED(self.dt * FIXED_TO_FLOAT(v_mid))
fixed_v = v_mid + FLOAT_TO_FIXED((0.5 * self.cb) * self.force_fxn(FIXED_TO_FLOAT(fixed_x)))
```
I was trying to figure out where `fixed_x` was being used since it was declared twice

timemachine/cpp/src/langevin_integrator.cu
badisa: In part it is more consistent with our reference implementations, and it also allows the integrator to compute the forces as many times as it would like. Which seems like a cleaner separation of responsibilities.

Currently the only time we compute additional values at the same time as `dU_dx` is `dU_dl` which this has a carve out for by passing in a buffer. The other case where we compute values on frames (`multiple_steps_U`) is already a separate potential call to the force calculation.

proteneer: hm okay - I see! I'm okay with this.

proteneer:
lgtm! and nice contribution

PR #790, opened 2022-08-10
Improve robustness of rbfe.py

proteneer:
This is a minor PR that:

-Handles exceptions more gracefully in the rbfe.py code when an exception is thrown
-Adds a test for bitwise reproducibility using the initial_state and protocol object
-Removes an extra print statement related to plotting
-Improves memory usage to that of 2 states for both trajectory generation and free energy estimates
-Uses the new atom mapping protocol that reduces the number of dummy atoms

### DIFF
--- a/examples/rbfe_edge_list.py
+++ b/examples/rbfe_edge_list.py
@@ -82,12 +82,12 @@ def read_from_args():
 
     for fut, meta in zip(futures, metadata):
 
+        mol_a, mol_b, _, _, exp_ddg, fep_ddg, fep_ddg_err, ccc_ddg, ccc_ddg_err = meta
+        mol_a_name = get_mol_name(mol_a)
+        mol_b_name = get_mol_name(mol_b)
+
         try:
             solvent_res, solvent_top, complex_res, complex_top = fut.result()
-
-            mol_a, mol_b, _, _, exp_ddg, fep_ddg, fep_ddg_err, ccc_ddg, ccc_ddg_err = meta
-            mol_a_name = get_mol_name(mol_a)
-            mol_b_name = get_mol_name(mol_b)
 
             with open(f"success_rbfe_result_{mol_a_name}_{mol_b_name}.pkl", "wb") as fh:
                 pkl_obj = (meta, solvent_res, solvent_top, complex_res, complex_top)
@@ -106,6 +106,8 @@ def read_from_args():
             )
 
         except SimulationException as sim_exc:
+
+            print("Failed", sim_exc)
             with open(f"failed_rbfe_result_{mol_a_name}_{mol_b_name}.pkl", "wb") as fh:
                 pickle.dump(sim_exc, fh)
 
--- a/examples/rbfe_edge_list.py
+++ b/examples/rbfe_edge_list.py
@@ -1,6 +1,7 @@
 import argparse
 import csv
 import pickle
+import traceback
 
 import numpy as np
 from rdkit import Chem
@@ -106,8 +107,7 @@ def read_from_args():
             )
 
         except SimulationException as sim_exc:
-
-            print("Failed", sim_exc)
+            traceback.print_exc()
             with open(f"failed_rbfe_result_{mol_a_name}_{mol_b_name}.pkl", "wb") as fh:
                 pickle.dump(sim_exc, fh)
 
--- a/examples/rbfe_edge_list.py
+++ b/examples/rbfe_edge_list.py
@@ -9,10 +9,8 @@ Chem.SetDefaultPickleProperties(Chem.PropertyPickleOptions.AllProps)
 
 from timemachine.constants import KCAL_TO_KJ
 from timemachine.fe import atom_mapping
-from timemachine.fe.rbfe import HostConfig, estimate_relative_free_energy
+from timemachine.fe.rbfe import SimulationException, run_pair
 from timemachine.fe.utils import get_mol_name
-from timemachine.ff import Forcefield
-from timemachine.md import builders
 from timemachine.parallel.client import CUDAPoolClient
 
 
@@ -22,30 +20,6 @@ def get_mol_by_name(mols, name):
             return m
 
     assert 0, "Mol not found"
-
-
-def run_pair(mol_a, mol_b, core, forcefield_path, protein_path, n_frames, seed):
-
-    box_width = 4.0
-    solvent_sys, solvent_conf, solvent_box, solvent_top = builders.build_water_system(box_width)
-    solvent_box += np.diag([0.1, 0.1, 0.1])  # remove any possible clashes, deboggle later
-    solvent_host_config = HostConfig(solvent_sys, solvent_conf, solvent_box)
-
-    forcefield = Forcefield.load_from_file(forcefield_path)
-
-    solvent_res = estimate_relative_free_energy(
-        mol_a, mol_b, core, forcefield, solvent_host_config, seed, n_frames=n_frames, prefix="solvent"
-    )
-
-    complex_sys, complex_conf, _, _, complex_box, complex_top = builders.build_protein_system(protein_path)
-    complex_box += np.diag([0.1, 0.1, 0.1])  # remove any possible clashes, deboggle later
-    complex_host_config = HostConfig(complex_sys, complex_conf, complex_box)
-
-    complex_res = estimate_relative_free_energy(
-        mol_a, mol_b, core, forcefield, complex_host_config, seed + 1, n_frames=n_frames, prefix="complex"
-    )
-
-    return solvent_res, solvent_top, complex_res, complex_top
 
 
 def read_from_args():
@@ -83,11 +57,12 @@ def read_from_args():
             mol_b = get_mol_by_name(mols, mol_b_name)
 
             print(f"Submitting job for {mol_a_name} -> {mol_b_name}")
-            mcs_result = atom_mapping.mcs_map_graph_only_complete_rings(mol_a, mol_b)
+            mcs_threshold = 2.0
+            mcs_result = atom_mapping.mcs_map(mol_a, mol_b, threshold=mcs_threshold)
             query_mol = Chem.MolFromSmarts(mcs_result.smartsString)
-            core = atom_mapping.get_core_by_mcs(mol_a, mol_b, query_mol, threshold=2.0)
+            core = atom_mapping.get_core_by_mcs(mol_a, mol_b, query_mol, threshold=mcs_threshold)
             fut = cpc.submit(
-                run_pair, mol_a, mol_b, core, args.forcefield, args.protein, args.seed + row_idx, args.n_frames
+                run_pair, mol_a, mol_b, core, args.forcefield, args.protein, args.n_frames, args.seed + row_idx
             )
             futures.append(fut)
 
@@ -106,26 +81,33 @@ def read_from_args():
             )
 
     for fut, meta in zip(futures, metadata):
-        solvent_res, solvent_top, complex_res, complex_top = fut.result()
-        solvent_ddg = np.sum(solvent_res.all_dGs)
-        solvent_ddg_err = np.linalg.norm(solvent_res.all_errs)
-        complex_ddg = np.sum(complex_res.all_dGs)
-        complex_ddg_err = np.linalg.norm(complex_res.all_errs)
 
-        mol_a, mol_b, _, _, exp_ddg, fep_ddg, fep_ddg_err, ccc_ddg, ccc_ddg_err = meta
-        mol_a_name = get_mol_name(mol_a)
-        mol_b_name = get_mol_name(mol_b)
+        try:
+            solvent_res, solvent_top, complex_res, complex_top = fut.result()
 
-        with open(f"rbfe_result_{mol_a_name}_{mol_b_name}.pkl", "wb") as fh:
-            pkl_obj = (meta, solvent_res, solvent_top, complex_res, complex_top)
-            pickle.dump(pkl_obj, fh)
+            mol_a, mol_b, _, _, exp_ddg, fep_ddg, fep_ddg_err, ccc_ddg, ccc_ddg_err = meta
+            mol_a_name = get_mol_name(mol_a)
+            mol_b_name = get_mol_name(mol_b)
 
-        tm_ddg = complex_ddg - solvent_ddg
-        tm_err = np.linalg.norm([complex_ddg_err, solvent_ddg_err])
+            with open(f"success_rbfe_result_{mol_a_name}_{mol_b_name}.pkl", "wb") as fh:
+                pkl_obj = (meta, solvent_res, solvent_top, complex_res, complex_top)
+                pickle.dump(pkl_obj, fh)
 
-        print(
-            f"finished {mol_a_name} -> {mol_b_name} (kJ/mol) | complex {complex_ddg:.2f} +- {complex_ddg_err:.2f} | solvent {solvent_ddg:.2f} +- {solvent_ddg_err:.2f} | tm_pred {tm_ddg:.2f} +- {tm_err:.2f} | exp_ddg {exp_ddg:.2f} | fep_ddg {fep_ddg:.2f} +- {fep_ddg_err:.2f}"
-        )
+            solvent_ddg = np.sum(solvent_res.all_dGs)
+            solvent_ddg_err = np.linalg.norm(solvent_res.all_errs)
+            complex_ddg = np.sum(complex_res.all_dGs)
+            complex_ddg_err = np.linalg.norm(complex_res.all_errs)
+
+            tm_ddg = complex_ddg - solvent_ddg
+            tm_err = np.linalg.norm([complex_ddg_err, solvent_ddg_err])
+
+            print(
+                f"finished: {mol_a_name} -> {mol_b_name} (kJ/mol) | complex {complex_ddg:.2f} +- {complex_ddg_err:.2f} | solvent {solvent_ddg:.2f} +- {solvent_ddg_err:.2f} | tm_pred {tm_ddg:.2f} +- {tm_err:.2f} | exp_ddg {exp_ddg:.2f} | fep_ddg {fep_ddg:.2f} +- {fep_ddg_err:.2f}"
+            )
+
+        except SimulationException as sim_exc:
+            with open(f"failed_rbfe_result_{mol_a_name}_{mol_b_name}.pkl", "wb") as fh:
+                pickle.dump(sim_exc, fh)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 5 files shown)


proteneer:
@mcwitt I think I'm going to defer the exception handling comment to a later revision - I'm not sure how to think about handling exceptions across processes/gRPC calls properly right now and I'd rather they be handled explicitly (so we can do recovery of things like initial states).

timemachine/fe/rbfe.py
```diff
@@ -125,34 +85,70 @@ def __init__(self, omm_system, conf, box):
...
+        lambda_windows=[initial_state.lamb],
+        store_u_interval=0,
+        store_x_interval=0,
+    )
 
-    lambda_schedule: np.array(L,)
-        Sequence of lambda windows
+    # a crude, and probably not great, guess on the decorrelation time
+    n_steps = protocol.n_frames * protocol.steps_per_frame
+    all_nrgs, all_coords, all_boxes = ctxt.multiple_steps_U(
```
badisa, line 110: Seems that we aren't using the energies, should we disable the lambda windows to avoid the additional cost?

timemachine/fe/rbfe.py
```diff
@@ -125,34 +85,70 @@ def __init__(self, omm_system, conf, box):
...
+    assert all_boxes.shape[0] == protocol.n_frames
 
-    Returns
-    -------
-    3-tuple of all_frames, all_boxes, all_U_fns
-        all_frames has shape (L,T,N,3), all_boxes has shape (L,T,3,3), all_U_fns has shape (L,)
+    return all_coords, all_boxes
 
+
+class SimulationProtocol:
```
badisa, line 124: nit: dataclass for SimulationProtocol and InitialState

timemachine/fe/rbfe.py
```diff
@@ -408,21 +431,25 @@ def estimate_relative_free_energy(
...
+        prefix = get_mol_name(mol_a) + "_" + get_mol_name(mol_b) + "_" + prefix
+
+        return estimate_free_energy_given_initial_states(initial_states, protocol, temperature, prefix, keep_idxs)
+
+    except Exception as e:
 
-    keep_frames = []
-    keep_boxes = []
-    keep_U_fns = []
+        import traceback
```
badisa, line 450: Move to the top of the file

proteneer: good call

timemachine/fe/rbfe.py
proteneer: whoops - forgot to cherry pick the dataclass. Also still struggling to come up with a better name than "InitialState", maybe "DeterministicSetup", "DeterministicInitializer", "SimulationSettings", all sound equally crappy

proteneer:
also: noticed an issue with the atom-mapping code - will fix.

timemachine/fe/rbfe.py
```diff
@@ -408,21 +431,25 @@ def estimate_relative_free_energy(
...
+        if keep_idxs is None:
+            keep_idxs = [0, -1]  # keep first and last frames
+
+        assert len(keep_idxs) <= len(lambda_schedule)
+
+        prefix = get_mol_name(mol_a) + "_" + get_mol_name(mol_b) + "_" + prefix
+
+        return estimate_free_energy_given_initial_states(initial_states, protocol, temperature, prefix, keep_idxs)
+
+    except Exception as e:
```
badisa, line 448: Seems cleaner to push the exception handling into the `estimate_free_energy_given_initial_states` as it returns a SimulationResult, so it would seem like it could set the exception field.

timemachine/fe/rbfe.py
```diff
@@ -226,11 +220,22 @@ def plot_BAR(df, df_err, fwd_delta_u, rev_delta_u, title, axes):
...
+        "boxes",
+        "initial_states",
+        "protocol",
+        "exception",
+    ],
 )
 
 
-def estimate_free_energy_given_samples(all_frames, all_boxes, all_U_fns, temperature, lambda_schedule, prefix):
+# def estimate_free_energy_given_samples(all_frames, all_boxes, all_U_fns, temperature, lambda_schedule, prefix):
```
badisa, line 237: Remove commented out signature?

timemachine/fe/rbfe.py
```diff
@@ -125,34 +85,70 @@ def __init__(self, omm_system, conf, box):
...
-        Configuration of the host
+    ctxt = custom_ops.Context(initial_state.x0, initial_state.v0, initial_state.box0, intg_impl, bound_impls, baro_impl)
 
-    temperature: float
-        Temperature of the system, in Kelvins
+    # burn-in
+    ctxt.multiple_steps_U(
+        lamb=initial_state.lamb,
+        n_steps=protocol.burn_in,
+        lambda_windows=[initial_state.lamb],
```
badisa, line 103: No need to compute energies for burn in.

examples/rbfe_edge_list.py
```diff
@@ -83,11 +82,12 @@ def read_from_args():
             mol_b = get_mol_by_name(mols, mol_b_name)
 
             print(f"Submitting job for {mol_a_name} -> {mol_b_name}")
+            mcs_threshold = 100.0
```
badisa, line 85: This seems really large. 100 angstroms of distance seems to pretty much mean no threshold?

proteneer: this is actually a problem it turns out - (i will patch this a bit)

timemachine/fe/rbfe.py
```diff
@@ -125,34 +88,81 @@ def __init__(self, omm_system, conf, box):
...
+    x0: np.ndarray
+    v0: np.ndarray
+    box0: np.ndarray
+    lamb: float
+
+
+@dataclass
+class SimulationResult:
+    all_dGs: List[np.ndarray]
+    all_ers: List[float]
```
mcwitt, line 154: nit: typo

timemachine/fe/rbfe.py
proteneer: willfix

timemachine/fe/rbfe.py
proteneer: this was de-boggled! all of the exception handling is done on the client side

timemachine/fe/rbfe.py
proteneer: i'm going to leave this in for debugging purposes -  sometimes its useful to monitor the progress of the nrgs

examples/rbfe_edge_list.py
proteneer: resolved - there was a bug in the MCS logic (dealing with how the threshold is mapped, there was a corner cases where the lowest cost[i,j] may still contain bad pair that was bigger than the distance threshold.

timemachine/fe/rbfe.py
proteneer: fixed

examples/run_from_pickle.py
```diff
@@ -0,0 +1,31 @@
...
+
+    with open(args.pickle_path, "rb") as fh:
+        (meta, solvent_res, solvent_top, complex_res, complex_top) = pickle.load(fh)
+        print("running solvent")
+        for initial_state in solvent_res.initial_states:
+            all_coords, all_boxes = rbfe.sample(initial_state, solvent_res.protocol)
+        print("running complex")
+        for initial_state in complex_res.initial_states:
+            all_coords, all_boxes = rbfe.sample(initial_state, complex_res.protocol)
+        print("SUCCESS")
```
mcwitt: Could be dedented?

timemachine/fe/atom_mapping.py
```diff
@@ -52,13 +52,19 @@ def compare(self, p, mol1, atom1, mol2, atom2):
             return True
 
 
-def mcs_map(a, b, threshold: float = 0.5, timeout: int = 5, smarts: Optional[str] = None):
+def mcs_map(a, b, threshold=2.0, timeout: int = 5, smarts: Optional[str] = None):
```
mcwitt, line 55: nit: lost annotation

timemachine/fe/atom_mapping.py
```diff
@@ -52,13 +52,19 @@ def compare(self, p, mol1, atom1, mol2, atom2):
...
     params.BondCompareParameters.RingMatchesRingOnly = 1
-    params.AtomCompareParameters.matchValences = 1
-    params.AtomTyper = CompareDist(threshold=threshold)
+    params.BondTyper = rdFMCS.BondCompare.CompareAny
+
+    params.AtomCompareParameters.CompleteRingsOnly = 1
+    params.AtomCompareParameters.RingMatchesRingOnly = 1
+    params.AtomCompareParameters.matchValences = 0
+    params.AtomCompareParameters.MaxDistance = threshold
+    params.AtomTyper = rdFMCS.AtomCompare.CompareAny
```
mcwitt, line 66: Commenting on some of these settings might be useful (esp. those that changed). E.g., why did we change `matchValences` to 0?

timemachine/fe/rbfe.py
badisa: Should we print these values then?

timemachine/fe/rbfe.py
```diff
@@ -125,34 +88,81 @@ def __init__(self, omm_system, conf, box):
...
+    integrator: LangevinIntegrator
+    barostat: MonteCarloBarostat
+    x0: np.ndarray
+    v0: np.ndarray
+    box0: np.ndarray
+    lamb: float
+
+
+@dataclass
+class SimulationResult:
```
badisa, line 152: Should we differentiate this with this other class: https://github.com/proteneer/timemachine/blob/master/timemachine/fe/estimator.py#L18-L22

timemachine/fe/atom_mapping.py
```diff
@@ -187,11 +193,15 @@ def get_core_by_mcs(mol_a, mol_b, query, threshold=0.5):
     matches_a = mol_a.GetSubstructMatches(query, uniquify=False)
     matches_b = mol_b.GetSubstructMatches(query, uniquify=False)
 
-    # cost[i, j] = sum_i distance(conf)
     cost = np.zeros((len(matches_a), len(matches_b)))
     for i, a in enumerate(matches_a):
```
mcwitt, line 197: I noticed that tests still pass when `matches_a` is substituted with `matches_b` on this line. Is it easy to add an example that would fail?

examples/run_from_pickle.py
```diff
@@ -0,0 +1,31 @@
+# run from a simulation state pickle to help with debugging
```
badisa, line 1: Add a test for this in `tests/test_examples.py`?

proteneer: i didn't mean to check this in - the tests already test reproduciblity from the `(initial_state, protocol)` pair

timemachine/fe/atom_mapping.py
```diff
@@ -187,11 +193,15 @@ def get_core_by_mcs(mol_a, mol_b, query, threshold=0.5):
...
         for j, b in enumerate(matches_b):
-            cost[i, j] = np.linalg.norm(conf_a[np.array(a)] - conf_b[np.array(b)], axis=1).sum()
+            # if a single pair is outside of threshold, we set the cost to inf
+            dij = np.linalg.norm(conf_a[np.array(a)] - conf_b[np.array(b)], axis=1)
+            if np.any(dij > threshold):
+                cost[i, j] = np.inf
+            else:
+                cost[i, j] = dij.sum()
```
mcwitt, line 204: nit: slightly cleaner as ternary expression
```suggestion
            cost[i, j] = dij.sum() if np.all(dij < threshold) else np.inf
```

timemachine/fe/rbfe.py
proteneer: hm I'm not sure - I'd like to merge them (but didn't want to break the RABFE code). Will resolve during deboggling week.

timemachine/fe/rbfe.py
proteneer: I'm only printing these out when a simulation crashes - we probably don't want to print them out generally.

timemachine/fe/rbfe.py
```diff
@@ -125,34 +88,81 @@ def __init__(self, omm_system, conf, box):
...
 
+@dataclass
+class InitialState:
     """
+    An initial contains everything that is needed to bitwise reproduce a trajectory given a SimulationProtocol
+
+    This object can be pickled safely.
+    """
+
+    U_fns: List[CustomOpWrapper]
```
mcwitt, line 142: naming nit: the name `U_fns` suggests that these are functions rather than potential objects. Would it be preferable to call this `potentials` instead?

timemachine/fe/rbfe.py
proteneer: also - I think `store_u_interval=0` should avoid computing the energies directly

timemachine/fe/rbfe.py
```diff
@@ -125,34 +88,81 @@ def __init__(self, omm_system, conf, box):
...
+    v0: np.ndarray
+    box0: np.ndarray
+    lamb: float
+
+
+@dataclass
+class SimulationResult:
+    all_dGs: List[np.ndarray]
+    all_errs: List[float]
+    plot_png: Any
```
mcwitt, line 155: nit: a more precise type here could be `bytes`

timemachine/fe/rbfe.py
proteneer: correct we're not using them right now, they were mostly for debugging purposes. These are computed once every 1000 steps right now, so it probably doesn't hurt to keep them.

timemachine/fe/rbfe.py
badisa: `store_u_interval=0` will run the energy on the last step. Not a big cost, just doesn't seem like it will be ever seen if run in production.

timemachine/fe/rbfe.py
```diff
@@ -225,12 +233,14 @@ def plot_BAR(df, df_err, fwd_delta_u, rev_delta_u, title, axes):
     axes.legend()
 
 
-SimulationResult = namedtuple(
-    "SimulationResult", ["all_dGs", "all_errs", "plot_png", "frames", "boxes", "U_fns", "lambda_schedule"]
-)
+class DebugException(Exception):
```
mcwitt, line 236: This sounds pretty generic. Would `SimulationException` be an improvement?

timemachine/fe/rbfe.py
```diff
@@ -243,80 +253,75 @@ def estimate_free_energy_given_samples(all_frames, all_boxes, all_U_fns, tempera
...
+        Detailing specifics of each simulation
 
     temperature: float
         Temperature the system was run at
 
-    lambda_schedule: np.array of shape (L,)
-        Lambda schedule
-
     prefix: str
         A prefix to append to overlap figures
```
mcwitt, line 266: Typo?

badisa (approved):

timemachine/fe/rbfe.py
```diff
@@ -392,6 +411,13 @@ def estimate_relative_free_energy(
     lambda_schedule: list of float
         This should only be set when debugging or unit testing. This argument may be removed later.
 
+    keep_idxs: list of int or None
+        If None, return only the end-state frames. Otherwise if not None, use only for debugging, and this
+        will return the frames corresponding to the idxs of interest.
+
+    prepare_only: bool
```
mcwitt, line 418: Missing from the arguments?

timemachine/fe/rbfe.py
```diff
@@ -408,21 +434,40 @@ def estimate_relative_free_energy(
...
-    )
-    all_dGs, all_errs, plot_buffer = estimate_free_energy_given_samples(
-        all_frames, all_boxes, all_U_fns, temperature, lambda_schedule, prefix
-    )
+    initial_states = setup_initial_states(single_topology, host_config, temperature, lambda_schedule, seed)
+    protocol = SimulationProtocol(n_frames=n_frames, burn_in=10000, steps_per_frame=1000)
+
+    try:
+        if keep_idxs is None:
+            keep_idxs = [0, -1]  # keep first and last frames
```
mcwitt, line 442: nit: move outside of try block

timemachine/fe/rbfe.py
```diff
@@ -408,21 +434,40 @@ def estimate_relative_free_energy(
...
+        assert len(keep_idxs) <= len(lambda_schedule)
+        combined_prefix = get_mol_name(mol_a) + "_" + get_mol_name(mol_b) + "_" + prefix
+        return estimate_free_energy_given_initial_states(
+            initial_states, protocol, temperature, combined_prefix, keep_idxs
+        )
+
+    except Exception as old_exc:
+
+        raise DebugException(initial_states, protocol, prefix) from old_exc
```
mcwitt, line 451: should this be
```suggestion
        raise DebugException(initial_states, protocol, combined_prefix) from old_exc
```
?

mcwitt (approved):

examples/run_from_pickle.py
proteneer: sorry this was a debug file - didn't mean for it to be added into the PR.

timemachine/fe/atom_mapping.py
proteneer: willfix

timemachine/fe/atom_mapping.py
proteneer: matchValences was unnecessary (and was a default from an older protocol that tried to avoid chiral additions, eg. amides into tetrahedral centers). Chiral restraints should naturally deal with this issue.

timemachine/fe/rbfe.py
proteneer: d'oh this was from a debug session

proteneer: will remove

timemachine/fe/rbfe.py
proteneer: will minimize scope of try block

timemachine/fe/rbfe.py
proteneer: good idea - willfix inline

timemachine/fe/atom_mapping.py
proteneer: I'm not sure what the old tests are doing - I think we're going to 1) remove the old tests and the old atom_mapping code and 2) rewrite them based on this class (and whatever protocols we support).

timemachine/fe/rbfe.py
proteneer: good call

timemachine/fe/rbfe.py
proteneer: I agree - note: I should get rename the `HostGuestSystem::get_U_fns()` to `get_potentials` instead

proteneer:
rebased and will merge once tests pass

timemachine/fe/rbfe.py
proteneer: fixed

timemachine/fe/rbfe.py
proteneer: (will do the HGS one in deboggling week)

PR #793, opened 2022-08-10
Queries Nvidia SMI for compute capability

### LINKED ISSUE #728: Precommit hook to verify pybind stubs can require a complete new build
Right now if you do `make CUDA_ARCH=86 build` then run `git commit` the pre-commit trigger will rebuild everything. Seems like there should be a way to just generate the stubs and not have to build all of the source.

badisa:
* Unsure how this will work if you have two GPUs with different compute capabilities
* Fixes #728 by having `make build` use the compute capability of the GPU rather than the default


badisa:
I am unsure how this behaves if you have multiple GPUs, especially if you have different capabilities on different GPUs. 

@jkausrelay I believe you have two GPUs with different capabilities if you could try it.

mcwitt:
This doesn't seem to quite work with multiple CUDA devices (even if they have the same compute capability). On my local setup (2 x CC 6.1 GPUs) this fails with

```
Cuda Arch: 61
61
-- Configuring done
-- Generating done
-- Build files have been written to: /home/matt-relay/src/timemachine/timemachine/cpp/build
make[1]: Entering directory '/home/matt-relay/src/timemachine/timemachine/cpp/build'
make[2]: Entering directory '/home/matt-relay/src/timemachine/timemachine/cpp/build'
make[3]: Entering directory '/home/matt-relay/src/timemachine/timemachine/cpp/build'
CMakeFiles/custom_ops.dir/flags.make:12: *** missing separator.  Stop.
make[3]: Leaving directory '/home/matt-relay/src/timemachine/timemachine/cpp/build'
make[2]: *** [CMakeFiles/Makefile2:101: CMakeFiles/custom_ops.dir/all] Error 2
make[2]: Leaving directory '/home/matt-relay/src/timemachine/timemachine/cpp/build'
make[1]: *** [Makefile:136: all] Error 2
make[1]: Leaving directory '/home/matt-relay/src/timemachine/timemachine/cpp/build'
make: *** [Makefile:15: build] Error 2
```

This is the output I get from the query command:
```console
$ nvidia-smi --query-gpu=compute_cap --format=csv | grep -v compute_cap
6.1
6.1
```

Maybe something like
```bash
nvidia-smi --query-gpu=compute_cap --format=csv | grep -v compute_cap | sort -n | tail -n1
```
would handle this case gracefully?

jkausrelay:
@badisa This is what I get:

`nvidia-smi --query-gpu=compute_cap --format=csv | grep -v compute_cap | sort -n | head -n1`
> 6.1

which corresponds to the lower CC of my two cards. But I'm not sure this is what we really want, since it'll build with the lower CC card which is not really used for running jobs

nvidia-smi --query-gpu=compute_cap,name --format=csv
compute_cap, name
8.6, NVIDIA RTX A4000
6.1, Quadro P2000

badisa:
> @badisa This is what I get:
> 
> `nvidia-smi --query-gpu=compute_cap --format=csv | grep -v compute_cap | sort -n | head -n1`
> 
> > 6.1
> 
> which corresponds to the lower CC of my two cards. But I'm not sure this is what we really want, since it'll build with the lower CC card which is not really used for running jobs
> 
> nvidia-smi --query-gpu=compute_cap,name --format=csv compute_cap, name 8.6, NVIDIA RTX A4000 6.1, Quadro P2000

This is intended to be the 'safest' option, as it could build for `8.6` but then fail on the Quadro card. We could perhaps disable this if there are multiple GPUs and instead go back to defaulting to 75? That or we could just leave things as they are.

proteneer:
I think if there are two GPUs - we should just throw an exception and say "this isn't supported" as opposed to silent picking the lower CC (and just have @jkausrelay specify the CC manually)

(otherwise LGTM)!

proteneer (approved):

jkausrelay (approved):

jkausrelay (approved):

PR #796, opened 2022-08-11
Require setuptools < 64.0.0

mcwitt:
Temporary workaround for incompatibility with https://peps.python.org/pep-0660/

On setuptools [v64.0.0](https://github.com/pypa/setuptools/blob/main/CHANGES.rst#v6400), `pip install .` fails with

```
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmplzftnbox.build-lib/timemachine/_version.py'
```

A possible workaround is to set `SETUPTOOLS_ENABLE_FEATURES="legacy-editable"`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ follow_imports = "silent"
 requires = [
   "cmake==3.22.1",
   "mypy==0.942",
-  "setuptools>=43.0.0",
+  "setuptools >= 43.0.0, < 64.0.0",
   "wheel",
   "versioneer-518"
 ]


badisa:
Any idea why this hasn't shown up on CI?

mcwitt:
> Any idea why this hasn't shown up on CI?

I think it's just a matter of time. v64.0.0 was released [4 hours ago](https://github.com/pypa/setuptools/commit/bc39d28bda2a1faee6680ae30e42526b9d775151).

badisa:
> > Any idea why this hasn't shown up on CI?
> 
> I think it's just a matter of time. v64.0.0 was released [4 hours ago](https://github.com/pypa/setuptools/commit/bc39d28bda2a1faee6680ae30e42526b9d775151).

This job ran within the last 4 hours.

https://gitlab.com/relaytx/physics-platform/timemachine/-/jobs/2845959786

Prior to running the tests it does `pip install .[test]` which should rebuild the C++ and pull in setuptools

badisa:
> @badisa hmm, are you sure dependencies aren't being cached in the image? I'm able to reproduce the error locally on with a fresh clone (on Python 3.7)

That seems likely. We might want to enable the verbose mode in the docker build to ensure we have more visibility. Should probably also have a nightly task that rebuilds with no cache to ensure we catch these sorts of issues

mcwitt:
Actually it seems like this might be a quirk of my local dev setup. I'm unable to reproduce the error with `setuptools=64.0.0` on a controller instance. I'll close this for now, and can resurrect if this ever surfaces in CI. Sorry for the noise.

mcwitt:
Reopened since we hit this in a Docker build.

PR #797, opened 2022-08-11
Default to parallel setuptools build

mcwitt:
Currently `pip install .` will build the C++ extension using only a single process. It's possible to use an incantation like `CMAKE_BUILD_PARALLEL_LEVEL=$(nproc) pip install .` to enable build parallelism, but this is hard to remember.

This PR changes the default build parallelism to `multiprocessing.cpu_count()` (it can still be overriden with `CMAKE_BUILD_PARALLEL_LEVEL`).

This also cleans up some unused CMake-related configuration in `setup.py` that was carried over from the example this was originally adapted from.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@
 Adapted from https://github.com/pypa/sampleproject/blob/main/setup.py
 """
 
+import multiprocessing
 import os
 import pathlib
 import subprocess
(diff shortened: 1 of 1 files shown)


proteneer (approved):
lgtm!

badisa (approved):

PR #798, opened 2022-08-15
Serialize inputs

proteneer:
This should resolve some of the issues re: files vs serialized systems.


timemachine/fe/rbfe.py
```diff
@@ -450,9 +449,7 @@ def estimate_relative_free_energy(
         raise SimulationException(initial_states, protocol, combined_prefix) from old_exc
 
 
-def run_pair(mol_a, mol_b, core, forcefield_path, protein_path, n_frames, seed):
-
-    forcefield = Forcefield.load_from_file(forcefield_path)
+def run_pair(mol_a, mol_b, core, forcefield, protein_path, n_frames, seed):
```
badisa, line 452: Might be worth renaming `protein_path` to `protein`. Or adding docstring/type hints.

badisa (approved):

timemachine/fe/rbfe.py
proteneer: willfix

PR #801, opened 2022-08-19
Mapped works, mapped u_kn matrix, terminal bond length maps

maxentile:
Introduces a new submodule `timemachine.maps`

- [x] Add and test utilities for computing mapped works or mapped `u_kn` matrix, incorporating invertible `map_fxns`
- [x] Add and test utilities for mapping between states where terminal bond lengths and/or force constants are changing


timemachine/maps/estimators.py
```diff
@@ -0,0 +1,56 @@
+"""Utilities for computing mapped works or mapped u_kn matrix, incorporating invertible map_fxns"""
```
maxentile, line 1: Naming: `estimators.py` isn't quite right (doesn't contain estimators, but instead _adaptors_ for other estimators like EXP, BAR, MBAR that expect `works` array or `u_kn` matrix), but other naming ideas like `adaptors.py` didn't seem quite right either.

timemachine/maps/estimators.py
```diff
@@ -0,0 +1,56 @@
...
+
+import numpy as np
+from pymbar.utils import kln_to_kn  # TODO: should this be "vendored"?
```
maxentile, line 4: Currently importing from https://github.com/choderalab/pymbar/blob/1d14e80d3320a5ae0651fd002a8cb763a900190c/pymbar/utils.py#L46-L85 , but could be replicated here

timemachine/maps/estimators.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+
+    # compare with version without map_fxn
+    # return dst_reduced_energy_fxn(src_samples) - src_reduced_energy_fxn(src_samples)
+
+    mapped_samples, logdetjacs = map_fxn(src_samples)
+    return dst_reduced_energy_fxn(mapped_samples) - src_reduced_energy_fxn(src_samples) - logdetjacs
+
+
+def compute_mapped_u_kn(sample_lists, reduced_energy_fxns, map_fxns):
```
maxentile, line 25: Signature: assumes lists / arrays of callables -- rather than lists / arrays of parameters + a single u_fxn, map_fxn that accepts parameters -- which deviates from idiomatic jax and rules out some jax transformations. Each callable `map_fxn[i, j]` is assumed to act on whole trajectories (possibly including manual or automatic optimizations for processing whole trajectories).

This seemed the most concise and general way to write this utility (e.g. it allows independent implementations for each pair of states -- maybe you want to use the identity fxn for `map_fxns[i, i]`, a simple implementation for `map_fxns[i, j]` where `|j - i| == 1`, a more complicated map for `map_fxns[i, j]` where `|j - i| == 2`, and to not even bother for distant pairs `map_fxns[i, j]` where `|j - i| > 2` or something). However, this may need to be modified for performance if K^2 Python loop overhead is large, compared with the cost of processing each individual trajectory, and if a Jax version of this loop can be made significantly faster.

The current implementation of this utility is a good fit for the terminal bond map implementation, where a reasonably fast function for processing `(traj, map_params)` can be prepared [using `jit/vmap` once](https://github.com/proteneer/timemachine/blob/3602dd0bc89bf20ecdd24a83aee4f8df61496c8e/timemachine/maps/terminal_bonds.py#L100), and then essentially `partial`'d with differing values of `map_params` to [produce](https://github.com/proteneer/timemachine/blob/3602dd0bc89bf20ecdd24a83aee4f8df61496c8e/tests/test_terminal_bond_maps.py#L153-L158) a `K`-by-`K` array of callables (without triggering JIT `K^2` times). I don't anticipate issues with using a similar pattern for other maps in the future.

timemachine/maps/estimators.py
```diff
@@ -0,0 +1,56 @@
...
+    """Evaluate every sample from state i in every state j,
+    incorporating user-defined change of variables supplied in map_fxns[i, j]
```
maxentile, line 27: Possible documentation / runtime check to be added: This function assumes (without verifying) that:
(1) `map_fxns[i, j]` is the inverse of `map_fxns[j, i]`,
(2) the domain of `map_fxns[i]` is the support of state `i`,
(3) the codomain of `map_fxns[i, j]` is the support of state `j`.

timemachine/maps/terminal_bonds.py
```diff
@@ -0,0 +1,223 @@
...
+
+from typing import List
+
+import networkx as nx
+
+from timemachine.constants import BOLTZ, DEFAULT_TEMP
+
+
+@dataclass
+class Interval:
```
maxentile, line 21: Naming: Perhaps `PositiveInterval` would be more descriptive?

timemachine/maps/estimators.py
```diff
@@ -0,0 +1,56 @@
+"""Utilities for computing mapped works or mapped u_kn matrix, incorporating invertible map_fxns"""
+
+import numpy as np
+from pymbar.utils import kln_to_kn  # TODO: should this be "vendored"?
+
+__all__ = ["compute_mapped_reduced_work", "compute_mapped_u_kn"]
+
+
+def compute_mapped_reduced_work(src_samples, src_reduced_energy_fxn, dst_reduced_energy_fxn, map_fxn):
```
maxentile, line 9: Naming: (`samples`, `reduced_energy_fxn`) rather than (`xs`, `u_fxn`) (or (`confs`, `reduced_potential_fxn`) or similar), to anticipate the case where the samples / states / maps are augmented in some convenient way (e.g. `sample = (x, v)` to use a VelocityVerlet map, or `sample = (x, v, particle_selection)` to use a local MD map).

timemachine/maps/estimators.py
mcwitt: `kln_to_kn(u_kln)` seems like a synonym for `np.concatenate(u_kln, axis=1)`? The latter seems to be ~10x faster according to some rough timing. Is there any reason to prefer the former?

timemachine/maps/estimators.py
mcwitt: This sounds reasonable to me.

> e.g. it allows independent implementations for each pair of states

Could this also be reasonably achieved using a piecewise function? E.g. it seems relatively straightforward to implement the case of the functional form depending on $|i-j|$ using `jnp.where` or similar.

timemachine/maps/terminal_bonds.py
```diff
@@ -0,0 +1,239 @@
...
+    lower: float
+    upper: float
+
+    @property
+    def width(self) -> float:
+        return self.upper - self.lower
+
+    def validate(self) -> None:
```
mcwitt, line 30: Validation seems better done at construction (unless there's a good reason to defer?). For dataclasses, can do validation in the `__post_init__` method ([example](https://stackoverflow.com/a/60179826))

timemachine/maps/terminal_bonds.py
```diff
@@ -0,0 +1,239 @@
...
+        r = self.stddev * sigma_thresh
+        interval = Interval(self.mean - r, self.mean + r)
+        interval.validate()
+        return interval
+
+
+def gaussians_from_harmonic_bonds(ks, eq_lengths, temperature=DEFAULT_TEMP) -> List[Gaussian]:
+    kBT = BOLTZ * temperature
+    params = zip(ks, eq_lengths)
+    return [Gaussian(eq_length, np.sqrt(kBT / k)) for (k, eq_length) in params]
```
mcwitt, line 50: nit: this function seems like it does 2 distinct things:
1. translating harmonic bond parameters to mu, sigma
2. mapping over a list of inputs

timemachine/maps/terminal_bonds.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+@jit
+def interval_map(x, src_lb, src_ub, dst_lb, dst_ub):
+    """map x in Interval[src_lb, src_ub] to y in Interval[dst_lb, dst_ub]"""
+
+    scale_factor = (dst_ub - dst_lb) / (src_ub - src_lb)
+
+    in_support = (x >= src_lb) * (x <= src_ub)
```
mcwitt, line 59: nit: should we prefer `&` for logical AND?

timemachine/maps/terminal_bonds.py
```diff
@@ -0,0 +1,239 @@
...
+    mapped = dst_lb + (x - src_lb) * scale_factor
+
+    return jnp.where(in_support, mapped, np.nan)
```
mcwitt, line 62: Could we assert as a precondition that all inputs are in the domain?

mcwitt (commented):
Looks great so far! Left some initial questions/comments.

timemachine/maps/estimators.py
maxentile: >kln_to_kn(u_kln) seems like a synonym for np.concatenate(u_kln, axis=1)?

Not quite, in the general case that N_k are not all equal.

>The latter seems to be ~10x faster according to some rough timing. Is there any reason to prefer the former?

Other simplifications can be applied relative to the `kln_to_kn` utility in pymbar (e.g. assume `N_k` is always given / doesn't have to be inferred) -- just used this initially without peaking inside the definition.

tests/test_mapped_estimators.py
```diff
@@ -0,0 +1,135 @@
...
+
+        # since map_fxn is perfect, variance should be == 0...
+        mapped_w_F = compute_mapped_reduced_work(src_samples, u_src, u_dst, map_fxn)
+        assert np.std(mapped_w_F) < eps
+
+        # ... and estimated_delta_f should be == exact_delta_f
+        estimated_delta_f = EXP(mapped_w_F)[0]
+        exact_delta_f = dst_state.reduced_free_energy - src_state.reduced_free_energy
+
+        np.testing.assert_almost_equal(estimated_delta_f, exact_delta_f)
```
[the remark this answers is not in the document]
maxentile: Yes, will fix!

tests/test_mapped_estimators.py
```diff
@@ -0,0 +1,135 @@
...
+        exact_delta_f = state_b.reduced_free_energy - state_a.reduced_free_energy
+
+        np.testing.assert_almost_equal(estimated_delta_f, exact_delta_f)
+
+
+def test_multistate_estimates():
+
+    # define a collection of states
+    K = 10
+    states = [UnnormalizedGaussian.initialize_randomly() for _ in range(K)]
```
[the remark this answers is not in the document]
maxentile: Yes -- this omission was not intentional!

timemachine/maps/estimators.py
maxentile: >Could this also be reasonably achieved using a piecewise function? E.g. it seems relatively straightforward to implement the case of the functional form depending on  using jnp.where or similar.

In the first example, yes, but in other foreseeable experimental usages (e.g. if each state is a different molecule, and each pair is augmented with a different set of dummy atoms*), `np.where` mask wouldn't be as readily applicable.

(*Currently unclear if there's a correct way to do this, but it's something that I'll want to try soon.)

timemachine/maps/terminal_bonds.py
maxentile: Thank you -- your suggestion is clearly better!

>unless there's a good reason to defer?

No reason -- just ignorance of the `__post_init__` method

timemachine/maps/terminal_bonds.py
maxentile: Yes -- that would be better than nan-poison -- thanks for suggesting https://jax.readthedocs.io/en/latest/_autosummary/jax.experimental.checkify.checkify.html offline!

timemachine/maps/terminal_bonds.py
maxentile: Yes, that would be clearer!

timemachine/maps/terminal_bonds.py
maxentile: Good point -- would be cleaner to remove this free-floating function, add a `Gaussian.from_harmonic_bond`, and do the list comprehension where needed -- will fix!

maxentile:
Oof -- looks like my batch of comments pointing to the commits that address each review comment were stuck Pending in drafts...

Anyway, all suggestions were applied except for the one about switching the implementation of an assertion from nan-poison to jax.experimental.checkify.

tests/test_mapped_estimators.py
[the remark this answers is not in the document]
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/801/commits/9594dac36f9156e434241ba4a3f175818320ffb8

timemachine/maps/terminal_bonds.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/801/commits/28379e03a404d1f032b40b22fcb2175c336f6451

timemachine/maps/terminal_bonds.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/801/commits/ea1760e95725e4b8e97daef6bf99648ea561c79a

timemachine/maps/terminal_bonds.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/801/commits/0f07fc1cbefb6cf07377d7dd130aa724b861e84d

timemachine/maps/terminal_bonds.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+def conf_map(x, bond, param) -> Tuple[Array, Array]:
+    """Apply map to a single bond in a conformer x, return (updated x, logdetjac)"""
+    a, b = bond
+    dim = 3
+
+    def f(xy, param):
```
jkausrelay, line 67: nit: change f to a more descriptive name

timemachine/maps/terminal_bonds.py
```diff
@@ -0,0 +1,236 @@
...
+
+    return x_updated, logdetjac
+
+
+apply_conf_map_to_traj = jit(vmap(conf_map, in_axes=(0, None, None)))
+
+
+def apply_conf_maps_to_traj(xs, bond_idxs, params) -> Tuple[Array, Array]:
+    """Apply maps to several bonds in a trajectory of conformers xs, return (updated xs, logdetjacs)"""
+    xs_traj, logdetjac_increments_traj = [jnp.array(xs)], [np.zeros(len(xs))]
```
jkausrelay, line 105: Does xs_traj need to be a list?

timemachine/maps/terminal_bonds.py
```diff
@@ -0,0 +1,236 @@
...
+    bonds = [sort_by_descending_degree(bond) for bond in bond_idxs]
+
+    is_terminal = lambda bond: degrees[bond[1]] == 1
+
+    terminal_bonds = sorted(filter(is_terminal, bonds))
+    return terminal_bonds
+
+
+class TerminalMappableState:
+    def __init__(self, terminal_bond_idxs, ks, eq_lengths, temperature=DEFAULT_TEMP, sigma_thresh=20):
```
jkausrelay, line 138: nit: Define DEFAULT_SIGMA_THRESHOLD=20 somewhere, since it appears repeatedly

timemachine/maps/terminal_bonds.py
```diff
@@ -0,0 +1,236 @@
...
+
+        terminal_bond_tuples = get_terminal_bonds(bond_idxs)  # each tuple will be sorted in order (anchor, terminal)
+
+        ks, eq_lengths = np.array([param_dict[tuple(sorted(bond))] for bond in terminal_bond_tuples]).T
+
+        return cls(np.array(terminal_bond_tuples), ks, eq_lengths, temperature=temperature, sigma_thresh=sigma_thresh)
+
+
+def states_to_conf_map_params(src: TerminalMappableState, dst: TerminalMappableState):
+    # find bond idxs in common
```
jkausrelay, line 203: Can you add a brief summary of this function?

tests/test_terminal_bond_maps.py
```diff
@@ -0,0 +1,170 @@
...
+
+    src_mol = Chem.MolFromSmiles("C([H])([H])([H])([H])", parser_params)
+    dst_mol = Chem.MolFromSmiles("C(Br)(Cl)(F)(I)", parser_params)
+    AllChem.EmbedMolecule(src_mol)
+    AllChem.EmbedMolecule(dst_mol)
+
+    # define src and dst states
+    u_src = get_vacuum_u_fxn(src_mol)
+    src_params, src_bond_idxs = get_hb_params(src_mol)
+    src_state = TerminalMappableState.from_harmonic_bond_params(src_bond_idxs, src_params)
```
jkausrelay, line 104: Do any angle terms change in this example?

timemachine/maps/terminal_bonds.py
jkausrelay: `PositiveInterval` sgtm

timemachine/maps/terminal_bonds.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/801/commits/c3c8d0a4f15211de051e53e64fb8f86ea9037a01

timemachine/maps/terminal_bonds.py
maxentile: No -- could switch to a fixed-size array with leading dimension = `len(bond_idxs)` if that would be preferable

jkausrelay: I meant just replace it with a variable equal to the last value:
```
    xs_updated = jnp.array(xs)
    for (bond, param) in zip(bond_idxs, params):  # TODO: jax.lax for-loop?
        xs_updated, logdetjac_increments = apply_conf_map_to_traj(xs_updated, bond, param)
        logdetjac_increments_traj.append(logdetjac_increments)

    return xs_updated, np.sum(logdetjac_increments_traj, axis=0)
```

timemachine/maps/terminal_bonds.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/801/commits/d99f460f5eab2f0afb218e4cce8f26a75034d03d

jkausrelay:
Re:
> Anyway, most suggestions were applied except for switching the implementation of an assertion from nan-poison to jax.experimental.checkify, and switching test assertions from almost_equal to allclose.

Do you want to implement these, or ignore for now?

timemachine/maps/terminal_bonds.py
```diff
@@ -0,0 +1,244 @@
+"""Summarize HarmonicBonds as Intervals in R+, construct maps between Intervals, & apply these maps to bond lengths"""
```
jkausrelay, line 1: Can you add documentation to clarify when this can be used?

tests/test_terminal_bond_maps.py
maxentile: Yes -- this is why standard deviation of mapped works in this example is not 0.

jkausrelay: Right, can you add this as a comment below

timemachine/maps/terminal_bonds.py
maxentile: Added in https://github.com/proteneer/timemachine/pull/801/commits/5db990247755a73f4b328962114a9960a862c4f4

timemachine/maps/terminal_bonds.py
maxentile: Ahh, I misunderstood -- applied this change in https://github.com/proteneer/timemachine/pull/801/commits/652211ab83c4f24e6b7900382920e122003e06e2

tests/test_terminal_bond_maps.py
maxentile: Added comment in https://github.com/proteneer/timemachine/pull/801/commits/9130b4d23fb0bf8b0d4814c239314556f6e7d882

timemachine/maps/terminal_bonds.py
maxentile: Added docstring in https://github.com/proteneer/timemachine/pull/801/commits/d4339076819ebb2aaf464babb646e0b9a5199b6c .

Note that you don't need to call this function directly -- intended usage is [`TerminalBondMap.from_states(src, dst)`](https://github.com/proteneer/timemachine/blob/d4339076819ebb2aaf464babb646e0b9a5199b6c/timemachine/maps/terminal_bonds.py#L265) , as [in the tests](https://github.com/proteneer/timemachine/blob/d4339076819ebb2aaf464babb646e0b9a5199b6c/tests/test_terminal_bond_maps.py#L122-L124), but figured I should expose the implementation as a free function as well, for ease of tinkering.

tests/test_mapped_estimators.py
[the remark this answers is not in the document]
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/801/commits/ad01ff3151c15cbebf36f8715110614567af1b4d

maxentile:
>Anyway, most suggestions were applied except for switching the implementation of an assertion from nan-poison to jax.experimental.checkify, and switching test assertions from almost_equal to allclose.
>Do you want to implement these, or ignore for now?

Thanks for the reminder --
* checkify -- I think I'll hold off until this feature is moved out of `jax.experimental`
* almost_equal -> allclose -- done

jkausrelay (approved):

mcwitt (approved):

PR #803, opened 2022-08-22
Baseline method for resampling p(x[dummy] | x[anchor])

maxentile:
- [x] Add a "brute-force" method for resampling the coordinates of dummy atoms, conditioned on the coordinates of anchor atoms
- [x] Add a basic test (asserting that only the expected set of updated atom idxs are updated)
- [ ] TBD[current PR]: think of + add a more stringent test
- [ ] TBD[later PR]: apply performance optimizations (rather than use the whole intramolecular `U_fn` with some coords frozen, construct and use a cheaper `U_fn` using information from `timemachine.fe.single_topology_v3.find_dummy_groups_and_anchors`)

----------------------

Illustration:

For the methane -> fluoromethane example in the test (with a single dummy atom at each endstate), the resampled atoms flap around as expected in each endstate:

image


timemachine/fe/utils.py
```diff
@@ -402,3 +409,62 @@ def delta_U(from_idx, to_idx):
...
+
+
+def minimize_energy(U_fxn, x0, verbose=True):
+    """using L-BFGS"""
+
+    fun = wrap_for_scipy_optimize(U_fxn, x0)
+    res = minimize(fun, x0.flatten(), jac=True, method="L-BFGS-B")
+    if verbose:
+        print(res)
+    return res.x.reshape(x0.shape)
```
maxentile, line 433: Introduces duplication with https://github.com/proteneer/timemachine/blob/df26704fdd7e5c67abf391348d37ce8c91753ea6/timemachine/fe/system.py#L14-L34 -- should merge

timemachine/fe/utils.py
```diff
@@ -402,3 +409,62 @@ def delta_U(from_idx, to_idx):
...
+    friction = 1
+    n_steps = 1_000
+
+    force = jit(lambda x_subset: -grad(U_subset)(x_subset))
+    intg = LangevinIntegrator(force, masses, temperature=temperature, dt=dt, friction=friction)
+
+    v0_subset = np.random.randn(*x0_subset.shape) * np.sqrt(BOLTZ * temperature / masses[:, np.newaxis])
+    assert v0_subset.shape == x0_subset.shape
+
+    xs, _ = intg.multiple_steps(x0_subset, v0_subset, n_steps)
```
maxentile, line 466: Possibly replace with call to `timemachine.fe.system.simulate_system` https://github.com/proteneer/timemachine/blob/df26704fdd7e5c67abf391348d37ce8c91753ea6/timemachine/fe/system.py#L37

timemachine/fe/utils.py
```diff
@@ -402,3 +409,62 @@ def delta_U(from_idx, to_idx):
...
+
+    # return final state of MD trajectory
+    # TODO[decision]: return just one sample? or return a collection of samples?
+    return x0.at[selection_mask].set(xs[-1])
```
maxentile, line 470: Usage: should this be hard-coded to return 1 sample, or should the number of samples returned be controlled by an argument `n_samples`?

(I restricted this assuming we would only ever need to look at 1 sample from this function, but maybe there are cases where > 1 sample would be useful...)

timemachine/fe/utils.py
```diff
@@ -250,3 +257,62 @@ def delta_U(from_idx, to_idx):
...
+    x0 = jnp.array(x0)
+    x0_subset = x0[selection_mask]
+
+    # prepare energy function defined on selected atoms only
+    def U_subset(x_subset):
+        return U(x0.at[selection_mask].set(x_subset))
+
+    # optionally minimize
+    if minimize_first:
+        x0_subset = minimize_energy(jit(U_subset), x0_subset)
```
maxentile, line 299: May be superseded by https://github.com/proteneer/timemachine/pull/821/files#diff-bb2d19446b7ea37b639b8b5a06a398d7be930ad81e47d67e143e2da183f07e1cR325-R343

PR #804, opened 2022-08-24
Move hif2a_nowater_min.pdb to testsystems

jkausrelay:
- Move hif2a_nowater_min.pdb to testsystems
- Use resources to find pdb path instead of manual path manipulation


mcwitt (approved):
LGTM

PR #806, opened 2022-08-25
Local MD: No interaction group

### LINKED ISSUE #792: Move processing of integrator parameters into C++
Move the pre-processing of the integrator parameters to the C++.

Should also attempt to clarify the names

_Originally posted by @maxentile in https://github.com/proteneer/timemachine/pull/789#discussion_r942600565_

badisa:
- Adds Local MD
- Resolves #792 
- Other minor aesthetic changes 

Currently this is implemented without the Interaction group. The logic becomes more complex as you have to construct a bound potential in C++ from the existing and potentially nested Nonbonded object. In initial pass it didn't behave, and will leave to another PR. 

Benchmarks
image


timemachine/cpp/src/bound_potential.cu
```diff
@@ -37,9 +37,9 @@ void BoundPotential::execute_host(
     gpuErrchk(cudaMemset(d_du_dl.data, 0, d_du_dl.size));
     gpuErrchk(cudaMemset(d_u.data, 0, d_u.size));
 
-    this->execute_device(
-        N, d_x.data, d_box.data, lambda, d_du_dx.data, nullptr, d_du_dl.data, d_u.data, static_cast<cudaStream_t>(0));
-
+    cudaStream_t stream = static_cast<cudaStream_t>(0);
+    this->execute_device(N, d_x.data, d_box.data, lambda, d_du_dx.data, nullptr, d_du_dl.data, d_u.data, stream);
+    gpuErrchk(cudaStreamSynchronize(stream));
```
badisa, line 42: If we want to use streams more fully we should be more explicit. Not all streams sync with the default stream which can cause unexpected behavior.

This was added as the error for the execute device popped up in the copy line due to an implicit sync.

timemachine/cpp/src/bound_potential.cu
```diff
@@ -37,14 +37,30 @@ void BoundPotential::execute_host(
...
     d_du_dx.copy_to(h_du_dx);
     d_du_dl.copy_to(h_du_dl);
     d_u.copy_to(h_u);
 };
 
+void BoundPotential::set_params_device(
+    const std::vector<int> device_shape, const double *d_new_params, const cudaStream_t stream) {
```
mcwitt, line 49: nit: could this take a `DeviceBuffer<double>` instead of the raw buffer?
```suggestion
    const std::vector<int> device_shape, const DeviceBuffer<double> &new_params, const cudaStream_t stream) {
```

mcwitt: Would `new_shape` be a more descriptive name than `device_shape`?

timemachine/cpp/src/bound_potential.cu
```diff
@@ -37,14 +37,30 @@ void BoundPotential::execute_host(
...
+    int updated_size = 1;
+    for (auto s : device_shape) {
+        updated_size *= s;
+    }
+    if (updated_size > 0) {
+        if (updated_size > this->size()) {
+            d_p.reset(new DeviceBuffer<double>(updated_size));
+        }
+        gpuErrchk(cudaMemcpyAsync(
+            d_p->data, d_new_params, updated_size * sizeof(*d_p->data), cudaMemcpyDeviceToDevice, stream));
```
mcwitt, line 59: If the argument is changed to a DeviceBuffer, this could be
```suggestion
            d_p->data, new_params.data, new_params.size, cudaMemcpyDeviceToDevice, stream));
```

timemachine/cpp/src/constants.hpp
```diff
@@ -0,0 +1,3 @@
+
+static const double BOLTZ = 0.008314462618;
+static const double AVOGADRO = 6.0221367e23;
```
mcwitt, line 3: 👍

timemachine/cpp/src/context.cu
```diff
@@ -50,6 +64,342 @@ Context::~Context() {
...
+               nb_pot) {
+        return nb_pot->get_cutoff();
+    } else if (std::shared_ptr<NonbondedAllPairs<double, false>> nb_pot =
+                   std::dynamic_pointer_cast<NonbondedAllPairs<double, false>>(pot);
+               nb_pot) {
+        return nb_pot->get_cutoff();
+    } else {
+        throw std::runtime_error("Unable to cast potential to NonbondedAllPairs");
+    }
+}
```
mcwitt, line 87: nit: It feels cleaner to me to split this into 2 pieces:
1. Potential -> Optional[NonbondedAllPairs]
2. NonbondedAllPairs -> cutoff

mcwitt: That this is necessary makes me want to consider moving `local_md` into a subclass of Context with sigature like (Python pseudocode):
```
LocalMDContext(nonbonded: NonbondedAllPairs, others: List[Potential], ...)
```

timemachine/cpp/src/context.cu
```diff
@@ -50,6 +64,342 @@ Context::~Context() {
...
+               nb_pot) {
+        nb_pot->set_atom_idxs_device(num_idxs, d_idxs, stream);
+    } else if (std::shared_ptr<NonbondedAllPairs<double, false>> nb_pot =
+                   std::dynamic_pointer_cast<NonbondedAllPairs<double, false>>(pot);
+               nb_pot) {
+        nb_pot->set_atom_idxs_device(num_idxs, d_idxs, stream);
+    } else {
+        throw std::runtime_error("unable to cast potential to NonbondedAllPairs");
+    }
+}
```
mcwitt, line 131: That this verbose casting and error checking is necessary makes me want to consider moving `local_md` into a subclass of Context with sigature like (Python pseudocode):
```
LocalMDContext(nonbonded: NonbondedAllPairs, others: List[Potential], ...)
```

timemachine/cpp/src/context.cu
```diff
@@ -50,6 +64,342 @@ Context::~Context() {
...
+
+void Context::_assert_temperature_matches(const double temperature) {
```
mcwitt, line 133: nit: name sounds general, but this is specific to local MD?

timemachine/cpp/src/device_buffer.cu
```diff
@@ -4,34 +4,60 @@
...
+        gpuErrchk(cudaMallocHost(&buffer, length * sizeof(T)));
+    } else {
+        gpuErrchk(cudaMalloc(&buffer, length * sizeof(T)));
+    }
     return buffer;
 }
 
-template <typename T>
-DeviceBuffer<T>::DeviceBuffer(const std::size_t length) : size(length * sizeof(T)), data(allocate<T>(length)) {}
+template <typename T, bool PINNED>
```
mcwitt, line 20: It seems like it might be cleaner to introduce a separate class for pinned buffers, since the implementations seem almost entirely disjoint?

timemachine/cpp/src/gpu_utils.cu
```diff
@@ -1,5 +1,13 @@
 #include "gpu_utils.cuh"
 
+int round_up_even(int count) {
+    if (count % 2 == 1) {
+        return count += 1;
+    } else {
+        return count;
+    }
```
mcwitt, line 8: nit: simplify (also I think `+=` is a typo?)
```suggestion
    return count + (count % 2)
```

timemachine/cpp/src/kernels/k_indices.cu
```diff
@@ -0,0 +1,38 @@
+#include "k_indices.cuh"
+
+// Takes a source and destination array. Assumes K <= N with values in the src are less than or equal
+// to K. The value of the src is used as the indice and the value in the destination array. Allows combining
+// a series of indices to get a unique set of values.
```
mcwitt, line 5: Might be worth documenting with equivalent numpy code. I think it'd be something like `dst[src] = src` in this case?

mcwitt (commented):
Looking good so far! Just took a high-level initial pass; will spend more time early next week on understanding the algorithmic changes

tests/test_md.py
```diff
@@ -354,7 +357,114 @@ def compute_reference_values():
...
+        ctxt = custom_ops.Context(coords, v0, box, intg_impl, bps * 2)
+        with pytest.raises(RuntimeError) as e:
+            ctxt.local_md(np.zeros(100), local_idxs, radius=radius, temperature=200.0)
+        assert "Local MD temperature didn't match Thermostat's temperature." == str(e.value)
+
+        # Construct context with no potentials, local MD should fail.
+        ctxt = custom_ops.Context(coords, v0, box, intg_impl, [])
+        with pytest.raises(RuntimeError) as e:
+            ctxt.local_md(np.zeros(100), local_idxs, radius=radius)
+        assert "unable to find a NonbondedAllPairs potential" == str(e.value)
```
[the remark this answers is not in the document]
badisa: I took this approach as it puts less burden on the user to track potentials, which I don't find trivial especially since we combined `NonbondedAllPairs` and `NonbondedPairList` in python using a `NonbondedImplWrapper`. This also seems easily testable, so don't see much risk in this approach, but subclass is doable.

tests/test_velocity_verlet_integrator.py
```diff
@@ -184,3 +185,25 @@ def test_initialization_and_finalization():
...
+
+    mol_a = rfe.mol_a
+
+    local_idxs = np.arange(len(coords) - mol_a.GetNumAtoms(), len(coords), dtype=np.uint32)
+
+    intg_impl, ctxt = setup_velocity_verlet(bound_potentials, coords, solvent_box, dt, masses)  # noqa
+
+    n_steps = 10
+
+    ctxt.local_md(np.ones(n_steps), local_idxs, temperature=100)
```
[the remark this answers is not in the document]
badisa: Will add comment, was just to test a value that wasn't the default in the case that the integrator is not a thermostat.

timemachine/cpp/src/bound_potential.cu
badisa: Right now our convention is that the `_device` methods take raw pointers. Not sure I would want to change it anyways in the case that you have say a slice of a DeviceBuffer.

timemachine/cpp/src/context.cu
badisa: Right now it is specific to local MD. But in any case where MD moves are dependent on temperature, we will want to verify that the thermostat agrees. Could rename in the mean time.

timemachine/cpp/src/device_buffer.cu
badisa: Will do

timemachine/cpp/src/context.hpp
```diff
@@ -34,6 +34,15 @@ class Context {
         int store_u_interval,
         int store_x_interval);
 
+    std::array<std::vector<double>, 2> local_md(
```
mcwitt, line 37: For naming consistency with the global MD methods, could this be called `multiple_steps_local` or similar?

timemachine/cpp/src/context.cu
badisa: As far as I understand templates, there is not generic NonbondedAllPairs type. It needs an explicit template values for the return return in 1) and the passed in value for 2). I have looked around, but might have missed it.

timemachine/cpp/src/context.cu
badisa: Refer to https://github.com/proteneer/timemachine/pull/806#discussion_r957379690

timemachine/cpp/src/context.hpp
badisa: Done in [574268f](https://github.com/proteneer/timemachine/pull/806/commits/574268fd50e5ae48be677b1522d637a33284bfba)

timemachine/cpp/src/gpu_utils.cu
badisa: Done in [30e42fa](https://github.com/proteneer/timemachine/pull/806/commits/30e42fa1ea70ae75a9506120bdc19a1bb2bc5f6b)

tests/test_md.py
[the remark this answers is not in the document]
mcwitt: > which I don't find trivial especially since we combined NonbondedAllPairs and NonbondedPairList in python using a NonbondedImplWrapper

Ah, good point that the user may not be aware of the `NonbondedAllPairs` behind the scenes if they're using `Nonbonded`. So maybe it'd make sense for `LocalMDContext` to accept either `Nonbonded` or `NonbondedAllPairs` as a separate argument? In any case, I don't mean for this to be a blocker. Maybe we can revisit during the next deboggling week?

timemachine/cpp/src/context.cu
mcwitt: Makes sense; the suggestion seems awkward to implement after some more thought.

badisa:
Closing this to avoid distraction, as other issues are more pressing..

PR #807, opened 2022-08-26
Print warning and fall back to default arch if nvidia-smi command fails

mcwitt:
Small quality of life improvement. Previously, if `nvidia-smi` returned an error, `CUDA_ARCH` would be set to the failure message, resulting in failures further on in the build.

This adds a check for the exit code of the `nvidia-smi` command. If the code indicates an error, we print a warning including the output and fall back to the default `CUDA_ARCH` (as we do if `nvidia-smi` is not found).


badisa (approved):
Nice!

PR #808, opened 2022-08-29
Add option for rbfe eq_steps and fix chiral restraint issue

jkausrelay:
- Add option to change the number of eq steps (currently is only 10 ps)
- Fix issue in checking for duplicates of chiral restraints (np array is not hashable)


timemachine/fe/interpolate.py
```diff
@@ -87,7 +87,7 @@ def make_kv(all_idxs, all_params):
 
         def has_duplicates(x):
             x = list(x)
-            return len(set(x)) < len(x)
+            return len(set(to_hashable(x))) < len(x)
```
mcwitt, line 90: Was this change left in unintentionally? (tests still pass without it)

timemachine/fe/interpolate.py
```diff
@@ -138,8 +138,8 @@ def align_chiral_bond_idxs_and_params(src_idxs, src_params, src_signs, dst_idxs,
             dst_idxs,
             zip(dst_signs, dst_params),
             make_default=lambda p: (p[0], 0),  # p[0] is sign, 0 is force constant
-            key=lambda idxs, p: (idxs, p[0]),  # use idxs and sign as key
-            get_idxs=lambda key: key[0],
+            key=lambda idxs, p: (tuple(idxs), p[0]),  # use idxs and sign as key
+            get_idxs=lambda key: tuple(key[0]),
```
mcwitt, line 142: another (maybe nicer?) fix might be to change the definition of `to_hashable` to also convert nested arrays to tuples. This seems to work:
```python
    def to_hashable(x):
        return tuple(to_hashable(e) for e in x) if isinstance(x, Iterable) else x
```

timemachine/fe/rbfe.py
```diff
@@ -431,38 +435,46 @@ def estimate_relative_free_energy(
...
-    try:
-
-        return estimate_free_energy_given_initial_states(
-            initial_states, protocol, temperature, combined_prefix, keep_idxs
-        )
-
-    except Exception as old_exc:
-
-        raise SimulationException(initial_states, protocol, combined_prefix) from old_exc
+    return estimate_free_energy_given_initial_states(initial_states, protocol, temperature, combined_prefix, keep_idxs)
```
mcwitt, line 445: Was this change intentional? Seems like we still want the `SimulationException` to be able to attach additional info?

mcwitt (changes requested):
Thanks for fixing the chiral bond interpolation bug! Looks good overall. Was wondering why we're no longer raising a `SimulationException`?

timemachine/fe/interpolate.py
jkausrelay: reverted. This was my first attempt at a fix, and isn't needed anymore

timemachine/fe/interpolate.py
jkausrelay: updated

timemachine/fe/rbfe.py
jkausrelay: Updated to store the pickle on exception in rbfe

mcwitt (approved):

PR #811, opened 2022-08-30
Add overlap summary fields, plot to RBFE simulation results

mcwitt:
Adds and populates new fields of the RBFE simulation result object (`timemachine.fe.rbfe.SimulationResult`):

- **`overlaps_by_lambda`**: `(L-1,)` array of BAR overlaps between neighboring pairs of states
- **`overlaps_by_term_lambda`**: `(len(potentials), L-1)` array of BAR overlaps between neighboring pairs of states, for each potential (bond, angle, etc.)
- **`overlap_summary_png`**: plot summarizing the above data (as PNG-encoded binary buffer, similar to existing `plot_png`. Example:
   image

Also renames `plot_png` to `overlap_detail_png` to disambiguate with the new plot.

### DIFF
--- a/timemachine/fe/rbfe.py
+++ b/timemachine/fe/rbfe.py
@@ -151,7 +151,7 @@ class InitialState:
 class SimulationResult:
     all_dGs: List[np.ndarray]
     all_errs: List[float]
-    plot_png: bytes
+    overlap_detail_png: bytes
     frames: List[np.ndarray]
     boxes: List[np.ndarray]
     initial_states: List[InitialState]
@@ -360,9 +360,17 @@ def estimate_free_energy_given_initial_states(initial_states, protocol, temperat
     buffer = io.BytesIO()
     plt.savefig(buffer, format="png")
     buffer.seek(0)
-    img_as_bytes = buffer.read()
+    overlap_detail_png = buffer.read()
 
-    return SimulationResult(all_dGs, all_errs, img_as_bytes, stored_frames, stored_boxes, initial_states, protocol)
+    return SimulationResult(
+        all_dGs,
+        all_errs,
+        overlap_detail_png,
+        stored_frames,
+        stored_boxes,
+        initial_states,
+        protocol,
+    )
 
 
 def estimate_relative_free_energy(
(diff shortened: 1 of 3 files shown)


timemachine/fe/rbfe.py
```diff
@@ -311,23 +326,29 @@ def estimate_free_energy_given_initial_states(initial_states, protocol, temperat
...
-                all_fwd_delta_us.append(fwd_delta_u)
-                all_rev_delta_us.append(rev_delta_u)
 
             # sanity check - I don't think the dG calculation commutes with its components, so we have to re-estimate
             # the dG from the sum of the delta_us as opposed to simply summing the component dGs
-            total_fwd_delta_us = np.sum(all_fwd_delta_us, axis=0)
-            total_rev_delta_us = np.sum(all_rev_delta_us, axis=0)
+            ukln_by_term = np.array(ukln_by_term, dtype=np.float64)
+            total_fwd_delta_us = (ukln_by_term[:, 1, 0, :] - ukln_by_term[:, 0, 0, :]).sum(axis=0)
+            total_rev_delta_us = (ukln_by_term[:, 0, 1, :] - ukln_by_term[:, 1, 1, :]).sum(axis=0)
```
maxentile, line 351: I found this difficult to read at first, but from comparison to L335-339 it looks correct.

Suggestion: add a comment to `ukln_by_term` to name the 4 axes
```suggestion
            # (num components, energy fxn idx, sampled state idx, num samples)
            ukln_by_term = np.array(ukln_by_term, dtype=np.float64)
            total_fwd_delta_us = (ukln_by_term[:, 1, 0, :] - ukln_by_term[:, 0, 0, :]).sum(axis=0)
            total_rev_delta_us = (ukln_by_term[:, 0, 1, :] - ukln_by_term[:, 1, 1, :]).sum(axis=0)
```

maxentile (approved):
Nice! Good to be able to summarize the calculation on a single plot (in addition to the more detailed set of ((# lambda windows - 1) x # energy components) plots)

timemachine/fe/rbfe.py
mcwitt: Thanks! Applied in [68229f1](https://github.com/proteneer/timemachine/pull/811/commits/68229f10602559de212a3a24cff6bf7bf78b331f). I also slightly prefer your usage of "component" instead of "term" and updated the naming in [2dd0942](https://github.com/proteneer/timemachine/pull/811/commits/2dd09428c53aac1fb88d0087c88f8a0e6b305663).

PR #812, opened 2022-08-30
Add fallback to avoid trivial atom mappings in some cases

maxentile:
In some cases involving ring size changes, RDKit 2021.03.1 MCS with [our preferred settings](https://github.com/proteneer/timemachine/blob/13ca9fd3f27210a7a116a7ca2747b5cd5b0cc4f6/timemachine/fe/atom_mapping.py#L37-L46) returns a trivial substructure

image

(this seems to be due to a bad interaction between explicit hydrogens and the `CompleteRingsOnly = 1` setting)

--------

However, MCS with these same settings works just fine if we `RemoveHs`:
image

--------

This PR adds a fall-back to reseed the MCS with the result of an easier MCS on `RemoveHs(a), RemoveHs(b)`

image

### DIFF
--- a/attic/atom_mapping/mcs_atom_compare.py
+++ b/attic/atom_mapping/mcs_atom_compare.py
@@ -18,3 +18,29 @@ class CompareDist(rdFMCS.MCSAtomCompare):
         x_i = mol1.GetConformer(0).GetPositions()[atom1]
         x_j = mol2.GetConformer(0).GetPositions()[atom2]
         return bool(np.linalg.norm(x_i - x_j) <= self.threshold)  # must convert from np.bool_ to Python bool!
+
+
+
+class CompareDistNonterminal(rdFMCS.MCSAtomCompare):
+    """
+    Custom comparator used in the FMCS code.
+    This allows two atoms to match if:
+        1. Neither atom is a terminal atom (H, F, Cl, Halogens etc.)
+        2. They are within 1 angstrom of each other.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def compare(self, p, mol1, atom1, mol2, atom2):
+
+        if mol1.GetAtomWithIdx(atom1).GetDegree() == 1:
+            return False
+        if mol2.GetAtomWithIdx(atom2).GetDegree() == 1:
+            return False
+
+        x_i = mol1.GetConformer(0).GetPositions()[atom1]
+        x_j = mol2.GetConformer(0).GetPositions()[atom2]
+
+        threshold = 1.0  # angstroms
+        return bool(np.linalg.norm(x_i - x_j) <= threshold)
--- a/examples/relative_free_energy.py
+++ b/examples/relative_free_energy.py
@@ -39,7 +39,7 @@ def run_pair(mol_a, mol_b, core, forcefield, n_frames, protein_path, seed):
     )
 
     with open("solvent_overlap.png", "wb") as fh:
-        fh.write(solvent_res.plot_png)
+        fh.write(solvent_res.overlap_detail_png)
 
     # this st is only needed to deal with visualization jank
     write_trajectory_as_pdb(mol_a, mol_b, core, solvent_res.frames, solvent_top, "solvent_traj")
@@ -53,7 +53,7 @@ def run_pair(mol_a, mol_b, core, forcefield, n_frames, protein_path, seed):
         mol_a, mol_b, core, forcefield, complex_host_config, seed + 1, n_frames=n_frames, prefix="complex"
     )
     with open("complex_overlap.png", "wb") as fh:
-        fh.write(complex_res.plot_png)
+        fh.write(complex_res.overlap_detail_png)
     write_trajectory_as_pdb(mol_a, mol_b, core, complex_res.frames, complex_top, "complex_traj")
 
     print(f"complex dG: {np.sum(solvent_res.all_dGs):.3f} +- {np.linalg.norm(solvent_res.all_errs):.3f} kJ/mol")
--- a/examples/relative_free_energy.py
+++ b/examples/relative_free_energy.py
@@ -103,7 +103,7 @@ def read_from_args():
     mol_b = get_mol_by_name(mols, args.mol_b_name)  # 30 in test pair
 
     print("Searching for the maximum common substructure...")
-    mcs_result = atom_mapping.mcs(mol_a, mol_b, conformer_aware=False, timeout=60)
+    mcs_result = atom_mapping.mcs(mol_a, mol_b)
     query_mol = Chem.MolFromSmarts(mcs_result.smartsString)
 
     print("mol_a SMILES:", Chem.MolToSmiles(mol_a, isomericSmiles=False))
--- a/examples/relative_free_energy.py
+++ b/examples/relative_free_energy.py
@@ -103,7 +103,7 @@ def read_from_args():
     mol_b = get_mol_by_name(mols, args.mol_b_name)  # 30 in test pair
 
     print("Searching for the maximum common substructure...")
-    mcs_result = atom_mapping.mcs_map_graph_only_complete_rings(mol_a, mol_b)
+    mcs_result = atom_mapping.mcs_graph_only_complete_rings(mol_a, mol_b)
     query_mol = Chem.MolFromSmarts(mcs_result.smartsString)
 
     print("mol_a SMILES:", Chem.MolToSmiles(mol_a, isomericSmiles=False))
--- a/examples/relative_free_energy.py
+++ b/examples/relative_free_energy.py
@@ -103,7 +103,7 @@ def read_from_args():
     mol_b = get_mol_by_name(mols, args.mol_b_name)  # 30 in test pair
 
     print("Searching for the maximum common substructure...")
-    mcs_result = atom_mapping.mcs(mol_a, mol_b, conformer_aware=False)
+    mcs_result = atom_mapping.mcs(mol_a, mol_b, conformer_aware=False, timeout=60)
     query_mol = Chem.MolFromSmarts(mcs_result.smartsString)
 
     print("mol_a SMILES:", Chem.MolToSmiles(mol_a, isomericSmiles=False))
--- a/examples/relative_free_energy.py
+++ b/examples/relative_free_energy.py
@@ -103,7 +103,7 @@ def read_from_args():
     mol_b = get_mol_by_name(mols, args.mol_b_name)  # 30 in test pair
 
     print("Searching for the maximum common substructure...")
-    mcs_result = atom_mapping.mcs_graph_only_complete_rings(mol_a, mol_b)
+    mcs_result = atom_mapping.mcs(mol_a, mol_b, conformer_aware=False)
     query_mol = Chem.MolFromSmarts(mcs_result.smartsString)
 
     print("mol_a SMILES:", Chem.MolToSmiles(mol_a, isomericSmiles=False))
--- a/tests/test_atom_mapping.py
+++ b/tests/test_atom_mapping.py
@@ -1,18 +1,18 @@
 from rdkit import Chem
 
-from timemachine.fe.atom_mapping import get_core_by_mcs, mcs_map
+from timemachine.fe.atom_mapping import get_core_by_mcs, mcs_conformer_aware
 from timemachine.testsystems.relative import hif2a_ligand_pair
 
 
 def test_mcs_map():
-    mcs_result = mcs_map(hif2a_ligand_pair.mol_a, hif2a_ligand_pair.mol_b, threshold=0.5)
+    mcs_result = mcs_conformer_aware(hif2a_ligand_pair.mol_a, hif2a_ligand_pair.mol_b, threshold=0.5)
     assert mcs_result.queryMol is not None
     assert mcs_result.numAtoms > 1
 
 
 def test_get_core_by_mcs():
     mol_a, mol_b = hif2a_ligand_pair.mol_a, hif2a_ligand_pair.mol_b
-    query = mcs_map(mol_a, mol_b).queryMol
+    query = mcs_conformer_aware(mol_a, mol_b).queryMol
     core = get_core_by_mcs(mol_a, mol_b, query)
     assert core.shape[1] == 2
 
@@ -30,5 +30,5 @@ def test_ring_size_change():
     mol_a.Compute2DCoords()
     mol_b.Compute2DCoords()
 
-    mcs_result = mcs_map(mol_a, mol_b)
+    mcs_result = mcs_conformer_aware(mol_a, mol_b)
     assert mcs_result.numBonds > 1
--- a/tests/test_atom_mapping.py
+++ b/tests/test_atom_mapping.py
@@ -31,4 +31,4 @@ def test_ring_size_change():
     mol_b.Compute2DCoords()
 
     mcs_result = mcs_map(mol_a, mol_b)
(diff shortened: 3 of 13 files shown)


timemachine/fe/atom_mapping.py
```diff
@@ -51,6 +58,36 @@ def mcs_map(a, b, threshold: float = 2.0, timeout: int = 5, smarts: Optional[str
     return rdFMCS.FindMCS([a, b], params)
 
 
+def mcs_map(a, b, threshold: float = 2.0, timeout: int = 5, smarts: Optional[str] = None):
```
maxentile, line 61: Need to refactor + sort out some naming: 
* This reseeding fallback should be equally applicable to the conformer-aware default `mcs` and to the conformer-oblivious `mcs_map_graph_only_complete_rings`, but currently I only applied it to one... Should I factor out this logic so that it can be used with arbitrary `mcs_fxn(Mol, Mol) -> MCSResult`?
* Can we remove the word `_map` and use similar levels of descriptiveness for the two `mcs_fxns` in this file? (`mcs_map` -> `mcs_conformer_aware`, `mcs_map_graph_only_complete_rings` -> `mcs_graph_only_complete_rings`) "`_map`" seems misleading here since these don't return atom maps, but RDKit `MCSResult`s...

maxentile: Applied name change and factored out `possibly_fallback_to_heavy_atom_mcs`, but I didn't immediately find a clean way to implement using `functools.wraps` or similar.

Current refactoring looks like:
https://github.com/proteneer/timemachine/blob/5c7a0379ee086a442eff7191b232e6a2981c8066/timemachine/fe/atom_mapping.py#L104-L126

This is not ideal -- now introduces two copies of each MCS method -- requiring docstrings to be either duplicated or less informative, and requiring the signature to be partially duplicated...

maxentile: @mcwitt -- is there a better way to do this off the top of your head? Seems like this should have a neater solution using something in `functools`...

timemachine/fe/atom_mapping.py
```diff
@@ -64,9 +115,18 @@ def mcs_map_graph_only_complete_rings(a, b, timeout: int = 3600, smarts: Optiona
         matchChiralTag=False,
         atomCompare=Chem.rdFMCS.AtomCompare.CompareAny,
         bondCompare=Chem.rdFMCS.BondCompare.CompareAny,
+        seedSmarts=smarts if type(smarts) == str else "",
     )
 
 
+def mcs_graph_only_complete_rings(a, b, timeout: int = 3600, smarts: Optional[str] = None):
```
badisa, line 122: We should probably reduce this default timeout. An hour to find MCS seems extreme as a default. 
```suggestion
def mcs_graph_only_complete_rings(a, b, timeout: int = 60, smarts: Optional[str] = None):
```

maxentile: Agreed -- that seems like a good change to introduce.

Is there a reason this timeout should remain higher than for the conformer-aware MCS?

(Before this PR, conformer-aware MCS has timeout = 5 seconds and graph-only MCS has timeout = 1 hr, but perhaps both should = 5 seconds or both = 1 minute.)

timemachine/fe/atom_mapping.py
```diff
@@ -64,9 +115,18 @@ def mcs_map_graph_only_complete_rings(a, b, timeout: int = 3600, smarts: Optiona
         matchChiralTag=False,
         atomCompare=Chem.rdFMCS.AtomCompare.CompareAny,
         bondCompare=Chem.rdFMCS.BondCompare.CompareAny,
+        seedSmarts=smarts if type(smarts) == str else "",
```
jkausrelay, line 118: nit: `smarts or ""`

timemachine/fe/atom_mapping.py
maxentile: Applied in https://github.com/proteneer/timemachine/pull/812/commits/354f0dc3efa751a8d2053b0614ebc9f05c1a1977

timemachine/fe/atom_mapping.py
maxentile: Applied in https://github.com/proteneer/timemachine/pull/812/commits/c8f253662ec35f6c44a7b0718c761c88e8622479

timemachine/fe/atom_mapping.py
maxentile: (Side note: increasing timeout from 5 seconds to 1 minute for conformer-aware MCS does not avoid timing out and returning a trivial (1-bond) substructure in the example in https://github.com/proteneer/timemachine/pull/812#issue-1356239660 )

maxentile:
Another thought -- would it make sense to consolidate `mcs_conformer_aware` and `mcs_graph_only_complete_rings`, and just have an option to toggle `mcs(a, b, ..., conformer_aware=True)`?

Aside from conformer-awareness, there are a couple minor differences between the two functions but I don't think they're significant. (Passing an `MCSParameters` object to `FindMCS` instead of using keywords, and whether to explicitly set `MatchChiralTag = False` vs. rely on this being the default.)

timemachine/fe/atom_mapping.py
maxentile: Simplified by consolidating the two graph-only and conformer-aware mcs fxns in this file and inlining this check -- may need to refactor again if we reintroduce multiple mcs fxns in this file

timemachine/fe/atom_mapping.py
maxentile: Consolidated with default timeout = 5 seconds, and set timeout = 60 seconds in the one usage where conformer_aware=False https://github.com/proteneer/timemachine/pull/812/commits/7588bcc945c7b701540d3bca6f546177853d1650

attic/atom_mapping/mcs_atom_compare.py
```diff
@@ -18,3 +18,29 @@ def compare(self, p, mol1, atom1, mol2, atom2):
         x_i = mol1.GetConformer(0).GetPositions()[atom1]
         x_j = mol2.GetConformer(0).GetPositions()[atom2]
         return bool(np.linalg.norm(x_i - x_j) <= self.threshold)  # must convert from np.bool_ to Python bool!
+
+
+
+class CompareDistNonterminal(rdFMCS.MCSAtomCompare):
```
maxentile, line 24: This was only used in one test of `setup_relative_restraints_using_smarts` that could just as well use any smarts string -- modified this test to generate smarts using MCS with `rdFMCS.AtomCompare.CompareAnyHeavyAtom` instead of MCS with this custom atom comparison, and moved this custom atom comparison to attic for future reference

timemachine/fe/atom_mapping.py
```diff
@@ -7,64 +8,64 @@
...
-    This allows two atoms to match if:
-        1. Neither atom is a terminal atom (H, F, Cl, Halogens etc.)
-        2. They are within 1 angstrom of each other.
-    """
-
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-
-    def compare(self, p, mol1, atom1, mol2, atom2):
+def mcs(a, b, threshold: float = 2.0, timeout: int = 5, smarts: Optional[str] = None, conformer_aware=True, retry=True):
```
maxentile, line 11: Renamed `mcs_map` -> `mcs` + consolidated with `mcs_map_graph_only_complete_rings`  which can now be emulated by setting `conformer_aware=False`

timemachine/fe/atom_mapping.py
badisa: > Is there a reason this timeout should remain higher than for the conformer-aware MCS?

Only suggested greater than 5 because it was larger initially, so there might have been a justification at some point for a longer time out. Like consolidating to 5 seconds and adjusting in example!

tests/test_atom_mapping.py
```diff
@@ -1,15 +1,34 @@
-from timemachine.fe.atom_mapping import get_core_by_mcs, mcs_map
+from rdkit import Chem
+
+from timemachine.fe.atom_mapping import get_core_by_mcs, mcs
 from timemachine.testsystems.relative import hif2a_ligand_pair
 
 
 def test_mcs_map():
-    mcs_result = mcs_map(hif2a_ligand_pair.mol_a, hif2a_ligand_pair.mol_b, threshold=0.5)
+    mcs_result = mcs(hif2a_ligand_pair.mol_a, hif2a_ligand_pair.mol_b, threshold=0.5)
```
mcwitt, line 8: nit: update test name?

timemachine/fe/atom_mapping.py
```diff
@@ -7,64 +8,64 @@
...
+        params.InitialSeed = heavy_atom_result.smartsString
+
+        result = rdFMCS.FindMCS([a, b], params)
+
+    if is_trivial(result):
+        message = f"""MCS result trivial!
+            timed out: {result.canceled}
+            # atoms in MCS: {result.numAtoms}
+            # bonds in MCS: {result.numBonds}
+        """
```
mcwitt, line 65: nit: to avoid unwanted whitespace in the string, can do
```suggestion
        message = "\n".join(
            [
                "MCS result trivial!",
                f"timed out: {result.canceled}",
                f"# atoms in MCS: {result.numAtoms}",
                f"# bonds in MCS: {result.numBonds}",
            ]
        )
```

or, alternatively, apply [`textwrap.dedent`](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) to the indented multiline string.

mcwitt (approved):
LGTM! (just a couple of nits)

attic/atom_mapping/mcs_atom_compare.py
maxentile: Further simplified the test to use a smarts string rather than go through (conformer-aware or graph-only) MCS at all in https://github.com/proteneer/timemachine/pull/812/commits/eb541de292986245b175198e228009c89781d0d9

tests/test_atom_mapping.py
maxentile: Ahh, good call -- fixed in https://github.com/proteneer/timemachine/pull/812/commits/23aa25ee24ca8368f63cedb91c1716a15b3fee8c

examples/relative_free_energy.py
```diff
@@ -103,7 +103,7 @@ def read_from_args():
     mol_b = get_mol_by_name(mols, args.mol_b_name)  # 30 in test pair
 
     print("Searching for the maximum common substructure...")
-    mcs_result = atom_mapping.mcs_map_graph_only_complete_rings(mol_a, mol_b)
+    mcs_result = atom_mapping.mcs(mol_a, mol_b, conformer_aware=False, timeout=60)
```
jkausrelay, line 106: I think this was an oversight, rather than intentional to use a different mcs mapping here. This is the old mapping prior to (https://github.com/proteneer/timemachine/pull/790/files#diff-83e3b80af4d75c2db550a5e4f928c4eb03e955ceacae8d6695e4e07cb1558df8).

tests/test_atom_mapping.py
```diff
@@ -1,15 +1,34 @@
...
+    smi_b = "C(C1CCC1)C1=CC2=C(CCC2)C=C1"
+
+    mol_a = Chem.AddHs(Chem.MolFromSmiles(smi_a))
+    mol_b = Chem.AddHs(Chem.MolFromSmiles(smi_b))
+
+    mol_a.Compute2DCoords()
+    mol_b.Compute2DCoords()
+
+    mcs_result = mcs(mol_a, mol_b)
+    assert mcs_result.numBonds > 1
```
jkausrelay, line 34: Can you sanity check that mol_a/mol_b aren't modified here?

tests/test_single_topology_v3.py
```diff
@@ -206,7 +206,7 @@ def get_max_distance(x0):
 
         print("Checking", get_mol_name(mol_a), "->", get_mol_name(mol_b))
         mcs_threshold = 0.75  # distance threshold, in nanometers
```
jkausrelay, line 208: 0.75 -> 2.0 for consistency?

tests/test_single_topology_v3.py
```diff
@@ -288,7 +288,7 @@ def setup_arbitary_transformation():
         mol_b = mols["57"]
 
         mcs_threshold = 0.75
-        res = atom_mapping.mcs_map(mol_a, mol_b, mcs_threshold)
+        res = atom_mapping.mcs(mol_a, mol_b, mcs_threshold)
```
jkausrelay, line 291: same with 0.75 here ^

tests/test_atom_mapping.py
maxentile: Good call -- added in https://github.com/proteneer/timemachine/pull/812/commits/81a8ebff566d72cbdeca9ac1a59f119a868cd729

tests/test_single_topology_v3.py
maxentile: Modified all instances of `mcs_threshold = 0.75` to `mcs_threshold = 2.0` in https://github.com/proteneer/timemachine/pull/812/commits/2b91826c31d69235fdd90b5b9c3b31cc82533bf4

tests/test_single_topology_v3.py
maxentile: Modified all instances of `mcs_threshold = 0.75` to `mcs_threshold = 2.0` in https://github.com/proteneer/timemachine/pull/812/commits/2b91826c31d69235fdd90b5b9c3b31cc82533bf4

timemachine/fe/atom_mapping.py
```diff
@@ -7,64 +8,64 @@
...
+    # try on given mols
+    result = rdFMCS.FindMCS([a, b], params)
+
+    # optional fallback
+    def is_trivial(mcs_result) -> bool:
+        return mcs_result.numBonds < 2
+
+    if retry and is_trivial(result) and smarts is None:
+        # try again, but seed with MCS computed without explicit hydrogens
+        a_without_hs = Chem.RemoveHs(deepcopy(a))
```
jkausrelay, line 52: Actually from reading the docs, this always returns a copy anyway (https://www.rdkit.org/docs/source/rdkit.Chem.rdmolops.html)

RETURNS: a new molecule with the Hs removed

NOTES:

The original molecule is not modified.

maxentile: Even so, I think the extra few microseconds it costs to `deepcopy` is worth it to avoid having a reader ever be prompted to confirm this, and I'd be inclined to keep the redundant `deepcopy` here.

timemachine/fe/atom_mapping.py
maxentile: The whitespace is mostly intentional! (Details are indented, but first line summary is not.)

Preference for
```text
AtomMappingError: MCS result trivial!
            timed out: False
            # atoms in MCS: 13
            # bonds in MCS: 12
```

over

```text
AtomMappingError: MCS result trivial!
timed out: False
# atoms in MCS: 13
# bonds in MCS: 12
```
or
```text
AtomMappingError:
MCS result trivial!
timed out: False
# atoms in MCS: 13
# bonds in MCS: 12
```
or similar

---
Might be good to reduce the indent level in the error message, but I'm inclined to leave as-is so that the source is indented readably as well.

examples/relative_free_energy.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/812/commits/627816824c49d8f7d30631741a19b3a4386f33d5

jkausrelay (approved):

PR #813, opened 2022-08-31
Align Mols by Core

badisa:
An issue was seen in finding the MCS of two compounds due to slight conformational changes. This PR adds the ability to set up a set of restraints between two compounds in vacuum to align their cores together. This can be used to construct a valid MCS mapping that is stable for SingleTopologyV3

Test Case
Two cyclohexanes that have different conformations.
Screenshot from 2022-08-31 14-07-44

If mapped using `mcs_map` and `get_core_by_mcs` the hydrogens mapped are invalid due to the relative position to the ring
Screenshot from 2022-08-31 14-10-02
When this core is applied to SingleTopologyV3, the initial conformation is wrong and is an unstable transformation.
Screenshot from 2022-08-31 14-08-16

Aligning to the carbon ring then running `mcs_map` and `get_core_by_mcs`, the mapping is as expected.
Screenshot from 2022-08-31 14-10-14

### DIFF
--- a/tests/test_align.py
+++ b/tests/test_align.py
@@ -1,0 +1,180 @@
+from copy import deepcopy
+
+import numpy as np
+import pytest
+from rdkit import Chem
+from rdkit.Chem import AllChem
+
+from timemachine.constants import DEFAULT_FF
+from timemachine.fe.atom_mapping import get_core_by_mcs, mcs
+from timemachine.fe.topology import AtomMappingError
+from timemachine.fe.utils import set_romol_conf
+from timemachine.ff import Forcefield
+from timemachine.md import align
+
+
+def get_cyclohexanes_different_confs():
+    """Two cyclohexane structures that differ enough in conformations to map poorly by MCS with threshold of 2.0"""
+    mol_a = Chem.MolFromMolBlock(
+        """
+ cyclo_1
+
+ 18 18  0  0  1  0  0  0  0  0999 V2000
+    0.7780    1.1695    0.1292 C   0  0  0  0  0  0  0  0  0  0  0  0
+    1.3871   -0.1008    0.2959 C   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.6896    1.3214   -0.2192 C   0  0  0  0  0  0  0  0  0  0  0  0
+   -1.5088    0.0613    0.0503 C   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.7859   -1.2096   -0.4242 C   0  0  0  0  0  0  0  0  0  0  0  0
+    0.6085   -1.3920    0.2133 C   0  0  0  0  0  0  0  0  0  0  0  0
+   -1.1105    2.1590    0.3356 H   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.7549    1.5841   -1.2762 H   0  0  0  0  0  0  0  0  0  0  0  0
+   -1.6874   -0.0047    1.1175 H   0  0  0  0  0  0  0  0  0  0  0  0
+   -2.4858    0.1560   -0.4244 H   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.6117   -1.0357   -1.4891 H   0  0  0  0  0  0  0  0  0  0  0  0
+    1.1610   -2.0015   -0.5036 H   0  0  0  0  0  0  0  0  0  0  0  0
+    0.5422   -1.8809    1.1852 H   0  0  0  0  0  0  0  0  0  0  0  0
+   -1.4054   -2.0928   -0.2686 H   0  0  0  0  0  0  0  0  0  0  0  0
+    1.3677    1.7499   -0.5802 H   0  0  0  0  0  0  0  0  0  0  0  0
+    0.9940    1.7789    1.0067 H   0  0  0  0  0  0  0  0  0  0  0  0
+    1.9567   -0.0955    1.2253 H   0  0  0  0  0  0  0  0  0  0  0  0
+    2.2449   -0.1670   -0.3734 H   0  0  0  0  0  0  0  0  0  0  0  0
+  1  2  1  0
+  1  3  1  0
+  1 15  1  0
+  1 16  1  0
+  2  6  1  0
+  2 17  1  0
+  2 18  1  0
+  3  4  1  0
+  3  7  1  0
+  3  8  1  0
+  4  5  1  0
+  4  9  1  0
+  4 10  1  0
+  5  6  1  0
+  5 11  1  0
+  5 14  1  0
+  6 12  1  0
+  6 13  1  0
+M  END
+$$$$""",
+        removeHs=False,
+    )
+
+    mol_b = Chem.MolFromMolBlock(
+        """
+ cyclo_2
+
+ 18 18  0  0  1  0  0  0  0  0999 V2000
+    0.7953    1.1614    0.0469 C   0  0  0  0  0  0  0  0  0  0  0  0
+    1.3031   -0.0613    0.5362 C   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.6118    1.1962   -0.5144 C   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.9934   -0.1785   -1.1042 C   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.6439   -1.3144   -0.1494 C   0  0  0  0  0  0  0  0  0  0  0  0
+    0.4262   -1.2251    0.6719 C   0  0  0  0  0  0  0  0  0  0  0  0
+   -1.2949    1.4641    0.2937 H   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.6968    1.9715   -1.2775 H   0  0  0  0  0  0  0  0  0  0  0  0
+   -2.0662   -0.1837   -1.3042 H   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.4970   -0.3613   -2.0575 H   0  0  0  0  0  0  0  0  0  0  0  0
+    0.6428   -1.9811    1.4121 H   0  0  0  0  0  0  0  0  0  0  0  0
+   -1.2637   -2.1987   -0.1345 H   0  0  0  0  0  0  0  0  0  0  0  0
+    1.4850    1.5611   -0.6965 H   0  0  0  0  0  0  0  0  0  0  0  0
+    0.8877    1.9212    0.8230 H   0  0  0  0  0  0  0  0  0  0  0  0
+    1.8010    0.1189    1.4889 H   0  0  0  0  0  0  0  0  0  0  0  0
+    2.1753   -0.3430   -0.0537 H   0  0  0  0  0  0  0  0  0  0  0  0
+   -1.2711   -0.8618    0.6186 H   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.1781   -0.6848    1.4006 H   0  0  0  0  0  0  0  0  0  0  0  0
+  1  2  1  0
+  1  3  1  0
+  1 13  1  0
+  1 14  1  0
+  2  6  1  0
+  2 15  1  0
+  2 16  1  0
+  3  4  1  0
+  3  7  1  0
+  3  8  1  0
+  4  5  1  0
+  4  9  1  0
+  4 10  1  0
+  5  6  1  0
+  5 12  1  0
+  5 17  1  0
+  6 11  1  0
+  6 18  1  0
+M  END
+$$$$""",
+        removeHs=False,
+    )
+    return mol_a, mol_b
+
+
+def get_core(mol_a, mol_b, threshold=2.0):
+    """Simple utility that finds a core by using the conformers with a threshold"""
+    mcs_result = mcs(mol_a, mol_b, threshold=threshold)
+    query = Chem.MolFromSmarts(mcs_result.smartsString)
+    return get_core_by_mcs(mol_a, mol_b, query, threshold=threshold)
+
+
+@pytest.mark.nogpu
+def test_align_mols_by_core():
+    """Uses a core to align two compounds. Uses two cyclohexanes with different conformers as the test case in the case of
+    finding the best MCS"""
+
+    mol_a, mol_b = get_cyclohexanes_different_confs()
+
+    assert mol_a.GetNumAtoms() == mol_b.GetNumAtoms()
+
+    core = get_core(mol_a, mol_b)
+
+    # The difference in conformer prevents a complete mapping
+    assert len(core) != mol_a.GetNumAtoms()
+
+    hydrogenless_mol_a = Chem.RemoveHs(deepcopy(mol_a))
+    hydrogenless_mol_b = Chem.RemoveHs(deepcopy(mol_b))
+
+    restraint_core = get_core(hydrogenless_mol_a, hydrogenless_mol_b)
+
+    assert len(restraint_core) == 6  # Number of carbons in mol
+
+    ff = Forcefield.load_from_file(DEFAULT_FF)
+
+    conf_a, conf_b = align.align_mols_by_core(mol_a, mol_b, restraint_core, ff)
+
+    set_romol_conf(mol_a, conf_a)
+    set_romol_conf(mol_b, conf_b)
+
+    core = get_core(mol_a, mol_b)
+
+    # The mapping should cover the entirety of the two compounds
+    assert len(core) == mol_a.GetNumAtoms()
+
+
+def test_align_core_different_size_mols():
+    """Verifies that this works as expected if the size of the molecules aren't the same."""
+    mol_a = Chem.AddHs(Chem.MolFromSmiles("c1ccccc1"))  # benzene
+    mol_b = Chem.AddHs(Chem.MolFromSmiles("c1ccccc1-c2ccccc2"))  # bi-phyenl
+    seed = 2022
+    AllChem.EmbedMolecule(mol_a, randomSeed=seed)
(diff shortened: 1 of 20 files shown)


timemachine/md/minimizer.py
```diff
@@ -277,3 +282,85 @@ def equilibrate_host(
     ctxt.multiple_steps(np.linspace(0.0, 0.0, n_steps))
 
     return ctxt.get_x_t(), ctxt.get_box()
+
+
+def align_mols_by_core(
```
badisa, line 287: Should probably move to a `md/align.py` module

timemachine/fe/topology.py
```diff
@@ -646,6 +646,92 @@ def parameterize_nonbonded(self, ff_q_params, ff_lj_params):
             cutoff,
         )
 
+    def parameterize_nonbonded_pairlist(self, ff_q_params, ff_lj_params):
```
badisa, line 649: Needed to use DualTopology with VacuumSystem, was also broken previously.

timemachine/fe/topology.py
```diff
@@ -629,8 +629,8 @@ def parameterize_nonbonded(self, ff_q_params, ff_lj_params):
             ]
         ).astype(np.float64)
 
-        combined_lambda_plane_idxs = None
-        combined_lambda_offset_idxs = None
+        combined_lambda_plane_idxs = np.zeros(NA + NB, dtype=np.int32)
```
badisa, line 632: The potential is effectively broken without this. Because we only used it parameterize other nonbonded potentials we hadn't seen an error.

timemachine/potentials/nonbonded.py
```diff
@@ -192,6 +192,10 @@ def nonbonded_v3(
         assert (lj_rescale_mask == lj_rescale_mask.T).all()
 
     N = conf.shape[0]
+    if lambda_plane_idxs is None:
```
badisa, line 195: The docstring says these are optional, if the conf is 3D, they are not.

timemachine/md/align.py
```diff
@@ -0,0 +1,96 @@
...
+
+    grad_fn = jax.jit(jax.grad(combined_U, argnums=(0)))
+
+    def force_func(x):
+        du_dx = grad_fn(x)
+        return -du_dx
+
+    # TBD 1e-3 step size is arbitrary
+    intg = LangevinIntegrator(force_func, masses, 0.0, 1e-3, 0.0)
+    key = jax.random.PRNGKey(2022)  # seed value doesn't matter with no friction and 0 temp
```
maxentile, line 84: Possibility: Could [retain 0 temp but set friction > 0](https://github.com/proteneer/timemachine/blob/05d60b173f92ed18b8c63fbf05cd3249a015711e/tests/test_reference_langevin_integrator.py#L152-L158) to get more minimizer-like behavior.

(With friction=temp=0, this may not effectively minimize energy, I think)

timemachine/md/align.py
```diff
@@ -0,0 +1,96 @@
...
+    xs, _ = intg.multiple_steps_lax(key, x0, np.zeros_like(x0), n_steps=n_steps)
+    xs = np.array(xs)
+
+    mol_a_core = xs[:, bond_indices[:, 0]]
+    mol_b_core = xs[:, bond_indices[:, 1]]
+    deltas = mol_a_core - mol_b_core
+    rmsds = [np.linalg.norm(delta) for delta in deltas]
+    best_alignment_idx = np.argmin(rmsds)
```
maxentile, line 92: Question: Is taking `idx = argmin(rmsds)` preferable to taking last frame `idx = -1`?

With default restraint force constant k = 10000 kJ/mol / nm^2, I would expect very small restraint RMSDs here (if these are fluctuating @ 300K and there are no other competing energy terms, each restrained atom pair might fluctuate roughly on order of stddev = sqrt(kBT/ k) = 0.015 nm).

Another possibility may be to gradually increase the strength of the core restraint over time and take the last frame, if driving RMSD close to 0 is crucial.

maxentile (commented):
Nice approach! This should reduce sensitivity of distance-threshold-based MCS to imperfectly aligned initial conformers.

(Minor nit inline about how the minimization is performed.)

timemachine/md/align.py
badisa: Doing `idx = argmin(rmsds)` definitely has an impact. Have seen selecting the last frame produce cores just as bad the current core finding approach we use. It can also explode if the initial core is broken, so this will avoid taking a frame that is outright bad, though could have a separate check outside of `rmsd`. 

> Another possibility may be to gradually increase the strength of the core restraint over time and take the last frame, if driving RMSD close to 0 is crucial.

That certainly seems viable.

timemachine/md/align.py
```diff
@@ -0,0 +1,98 @@
...
+
+from timemachine.fe import topology
+from timemachine.fe.utils import get_mol_masses, get_romol_conf
+from timemachine.ff import Forcefield
+from timemachine.integrator import LangevinIntegrator
+from timemachine.potentials import generic
+
+
+def align_mols_by_core(
+    mol_a: Chem.Mol, mol_b: Chem.Mol, core: NDArray, ff: Forcefield, n_steps: int = 200, k: float = 10000
```
badisa, line 18: Increased to 200 with the increase of friction to 10.0, otherwise some test cases that passed before fail.

timemachine/md/align.py
badisa: Resolved in https://github.com/proteneer/timemachine/pull/813/commits/eba33b2cc34690fd53adc6a7399301203d7f3518

timemachine/md/align.py
badisa: I am inclined to hold off with the increasing the strength, as it seems like we need to deboggle the integrator and how we configure JAX vacuum systems. Using the GPU systems would work, but this seems too cheap to justify pushing onto the GPU.

badisa:
Been running this on CDK2, Hif2a, SHP2 from here: https://github.com/MCompChem/fep-benchmark/

So far things are looking pretty good. Takes about 15 seconds per pair at the moment (32 core machine, but looks to only be about 2 cores really doing work)

timemachine/fe/atom_mapping.py
```diff
@@ -2,10 +2,17 @@
...
 from rdkit import Chem
 from rdkit.Chem import rdFMCS
 
+from timemachine.constants import DEFAULT_FF
 from timemachine.fe.topology import AtomMappingError
+from timemachine.fe.utils import set_romol_conf
+from timemachine.ff import Forcefield
+from timemachine.md.align import align_mols_by_core
+
+# def heavy_atom_mcs(a, b, threshold)
```
jkausrelay, line 15: remove

timemachine/md/align.py
```diff
@@ -0,0 +1,98 @@
...
+) -> Tuple[NDArray, NDArray]:
+    """
+    Given two mols and a core mapping, simulate the mols in vacuum with harmonic bounds between them to get aligned poses.
```
jkausrelay, line 21: bounds -> bonds

badisa: Fixed in https://github.com/proteneer/timemachine/pull/813/commits/c45d947dbfd2941220a5e5959ed9dd3c1b623cc3

timemachine/fe/atom_mapping.py
badisa: Removed in https://github.com/proteneer/timemachine/pull/813/commits/c45d947dbfd2941220a5e5959ed9dd3c1b623cc3

timemachine/md/align.py
mcwitt: > as it seems like we need to deboggle the integrator and how we configure JAX vacuum systems

Not sure if this is what needs deboggling, but just a note that now you can convert GPU potentials to JAX reference potentials without too much trouble:

```
from timemachine.potentials import generic

generic.Nonbonded.from_gpu(nonbonded_gpu).to_reference()
```

tests/test_fe_utils.py
```diff
@@ -103,3 +104,22 @@ def test_get_mol_name():
...
+
+
+def test_set_mol_coords():
+    mol = Chem.AddHs(Chem.MolFromSmiles("c1ccccc1"))
+    AllChem.EmbedMolecule(mol)
+
+    x0 = utils.get_romol_conf(mol)
+
+    # Make some random move
+    x1 = x0 + np.random.randn(*x0.shape)
```
mcwitt, line 116: Should we set a seed above?

timemachine/md/align.py
badisa: For sure, those would certainly help getting a reference version ready. I like having this a CPU only method and its the JAX potentials and the integrator that would need to be modified to interpolate.

Unless we want to do something like this https://github.com/proteneer/timemachine/blob/master/timemachine/fe/single_topology_v3.py#L656-L679 and change lambda in a loop?

timemachine/fe/utils.py
```diff
@@ -148,6 +149,16 @@ def get_romol_conf(mol):
     return guest_conf / 10  # from angstroms to nm
 
 
+def set_romol_conf(mol, new_coords):
+    """Sets coordinates of mol's 0th confomrmer"""
+    assert new_coords.shape[0] == mol.GetNumAtoms()
+    # convert from nm to angstroms
+    angstrom_coords = new_coords * 10
```
mcwitt, line 156: Should we indicate in the function name and/or the docstring that this also performs unit conversion?

timemachine/fe/utils.py
```diff
@@ -148,6 +149,16 @@ def get_romol_conf(mol):
...
+    conf = mol.GetConformer(0)
+    for i, (x, y, z) in enumerate(angstrom_coords):
+        conf.SetAtomPosition(i, Point3D(float(x), float(y), float(z)))
```
mcwitt, line 159: nit: is there a downside to this slightly simpler version?
```suggestion
    for i, pos in enumerate(angstrom_coords):
        conf.SetAtomPosition(i, pos)
```

timemachine/md/align.py
mcwitt: Hmm, I think I understand. Could you modify the force function signature to take `k` as an argument, and then step via a for loop while modifying k? Maybe something like

```python
    @jax.jit
    def force_func(x, k):
        restraint_params = jnp.zeros((len(core), 2)).at[:, 0].set(k)

        bond_U = functools.partial(
            U_ref,
            params=jnp.array(restraint_params),
            box=None,
            lam=0.0,
        )

        def combined_U(x):
            return U(x) + bond_U(x)

        dU_dx = jax.grad(combined_U)(x)

        return -du_dx
```

timemachine/fe/topology.py
```diff
@@ -646,6 +646,92 @@ def parameterize_nonbonded(self, ff_q_params, ff_lj_params):
...
+            params.append((q * sf_q, s, e * sf_lj))
+
+        params = np.array(params)
+
+        beta = _BETA
+        cutoff = _CUTOFF  # solve for this analytically later
+
+        offsets = np.zeros(len(inclusion_idxs))
+
+        return params, potentials.NonbondedPairListPrecomputed(inclusion_idxs, offsets, beta, cutoff)
```
mcwitt, line 733: nit: it seems like this function might factorize into something like
```
params_a, nb_a = BaseTopology(mol_a, ff).parameterize_nonbonded_pairlist(ff_q_params, ff_lj_params)
params_b, nb_b = BaseTopology(mol_b, ff).parameterize_nonbonded_pairlist(ff_q_params, ff_lj_params)
params, nonbonded = combine_(params_a, nb_a, params_b, nb_b)
```

(with `combine_` to be defined) and this might reduce the repetition here relative to `BaseTopology`?

timemachine/fe/utils.py
badisa: The API requires a Point3D object and each value must be in a `float` 

https://www.rdkit.org/docs/source/rdkit.Chem.rdchem.html#rdkit.Chem.rdchem.Conformer.SetAtomPosition

Unless you are indicating a larger change?

mcwitt: No larger change; the Python API seems to also accept iterables of objects that can be converted to float (the docs are not clear on this, but do seem to show a variant that takes `AtomPairsParameters` 😕 ). In any case, there are many examples of passing e.g. lists and tuples as the second argument in the [rdkit repo](https://github.com/rdkit/rdkit/search?q=SetAtomPosition), so either is probably fine?

timemachine/md/align.py
badisa: That would work, just then need to have a way to modify k as it moves forward. Could recreate a new force func every step along with and integrator, but seems like we should support this more generally rather than hard coding it here.

timemachine/fe/topology.py
badisa: Done in [e0e77cc](https://github.com/proteneer/timemachine/pull/813/commits/e0e77cc7c079ffc5b176201af4bfb4170514b483)

timemachine/fe/utils.py
badisa: Done in [5ab4c26](https://github.com/proteneer/timemachine/pull/813/commits/5ab4c26e0267f6ba02fee9a8f05a3764deb3c81f)

timemachine/fe/atom_mapping.py
```diff
@@ -119,3 +124,64 @@ def get_core_by_mcs(mol_a, mol_b, query, threshold=0.5):
...
+        return core
+
+    heavy_atom_core = setup_core(a_without_hs, b_without_hs)
+
+    conf_a, conf_b = align_mols_by_core(mol_a, mol_b, heavy_atom_core, ff, n_steps=n_steps, k=k)
+    set_romol_conf(a_copy, conf_a)
+    set_romol_conf(b_copy, conf_b)
+
+    core = setup_core(a_copy, b_copy)
+    return core
```
jkausrelay, line 187: It's useful for debugging to return the smarts string too

timemachine/md/align.py
mcwitt: > just then need to have a way to modify k as it moves forward. Could recreate a new force func every step along with and integrator

I'm not sure I understand. In the example the force function takes `k` as a parameter, so it seems like we shouldn't need to create a new force function on every step?

tests/test_align.py
```diff
@@ -0,0 +1,180 @@
...
+
+    conf_a, conf_b = align.align_mols_by_core(mol_a, mol_b, core, ff)
+
+    set_romol_conf(mol_a, conf_a)
+    set_romol_conf(mol_b, conf_b)
+
+    core = get_core(mol_a, mol_b)
+
+    # The mapping should cover the entirety of the benzene minus the hydrogen covering the carbon linker
+    assert len(core) == mol_a.GetNumAtoms() - 1
```
maxentile, line 180: Testing coverage: Can we add a test of `atom_mapping.get_core_with_alignment` to `test_align.py` or `test_atom_mapping.py`?

(As of https://github.com/proteneer/timemachine/commit/68690b35b7e773a3d11f3738c5015638be2187f9 , getting a`ValueError: cannot extract desired type from sequence`  at [L183](https://github.com/proteneer/timemachine/pull/813/files#diff-79ea6707b80f02c8fa97a1b42d9e52116bd31149cce41220eefbb0ca7a56191cR183) 
, even though `align_mols_by_core` and `set_romol_conf` look good individually)

timemachine/md/align.py
maxentile: On further thought -- is there a strong reason to prefer using LangevinIntegrator settings to emulate a minimizer, instead of using e.g. `scipy.optimize.minimize`?

May be preferable since no settings need to be potentially tuned for stability / performance (no stepsize, masses, or damping parameters). (Would expect to be slightly cheaper too -- probably converging in < 100 gradient evaluations.)

badisa: Don't think there is a very strong case. I would hesitate to use `scipy.optimize.minimize` due to the issues we saw with trying to minimize using scipy for `minimize_host_4d`, though that is a more complicated case admittedly. Would be comfortable using the fire minimizer though as that didn't exhibit the issues?

tests/test_align.py
badisa: Where do you see that error? Testing with the cyclohexane case doesn't seem to raise the error locally.

timemachine/fe/atom_mapping.py
badisa: Done in [4e0b6e3](https://github.com/proteneer/timemachine/pull/813/commits/4e0b6e3ae5aa5b4df619e835571c541cbc972436)

tests/test_align.py
badisa: Fixed here: [6cce4b0](https://github.com/proteneer/timemachine/pull/813/commits/6cce4b0e5132d23081fe2e3e62a684c1ed141804)

The test fails without the cast to float64 in `set_romol_conf`

tests/test_fe_utils.py
badisa: Done in [4d77204](https://github.com/proteneer/timemachine/pull/813/commits/4d77204bd5168b5dfb06c163774094b2a851139a)

mcwitt (approved):

timemachine/md/align.py
maxentile: Hmm, fair enough -- would be in favor of tabling this for now.

I don't remember the details of the case where FIRE appears more robust than L-BFGS, or if any causes are in common with this application... Switching optimizer maybe something to consider benchmarking in a follow-up, since this works as-is and the application doesn't seem to be very sensitive to whether the alignment dynamics is energy-minimizing... 

(Main benefit of the suggestion would be to _reduce_ the amount of investigation required to tune parameters / confirm stability etc. -- would defeat the purpose if this instead required new investigation!)

jkausrelay (approved):

maxentile (approved):

PR #819, opened 2022-09-08
Improve `get_core_by_mcs`

maxentile:
Handle more cases where highly symmetric MCS results cause AtomMappingErrors

Increase size-limit on exhaustive search
* max_matches 1000 -> 10_000
* warn if max_matches exceeded

Performance optimizations
* Move some stuff out of double for loop
* Use cheaper np.any reduction before np.sum reduction

### DIFF
--- a/timemachine/fe/atom_mapping.py
+++ b/timemachine/fe/atom_mapping.py
@@ -5,6 +5,7 @@ import numpy as np
 from numpy.typing import NDArray
 from rdkit import Chem
 from rdkit.Chem import rdFMCS
+from scipy.spatial.distance import cdist
 
 from timemachine.constants import DEFAULT_FF
 from timemachine.fe.topology import AtomMappingError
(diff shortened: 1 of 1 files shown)


timemachine/fe/atom_mapping.py
```diff
@@ -94,24 +95,39 @@ def get_core_by_mcs(mol_a, mol_b, query, threshold=0.5):
...
     """
 
     # fetch conformer, assumed aligned
     conf_a = mol_a.GetConformer(0).GetPositions()
     conf_b = mol_b.GetConformer(0).GetPositions()
 
     # note that >1 match possible here -- must pick minimum-cost match
-    matches_a = mol_a.GetSubstructMatches(query, uniquify=False)
-    matches_b = mol_b.GetSubstructMatches(query, uniquify=False)
+    max_matches = 10_000
```
bp-kelley, line 107: Are you matching with or without hydrogens?  I will probably be cheaper to match without hydrogens and then match the hydrogens after the fact.

maxentile: Attempted, and I think your suggestion would be faster, but I don't know how to implement the "match the hydrogens after the fact" step without making this function significantly more complicated or possibly returning matches that could be inconsistent with `query`.

jkausrelay (approved):

maxentile:
@proteneer approved offline

PR #820, opened 2022-09-08
Improve protocol for ring opening/closing transformations

mcwitt:
- Changes the functional form used for interpolation of harmonic bond and angle force constants in single topology transformations from linear to log-linear
   - This is motivated by the observation that the optimal transformation between gaussians is approximately log-linear in sigma in the limit that initial and final mu are identical ([theory reference](http://www.stat.columbia.edu/~gelman/research/published/path2.pdf), [numerical evidence (cell 75)](https://gitlab.com/relaytx/users/jfass/notebooks/-/blob/master/protocol_optimization/RingClosingTwoWays.ipynb)).

- For bonds, angles and torsions In the case where `src_k=0` or `dst_k=0` (e.g. ring opening/closing), sequences force constant transformations so that force constants "turn on" in the order (bonds, angles, torsions).
   - This is to prevent numerical instability seen when an angle with nonzero k spans a bond with near-zero k (due to potential colocation of atoms), or when a torsion with nonzero k spans an angle with near-zero k (due to potential collinearity of bonds).

### hif2a vacuum single topology test

- 32 windows
- 750 ps / window
- 80 edges (from `results_edges_20ns.csv`)

#### Before
Simultaneous linear interpolation for bonds and angle force constants
image

#### After
Log-linear interpolation for bond, angle force constants; sequenced (bonds, angles, torsions)
image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -100,7 +100,7 @@ setup(
             "mypy==0.942",
             "pre-commit==2.17.0",
         ],
-        "test": ["pytest", "pytest-cov", "hilbertcurve==1.0.5", "hypothesis==6.54.6"],
+        "test": ["pytest", "pytest-cov", "hilbertcurve==1.0.5"],
     },
     package_data={
         "timemachine": [
--- a/setup.py
+++ b/setup.py
@@ -100,7 +100,7 @@ setup(
             "mypy==0.942",
             "pre-commit==2.17.0",
         ],
-        "test": ["pytest", "pytest-cov", "hilbertcurve==1.0.5"],
+        "test": ["pytest", "pytest-cov", "hilbertcurve==1.0.5", "hypothesis==6.54.6"],
     },
     package_data={
         "timemachine": [
--- a/tests/test_single_topology_v3.py
+++ b/tests/test_single_topology_v3.py
@@ -455,7 +455,7 @@ def test_interpolate_harmonic_force_constant(src_k, dst_k, k_min, lambda_interva
 
     lambdas = np.arange(0.01, 1.0, 0.01)
 
-    # all interpolated values less than k_min
+    # all interpolated values >= k_min
     np.testing.assert_array_less(1.0, f(src_k, dst_k, lambdas) / k_min + 1e-9)
 
     def assert_nondecreasing(f):
--- a/tests/test_single_topology_v3.py
+++ b/tests/test_single_topology_v3.py
@@ -14,6 +14,8 @@ from importlib import resources
 import jax.numpy as jnp
 import numpy as np
 import pytest
+from hypothesis import given
+from hypothesis.strategies import floats
 from jax import vmap
 from jax.experimental.checkify import checkify
 from rdkit import Chem
@@ -494,8 +496,25 @@ def test_cyclic_difference():
     assert cyclic_difference(0, 0, 3) == 0
     assert cyclic_difference(0, 1, 3) == 1
     assert cyclic_difference(0, 2, 3) == -1
-
     _ = jax.jit(cyclic_difference)(0, 1, 1)
+
+
+finite_floats = functools.partial(floats, allow_subnormal=False, allow_nan=False, allow_infinity=False)
+
+
+@given(finite_floats(-1e6, 1e6), finite_floats(-1e6, 1e6), finite_floats(min_value=0.0, exclude_min=True))
+def test_cyclic_difference_prop(a, b, period):
+    cdiff = cyclic_difference(a, b, period)
+
+    def assert_allclose_cyclic(a, b, **kwargs):
+        def f(x):
+            x_mod = x % period
+            return np.minimum(x_mod, period - x_mod)
+
+        return np.testing.assert_allclose(f(a), f(b), **kwargs)
+
+    assert_allclose_cyclic(a + cdiff, b, atol=1e-10)
+    assert_allclose_cyclic(np.abs(cdiff), np.abs(a - b), atol=1e-12)
 
 
 def assert_linear(f, x1, x2, x3):
--- a/tests/test_single_topology_v3.py
+++ b/tests/test_single_topology_v3.py
@@ -421,6 +421,16 @@ def test_handle_ring_opening_closing_symmetric(k, lambda_interval, lam):
     )
 
 
+@given(nonzero_force_constants, st.lists(lambdas, min_size=3, max_size=3, unique=True).map(sorted))
+@seed(2022)
+def test_handle_ring_opening_closing_pin_to_end_states(k, lambdas):
+    lam, lambda_min, lambda_max = lambdas
+    assert handle_ring_opening_closing(linear_interpolation, 0.0, k, lam, lambda_min, lambda_max) == 0.0
+
+    lambda_min, lambda_max, lam = lambdas
+    assert handle_ring_opening_closing(linear_interpolation, 0.0, k, lam, lambda_min, lambda_max) == k
+
+
 @given(
     nonzero_force_constants,
     nonzero_force_constants,
@@ -517,6 +527,16 @@ def test_cyclic_difference_inverse(a, b, period):
 @seed(2022)
 def test_cyclic_difference_antisymmetric(a, b, period):
     assert cyclic_difference(a, b, period) + cyclic_difference(b, a, period) == 0
+
+
+@given(bounded_ints, bounded_ints, bounded_ints, bounded_ints, periods)
+@seed(2022)
+def test_cyclic_difference_shift_by_n_periods(a, b, m, n, period):
+    assert_equal_cyclic(
+        cyclic_difference(a + m * period, b + n * period, period),
+        cyclic_difference(a, b, period),
+        period,
+    )
 
 
 @given(bounded_ints, bounded_ints, bounded_ints, periods)
--- a/tests/test_single_topology_v3.py
+++ b/tests/test_single_topology_v3.py
@@ -492,6 +492,16 @@ def test_cyclic_difference():
     assert cyclic_difference(0, 0, 3) == 0
     assert cyclic_difference(0, 1, 3) == 1
     assert cyclic_difference(0, 2, 3) == -1
+
+    # antisymmetric
+    assert cyclic_difference(0, 1, 3) == -cyclic_difference(1, 0, 3)
+    assert cyclic_difference(0, 2, 3) == -cyclic_difference(2, 0, 3)
+
+    # translation invariant
+    assert cyclic_difference(0, 1, 3) == cyclic_difference(-1, 0, 3)
+    assert cyclic_difference(0, 4, 8) == cyclic_difference(-2, 2, 8) == cyclic_difference(-4, 0, 8)
+
+    # jittable
     _ = jax.jit(cyclic_difference)(0, 1, 1)
 
 
--- a/tests/test_single_topology_v3.py
+++ b/tests/test_single_topology_v3.py
@@ -14,8 +14,6 @@ from importlib import resources
 import jax.numpy as jnp
 import numpy as np
 import pytest
-from hypothesis import given
-from hypothesis.strategies import floats
 from jax import vmap
 from jax.experimental.checkify import checkify
 from rdkit import Chem
@@ -497,24 +495,6 @@ def test_cyclic_difference():
     assert cyclic_difference(0, 1, 3) == 1
     assert cyclic_difference(0, 2, 3) == -1
     _ = jax.jit(cyclic_difference)(0, 1, 1)
-
-
-finite_floats = functools.partial(floats, allow_subnormal=False, allow_nan=False, allow_infinity=False)
-
-
-@given(finite_floats(-1e6, 1e6), finite_floats(-1e6, 1e6), finite_floats(min_value=0.0, exclude_min=True))
-def test_cyclic_difference_prop(a, b, period):
-    cdiff = cyclic_difference(a, b, period)
-
-    def assert_allclose_cyclic(a, b, **kwargs):
-        def f(x):
-            x_mod = x % period
-            return np.minimum(x_mod, period - x_mod)
-
-        return np.testing.assert_allclose(f(a), f(b), **kwargs)
-
-    assert_allclose_cyclic(a + cdiff, b, atol=1e-10)
-    assert_allclose_cyclic(np.abs(cdiff), np.abs(a - b), atol=1e-12)
 
 
 def assert_linear(f, x1, x2, x3):
--- a/tests/test_single_topology_v3.py
+++ b/tests/test_single_topology_v3.py
@@ -399,10 +399,10 @@ nonzero_force_constants = finite_floats(1e-9, 1e9)
 
 lambdas = finite_floats(0.0, 1.0)
 
-open_lambda_intervals = st.lists(finite_floats(1e-9, 1.0 - 1e-9), min_size=2, max_size=2, unique=True).map(sorted)
+lambda_intervals = st.lists(finite_floats(1e-9, 1.0 - 1e-9), min_size=2, max_size=2, unique=True).map(sorted)
 
 
-@given(nonzero_force_constants, open_lambda_intervals, lambdas)
+@given(nonzero_force_constants, lambda_intervals, lambdas)
 def test_handle_ring_opening_closing_symmetry(k, lambda_interval, lam):
     lambda_min, lambda_max = lambda_interval
 
@@ -424,7 +424,7 @@ def test_handle_ring_opening_closing_symmetry(k, lambda_interval, lam):
     nonzero_force_constants,
     nonzero_force_constants,
     nonzero_force_constants,
-    open_lambda_intervals,
+    lambda_intervals,
 )
 def test_interpolate_harmonic_force_constant(src_k, dst_k, k_min, lambda_interval):
     lambda_min, lambda_max = lambda_interval
--- a/tests/test_single_topology_v3.py
+++ b/tests/test_single_topology_v3.py
@@ -522,6 +522,7 @@ bounded_ints = st.integers(-int(1e9), int(1e9))
 @seed(2022)
 def test_cyclic_difference_inverse(a, b, period):
     x = cyclic_difference(a, b, period)
+    assert np.abs(x) <= period / 2
     assert_equal_cyclic(a + x, b, period)
 
 
--- a/tests/test_single_topology_v3.py
+++ b/tests/test_single_topology_v3.py
@@ -15,6 +15,7 @@ import jax.numpy as jnp
 import numpy as np
 import pytest
 from jax import vmap
+from jax.experimental.checkify import checkify
 from rdkit import Chem
 from rdkit.Chem import AllChem
 
@@ -27,11 +28,10 @@ from timemachine.fe.single_topology_v3 import (
     MultipleAnchorWarning,
     SingleTopologyV3,
     canonicalize_improper_idxs,
+    cyclic_difference,
     handle_ring_opening_closing,
-    interpolate_harmonic_force,
-    interpolate_periodic_torsion,
+    interpolate_harmonic_force_constant,
     setup_dummy_interactions_from_ff,
-    standardize_angle_radians,
 )
 from timemachine.fe.system import convert_bps_into_system, minimize_scipy, simulate_system
 from timemachine.fe.utils import get_mol_name, get_romol_conf
@@ -326,7 +326,7 @@ def test_jax_transform_intermediate_potential():
     def U(x, lam):
         return st.setup_intermediate_state(lam).get_U_fn()(x)
 
-    _ = jax.jit(U)(conf, 0.1)
+    _ = jax.jit(checkify(U))(conf, 0.1)
 
     confs = jnp.array([conf for _ in range(10)])
     lambdas = jnp.linspace(0, 1, 10)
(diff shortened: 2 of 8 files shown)


tests/test_single_topology_v3.py
```diff
@@ -340,3 +347,109 @@ def test_combine_with_host():
...
+    # check that force constants are interpolated linearly in the intermediate region
+    dst_params = random_params()
+    assert_linear(lambda x: f(src_params, dst_params, x)[0], lambda_min, 0.5, 1 - lambda_min)
+
+    # check that k(lambda) = 0 when 1 - lambda_min < lambda for ring opening
+    opening_idxs = rng.choice(n, 10)
+    dst_params = random_params()
+    dst_params[0, opening_idxs] = 0.0
+    assert np.all(f(random_params(), dst_params, 1 - dlam)[0, opening_idxs] == 0.0)
+
```
mcwitt, line 449: TODO: Add test for simultaneous ring opening and closing

timemachine/fe/single_topology_v3.py
```diff
@@ -483,6 +484,103 @@ def find_dummy_groups_and_anchors(mol_a, mol_b, core_a, core_b):
...
+        jnp.array([k0, x0]),
+        jnp.where(
+            lamb == 1.0,
+            jnp.array([k1, x1]),
+            f(jnp.maximum(k0, k_min), jnp.maximum(k1, k_min)),
+        ),
+    )
+
+
+def standardize_angle_radians(x):
```
mcwitt, line 532: TODO: investigate period-aware interpolation

timemachine/fe/interpolate.py
```diff
@@ -148,4 +148,29 @@ def linear_interpolation(src_params, dst_params, lamb):
...
+        linear_interpolation(
+            jnp.log(jnp.maximum(src_params, min_value)),
+            jnp.log(jnp.maximum(dst_params, min_value)),
+            lamb,
+        )
+    )
+
+
+def pad(f, src_params, dst_params, lamb, lambda_min, lambda_max):
+    return jnp.where(
```
mcwitt, line 168: TODO: docstring

setup.py
```diff
@@ -100,7 +100,7 @@ def build_extension(self, ext):
             "mypy==0.942",
             "pre-commit==2.17.0",
         ],
-        "test": ["pytest", "pytest-cov", "hilbertcurve==1.0.5"],
+        "test": ["pytest", "pytest-cov", "hilbertcurve==1.0.5", "hypothesis==6.54.6"],
```
mcwitt, line 103: Note: not essential, but I found property testing useful for getting the `cyclic_difference` function correct. Open to removing

timemachine/fe/single_topology_v3.py
```diff
@@ -483,6 +484,211 @@ def find_dummy_groups_and_anchors(mol_a, mol_b, core_a, core_b):
...
+
+    with minimum abs(d).
+    """
+
+    d = jnp.fmod(b - a, period)
+
+    def f(d):
+        return jnp.where(d <= period / 2, d, d - period)
+
+    return jnp.where(0 < d, f(d), -f(-d))
```
mcwitt, line 609: It feels like there might be a simpler formula / builtin function for this definition of cyclic difference, but I wasn't able to figure it out yet.

timemachine/fe/interpolate.py
```diff
@@ -148,4 +148,33 @@ def linear_interpolation(src_params, dst_params, lamb):
...
+    """
+    Linear interpolation in log space
+    """
+    return jnp.exp(
+        linear_interpolation(
+            jnp.log(jnp.maximum(src_params, min_value)),
+            jnp.log(jnp.maximum(dst_params, min_value)),
+            lamb,
+        )
+    )
```
maxentile, line 164: 1. Clarification: if `src_params` and `dst_params` < `min_val`, does this not return `src_params` @ `lamb=0`, `dst_params` @ `lamb=1`?

2. Style opinion: preference for simpler compound statements 
```suggestion
    src = jnp.maximum(src_params, min_value)
    dst = jnp.maximum(dst_params, min_value)
    return jnp.exp(linear_interpolation(jnp.log(src), jnp.log(dst), lamb))
```

maxentile: Ahh, I see -- it's safe if combined with `pad`, which pins to end-state values outside of `[lambda_min, lambda_max]`

timemachine/fe/single_topology_v3.py
```diff
@@ -483,6 +484,211 @@ def find_dummy_groups_and_anchors(mol_a, mol_b, core_a, core_b):
     return dummy_groups_b, all_jks
 
 
+def handle_ring_opening_closing(
+    f: Callable[[float, float, float], float],
```
maxentile, line 488: Preference: type alias `InterpolationFxn`, more suggestive name `interpolate_fxn`, or docstring line indicating parameters of `f(src, dst, lamb) -> interpolated`

timemachine/fe/single_topology_v3.py
maxentile, line 609: Nice! This should be more efficient in the worst case than the previous approaches of linearly interpolating raw angles, or linearly interpolating angles mapped to a fixed interval.

tests/test_single_topology_v3.py
```diff
@@ -383,3 +394,131 @@ def test_no_chiral_bond_restraints():
...
+    np.testing.assert_allclose(f(a, b, 1.0), b)
+
+    rng = np.random.default_rng(2022)
+    lambdas = rng.uniform(0, 1, (30,))
+    ys = vmap(functools.partial(f, a, b))(lambdas)
+    assert np.all(np.minimum(a, b) <= ys)
+    assert np.all(ys <= np.maximum(a, b))
+
+
+def test_handle_ring_opening_closing():
```
maxentile, line 410: Sorry if I missed it, but also add test asserting vacuum stability on hif2a dataset examples of ring opening/closing?

maxentile (commented):
Still reviewing, but looks great so far! A few initial comments.

timemachine/fe/interpolate.py
mcwitt: Yes; also in cases where we don't use `pad` (if there is no ring opening/closing), the end-states are special cases in `interpolate_harmonic_force_constant`

tests/test_single_topology_v3.py
mcwitt: Good call, I had not done this yet but will add.

timemachine/fe/single_topology_v3.py
mcwitt: Good call. I added the suggestion to the docstring in 861cc6d (coming up with a general type for `InterpolationFxn` at this stage seemed a bit complicated due to mixed conventions with passing scalars and arrays)

timemachine/fe/single_topology_v3.py
```diff
@@ -483,6 +484,221 @@ def find_dummy_groups_and_anchors(mol_a, mol_b, core_a, core_b):
...
+    Parameters
+    ----------
+    src_params : array-like, float, (2,)
+        force constant and equilibrium length at lambda=0
+
+    dst_params : array-like, float, (2,)
+        force constant and equilibrium length at lambda=1
+
+    lamb : float
+        control parameter
```
mcwitt, line 591: TODO: control -> alchemical

timemachine/fe/interpolate.py
mcwitt: Hopefully further addressed potential confusion in 4f91650b4f42380f1df478a098d15865261a4f60.

tests/test_single_topology_v3.py
```diff
@@ -383,3 +390,113 @@ def test_no_chiral_bond_restraints():
...
+
+
+def assert_interpolation_valid(f, a, b):
+    np.testing.assert_allclose(f(a, b, 0.0), a)
+    np.testing.assert_allclose(f(a, b, 1.0), b)
+
+    rng = np.random.default_rng(2022)
+    lambdas = rng.uniform(0, 1, (30,))
+    ys = vmap(functools.partial(f, a, b))(lambdas)
+    assert np.all(np.minimum(a, b) <= ys)
```
proteneer, line 402: is there a stronger condition we can check? i.e. not just that min(a,b) <= for all f(a,b,lambdas) <= max(a,b) but that it's monotonically increasing?

tests/test_single_topology_v3.py
proteneer: can we give you a couple of more test-cases of the form:

mol_a_sdf, mol_b_sdf, core_ab?

tests/test_single_topology_v3.py
```diff
@@ -383,3 +390,113 @@ def test_no_chiral_bond_restraints():
...
+    _ = f(src_k, dst_k, 0.1)
+
+
+def test_cyclic_difference():
+    assert cyclic_difference(0, 0, 1) == 0
+    assert cyclic_difference(0, 1, 2) == 1  # arbitrary, positive by convention
+    assert cyclic_difference(0, 0, 3) == 0
+    assert cyclic_difference(0, 1, 3) == 1
+    assert cyclic_difference(0, 2, 3) == -1
+    _ = jax.jit(cyclic_difference)(0, 1, 1)
```
proteneer, line 495: should we add a test 

1) with some negative numbers?  I'm not sure if we're guaranteed positive phases in the OFF spec
2) handle the case where accidentally b < a?

tests/test_single_topology_v3.py
mcwitt: oops, apparently I'd accidentally marked this as resolved. Still in progress..

timemachine/fe/interpolate.py
```diff
@@ -148,4 +148,36 @@ def linear_interpolation(src_params, dst_params, lamb):
...
+    return (1 - lamb) * jnp.asarray(src_params) + lamb * jnp.asarray(dst_params)
+
+
+def log_linear_interpolation(src_params, dst_params, lamb, min_value):
+    """
+    Linear interpolation in log space.
+
+    Notes
+    ----
+    * f(0) = src_params, f(1) = dst_params only if src_params and dst_params are strictly positive, respectively.
```
maxentile, line 160:
```suggestion
    * f(0) = src_params, f(1) = dst_params only if src_params and dst_params are >= min_value, respectively.
```

timemachine/fe/single_topology_v3.py
```diff
@@ -779,34 +995,71 @@ def setup_intermediate_state(self, lamb):
...
+
         angle = self._setup_intermediate_bonded_term(
-            src_system.angle, dst_system.angle, lamb, interpolate.align_harmonic_angle_idxs_and_params, interpolate_fn
+            src_system.angle,
+            dst_system.angle,
+            lamb,
+            interpolate.align_harmonic_angle_idxs_and_params,
+            partial(
+                interpolate_harmonic_angle_params,
+                k_min=0.05,
```
maxentile, line 1024: Optional nit: Add line comment (similar to the one for the value of `k_min` for harmonic_bonds above) if this value has been tuned / if there is a known acceptable range for this parameter?

maxentile (changes requested):
Approach and implementation both look solid to me!

My approval is just pending expansion of tests to assert vacuum stability on inputs in format of `mol_a, mol_b, core`, such as those plotted in the PR description.

tests/test_single_topology_v3.py
```diff
@@ -1,33 +1,40 @@
-# test that end-states are setup correctly in single topology calculations.
-import functools
 import multiprocessing
 import os
-from importlib import resources
-
-from timemachine.constants import DEFAULT_FF
-from timemachine.ff.handlers import openmm_deserializer
 
 os.environ["XLA_FLAGS"] = "--xla_force_host_platform_device_count=" + str(multiprocessing.cpu_count())
```
badisa, line 4: Should we set this in the makefile? Maybe thats not why the imports changed, if so this is just a nit?

timemachine/fe/single_topology_v3.py
```diff
@@ -483,6 +484,221 @@ def find_dummy_groups_and_anchors(mol_a, mol_b, core_a, core_b):
...
+    ----------
+    f : callable, (src_k: float, dst_k: float, lam: float) -> float
+        interpolation function; should satisfy f(0) = src_k, f(1) = dst_k
+
+    src_k, dst_k : float, k >= 0
+        force constants at lambda=0 and lambda=1, respectively
+
+    lambda_min, lambda_max : float, in 0 < lambda_min < lambda_max < 1
+        interpolate in range [lambda_min, lambda_max] (pin to end states otherwise). Note that if dst_k=0, the
+        convention is flipped so that 1 - lambda_min corresponds to f(0) and 1 - lambda_max corresponds to f(1).
```
badisa, line 511: Missing return description

timemachine/fe/single_topology_v3.py
```diff
@@ -483,6 +484,221 @@ def find_dummy_groups_and_anchors(mol_a, mol_b, core_a, core_b):
...
+
+    return jnp.array([k, phase])
+
+
+def interpolate_periodic_torsion_params(src_params, dst_params, lamb, lambda_min, lambda_max):
+    """
+    Interpolate periodic torsion parameters using
+
+    1. Linear interpolation for force constants (see note on ring
+       closing in `interpolate_harmonic_bond_params` docstring)
```
badisa, line 667: Not sure what note this is referring to. Don't see anything about ring closing in that docstring

timemachine/fe/single_topology_v3.py
```diff
@@ -483,6 +484,221 @@ def find_dummy_groups_and_anchors(mol_a, mol_b, core_a, core_b):
...
+    src_params : array-like, float, (2,)
+        force constant, equilibrium dihedral angle, and periodicity at lambda=0
+
+    dst_params : array-like, float, (2,)
+        force constant and equilibrium dihedral angle, and periodicity at lambda=1
+
+    lamb : float
+        alchemical parameter
+
+    k_min, lambda_min, lambda_max : float
```
badisa, line 684: There is no k_min in the signature.

timemachine/fe/single_topology_v3.py
```diff
@@ -779,34 +995,71 @@ def setup_intermediate_state(self, lamb):
         """
         src_system = self.src_system
         dst_system = self.dst_system
-        interpolate_fn = interpolate.linear_interpolation
 
-        # tbd: use different interpolation functions later
+        # alchemical parameter values at which angle and torsion force constants transition from zero to nonzero for
+        # ring closing transformations
+        lambda_angles = 0.4  # for angles, src_k = 0 => k(lamb) = 0 when lamb < lambda_angles
```
badisa, line 1001: Pretty sure I understand what this is getting at, but not really clear on how to read this syntax
```suggestion
        lambda_angles = 0.4  # angle param interpolation for lamb < lambda_angles = 0
```

tests/test_single_topology_v3.py
mcwitt: Hm, not sure I understand. As far as I understand this needs to be set in the environment when jax is loaded at runtime.

This was not added in the PR; the change in the imports was just to combine two timemachine import sections that were previously separated by this line (so isort wouldn't automatically fix them).

timemachine/fe/single_topology_v3.py
mcwitt: Fixed in [4acc511](https://github.com/proteneer/timemachine/pull/820/commits/4acc511b0414be6e9286a1e7c2c9b1149cb81e87)

timemachine/fe/single_topology_v3.py
mcwitt: Fixed in [4acc511](https://github.com/proteneer/timemachine/pull/820/commits/4acc511b0414be6e9286a1e7c2c9b1149cb81e87).

timemachine/fe/single_topology_v3.py
mcwitt: Added in [df8632](https://github.com/proteneer/timemachine/pull/820/commits/cdf8632e032a7057ad4ba20f6c74cf7c50224c7f)

timemachine/fe/single_topology_v3.py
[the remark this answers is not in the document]
mcwitt: Added the ability to pass overrides for the parameter interpolation functions to `setup_intermediate_state` in [3003770](https://github.com/proteneer/timemachine/pull/820/commits/300377099ccf6981338d5c196bc508fc3289f815) (this will also be useful for setting up controls in stability tests)

timemachine/fe/single_topology_v3.py
mcwitt: Addressed in [3003770](https://github.com/proteneer/timemachine/pull/820/commits/300377099ccf6981338d5c196bc508fc3289f815)

tests/test_single_topology_v3.py
mcwitt: Should be addressed by [7960a0a](https://github.com/proteneer/timemachine/pull/820/commits/7960a0ad1e7b977e8760a3ed5fd682b9004b3bac)

tests/test_single_topology_v3.py
mcwitt: This is now checked here: https://github.com/proteneer/timemachine/pull/820/files#diff-e679449b0c7d9911ad6910fd3cac904a8aaa1af2b02ad5d7873de170daa291bcR448-R450

tests/test_single_topology_v3.py
mcwitt: > Sorry if I missed it, but also add test asserting vacuum stability on hif2a dataset examples of ring opening/closing?

This is so far seeming a bit tricky -- the cases I've uncovered so far seem to have pretty low probability / step of blowing up (i.e. require ~100k steps to blow up), so it seems like we'd need a lot of samples to be reasonably sure that we've eliminated the instabilities. Thinking of a couple options to test this:

1. Start from a seed and initial conformation that's known to blow up in a small number of steps; assert that it does blow up when we interpolate bonds/angles/torsions simultaneously, but is safe with sequencing. Downside: if we rely on a seed, this wouldn't be robust to minor code changes.

2. Test that assertions of the form `k(i, j, k, l) > 0 => k(i, j, k) > k_min and k(j, k, l) > k_min` hold (where `k_min` is the minimum safe angle force constant for enabling torsions)

timemachine/fe/single_topology_v3.py
mcwitt: Hopefully this is clarified by the docstring addition in https://github.com/proteneer/timemachine/pull/820/commits/300377099ccf6981338d5c196bc508fc3289f815

tests/test_single_topology_v3.py
proteneer: Okay - let's set up the examples in a separate PR then. Don't want to block the PR.

proteneer: (Ideally I still want to do something that involves sampling - but this may be a bit tricky as you've noted)

proteneer (approved):
lgtm - can't wait to test this out.

maxentile (approved):
>My approval is just pending expansion of tests to assert vacuum stability on inputs in format of mol_a, mol_b, core, such as those plotted in the PR description.

Never mind -- can go in a later PR, following discussion here: https://github.com/proteneer/timemachine/pull/820#discussion_r980635716

tests/test_single_topology_v3.py
mcwitt: @badisa were you thinking to set this in the Makefile `*_tests` targets? I think that would make sense for CI, but would break running individual tests with `pytest -k`

badisa: Perhaps at some point we should investigate `pytest-env` to set this in our pytest.ini. 

In the mean time perhaps just add a comment about not importing timemachine above the line? But not important to PR, so do at your discretion.

tests/test_single_topology_v3.py
```diff
@@ -392,98 +393,83 @@ def test_no_chiral_bond_restraints():
...
-
-    closing_idxs = idxs[: k // 2]
-    opening_idxs = idxs[k // 2 :]
-
-    src_k[closing_idxs] = 0.0
-    dst_k[opening_idxs] = 0.0
-
-    lambda_min = 0.3
-    lambda_max = 0.7
+@given(nonzero_force_constants, open_lambda_intervals, lambdas)
```
badisa: https://hypothesis.readthedocs.io/en/latest/details.html#hypothesis.given says that this is a `randomized` test. Is it deterministic randomness?

badisa: Do we need to do something like this? https://hypothesis.readthedocs.io/en/latest/details.html?highlight=random#making-random-code-deterministic

tests/test_single_topology_v3.py
```diff
@@ -439,7 +439,7 @@ def test_interpolate_harmonic_force_constant(src_k, dst_k, k_min, lambda_interva
     assert f(src_k, dst_k, 0.0) == src_k
     assert f(src_k, dst_k, 1.0) == dst_k
 
-    lambdas = np.linspace(lambda_min, lambda_max, 10)
+    lambdas = np.arange(0.01, 1.0, 0.01)
```
badisa, line 442: If I understand the `given` code, lambdas[0] can be greater than lambda_min, is that true and expected?

tests/test_single_topology_v3.py
badisa: Or perhaps the more explicit https://hypothesis.readthedocs.io/en/latest/reproducing.html?highlight=random#reproducing-a-test-run-with-seed?

mcwitt: Yes, good call, I think it makes sense to explicitly use `@seed`, otherwise it seems like we might see nondeterministic (if reproducible) test failures. It looks like [`derandomize`](https://hypothesis.readthedocs.io/en/latest/settings.html#hypothesis.settings.derandomize) might be another option

badisa: `derandomize` seems to depend on the actual function hash though, which seems dangerous. `@seed` makes things the most explicit.

tests/test_single_topology_v3.py
mcwitt: Yes, that is true and expected -- the interpolation function should be nondecreasing over the whole interval (excluding the end points) regardless of lambda_{min,max}

badisa (approved):

tests/test_single_topology_v3.py
mcwitt: > Do we need to do something like this?

My understanding is that this is only needed for tests that have sources of nondeterminism independent of the example generation, so I don't think it's needed.

tests/test_single_topology_v3.py
mcwitt: Created #850 to track

tests/test_single_topology_v3.py
```diff
@@ -383,3 +393,163 @@ def test_no_chiral_bond_restraints():
...
+    # antisymmetric
+    assert cyclic_difference(0, 1, 3) == -cyclic_difference(1, 0, 3)
+    assert cyclic_difference(0, 2, 3) == -cyclic_difference(2, 0, 3)
+
+    # translation invariant
+    assert cyclic_difference(0, 1, 3) == cyclic_difference(-1, 0, 3)
+    assert cyclic_difference(0, 4, 8) == cyclic_difference(-2, 2, 8) == cyclic_difference(-4, 0, 8)
+
+    # jittable
+    _ = jax.jit(cyclic_difference)(0, 1, 1)
```
maxentile, line 509: tests are currently failing due to a ​`hypothesis​` exception related to inconsistent timings of `test_cyclic_difference_antisymmetric` -- possibly the first call is invoking the compiler https://gitlab.com/relaytx/physics-platform/timemachine/-/jobs/3092557213#L1431 . If so, possibly setting​`deadline=None`​ , or adding a `​jit(cyclic_difference)(inputs_with_expected_shapes_and_types)` ​statement prior to all tests would resolve.

mcwitt: Just saw this and had arrived at a similar conclusion. I added a CI configuration profile setting `deadline=None` in [bbe57bc](https://github.com/proteneer/timemachine/pull/820/commits/bbe57bc7d013d6bfe387cd9b9327e6115bcd1546).

PR #821, opened 2022-09-08
Local minimization

proteneer:
This PR adds a local minimization step where a local region (defined by any atom near any ligand atom, including itself) is minimized using BFGS. This is used to deal with bad initial states containing dummy atoms that are stretched due to poor core geometries. The BFGS routine also adds a sanity guard to deal with the fixed point overflow, however, unlike:

https://github.com/proteneer/timemachine/blob/cd360eb4d26894eedcee10e126cbc9cc13ddd109/timemachine/md/ensembles.py#L40-L41

Which uses a fixed abs on a size-extensive energy, this instead looks at the energy *difference* relative to the start geometry (which is still size extensive, but less so).

### DIFF
--- a/tests/test_minimizer.py
+++ b/tests/test_minimizer.py
@@ -57,11 +57,7 @@ def test_local_minimize_water_box():
     bps, _ = openmm_deserializer.deserialize_system(system, cutoff=1.2)
     box0 += np.diag([0.1, 0.1, 0.1])  # remove any possible clashes at the boundary
 
-    impls = []
-    for bp in bps:
-        impls.append(bp.bound_impl(np.float32))
-
-    val_and_grad_fn = minimizer.get_val_and_grad_fn(impls, box0, lamb)
+    val_and_grad_fn = minimizer.get_val_and_grad_fn(bps, box0, lamb)
 
     free_idxs = [0, 2, 3, 6, 7, 9, 15, 16]
     frozen_idxs = set(range(len(x0))).difference(set(free_idxs))
--- a/tests/test_minimizer.py
+++ b/tests/test_minimizer.py
@@ -1,9 +1,12 @@
 from importlib import resources
 
+import numpy as np
 from rdkit import Chem
 
 from timemachine.constants import DEFAULT_FF
+from timemachine.fe.utils import to_md_units
 from timemachine.ff import Forcefield
+from timemachine.ff.handlers import openmm_deserializer
 from timemachine.md import builders, minimizer
 
 
@@ -41,3 +44,32 @@ def test_equilibrate_host():
     assert coords.shape[0] == host_coords.shape[0] + mol.GetNumAtoms()
     assert coords.shape[1] == host_coords.shape[1]
     assert box.shape == host_box.shape
+
+
+def test_local_minimize_water_box():
+    """
+    Test that we can locally relax a box of water by selecting some random indices.
+
+    Furthermore, assert that we can
+    """
+
+    system, x0, box0, _ = builders.build_water_system(4.0)
+    x0 = to_md_units(x0)
+    lamb = 0.0
+    bps, _ = openmm_deserializer.deserialize_system(system, cutoff=1.2)
+    box0 += np.diag([0.1, 0.1, 0.1])  # remove any possible clashes at the boundary
+
+    impls = []
+    for bp in bps:
+        impls.append(bp.bound_impl(np.float32))
+
+    val_and_grad_fn = minimizer.get_val_and_grad_fn(impls, box0, lamb)
+
+    free_idxs = [0, 2, 3, 6, 7, 9, 15, 16]
+    frozen_idxs = set(range(len(x0))).difference(set(free_idxs))
+    frozen_idxs = list(frozen_idxs)
+
+    x_opt = minimizer.local_minimize(x0, val_and_grad_fn, free_idxs)
+
+    np.testing.assert_array_equal(x0[frozen_idxs], x_opt[frozen_idxs])
+    assert np.linalg.norm(x0[free_idxs] - x_opt[free_idxs]) > 0.01
--- a/tests/test_minimizer.py
+++ b/tests/test_minimizer.py
@@ -49,8 +49,6 @@ def test_equilibrate_host():
 def test_local_minimize_water_box():
     """
     Test that we can locally relax a box of water by selecting some random indices.
-
-    Furthermore, assert that we can
     """
 
     system, x0, box0, _ = builders.build_water_system(4.0)
--- a/timemachine/fe/rbfe.py
+++ b/timemachine/fe/rbfe.py
@@ -109,8 +109,7 @@ def sample(initial_state, protocol):
 
     val_and_grad_fn = minimizer.get_val_and_grad_fn(bound_impls, initial_state.box0, initial_state.lamb)
 
-    # note, can't use np.any(np.isnan(initial_state.x0)) is False since np.any returns a bool type
-    assert not np.any(np.isnan(initial_state.x0)), "Initial coordinates contain nan"
+    assert np.all(np.isfinite(initial_state.x0)), "Initial coordinates contain nan or inf"
 
     x0_min = minimizer.local_minimize(initial_state.x0, val_and_grad_fn, free_idxs)
 
--- a/timemachine/fe/rbfe.py
+++ b/timemachine/fe/rbfe.py
@@ -108,7 +108,13 @@ def sample(initial_state, protocol):
     free_idxs = np.where(np.any(d_ij < cutoff, axis=0))[0].tolist()
 
     val_and_grad_fn = minimizer.get_val_and_grad_fn(bound_impls, initial_state.box0, initial_state.lamb)
+
+    # note, can't use np.any(np.isnan(initial_state.x0)) is False since np.any returns a bool type
+    assert not np.any(np.isnan(initial_state.x0)), "Initial coordinates contain nan"
+
     x0_min = minimizer.local_minimize(initial_state.x0, val_and_grad_fn, free_idxs)
+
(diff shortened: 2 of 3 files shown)


timemachine/md/minimizer.py
```diff
@@ -277,3 +278,83 @@ def equilibrate_host(
...
+    """
+
+    def val_and_grad_fn(coords):
+        nrgs = []
+        grads = []
+        for impl in impls:
+            g, _, u = impl.execute(coords, box, lamb)
+            nrgs.append(u)
+            grads.append(g)
+        return np.sum(nrgs, axis=0), np.sum(grads, axis=0)
```
maxentile, line 295: Nit: performance -- should this use SummedPotential?

timemachine/fe/rbfe.py
```diff
@@ -98,7 +99,25 @@ def sample(initial_state, protocol):
...
+    free_idxs = np.where(np.any(d_ij < cutoff, axis=0))[0].tolist()
+
+    val_and_grad_fn = minimizer.get_val_and_grad_fn(bound_impls, initial_state.box0, initial_state.lamb)
+
+    # note, can't use np.any(np.isnan(initial_state.x0)) is False since np.any returns a bool type
+    assert not np.any(np.isnan(initial_state.x0)), "Initial coordinates contain nan"
+
+    x0_min = minimizer.local_minimize(initial_state.x0, val_and_grad_fn, free_idxs)
+
+    assert not np.any(np.isnan(x0_min)), "Minimization resulted in a nan"
```
maxentile, line 117: Curiosity: Is there an example where this has occurred?

timemachine/fe/rbfe.py
```diff
@@ -98,7 +99,25 @@ def sample(initial_state, protocol):
...
+    # local minimize region around ligand
+    ligand_coords = initial_state.x0[initial_state.ligand_idxs]
+    d_ij = cdist(ligand_coords, initial_state.x0)
+    # if any atom is within any of the ligand's atom's ixn radius, flag it for minimization
+    free_idxs = np.where(np.any(d_ij < cutoff, axis=0))[0].tolist()
+
+    val_and_grad_fn = minimizer.get_val_and_grad_fn(bound_impls, initial_state.box0, initial_state.lamb)
+
+    # note, can't use np.any(np.isnan(initial_state.x0)) is False since np.any returns a bool type
+    assert not np.any(np.isnan(initial_state.x0)), "Initial coordinates contain nan"
```
maxentile, line 113:
```suggestion
    assert np.all(np.isfinite(initial_state.x0)), "Initial coordinates contain nan or inf"
```

timemachine/md/minimizer.py
```diff
@@ -277,3 +278,83 @@ def equilibrate_host(
...
+    """
+    Minimize a local region given selected idxs.
+
+    Parameters:
+    -----------
+    x0: np.array (N,3)
+        Coordinates
+
+    box0: np.array (3,3)
+        Frozen box
```
maxentile, line 310:
```suggestion
```

timemachine/md/minimizer.py
```diff
@@ -277,3 +278,83 @@ def equilibrate_host(
...
+
+    val_and_grad_fn: f: R^(Nx3,3x3) -> (R^1, R^Nx3)
+        Energy function
+
+    local_idxs: list of int
+        Unique idxs we allow to move.
+
+    Returns
+    -------
+    Optimized set of coordinates.
```
maxentile, line 320: Possibly ambiguous whether return has `len(x0)` or `len(local_idxs)`
```suggestion
    Optimized set of coordinates (N,3)
```

timemachine/md/minimizer.py
```diff
@@ -277,3 +278,83 @@ def equilibrate_host(
...
+    u_0, _ = val_and_grad_fn(x0)
+
+    # deal with overflow
+    guard_threshold = 1e6
+
+    def val_and_grad_fn_local(x_local):
+        x_prime = x0.copy()
+        x_prime[local_idxs] = x_local
+        u_full, grad_full = val_and_grad_fn(x_prime)
+        # we minimized the energy too much, likely an overflow
```
maxentile, line 336: "we minimized the energy too much" is not quite the right description of what we were seeing, as this issue was encountered on the first step of line search (and all subsequent configurations tried) on one of four test systems.

```suggestion
        # avoid being trapped when overflows spuriously appear as large negative numbers
        # remove after resolution of https://github.com/proteneer/timemachine/issues/481
```

tests/test_minimizer.py
```diff
@@ -41,3 +44,32 @@ def test_equilibrate_host():
     assert coords.shape[0] == host_coords.shape[0] + mol.GetNumAtoms()
     assert coords.shape[1] == host_coords.shape[1]
     assert box.shape == host_box.shape
+
+
+def test_local_minimize_water_box():
+    """
+    Test that we can locally relax a box of water by selecting some random indices.
+
+    Furthermore, assert that we can
```
maxentile, line 53:
```suggestion
```

timemachine/md/minimizer.py
```diff
@@ -277,3 +278,83 @@ def equilibrate_host(
...
+            nrgs.append(u)
+            grads.append(g)
+        return np.sum(nrgs, axis=0), np.sum(grads, axis=0)
+
+    return val_and_grad_fn
+
+
+def local_minimize(x0, val_and_grad_fn, local_idxs):
+    """
+    Minimize a local region given selected idxs.
```
maxentile, line 302: Should this be configurable to use either L-BFGS or FIRE?

maxentile (approved):

timemachine/fe/rbfe.py
proteneer: not yet (and hopefully never!)

timemachine/md/minimizer.py
proteneer: good call - let me update.

timemachine/md/minimizer.py
proteneer: I don't know the optimal initial step sizes for FIRE when dealing with over stretched bonds (and hoping that line search from BFGS takes care of this)

tests/test_minimizer.py
```diff
@@ -41,3 +44,26 @@ def test_equilibrate_host():
...
+    val_and_grad_fn = minimizer.get_val_and_grad_fn(bps, box0, lamb)
+
+    free_idxs = [0, 2, 3, 6, 7, 9, 15, 16]
+    frozen_idxs = set(range(len(x0))).difference(set(free_idxs))
+    frozen_idxs = list(frozen_idxs)
+
+    x_opt = minimizer.local_minimize(x0, val_and_grad_fn, free_idxs)
+
+    np.testing.assert_array_equal(x0[frozen_idxs], x_opt[frozen_idxs])
+    assert np.linalg.norm(x0[free_idxs] - x_opt[free_idxs]) > 0.01
```
mcwitt, line 69: nit: worth checking that the energy has decreased?

timemachine/fe/rbfe.py
```diff
@@ -98,7 +99,24 @@ def sample(initial_state, protocol):
...
+    # if any atom is within any of the ligand's atom's ixn radius, flag it for minimization
+    free_idxs = np.where(np.any(d_ij < cutoff, axis=0))[0].tolist()
+
+    val_and_grad_fn = minimizer.get_val_and_grad_fn(initial_state.potentials, initial_state.box0, initial_state.lamb)
+
+    assert np.all(np.isfinite(initial_state.x0)), "Initial coordinates contain nan or inf"
+
+    x0_min = minimizer.local_minimize(initial_state.x0, val_and_grad_fn, free_idxs)
+
+    assert not np.any(np.isnan(x0_min)), "Minimization resulted in a nan"
```
mcwitt, line 116: nit: any reason not to assert all finite here (as opposed to just no NaNs) as above?

timemachine/fe/rbfe.py
```diff
@@ -98,7 +99,24 @@ def sample(initial_state, protocol):
...
+
+    # (ytz): re-use the initial states' v0?
```
mcwitt, line 118: Is this comment still relevant? (it looks like we do use `initial_state.v0` below)

timemachine/fe/rbfe.py
```diff
@@ -213,7 +236,11 @@ def setup_initial_states(st, host_config, temperature, lambda_schedule, seed):
         ligand_conf = st.combine_confs(mol_a_conf, mol_b_conf)
         combined_conf = np.concatenate([host_conf, ligand_conf])
         x0 = combined_conf
-        v0 = np.zeros_like(x0)
+        v0 = np.zeros_like(x0)  # tbd resample from Maxwell-boltzman?
+        num_ligand_atoms = len(ligand_conf)
+        num_total_atoms = len(x0)
+        ligand_idxs = list(range(num_total_atoms - num_ligand_atoms, num_total_atoms))
```
mcwitt, line 242: nit: maybe
```suggestion
        ligand_idxs = np.arange(num_total_atoms - num_ligand_atoms, num_total_atoms)
```

tests/test_minimizer.py
mcwitt: oops, never mind, I see that we already have an assertion for this in the minimizer.

mcwitt (approved):

timemachine/fe/rbfe.py
proteneer: willfix (forgot to copy the above test)

timemachine/fe/rbfe.py
proteneer: will remove

timemachine/fe/rbfe.py
proteneer: willfix inline

PR #822, opened 2022-09-08
[Draft] updates to rbfe protocol

jkausrelay:
- Currently an edge with default parameters will generate trajectories that are 4 GB (complex) and 1.2 GB (solvent).
- Make option to set steps_per_frame as a kwarg, so the file size can be decreased
- Decrease # of lambdas from 21 to 16

### DIFF
--- a/timemachine/fe/rbfe.py
+++ b/timemachine/fe/rbfe.py
@@ -510,7 +510,7 @@ def estimate_relative_free_energy(
     single_topology = SingleTopologyV3(mol_a, mol_b, core, ff)
 
     if lambda_schedule is None:
-        lambda_schedule = np.array([0.0, 0.01, 0.02, 0.04, 0.06, 0.08, 0.11, 0.15, 0.20, 0.32, 0.42])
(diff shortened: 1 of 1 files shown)


timemachine/fe/rbfe.py
```diff
@@ -116,7 +116,7 @@ def sample(initial_state, protocol):
         n_steps=n_steps,
         lambda_windows=[initial_state.lamb],
         store_u_interval=protocol.steps_per_frame,
-        store_x_interval=protocol.steps_per_frame,
+        store_x_interval=int(protocol.steps_per_frame / protocol.subsample_traj),
```
proteneer, line 119: 1) I think this also affects the # of samples used in the free energy estimate.
2) Diving by subsample_traj would *increase* number of frames we'd get.  (eg `subsample_traj=2`, `steps_per_frame=100` would result a `store_x_interval` of `50`


I'd rather subsampling of frames be done *outside* of the estimator, and it should only affect the storage component. It might be easier to instead do `simulation_result.frames[::subsample_traj]` instead.

jkausrelay: good point

PR #823, opened 2022-09-08
Fix overlap summary plot

mcwitt:
Fixes a bug that resulted in incorrect overlap values in the RBFE overlap summary plots.

```python
u_kn = np.concatenate(u_kln, axis=1)  # (k, l, n) -> (l, k*n) INCORRECT
u_kn = u_kln.reshape(k, -1)           # (k, l, n) -> (k, l*n)
```

### DIFF
--- a/timemachine/fe/rbfe.py
+++ b/timemachine/fe/rbfe.py
@@ -408,9 +408,10 @@ def estimate_free_energy_given_initial_states(initial_states, protocol, temperat
     ukln_by_lambda_by_component = np.array(ukln_by_component_by_lambda).swapaxes(0, 1)
 
     def pair_overlap(u_kln):
-        assert u_kln.shape[:2] == (2, 2)
-        u_kn = np.concatenate(u_kln, axis=1)
-        N_k = u_kln.shape[2] * np.ones(u_kn.shape[0])
+        k, l, n = u_kln.shape
+        assert k == l == 2
+        u_kn = u_kln.reshape(k, -1)  # (k, l * n)
+        N_k = n * np.ones(l)
         return pymbar.MBAR(u_kn, N_k).computeOverlap()["matrix"][0, 1]  # type: ignore
 
     lambdas = [s.lamb for s in initial_states]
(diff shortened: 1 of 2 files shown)


timemachine/fe/rbfe.py
```diff
@@ -408,9 +408,10 @@ def estimate_free_energy_given_initial_states(initial_states, protocol, temperat
     ukln_by_lambda_by_component = np.array(ukln_by_component_by_lambda).swapaxes(0, 1)
 
     def pair_overlap(u_kln):
-        assert u_kln.shape[:2] == (2, 2)
-        u_kn = np.concatenate(u_kln, axis=1)
-        N_k = u_kln.shape[2] * np.ones(u_kn.shape[0])
+        k, l, n = u_kln.shape
+        assert k == l == 2
+        u_kn = u_kln.reshape(k, -1)  # (k, l * n)
```
maxentile, line 413: Maybe comment -> assertion?
```suggestion
        u_kn = u_kln.reshape(k, -1)
        assert u_kn.shape == (k, l * n)
```

maxentile (approved):
Good catch!

timemachine/fe/rbfe.py
```diff
@@ -408,9 +408,10 @@ def estimate_free_energy_given_initial_states(initial_states, protocol, temperat
     ukln_by_lambda_by_component = np.array(ukln_by_component_by_lambda).swapaxes(0, 1)
 
     def pair_overlap(u_kln):
```
jkausrelay, line 410: Might be useful to move out of this function and add a basic test.

timemachine/fe/rbfe.py
mcwitt: Good call, added in [85f052f](https://github.com/proteneer/timemachine/pull/823/commits/85f052f684fe01cc569212f384ec8944b54d1c26)

timemachine/fe/rbfe.py
mcwitt: Done in [986dc1e](https://github.com/proteneer/timemachine/pull/823/commits/986dc1e3cd964901dcc7538b3a91d6d12545094d). Unfortunately, I wasn't able to figure out a simple example where the pair BAR overlap computed on `u_kln` is different from that on `u_kln.swapaxes(0, 1)` (i.e. the test added also passes with the bug present). I still want to understand why the ordering of axes doesn't seem to matter for gaussians, but maybe this is good enough to merge for now?

jkausrelay (approved):

PR #824, opened 2022-09-09
Asserts that charge between mols match in SingleTopologyV3

badisa:
* Avoids running charge perturbations

### DIFF
--- a/timemachine/fe/single_topology_v3.py
+++ b/timemachine/fe/single_topology_v3.py
@@ -4,6 +4,7 @@ from typing import Tuple
 
 import jax.numpy as jnp
 import numpy as np
+from rdkit import Chem
 
 from timemachine.fe import interpolate, system, topology, utils
 from timemachine.fe.dummy import canonicalize_bond, identify_dummy_groups, identify_root_anchors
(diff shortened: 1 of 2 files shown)


timemachine/fe/single_topology_v3.py
```diff
@@ -508,6 +513,11 @@ def __init__(self, mol_a, mol_b, core, forcefield):
         self.ff = forcefield
         assert core.shape[1] == 2
 
+        a_charge = Chem.GetFormalCharge(mol_a)
```
proteneer, line 516: lgtm - I think this is fine. We probably want to actually assert that the nonbonded charges themselves have the same sum later on, but it'd be a pretty egregious bug if nonbonded charges did not sum to the formal charge.

proteneer (approved):

jkausrelay (approved):

timemachine/fe/single_topology_v3.py
badisa: We do make the assumption that the `sum(atom_charges) == sum(parameterized_atom_charges)` if that is what you mean?

proteneer: ah great - i forgot

PR #828, opened 2022-09-14
Local minimization (#821)

proteneer:
This PR:

-Reduces search radius of BFGS from 1nm to 0.5nm (reducing number of atoms from 1200 to 300)
-Improve the guard threshold from `1e6` to `5e4`, improving the overflow behavior since we're operating on delta Us

### DIFF
--- a/timemachine/fe/rbfe.py
+++ b/timemachine/fe/rbfe.py
@@ -99,9 +99,9 @@ def sample(initial_state, protocol):
     intg_impl = initial_state.integrator.impl()
     baro_impl = initial_state.barostat.impl(bound_impls)
 
-    # minimize the local region, removing clashes and unstretching some bonds
-    cutoff = 1.0  # in nanometers
-    # local minimize region around ligand
+    cutoff = 0.5  # in nanometers
+    # local minimize region around ligand, should result in ~300 atoms
+
     ligand_coords = initial_state.x0[initial_state.ligand_idxs]
     d_ij = cdist(ligand_coords, initial_state.x0)
     # if any atom is within any of the ligand's atom's ixn radius, flag it for minimization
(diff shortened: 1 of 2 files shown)


maxentile (approved):

timemachine/fe/rbfe.py
```diff
@@ -99,9 +99,9 @@ def sample(initial_state, protocol):
     intg_impl = initial_state.integrator.impl()
     baro_impl = initial_state.barostat.impl(bound_impls)
 
-    # minimize the local region, removing clashes and unstretching some bonds
-    cutoff = 1.0  # in nanometers
-    # local minimize region around ligand
+    cutoff = 0.5  # in nanometers
+    # local minimize region around ligand, should result in ~300 atoms
```
maxentile, line 103: Possible alternatives to consider in the future:
* If we want to directly control the number of atoms, we could sort by distance to ligand, and take the top `K=300` atoms, rather than taking all atoms within a radius
* Could modify the selection to include all clashy atoms, rather than atoms based on distance to a ligand assumed to be the sole source of clashes

timemachine/md/minimizer.py
```diff
@@ -326,8 +326,8 @@ def local_minimize(x0, val_and_grad_fn, local_idxs):
     x_local_shape = (len(local_idxs), 3)
     u_0, _ = val_and_grad_fn(x0)
 
-    # deal with overflow
-    guard_threshold = 1e6
+    # deal with overflow, empirically obtained by testing on some real systems.
+    guard_threshold = 5e4
```
maxentile, line 330: Possible alternatives to consider in the future:
* Detect possible precision loss in more cases by also inspecting the gradient `norm(grad(f)(x))` rather than only the function values `abs(f(x))` or `abs(f(x) - f(x0))` or similar
* This can be interpreted as restricting the search to the set {x : abs(f(x) - f(x0)) < 5e4}. Could make this constraint more restrictive, e.g. by also enforcing small distance from x0.

Possible investigation in the future:
* Reproduce this issue as a possible symptom of https://github.com/proteneer/timemachine/issues/481 , since violation of line search assumptions is likely a canary for self-inconsistency of function, gradient implementations at some values of x.

PR #829, opened 2022-09-14
Bootstrap BAR uncertainty

maxentile:
Replaces pymbar3 BAR uncertainty with bootstrap in `rbfe.py` and `estimator.py`
* bootstrap samples are computed at looser relative tolerance, for speed
* a timeout is provided, in case instance size or difficulty is greater than expected
* a warning is logged if distribution of bootstrapped estimates deviates a lot from normality

Introduces a deboggling to-do: remove this code after upgrade to pymbar4, which supports bootstrapping.

### DIFF
--- a/timemachine/fe/bar.py
+++ b/timemachine/fe/bar.py
@@ -107,7 +107,7 @@ def bootstrap_bar(w_F, w_R, n_bootstrap=1000, timeout=10):
     w_R : array
         reverse works
     n_bootstrap : int
-        # boostrap samples
+        # bootstrap samples
     timeout : int
         in seconds
 
--- a/timemachine/fe/bar.py
+++ b/timemachine/fe/bar.py
@@ -1,7 +1,14 @@
+import logging
+from time import time
+
 import jax
 import jax.numpy as jnp
+import numpy as np
 import pymbar
 from jax.scipy.special import logsumexp
+from scipy.stats import normaltest
+
+logger = logging.getLogger(__name__)
 
 
 def EXP(w_raw):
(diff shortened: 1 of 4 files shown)


timemachine/fe/bar.py
```diff
@@ -88,3 +95,76 @@ def dG_dw(w):
...
+        if elapsed_time > timeout:
+            break
+
+        inds_F = np.random.randint(0, n_F, n_F)
+        inds_R = np.random.randint(0, n_R, n_R)
+
+        bar_result = pymbar.BAR(
+            w_F=w_F[inds_F],
+            w_R=w_R[inds_R],
+            DeltaF=full_bar_result,  # warm start
```
proteneer, line 146: nice!

timemachine/fe/bar.py
```diff
@@ -88,3 +95,76 @@ def dG_dw(w):
...
+    df, bootstrap_dfs = bootstrap_bar(w_F, w_R, n_bootstrap=n_bootstrap, timeout=timeout)
+
+    # warn if bootstrap distribution deviates significantly from normality
+    normaltest_result = normaltest(bootstrap_dfs)
+    pvalue_threshold = 1e-3  # arbitrary, small
+    if normaltest_result.pvalue < pvalue_threshold:
+        logger.warning(f"bootstrapped errors non-normal: {normaltest_result}")
+
+    # regardless, summarize as if normal
+    ddf = np.std(bootstrap_dfs)
```
proteneer, line 169: later on it might be useful to just return the 95% CI, but 1) I have no idea how to propagate the CI errors 2) how to use them in MLE settings, this is probably fine.

proteneer (approved):
slick - any idea on performance? Guessing it will be # of times to do bootstrap * time to do one BAR calculation in the worst case?

timemachine/fe/bar.py
maxentile: This is why I exposed two functions: one that returns an array of bootstrap samples (which can be propagated through later steps, visualized in a histogram, or summarized directly `ci_95 = np.quantile(bootstrap_dfs, [0.025, 0.975])` if desired).

For (1) and (2), it would be most general to propagate the samples and summarize at the end, rather than to write code for propagating specific kinds of summaries such as intervals or Gaussians.

However, I think the requirements we discussed yesterday were to return Gaussian error summaries from each edge, and then use diffnet for error propagation from Gaussian errors on edges to Gaussian errors on nodes. We could also avoid ever assuming Gaussian errors, e.g. by bootstrapping the whole diffnet-included estimator as a function of the works.

maxentile:
> slick - any idea on performance? Guessing it will be # of times to do bootstrap * time to do one BAR calculation in the worst case?

About 1 second for each of the test instances, which have `n_F=n_R=2000` and varying difficulty. (Chosen to be the maximum size instance likely to come out of rbfe.py with default settings.)

`time to do one BAR calculation` depends on `n_F`, `n_R`, initialization, difficulty, solver tolerance, and chosen solver method. Of the controllable factors, I make initialization and solver tolerance slightly easier than pymbar default in the loop that generates bootstrap samples. This seems to have minimal impact on the result, but provides a decent speed-up on more difficult instances (i.e. instances where stddev(w_F), stddev(w_R) are large).

Currently `# of times to do bootstrap` work is done serially, but if this becomes a bottleneck, these tasks are all independent and would be suitable for `multiprocessing`.

PR #815, opened 2022-09-06
disable chiral restraints

jkausrelay:
- Based on offline discussion w/Yutong, disable the chiral restraints for now

PR #832, opened 2022-09-14
Fix failure to evaluate intermediate potential when there are no chiral restraints, disable chiral restraints in vacuum simulations

mcwitt:
## What
- Fixes bug causing intermediate vacuum potential evaluation to fail when there are no chiral atom restraints or no chiral bond restraints
- Adds test cases
- Disables chiral restraints in vacuum simulations (similar to #815)

## Why
Currently
```python
U_fn = st.setup_intermediate_state(0.1).get_U_fn()
_ = U_fn(conf)
```
fails with `TypeError: len() of unsized object` if the intermediate does not have any chiral atom (or bond) restraints (as is the case for the benzene-iodobenzine transformation in the added test). This adds a branch that returns zero energy if `idxs` is empty.

~Should we also disable chiral restraints for vacuum simulations in this PR (as in #815 ?)~ Done.

### DIFF
--- a/tests/test_chiral_restraints.py
+++ b/tests/test_chiral_restraints.py
@@ -5,6 +5,7 @@ os.environ["XLA_FLAGS"] = "--xla_force_host_platform_device_count=" + str(multip
 
 import jax.numpy as jnp
 import numpy as np
+import pytest
 import scipy
 from rdkit import Chem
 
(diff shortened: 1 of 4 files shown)


maxentile (approved):
Good catch!

mcwitt:
@jkausrelay chiral restraints disabled for vacuum simulations in 752d603.

jkausrelay (approved):

proteneer (approved):
lgtm

mcwitt:
Manually squashed to separate https://github.com/proteneer/timemachine/pull/832/commits/c0c641ba2d719b0edcb7ed57dc06959d297e8c8d for eventual rollback.

I think we also need to disable `test_chiral_topology` until chiral restraints are added back (skipped in c0c641b).

PR #833, opened 2022-09-15
Run vacuum leg

proteneer:
This PR adds a vacuum leg to allow for quick failure. If the BAR error of the vacuum leg is `>1kJ/mol` then the simulation is aborted. Note that we currently don't save any statistics, and only assert, since saving the results would break the `run_pair` API (since it would return an extra `vacuum_res`)

### DIFF
--- a/timemachine/fe/rbfe.py
+++ b/timemachine/fe/rbfe.py
@@ -98,7 +98,10 @@ def sample(initial_state, protocol):
 
     bound_impls = [p.bound_impl(np.float32) for p in initial_state.potentials]
     intg_impl = initial_state.integrator.impl()
-    baro_impl = initial_state.barostat.impl(bound_impls)
+    if initial_state.barostat:
+        baro_impl = initial_state.barostat.impl(bound_impls)
+    else:
+        baro_impl = None
 
     cutoff = 0.5  # in nanometers
     # local minimize region around ligand, should result in ~300 atoms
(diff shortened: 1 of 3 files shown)


timemachine/fe/rbfe.py
```diff
@@ -562,6 +569,24 @@ def estimate_relative_free_energy(
...
+        forcefield,
+        vacuum_host_config,
+        seed,
+        n_frames=n_frames,
+        prefix="vacuum",
+        n_eq_steps=n_eq_steps,
+    )
+
+    # vacuum error should be below a threshold, otherwise this edge is likely problematic
+    assert np.linalg.norm(vacuum_res.all_dGs) < 1.0
```
maxentile, line 588: Curiosity: Has the false positive rate of this assertion been checked in some way? (Intuitively this looks likely to introduce spurious assertion errors.)

Clarification: Is this assertion intended to reference the estimated dGs or the estimated errors? 

```suggestion
    # vacuum error should be below a threshold, otherwise this edge is likely problematic
    assert np.linalg.norm(vacuum_res.all_errs) < 1.0
```

timemachine/fe/rbfe.py
```diff
@@ -562,6 +569,24 @@ def estimate_relative_free_energy(
...
+
```
jkausrelay, line 589: I'd rather have this as a separate function that just returns the vacuum_res. Then caller can decide what checks should be done and how to handle them.

maxentile (changes requested):

timemachine/fe/rbfe.py
jkausrelay: Or just call estimate_relative_free_energy directly.

proteneer:
> Curiosity: Has the false positive rate of this assertion been checked in some way? (Intuitively this looks likely to introduce spurious assertion errors.)

Usually the complex/solvent errors are greater than that of vacuum errors. A 1 kJ/mol vacuum error seems pretty high to me, I could change this into a warning?

maxentile:
> > Curiosity: Has the false positive rate of this assertion been checked in some way? (Intuitively this looks likely to introduce spurious assertion errors.)
> 
> Usually the complex/solvent errors are greater than that of vacuum errors. A 1 kJ/mol vacuum error seems pretty high to me, I could change this into a warning?

Was just curious how this had been tested, since prior to https://github.com/proteneer/timemachine/pull/833/commits/697dd9213737706779de6da6da4cd10da4991d11 it was an assertion that `norm(all_dGs) < 1`, which intuitively seems likely to introduce spurious assertion errors.

After https://github.com/proteneer/timemachine/pull/833/commits/697dd9213737706779de6da6da4cd10da4991d11 , this looks much less likely to cause spurious assertion errors. The numerical value of the threshold sounds reasonable to me, and I am no longer curious if the false positive rate at that threshold has been checked in some way.

jkausrelay (approved):

PR #834, opened 2022-09-15
Fixes dependence on atom ordering for get_core_with_alignment

badisa:
* Adds match_hydrogens flag to timemachine.fe.atom_mapping::mcs

### Problem
In `get_core_with_alignment` the mols would have their hydrogens removed then build a core using the hydrogen free mols. It assumed that all of the hydrogens were at the end of the molecules so that the core would be valid. This is not always true though, result in an unstable core for use during alignment which caused fewer atoms to be mapped.

### Solution
In `mcs` adds a flag `match_hydrogens` that if it is set to False, will generate the match query using the hydrogen less mols which works since it uses smarts rather than indexs.

### DIFF
--- a/timemachine/fe/atom_mapping.py
+++ b/timemachine/fe/atom_mapping.py
@@ -70,7 +70,7 @@ def mcs(
     # try on given mols
     result = rdFMCS.FindMCS([a, b], params)
 
-    # optional fallbackmcs
+    # optional fallback
     def is_trivial(mcs_result) -> bool:
         return mcs_result.numBonds < 2
 
--- a/timemachine/fe/atom_mapping.py
+++ b/timemachine/fe/atom_mapping.py
@@ -14,7 +14,16 @@ from timemachine.ff import Forcefield
 from timemachine.md.align import align_mols_by_core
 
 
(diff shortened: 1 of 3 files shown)


timemachine/fe/atom_mapping.py
```diff
@@ -40,12 +51,22 @@ def mcs(a, b, threshold: float = 2.0, timeout: int = 5, smarts: Optional[str] =
...
     params.Timeout = timeout
     if smarts is not None:
         params.InitialSeed = smarts
 
+    def strip_hydrogens(mol):
+        """Strip hydrogens with deepcopy to be extra safe"""
+        return Chem.RemoveHs(deepcopy(mol))
+
+    if not match_hydrogens:
+        # Setting CompareAnyHeavyAtom doesn't handle this correctly, strip hydrogens explicitly
```
maxentile, line 64: If possible, are there any details to share on how `CompareAnyHeavyAtom` fails to restrict the MCS to heavy atoms only?

jkausrelay (approved):

maxentile (approved):

timemachine/fe/atom_mapping.py
maxentile: (Sorry, meant to delete this comment before posting review-- resolving.)

proteneer (approved):
lgtm

PR #836, opened 2022-09-16
Freeze random seed in `bootstrap_bar`

maxentile:
Makes `bootstrap_bar` deterministic aside from possible `timeout`.

Later, can modify to allow passing an optional random seed / RNG state, but the current PR just hard-codes the random seed, to avoid changing the signature in any way.

### DIFF
--- a/timemachine/fe/bar.py
+++ b/timemachine/fe/bar.py
@@ -140,12 +140,12 @@ def bootstrap_bar(w_F, w_R, n_bootstrap=1000, timeout=10):
         if elapsed_time > timeout:
             break
 
-        w_F_sample = rng.choice(w_F, replace=True)
-        w_R_sample = rng.choice(w_R, replace=True)
+        w_F_sample = rng.choice(w_F, size=(n_F,), replace=True)
+        w_R_sample = rng.choice(w_R, size=(n_R,), replace=True)
 
         bar_result = pymbar.BAR(
-            w_F=w_F[inds_F],
-            w_R=w_R[inds_R],
+            w_F=w_F_sample,
+            w_R=w_R_sample,
             DeltaF=full_bar_result,  # warm start
             compute_uncertainty=False,
             relative_tolerance=1e-6,  # reduce cost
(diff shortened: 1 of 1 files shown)


timemachine/fe/bar.py
```diff
@@ -132,13 +132,16 @@ def bootstrap_bar(w_F, w_R, n_bootstrap=1000, timeout=10):
 
     t0 = time()
 
+    seed = 2022
+    rng = np.random.RandomState(seed)
```
mcwitt, line 136: nit: any reason not to use the newer [`Generator` API](https://numpy.org/doc/stable/reference/random/generator.html) instead?
```suggestion
    rng = np.random.default_rng(seed)
```

maxentile: Nope! Updated

timemachine/fe/bar.py
```diff
@@ -132,13 +132,16 @@ def bootstrap_bar(w_F, w_R, n_bootstrap=1000, timeout=10):
...
+
     for _ in range(n_bootstrap):
         elapsed_time = time() - t0
         if elapsed_time > timeout:
             break
 
-        inds_F = np.random.randint(0, n_F, n_F)
-        inds_R = np.random.randint(0, n_R, n_R)
+        inds_F = rng.randint(0, n_F, n_F)
+        inds_R = rng.randint(0, n_R, n_R)
```
mcwitt, line 144: nit: to me it's slightly more obvious to do something like
```suggestion
        w_F_sample = rng.choice(w_F, replace=True)
        w_R_sample = rng.choice(w_R, replace=True)
```
(e.g. don't need to think about whether the bounds of `randint` are inclusive or not)

maxentile: Applied. One reason to prefer `inds` in the future might be if we model correlation between neighboring pair bar estimates (0 -> 1, 1 -> 2 are not independent, and could reference the same bootstrap sample from lambda idx 1 twice), but this is not possible with the current signature anyway.

mcwitt: Makes sense. In that case I think you could still use `inds_F = rng.choice(n_F, size=(n_F,))` and not have to think about inclusivity of bounds.

mcwitt (approved):

PR #838, opened 2022-09-19
Overlap summary improvements

mcwitt:
- Scale pairwise overlap statistic by 2 so its range is [0, 1] instead of [0, 0.5]
- Set y axis limits in overlap summary plot to the allowed range of overlap values, [0, 1]


proteneer (approved):
lgtm!

maxentile (approved):
Looks good!

PR #839, opened 2022-09-20
Avoid computing BAR on energy components

maxentile:
Based on observation from @badisa that `WARNING:timemachine.fe.bar:bootstrapped errors non-normal: NormaltestResult(statistic=2705.698965619963, pvalue=0.0)` is being repeated many times (with identical numerical values!) in logs:

~Revert from `logger.warning` -> `print`, to eliminate possibility of repeated message bug due to some unexpected interaction with `logging`. (I have not confirmed that this is the cause, because I have not reproduced the issue locally.)~ (@badisa established this is not the culprit!)

Modify inner-loop diagnostic plots -- which show histograms of w_f_component, w_r_component -- to avoid calling BAR on w_f_component, w_r_component. (I don't think it makes sense to decompose the free energy difference per potential energy component, although it is intuitively still helpful to look at work variance/overlap/etc by energy component. This retains the descriptive plots decomposed by energy component, but avoids computing two quantitative results `dg, dg_err` for each component.)

### DIFF
--- a/timemachine/fe/bar.py
+++ b/timemachine/fe/bar.py
@@ -1,3 +1,4 @@
+import logging
 from time import time
 
 import jax
@@ -6,6 +7,8 @@ import numpy as np
 import pymbar
 from jax.scipy.special import logsumexp
 from scipy.stats import normaltest
+
+logger = logging.getLogger(__name__)
 
 
 def EXP(w_raw):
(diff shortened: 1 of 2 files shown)


maxentile:
Thanks to @badisa for tracking this down: the warning was from calling BAR on vectors of zeros (which arises when calling BAR on energy components that are not changing).

It can be reproduced by:

```python
bar_with_bootstrapped_uncertainty(np.zeros(1000), np.zeros(1000))
```

```text
WARNING:timemachine.fe.bar:bootstrapped errors non-normal: NormaltestResult(statistic=2705.698965619963, pvalue=0.0)
```

The repeated spurious warnings should be avoided by the change to avoid calling BAR on individual energy components, so I have [reverted](https://github.com/proteneer/timemachine/pull/839/commits/4894f583ad7be300d0289c9ec2bcaa7724dcd6f7) an unnecessary `logging` change initially included in this PR.

maxentile:
@proteneer gave a 👍 offline, merging

PR #842, opened 2022-09-21
Minor clean-up in reference potentials

maxentile:
Mostly benign unopinionated fixes.

Two possibly objectionable style changes:

1. Edit a few parameter docstrings from

```python
    np.array: (3,)
        First point
    np.array: (3,)
        Second point
    np.array: (3,)
        Third point
    np.array: (3,)
        Fourth point
```

to

```python
    ci, cj, ck, cl: np.arrays of shape (3,)
        four points
```

2. Reduce duplication but increase indirection
```python
x = b - a

...

x = x / jnp.linalg.norm(x)
```
to
```python
x = normalize(b - a)
```


timemachine/potentials/chiral_restraints.py
```diff
@@ -2,6 +2,10 @@
 import jax.numpy as jnp
 
 
+def normalize(x):
```
proteneer, line 5: maybe rename to `unit_vector(x)`

timemachine/potentials/chiral_restraints.py
```diff
@@ -12,14 +16,8 @@ def pyramidal_volume(xc, x1, x2, x3):
...
 
-    x1: np.array: (3,)
-        First point
-
-    x2: np.array: (3,)
-        Second point
-
-    x3: np.array: (3,)
-        Third point
+    x1, x2, x3: np.arrays of shape (3,)
```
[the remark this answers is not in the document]
proteneer: (it's okay if it's not, since we don't generate docstrings, but just wanted to double check)

timemachine/potentials/bonded.py
```diff
@@ -175,23 +175,23 @@ def harmonic_angle(conf, params, box, lamb, angle_idxs, lamb_mult=None, lamb_off
     lamb: float
         alchemical lambda parameter, linearly rescaled
 
+    angle_idxs: shape [num_angles, 3] np.array
```
mcwitt, line 178: super tiny nit:
```suggestion
    angle_idxs: shape [num_angles, 3] np.ndarray
```
(`np.array` is a function)

timemachine/potentials/chiral_restraints.py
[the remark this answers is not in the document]
mcwitt: This seems to be allowed by the [numpydoc convention](https://numpydoc.readthedocs.io/en/latest/format.html#:~:text=when%20two%20or%20more%20input%20parameters%20have%20exactly%20the%20same%20type%2C%20shape%20and%20description%2C%20they%20can%20be%20combined%3A)

mcwitt (approved):
LGTM!

proteneer (approved):

timemachine/potentials/chiral_restraints.py
maxentile: This would be less clear in my opinion (noun vs. verb), and I'm not aware of an ambiguity with the current name ("normalize" doesn't seem easily confused with anything else in this context).

Can change this name in a subsequent pass if needed.

timemachine/potentials/chiral_restraints.py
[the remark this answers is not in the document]
maxentile: Thanks for confirming, @mcwitt .

(Mostly, was prompted to make this change when reading a similar docstring in `torsion_volume` https://github.com/proteneer/timemachine/pull/842/files#diff-604fade1f661f8c0a708698091a5188ad90784fcaa0b426054e4eeb08fef2dedL50-R45 that was also missing parameter names. PyCharm signature reminder stuff also seems to permit this style.)

timemachine/potentials/bonded.py
maxentile: Applied this change to all instances of `"np.array"` as type in `timemachine/potentials` docstrings in https://github.com/proteneer/timemachine/pull/842/commits/200ebce9d8faa508b2cca46aecfe639d9f0c5cda

PR #844, opened 2022-09-23
Make scipy.optimize warnings optional, formatted

maxentile:
Adds context for the internal scipy.optimize warnings emitted by `local_minimize` (https://github.com/proteneer/timemachine/pull/821, https://github.com/proteneer/timemachine/pull/828), to clarify that the messages apply to initial energy minimization and are not a diagnostic about the broader simulation.

(See e.g. the output in cell [27] of https://github.com/mcwitt/timemachine-demo/blob/a4a3809bd289257940ac41d6d564eaab552afbb5/examples/relative-hydration.ipynb for how this can be ambiguous.)


timemachine/md/minimizer.py
```diff
@@ -349,10 +354,36 @@ def val_and_grad_fn_bfgs(x_local_flattened):
...
+    method = "BFGS"
+
+    if verbose:
+        print("-" * 70)
+        n_local = len(local_idxs)
+        n_frozen = len(x0) - n_local
+        print(f"performing {method} minimization on {n_local} atoms\n(holding the other {n_frozen} atoms frozen)")
+        U_0, grad_0 = val_and_grad_fn_bfgs(x_local_0_flat)
+        print(f"U(x_0) = {U_0:.3f} kJ/mol")
+        print(f"|force(x_0)| = {np.linalg.norm(grad_0):.3f} kJ/mol / nm^2")
```
proteneer, line 366: is the intention to help with debugging the atom that has the highest force? If so, may want `norm` or `argmax` over per-atom force norm here (and print the offending atom idx). Otherwise, if the intention is to check convergence, not sure how useful this is since the norm force will pretty much never be zero.

mcwitt: nit: units are determined by the potential function we pass in, but the log messages are specialized to energies in kJ/mol. Should we specify that the potential function should return energies in kJ/mol?

timemachine/md/minimizer.py
```diff
@@ -349,10 +354,36 @@ def val_and_grad_fn_bfgs(x_local_flattened):
     x_local_0 = x0[local_idxs]
     x_local_0_flat = x_local_0.reshape(-1)
 
-    res = scipy.optimize.minimize(val_and_grad_fn_bfgs, x_local_0_flat, method="BFGS", jac=True, options={"disp": True})
+    method = "BFGS"
+
+    if verbose:
+        print("-" * 70)
```
mcwitt, line 360: Non blocking: in the future we might consider using [rich](https://github.com/Textualize/rich) to format messages like this (here could use [`console.rule`](https://rich.readthedocs.io/en/stable/console.html#rules))

mcwitt (approved):

timemachine/md/minimizer.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/844/commits/468194a525b5d4908be0e52ca093d92a286b81c0, https://github.com/proteneer/timemachine/pull/844/commits/8d06ec279d97e8533a19d1daa97a4061d7180704

PR #847, opened 2022-09-26
split run pair into run_leg

jkausrelay:
- split run pair into separate functions for each leg

### DIFF
--- a/examples/rbfe_edge_list.py
+++ b/examples/rbfe_edge_list.py
@@ -12,7 +12,7 @@ from simtk.openmm import app
 
 from timemachine.constants import KCAL_TO_KJ
 from timemachine.fe import atom_mapping
-from timemachine.fe.rbfe import run_pair
+from timemachine.fe.rbfe import run_complex, run_solvent
 from timemachine.fe.utils import get_mol_name
 from timemachine.ff import Forcefield
 from timemachine.parallel.client import CUDAPoolClient
@@ -45,7 +45,8 @@ def read_from_args():
 
     mols = [mol for mol in Chem.SDMolSupplier(str(args.ligands), removeHs=False)]
 
-    futures = []
+    cfutures = []
+    sfutures = []
     metadata = []
 
     cpc = CUDAPoolClient(args.n_gpus)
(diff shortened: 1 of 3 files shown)


timemachine/fe/rbfe.py
```diff
@@ -620,11 +638,10 @@ def run_pair(mol_a, mol_b, core, forcefield, protein, n_frames, seed, n_eq_steps
         core,
         forcefield,
         complex_host_config,
-        seed + 1,
+        seed,  # this was seed + 1, why?
```
proteneer, line 641: this was to avoid complex and solvent re-using the same seed. Not sure if it matters at all in practice.

examples/rbfe_edge_list.py
```diff
@@ -66,8 +67,12 @@ def read_from_args():
             print(f"Submitting job for {mol_a_name} -> {mol_b_name}")
             mcs_threshold = 2.0
             core, smarts = atom_mapping.get_core_with_alignment(mol_a, mol_b, threshold=mcs_threshold)
-            fut = cpc.submit(run_pair, mol_a, mol_b, core, forcefield, protein, args.n_frames, args.seed + row_idx)
-            futures.append(fut)
+            cfutures.append(
+                cpc.submit(run_complex, mol_a, mol_b, core, forcefield, protein, args.n_frames, args.seed + row_idx)
+            )
+            sfutures.append(
```
proteneer, line 73: does it make sense to append sfutures first? (since they're faster and will likely be finished first)

examples/rbfe_edge_list.py
```diff
@@ -83,14 +88,14 @@ def read_from_args():
...
-
+    for i, meta in enumerate(metadata):
         mol_a, mol_b, _, _, exp_ddg, fep_ddg, fep_ddg_err, ccc_ddg, ccc_ddg_err = meta
         mol_a_name = get_mol_name(mol_a)
         mol_b_name = get_mol_name(mol_b)
 
         try:
-            solvent_res, solvent_top, complex_res, complex_top = fut.result()
+            complex_res, complex_top = cfutures[i].result()
+            solvent_res, solvent_top = sfutures[i].result()
```
proteneer, line 98: oops - meant to put the comment here. might be good to get the `result()` of the solvent futures first since its faster, but it probably doesn't matter much in practice since the analysis needs both `complex` and `solvent`, maybe later on when we can decouple the windows entirely we can be more clever about this.

proteneer (approved):
lgtm

timemachine/fe/rbfe.py
jkausrelay: I doubt it would, since the dof are different anyways

examples/rbfe_edge_list.py
jkausrelay: Right. I'll swap in case the result for solvent throws first. Definitely can be improved later.

PR #853, opened 2022-09-28
Remove JIT compilation of custom interpolation expressions, jitify dependency

mcwitt:
This is the first part of the planned transition away from computing interpolated parameters in C++/CUDA.

- Renames `k_lambda_transformer_jit` -> `k_lambda_transformer` and specializes to linear interpolation
- Switches to static compilation of `k_lambda_transformer`
- Replaces driver calls (`kernel.configure().launch()`) with static kernel calls
- Removes kernel sources from Python `package_data`
- Removes vendored jitify dependency


badisa (approved):
🚀

timemachine/cpp/CMakeLists.txt
```diff
@@ -84,6 +84,7 @@ pybind11_add_module(${LIBRARY_NAME} SHARED NO_EXTRAS
   src/summed_potential.cu
   src/fanout_summed_potential.cu
   src/device_buffer.cu
+  src/kernels/k_lambda_transformer.cu
```
maxentile, line 87: May also need to modify another line in CMakeLists.txt (I think we can revert https://github.com/proteneer/timemachine/pull/403/files#diff-91d026c1fcb8424ad567b2343d63101fc584eb17536930067d7241f94ba7ae7eL63-R63 ?)

maxentile (approved):
🚀

timemachine/cpp/CMakeLists.txt
mcwitt: Good call, removed in 5829a43. Also removed `cudart` and `cudadevrt`, which seem to be unnecessary, in 4e1e72

timemachine/cpp/CMakeLists.txt
```diff
@@ -133,7 +134,7 @@ message("Cuda Arch: " ${CUDA_ARCH})
 set_property(TARGET ${LIBRARY_NAME} PROPERTY CUDA_ARCHITECTURES "${CUDA_ARCH}")
 unset(CUDA_ARCH CACHE)
 
-target_link_libraries(${LIBRARY_NAME} PRIVATE -lcurand -lcuda -lcudart -lcudadevrt -lnvrtc)
```
proteneer, line 136: wow!

proteneer (approved):
awesome changes :)

PR #855, opened 2022-09-29
remove hif2a_ligand_pair and setup_hif2a_ligand_pair

jkausrelay:


tests/test_benchmark.py
```diff
@@ -388,39 +345,21 @@ def test_hif2a():
 def test_nonbonded_interaction_group_potential(hi2fa_test_frames):
     pots, sys_params, frames, boxes, ligand_idxs = hi2fa_test_frames
     lambdas = np.array([0.0, 1.0])
-    nonbonded_interp = pots[3]
-    assert isinstance(nonbonded_interp, NonbondedInterpolated)
+    U_names = [type(U_fn).__name__ for U_fn in pots]
+    nb_idx = U_names.index("Nonbonded")
```
proteneer, line 349: nice! I was just doing `pots[-1]` - I like your approach better.

tests/test_velocity_verlet_integrator.py
```diff
@@ -1,3 +1,7 @@
+import jax
+
+jax.config.update("jax_enable_x64", True)
```
proteneer, line 3: is this new? (not sure if we just want to stuff this in a `conftest` later)

proteneer:
removing the `SingleTopology` from the return object has another nicer advantage too - the new `SingleTopologyV3` class is substantially slower to instantiate since it needs to setup the end-state. Simply returning the `mol_a, mol_b, and ff` obviates this issue.

tests/test_velocity_verlet_integrator.py
jkausrelay: Yeah, will fix with the conftest update, put here for now so the tests pass

tests/test_benchmark.py
mcwitt: nit: I think this also works

```python
nonbonded = next(p for p in pots if isinstance(p, Nonbonded))
```

jkausrelay: Sure, but I need the index for sys_params too.

mcwitt (approved):
Nice cleanup!

PR #857, opened 2022-09-30
Task/reduce docker container size

badisa:
Reduces the production container to ~4GB from ~8GB.
- Switch to the runtime container for the final container, means you can't build timemachine from the final container thought there is an intermediate container still that supports building.
- Remove the build directory (~200MB)

Some minor changes to the gitlab config to remove what is now a redundant env variable.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -99,9 +99,9 @@ ARG CUDA_ARCH=75
 ENV CMAKE_ARGS -DCUDA_ARCH=${CUDA_ARCH}
 COPY . /code/timemachine/
 WORKDIR /code/timemachine/
-RUN pip install --no-cache-dir -e .[dev,test]
+RUN pip install --no-cache-dir -e . && rm -rf ./build
 
-# Container with only cuda runtime
+# Container with only cuda runtime, half the size of dev container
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -93,9 +93,20 @@ RUN cd /code/timemachine && git init . && pre-commit install-hooks
 COPY ci/requirements.txt /code/timemachine/ci/requirements.txt
 RUN pip install --no-cache-dir -r timemachine/ci/requirements.txt
 
-FROM tm_base_env AS timemachine
+# Dev container that contains the cuda developer tools
+FROM tm_base_env AS timemachine_dev
```
mcwitt, line 97: Is the idea to leave `timemachine_dev` and define a new target in case the former is useful for debugging? (versus deleting the extra dev stuff not needed at runtime; seems fine to me, was just curious)

mcwitt (approved):

jkausrelay (approved):

Dockerfile
mcwitt: Got it, I think that makes sense. I'd missed that the new target was using a different base image.

PR #859, opened 2022-09-30
Deprecate old SingleTopology code

proteneer:
This PR removes the old SingleTopology class.

The new SingleTopologyV3 now extends a `AtomMapMixin`, for convenience when using the `pdb_writer`. This is because the constructor does a fair bit more work since it sets up the end-state systems. The performance benchmark tests also now default to single topology as opposed to dual topology. 

I've also eagerly removed various tests dealing with parameter interpolation as well, since they will be removed by @mcwitt soon. 

Note for @jkausrelay the `testsystems` in this PR still return a `SingleTopologyV3` object - but this can be easily resolved (regardless of the order of the merge). 

TODO:

-I will rename `SingleTopologyV3` back to `SingleTopology` in a final rebase (left as is to improve clarity for reviewers) 
-Rename `fe/single_topology_v3.py to fe/single_topology.py`


proteneer:
Removing reviewers for now - going to deal with the rebase before re-requesting a review.

timemachine/fe/single_topology_v3.py
```diff
@@ -799,6 +784,48 @@ def __init__(self, mol_a, mol_b, core, forcefield, lambda_angles=0.4, lambda_tor
...
+            For ring opening/closing transformations, alchemical parameter values controlling the intervals over which
+            bonds, angles, and torsions are interpolated. Note that these have no effect on terms not involved in ring
+            opening/closing.
+
+            - Bonds are interpolated in the interval [0, lambda_angles]
+            - Angles are interpolated in the interval [lambda_angles, lambda_torsions]
+            - Torsions are interpolated in the interval [lambda_torsions, 1]
+        """
+        # initialize the mixin to get the a_to_c, b_to_c, c_to_a, c_to_b, and c_flags
+        AtomMapMixin.__init__(self, mol_a, mol_b, core)
```
mcwitt, line 819: I'm curious, is there an advantage to using a mixin here versus modifying the `SingleTopologyV3` constructor to take an `AtomMap` instead of `mol_a`, `mol_b`, `core`? For the uses I'm imagining the latter seems a bit simpler (though not a huge difference)

proteneer: Going to answer this in the other PR #862

PR #860, opened 2022-09-30
Draft: Move jax related global config to conftest

jkausrelay:
- move XLA_FLAGS and jax_enable_x64 to conftest

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,21 @@
 # This file defines fixtures shared across multiple test modules.
 # https://docs.pytest.org/en/latest/how-to/fixtures.html#scope-sharing-fixtures-across-classes-modules-packages-or-session
 
+import multiprocessing
+import os
+
+# (ytz): not pretty, but this is needed to get XLA to be less stupid
+# see https://github.com/google/jax/issues/1408 for more information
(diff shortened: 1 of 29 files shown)


tests/conftest.py
```diff
@@ -1,6 +1,21 @@
...
+import multiprocessing
+import os
+
+# (ytz): not pretty, but this is needed to get XLA to be less stupid
+# see https://github.com/google/jax/issues/1408 for more information
+# needs to be set before xla/jax is initialized, and is set to a number
+# suitable for running on CI
+# NOTE: To have an effect, XLA_FLAGS must be set in the environment before loading JAX (whether directly or transitively
+# through another import)
+os.environ["XLA_FLAGS"] = "--xla_force_host_platform_device_count=" + str(multiprocessing.cpu_count())
```
badisa, line 13: nit: Might be good to see if it already exists in the env, and if so don't set it. Just in case someone wants to pass their own set of flags.

proteneer: +1 - assert if already set

jkausrelay: @proteneer  assert or use their settings?

jkausrelay: Not sure, but I'll add a warning

PR #861, opened 2022-09-30
Add reference distance with 4D offset, clean nonbonded reference potentials

mcwitt:
In preparation for switching to specifying 4-d coordinates for decoupling as parameters,

- renames `jax_utils.distance` -> `jax_utils.pairwise_distances` to be more specific
- updates JAX reference code to accept 4-d coordinates as a separate argument
- removes hack previously used to simulate an aperiodic 4th dimension via using a box that is ~infinite in the last dimension
- removes helper functions related to 3-d -> 4-d conversion previously used to implement the above, now unused

### DIFF
--- a/timemachine/potentials/nonbonded.py
+++ b/timemachine/potentials/nonbonded.py
@@ -9,11 +9,12 @@ from typing_extensions import TypeAlias
 
 from timemachine.potentials import jax_utils
 from timemachine.potentials.jax_utils import (
+    compute_lifting_parameter,
     convert_to_4d,
     delta_r,
-    distance,
     distance_on_pairs,
     pairs_from_interaction_groups,
+    pairwise_distances,
 )
 
 Array: TypeAlias = NDArray
@@ -223,10 +224,9 @@ def nonbonded(
 
     eps_i = jnp.expand_dims(eps, 0)
     eps_j = jnp.expand_dims(eps, 1)
-
     eps_ij = combining_rule_epsilon(eps_i, eps_j)
 
-    dij = distance(conf, box)
+    dij = pairwise_distances(conf, box)
 
     keep_mask = jnp.ones((N, N)) - jnp.eye(N)
     keep_mask = jnp.where(eps_ij != 0, keep_mask, 0)
--- a/timemachine/potentials/nonbonded.py
+++ b/timemachine/potentials/nonbonded.py
@@ -10,7 +10,6 @@ from typing_extensions import TypeAlias
 from timemachine.potentials import jax_utils
 from timemachine.potentials.jax_utils import (
     compute_lifting_parameter,
-    convert_to_4d,
     delta_r,
     distance_on_pairs,
     pairs_from_interaction_groups,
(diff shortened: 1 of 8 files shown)


maxentile (approved):

timemachine/potentials/nonbonded.py
```diff
@@ -270,7 +256,14 @@ def nonbonded(
...
 
 def nonbonded_on_specific_pairs(
-    conf, params, box, pairs, beta: float, cutoff: Optional[float] = None, rescale_mask=None
+    conf,
+    params,
+    box,
+    pairs,
+    beta: float,
+    cutoff: Optional[float] = None,
+    w_coords: Optional[Array] = None,
```
maxentile, line 265: Optional clarification: In docstring for pairlist variant indicate (1) whether `w_coords` has length `conf` or length `params`, (2) if `w_coords` must be non-negative.

mcwitt: I tried to follow the convention that `w_coords` is per atom, `w_offsets` is per interaction. Will document!

mcwitt: Added clarifying comments in 9afae5b

maxentile: 👍 -- this convention makes sense, and thanks for adding the comments!

maxentile: Actually, follow-up question: Should the pairlist variant be modified from accepting per-atom `w_coords` to accepting per-interaction `w_offsets`?

(For the other ones (all-pairs and interaction-group) `w_coords` seems like the natural input format, but this pairlist one looks like the right place to possibly allow more fine-grained control.)

mcwitt: > Should the pairlist variant be modified from accepting per-atom w_coords to accepting per-interaction w_offsets?

I think I see the utility of this, but would it make sense to wait to implement this capability for the GPU and reference potentials at the same time? (right now the GPU `NonbondedPairList` only supports per-atom 4-d coordinates)

maxentile: Oh, right! Sorry, I had overlooked that -- makes sense to do both consistently, not in this PR.

proteneer: It's a little tricky, currently the `nonbonded-pairlist` is only used for exclusions (via cancellation of nans), we use the `nonbonded-pairlist-precomputed` variant for actual intramolecular nonbonded interactions. Right now the `nonbonded-pairlist` and `all-pairs` form the `Nonbonded` class on the GPU.

To be clear:

Suppose we have a butane with 4 carbons, `c1-c2-c3-c4`, we have a single 1-4 exclusion term at 50% of the nonbonded ixn nrg (scale factors set to 0.5). What currently happens is with the `c1-c4` ixn is:

1) `c1-c4` is computed using `all-pairs`, with 100% ixn nrg, with whatever per-atom lifting parameters are present. 
2) `c1-c4` is computed a second time using the `nonbonded-pairlist`, again with -100% ixn nrg, with whatever per-atom lifting parameters are present to cancel out 1)
3) `c1-c4` is computed a third time using the `nonbonded-pairlist-precomputed` at 50% of the ixn nrg, with a per-interaction scale factor. 

So for consistency, the `nonbonded-pairlist` is set using `w_coords` with shape `(num_atoms, 3)` since its meant to be used as exclusions, and `nonbonded-pairlist-precomputed` is set using `w_offsets` with shape `(num_pairs_of_ixns,)`

proteneer: (unresolving to make sure that @maxentile and @mcwitt - feel free to resolve again)

proteneer (approved):
lgtm!

timemachine/potentials/nonbonded.py
proteneer: note that OpenMM has the following convention:

1) exclusions are fully removed from the neighborlist when the interacting tiles are built up
2) exceptions are re-computed and the interaction strength is set to whatever the scale factors are

PR #862, opened 2022-09-30
Deprecate old single topology class.

proteneer:
This PR removes the old SingleTopology class.

The new SingleTopologyV3 now extends a AtomMapMixin, for convenience when using the pdb_writer. This is because the constructor does a fair bit more work since it sets up the end-state systems. The performance benchmark tests also now default to single topology as opposed to dual topology.

I've also eagerly removed various tests dealing with parameter interpolation as well, since they will be removed by @mcwitt soon.

Note for @jkausrelay the testsystems in this PR still return a SingleTopologyV3 object - but this can be easily resolved (regardless of the order of the merge).

Done:
-SingleTopologyV3 renamed to SingleTopology
-fe/single_topology_v3.py renamed to fe/single_topology.py


proteneer:
> I'm curious, is there an advantage to using a mixin here versus modifying the SingleTopologyV3 constructor to take an AtomMap instead of mol_a, mol_b, core? For the uses I'm imagining the latter seems a bit simpler (though not a huge difference).

I wanted to make minimal changes to the constructor signature (and avoid breaking the API if possible). I think when doing refactoring stuff like this its best to keep scope of API changes as narrow as possible.

tests/test_free_energy.py
```diff
@@ -223,18 +234,28 @@ def low_dim_f(perturb: float) -> float:
...
-    unbound_potentials, sys_params, coords, _ = get_relative_hif2a_in_vacuum()
-
+    mol_a, mol_b, core = get_hif2a_ligand_pair_single_topology()
+    forcefield = Forcefield.load_from_file(DEFAULT_FF)
+    st = SingleTopology(mol_a, mol_b, core, forcefield)
+    vac_sys = st.setup_intermediate_state(0.5)
+    x_a = utils.get_romol_conf(st.mol_a)
+    x_b = utils.get_romol_conf(st.mol_b)
+    coords = st.combine_confs(x_a, x_b)
+    box = np.eye(3) * 100
```
jkausrelay, line 249: duplicate box

tests/test_free_energy.py
```diff
@@ -223,18 +234,28 @@ def low_dim_f(perturb: float) -> float:
         assert "box" in str(e).lower()
 
 
+#
```
jkausrelay, line 237: nit: remove #

tests/test_benchmark.py
```diff
@@ -252,11 +289,30 @@ def benchmark_dhfr(verbose=False, num_batches=100, steps_per_batch=1000):
     )
 
 
+def prepare_hif2a_initial_states(st, host_system, host_coords, host_box):
+    st = rbfe.SingleTopology(st.mol_a, st.mol_b, st.core, st.ff)
+    host_config = rbfe.HostConfig(host_system, host_coords, host_box)
+    temperature = 300.0
+    lamb = 0.1
+    initial_state = rbfe.setup_initial_states(st, host_config, temperature, [lamb], seed=2022)[0]
```
mcwitt, line 297: nit: should this name be pluralized?
```suggestion
    initial_states = rbfe.setup_initial_states(st, host_config, temperature, [lamb], seed=2022)[0]
```

mcwitt: sorry, I misread. Deleting this comment..

tests/test_benchmark.py
```diff
@@ -252,11 +289,30 @@ def benchmark_dhfr(verbose=False, num_batches=100, steps_per_batch=1000):
     )
 
 
+def prepare_hif2a_initial_states(st, host_system, host_coords, host_box):
```
mcwitt, line 292: nit: should the name of this function be singular?

timemachine/fe/single_topology.py
```diff
@@ -800,6 +785,48 @@ def __init__(self, mol_a, mol_b, core, forcefield, lambda_angles=0.4, lambda_tor
...
+            For ring opening/closing transformations, alchemical parameter values controlling the intervals over which
+            bonds, angles, and torsions are interpolated. Note that these have no effect on terms not involved in ring
+            opening/closing.
+
+            - Bonds are interpolated in the interval [0, lambda_angles]
+            - Angles are interpolated in the interval [lambda_angles, lambda_torsions]
+            - Torsions are interpolated in the interval [lambda_torsions, 1]
+        """
+        # initialize the mixin to get the a_to_c, b_to_c, c_to_a, c_to_b, and c_flags
+        AtomMapMixin.__init__(self, mol_a, mol_b, core)
```
jkausrelay, line 820: nit: `super().__init__(self, ...)`

tests/test_benchmark.py
```diff
@@ -378,16 +418,15 @@ def test_nonbonded_interaction_group_potential(hi2fa_test_frames):
...
 
 def test_nonbonded_potential(hi2fa_test_frames):
-    pots, sys_params, frames, boxes, _ = hi2fa_test_frames
-    U_names = [type(U_fn).__name__ for U_fn in pots]
-    nb_idx = U_names.index("Nonbonded")
-    nonbonded_pot = pots[nb_idx]
-    assert isinstance(nonbonded_pot, Nonbonded)
+    pots, frames, boxes, _ = hi2fa_test_frames
+
+    nonbonded_pot = pots[-1]
```
mcwitt, line 423: nit: probably safer not to assume an ordering here? Maybe
```suggestion
    nonbonded_pot = next(p for p in pots if isinstance(p, Nonbonded))
```

timemachine/fe/free_energy.py
```diff
@@ -245,7 +245,7 @@ def _combine(self, ligand_values, host_values=None):
 
 # this class is serializable.
 class RelativeFreeEnergy(BaseFreeEnergy):
-    def __init__(self, top: Union[topology.SingleTopology, topology.DualTopology], label=None, complex_path=None):
+    def __init__(self, top: topology.DualTopology, label=None, complex_path=None):
```
mcwitt, line 248: It looks like there's still code in this class to deal with `SingleTopology` (L345). Can that be removed, or should we continue to accept `Union[SingleTopology, DualTopology]` here?

timemachine/potentials/generic.py
```diff
@@ -22,6 +23,7 @@
 BondedGpuPotential = TypeVar("BondedGpuPotential", bound=gpu.BondedWrapper)
 
 
+@typing_extensions.runtime
```
mcwitt, line 26: This seems to be a [deprecated alias]((https://github.com/python/typing_extensions/blob/2979419f7bb10b554102e972351f8d03c8ffd1fa/src/typing_extensions.py#L635)) for `runtime_checkable`. Should we use the latter instead?

timemachine/fe/free_energy.py
```diff
@@ -342,7 +342,7 @@ def _combine(self, ligand_values_a, ligand_values_b, host_values=None):
         -------
             combined_values
         """
-        if isinstance(self.top, topology.SingleTopology):
+        if isinstance(self.top, single_topology.SingleTopology):
             ligand_values = [np.mean(self.top.interpolate_params(ligand_values_a, ligand_values_b), axis=0)]
```
mcwitt, line 346: Can this branch be removed? According to the constructor annotation, `self.top` should be `DualTopology`.

tests/test_free_energy.py
proteneer: willfix

tests/test_free_energy.py
proteneer: willfix

tests/test_benchmark.py
proteneer: willfix

timemachine/fe/single_topology.py
proteneer: willfix

tests/test_benchmark.py
proteneer: good idea, willfix inline

timemachine/potentials/generic.py
proteneer: willfix

timemachine/fe/free_energy.py
proteneer: good call, willfix.

PR #865, opened 2022-10-04
Handle `box=None` in `pairwise_distances`, add tests

mcwitt:
* Fixes regression introduced in 95dc14a32cb7c07b2a02d3fbd72811f0263c130d where passing `box=None` fails with a `TypeError` (this affects nightly tests)
* Refines assertion on shape of `w`
* Adds tests for `jax_utils.pairwise_distances`

### DIFF
--- a/tests/test_jax_utils.py
+++ b/tests/test_jax_utils.py
@@ -195,5 +195,5 @@ def test_pairwise_distances_periodic_lifting(coords_box_w):
     assert (dij >= dij0).all()
 
     # guard assertion to prevent spurious failure due to roundoff error
-    if np.std(w) / (1.0 + np.std(x)) > 1e-6:
+    if np.std(w) / (1.0 + np.linalg.norm(x)) > 1e-6:
         assert dij.sum() > dij0.sum()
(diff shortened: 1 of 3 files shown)


maxentile (approved):

tests/test_jax_utils.py
```diff
@@ -133,3 +139,57 @@ def d(conf, pairs, box):
...
+    w = draw(arrays(np.float64, n, elements=finite_floats(-1e6, 1e6)))
+    return x, box_diag, w
+
+
+@given(coords_box_w_triples())
+@seed(2022)
+def test_pairwise_distances_periodic(coords_box_w):
+    x, box_diag, _ = coords_box_w
+    n, _ = x.shape
+    dij = pairwise_distances(x, np.diagflat(box_diag))
```
badisa, line 181: Not sure I understand hypothesis well enough, but is this always creating a box small enough to impose PBCs?

badisa (approved):

tests/test_jax_utils.py
mcwitt: The coordinates and box are uncorrelated, so this should test some examples where we expect wrapping and some where we don't.

This got me thinking that it might be good practice to add explicit examples of cases that we want to ensure are tested. E.g. I think there's some (probably vanishingly small) probability that the arbitrary examples don't include a case with a small enough box to test PBCs. I added an explicit example in 2a657ab.

PR #868, opened 2022-10-04
include protein and water model in ff

jkausrelay:
- Update ff to add protein / water model attributes
- Update builders to take ff
- New attributes will be serialized out
- Prefer ff.serialize() to serialize_handlers(ff.get_ordered_handlers())


proteneer:
1) should we also rename DEFAULT_FF to DEFAULT_LIGAND_FF
2) are we hardcoding the water/protein format to xml?

tests/test_forcefields.py
```diff
@@ -14,7 +15,10 @@
 
 def test_serialization_of_ffs():
     for path in glob("timemachine/ff/params/smirnoff_*.py"):
-        ff = Forcefield(deserialize_handlers(open(path).read()))
+        handlers, protein_ff, water_model = deserialize_handlers(open(path).read())
+        ff = Forcefield(handlers, protein_ff=protein_ff, water_model=water_model)
+        assert ff.protein_ff == constants.DEFAULT_PROTEIN_FF
+        assert ff.water_model == constants.DEFAULT_WATER_MODEL
```
proteneer, line 21: should we rename this to `DEFAULT_WATER_FF` to be consistent?

jkausrelay:
> should we also rename DEFAULT_FF to DEFAULT_LIGAND_FF
Yes, this makes sense.

> are we hardcoding the water/protein format to xml

In the ff file it's just the name (amber14, tip3p, ...). For now that is the name of the xml file, but I don't see why that can't be changed in the future.

proteneer (approved):
lgtm, mostly minor comments.

timemachine/md/builders.py
```diff
@@ -19,7 +21,8 @@ def build_protein_system(host_pdbfile):
         PDB of the host structure
 
     """
-    host_ff = app.ForceField("amber99sbildn.xml", "tip3p.xml")
+
+    host_ff = app.ForceField(f"{tm_ff.protein_ff}.xml", f"{tm_ff.water_model}.xml")
```
badisa, line 25: Should the users have to add the extension?

jkausrelay:
>>should we also rename DEFAULT_FF to DEFAULT_LIGAND_FF
>Yes, this makes sense.

On second thought, I might just leave it as is. The file can have the protein/water ffs added to it now, so it could be considered the full ff, not just for the ligand.

badisa (approved):

timemachine/md/builders.py
jkausrelay: The extension seems like an implementation detail. We may switch formats if we start fitting the water/protein models.

timemachine/ff/handlers/deserialize.py
```diff
@@ -20,8 +21,19 @@ def deserialize_handlers(obj):
...
     handlers = []
 
+    protein_ff = constants.DEFAULT_PROTEIN_FF
+    water_ff = constants.DEFAULT_WATER_FF
+
     for k, v in obj_dict.items():
 
+        if k == constants.PROTEIN_FF_TAG:
+            protein_ff = v
+            continue
```
mcwitt, line 31: nit: is this equivalent to
```python
protein_ff = obj_dict.get(constants.PROTEIN_FF_TAG)
```
outside of the loop?

timemachine/md/builders.py
```diff
@@ -50,16 +55,16 @@ def build_protein_system(host_pdbfile):
     return solvated_host_system, solvated_host_coords, nwa, nha, box, modeller.topology
 
 
-def build_water_system(box_width):
-    ff = app.ForceField("tip3p.xml")
+def build_water_system(box_width, tm_ff: Forcefield):
+    ff = app.ForceField(f"{tm_ff.water_ff}.xml")
```
mcwitt, line 59: nit: seems like it might be more precise to pass in just `water_ff` and `sanitized_water_ff` here instead of the full `Forcefield` object

timemachine/ff/handlers/deserialize.py
jkausrelay: yup, will fix

badisa: Wouldn't this need to be `.pop` else it will fail on ln 52?

timemachine/md/builders.py
jkausrelay: updated

PR #869, opened 2022-10-04
Moves RABFE code to attic

badisa:


proteneer:
are we able to clean up the `DualTopology`, `*Conversion`, `*Decouple` classes from `fe/topology.py` as well?

proteneer (approved):
lgtm!

PR #872, opened 2022-10-05
Replace forcefield parameter lists with dataclass, attic unused optimization code

mcwitt:
- Quality of life improvement adding `ForcefieldParams` dataclass to replace lists of forcefield parameters with conventional ordering of elements. This should remove potential confusion about the ordering and improve the quality of static checks.
- Moves unused code from `timemachine/optimize/` to `attic/`


tests/test_topology.py
```diff
@@ -25,22 +25,15 @@ def test_relative_free_energy_forcefield():
...
-    ff1_params = ff1.get_ordered_params()
-
-    combined_params = combine_ordered_params(ff0, ff1)
+    ff0_params = ff0.get_params()
+    ff1_params = ff1.get_params()
+
+    combined_params = combine_params(ff0_params, ff1_params)
     combined_qlj_params, combined_ubp = fftop.parameterize_nonbonded(
-        combined_params[charge_idx], combined_params[lj_idx]
+        combined_params.q_params, combined_params.lj_params
```
mcwitt, line 33: Naming question: `ff_params.q_params` seems a bit redundant, but `ff_params.q` is a bit weird too (though maybe better?)

tests/test_dummy.py
```diff
@@ -625,7 +625,7 @@ def test_parameterize_and_draw_ixns():
     """
 
     ff_handlers = deserialize_handlers(open("timemachine/ff/params/smirnoff_1_1_0_sc.py").read())
-    ff = Forcefield(ff_handlers)
+    ff = Forcefield.from_handlers(ff_handlers)
```
mcwitt, line 628: This is a breaking API change; I didn't immediately see any usage of the `Forcefield.__init__` outside of the main repo, but might be good to double-check

timemachine/md/minimizer.py
```diff
@@ -235,23 +231,18 @@ def equilibrate_host(
...
-        (hgt.parameterize_harmonic_bond, [ff.hb_handle]),
-        (hgt.parameterize_harmonic_angle, [ff.ha_handle]),
-        (hgt.parameterize_periodic_torsion, [ff.pt_handle, ff.it_handle]),
-        (hgt.parameterize_nonbonded, [ff.q_handle, ff.lj_handle]),
+    ff_params = ff.get_params()
+    hb_params, hb_potential = hgt.parameterize_harmonic_bond(ff_params.hb_params)
+
+    params_potential_pairs = [
+        (hb_params, hb_potential),
+        hgt.parameterize_harmonic_angle(ff_params.ha_params),
```
maxentile, line 239: Curiosity: is the different syntax for harmonic bonds in this list significant?

(Is this mostly to document the tuple contents of the subsequent ones?)

tests/test_topology.py
```diff
@@ -25,22 +25,15 @@ def test_relative_free_energy_forcefield():
...
     )
-    qlj0_params, ubp0 = bt0.parameterize_nonbonded(ff0_params[charge_idx], ff0_params[lj_idx])
-    qlj1_params, ubp1 = bt1.parameterize_nonbonded(ff1_params[charge_idx], ff1_params[lj_idx])
+    qlj0_params, ubp0 = bt0.parameterize_nonbonded(ff0_params.q_params, ff0_params.lj_params)
+    qlj1_params, ubp1 = bt1.parameterize_nonbonded(ff1_params.q_params, ff1_params.lj_params)
```
maxentile, line 36: Nice

maxentile (approved):

jkausrelay:
@mcwitt Overall lgtm, but will conflict with #868

timemachine/md/minimizer.py
mcwitt: This is just to have a named reference to `hb_potential`, since it's used again a few lines below to get the bond list.

tests/test_topology.py
mcwitt: > this can contain either "generalizable" forcefield params (of length # types) or "system params" (of length # atoms or # interactions)

I didn't (intentionally) change the representation of system params in this PR; `ForcefieldParams` was intended to only be used for the former kind. The addition of the new dataclass instead of adding to `Forcefield` was to have a type that is 1-1 with the existing representation as a list of parameter arrays. `ForcefieldParams` is generic because sometimes (e.g. in `RelativeFreeEnergyForcefield`) the parameters are tuples of (src, dst).

mcwitt:
@jkausrelay I merged in the changes from #868. Summary of changes:

- Updated `from_handlers` to accept `protein_ff` and `water_ff`
- Switched from properties for `protein_ff`, `water_ff` to dataclass fields
- Set `frozen=True` on the dataclass so `protein_ff`, `water_ff` are still read-only (as well as the handlers, but I haven't seen a use case so far for assigning to these after ff creation)

jkausrelay (approved):

PR #873, opened 2022-10-06
Refactors get_val_and_grad_fn to take bound potentials

badisa:
* Makes memory management easier
* Reduces allocating 2x potentials for RBFE sample method

### DIFF
--- a/timemachine/md/minimizer.py
+++ b/timemachine/md/minimizer.py
@@ -11,7 +11,6 @@ from timemachine.fe.utils import get_romol_conf
 from timemachine.ff import Forcefield
 from timemachine.ff.handlers import openmm_deserializer
 from timemachine.lib import LangevinIntegrator, MonteCarloBarostat, custom_ops
-from timemachine.lib.potentials import SummedPotential
 from timemachine.md.barostat.utils import get_bond_list, get_group_indices
 from timemachine.md.fire import fire_descent
 
@@ -284,17 +283,28 @@ def equilibrate_host(
 def get_val_and_grad_fn(bps, box, lamb):
     """
     Convert impls, box, lamb into a function that only takes in coords.
+
+    Parameters
+    ----------
+    bps: List of BoundPotentials
+
+    box: np.array (3,3)
+
+    lamb: float
+
+    Returns
+    -------
+    Energy function with gradient
+        f: R^(Nx3) -> (R^1, R^Nx3)
     """
 
-    params = [np.array(bp.params) for bp in bps]
-    flat_params = np.concatenate([p.reshape(-1) for p in params])
-    sum_potential = SummedPotential(bps, params)
-    sum_potential.bind(flat_params)
-
-    impl = sum_potential.bound_impl(np.float32)
-
     def val_and_grad_fn(coords):
-        g, _, u = impl.execute(coords, box, lamb)
+        g = np.zeros_like(coords)
+        u = 0.0
+        for impl in bps:
+            g_bp, _, u_bp = impl.execute(coords, box, lamb)
+            g += g_bp
+            u += u_bp
         return u, g
 
     return val_and_grad_fn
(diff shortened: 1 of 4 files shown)


timemachine/md/minimizer.py
```diff
@@ -284,17 +283,28 @@ def equilibrate_host(
 def get_val_and_grad_fn(bps, box, lamb):
     """
     Convert impls, box, lamb into a function that only takes in coords.
-    """
 
-    params = [np.array(bp.params) for bp in bps]
-    flat_params = np.concatenate([p.reshape(-1) for p in params])
-    sum_potential = SummedPotential(bps, params)
-    sum_potential.bind(flat_params)
```
mcwitt, line 292: Would it work to still use the custom_ops `SummedPotential` here with a more minimal change?

```python
    ubps = [bp.potential for bp in bps]
    params = [np.array(bp.params) for bp in bps]
    flat_params = np.concatenate([p.reshape(-1) for p in params])
    impl = custom_ops.SummedPotential(ubps, [len(p) for p in flat_params])
```

proteneer: This does introduce more overhead though from the IO point of view, but it's probably insignificant relative to the cost of the nonbonded interactions. 

> The params are not available directly from the `C++` bound potentials. 

More importantly, remember that `SummedPotential` takes in `shared_ptr` to unbound potentials. I think the ownership semantic gets a bit more nuanced if they get wrapped. (And we still need a list of bound_impls for other calls.)

proteneer: Actually - one reason against this style change is that now, the subsequent calls in production will use "dirty" bound_impls that have already been used, eg. neighborlist has been pre-constructed, etc., during minimization.

In theory it should be safe to re-use them for equilibration/production even though they were used for local minimization but it's still a minor risk. We should definitely triple check this.

timemachine/md/minimizer.py
```diff
@@ -309,7 +319,7 @@ def local_minimize(x0, val_and_grad_fn, local_idxs, verbose=True):
     x0: np.array (N,3)
         Coordinates
 
-    val_and_grad_fn: f: R^(Nx3,3x3) -> (R^1, R^Nx3)
```
proteneer, line 312: good call

timemachine/md/minimizer.py
badisa: > Actually - one reason against this change is that now, the subsequent calls in production will use "dirty" bound_impls that have already been used, eg. neighborlist has been pre-constructed, etc., during minimization.
> 
> In theory it should be safe to re-use them for equilibration/production even though they were used for local minimization but it's still a minor risk. We should definitely triple check this.

Will put together a test for that.

Though presumably we have this same problem with Pair bar? We take a set of potentials, compute du_dx, du_dl and energies (build nblist) for x0, then compute du_dx, du_dl and energies for x1.

proteneer: yes - absolutely. I think its still safe to merge, but just wanted to note the change!

mcwitt: > The params are not available directly from the bound potentials.

Oh right, but couldn't this work with a slight change to the signature to

```python
def get_val_and_grad_fn(ubps: List[custom_ops.Potential], params, box, lamb):
```
?

proteneer (approved):
lgtm - I think this is safe to merge!

timemachine/md/minimizer.py
badisa: @proteneer makes a good point about the ownership once you do `[bound.get_potential() for bound in bound_pots]`. 

I would argue if there is utility in the summed potential, it should be passed in as a list with the only element of SummedPotential. This is still deterministic, even if the sums aren't done in fixed point.

mcwitt: > once you do [bound.get_potential() for bound in bound_pots].

(sure, but should be OK as long as you have a reference from Python. It's the same issue you'd run into passing in `[p.bound_impl(np.float32) for p in initial_state.potentials]` anonymously)

> I would argue if there is utility in the summed potential, it should be passed in as a list with the only element of SummedPotential

To me that suggests simplifying `get_value_and_grad_fn` to only accept a single potential (which could be a sum)?

Not blocking, but it seems preferable to me to use `custom_ops.SummedPotential` here, since

1. This will be run inside of a hot optimization loop (maybe only slightly preferable, since the overhead may amortize for large systems)
2. It might be useful for issues with clashes leading to fixed-point overflow to be exposed in the initial optimization, rather than later on in MD. (On the other hand, maybe this is a reason to prefer a floating-point sum here, if we think it's likely that the initial state will be clash-ier than the final)

PR #874, opened 2022-10-06
Removes try/catch from context

badisa:
* Uses DeviceBuffers instead

### DIFF
--- a/timemachine/cpp/src/context.cu
+++ b/timemachine/cpp/src/context.cu
@@ -71,8 +71,8 @@ Context::multiple_steps(const std::vector<double> &lambda_schedule, int store_du
     std::unique_ptr<DeviceBuffer<unsigned long long>> d_du_dl_buffer(nullptr);
     if (du_dl_buffer_size > 0) {
         d_du_dl_buffer.reset(new DeviceBuffer<unsigned long long>(du_dl_buffer_size));
+        gpuErrchk(cudaMemsetAsync(d_du_dl_buffer->data, 0, d_du_dl_buffer->size, stream));
     }
-    gpuErrchk(cudaMemsetAsync(d_du_dl_buffer->data, 0, d_du_dl_buffer->size, stream));
 
     intg_->initialize(bps_, lambda_schedule[0], d_x_t_, d_v_t_, d_box_t_, stream);
     for (int i = 1; i <= lambda_schedule.size(); i++) {
--- a/timemachine/cpp/src/context.cu
+++ b/timemachine/cpp/src/context.cu
@@ -29,12 +29,6 @@ Context::Context(
     cub::DeviceReduce::Sum(d_sum_storage_, d_sum_storage_bytes_, d_in_tmp, d_out_tmp, N_);
     gpuErrchk(cudaPeekAtLastError());
     gpuErrchk(cudaMalloc(&d_sum_storage_, d_sum_storage_bytes_));
-
-    // for(int i=0; i < bps.size(); i++) {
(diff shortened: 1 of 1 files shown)


timemachine/cpp/src/context.cu
```diff
@@ -74,65 +61,65 @@ Context::multiple_steps(const std::vector<double> &lambda_schedule, int store_du
...
-            }
+    std::unique_ptr<DeviceBuffer<double>> d_box_buffer(nullptr);
+    if (box_buffer_size > 0) {
+        d_box_buffer.reset(new DeviceBuffer<double>(box_buffer_size));
+    }
+    std::unique_ptr<DeviceBuffer<unsigned long long>> d_du_dl_buffer(nullptr);
+    if (du_dl_buffer_size > 0) {
+        d_du_dl_buffer.reset(new DeviceBuffer<unsigned long long>(du_dl_buffer_size));
+    }
+    gpuErrchk(cudaMemsetAsync(d_du_dl_buffer->data, 0, d_du_dl_buffer->size, stream));
```
badisa, line 75: This is wrong

timemachine/cpp/src/context.cu
```diff
@@ -74,65 +61,65 @@ Context::multiple_steps(const std::vector<double> &lambda_schedule, int store_du
...
-            }
+    std::unique_ptr<DeviceBuffer<double>> d_box_buffer(nullptr);
+    if (box_buffer_size > 0) {
+        d_box_buffer.reset(new DeviceBuffer<double>(box_buffer_size));
+    }
+    std::unique_ptr<DeviceBuffer<unsigned long long>> d_du_dl_buffer(nullptr);
+    if (du_dl_buffer_size > 0) {
+        d_du_dl_buffer.reset(new DeviceBuffer<unsigned long long>(du_dl_buffer_size));
+        gpuErrchk(cudaMemsetAsync(d_du_dl_buffer->data, 0, d_du_dl_buffer->size, stream));
+    }
```
mcwitt, line 75: Just noticing that we could avoid `unique_ptr` and dynamic allocation here except that `DeviceBuffer` currently has a check that fails if the size is zero. Would it be worth relaxing that restriction? (conflicts with e.g. the behavior of `malloc` AFAIK)

badisa: I would be for removing the check as it is consistent with cudaMalloc's behavior (nullptr if size is 0). Last time this came up though we decided against it, though I can't find a reference to that conversation.

Easily done in this PR and would simplify a fair amount of code.

mcwitt: To be clear, I wouldn't argue for using a literal `DeviceBuffer(0)` in cases where we don't want to allocate (I think the discussion starting [here](https://github.com/proteneer/timemachine/pull/726#discussion_r853317229) was more about that case)

mcwitt (approved):
LGTM. Nice cleanup!

proteneer:
hopefully once @mcwitt finishes the du_dl item and we remove lamb from the signature of U, we can completely remove d_du_dl  from these context methods!

timemachine/cpp/src/context.cu
proteneer: 1) I agree that wrapping a `DeviceBuffer` in a `unique_ptr` is unwieldy.
2) Do we actually have any practical situations where `box_buffer_size`, and `du_dl_buffer_size` is zero? If not, I'd just turn assert that those two quantities are `> 0` and simplify.

proteneer (approved):

PR #875, opened 2022-10-11
Use separate Python API type for potentials bound to parameters

mcwitt:
Small quality of life change intended to reduce confusion over whether a `CustomOpWrapper` is expected to be bound to parameters (i.e. whether we previously called `.bind(params)`).

For example, forgetting to call `bind` before `bound_impl`

```python
HarmonicBond(bond_idxs).bound_impl(np.float32)
```

will now fail to typecheck (in addition to failing at runtime).


mcwitt:
Opened prematurely; the current approach will not work. Deferring until larger refactoring of potential API.

PR #877, opened 2022-10-12
Remove unused restraint fxn

maxentile:
Moves `restraint` function to attic
Deletes an unused reference to this function from `tests/common.py`


tests/common.py
```diff
@@ -150,29 +150,6 @@ def prepare_nb_system(x, E, lambda_plane_idxs, lambda_offset_idxs, p_scale, cuto
...
-    params = np.random.randn(B, 3).astype(np.float64)
-
-    bond_idxs = []
-    for _ in range(B):
-        bond_idxs.append(np.random.choice(atom_idxs, size=2, replace=False))
-    bond_idxs = np.array(bond_idxs, dtype=np.int32)
-
-    lambda_flags = np.random.randint(0, 2, size=(B,)).astype(np.int32)
-
-    custom_restraint = potentials.Restraint(bond_idxs, params, lambda_flags, precision=precision)
```
maxentile, line 170: (Note: the test function to be removed was also broken at this line -- the reference to `potentials.Restraint` does not resolve, so the reference to `bonded.restraint` (removed in this PR) would not currently be reached if `prepare_restraints` were called.)

attic/modules/potentials/bonded.py
```diff
@@ -0,0 +1,54 @@
...
+    Compute the harmonic bond energy given a collection of molecules.
+
+    This implements a harmonic bond potential:
+        V(conf; lamb) = \sum_bond kbs[bond] term[bond]^2
+
+        where term[bond] = 1 - exp(-a0s[bond]*(dij[bond] - b0s[bond]))
+        and where
+            dij[bond] = distance[bond] + f_lambda^2
+            and where
+                f_lambda = lamb * lamb_flags
```
maxentile, line 14: Main motivation for removal: This dependence on lamb and lamb_flags is a bit different from that of other `bonded` potentials.

(Changing the signatures of the other `bonded` potentials will be marginally simpler if this potential doesn't have to be updated also.)

mcwitt (approved):

PR #880, opened 2022-10-13
Reduces memory usage in RBFE

badisa:
* Minor clean up
* Reuse coords already in memory

### DIFF
--- a/timemachine/fe/rbfe.py
+++ b/timemachine/fe/rbfe.py
@@ -126,7 +126,7 @@ def sample(initial_state, protocol):
     ctxt.multiple_steps_U(
         lamb=initial_state.lamb,
         n_steps=protocol.n_eq_steps,
-        lambda_windows=[initial_state.lamb],
+        lambda_windows=[],
         store_u_interval=0,
         store_x_interval=0,
     )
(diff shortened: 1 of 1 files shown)


timemachine/fe/rbfe.py
```diff
@@ -408,12 +412,10 @@ def estimate_free_energy_given_initial_states(initial_states, protocol, temperat
 
             # loop over bond, angle, torsion, nonbonded terms etc.
             for u_idx, (prev_U_fn, cur_U_fn) in enumerate(zip(prev_batch_U_fns, cur_batch_U_fns)):
-                x_cur = (cur_frames, cur_boxes)
```
badisa, line 411: `x_cur` and `x_prev` stored references to frames that wouldn't be cleared until the next loop and could hold onto the previous frames even after having sampled the new current frames.

timemachine/fe/rbfe.py
```diff
@@ -135,18 +135,18 @@ def sample(initial_state, protocol):
...
+        lambda_windows=[],
+        store_u_interval=0,
         store_x_interval=protocol.steps_per_frame,
     )
 
     assert all_coords.shape[0] == protocol.n_frames
     assert all_boxes.shape[0] == protocol.n_frames
 
-    assert np.all(np.isfinite(ctxt.get_x_t())), "Production resulted in a nan"
+    assert np.all(np.isfinite(all_coords[-1])), "Production resulted in a nan"
```
badisa, line 149: This avoids allocating a new array, just use the coordinates that were returned.

timemachine/fe/rbfe.py
```diff
@@ -135,18 +135,18 @@ def sample(initial_state, protocol):
 
     # a crude, and probably not great, guess on the decorrelation time
     n_steps = protocol.n_frames * protocol.steps_per_frame
-    all_nrgs, all_coords, all_boxes = ctxt.multiple_steps_U(
+    _, all_coords, all_boxes = ctxt.multiple_steps_U(
         lamb=initial_state.lamb,
         n_steps=n_steps,
-        lambda_windows=[initial_state.lamb],
```
badisa, line 141: Avoids allocating an array on the GPU and avoids evaluating energies that we never use.

jkausrelay (approved):
nice!

timemachine/fe/rbfe.py
```diff
@@ -393,7 +393,11 @@ def estimate_free_energy_given_initial_states(initial_states, protocol, temperat
     ukln_by_component_by_lambda = []
 
     for lamb_idx, initial_state in enumerate(initial_states):
-
+        # Clear any old references to avoid holding on to objects in memory we don't need.
+        cur_frames = None
+        cur_boxes = None
+        bound_impls = None
+        cur_batch_U_fns = None
```
mcwitt, line 400: nit: alternative (maybe slightly clearer in intent)

```python
        del cur_frames, cur_boxes, bound_impls, cur_batch_U_fns
```

at the end of loop body

timemachine/fe/rbfe.py
mcwitt: Good catch! I think `del x_cur, x_prev` at the end of the loop body would also solve this?

PR #883, opened 2022-10-14
Steric clash detection

jkausrelay:
- Add a simple function to detect (self) steric clashes based on the frc
- Clean up duplicate plot_atom_mapping_grid fn


timemachine/fe/model_utils.py
```diff
@@ -1,14 +1,58 @@
...
+
+    def val_and_grad_fn(x):
+        return U(x), grad_fn(x)
+
+    return val_and_grad_fn
+
+
+def get_clashing_atoms(mol: Chem.Mol, ff: Forcefield, max_force: Optional[float] = 50000) -> List[float]:
+    """
+    Return a list of atom indices that are clashing based on the max_force.
```
proteneer, line 43: note that this doesn't *just* find clashes, since vacuum_val_and_grad_fn also can include bond terms (which can detect strains, weird angles/torsions) - maybe rename to get_strained_atoms?

timemachine/fe/model_utils.py
```diff
@@ -1,14 +1,58 @@
...
+    return atoms == mol_copy.GetNumAtoms()
+
+
+def assert_mol_has_all_hydrogens(mol: Chem.Mol):
+    assert mol_has_all_hydrogens(mol), "Hydrogens missing for mol"
+
+
+def get_vacuum_val_and_grad_fn(mol: Chem.Mol, ff: Forcefield):
+    """
+    Return a function which returns the potential energy and frcs
```
proteneer, line 26: du_dx, not frcs

timemachine/fe/model_utils.py
maxentile: Would it be convenient to return one list of atom indices per force component,

```text
strained atoms (HarmonicBond)    : [],
strained atoms (HarmonicAngle)   : [],
strained atoms (PeriodicTorsion) : [1, 2, 3, 4],
strained atoms (Nonbonded)       : [5, 6],
...
```

or similar?

If convenient, this would be more informative and would save a manual follow-up step.

maxentile (approved):

timemachine/fe/model_utils.py
jkausrelay: The main issue is I'm not sure what good values are for the max_force by component.

maxentile: >not sure what good values are for the max_force by component.

Curiosity: How was `max_force = 50000` chosen?

(Assumption: Comparing to the same value of `max_force` for each force component would be ~ equally justified.)

------

Alternative proposal: Could we return a single list of atom_indices based on the same overall `max_force` comparison as currently implemented, but just add a force decomposition to the return / display?

(`atom 53 was strained: |bond force| = 0.012, |angle force| = 0.34, |torsion force| = 210391023.012, |nonbonded force| = 1223923.13412, ...`)

------

Context: Aside from visual inspection, what will a user of this function typically with the list of atom indices?

(Assumption: The next step would frequently be to identify which force component is responsible.)

proteneer (approved):
lgtm! would be good to also do this for the solvent/complex, but vacuum is insightful already!

timemachine/fe/model_utils.py
maxentile: (Tangential to current PR: Moving to issue https://github.com/proteneer/timemachine/issues/886 )

jkausrelay: > Curiosity: How was max_force = 50000 chosen?

I looked at molecules with H-H clashes that were blowing up and other molecules that were fine. This force is in the gap between the two.

> (Assumption: Comparing to the same value of max_force for each force component would be ~ equally justified.)
Good point.

> Alternative proposal: Could we return a single list of atom_indices based on the same overall max_force comparison as currently implemented, but just add a force decomposition to the return / display?

Yes, I like this better for simplicity. Can be done later (#886).

> Context: Aside from visual inspection, what will a user of this function typically do with the list of atom indices?

I was going to have minimizer.local_minimize resolve minor clashes. In my test set, it helped for all but the one structure where the Hs where very close (<0.9Å).

PR #885, opened 2022-10-14
Removes usage of GRPC code

badisa:
* Move everything to the attic


jkausrelay:
lgtm. My only comment, do we need to store these in the attic? I thought attic was more for scientific code that isn't currently being maintained.

badisa:
> lgtm. My only comment, do we need to store these in the attic? I thought attic was more for scientific code that isn't currently being maintained.

I will blow it away, re-reading the description in the readme you are right it is meant for things we want as references. Don't think this qualifies.

jkausrelay (approved):

proteneer:
nice! glad to see this blown away. and +1 to @jkausrelay , no need to store this in attic.

proteneer (approved):

PR #887, opened 2022-10-17
Updates RDKit to 2022.3.5

badisa:
- Install via Pypi
- Requires additional packages libXext and libXrender which were previously installed by conda
- Reduces size of docker container using .dockerignore, helps with caching as well.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,17 @@
+ARG LIBXRENDER_VERSION=1:0.9.10-*
+ARG LIBXEXT_VERSION=2:1.3.4-*
+
 FROM nvidia/cuda:11.6.0-devel-ubuntu20.04 AS tm_base_env
+ARG LIBXRENDER_VERSION
+ARG LIBXEXT_VERSION
 
 # Copied out of anaconda's dockerfile
 ARG MINICONDA_VERSION=4.6.14
 ARG MAKE_VERSION=4.2.1-1.2
 ARG GIT_VERSION=1:2.25.1-*
 ARG WGET_VERSION=1.20.3-1ubuntu2
-RUN apt-get update && apt-get install --no-install-recommends -y wget=${WGET_VERSION} git=${GIT_VERSION} make=${MAKE_VERSION} vim \
+RUN apt-get update && apt-get install --no-install-recommends -y \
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -85,6 +85,7 @@ def build_extension(self, ext):
         "numpy",
         "pymbar>3.0.4,<4",
         "pyyaml",
+        "rdkit",
```
mcwitt, line 88: Good call -- I think this really should've been here before too. Do you think we should also add `OpenEye-toolkits` and `OpenMM` while we're at it? (not that they would be installed with pip; this would just be an assertion that those packages are available for the timemachine installation to succeed)

Dockerfile
```diff
@@ -1,11 +1,17 @@
+ARG LIBXRENDER_VERSION=1:0.9.10-*
+ARG LIBXEXT_VERSION=2:1.3.4-*
```
mcwitt, line 2: Maybe add a comment on why these are needed?

Dockerfile
```diff
@@ -103,6 +109,11 @@ RUN pip install --no-cache-dir -e .[test] && rm -rf ./build
 
 # Container with only cuda runtime, half the size of dev container
 FROM nvidia/cuda:11.6.0-runtime-ubuntu20.04 as timemachine
+ARG LIBXRENDER_VERSION
+ARG LIBXEXT_VERSION
+RUN apt-get update && apt-get install --no-install-recommends -y libxrender1=${LIBXRENDER_VERSION} libxext-dev=${LIBXEXT_VERSION} \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
```
mcwitt, line 116: Not sure if this is possible, but (assuming `tm_base_env` has more stuff than we need for the stripped-down image), could this be moved to a common base target extended by both?

mcwitt (approved):
LGTM. It seems a bit unfortunate to have to include X infrastructure, but maybe there's no way around this. Would be good to add a comment on why this is needed.

Dockerfile
badisa: I don't think it is possible because they have different base images. I had considered copying the libraries from the original container, but seemed safer to just install in the new container.

jkausrelay (approved):

proteneer (approved):

PR #888, opened 2022-10-18
Parallelize non-GPU tests

mcwitt:
Experiment with configuring `nogpu` tests to run in parallel using `pytest-xdist`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ verify:
 
 .PHONY: nogpu_tests
 nogpu_tests:
-	pytest -m $(NOGPU_MARKER) $(PYTEST_CI_ARGS)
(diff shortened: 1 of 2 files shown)


badisa:
I would advocate against xdist, have seen way too many weird failures that are difficult to reproduce.

With the parallelism issue we had with jax recently and this case https://github.com/google/jax/issues/10242 it seems like we could end up having to chase test failures for what is currently not a bottleneck (CPU tests are 8 minutes, GPU tests are 20 minutes.

mcwitt:
> I would advocate against xdist

Fair enough; I'd been using this to speed up local testing and was mostly just curious what the impact would be in CI. I agree that it doesn't seem worth the added complexity; it roughly halves the wall time of the nogpu tests, but these are not the bottleneck as you say.

PR #889, opened 2022-10-19
Save benchmark data as json, add script to plot comparison

mcwitt:
Updates benchmarks to emit JSON files. Motivation is to allow easy comparison of benchmarks between development branches and master.

- Updates `tests/test_benchmark.py` to save benchmark data as json files (with the output path including the current commit hash)
- Adds a script to generate plots comparing benchmarks at various versions

#### Examples
image
image
image


mcwitt:
Closing as deprioritized

PR #890, opened 2022-10-20
Improve minimization of the initial states

proteneer:
This PR improves the robustness of the initial states so that the intermediate states have an `x0` that is more numerically stable. Before we used a single structure (dummy atoms of mol_a attached to the full structure of mol_b, i.e. the `lambda=1` state)

1) It generates the end-states conformations at `lambda=0`, and `lambda=1` first
2) It then tries to meet in the middle by going from `lambda=[0 -> 0.5)` and `lambda=[1->0.5]` by allowing all the ligand atoms to move.

This does create a discontinuity in the initial structure when going from `lambda<0.50` to `lambda>=0.5`

This also fixes an issue where we used `cdist` as opposed to a proper periodic wrapping when doing the distance calculation for the selection of free idxs.

### DIFF
--- a/tests/test_single_topology_confgen.py
+++ b/tests/test_single_topology_confgen.py
@@ -1,0 +1,100 @@
+import csv
+
+import numpy as np
+import pytest
+from rdkit import Chem
+
+from timemachine.constants import DEFAULT_FF, DEFAULT_TEMP
+from timemachine.fe import atom_mapping, pdb_writer, utils
+from timemachine.fe.rbfe import HostConfig, setup_initial_states
+from timemachine.fe.single_topology import AtomMapMixin, SingleTopology
+from timemachine.fe.utils import get_mol_name
+from timemachine.ff import Forcefield
+from timemachine.md import builders
+
+
+def get_mol_by_name(mols, name):
+    for m in mols:
+        if get_mol_name(m) == name:
+            return m
+    assert 0, "Mol not found"
+
+
+def write_trajectory_as_pdb(mol_a, mol_b, core, all_frames, host_topology, out_path):
+    atom_map_mixin = AtomMapMixin(mol_a, mol_b, core)
+    writer = pdb_writer.PDBWriter([host_topology, mol_a, mol_b], out_path)
+    for frame in all_frames:
+        host_frame = frame[: host_topology.getNumAtoms()]
+        ligand_frame = frame[host_topology.getNumAtoms() :]
+        mol_ab_frame = pdb_writer.convert_single_topology_mols(ligand_frame, atom_map_mixin)
+        writer.write_frame(np.concatenate([host_frame, mol_ab_frame]) * 10)
+    writer.close()
+
+
+def run_edge(mol_a, mol_b, protein_path):
+
+    threshold = 2.0
+    core, smarts = atom_mapping.get_core_with_alignment(mol_a, mol_b, threshold=threshold)
+    res = utils.plot_atom_mapping_grid(mol_a, mol_b, smarts, core)
+    with open(f"edge_map_{get_mol_name(mol_a)}_{get_mol_name(mol_b)}.svg", "w") as fh:
+        fh.write(res)
+
+    ff = Forcefield.load_from_file(DEFAULT_FF)
+    st = SingleTopology(mol_a, mol_b, core, ff)
+
+    lambda_schedule = np.linspace(0, 1, 12)
+    seed = 2023
+
+    # solvent
+    box_width = 4.0
+    solvent_sys, solvent_conf, solvent_box, solvent_top = builders.build_water_system(box_width, ff.water_ff)
+    solvent_box += np.diag([0.1, 0.1, 0.1])  # remove any possible clashes
+    solvent_host_config = HostConfig(solvent_sys, solvent_conf, solvent_box)
+    initial_states = setup_initial_states(st, solvent_host_config, DEFAULT_TEMP, lambda_schedule, seed)
+    all_frames = [state.x0 for state in initial_states]
+    write_trajectory_as_pdb(
+        mol_a,
+        mol_b,
+        core,
+        all_frames,
+        solvent_top,
+        f"solvent_{get_mol_name(mol_a)}_{get_mol_name(mol_b)}.pdb",
+    )
+
+    # complex
+    complex_sys, complex_conf, _, _, complex_box, complex_top = builders.build_protein_system(
+        protein_path, ff.protein_ff, ff.water_ff
+    )
+    complex_box += np.diag([0.1, 0.1, 0.1])  # remove any possible clashes
+    complex_host_config = HostConfig(complex_sys, complex_conf, complex_box)
+    initial_states = setup_initial_states(st, complex_host_config, DEFAULT_TEMP, lambda_schedule, seed)
+    all_frames = [state.x0 for state in initial_states]
+    write_trajectory_as_pdb(
+        mol_a,
+        mol_b,
+        core,
+        all_frames,
+        complex_top,
+        f"complex_{get_mol_name(mol_a)}_{get_mol_name(mol_b)}.pdb",
+    )
+
+
+@pytest.mark.nightly(reason="Takes a while to run")
+def test_confgen():
+
+    protein_path = "timemachine/testsystems/data/hif2a_nowater_min.pdb"
+    ligands = "timemachine/datasets/fep_benchmark/hif2a/ligands.sdf"
+    mols = [mol for mol in Chem.SDMolSupplier(ligands, removeHs=False)]
+
+    results_csv = "timemachine/datasets/fep_benchmark/hif2a/results_edges_5ns.csv"
+    with open(results_csv) as csvfile:
+        reader = csv.reader(csvfile, delimiter=",")
+        next(reader)
+        rows = [row for row in reader]
+        for row_idx, row in enumerate(rows):
+            mol_a_name, mol_b_name, exp_ddg, fep_ddg, fep_ddg_err, ccc_ddg, ccc_ddg_err = row
+            print("Processing", mol_a_name, "->", mol_b_name)
+            mol_a = get_mol_by_name(mols, mol_a_name)
+            mol_b = get_mol_by_name(mols, mol_b_name)
+
+            run_edge(mol_a, mol_b, protein_path)
--- a/timemachine/fe/rbfe.py
+++ b/timemachine/fe/rbfe.py
@@ -8,7 +8,6 @@ from typing import List
 import matplotlib.pyplot as plt
 import numpy as np
 import pymbar
-from scipy.spatial.distance import cdist
 
 from timemachine.constants import BOLTZ, DEFAULT_TEMP
 from timemachine.fe import model_utils
@@ -22,6 +21,7 @@ from timemachine.lib import LangevinIntegrator, MonteCarloBarostat, custom_ops
 from timemachine.lib.potentials import CustomOpWrapper
 from timemachine.md import builders, minimizer
 from timemachine.md.barostat.utils import get_bond_list, get_group_indices
+from timemachine.potentials import jax_utils
 
 
 def get_batch_U_fns(bps, lamb):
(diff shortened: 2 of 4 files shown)


timemachine/potentials/nonbonded.py
```diff
@@ -261,8 +261,8 @@ def nonbonded_on_specific_pairs(
     box,
     pairs,
     beta: float,
-    cutoff: Optional[float] = None,
-    w_coords: Optional[Array] = None,  # per-particle 4-d coordinate
+    cutoff=None,
```
proteneer, line 264: @mcwitt: I'm going to remove these two typings because its giving me too much grief on my python. I'll add them back me once we migrate to python11

timemachine/fe/single_topology.py
```diff
@@ -925,14 +925,47 @@ def combine_confs(self, x_a, x_b):
...
+        # place a first, then b overrides a
         x0 = np.zeros((self.get_num_atoms(), 3))
         for src, dst in enumerate(self.a_to_c):
             x0[dst] = x_a[src]
         for src, dst in enumerate(self.b_to_c):
             x0[dst] = x_b[src]
+
+        return x0
+
+    def _combine_confs_ba(self, x_a, x_b):
```
jkausrelay, line 959: Would it be clearer to swap x_a/x_b in the args?

timemachine/fe/single_topology.py
```diff
@@ -925,14 +925,47 @@ def combine_confs(self, x_a, x_b):
             Combined conformation
 
         """
-        # tbd: allow input to take lambda schedule and return an array for each value of lambda
-        assert x_a.shape == (self.mol_a.GetNumAtoms(), 3)
-        assert x_b.shape == (self.mol_b.GetNumAtoms(), 3)
+        return self._combine_confs_ab(x_a, x_b)
+
+    def _combine_confs_ab(self, x_a, x_b):
```
jkausrelay, line 930: Make these public or have combine_confs take an extra argument for direction?

Otherwise you'll need to call a private function from outside the class.

tests/test_single_topology_confgen.py
```diff
@@ -0,0 +1,100 @@
...
+    all_frames = [state.x0 for state in initial_states]
+    write_trajectory_as_pdb(
+        mol_a,
+        mol_b,
+        core,
+        all_frames,
+        solvent_top,
+        f"solvent_{get_mol_name(mol_a)}_{get_mol_name(mol_b)}.pdb",
+    )
+
```
jkausrelay, line 64: Add some sanity check here (maybe rmsd to the mol_a/mol_b structures as appropriate?)

proteneer: 1) the internal BFGS code already checks that the energy decreases
2) i can add a check that the drift relative to the starting configuration is small

timemachine/fe/single_topology.py
proteneer: agreed, the naming is confusing as hell too. I'll rename this explicitly to `combine_confs_lhs` and `combine_confs_rhs` to denote setting up the conformation for the left and right end-states.

tests/test_single_topology_confgen.py
proteneer: I added a check that so that no atom drifts more than 7 angstroms from the initial state. (arbitrary)

This is now done directly in the setup code.

timemachine/fe/rbfe.py
```diff
@@ -459,6 +483,66 @@ def estimate_free_energy_given_initial_states(initial_states, protocol, temperat
...
+
+    # go from lambda 0 -> 0.5
+    if len(lhs_initial_states) > 0:
+        lhs_xs = _optimize_coords_along_states(lhs_initial_states)
+        for xs in lhs_xs:
+            all_xs.append(xs)
+
+    # go from lambda 1 -> 0.5 and reverse the coordinate trajectory and lambda schedule
+    if len(rhs_initial_states) > 0:
+        rhs_xs = _optimize_coords_along_states(lhs_initial_states)[::-1][::-1]
```
badisa, line 536: Should this be rhs and not lhs?

```suggestion
        rhs_xs = _optimize_coords_along_states(rhs_initial_states)[::-1][::-1]
```

proteneer: good catch! d'oh

timemachine/fe/rbfe.py
```diff
@@ -459,6 +483,66 @@ def estimate_free_energy_given_initial_states(initial_states, protocol, temperat
...
+    if len(lhs_initial_states) > 0:
+        lhs_xs = _optimize_coords_along_states(lhs_initial_states)
+        for xs in lhs_xs:
+            all_xs.append(xs)
+
+    # go from lambda 1 -> 0.5 and reverse the coordinate trajectory and lambda schedule
+    if len(rhs_initial_states) > 0:
+        rhs_xs = _optimize_coords_along_states(rhs_initial_states)[::-1][::-1]
+        for xs in rhs_xs:
+            all_xs.append(xs)
```
proteneer, line 538: ugh this code is buggy - it doesn't splice back the sequence correctly together.

proteneer: nm actually this is okay because the initial lambda schedule is monotonic. So all `lhs_initial_states` come before all `rhs_initial_states`

jkausrelay (approved):

timemachine/fe/rbfe.py
```diff
@@ -151,6 +142,27 @@ class SimulationResult:
 # setup the initial state so we can (hopefully) bitwise recover the identical simulation
 # to help us debug errors.
 def setup_initial_states(st, host_config, temperature, lambda_schedule, seed):
+    """
+    Setup the initial states for a series of lambda values. It is assumed that the lambda schedule
+    is a monotonically increasing sequence in the closed interval [0,1].
+
+    Parameters
+    ----------
```
badisa, line 150: nit: Doesn't mention the `st` parameter for the single topology object.

timemachine/fe/rbfe.py
```diff
@@ -164,11 +176,21 @@ def setup_initial_states(st, host_config, temperature, lambda_schedule, seed):
 
     initial_states = []
 
+    # check that the lambda schedule is monotonically increasing.
+    assert np.all(np.diff(lambda_schedule) > 0)
```
badisa, line 180: Should we use this? `from timemachine.fe.lambda_schedule import validate_lambda_schedule`

timemachine/fe/rbfe.py
```diff
@@ -459,6 +483,66 @@ def estimate_free_energy_given_initial_states(initial_states, protocol, temperat
...
+
+    # go from lambda 0 -> 0.5
+    if len(lhs_initial_states) > 0:
+        lhs_xs = _optimize_coords_along_states(lhs_initial_states)
+        for xs in lhs_xs:
+            all_xs.append(xs)
+
+    # go from lambda 1 -> 0.5 and reverse the coordinate trajectory and lambda schedule
+    if len(rhs_initial_states) > 0:
+        rhs_xs = _optimize_coords_along_states(rhs_initial_states)[::-1][::-1]
```
badisa, line 536: Did you mean to reverse the order of `rhs_initial_states`? As right now the end_state in `optimize_coords_along_states` will be `rhs_initial_states[0]` rather than `rhs_initial_sates[-1]`

timemachine/potentials/nonbonded.py
mcwitt: OK. Would it also fix to change `Array: TypeAlias = NDArray` to `Array = Any`? The latter is the convention we've adopted in other places, and IIUC it's the `TypeAlias` that's breaking with 3.7.0b0?

timemachine/fe/rbfe.py
proteneer: willfix

timemachine/fe/rbfe.py
proteneer: 1) API's not quite the same, i.e. `num_windows` is not specified
2) we don't strictly require that the schedule start at 0 and 1 at the end (some of our benchmark tests just set the lambda schedule to [0.1])

timemachine/fe/rbfe.py
proteneer: what's also kind of crazy is how the tests still pass for the hif2a set... maybe I need to find more problematic sets.

badisa (approved):

timemachine/potentials/nonbonded.py
proteneer: I think that works actually - I'll just set `Array = Any` instead of remove these annotations.

timemachine/fe/single_topology.py
```diff
@@ -925,14 +925,35 @@ def combine_confs(self, x_a, x_b):
...
             x0[dst] = x_b[src]
+
+        return x0
+
+    def combine_confs_lhs(self, x_a, x_b):
+        """
+        Combine x_a and x_b conformations for lambda=0
+        """
+        # place b first, then a overrides b
+        assert x_a.shape == (self.mol_a.GetNumAtoms(), 3)
```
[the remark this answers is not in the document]
proteneer: fixed

timemachine/potentials/nonbonded.py
proteneer: fixed - thanks for noticing this.

timemachine/fe/rbfe.py
```diff
@@ -459,6 +486,66 @@ def estimate_free_energy_given_initial_states(initial_states, protocol, temperat
...
+
+    # go from lambda 1 -> 0.5 and reverse the coordinate trajectory and lambda schedule
+    if len(rhs_initial_states) > 0:
+        rhs_xs = _optimize_coords_along_states(rhs_initial_states[::-1])[::-1]
+        for xs in rhs_xs:
+            all_xs.append(xs)
+
+    for state, coords in zip(initial_states, all_xs):
+        # sanity check that no atom has moved more than 7 angstroms away (arbitrary)
+        assert np.amax(np.linalg.norm(state.x0 - coords, axis=1)) < 0.7
```
maxentile, line 545: Should "close to initial coords" appear only as an assertion after optimization, or should this also be part of the optimization objective?

(If the optimization objective allows to go arbitrarily far away from initial coords, then an assertion afterwards will either have to be pretty loose, or be expected to fail frequently... But I haven't seen the validation, so maybe 7 angstroms is simultaneously tight enough to be informative, and loose enough not to cause spurious assertion errors in any of the practical cases tested...)

timemachine/fe/rbfe.py
```diff
@@ -459,6 +486,66 @@ def estimate_free_energy_given_initial_states(initial_states, protocol, temperat
...
+        x_traj.append(x_opt)
+        assert np.all(np.isfinite(x_opt)), "Minimization resulted in a nan"
+        del bound_impls
+
+    return x_traj
+
+
+def optimize_coordinates(initial_states):
+    """
+    Optimize geometries of the initial states. The initial states are modified in place.
```
maxentile, line 514: Nit: Is it beneficial to modify `initial_states` in-place, rather than returning a copy?

(Don't see a performance advantage since copies are created within this function, do see examples of where it would be useful to compare coordinates before / after optimization, and in general find it easier to think about less state...)

timemachine/fe/rbfe.py
```diff
@@ -164,11 +179,21 @@ def setup_initial_states(st, host_config, temperature, lambda_schedule, seed):
...
+    # check that the lambda schedule is monotonically increasing.
+    assert np.all(np.diff(lambda_schedule) > 0)
+
     for lamb_idx, lamb in enumerate(lambda_schedule):
 
         mol_a_conf = get_romol_conf(st.mol_a)
         mol_b_conf = get_romol_conf(st.mol_b)
-        ligand_conf = st.combine_confs(mol_a_conf, mol_b_conf)
+
+        assert lamb >= 0.0 or lamb <= 1.0
```
maxentile, line 190: nit: `or` or `and`?

timemachine/fe/rbfe.py
```diff
@@ -151,6 +142,30 @@ class SimulationResult:
...
+
+    host_config: HostConfig or None
+        Configurations of the host. If None, then a vacuum state will be setup.
+
+    temperature: float
+        Temperature to run the simulation at.
+
+    lambda_schedule: list of float of length K
+        Lambda schedule.
+
```
maxentile, line 162: nit: missing parameter

tests/test_single_topology_confgen.py
```diff
@@ -0,0 +1,100 @@
...
+        reader = csv.reader(csvfile, delimiter=",")
+        next(reader)
+        rows = [row for row in reader]
+        for row_idx, row in enumerate(rows):
+            mol_a_name, mol_b_name, exp_ddg, fep_ddg, fep_ddg_err, ccc_ddg, ccc_ddg_err = row
+            print("Processing", mol_a_name, "->", mol_b_name)
+            mol_a = get_mol_by_name(mols, mol_a_name)
+            mol_b = get_mol_by_name(mols, mol_b_name)
+
+            run_edge(mol_a, mol_b, protein_path)
```
maxentile, line 100: nit: testing: should this PR introduce a unit test of `optimize_coordinates`?

(In addition to the affected integration tests that rely on `setup_initial_states`...)

timemachine/fe/rbfe.py
proteneer: I'm not sure tbh - I think we can add this as an objective function but we'd need to figure out the penalty for the RMSD here. For complex/solvent minimization, the outer shell of the environment is frozen, so it'd be very unlikely to escape too much. 7 angstroms seems to have passed on the hif2a set - and any atom that moves further than that will be problematic

tests/test_single_topology_confgen.py
proteneer: I can cherry pick a couple of edges.

timemachine/fe/rbfe.py
maxentile: Hmm, fair enough -- I'm not sure what cases this assertion will catch / what the ideal intended behavior of the minimizer should be, but (1) picking a loose threshold like this seems safe as a post-condition, (2) seems somewhat safe to assume unrestrained minimization will introduce no other unwanted changes to the conformer...

(Although we may want to revisit (2) in the future to be more explicit, e.g. to prevent the minimizer from flipping important torsions, or to prevent the minimizer from moving any atom by more than 7 angstroms.)

timemachine/fe/rbfe.py
proteneer: I'm going to go with a hybrid solution - just have this return a list of coordinates and have the calling function do the  modification (which does the setup) so it's more explicit.

PR #891, opened 2022-10-21
Remove lambda dependence in bonded potentials

maxentile:
- [x] Modifies harmonic bond, harmonic angle, and periodic torsion to remove dependence on `lamb`, `lamb_mult` (sometimes aka `lambda_mult`), `lamb_offset` (sometimes aka `lambda_offset`). In all cases, the effect of these additional arguments can be reproduced by setting parameter `k`, so this change does not reduce expressivity. There are some sites where wrappers still accept `lam` and return `du_dlam` where needed for API consistency but in those cases `lam` is now ignored and `du_dlam` is always 0.
- [x] Renames `k_harmonic_angle_inference` to `k_harmonic_angle` for consistency with `k_harmonic_bond`
- [x] Moves unused k_restraint.cuh to attic


timemachine/cpp/src/harmonic_angle.cu
```diff
@@ -37,30 +27,9 @@ HarmonicAngle<RealType>::HarmonicAngle(
...
-    } else {
-        initializeArray(A_, d_lambda_mult_, 0);
-    }
-
-    if (lambda_offset.size() > 0) {
-        gpuErrchk(
-            cudaMemcpy(d_lambda_offset_, &lambda_offset[0], A_ * sizeof(*d_lambda_offset_), cudaMemcpyHostToDevice));
-    } else {
-        // can't memset this
-        initializeArray(A_, d_lambda_offset_, 1);
```
mcwitt, line 55: nit: it looks like `initializeArray` is now unused; do we want to remove it?

timemachine/cpp/src/kernels/k_harmonic_angle.cuh
```diff
@@ -58,35 +56,33 @@ void __global__ k_harmonic_angle_inference(
...
     }
 
     if (u) {
-        atomicAdd(u + i_idx, FLOAT_TO_FIXED_BONDED(ka / 2 * delta * delta * prefactor));
+        atomicAdd(u + i_idx, FLOAT_TO_FIXED_BONDED(ka / 2 * delta * delta));
     }
 
     if (du_dl) {
-        atomicAdd(du_dl + i_idx, FLOAT_TO_FIXED_BONDED(lambda_mult[a_idx] * ka / 2 * delta * delta));
+        atomicAdd(du_dl + i_idx, FLOAT_TO_FIXED_BONDED(0.0));
```
mcwitt, line 86: nit: is there a reason to leave this in (assuming it's a no-op)?

mcwitt (approved):
Looks great to me. Awesome simplification!

timemachine/cpp/src/kernels/k_harmonic_angle.cuh
maxentile: Good call -- removed from here (and `k_harmonic_bond`, `k_periodic_torsion`) in https://github.com/proteneer/timemachine/pull/891/commits/c60addb402e3d8a9a196ef9bb0795f912583ca6c

timemachine/cpp/src/harmonic_angle.cu
maxentile: Good catch -- `initializeArray` is now unused.

Considered removal in https://github.com/proteneer/timemachine/pull/891/commits/89be7845558a2ec4f70490c6d5fb3e723b88017c, but reverted in https://github.com/proteneer/timemachine/pull/891/commits/c915e9bbc378f91062b01c949d294a698a3b0ec5, since the kernel `k_initialize_array` it wraps is used in the neighborlist.

badisa (approved):

PR #893, opened 2022-10-25
Remove lambda dependence in nonbonded potentials

mcwitt:
This is the penultimate step of the project to remove lambda dependence from custom ops potentials. The remaining step is to remove lambda from the signatures of the `Potential` interface and reference potentials.

- Removes unused `LambdaPotential`
- Removes `lambda_plane_idxs` and `lambda_offset_idxs` constructor arguments from nonbonded potentials
- Augments nonbonded parameters with 4D offset per atom/pair: (charge, sigma, epsilon) -> (charge, sigma, epsilon, w)
- Updates topology classes to use parameters for decoupling
- Updates dU/dp computation with dU/dw component


tests/test_jax_nonbonded.py
```diff
@@ -158,17 +157,24 @@ def generate_random_inputs(n_atoms, dim, instance_flags=difficult_instance_flags
...
     eps = jnp.ones(n_atoms)
+    w_coords = jnp.zeros(n_atoms)
     if instance_flags["randomize_charges"]:
         charges = randn(n_atoms)
     if instance_flags["randomize_sigma"]:
         sig = min_dist * rand(n_atoms)
     if instance_flags["randomize_epsilon"]:
         eps = rand(n_atoms)
+    if instance_flags["randomize_w_coords"]:
+        w_coords = 2 * cutoff * rand(n_atoms)
```
mcwitt, line 175: TODO: test negative w

tests/test_reweighting.py
```diff
@@ -225,6 +225,7 @@ def batched_u_fxn(samples, ligand_nb_params):
     return samples_0, samples_1, batched_u_0, batched_u_1, ref_params, ref_delta_f
 
 
+@pytest.mark.skip(reason="needs update since removal of lambda dependence in nonbonded potentials")
```
mcwitt, line 228: TODO: investigate fix

tests/nonbonded/test_nonbonded_all_pairs.py
```diff
@@ -117,20 +104,23 @@ def test_nonbonded_all_pairs_correctness(
...
-    potential = generic.NonbondedAllPairs(lambda_plane_idxs, lambda_offset_idxs, beta, cutoff, atom_idxs)
-    lambda_values = [0.0, 0.1]
+    potential = generic.NonbondedAllPairs(num_atoms, beta, cutoff, atom_idxs)
+
+    def gen_params(n):
+        for _ in range(n):
+            w_coords = rng.uniform(-2, 2, (num_atoms,)) * cutoff
+            params_ = params.copy()
+            params_[:, 3] = w_coords
+            yield params_
```
mcwitt, line 118: TODO: replace with `gen_params_with_4d_offsets`

tests/common.py
```diff
@@ -340,14 +358,19 @@ def compare_forces(
...
         precision,
         atol: float = 1e-8,
     ):
         return self.compare_forces(
-            x, params, box, lambdas, potential.to_reference(), potential.to_gpu(), rtol, precision, atol
+            x, params_arrays, box, potential.to_reference(), potential.to_gpu(), rtol, precision, atol
         )
+
+
+def gen_params_with_4d_offsets(rng: np.random.Generator, params, w_min, w_max, n):
```
mcwitt, line 373: TODO: naming (gen_nb_params…?)

tests/nonbonded/test_consistency.py
```diff
@@ -87,9 +68,10 @@ def make_pairlist_potential(exclusion_idxs, exclusion_scales):
     ).unbound_impl(precision)
 
     def test():
-        for lam in [0.0, 0.1]:
-            du_dx_ref, du_dp_ref, du_dl_ref, u_ref = ref_impl.execute(conf, params, example_box, lam)
-            du_dx_test, du_dp_test, du_dl_test, u_test = test_impl.execute(conf, params, example_box, lam)
+        for w_coords in rng.uniform(-2, 2, (5, num_atoms)) * cutoff:
```
mcwitt, line 71: TODO: use `gen_params_with_4d_offsets`

tests/test_jax_nonbonded.py
```diff
@@ -34,12 +29,19 @@
...
+ChargeMask = Array
+LJMask = Array
+Lamb = float
+Beta = float
+Cutoff = float
+Energy = float
 
-nonbonded_args = Conf, Params, Box, Lamb, ChargeMask, LJMask, Beta, Cutoff, LambdaPlaneIdxs, LambdaOffsetIdxs
-NonbondedArgs = Tuple[nonbonded_args]
-NonbondedFxn = Callable[[*nonbonded_args], Energy]
```
mcwitt, line 42: This syntax looks a little strange to me (should it be `Tuple[*nonbonded_args]`), and isn't accepted by my IDE:

> Unpack operator in subscript requires Python 3.11 or newer

so I opted for a more verbose version for now.

tests/test_jax_nonbonded.py
```diff
@@ -186,31 +190,8 @@ def generate_random_inputs(n_atoms, dim, instance_flags=difficult_instance_flags
...
-        box,
-        lamb,
-        charge_rescale_mask,
-        lj_rescale_mask,
-        beta,
-        cutoff,
-        lambda_plane_idxs,
-        lambda_offset_idxs,
-    )
+    args = (conf, params, box, lamb, charge_rescale_mask, lj_rescale_mask, beta, cutoff)
```
badisa, line 194: Looks like lambda is still used in this file, is there something about the references that still require lambda?

timemachine/cpp/src/kernels/k_nonbonded_pair_list.cuh
```diff
@@ -156,6 +158,9 @@ void __global__ k_nonbonded_pair_list(
         g_qi += FLOAT_TO_FIXED_DU_DP<RealType, FIXED_EXPONENT_DU_DCHARGE>(charge_scale * qj * inv_dij * ebd);
         g_qj += FLOAT_TO_FIXED_DU_DP<RealType, FIXED_EXPONENT_DU_DCHARGE>(charge_scale * qi * inv_dij * ebd);
 
+        g_wi += FLOAT_TO_FIXED_NONBONDED(delta_prefactor * delta_w);
```
badisa, line 161: Any reason not to move this into the `if (du_dp)` block?

mcwitt: Not that I see; this was just keeping with the convention used for the other parameter gradients here (accumulate into the register outside the `if (du_dp)` block; accumulate using `atomicAdd` inside the block)

tests/test_jax_nonbonded.py
mcwitt: The references should be independent of lambda now. But my plan was to hold off on removing lambda from the reference signatures until we do the same for the custom ops potentials (which will be done in the final PR of this series)

tests/test_reweighting.py
maxentile: The skip annotations make sense for now, and I'll update the relevant functions in a follow-up!

timemachine/potentials/nonbonded.py
```diff
@@ -139,23 +138,19 @@ def nonbonded(
...
     * Coulomb terms are multiplied by erfc(beta * distance)
 
     Parameters
     ----------
-    conf : (N, 3) or (N, 4) np.ndarray
-        3D or 4D coordinates
-        if 3D, will be converted to 4D using (x,y,z) -> (x,y,z,w)
-            where w = cutoff * (lambda_plane_idxs + lambda_offset_idxs * lamb)
+    conf : (N, 3) np.ndarray
+        3D coordinates
```
maxentile, line 151: Good simplification

timemachine/lib/potentials.py
```diff
@@ -295,24 +277,18 @@ def get_idxs(self):
...
+        return self.args[3]
 
 
 class NonbondedPairListPrecomputed(CustomOpWrapper):
     """
-    This implements a pairlist with precomputed parameters. It differs from
-    the regular NonbondedPairlist in that it expects params of the form s0*q_ij, s_ij, and s1*e_ij
-    where s are the scaling factor and combining rules have already been applied.
+    This implements a pairlist with precomputed parameters. It differs from the regular NonbondedPairlist in that it
+    expects params of the form s0*q_ij, s_ij, s1*e_ij, and w_offsets, where s are the scaling factors and combining
```
maxentile, line 290: nit:
```suggestion
    expects params of the form s0*q_ij, s_ij, s1*e_ij, and w_offsets_ij, where s are the scaling factors and combining
```

(for consistency and to emphasize that this is one offset per pair)

timemachine/fe/topology.py
```diff
@@ -253,16 +246,21 @@ def parameterize_nonbonded_pairlist(self, ff_q_params, ff_lj_params):
...
         params = []
         for q, s, e, (sf_q, sf_lj) in zip(q_ij, sig_ij, eps_ij, rescale_mask):
-            params.append((q * sf_q, s, e * sf_lj))
+            params.append(
+                (
+                    q * sf_q,
+                    s,
+                    e * sf_lj,
+                    0.0,  # w offset
+                )
```
maxentile, line 255: nit: `s` -> `sig`, `e` -> `eps`?

timemachine/md/minimizer.py
```diff
@@ -145,12 +147,18 @@ def minimize_host_4d(mols, host_system, host_coords, ff, box, mol_coords=None) -
...
-    _, xs, _ = ctxt.multiple_steps(np.linspace(1.0, 0, 1000))
+    u_impls = bind_potentials(parameterize_system(hgt, ff, 1.0))
+    x = fire_minimize(x0, u_impls, box, np.ones(50))
+
+    for lamb in np.linspace(1.0, 0, 50):
+        u_impls = bind_potentials(parameterize_system(hgt, ff, lamb))
+        # context components: positions, velocities, box, integrator, energy fxns
+        ctxt = custom_ops.Context(x, v0, box, intg, u_impls)
+        _, xs, _ = ctxt.multiple_steps(lamb * np.ones(50))
+        x = xs[-1]
```
mcwitt, line 158: Note: this reduces the number of lambda windows used for minimization from 1000 down to 50 (while increasing the number of steps per window from 1 to 50). This is probably less efficient, but since we currently need to construct a new bound potential for each lambda value, it was necessary to reduce the number of lambda windows.

Aside: it might make sense for a future PR to add an interface to `Context` that allows passing unbound potentials and parameters separately? That way we could at least avoid the cost of constructing a new potential object on each iteration here.

timemachine/fe/single_topology.py
```diff
@@ -1197,55 +1192,53 @@ def _parameterize_host_guest_nonbonded(self, lamb, host_nonbonded):
...
             elif membership == 2:  # dummy_B
                 b_idx = self.c_to_b[idx]
-                plane = -1
-                offset = 1
                 q = guest_b_q[b_idx]
                 sig = guest_b_lj[b_idx, 0]
                 eps = guest_b_lj[b_idx, 1]
+
+                # w: -cutoff -> 0
+                w_coords = (lamb - 1.0) * cutoff
```
maxentile, line 1220: nit: took a few seconds to parse the comments as specifying boundary conditions for interpolation (possible style collision with function description `f: x -> y`).

May expand the wording a bit to `# interpolate from w = 0 to w = cutoff`, or `# w = -cutoff @ lamb=0, w = 0 @ lamb=1` or similar, or perhaps avoid need for comment using `w_coords = lerp(lamb, -cutoff, 0)` or similar...

maxentile (commented):
Still reviewing but looks great so far!

tests/common.py
```diff
@@ -72,6 +73,7 @@ def prepare_system_params(x: NDArray, sigma_scale: float = 5.0) -> NDArray:
             (np.random.rand(N).astype(np.float64) - 0.5) * np.sqrt(ONE_4PI_EPS0),  # q
             np.random.rand(N).astype(np.float64) / sigma_scale,  # sig
             np.random.rand(N).astype(np.float64),  # eps
+            np.random.rand(N).astype(np.float64) * 2 * cutoff,  # w
```
mcwitt, line 76: TODO: test with some negative `w`

timemachine/fe/topology.py
mcwitt: Done in [s -> sig, e -> eps](https://github.com/proteneer/timemachine/pull/893/commits/9c12e4a03859e21b77e91d63aad3f60a323c33f1)

timemachine/fe/single_topology.py
mcwitt: Good call, I like the suggestion to use `lerp`. Updated in faede63

tests/common.py
```diff
@@ -340,14 +334,39 @@ def compare_forces(
...
+    # all w_max
+    yield params_with_w_coords(w_max)
+
+    # half zero, half w_max
+    w_coords = jnp.zeros(num_atoms)
+    w_coords = w_coords.at[-num_atoms // 2 :].set(w_max)
+    yield params_with_w_coords(w_coords)
+
+    # discrete random uniform in {0, w_max}
+    w_coords = rng.choice([0.0, w_max], (num_atoms,))
```
maxentile, line 367: Nit / curiosity: Motivation for set `{0, w_max}`?

(From context, expected a slightly different set `{0, w_min, w_max}`, or possibly redundant `{0, +w_min, +w_max, -w_min, -w_max}`.)

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -292,55 +289,55 @@ void __device__ v_nonbonded_unified(
...
     unsigned long long gi_y = 0;
     unsigned long long gi_z = 0;
 
-    int charge_param_idx_i = atom_i_idx * 3 + 0;
-    int lj_param_idx_sig_i = atom_i_idx * 3 + 1;
-    int lj_param_idx_eps_i = atom_i_idx * 3 + 2;
+    int charge_param_idx_i = atom_i_idx * PARAMS_PER_ATOM + PARAM_OFFSET_CHARGE;
+    int lj_param_idx_sig_i = atom_i_idx * PARAMS_PER_ATOM + PARAM_OFFSET_SIG;
+    int lj_param_idx_eps_i = atom_i_idx * PARAMS_PER_ATOM + PARAM_OFFSET_EPS;
+    int w_param_idx_i = atom_i_idx * PARAMS_PER_ATOM + PARAM_OFFSET_W;
```
maxentile, line 304: Nit: style -- I do see the reason to replace `* 3` with `* PARAMS_PER_ATOM`, but personally these index expressions seem slightly noisier to skim now, just because there are more characters per line...

Perhaps extract repeated `atom_i_idx * PARAMS_PER_ATOM`?

timemachine/cpp/src/kernels/nonbonded_common.cuh
```diff
@@ -6,13 +6,15 @@
 #include <set>
 #include <vector>
 
+// each atom parameterized by a 4-tuple: charge, lj sigma, lj epsilon
```
maxentile, line 9: nit: comment is missing w

timemachine/cpp/src/kernels/nonbonded_common.cuh
```diff
@@ -6,13 +6,15 @@
...
+enum { PARAM_OFFSET_CHARGE = 0, PARAM_OFFSET_SIG, PARAM_OFFSET_EPS, PARAM_OFFSET_W, PARAMS_PER_ATOM };
```
maxentile, line 10: Nice!

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -393,12 +361,16 @@ void NonbondedInteractionGroup<RealType>::du_dp_fixed_to_float(
...
+        const int idx = i * PARAMS_PER_ATOM;
+        const int idx_charge = idx + PARAM_OFFSET_CHARGE;
+        const int idx_sig = idx + PARAM_OFFSET_SIG;
+        const int idx_eps = idx + PARAM_OFFSET_EPS;
+        const int idx_w = idx + PARAM_OFFSET_W;
+
         du_dp_float[idx_charge] = FIXED_TO_FLOAT_DU_DP<double, FIXED_EXPONENT_DU_DCHARGE>(du_dp[idx_charge]);
         du_dp_float[idx_sig] = FIXED_TO_FLOAT_DU_DP<double, FIXED_EXPONENT_DU_DSIG>(du_dp[idx_sig]);
         du_dp_float[idx_eps] = FIXED_TO_FLOAT_DU_DP<double, FIXED_EXPONENT_DU_DEPS>(du_dp[idx_eps]);
+        du_dp_float[idx_w] = FIXED_TO_FLOAT<double>(du_dp[idx_w]);
```
maxentile, line 373: Question: Should the fixed exponent for `du_dw` be exposed separately? (Or is it okay to keep it tied to global `FIXED_EXPONENT`?)

maxentile (approved):
Really nice simplification, thoroughly done! Looks complete and correct as far as I can tell.

(Only comments are non-blocking -- curiosities or weakly-held style opinions)

tests/common.py
mcwitt: The idea was just to test with sparse w coordinates (nonvanishing probability of zero), in contrast with the continuous uniform case (where there won't be any w that are exactly zero) and the ordered "half 0, half w_max" case. Since I was focused on "sparse, randomly allocated", I didn't think to test with other nonzero w values, but I don't see any harm in adding this.

timemachine/cpp/src/kernels/k_nonbonded.cuh
mcwitt: I was also kind of annoyed by this and tried out shorter names for the constants, but nothing stuck. I'll give your suggestion of factoring out the common offset a try.

timemachine/cpp/src/nonbonded_interaction_group.cu
mcwitt: I was on the fence about this; was thinking to do this later if we find it necessary in some case, but for now it seemed simplest to use the same fixed exponent that we use for the other positional gradients.

maxentile: Makes sense!

(To clarify, I wasn't suggesting to pick a different value than `FIXED_EXPONENT` initially, just to add a `static const auto FIXED_EXPONENT_DU_DW = FIXED_EXPONENT;` or something to https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/fixed_point.hpp , and add references to it from lines like this, so that (1) it's initially the same as `FIXED_EXPONENT`, (2) if tuning it separately becomes needed, only one site will need to be updated. It does seem quite different to me from spatial gradients, so I imagine the need could arise to tune a separate exponent here.)

mcwitt: Good call, I think this makes sense. Added in 08ddb1d. This also led me to discover a bug where I'd used the charge exponent by mistake (which just happens to be the same as the one for dU/dx)

timemachine/cpp/src/kernels/nonbonded_common.cuh
mcwitt: Thanks, fixed in c66f379

timemachine/cpp/src/kernels/k_nonbonded.cuh
mcwitt: Addressed in ea2a474

tests/common.py
mcwitt: > Another way to accomplish this

Thanks, I think your suggestion is better. Updated in bc14618.

tests/nonbonded/test_nonbonded_interaction_group.py
```diff
@@ -79,26 +67,14 @@ def test_nonbonded_interaction_group_correctness(
...
-        cutoff,
-    )
-
-    lambda_vals = [0.0, 0.1]
+    potential = generic.NonbondedInteractionGroup(num_atoms, ligand_idxs, beta, cutoff)
 
     GradientTest().compare_forces_gpu_vs_reference(
         conf,
-        params,
+        gen_nonbonded_params_with_4d_offsets(rng, params, -2 * cutoff, 2 * cutoff),
```
proteneer, line 76: what's the rationale for going beyond the cutoff?

tests/test_jax_nonbonded.py
```diff
@@ -158,17 +155,24 @@ def generate_random_inputs(n_atoms, dim, instance_flags=difficult_instance_flags
     min_dist = 0.1
     conf, box = resolve_clashes(conf, box, min_dist=min_dist)
 
+    cutoff = 1.2
+    if instance_flags["randomize_cutoff"]:
+        cutoff += rand()
```
proteneer, line 160: note that we have hard requirement that cutoff must be less than half of the box width (not sure what the box sizes here are).

tests/test_single_topology_combined.py
```diff
@@ -103,73 +104,77 @@ def _test_combined_parameters_impl_nonbonded(host_system_omm):
...
+                    if lamb == 0.0:
+                        assert w == 0.0
+                    elif lamb == 1.0:
+                        assert w == pytest.approx(cutoff)
                 elif indicator == 2:
                     # mol_b dummy
-                    assert offset_idx == 1
-                    assert plane_idx == -1
+                    if lamb == 0.0:
+                        assert w == pytest.approx(-cutoff)
```
proteneer, line 140: nit: (we can probably set this to +cutoff) later on.

timemachine/md/minimizer.py
proteneer: > Aside: it might make sense for a future PR to add an interface to Context that allows passing unbound potentials and parameters separately? That way we could at least avoid the cost of constructing a new potential object on each iteration here.

I think that's the cleanest way.

proteneer, line 158: did we want to carry over the velocities?  (Currently it resets them to zero each time)

proteneer (approved):
lgtm - mostly minor comments. 

Only application level change that we may need to make a mental note of is that it does change the host minimization routine slightly. But I think it should be fine.

tests/nonbonded/test_nonbonded_interaction_group.py
mcwitt: This is just for consistency with the existing tests, which tested lambda plane indices in {-2, 1, 0, 1, 2} (e.g. [here](https://github.com/proteneer/timemachine/pull/893/files#diff-bb6ec9fd4f933d4ba34a7034486e1d1d3be687475b700a38143fa656c04f9a89L82-L83)). It also seems useful to test nontrivial cases where some pairs are fully decoupled

proteneer: makes sense!

timemachine/md/minimizer.py
mcwitt: In offline discussion, decided to leave as-is; preserving velocities between lambda windows seems to decrease efficiency, requiring more windows to pass the check. Commented in e154a22).

tests/test_jax_nonbonded.py
mcwitt: It doesn't seem like we currently enforce this requirement in the reference potentials(?) (note this test code is for the reference nonbonded potential)

(also, this was not added in the PR but moved up from lower in the file so that the randomized cutoff could be used to scale the random w coordinates)

tests/nonbonded/test_nonbonded_interaction_group.py
mcwitt: After some more thought, I addressed this in 2914ed4. I think you are right that there's no benefit to testing `w` in `[-2 * cutoff, 2 * cutoff]`, since even with `[-cutoff, cutoff]` there will be fully decoupled pairs. Also fixed the generation of negative pair `delta_w`s in the test for `NonbondedPrecomputed`.

PR #896, opened 2022-10-28
Fix missing traj

jkausrelay:
- Code only works for positive idxs
- Add checks to ensure keep_idxs is positive


proteneer:
Good call:

```
if lamb_idx in keep_idxs:
    stored_frames.append(cur_frames)
    stored_boxes.append(cur_boxes)
```

PR #898, opened 2022-10-31
Add conformer-oblivious core option

maxentile:
Problem: `mcs` has an option to ignore conformer, but `get_core_by_mcs` does not.

(In some situations we may want to toggle off conformer-awareness in constructing atom-mapping, e.g. in relative hydration, and perhaps also to increase efficiency of the solvent leg in relative binding.)

Solution: Add a `conformer_aware` option (defaulting to `True`) to `get_core_by_mcs`.

### DIFF
--- a/tests/test_atom_mapping.py
+++ b/tests/test_atom_mapping.py
@@ -22,8 +22,9 @@ def test_mcs():
 def test_get_core_by_mcs():
     mol_a, mol_b, _ = get_hif2a_ligand_pair_single_topology()
     query = mcs(mol_a, mol_b).queryMol
-    core = get_core_by_mcs(mol_a, mol_b, query)
-    assert core.shape[1] == 2
+    for conformer_aware in [True, False]:
+        core = get_core_by_mcs(mol_a, mol_b, query, conformer_aware=conformer_aware)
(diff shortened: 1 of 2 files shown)


tests/test_atom_mapping.py
```diff
@@ -22,8 +22,9 @@ def test_mcs():
 def test_get_core_by_mcs():
     mol_a, mol_b, _ = get_hif2a_ligand_pair_single_topology()
     query = mcs(mol_a, mol_b).queryMol
-    core = get_core_by_mcs(mol_a, mol_b, query)
-    assert core.shape[1] == 2
+    for conformer_aware in [True, False]:
```
jkausrelay, line 25: nit: You can use pytest to run the test under different conditions
```
@pytest.mark.parametrize("conformer_aware", [True, False])
def test_get_core_by_mcs(conformer_aware):
    ...
```

maxentile: Addressed in https://github.com/proteneer/timemachine/pull/898/commits/d1ff144ad06052249a89840280c8c4b5c9dbd176

jkausrelay (approved):

PR #899, opened 2022-10-31
Adds cudaSafeMalloc macro

badisa:
* Macro to avoid templating and to provide file/lineno

### DIFF
--- a/timemachine/cpp/src/gpu_utils.cuh
+++ b/timemachine/cpp/src/gpu_utils.cuh
@@ -39,11 +39,11 @@ inline void curandAssert(curandStatus_t code, const char *file, int line, bool a
 */
 #define cudaSafeMalloc(ptr, size)                                                                                      \
     ({                                                                                                                 \
+        const int cudaSafeMalloc__line = __LINE__;                                                                     \
         if (size > (1 << 30)) {                                                                                        \
-            fprintf(stderr, "cudaSafeMalloc: allocation larger than 1GiB %s %d\n", __FILE__, __LINE__);                \
+            fprintf(stderr, "cudaSafeMalloc: allocation larger than 1GiB %s %d\n", __FILE__, cudaSafeMalloc__line);    \
         }                                                                                                              \
-        cudaError_t code = cudaMalloc(ptr, size);                                                                      \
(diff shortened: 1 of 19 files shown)


timemachine/cpp/src/gpu_utils.cuh
```diff
@@ -34,10 +34,22 @@ inline void curandAssert(curandStatus_t code, const char *file, int line, bool a
     }
 }
 
+/* cudaSafeMalloc is equivalent to gpuErrchk(cudaMalloc(...)) except that it prints a warning message if the
+* allocation is greater than a GiB.
+*/
+#define cudaSafeMalloc(ptr, size)                                                                                      \
```
badisa, line 40: I tried to follow the previous style of inlining a function in the macro. However it needs to handle all of the different types. Avoided adding templating, but could do that if it is preferred.

proteneer:
lgtm!

proteneer (approved):

timemachine/cpp/src/gpu_utils.cuh
```diff
@@ -34,10 +34,22 @@ inline void curandAssert(curandStatus_t code, const char *file, int line, bool a
...
+    ({                                                                                                                 \
+        if (size > (1 << 30)) {                                                                                        \
+            fprintf(stderr, "cudaSafeMalloc: allocation larger than 1GiB %s %d\n", __FILE__, __LINE__);                \
+        }                                                                                                              \
+        cudaError_t code = cudaMalloc(ptr, size);                                                                      \
```
mcwitt, line 45: nit: might be preferable to obfuscate `code` as something like `cudaSafeMalloc__code` (or just inline this definition to the next line) to prevent variable capture in the unlikely case that one of the macro args is `code`

badisa: Hmmm makes me think templating is the right way to do it, or at the very least move the cudaMalloc call into the gpuAssert to avoid the collision. Will play around with it.

badisa: Fixed in [ac06ffe](https://github.com/proteneer/timemachine/pull/899/commits/ac06ffe94bd1fa6d1d75fcb235dd6e03d6f8f419)

mcwitt (approved):

PR #900, opened 2022-10-31
Use linear lambda spacing, specialized interpolation for nonbonded decoupling

mcwitt:
This is a step toward independently parametrizing alchemical parameter transformations.

Previously, parameters for all interactions were linearly interpolated as a function of lambda, with the overall lambda spacing being adjusted to optimize overlap.

With #820, we introduced specialized parameter interpolation functions (e.g. log-linear for harmonic force constants) that generate reasonable protocols given linearly-spaced lambda windows, but did not yet modify the overall lambda schedule to be linear because we didn't have the ability to use a custom interpolation function for nonbonded decoupling parameters (which require a nonlinear schedule for reasonable efficiency). The result was suboptimal performance of harmonic force constant protocols, since their interpolation functions are chosen assuming near-linear lambda spacing.

This PR switches to a linear overall lambda spacing, using the existing pre-optimized nonlinear schedule for the nonbonded decoupling parameters only (possible since #893).

## Todo
- [x] Test on hif2a

### DIFF
--- a/tests/test_single_topology_combined.py
+++ b/tests/test_single_topology_combined.py
@@ -4,6 +4,7 @@ import numpy as np
 import pytest
 
 from timemachine.constants import DEFAULT_FF
+from timemachine.fe.interpolate import linear_interpolation
 from timemachine.fe.single_topology import SingleTopology
 from timemachine.fe.system import convert_bps_into_system
 from timemachine.ff import Forcefield
@@ -13,32 +14,42 @@ from timemachine.testsystems.dhfr import get_dhfr_system
 from timemachine.testsystems.relative import get_hif2a_ligand_pair_single_topology
 
 
-def test_combined_parameters_solvent():
+@pytest.fixture(scope="module")
+def hif2a_ligand_pair_single_topology():
+    mol_a, mol_b, core = get_hif2a_ligand_pair_single_topology()
+    forcefield = Forcefield.load_from_file(DEFAULT_FF)
+    return SingleTopology(mol_a, mol_b, core, forcefield)
+
+
+def convert_omm_system(omm_system):
+    bps, masses = openmm_deserializer.deserialize_system(omm_system, cutoff=1.2)
+    num_atoms = len(masses)
+    system = convert_bps_into_system(bps)
+    return system, num_atoms
+
+
+@pytest.fixture(scope="module")
+def complex_host_system():
+    host_sys_omm = get_dhfr_system()
+    return convert_omm_system(host_sys_omm)
+
+
+@pytest.fixture(scope="module")
+def solvent_host_system():
     forcefield = Forcefield.load_from_file(DEFAULT_FF)
     host_sys_omm, _, _, _ = builders.build_water_system(3.0, forcefield.water_ff)
-    _test_combined_parameters_impl_bonded(host_sys_omm)
-    _test_combined_parameters_impl_nonbonded(host_sys_omm)
+    return convert_omm_system(host_sys_omm)
 
 
-def test_combined_parameters_complex():
-    host_sys_omm = get_dhfr_system()
-    _test_combined_parameters_impl_bonded(host_sys_omm)
-    _test_combined_parameters_impl_nonbonded(host_sys_omm)
-
-
-def _test_combined_parameters_impl_bonded(host_system_omm):
+@pytest.mark.parametrize("host_system_fixture", ["solvent_host_system", "complex_host_system"])
+def test_combined_parameters_bonded(host_system_fixture, hif2a_ligand_pair_single_topology, request):
     # test bonded and nonbonded parameters are correct at the end-states.
     # 1) we expected bonded idxs in the ligand to be shifted by num_host_atoms
     # 2) we expected nonbonded lambda_idxs to be shifted
     # 3) we expected nonbonded parameters on the core to be linearly interpolated
 
-    mol_a, mol_b, core = get_hif2a_ligand_pair_single_topology()
-    forcefield = Forcefield.load_from_file(DEFAULT_FF)
-    st = SingleTopology(mol_a, mol_b, core, forcefield)
-
-    host_bps, masses = openmm_deserializer.deserialize_system(host_system_omm, cutoff=1.2)
-    num_host_atoms = len(masses)
-    host_sys = convert_bps_into_system(host_bps)
+    st = hif2a_ligand_pair_single_topology
+    host_sys, num_host_atoms = request.getfixturevalue(host_system_fixture)
 
     def check_bonded_idxs_consistency(bonded_idxs, num_host_idxs):
 
@@ -75,19 +86,15 @@ def _test_combined_parameters_impl_bonded(host_system_omm):
         check_bonded_idxs_consistency(hgs.nonbonded_guest_pairs.get_idxs(), 0)
 
 
-def _test_combined_parameters_impl_nonbonded(host_system_omm):
+@pytest.mark.parametrize("host_system_fixture", ["solvent_host_system", "complex_host_system"])
+def test_combined_parameters_nonbonded(host_system_fixture, hif2a_ligand_pair_single_topology, request):
     # test bonded and nonbonded parameters are correct at the end-states.
     # 1) we expected bonded idxs in the ligand to be shifted by num_host_atoms
     # 2) we expected nonbonded lambda_idxs to be shifted
     # 3) we expected nonbonded parameters on the core to be linearly interpolated
 
-    mol_a, mol_b, core = get_hif2a_ligand_pair_single_topology()
-    forcefield = Forcefield.load_from_file(DEFAULT_FF)
-    st = SingleTopology(mol_a, mol_b, core, forcefield)
-
-    host_bps, masses = openmm_deserializer.deserialize_system(host_system_omm, cutoff=1.2)
-    num_host_atoms = len(masses)
-    host_sys = convert_bps_into_system(host_bps)
+    st = hif2a_ligand_pair_single_topology
+    host_sys, num_host_atoms = request.getfixturevalue(host_system_fixture)
 
     for lamb in [0.0, 1.0]:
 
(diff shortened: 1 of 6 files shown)


timemachine/fe/single_topology.py
```diff
@@ -853,28 +844,6 @@ def __init__(self, mol_a, mol_b, core, forcefield, lambda_angles=0.4, lambda_tor
...
-            lambda_max=lambda_torsions,
-        )
-        self.default_interpolate_periodic_torsion_params_fxn = partial(
-            interpolate_periodic_torsion_params,
-            lambda_min=lambda_torsions,
-            lambda_max=1.0,
-        )
-        self.default_interpolate_nonbonded_params_fxn = interpolate.linear_interpolation
-        self.default_interpolate_chiral_atom_params_fxn = interpolate.linear_interpolation
-        self.default_interpolate_chiral_bond_params_fxn = interpolate.linear_interpolation
```
mcwitt, line 876: Removed this default/override mechanism for now as it wasn't being used and adds a fair amount of clutter. Interpolation functions are for the moment defined inline in `setup_intermediate_state` and `_parameterize_host_guest_nonbonded`.

timemachine/fe/single_topology.py
```diff
@@ -1088,18 +1057,22 @@ def _setup_intermediate_chiral_bond_term(self, src_bond, dst_bond, lamb, interpo
...
+        lamb: float
+
+        lambda_angles, lambda_torsions: float
+            For ring opening/closing transformations, alchemical parameter values controlling the intervals over which
+            bonds, angles, and torsions are interpolated. Note that these have no effect on terms not involved in ring
+            opening/closing.
+
+            - Bonds are interpolated in the interval [0, lambda_angles]
+            - Angles are interpolated in the interval [lambda_angles, lambda_torsions]
+            - Torsions are interpolated in the interval [lambda_torsions, 1]
```
maxentile, line 1075: nit[fool-proofing] -- Clearly documented, but would a couple extra assertions to confirm `1 > lambda_torsions > lambda_angles > 0` be useful?

tests/test_single_topology_combined.py
```diff
@@ -13,32 +13,42 @@
...
+def hif2a_ligand_pair_single_topology():
+    mol_a, mol_b, core = get_hif2a_ligand_pair_single_topology()
     forcefield = Forcefield.load_from_file(DEFAULT_FF)
-    host_sys_omm, _, _, _ = builders.build_water_system(3.0, forcefield.water_ff)
-    _test_combined_parameters_impl_bonded(host_sys_omm)
-    _test_combined_parameters_impl_nonbonded(host_sys_omm)
+    return SingleTopology(mol_a, mol_b, core, forcefield)
+
 
+def convert_omm_system(omm_system):
```
maxentile, line 23: nit[code location] -- looks potentially convenient outside of this test.

Should this instead live in `timemachine.fe.system`, near `convert_bps_into_system`?

maxentile (approved):

tests/test_single_topology_combined.py
mcwitt: Good call. Moved in 313135f and updated `rbfe.py` to use this in 64f18f.

timemachine/fe/single_topology.py
mcwitt: Thanks, added in 84a671d

timemachine/fe/rbfe.py
```diff
@@ -194,9 +196,10 @@ def setup_initial_states(st, host_config, temperature, lambda_schedule, seed):
 
         run_seed = seed + lamb_idx
 
-        if host_config:
+        if host is not None:
```
mcwitt, line 198: This change is to prevent linter errors like "host system may not be defined" (since it's defined conditionally if `host_config`)

tests/test_single_topology_combined.py
```diff
@@ -179,3 +179,42 @@ def _test_combined_parameters_impl_nonbonded(host_system_omm):
...
+        cutoff = hgs.nonbonded_host_guest.get_cutoff()
+
+        assert hgs.nonbonded_host_guest.params is not None
+        guest_params = hgs.nonbonded_host_guest.params[num_host_atoms:]
+        ws_core = [w for flag, (_, _, _, w) in zip(st.c_flags, guest_params) if flag == 0]
+        ws_a = [w for flag, (_, _, _, w) in zip(st.c_flags, guest_params) if flag == 1]
+        ws_b = [w for flag, (_, _, _, w) in zip(st.c_flags, guest_params) if flag == 2]
+
+        # core atoms fixed at w = 0
+        assert all(w == 0.0 for w in ws_core)
```
proteneer, line 205: might be good to check that the charges, sigma, epsilon are correctly set as well (and not just the w-coords), though not necessarily for this PR.

tests/test_single_topology_combined.py
```diff
@@ -179,3 +179,42 @@ def _test_combined_parameters_impl_nonbonded(host_system_omm):
...
+        assert len(set(ws_a)) == 1
+        assert len(set(ws_b)) == 1
+        (w_a,) = set(ws_a)
+        (w_b,) = set(ws_b)
+
+        # w in [0, cutoff]
+        assert 0 < w_a < cutoff
+        assert 0 < w_b < cutoff
+
+        if lamb < 0.5:
```
proteneer, line 217: are we still setting `w_b` to negative values here? (also, this may change if we do double insertion, i.e. at `lamb=0.5` `w_a=w_b=0` in certain cases)

timemachine/fe/single_topology.py
```diff
@@ -1088,19 +1057,26 @@ def _setup_intermediate_chiral_bond_term(self, src_bond, dst_bond, lamb, interpo
...
         """
+
+        assert 0.0 < lambda_angles < lambda_torsions < 1.0
```
proteneer, line 1078: nit: this is technically the open interval, but the docstrings seems to be the closed interval `<=`

timemachine/fe/single_topology.py
```diff
@@ -1109,66 +1085,80 @@ def setup_intermediate_state(
...
         )
 
         return system.VacuumSystem(bond, angle, torsion, nonbonded, chiral_atom, chiral_bond)
 
     def _parameterize_host_guest_nonbonded(self, lamb, host_nonbonded):
         # Parameterize nonbonded potential for the host guest interaction
 
-        guest_exclusions = []
-        guest_scale_factors = []
+        guest_exclusions_ = []
```
proteneer, line 1149: did you intend to add the extra `_` at the end?

proteneer: (is this to avoid aliasing/re-assignment with the guest_exclusions defined below?)

mcwitt: Correct, the motivation was to fix mypy errors of the form

```
error: Incompatible types in assignment (expression has type "ndarray[Any, dtype[floating[_64Bit]]]", variable has type "List[Tuple[float, float]]")
```

(this particular case could actually be reverted; I had changed this when the signature of `_parameterize_host_guest_nonbonded` had an annotated type, which triggered mypy to check the function body

tests/test_single_topology_combined.py
mcwitt: > are we still setting w_b to negative values here?

No, `w_b` is now positive (+cutoff -> 0) as of this PR (see [here](https://github.com/proteneer/timemachine/pull/900/files#diff-0b9bd14a098c7a129a18f14ffb94d84fc8e5c1020d1a55d0e6f53d545c1a4084R1218)).

timemachine/fe/single_topology.py
mcwitt: I might be misunderstanding the comment, but to me the current docstring doesn't indicate whether e.g. `lambda_angles` is allowed to be 0 or equal to `lambda_torsions`. Added a clarifying note in 25fd7aa

timemachine/fe/single_topology.py
mcwitt: Reverted this instance in fd37b16 to minimize the diff since ended up not being strictly necessary

tests/test_single_topology_combined.py
mcwitt: > also, this may change if we do double insertion, i.e. at lamb=0.5 w_a=w_b=0 in certain cases

To clarify, this would pertain to a future version of the SingleTopology class, right? (I don't think we currently have this capability)

timemachine/fe/single_topology.py
proteneer: I'm okay either way! Up to you

timemachine/fe/single_topology.py
proteneer: I think this assertion is fine, I was confused about where exactly the interpolation starts:

```
- Bonds are interpolated in the interval [0, lambda_angles)
- Angles are interpolated in the interval [lambda_angles, lambda_torsions)
- Torsions are interpolated in the interval [lambda_torsions, 1]
```
- bond interpolation is when `0 <= lambda < lambda_angles`  
- angle interpolation is when   `lambda_angles <= lambda < lambda_torsions`
- torsion interpolation is when `lambda_torsions <= torsion <= 1`

(i.e. being clear about `<` vs `<=`)

tests/test_single_topology_combined.py
proteneer: correct - just something to remember later on when we break this test!

proteneer (approved):
lgtm save for nits

timemachine/fe/single_topology.py
mcwitt: Oh, I see. I think the current docstring is correct in the usage of square brackets to indicate closed intervals (based on the implementation of `pad` [here](https://github.com/proteneer/timemachine/blob/fd37b16672994a4605a506c2f3380b6752c51aa8/timemachine/fe/interpolate.py#L175-L183)). I updated the docstring to use inequality notation to hopefully be less ambiguous in c3011a2.

PR #901, opened 2022-11-01
Removes get_U_fn from SingleTopology

badisa:
* Used linear intepolation, which means it is inconsistent with `topo.setup_intermediate_state(lam).get_U_fn()` which is what is used in production


maxentile (approved):

mcwitt (approved):

proteneer (approved):
lgtm

PR #904, opened 2022-11-04
Update minimization frc checks

jkausrelay:
- Use consistent value in checks between minimizer and get_strained_atoms
- Have minimizer raise MinimizationError so users can handle this error specially


proteneer (approved):
lgtm - I think the 25k is pretty arbitrary but seems to be okay so far.

PR #905, opened 2022-11-04
Image Frames from RBFE

badisa:


tests/test_relative_free_energy.py
```diff
@@ -74,10 +74,10 @@ def run_triple(mol_a, mol_b, core, forcefield, n_frames, protein_path, n_eq_step
     assert vacuum_res.overlap_summary_png is not None
     assert vacuum_res.overlap_detail_png is not None
     assert np.linalg.norm(vacuum_res.all_errs) < 0.1
-    assert len(vacuum_res.frames[0] == n_frames)
-    assert len(vacuum_res.frames[-1] == n_frames)
-    assert len(vacuum_res.boxes[0] == n_frames)
-    assert len(vacuum_res.boxes[-1] == n_frames)
+    assert len(vacuum_res.frames[0]) == n_frames
```
badisa, line 80: I noticed while testing these changes that if `.frames[0]` is not a `np.ndarray` this fails.

timemachine/fe/rbfe.py
```diff
@@ -630,11 +648,22 @@ def run_vacuum(
         n_eq_steps=n_eq_steps,
         n_windows=n_windows,
         steps_per_frame=steps_per_frame,
+        image_traj=image_traj,
```
badisa, line 651: Could probably force this to False all the time, as imaging vacuum frames doesn't do much

timemachine/fe/free_energy.py
```diff
@@ -57,6 +58,39 @@ class SimulationResult:
...
+    assert frames.shape[-1] == 3, "Frame coordinates are not 3D"
+    assert boxes.shape[1:] == (3, 3), "Boxes are not 3x3"
+    assert len(frames) == len(boxes), "Number of frames and boxes don't match"
+
+    hb_potential = next(p for p in initial_state.potentials if isinstance(p, HarmonicBond))
+    group_indices = get_group_indices(get_bond_list(hb_potential))
+    imaged_frames = np.empty_like(frames)
+    for i, frame in enumerate(frames):
+        imaged_coords = model_utils.image_frame(group_indices, frame, boxes[i])
+        imaged_frames[i] = imaged_coords
```
jkausrelay, line 90: nit: imaged_frames[i] = model_utils.image_frame(group_indices, frame, boxes[i])

timemachine/fe/rbfe.py
jkausrelay: still nice to have all legs have the same api

jkausrelay (approved):

badisa:
@proteneer  pointed out offline that this only images, doesn't recenter the system around the ligand. Need to do.

timemachine/fe/rbfe.py
```diff
@@ -605,17 +609,31 @@ def estimate_relative_free_energy(
     assert len(keep_idxs) <= len(lambda_schedule)
     combined_prefix = get_mol_name(mol_a) + "_" + get_mol_name(mol_b) + "_" + prefix
     try:
-        return estimate_free_energy_given_initial_states(
+        res = estimate_free_energy_given_initial_states(
             initial_states, protocol, temperature, combined_prefix, keep_idxs
         )
+        if image_traj:
+            for i, idx in enumerate(keep_idxs):
+                res.frames[i] = image_frames(res.initial_states[idx], res.frames[i], res.boxes[i])
```
jkausrelay, line 617: Can you update absolute hydration too: https://github.com/proteneer/timemachine/blob/master/timemachine/fe/absolute_hydration.py#L245 ?

jkausrelay: or move this into estimate_free_energy_given_initial_states

badisa: Moved in https://github.com/proteneer/timemachine/pull/905/commits/afaa10d72ca59511699e7d6bddcf96bc91ff242d

tests/test_fe_utils.py
```diff
@@ -43,6 +43,36 @@ def test_extract_delta_Us_from_U_knk():
...
+def test_image_frame():
+    rng = np.random.default_rng(2022)
+
+    coords = rng.random((90, 3))
+
+    max_dimensions = np.max(coords, axis=0)
+
+    # Add a random buffer to the dimensions of the box
+    box = np.eye(3) * (max_dimensions + rng.random(max_dimensions.shape))
+    group_indices = np.arange(len(coords)).reshape((-1, 3))
```
mcwitt, line 55: nit: idea for strengthening test: group of different sizes, including some empty?

tests/test_relative_free_energy.py
```diff
@@ -183,6 +179,55 @@ def test_steps_per_frames():
         np.testing.assert_array_equal(frame[-1], test_frame[-1])
 
 
+def test_imaging_frames():
+    """Should not be able to run a relative free energy calculation with a single window"""
```
mcwitt, line 183: Is this docstring intended? I don't immediately understand how this relates to the test.

badisa: Nope, copy pasta

tests/test_relative_free_energy.py
```diff
@@ -183,6 +179,55 @@ def test_steps_per_frames():
...
+        assert np.any(np.abs(np.max(res.frames[i], axis=(0, 1))) > box_extents + padding) or np.any(
+            np.min(res.frames[i], axis=(0, 1)) < -padding
+        )
+        # Ligand won't be near center of box
+        assert not np.allclose(np.mean(res.frames[i][0][initial_state.ligand_idxs], axis=0), box_center)
+
+        imaged = image_frames(initial_state, res.frames[i], res.boxes[i])
+
+        # Verify that after imaged, coordinates are within padding of the box extents
+        assert np.all(np.abs(np.max(imaged, axis=(0, 1))) <= box_extents + padding) and np.any(
```
mcwitt, line 224: Should the second condition also be `np.all`?

timemachine/fe/absolute_hydration.py
```diff
@@ -219,6 +220,9 @@ def estimate_absolute_free_energy(
     steps_per_frame: int
         The number of steps to take before collecting a frame
 
+    image_traj: bool
+        Images the trajectories returned with the SimulationResult
```
mcwitt, line 224: nit: I think it'd be helpful to briefly describe what imaging means here

badisa: Fixed in [59f52c1](https://github.com/proteneer/timemachine/pull/905/commits/59f52c120f00e9e81a6bcca2b3eecccc36ac75d9)

tests/test_relative_free_energy.py
badisa: Both fixed in [3b58a0b](https://github.com/proteneer/timemachine/pull/905/commits/3b58a0ba366ae6c2d9a7da2446961aa52271ccee)

tests/test_relative_free_energy.py
badisa: Changed in [3b58a0b](https://github.com/proteneer/timemachine/pull/905/commits/3b58a0ba366ae6c2d9a7da2446961aa52271ccee)

tests/test_fe_utils.py
badisa: Strengthened in [3b58a0b](https://github.com/proteneer/timemachine/pull/905/commits/3b58a0ba366ae6c2d9a7da2446961aa52271ccee)

mcwitt (approved):
Nice, LGTM!

proteneer:
overall this looks good but I would add one more sanity check, that imaging doesn't affect the energies, forces, etc. This is because we're directly modifying the stored trajectories now - and we should triple check the energies etc. (they're probably not going to be bitwise identical, but they should be very close), we may use the store frames later in re-weighting experiments etc. so we should bullet-proof this.

proteneer (changes requested):
add a test for energies, forces, du/dp

badisa:
Not getting the precision I would expect (1e-4), particularly `du_dp` is very different. Trying to track down the issue

tests/test_relative_free_energy.py
```diff
@@ -183,6 +180,82 @@ def test_steps_per_frames():
...
+        imaged = image_frames(initial_state, frames, boxes)
+
+        # Verify that after imaged, coordinates are within padding of the box extents
+        assert np.all(np.max(imaged, axis=(0, 1)) <= box_extents + padding) and np.all(
+            np.min(imaged, axis=(0, 1)) >= -padding
+        )
+        # Verify that ligand was centered in the box
+        np.testing.assert_allclose(np.mean(imaged[0][initial_state.ligand_idxs], axis=0), box_center)
+
+        for precision, rtol, atol in [(np.float64, 1e-8, 1e-8), (np.float32, 7e-3, 5e-4)]:
```
badisa, line 231: Surprised by the large relative tolerance `7e-3` here. Specifically it is the Nonbonded potential that fails here with a lower tolerance.

badisa:
> overall this looks good but I would add one more sanity check, that imaging doesn't affect the energies, forces, etc. This is because we're directly modifying the stored trajectories now - and we should triple check the energies etc. (they're probably not going to be bitwise identical, but they should be very close), we may use the store frames later in re-weighting experiments etc. so we should bullet-proof this.

Based on https://github.com/proteneer/timemachine/pull/905#discussion_r1025777531 it seems like as long as the frames are stored in float64 and float64 precision potentials are used for the re-weighting experiments you would be okay. If you preserve the frames in float32, that is no longer true.

maxentile:
> as long as the frames are stored in float64 and float64 precision potentials are used for the re-weighting experiments you would be okay. If you preserve the frames in float32, that is no longer true.

I don't immediately see that reweighting imposes requirements on this PR.

Revisiting an investigation from last year (calling float32 precision potential impls on frames stored in float16, float32, float64 precision) :

image

Although float64 storage would be required to recompute _energies_ ~losslessly, float32 storage is sufficient for reweighting since (1) the energy errors are small (~0.01 kJ/mol) relative to application requirements / other sources of noise, (2) reweighting only needs _energy differences_, which usually can be recomputed ~losslessly on frames stored in float32.

(Also note that we have since accepted a "linear basis function" optimization https://github.com/proteneer/timemachine/pull/685 , which removes the need to store frames at all to recompute energies at new ligand charge or epsilon parameters.)

proteneer:
This is pretty interesting the more I think about it. It's possible that the precision becomes arbitrarily worse as the particles drift further apart.

Consider computing the difference of 2 numbers, x and y:

let x = 0.1234 and y = 0.2345, their difference y-x=0.1111 can be computed to a very high level of precision. Suppose the two particles are bonded to each other (so they move together).

However,  suppose we run MD for a trillion steps, and now:

let x = 20000000.1234 and y = 20000000.2345, their difference should still be 0.1111 but the floating point difference can be very small. In fact, x ~= 20000000 and any precision is lost. 

Note that I haven't thought too much about:
1) how many steps until this happens
2) how this interacts with fixed point jank

badisa:
@proteneer 

Comparing differences in du_dp and u.

Note that rtol and atol are not quite comparable to absolute and relative difference. Because its atol + (rtol * desired) is how the tests are validated, while the absolute are abs(ref - desired) and relative is abs(diff / desired).

----- f32 Nonbonded du_dp -----
----------------------------------------
Absolute Difference 0.01623864471912384
Relative Difference 0.049392921012479926
Max Value 2758.2305112250615
Min Value -116.04206670634449

----- f32 Nonbonded u -----
----------------------------------------
Absolute Difference 0.022717295651091263
Relative Difference 2.1217467527803133e-07
Max Value -107068.83666169288
Min Value -107068.83666169288

----- f64 Nonbonded du_dp -----
----------------------------------------
Absolute Difference 4.3655745685100555e-11
Relative Difference 2.888259224116654e-10
Max Value 2758.229294277953
Min Value -116.04197075600678

----- f64 Nonbonded u -----
----------------------------------------
Absolute Difference 1.1641532182693481e-10
Relative Difference 1.0872953239176375e-15
Max Value -107068.7229735141
Min Value -107068.7229735141

Comparing the breakdown of the du_dp parameters.

Epsilons are the worst in absolute difference, but charges are worse by relative difference. 

***** f32_Nonbonded_diff_du_dp *****
----- du charges -----
----------------------------------------
Absolute Difference 0.005637174239382148
Relative Difference 0.049392921012479926

----- du eps -----
----------------------------------------
Absolute Difference 0.01623864471912384
Relative Difference 0.00011647463117692889

----- du sig -----
----------------------------------------
Absolute Difference 0.0004909959388896823
Relative Difference 0.00012128080843504878

----- du w -----
----------------------------------------
Absolute Difference 0.0
Relative Difference 0.0

***** f64_Nonbonded_diff_du_dp *****
----- du charges -----
----------------------------------------
Absolute Difference 2.9103830456733704e-11
Relative Difference 2.888259224116654e-10

----- du eps -----
----------------------------------------
Absolute Difference 4.3655745685100555e-11
Relative Difference 3.185900347826155e-13

----- du sig -----
----------------------------------------
Absolute Difference 7.275957614183426e-12
Relative Difference 7.610423668344361e-13

----- du w -----
----------------------------------------
Absolute Difference 0.0
Relative Difference 0.0

We could simply remove the image_frames from the `run_solvent` type methods and call it only when we explicitly want it. That way we have the originals and image as we need.

badisa:
After digging around a bit more, it seems that the number of steps this test takes has an impact on the error.

Also changing the nonbonded potentials to perform certain operations in double rather than float doesn't resolve the issue, instead it makes it worse.

Going to try a few more things before removing the imaging from the `run_*` methods.

maxentile:
>it seems that the number of steps this test takes has an impact on the error.
>Going to try a few more things before removing the imaging from the run_* methods.

Clarification: Is this error susbtantial for du_dx, or just for the du_dp's reported in https://github.com/proteneer/timemachine/pull/905#issuecomment-1324232580 ?

If the precision loss in du_dx is substantial at practical drift distances, then I'd interpret this as an argument for re-imaging into home box during MD, since the precision loss could have unknown impacts on MD quality that could worsen with simulation length.

(Specific scheme would need to be fleshed out, maybe every N steps or every time a molecule gets more than X boxes away from home box...)

badisa:
> > it seems that the number of steps this test takes has an impact on the error.
> > Going to try a few more things before removing the imaging from the run_* methods.
> 
> Clarification: Is this error susbtantial for du_dx, or just for the du_dp's reported in [#905 (comment)](https://github.com/proteneer/timemachine/pull/905#issuecomment-1324232580) ?
> 
> If the precision loss in du_dx is substantial at practical drift distances, then I'd interpret this as an argument for re-imaging into home box during MD, since the precision loss could have unknown impacts on MD quality that could worsen with simulation length.
> 
> (Specific scheme would need to be fleshed out, maybe every N steps or every time a molecule gets more than X boxes away from home box...)

The du_dx has a relative error of 2.5e-4, while du_dp has 4e-2 relative error. This is strictly due to the imaging that takes place.

I don't think we have any sense for the precision loss of MD at the moment, would need to set up a separate test for that. Like what is described here https://github.com/proteneer/timemachine/pull/905#issuecomment-1321431401

badisa:
@maxentile I put together a little test that shifts a vacuum system to see how much it has to shift to change.

https://github.com/proteneer/timemachine/commit/99c0dd2681f9091c3ce1867e72c199d9010f5da0

After just 3 boxes the float32 implementation is no longer stable at our standard float 32 tolerances rtol=1e-4 atol=5e-4

Which would seem to indicate that imaging after particles are more than 2 boxes away would be a good starting point.

maxentile:
Nice @badisa  -- that's highly informative! 2 boxes seems a good starting point. (Also that would be infrequent enough that any slow-down would be negligible.)

proteneer:
> After just 3 boxes the float32 implementation is no longer stable at our standard float 32 tolerances rtol=1e-4 atol=5e-4

Interesting observation! To be clear, is it just all three terms (U, du_dx, du_dp) that exhibits the drift behavior?

badisa:
> > After just 3 boxes the float32 implementation is no longer stable at our standard float 32 tolerances rtol=1e-4 atol=5e-4
> 
> Interesting observation! To be clear, is it just all three terms (U, du_dx, du_dp) that exhibits the drift behavior?

For posteriority, it is all three terms. Though U is less sensitive then du_dx, du_dp.

proteneer (approved):

PR #908, opened 2022-11-07
Check compatibility of atom map with chiral endstates

maxentile:
Defines a filter that can be applied to a trial atom mapping, to determine if it is in some sense inconsistent with the set of chiral atom restraints defined on the endstates A and B.

Inconsistency can currently take two forms: "flipped" (the atom mapping would flip the sign of chiral restraints defined in both endstates) or "undefined" (the atom mapping includes a chiral restraint that is defined in one endstate but not the other).

TODOs for this PR:
- [x] Expand tests
- [x] Decide defaults (PR currently changes default behavior of `get_core_by_mcs`, now filtering out "flipped" cases by default)
- [x] Clean-up: `"flipped" in message` -> `status.is_flipped`, `"undefined" in message` -> `status.is_undefined` or similar
- [x] Clean-up: Avoid 2x duplication in `find_atom_map_chiral_conflicts`

TODOs for later:
- [ ] Consider changing definition of `find_chiral_atoms`. (This PR does not change timemachine's current definition.)
- [ ] Consider checking consistency on bonds as well
- [ ] Consider porting to https://github.com/proteneer/bgl

### DIFF
--- a/tests/test_chiral_utils.py
+++ b/tests/test_chiral_utils.py
@@ -165,12 +165,15 @@ def test_chiral_flip_check():
     swap_map[1, 1] = 2
     swap_map[2, 1] = 1
 
-    identity_conflicts = find_atom_map_chiral_conflicts(identity_map, chiral_set_a, chiral_set_b)
-    assert len(identity_conflicts) == 0
+    identity_flips = find_atom_map_chiral_conflicts(identity_map, chiral_set_a, chiral_set_b, mode="flip")
+    identity_undefineds = find_atom_map_chiral_conflicts(identity_map, chiral_set_a, chiral_set_b, mode="undefined")
+    assert len(identity_flips) == 0
+    assert len(identity_undefineds) == 0
 
-    swap_conflicts = find_atom_map_chiral_conflicts(swap_map, chiral_set_a, chiral_set_b)
-    assert len(swap_conflicts) == 8  # TODO: deduplicate idxs?
-    assert all("flipped" in msg.lower() for msg in swap_conflicts)
+    swap_map_flips = find_atom_map_chiral_conflicts(swap_map, chiral_set_a, chiral_set_b, mode="flip")
+    swap_map_undefineds = find_atom_map_chiral_conflicts(swap_map, chiral_set_a, chiral_set_b, mode="undefined")
+    assert len(swap_map_flips) == 8  # TODO: deduplicate idxs?
+    assert len(swap_map_undefineds) == 0
 
     # test maps where atom chirality is defined in one endstate, undefined in other
     mol_b = Chem.AddHs(Chem.MolFromSmiles("N"))
@@ -180,6 +183,7 @@ def test_chiral_flip_check():
     assert len(chiral_set_b.restr_idxs) == 0
 
     partial_map = identity_map[:4]
-    partial_conflicts = find_atom_map_chiral_conflicts(partial_map, chiral_set_a, chiral_set_b)
-    assert len(partial_conflicts) > 0
-    assert all("undefined" in msg.lower() for msg in partial_conflicts)
+    partial_map_flips = find_atom_map_chiral_conflicts(partial_map, chiral_set_a, chiral_set_b, mode="flip")
+    partial_map_undefineds = find_atom_map_chiral_conflicts(partial_map, chiral_set_a, chiral_set_b, mode="undefined")
+    assert len(partial_map_flips) == 0
+    assert len(partial_map_undefineds) > 0
--- a/tests/test_chiral_utils.py
+++ b/tests/test_chiral_utils.py
@@ -141,9 +141,9 @@ def test_find_chiral_bonds():
     assert res == set([(0, 2)])
 
 
-def test_chiral_flip_check():
+def test_chiral_conflict_flip():
+    # exercise case of no conflicts or a flip conflict
 
-    # test maps that can swap "chirality of methane"
     mol_a = Chem.AddHs(Chem.MolFromSmiles("C"))
     mol_b = Chem.AddHs(Chem.MolFromSmiles("C"))
 
@@ -175,15 +175,57 @@ def test_chiral_flip_check():
     assert len(swap_map_flips) == 8  # TODO: deduplicate idxs?
     assert len(swap_map_undefineds) == 0
 
-    # test maps where atom chirality is defined in one endstate, undefined in other
+
+def test_chiral_conflict_undefined():
+    # exercise case where atom chirality is defined in one endstate, undefined in other
+
+    mol_a = Chem.AddHs(Chem.MolFromSmiles("C"))
     mol_b = Chem.AddHs(Chem.MolFromSmiles("N"))
+
+    AllChem.EmbedMolecule(mol_a, randomSeed=0)
     AllChem.EmbedMolecule(mol_b, randomSeed=0)
+
+    conf_a = mol_a.GetConformer(0).GetPositions()
     conf_b = mol_b.GetConformer(0).GetPositions()
+
+    chiral_set_a = ChiralRestrIdxSet.from_mol(mol_a, conf_a)
     chiral_set_b = ChiralRestrIdxSet.from_mol(mol_b, conf_b)
+
+    assert len(chiral_set_a.restr_idxs) == 4
     assert len(chiral_set_b.restr_idxs) == 0
 
-    partial_map = identity_map[:4]
+    partial_map = np.array([(i, i) for i in range(4)])
     partial_map_flips = find_atom_map_chiral_conflicts(partial_map, chiral_set_a, chiral_set_b, mode="flip")
     partial_map_undefineds = find_atom_map_chiral_conflicts(partial_map, chiral_set_a, chiral_set_b, mode="undefined")
     assert len(partial_map_flips) == 0
     assert len(partial_map_undefineds) == 1
+
+
+def test_chiral_conflict_mixed():
+    # test case containing both a flip and a partial undefined
+
+    mol_a = Chem.AddHs(Chem.MolFromSmiles("CC"))
+    mol_b = Chem.AddHs(Chem.MolFromSmiles("CN"))
+
+    AllChem.EmbedMolecule(mol_a, randomSeed=0)
+    AllChem.EmbedMolecule(mol_b, randomSeed=0)
+
+    conf_a = mol_a.GetConformer(0).GetPositions()
+    conf_b = mol_b.GetConformer(0).GetPositions()
+
+    chiral_set_a = ChiralRestrIdxSet.from_mol(mol_a, conf_a)
+    chiral_set_b = ChiralRestrIdxSet.from_mol(mol_b, conf_b)
+
+    assert len(chiral_set_a.restr_idxs) == 8
+    assert len(chiral_set_b.restr_idxs) == 4
+
+    mixed_map = np.array([[i, i] for i in range(mol_b.GetNumAtoms())])
+
+    mixed_map[2, 0] = 3
+    mixed_map[3, 0] = 2
+
+    mixed_map_flips = find_atom_map_chiral_conflicts(mixed_map, chiral_set_a, chiral_set_b, mode="flip")
+    mixed_map_undefineds = find_atom_map_chiral_conflicts(mixed_map, chiral_set_a, chiral_set_b, mode="undefined")
+
+    assert len(mixed_map_flips) == 8
+    assert len(mixed_map_undefineds) == 1
--- a/tests/test_chiral_utils.py
+++ b/tests/test_chiral_utils.py
@@ -234,6 +234,7 @@ def test_chiral_conflict_flip():
 
     identity_map = np.array([(i, i) for i in range(len(conf_a))])
 
+    # swap any pair of atoms around a tetrahedral center to flip chirality
     swap_map = np.array(identity_map)
     swap_map[1, 1] = 2
     swap_map[2, 1] = 1
@@ -288,6 +289,7 @@ def test_chiral_conflict_mixed():
 
     mixed_map = np.array([[i, i] for i in range(mol_b.GetNumAtoms())])
 
+    # swap any pair of atoms around a tetrahedral center to flip chirality
     mixed_map[2, 0] = 3
     mixed_map[3, 0] = 2
 
--- a/tests/test_chiral_utils.py
+++ b/tests/test_chiral_utils.py
@@ -186,4 +186,4 @@ def test_chiral_flip_check():
     partial_map_flips = find_atom_map_chiral_conflicts(partial_map, chiral_set_a, chiral_set_b, mode="flip")
     partial_map_undefineds = find_atom_map_chiral_conflicts(partial_map, chiral_set_a, chiral_set_b, mode="undefined")
     assert len(partial_map_flips) == 0
-    assert len(partial_map_undefineds) > 0
+    assert len(partial_map_undefineds) == 1
--- a/tests/test_chiral_utils.py
+++ b/tests/test_chiral_utils.py
@@ -4,6 +4,7 @@ from rdkit import Chem
 from rdkit.Chem import AllChem
 
 from timemachine.fe import chiral_utils, utils
+from timemachine.fe.chiral_utils import ChiralRestrIdxSet, find_atom_map_chiral_conflicts
 from timemachine.potentials.chiral_restraints import U_chiral_atom_batch, U_chiral_bond_batch
 
 pytestmark = [pytest.mark.nogpu]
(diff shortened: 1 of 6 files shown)


maxentile:
Ready for initial review of `find_atom_map_chiral_conflicts` --

(1) are these the right conditions to check?
(2) is the approach compatible with intended implementation in https://github.com/proteneer/bgl ?
(3) do we have more difficult test cases in mind?

(Re: (2) -- the checks just involve array indexing and set inclusion, which should be trivial to port to a higher-performance implementation if needed, e.g. if similar logic should be part of a C++ callback that will be invoked many times during a backtracking search. However, I might need to modify the implementation to be able to check more general candidate partial solutions represented by a `marcs` matrix as of https://github.com/proteneer/bgl/pull/4 , rather than only checking proposed atom pairings from a `core` array. My default approach would initially be to ignore any additional information provided by the partial solution `marcs` matrix beyond "these atom pairs are definitely included in this partial solution".)

tests/test_chiral_utils.py
```diff
@@ -138,3 +139,51 @@ def test_find_chiral_bonds():
...
+
+
+def test_chiral_flip_check():
+
+    # test maps that can swap "chirality of methane"
+    mol_a = Chem.AddHs(Chem.MolFromSmiles("C"))
+    mol_b = Chem.AddHs(Chem.MolFromSmiles("C"))
+
+    AllChem.EmbedMolecule(mol_a, randomSeed=0)
+    AllChem.EmbedMolecule(mol_b, randomSeed=0)
```
proteneer, line 151: is the intent here that they have the same conformation? (may be better to just copy over the conformer for clarity).

tests/test_chiral_utils.py
```diff
@@ -138,3 +139,51 @@ def test_find_chiral_bonds():
...
+    conf_b = mol_b.GetConformer(0).GetPositions()
+    chiral_set_b = ChiralRestrIdxSet.from_mol(mol_b, conf_b)
+    assert len(chiral_set_b.restr_idxs) == 0
+
+    partial_map = identity_map[:4]
+    partial_map_flips = find_atom_map_chiral_conflicts(partial_map, chiral_set_a, chiral_set_b, mode="flip")
+    partial_map_undefineds = find_atom_map_chiral_conflicts(partial_map, chiral_set_a, chiral_set_b, mode="undefined")
+    assert len(partial_map_flips) == 0
+    assert len(partial_map_undefineds) > 0
```
proteneer, line 189: should we set to the actual length? not sure why we prefer `> 0` here.

proteneer, line 189: I would add one more mixed test with the following mapping that mixes undefined with swaps:

```
[[0,0],
[1,2],
[3,3],
[4,4]]
```
this way we have both undefined and swaps

timemachine/fe/atom_mapping.py
```diff
@@ -169,10 +178,35 @@ def get_core_by_mcs(
...
+        undefineds = find_atom_map_chiral_conflicts(trial_core, chiral_set_a, chiral_set_b, mode="undefined")
+
+        num_flips = len(flips)
+        num_undefineds = len(undefineds)
+
+        # TODO: de-DeMorgan-ify
+        invalid_due_to_flips = (not allow_chiral_atom_flips) and (num_flips > 0)
+        invalid_due_to_undefineds = (not allow_chiral_atom_undefined) and (num_undefineds > 0)
+        invalid = invalid_due_to_flips or invalid_due_to_undefineds
+        valid = not invalid
```
proteneer, line 198: lol

proteneer (commented):
I need to do another pass on this - I have one high level question, is it possible to have this operate in an incremental manner? Currently the approach would work for leaf nodes given the MCS, but it would be great if we could also support this (cheaply) for two reasons:

1) it may be "too-late" by the time we reach the leaf node, i.e. all the maximal cores fail the chirality check, but there may have been an internal node that did not fail the chiral check

2) we can increase the pruning efficiency as well, currently this isn't a big issue because we can search through the tree so quickly, however, if we add a chiral check to internal nodes this may slow things down dramatically.

The mental picture I have in mind is this:

Suppose we have two molecules `mol_a, mol_b`, and we have an existing core of size `[k,2]`. Now we want to propose a new mapping (x,y) to append to the core. We don't want to have to re-check all of the `[k,2]` core atoms, but only the ones affected by (x,y):

a) the new atoms (x,y) would need to be checked
b) any adjacent nbs_x to x, and adjacent nbs_y to y, where nbs_x in core, and nbs_y in core, would need to be re-verified

But this should still be substantially faster than checking the entire core.

tests/test_chiral_utils.py
maxentile: will fix -- no reason to prefer `> 0`

tests/test_chiral_utils.py
maxentile: Do you have a preferred idiom for copying a conformer from one mol to another?

(First method I tried https://github.com/rdkit/rdkit/issues/5116 causes memory errors due to a low-priority object lifetime bug in rdkit, and it seemed simpler to create two independent mols than to copy conformers safely.)

tests/test_chiral_utils.py
maxentile: Good call -- would strengthen test!

timemachine/fe/atom_mapping.py
maxentile: a bit convoluted, will simplify

maxentile:
>I have one high level question, is it possible to have this operate in an incremental manner? 
>we have an existing core of size [k,2]. Now we want to propose a new mapping (x,y) to append to the core. We don't want to have to re-check all of the [k,2] core atoms, but only the ones affected by (x,y):

I think so -- can make this part incremental, rather than always looping over all restraint idxs : https://github.com/proteneer/timemachine/pull/908/files/f05b16d81e666bdd95fc228f3b1da1da661600f8#diff-cc54470389f7198a698e6ff9a5a6d7886cf32a4dc9c96bdd7a6c2f25d161e421R221-R222

(There are some easier performance optimizations too, if needed.)

tests/test_chiral_utils.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/908/commits/a8cedb3bd8059b5cd24d13c30e53120d101a83cc

maxentile: I didn't see a way to mix both of these cases using any mapping of this specific molecule pair, but added a test that mixes both cases in https://github.com/proteneer/timemachine/pull/908/commits/04ba0f0806ac5a9430444f2ea5c9a35f3d1a7c58

timemachine/fe/atom_mapping.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/908/commits/68cbae755248cbd306444fcfda290c45818a5042

tests/test_chiral_utils.py
proteneer: hm good point - i was going to suggest just embedding a pre-generated SDF perhaps, but maybe this isn't a big deal.

tests/test_chiral_utils.py
proteneer: your change makes sense, I think was thinking of setting up a case of sp3->sp3 but the core is only of size 4, and one of them is mismatched (eg. a down H matching an up H), so this wouldn't be "undefined" by rather "missing"

tests/test_chiral_utils.py
proteneer: I'm more inclined to just hard code an SDF here - since the conformer generation tool in the RDKit can vary between versions. Eg. rdkit 2022 generates an R version, but when we do an upgrade later on it instead generates an S version.

maxentile: Addressed in https://github.com/proteneer/timemachine/pull/908/commits/9eb5f255e8d16f9b1f15724bc6725a93baa4a00a

maxentile:
>I have one high level question, is it possible to have this operate in an incremental manner?
we have an existing core of size [k,2]. Now we want to propose a new mapping (x,y) to append to the core. We don't want to have to re-check all of the [k,2] core atoms, but only the ones affected by (x,y):

Clarification: Is the performance comment re: incremental checks blocking for this PR? Can this be investigated in a follow-up PR targeting [`proteneer/bgl:priority_list`](https://github.com/proteneer/bgl/pull/8) (or a follow-up PR into `timemachine` after the implementation being developed in `bgl` is merged into `timemachine`)?

(I would prefer to address performance in a separate PR if possible, guided by profiling in context. Making the check incremental etc. will add complexity and may or may not pay off performance-wise depending on integration context. Currently the check is slower than it needs to be -- around 0.1 milliseconds per candidate core -- but if just applied as a post-filter to a small number of candidate cores, this does not currently seem unacceptable to me. If integrated into the inner search loop of `proteneer/bgl:priority_list` it may cause unacceptable performance (aka exceeding 1 second when the number of cores to check exceeds ~10000). In the context of the inner search loop of  `proteneer/bgl:priority_list`, many lower-hanging-fruit performance optimizations are possible (just returning a bool rather than returning the whole list of tuples, avoiding some repeated work per check, avoiding unnecessary conversions from `bgl`'s internal representations to a numpy array of ints, etc.).)

tests/test_chiral_utils.py
```diff
@@ -138,3 +139,160 @@ def test_find_chiral_bonds():
...
+
+    chiral_set_a = ChiralRestrIdxSet.from_mol(mol_a, conf_a)
+    chiral_set_b = ChiralRestrIdxSet.from_mol(mol_b, conf_b)
+
+    assert len(chiral_set_a.restr_idxs) == 4
+    assert len(chiral_set_b.restr_idxs) == 4
+
+    identity_map = np.array([(i, i) for i in range(len(conf_a))])
+
+    swap_map = np.array(identity_map)
```
jkausrelay, line 237: nit: Add comment # swap hydrogens to flip chirality

tests/test_chiral_utils.py
```diff
@@ -138,3 +139,160 @@ def test_find_chiral_bonds():
...
+
+    chiral_set_a = ChiralRestrIdxSet.from_mol(mol_a, conf_a)
+    chiral_set_b = ChiralRestrIdxSet.from_mol(mol_b, conf_b)
+
+    assert len(chiral_set_a.restr_idxs) == 8
+    assert len(chiral_set_b.restr_idxs) == 4
+
+    mixed_map = np.array([[i, i] for i in range(mol_b.GetNumAtoms())])
+
+    mixed_map[2, 0] = 3
```
jkausrelay, line 291: nit: # swap hydrogens part of methyl group to flip chirality

timemachine/fe/chiral_utils.py
```diff
@@ -132,6 +142,131 @@ def find_chiral_atoms(mol):
...
+    core
+        atom map, establishing correspondences
+            mol_a[a_i] <-> mol_b[b_i]
+        for (a_i, b_i) in core
+
+    chiral_set_a, chiral_set_b
+        chiral restraint sets for mols a and b
+
+    mode : str, one of "flip" or "undefined"
+        "flip" : find cases where chiral atom restraints are defined for both a and b with opposite signs
```
jkausrelay, line 250: nit: for both mols a and b

timemachine/fe/chiral_utils.py
```diff
@@ -132,6 +142,131 @@ def find_chiral_atoms(mol):
...
+
+    def disallows(self, trial_tuple: FourTuple) -> bool:
+        return trial_tuple in self.disallowed_set
+
+
+def _find_atom_map_chiral_conflicts_one_direction(
+    core: np.ndarray,
+    chiral_set_a: ChiralRestrIdxSet,
+    chiral_set_b: ChiralRestrIdxSet,
+    mode="flip",
```
jkausrelay, line 201: nit: You could create constants for the modes (CHIRAL_MODE_FLIP, CHIRAL_MODE_UNDEFINED), since they are used repeatedly.

timemachine/fe/chiral_utils.py
```diff
@@ -132,6 +142,131 @@ def find_chiral_atoms(mol):
...
+
+
+def find_atom_map_chiral_conflicts(
+    core: np.ndarray,
+    chiral_set_a: ChiralRestrIdxSet,
+    chiral_set_b: ChiralRestrIdxSet,
+    mode="flip",
+) -> Set[Tuple[FourTuple, FourTuple]]:
+    """
+
```
jkausrelay, line 238: Add a note that chiral here treats all atoms as distinguishable (methane is chiral).

jkausrelay (commented):
LGTM, minor comments

tests/test_chiral_utils.py
maxentile: Added in https://github.com/proteneer/timemachine/pull/908/commits/48f909e92fef38757cfc002b3ff10ca62e131a0f

tests/test_chiral_utils.py
maxentile: Added in https://github.com/proteneer/timemachine/pull/908/commits/48f909e92fef38757cfc002b3ff10ca62e131a0f

timemachine/fe/chiral_utils.py
maxentile: Added references to definition + more context for why we adopted this definition in https://github.com/proteneer/timemachine/pull/908/commits/71f83c012c6da720c18503adaa8bf5610977295f

proteneer:
> > I have one high level question, is it possible to have this operate in an incremental manner?
> > we have an existing core of size [k,2]. Now we want to propose a new mapping (x,y) to append to the core. We don't want to have to re-check all of the [k,2] core atoms, but only the ones affected by (x,y):
> 
> Clarification: Is the performance comment re: incremental checks blocking for this PR? Can this be investigated in a follow-up PR targeting [`proteneer/bgl:priority_list`](https://github.com/proteneer/bgl/pull/8) (or a follow-up PR into `timemachine` after the implementation being developed in `bgl` is merged into `timemachine`)?
> 
> (I would prefer to address performance in a separate PR if possible, guided by profiling in context. Making the check incremental etc. will add complexity and may or may not pay off performance-wise depending on integration context. Currently the check is slower than it needs to be -- around 0.1 milliseconds per candidate core -- but if just applied as a post-filter to a small number of candidate cores, this does not currently seem unacceptable to me. If integrated into the inner search loop of `proteneer/bgl:priority_list` it may cause unacceptable performance (aka exceeding 1 second when the number of cores to check exceeds ~10000). In the context of the inner search loop of `proteneer/bgl:priority_list`, many lower-hanging-fruit performance optimizations are possible (just returning a bool rather than returning the whole list of tuples, avoiding some repeated work per check, avoiding unnecessary conversions from `bgl`'s internal representations to a numpy array of ints, etc.).)

I think that's fine. As is this is a big improvement already :)

proteneer (approved):

timemachine/fe/chiral_utils.py
maxentile: The suggestion to use a named constant instead of a string is good, but this setting seems too specific to add to global constants.py -- added an enum instead https://github.com/proteneer/timemachine/pull/908/commits/3e8738118197b6826f1715a0d3d426b67bda33c8

(and removed an option in `get_core_by_mcs`, to better enforce reasonable default)

timemachine/fe/chiral_utils.py
```diff
@@ -132,6 +142,131 @@ def find_chiral_atoms(mol):
...
+            # swaps
+            disallowed_set.add((center, i, k, j))
+            disallowed_set.add((center, j, i, k))
+            disallowed_set.add((center, k, j, i))
+
+        assert allowed_set.isdisjoint(disallowed_set)
+
+        return allowed_set, disallowed_set
+
+    def __post_init__(self):
```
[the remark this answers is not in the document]
maxentile: Adjusted in https://github.com/proteneer/timemachine/pull/908/commits/6e71f8afad1a7fc5bf57c1f81331cb38fcda3665

timemachine/fe/chiral_utils.py
```diff
@@ -132,6 +147,136 @@ def find_chiral_atoms(mol):
...
+            disallowed_set.add((center, i, k, j))
+            disallowed_set.add((center, j, i, k))
+            disallowed_set.add((center, k, j, i))
+
+        assert allowed_set.isdisjoint(disallowed_set)
+
+        return allowed_set, disallowed_set
+
+    def defines(self, trial_tuple: FourTuple) -> bool:
+        return (trial_tuple in self.allowed_set) or (trial_tuple in self.disallowed_set)
```
mcwitt, line 194: minor nit: would it be slightly more efficient to represent the allowed and disallowed sets as a `is_allowed: Dict[FourTuple, bool]`? (e.g. `idxs in allowed_set or idxs in disallowed_set === idxs in is_allowed` and `idxs in disallowed_set === not is_allowed.get(idxs)`

tests/test_single_topology.py
```diff
@@ -39,6 +39,8 @@ def test_phenol():
     are present when either a single root anchor is provided, or when a root anchor and a neighbor anchor is provided.
     """
     mol = Chem.AddHs(Chem.MolFromSmiles("c1ccccc1O"))
+    AllChem.EmbedMolecule(mol, randomSeed=2022)
```
mcwitt, line 42: Good to remember that `EmbedMolecule` is nondeterministic! (just noticing that we have several instances in this in other tests that do not pass a seed)

mcwitt (approved):
LGTM! Only non-blocking minor comments/curiosities

timemachine/fe/chiral_utils.py
maxentile: Hmm, haven't profiled for efficiency differences like that yet.

Sets seem marginally simpler to read than dicts (the only important thing added by this PR is [`trial_tuple in self.disallowed_set`](https://github.com/proteneer/timemachine/pull/908/files/7787f1e2a9036da81ec363b9ab5a524912a568f0#diff-cc54470389f7198a698e6ff9a5a6d7886cf32a4dc9c96bdd7a6c2f25d161e421R197)), but either spelling would be easily readable imo.

I would have a slight preference to leave this as a set for now, and defer performance to a second (profile-guided) pass, unless there's a readability advantage to switching now.

mcwitt: Sounds good!

tests/test_single_topology.py
maxentile: Based on a quick scan I didn't see any existing examples in the test where non-deterministic conformer generation would cause a problem, but good to keep in mind.

(This PR did introduce a test with a dependence on an aspect of the generated conformer that we have no reason to be conserved across RDKit updates, so it was important to freeze the conformer used by this PR's new tests -- https://github.com/proteneer/timemachine/pull/908#discussion_r1020432268 )

PR #911, opened 2022-11-10
Reduce memory usage in RBFE example script

mcwitt:
Previously, references to results for all edges were retained in a list of futures. This change avoids maintaining references to futures by moving the saving of results into the concurrently-executed function. Overall memory usage should be reduced from O(edges) to O(max_workers).

Note: this reduces the granularity of the concurrency by a factor of 2; here we parallelize over edges instead of over (edge, leg) pairs as we did previously.

### DIFF
--- a/examples/rbfe_edge_list.py
+++ b/examples/rbfe_edge_list.py
@@ -26,6 +26,65 @@ def get_mol_by_name(mols, name):
     assert 0, "Mol not found"
 
 
+def run_edge_and_save_results(
+    run_complex_leg_fn,
+    run_solvent_leg_fn,
+    mol_a,
+    mol_b,
+    core,
+    forcefield,
+    protein,
+    n_frames,
+    seed,
+    smarts,
+    exp_ddg,
+    fep_ddg,
+    fep_ddg_err,
+    ccc_ddg,
+    ccc_ddg_err,
+):
+    mol_a_name = get_mol_name(mol_a)
+    mol_b_name = get_mol_name(mol_b)
+    args = mol_a, mol_b, core, forcefield, protein, n_frames, seed
+
+    try:
+        complex_res, complex_top = run_complex_leg_fn(*args)
+        solvent_res, solvent_top = run_solvent_leg_fn(*args)
+
+        meta = (
+            mol_a,
+            mol_b,
+            smarts,
+            core,
+            float(exp_ddg) * KCAL_TO_KJ,
+            float(fep_ddg) * KCAL_TO_KJ,
+            float(fep_ddg_err) * KCAL_TO_KJ,
+            float(ccc_ddg) * KCAL_TO_KJ,
+            float(ccc_ddg_err) * KCAL_TO_KJ,
+        )
+
+        with open(f"success_rbfe_result_{mol_a_name}_{mol_b_name}.pkl", "wb") as fh:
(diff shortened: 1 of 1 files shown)


examples/rbfe_edge_list.py
```diff
@@ -59,65 +112,31 @@ def read_from_args():
         reader = csv.reader(csvfile, delimiter=",")
         next(reader)
         rows = [row for row in reader]
-        for row_idx, row in enumerate(rows):
+        for row in rows:
```
mcwitt, line 115: Small orthogonal change: seemed simpler to use the same seed for each edge. Can add this back in if there's a good reason.

mcwitt: Added this back in to reduce the diff.

examples/rbfe_edge_list.py
```diff
@@ -26,6 +26,62 @@ def get_mol_by_name(mols, name):
...
+            float(exp_ddg) * KCAL_TO_KJ,
+            float(fep_ddg) * KCAL_TO_KJ,
+            float(fep_ddg_err) * KCAL_TO_KJ,
+            float(ccc_ddg) * KCAL_TO_KJ,
+            float(ccc_ddg_err) * KCAL_TO_KJ,
+        )
+
+        with open(f"success_rbfe_result_{mol_a_name}_{mol_b_name}.pkl", "wb") as fh:
+            pkl_obj = (meta, solvent_res, solvent_top, complex_res, complex_top)
+            pickle.dump(pkl_obj, fh)
```
jkausrelay, line 65: If the submit goes to a remote host this will write the results to the remote host and not be pulled back to the launcher host.

examples/rbfe_edge_list.py
```diff
@@ -64,60 +116,26 @@ def read_from_args():
...
+                protein,
+                args.n_frames,
+                args.seed + row_idx,
+                smarts,
+                exp_ddg,
+                fep_ddg,
+                fep_ddg_err,
+                ccc_ddg,
+                ccc_ddg_err,
             )
```
jkausrelay, line 138: fut.result() blocks until the subjob completes. If the process exits before the subjobs complete, I'm not sure what will happen.

mcwitt: In my testing the process has stayed open until subprocesses complete. But maybe best to add an explicit call to `.result()`?

examples/rbfe_edge_list.py
mcwitt: Good point; for now I don't think this is an issue because this code is specialized to `CUDAPoolClient`, which uses local processes. But I was planning to make this abstract over the client as a next step to allow running on remote hosts. Do you think it makes sense to address this then?

examples/rbfe_edge_list.py
jkausrelay: Yeah it seems ProcessPoolExecutor will wait, but it might be better to be explicit.

examples/rbfe_edge_list.py
jkausrelay: That's fine too.

jkausrelay (approved):

proteneer (approved):
lgtm!

examples/rbfe_edge_list.py
mcwitt: Addressed in 7addb66

PR #913, opened 2022-11-11
Move RBFE driver into module, make client-agnostic

mcwitt:
* Separates command- and file-parsing logic from simulation driver code; moves the latter into the `timemachine.fe.rbfe` library module (as `run_edges_parallel`)
* Adds `pool_client: AbstractClient` and `file_client: AbstractFileClient` args to `run_edges_parallel`; this allows running e.g. on remote machines or AWS Batch by substituting different clients
* Deletes references to future objects as they complete[^1]; this should ensure that they are GC'd and memory is returned to the process pool ~(0770b54 .. c4942fc, [relevant SO thread](https://stackoverflow.com/questions/34770169/using-concurrent-futures-without-running-out-of-ram))~

[^1]: this is only approximate; we actually process futures in the order that they were submitted (processing in order of completion would require additional API changes for custom future objects)

### DIFF
--- a/examples/rbfe_edge_list.py
+++ b/examples/rbfe_edge_list.py
@@ -2,6 +2,8 @@ import argparse
 import csv
 from typing import List
 
+from simtk.openmm import app
+
 from timemachine.constants import KCAL_TO_KJ
 from timemachine.fe import rbfe
 from timemachine.fe.utils import read_sdf
@@ -52,6 +54,7 @@ if __name__ == "__main__":
     ligands = read_sdf(args.ligands)
     edges = read_edges_csv(args.results_csv)
     forcefield = Forcefield.load_from_file(args.forcefield)
+    protein = app.PDBFile(str(args.protein))
 
     _ = rbfe.run_edges_parallel(
         ligands,
--- a/examples/rbfe_edge_list.py
+++ b/examples/rbfe_edge_list.py
@@ -57,11 +57,11 @@ if __name__ == "__main__":
     protein = app.PDBFile(str(args.protein))
 
     paths = rbfe.run_edges_parallel(
-        args.n_frames,
         ligands,
         edges,
         forcefield,
         args.protein,
+        args.n_frames,
         args.n_gpus,
         args.seed,
     )
--- a/examples/rbfe_edge_list.py
+++ b/examples/rbfe_edge_list.py
@@ -2,8 +2,6 @@ import argparse
 import csv
 from typing import List
 
-# HACK: import rdkit first to work around free(): invalid pointer
-import rdkit  # noqa: F401
 from simtk.openmm import app
 
 from timemachine.constants import KCAL_TO_KJ
--- a/examples/rbfe_edge_list.py
+++ b/examples/rbfe_edge_list.py
@@ -1,85 +1,6 @@
 import argparse
-import csv
-import pickle
-import traceback
 
-import numpy as np
-from rdkit import Chem
-
-Chem.SetDefaultPickleProperties(Chem.PropertyPickleOptions.AllProps)
-
-from simtk.openmm import app
-
-from timemachine.constants import KCAL_TO_KJ
-from timemachine.fe import atom_mapping
-from timemachine.fe.rbfe import run_complex, run_solvent
-from timemachine.fe.utils import get_mol_name, read_sdf
-from timemachine.ff import Forcefield
-from timemachine.parallel.client import CUDAPoolClient
-
-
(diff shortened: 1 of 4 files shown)


jkausrelay (approved):

timemachine/fe/rbfe.py
```diff
@@ -679,3 +686,119 @@ def run_complex(
...
+                seed + edge_idx,
+                file_client,
+            )
+        )
+
+    # Block until jobs finish
+    paths = []
+    for fut in futures.as_completed(jobs):
+        paths.append(fut.result())
+        del fut  # allow future to be GC'd
```
mcwitt, line 802: TODO: fix this (doesn't delete the reference through `jobs`)

mcwitt: Fixed in e8a746a

timemachine/fe/rbfe.py
```diff
@@ -679,3 +685,114 @@ def run_complex(
...
+            ff,
+            protein,
+            n_frames,
+            seed + edge_idx,
+            file_client,
+        )
+        for edge_idx, edge in enumerate(edges)
+    )
+
+    paths = [job.result() for job in jobs]
```
mcwitt, line 797: Would be nice to do
```suggestion
    paths = [job.result() for job in futures.as_completed(jobs)]
```
here instead, but this doesn't seem possible without reimplementing some `concurrent.futures` internals for our custom futures (e.g. for AWS Batch)

jkausrelay (approved):

timemachine/fe/rbfe.py
```diff
@@ -792,7 +791,7 @@ def run_edges_parallel(
             edge,
             mols,
             ff,
-            protein,
+            protein_path,
```
jkausrelay, line 794: This should be the protein object, the path won't be accessible on a remote server.

timemachine/fe/rbfe.py
```diff
@@ -707,8 +706,8 @@ def run_edge_and_save_results(
         mol_b = mols[edge.mol_b_name]
         core, smarts = atom_mapping.get_core_with_alignment(mol_a, mol_b, threshold=2.0)
 
-        complex_res, complex_top = run_complex(mol_a, mol_b, core, forcefield, protein, n_frames, seed)
-        solvent_res, solvent_top = run_solvent(mol_a, mol_b, core, forcefield, protein, n_frames, seed)
+        complex_res, complex_top = run_complex(mol_a, mol_b, core, forcefield, protein_path, n_frames, seed)
```
jkausrelay, line 709: run_complex/run_solvent will take either the path or the object. But the object is better for this use case.

timemachine/fe/rbfe.py
mcwitt: Ah, thanks. I was seeing an error from the remote jobs that indicated they were trying to read in the PDB file, so I thought the driver script would need to store it using the remote file client and pass the path into the job. But looks like the error might've been due to another bug. Will revert this.

maxentile (approved):
Looks good!

(One unit conversion to double-check appears in this diff but was present before)

examples/rbfe_edge_list.py
```diff
@@ -97,54 +25,43 @@ def read_from_args():
...
+        return [
+            rbfe.Edge(
+                mol_a_name,
+                mol_b_name,
+                {
+                    "exp_ddg_kcal": float(exp_ddg) * KCAL_TO_KJ,
+                    "fep_ddg_kcal": float(fep_ddg) * KCAL_TO_KJ,
+                    "fep_ddg_err_kcal": float(fep_ddg_err) * KCAL_TO_KJ,
+                    "ccc_ddg_kcal": float(ccc_ddg) * KCAL_TO_KJ,
+                    "ccc_ddg_err_kcal": float(ccc_ddg_err) * KCAL_TO_KJ,
```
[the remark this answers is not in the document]
mcwitt: Hmm, good catch, this looks wrong. 1 = 1 kcal / 4.184 kJ and `KCAL_TO_KJ == 4.184`, so I think you're right that we should be dividing here.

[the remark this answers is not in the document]
mcwitt: I think I just misunderstood the intended units of the output. I think it's best to remove the suffix from the outputs instead (since I think we generally follow the convention that no suffix means kJ)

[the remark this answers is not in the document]
mcwitt: Addressed in f3c923f

PR #914, opened 2022-11-17
Fix broken parallelism in RBFE driver

mcwitt:
* Fixes bug in `rbfe.run_edges_parallel` forcing edges to run sequentially (9210947)
* Fixes bugs preventing metadata from being logged and failure results from being stored (0b34d2d)

#913 introduced a bug that forced `rbfe.run_edges_parallel` to run edges in sequence. The change was to call `client.submit()` inside of a generator expression, which was intended to discard references to futures after completion to address OOM failure. But this also meant that jobs were submitted lazily by `[job.result() for job in jobs]`, each only after the preceding job had finished.

This PR switches to a different method of removing references after completion that does not have this issue.

### DIFF
--- a/timemachine/fe/rbfe.py
+++ b/timemachine/fe/rbfe.py
@@ -800,6 +800,10 @@ def run_edges_parallel(
         for edge_idx, edge in enumerate(edges)
     ]
 
+    # Remove references to completed jobs to allow garbage collection.
+    # TODO: The current approach uses O(edges) memory in the worst case (e.g. if the first job gets stuck). Ideally we
+    # should process and remove references to jobs in the order they complete, but this would require an interface
+    # presently not implemented in our custom future classes.
     paths = []
     while jobs:
         job = jobs.pop(0)
--- a/timemachine/fe/rbfe.py
+++ b/timemachine/fe/rbfe.py
@@ -716,8 +716,10 @@ def run_edge_and_save_results(
             " | ".join(
                 [
                     f"{edge.mol_a_name} -> {edge.mol_b_name} (kJ/mol)",
-                    f"exp_ddg {edge.metadata['exp_ddg_kcal']:.2f}" if "exp_ddg_kcal" in edge.metadata else "",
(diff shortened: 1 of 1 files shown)


timemachine/fe/rbfe.py
```diff
@@ -784,9 +784,7 @@ def run_edges_parallel(
     # Without this get_mol_name(mol) will fail on roundtripped mol
     Chem.SetDefaultPickleProperties(Chem.PropertyPickleOptions.AllProps)
 
-    # NOTE: using a generator expression here ensures that no references are kept to future objects after we've
-    # retrieved their results
```
mcwitt, line 788: :facepalm:

timemachine/fe/rbfe.py
```diff
@@ -798,7 +796,11 @@ def run_edges_parallel(
             file_client,
         )
         for edge_idx, edge in enumerate(edges)
-    )
+    ]
+
+    paths = []
+    while jobs:
+        job = jobs.pop(0)
+        paths.append(job.result())
```
mcwitt, line 804: Tested this more carefully to verify that jobs actually run in parallel now.

timemachine/fe/rbfe.py
```diff
@@ -798,7 +796,11 @@ def run_edges_parallel(
...
 
-    paths = [job.result() for job in jobs]
```
maxentile, line 803: Curiosity / nit: is this second change necessary?

(Would have thought the above change from lazy `( ... )` -> eager `[ ... ]` is sufficient to ensure all edge jobs are enqueued before we start waiting on results, don't immediately see the necessity for this additional change.)

maxentile (approved):

timemachine/fe/rbfe.py
maxentile: Oh, right 🤦 , not required for the parallelism fix but required to remove references to jobs after completion / keep memory usage from growing with `len(job)` -- sorry, had fixated on just one part of the PR description. This looks like a clean way to accomplish both goals. 👍

mcwitt: Added a comment in 14e2764 to clarify the motivation for this.

badisa (approved):

PR #915, opened 2022-11-18
Local MD without IxnGroup

badisa:
- Runs Local MD by applying restraints using `FlatBottomBond` and modifying the particles of the context's `NonbondedAllPairs`. Which does mean it doesn't support vacuum simulations
- Adds a `Thermostat` C++ type for integrators that have a temperature
- Adds a `PinnedHostBuffer` for handling allocation and clean up of pinned host memory


timemachine/cpp/src/bound_potential.cu
```diff
@@ -40,6 +40,22 @@ void BoundPotential::execute_host(
...
+    }
+    if (updated_size > 0) {
+        if (updated_size > this->size()) {
+            d_p.reset(new DeviceBuffer<double>(updated_size));
+        }
+        gpuErrchk(cudaMemcpyAsync(
+            d_p->data, d_new_params, updated_size * sizeof(*d_p->data), cudaMemcpyDeviceToDevice, stream));
+    }
+    shape = device_shape;
+}
```
mcwitt, line 57: Idea for potential simplification: does BoundPotential need to know the shape of the parameters, or could it get away with just the size? It seems like switching to size would simplify a few places (and allow removing `BoundPotential::size()`)

timemachine/cpp/src/context.hpp
```diff
@@ -28,6 +28,15 @@ class Context {
 
     std::array<std::vector<double>, 3> multiple_steps_U(const int n_steps, int store_u_interval, int store_x_interval);
 
+    std::array<std::vector<double>, 2> multiple_steps_local(
+        const int local_steps,
```
mcwitt, line 32: nit: `n_steps`, to align with the signature of `multiple_steps`?

timemachine/cpp/src/kernels/k_flat_bottom_bond.cuh
```diff
@@ -1,6 +1,65 @@
...
+template <typename RealType>
+RealType __device__ __forceinline__ compute_flat_bottom_energy(
+    RealType k, RealType r_lt_rmin, RealType r_gt_rmax, RealType r, RealType rmin, RealType rmax) {
+    return (k / 4) * (r_lt_rmin * (pow(r - rmin, 4)) + r_gt_rmax * (pow(r - rmax, 4)));
+}
+
+template <typename RealType>
+void __global__ k_log_probability_selection(
+    const int N,                             // Num atoms
+    const double kBt,                        // BOLTZ * temperature
```
mcwitt, line 14: tiny nit: capital T for temperature? (`kBT`)

timemachine/cpp/src/kernels/k_flat_bottom_bond.cuh
```diff
@@ -1,6 +1,65 @@
...
+    unsigned int *__restrict__ selected      // [N] idx array, N if idx is not selected, else idx of coordinate
+) {
+    const unsigned int idx = blockDim.x * blockIdx.x + threadIdx.x;
+    if (idx >= N) {
+        return;
+    }
+    const float radius_sq = radius * radius;
+    const RealType bx = box[0 * 3 + 0];
+    const RealType by = box[1 * 3 + 1];
+    const RealType bz = box[2 * 3 + 2];
```
mcwitt, line 30: Should we throw an exception in `FlatBottomBond::execute_device` if passed a non-ortholinear box, as we do for the nonbonded potentials?

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -786,6 +860,7 @@ template <typename RealType> void declare_nonbonded_all_pairs(py::module &m, con
         m, pyclass_name.c_str(), py::buffer_protocol(), py::dynamic_attr())
         .def("set_nblist_padding", &timemachine::NonbondedAllPairs<RealType>::set_nblist_padding, py::arg("val"))
         .def("disable_hilbert_sort", &timemachine::NonbondedAllPairs<RealType>::disable_hilbert_sort)
+        .def("set_atom_idxs", &timemachine::NonbondedAllPairs<RealType>::set_atom_idxs)
```
mcwitt, line 863: nit: missing argument name

timemachine/cpp/src/kernels/k_flat_bottom_bond.cuh
badisa: Seems like this would be well suited to https://github.com/proteneer/timemachine/pull/916

We should probably verify non-ortholinear boxes as we go from python -> C++ and avoid runtime checks.

timemachine/cpp/src/wrap_kernels.cpp
badisa: Fixed in https://github.com/proteneer/timemachine/pull/915/commits/ee6008742be225302c0e4fd51cd0a38d884e810b

timemachine/cpp/src/kernels/k_flat_bottom_bond.cuh
badisa: Fixed in [ff96b2b](https://github.com/proteneer/timemachine/pull/915/commits/ff96b2b513c05b7cd3f1aec16f5550f71617f1bc)

timemachine/cpp/src/context.hpp
badisa: Fixed in https://github.com/proteneer/timemachine/pull/915/commits/8c4e10b69805f1531791636d369a85b2e701dbce

timemachine/cpp/src/bound_potential.cu
badisa: I agree, the shape is confusing and never used as a shape.

Am inclined to doing this in a follow up PR, to avoid expanding the scope of this PR.

timemachine/cpp/src/nonbonded_all_pairs.cu
```diff
@@ -123,21 +131,60 @@ template <typename RealType> NonbondedAllPairs<RealType>::~NonbondedAllPairs() {
...
-    const int tpb = 32;
+    const int tpb = warp_size;
     const int B = ceil_divide(K_, tpb);
 
-    if (d_atom_idxs_) {
-        k_coords_to_kv_gather<<<B, tpb, 0, stream>>>(
-            K_, d_atom_idxs_, d_coords, d_box, d_bin_to_idx_, d_sort_keys_in_, d_sort_vals_in_);
-    } else {
-        // N_ == K_
-        k_coords_to_kv<<<B, tpb, 0, stream>>>(K_, d_coords, d_box, d_bin_to_idx_, d_sort_keys_in_, d_sort_vals_in_);
```
mcwitt, line 139: Can `k_coords_to_kv` be removed? (then might also rename `k_coords_to_kv{_gather,}`)

timemachine/cpp/src/nonbonded_all_pairs.cu
```diff
@@ -123,21 +131,60 @@ template <typename RealType> NonbondedAllPairs<RealType>::~NonbondedAllPairs() {
...
 void NonbondedAllPairs<RealType>::hilbert_sort(const double *d_coords, const double *d_box, cudaStream_t stream) {
 
-    const int tpb = 32;
+    const int tpb = warp_size;
     const int B = ceil_divide(K_, tpb);
 
-    if (d_atom_idxs_) {
-        k_coords_to_kv_gather<<<B, tpb, 0, stream>>>(
-            K_, d_atom_idxs_, d_coords, d_box, d_bin_to_idx_, d_sort_keys_in_, d_sort_vals_in_);
-    } else {
```
mcwitt, line 137: Is there any performance penalty to removing this fast path in the global MD case?

timemachine/cpp/src/kernels/k_flat_bottom_bond.cuh
mcwitt: Makes sense to add to #916, assuming that all potentials require `box` to be diagonal. Just noting that there's currently an inconsistency with the nonbonded potentials, which check this assumption in `execute_device`.

timemachine/cpp/src/nonbonded_all_pairs.cu
badisa: Based on the benchmarks there is not. Only run it a couple of times, can put together something more robust next week.

timemachine/cpp/src/nonbonded_all_pairs.cu
badisa: Removed here: https://github.com/proteneer/timemachine/pull/915/commits/89fb1d8e41e5b6c9df4fb521da1199dce1d023e9

Not clear what you mean by rename. If you mean to rename `k_coords_to_kv_gather-> k_coords_to_kv`, seems like the convention of `scatter` and `gather` in the signature is nice.

mcwitt: > If you mean to rename k_coords_to_kv_gather-> k_coords_to_kv

Yep, that's what I meant. Was thinking to remove the suffix since it's now the only variant. But I can see the utility of keeping `gather` in the name.

timemachine/cpp/src/kernels/k_flat_bottom_bond.cuh
```diff
@@ -1,6 +1,65 @@
 #include "../fixed_point.hpp"
 #include "k_fixed_point.cuh"
 
+// branchless implementation of piecewise function
+template <typename RealType>
+RealType __device__ __forceinline__ compute_flat_bottom_energy(
+    RealType k, RealType r_lt_rmin, RealType r_gt_rmax, RealType r, RealType rmin, RealType rmax) {
```
maxentile, line 7: Don't immediately see need for intermediate variables `r_lt_rmin`, `r_gt_rmax` to be part of the signature, rather than computed from the other arguments -- can these be removed from the signature?

(Expanding in this way looks unsafe. For example, a user of this function can pass in `r, r_min, r_lt_rmin`, such that `r_lt_min != (r < rmin)` but no exception etc. would be surfaced.)

timemachine/cpp/src/kernels/k_flat_bottom_bond.cuh
```diff
@@ -1,6 +1,65 @@
...
+    atom_atom_dx -= bx * nearbyint(atom_atom_dx * inv_bx);
+    atom_atom_dy -= by * nearbyint(atom_atom_dy * inv_by);
+    atom_atom_dz -= bz * nearbyint(atom_atom_dz * inv_bz);
+
+    const RealType distance_sq =
+        atom_atom_dx * atom_atom_dx + atom_atom_dy * atom_atom_dy + atom_atom_dz * atom_atom_dz;
+
+    RealType prob = 1.0;
+    if (distance_sq >= radius_sq) {
+        RealType energy = compute_flat_bottom_energy<RealType>(
```
maxentile, line 49: Curiosity:
(1) should `k_log_probability_selection` be moved somewhere local md specific rather than flat-bottom bond specific,
(2) would it be hard to swap in other restraint functions besides this specific "flat bottom bond" in the future?

(The x^4 was a specific way to preempt a concern about the more widely used x^2 flat-bottom restraint raised about a prototype implementation, but I'm not aware if the concern has also arisen in practice / not sure this specific way of resolving the concern is an intrinsic part of local MD. Initially I think the logic of `k_log_probability_selection` is independent of the implementation details of the restraint function, and should remain mostly unchanged for any choice of non-negative valued restraint function, as in https://github.com/proteneer/timemachine/pull/785 )

maxentile (commented):
Initially looks very good!

(But there's a lot to cover, will review in stages)

timemachine/cpp/src/context.cu
```diff
@@ -38,6 +52,326 @@ Context::~Context() {
...
+
+    // Pinned memory for getting lengths of indice arrays
+    PinnedHostBuffer<int> p_num_selected(1);
+    DeviceBuffer<int> num_selected_buffer(1);
+    LessThan select_op(N_);
+
+    std::size_t temp_storage_bytes = 0;
+    cub::DevicePartition::If(
+        nullptr, temp_storage_bytes, d_shell_idxs_inner.data, d_row_idxs.data, num_selected_buffer.data, N_, select_op);
+    DeviceBuffer<char> d_temp_storage_buffer(temp_storage_bytes);
```
maxentile, line 192: Not quite following this -- is it worth a line comment why this temporary storage buffer is of bytes rather than ints / uints?

(Will it only ever contain idxs? Or will it also be used for something else?)

maxentile (approved):

timemachine/cpp/src/kernels/k_flat_bottom_bond.cuh
badisa: 1) Keeping `k_log_probability_selection` here was intended to be very explicit about its implementation being specific to the flat bottom restraint. Don't have a strong opinion on this, could be moved elsewhere.

2) Would be pretty easy to swap out other restraint functions, would mostly depend on the ability to construct the restraint on the fly in C++. 

 `k_log_probability_selection` could probably be templated to use different energy functions in the future. Are you thinking we should support HarmonicBond and FlatBottomBond as different restraints?

maxentile: (1) Fair enough!

(2) Agreed -- I think that's the main requirement.

>k_log_probability_selection could probably be templated to use different energy functions in the future. Are you thinking we should support HarmonicBond and FlatBottomBond as different restraints?

Fine as-is. I didn't mean to advocate a templated or programmable choice among more than one option -- and I don't think there is more than one suitable option currently available. HarmonicBond is probably not immediately suitable.

Just trying to think ahead to what would be involved in revising any details of the implementation that are somewhat arbitrary / tentative (such as the functional form of the restraint).

timemachine/cpp/src/context.cu
maxentile: Addressed by comment in https://github.com/proteneer/timemachine/pull/915/commits/57ef840f794f31cee087c2686d14974f55c787e1

timemachine/cpp/src/kernels/k_flat_bottom_bond.cuh
maxentile: https://github.com/proteneer/timemachine/pull/915/commits/3065e28567d3b47c0ac85f00be856e3847f4ad2e

timemachine/cpp/src/kernels/k_flat_bottom_bond.cuh
badisa: @mcwitt Was resolved in #916

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -178,6 +179,87 @@ void declare_context(py::module &m) {
...
+        F = iterations / store_x_interval
+
+        Parameters
+        ----------
+        n_steps: int
+            Number of steps to run.
+
+        local_idxs: np.array of int32
+            The idxs that defines the atoms to use as the region(s) to run local MD. A random idx will be
+            selected to be frozen and used as the center of the shell of particles to be simulated.
```
mcwitt, line 239: Do we select a different random index at every step, or only once at the beginning? (seems helpful to specify here)

tests/test_local_move.py
```diff
@@ -181,3 +185,52 @@ def assert_correctness(local_move):
...
+    temperature = 300
+    dt = 1.5e-3
+    friction = 0.0
+    seed = 2022
+    cutoff = 1.2
+
+    # Have to minimize, else there can be clashes and the local moves will cause crashes
+    unbound_potentials, sys_params, masses, coords, box = get_solvent_phase_system(mol, ff, 0.0)
+
+    local_idxs = np.array([len(coords) - 1], dtype=np.int32)
```
mcwitt, line 203: Is it worth parameterizing this test to run with different numbers of local_idxs?

tests/test_local_move.py
```diff
@@ -181,3 +185,52 @@ def assert_correctness(local_move):
...
+        subsystem_idxs = np.unique(np.concatenate([inner_shell_idxs, np.arange(num_host_atoms, len(coords))]))
+        return len(subsystem_idxs)
+
+    ctxt = custom_ops.Context(coords, v0, box, intg_impl, bps)
+
+    def local_move(x, steps=1):
+        xs, boxes = ctxt.multiple_steps_local(steps, local_idxs)
+        return xs, boxes
+
+    expect_no_drift(coords, local_move, observable, n_local_resampling_iterations=50)
```
mcwitt, line 236: nit/idea: if it's possible to disable the restraint, it might be useful to add an expected failure to verify the sensitivity of this test

tests/test_md.py
```diff
@@ -277,6 +282,114 @@ def compute_reference_values():
...
+        ctxt = custom_ops.Context(coords, v0, box, intg_impl, bps)
+        # Run steps of local MD
+        xs, boxes = ctxt.multiple_steps_local(num_steps, local_idxs, store_x_interval=x_interval, radius=radius)
+
+        assert xs.shape[0] == num_steps // x_interval
+        assert boxes.shape[0] == num_steps // x_interval
+
+        # Indices in mol that weren't the last atom should have moved
+        assert np.all(coords[:-1][-(mol.GetNumAtoms() - 1) :] != xs[-1][:-1][-(mol.GetNumAtoms() - 1) :])
+        assert np.any(coords[: -mol.GetNumAtoms()] == xs[-1][: -mol.GetNumAtoms()])
```
mcwitt, line 343: nit: could this use `local_idxs`?

tests/test_md.py
```diff
@@ -277,6 +282,114 @@ def compute_reference_values():
...
+
+        assert xs.shape[0] == num_steps // x_interval
+        assert boxes.shape[0] == num_steps // x_interval
+
+        # Running with Barostat should similarly not impact the positions of the global system, as it doesn't
+        # run during local MD
+        assert np.all(coords[:-1][-(mol.GetNumAtoms() - 1) :] != xs[-1][:-1][-(mol.GetNumAtoms() - 1) :])
+        assert np.any(coords[: -mol.GetNumAtoms()] == xs[-1][: -mol.GetNumAtoms()])
+
+        summed_potential = SummedPotential(unbound_potentials, sys_params)
```
mcwitt, line 373: nit: could this test be split up into several smaller tests? (e.g. separating parameter validation, checks for expected side effects, consistency with and without summed potential, etc.). I think that would make it easier to follow what is being tested.

timemachine/cpp/src/context.hpp
```diff
@@ -28,6 +28,15 @@ class Context {
 
     std::array<std::vector<double>, 3> multiple_steps_U(const int n_steps, int store_u_interval, int store_x_interval);
 
+    std::array<std::vector<double>, 2> multiple_steps_local(
+        const int n_steps,
+        const std::vector<int> &local_idxs,
+        const int store_x_interval,
+        const double radius,
+        const double k,
+        const double temperature,
```
mcwitt, line 37: Curious if it would make sense to parameterize this instead by the ratio `k / temperature`? (assuming that e.g. doubling both `k` and `temperature` here has no effect on the selection probabilities). Would this sidestep the need to check that the temperature here is consistent with the integrator's temperature?

mcwitt: Sorry, disregard the suggestion of parameterizing with k / temperature instead of (k, temperature) - I'd misunderstood that temperature wasn't required except for this ratio.

Related question though: would it make sense to remove temperature as a parameter and just use the integrator temperature?

maxentile: >Related question though: would it make sense to remove temperature as a parameter and just use the integrator temperature?

(👍 , this seems cleaner/safer, if possible)

tests/test_local_move.py
badisa: I am hesitant to expose this, as it would be an invalid mode of sampling. 

I feel that this test that verifies that the naive approach to sampling is invalid is sufficient: https://github.com/proteneer/timemachine/blob/master/tests/test_local_move.py#L182-L183

timemachine/cpp/src/context.hpp
maxentile: >If we remove temperature as a parameter, when you use an integrator (not a thermostat, IE Verlet) would it be alright to use a default temperature that is not tunable?

Ahh, I think that shows a possible inconsistency in allowing to pair "Local MD" with a "raw VelocityVerlet" integrator.

In this context I assumed we were treating "VelocityVerlet" as an approximately correct MCMC move for a specific NVT ensemble (i.e. it _is_ implicitly associated with a temperature in this context).

One correct way to use VelocityVerlet as an MCMC move would be to use [an HMC move](https://github.com/proteneer/timemachine/pull/787). ("Randomly sample velocities (depending on temperature/masses), integrate with velocity verlet (depending on masses), accept or reject (depending on temperature/masses).")

In practice this correct implementation could be close to a less explicit approximate implementation. (E.g. if the temperature is always the default temperature, initial velocities are distributed as expected given temperature, and stepsize etc. are small enough that nearly 100% of moves would be accepted, then "raw `VelocityVerlet(dt, masses)`" could have pretty similar behavior to "`HMC(dt, masses, temperature)`".)

Agnostic whether "it's the user's responsibility" to understand this and make sure masses/temperatures/initial velocities etc. are self-consistent + validate that the stepsize is small enough that approximation errors are negligible + ..., or if "it's the implementer's responsibility" to put guardrails here.

tests/test_local_move.py
badisa: After further investigation, it doesn't seem like it. I switched to what seems like the more typical case, of the entire ligand.

Testing the case with random idxs is a bit more difficult, as a random index will be selected in C++, and doesn't seem to cover a different set of logic.

tests/test_local_move.py
```diff
@@ -181,3 +185,52 @@ def assert_correctness(local_move):
...
+    cutoff = 1.2
+
+    # Have to minimize, else there can be clashes and the local moves will cause crashes
+    unbound_potentials, sys_params, masses, coords, box = get_solvent_phase_system(mol, ff, 0.0)
+
+    local_idxs = np.array([len(coords) - 1], dtype=np.int32)
+
+    nblist = custom_ops.Neighborlist_f32(coords.shape[0])
+    num_local_idxs = len(local_idxs)
+    num_host_atoms = len(coords) - num_local_idxs
```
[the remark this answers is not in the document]
badisa: Fixed in https://github.com/proteneer/timemachine/pull/915/commits/b8bb42f0225fa2c5862478a26baf44073a04ea9d

timemachine/cpp/src/context.hpp
badisa: Talking to @maxentile was decided to require a thermostat, and we will wrap the Verlet integrator into a thermostat at a later date.

https://github.com/proteneer/timemachine/pull/915/commits/0165635690f13fbc12093e1bfa0fbb3dc406afbf

tests/test_local_move.py
```diff
@@ -181,3 +185,52 @@ def assert_correctness(local_move):
     # expect failure with ablated version of local move
     with pytest.raises(RuntimeError):
         assert_correctness(incorrect_local_move)
+
+
+def test_local_md_particle_density():
```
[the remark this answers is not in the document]
badisa: Added docstring: https://github.com/proteneer/timemachine/pull/915/commits/3e045173b1e1c31abd6fe5fdd2a75f4aee2629b5

tests/test_md.py
badisa: Split up into two tests: https://github.com/proteneer/timemachine/pull/915/commits/2bbc6a37ada043c27f9a49c447fbd31226b2fb6d

Kept the rest of the checks together as they should all be identical

timemachine/cpp/src/wrap_kernels.cpp
badisa: Specified here: https://github.com/proteneer/timemachine/pull/915/commits/c13e0e4522e9d7f52f681291347289a6a66c81bf

mcwitt (approved):

timemachine/cpp/src/device_buffer.cu
```diff
@@ -32,6 +32,9 @@ template <typename T> void DeviceBuffer<T>::copy_to(T *host_buffer) const {
 }
```
proteneer, line 33: at some point we should be more precise about these, i.e. use the `uint64_t`, `int64_t`, etc. types from `<cstdint>`

timemachine/cpp/src/nonbonded_all_pairs.cu
```diff
@@ -31,19 +35,15 @@ NonbondedAllPairs<RealType>::NonbondedAllPairs(
...
 
-    if (atom_idxs) {
-        cudaSafeMalloc(&d_atom_idxs_, K_ * sizeof(*d_atom_idxs_));
-        std::vector<int> atom_idxs_v(atom_idxs->begin(), atom_idxs->end());
-        gpuErrchk(cudaMemcpy(d_atom_idxs_, &atom_idxs_v[0], K_ * sizeof(*d_atom_idxs_), cudaMemcpyHostToDevice));
-    }
+    cudaSafeMalloc(&d_atom_idxs_, N_ * sizeof(*d_atom_idxs_));
 
-    cudaSafeMalloc(&d_sorted_atom_idxs_, K_ * sizeof(*d_sorted_atom_idxs_));
+    cudaSafeMalloc(&d_sorted_atom_idxs_, N_ * sizeof(*d_sorted_atom_idxs_));
```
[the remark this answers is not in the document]
proteneer: (guessing it had to do with letting K_ be variable across different calls, but wasn't 100% sure)

timemachine/cpp/src/kernels/k_integrator.cuh
```diff
@@ -47,6 +55,12 @@ __global__ void half_step_velocity_verlet(
     if (atom_idx >= N) {
         return;
     }
+    if (idxs) {
```
proteneer, line 58: (same as comment above re: clarity)

timemachine/cpp/src/context.cu
```diff
@@ -38,6 +52,330 @@ Context::~Context() {
...
+            nonbonded_potential = pot;
+        }
+    }
+    if (!nonbonded_potential) {
+        throw std::runtime_error("unable to find a NonbondedAllPairs potential");
+    }
+
+    std::mt19937 rng;
+    rng.seed(seed);
+    std::uniform_int_distribution<unsigned int> random_dist(0, local_idxs.size() - 1);
```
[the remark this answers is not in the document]
badisa: Good catch, is broken.

timemachine/cpp/src/nonbonded_all_pairs.cu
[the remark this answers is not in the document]
badisa: Correct, it is so that the nonbonded potential can use different values of K. Could also reconstruct a new NB potential in the local md, but this seemed cleaner

timemachine/cpp/src/context.cu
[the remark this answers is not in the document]
badisa: Fixed in https://github.com/proteneer/timemachine/pull/915/commits/916955f15457b5024720533812f53cc8faeb0be8

timemachine/cpp/src/nonbonded_all_pairs.cu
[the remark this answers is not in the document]
proteneer: Okay - wasn't sure if it was safe to malloc now of size `N_` instead of size `K_` - i.e. if we only initially the `K_` subsets of the resulting arrays, is it safe that the rest of the array is uninitialized?

timemachine/cpp/src/context.cu
```diff
@@ -38,6 +52,330 @@ Context::~Context() {
...
+        if (summed_potential != nullptr) {
+            flatten_potentials(summed_potential->get_potentials(), flattened);
+            continue;
+        }
+        flattened.push_back(pot);
+    }
+}
+
+double Context::_get_temperature() {
+    if (Thermostat *thermostat = dynamic_cast<Thermostat *>(intg_); thermostat != nullptr) {
```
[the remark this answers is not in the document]
badisa: Its `Integrator -> Thermostat`, the context can take any integrator. For the Verlet integrator (inherits from Integrator, not Thermostat), it has no temperature and so local md is potentially inconsistent. We are going to add a wrapper around integrators to have a temperature, but to ensure consistency right now this requires a thermostat.

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -200,6 +201,106 @@ void declare_context(py::module &m) {
...
+            Number of steps to run.
+
+        local_idxs: np.array of int32
+            The idxs that defines the atoms to use as the region(s) to run local MD. A random idx will be
+            selected to be frozen and used as the center of the shell of particles to be simulated. The selected
+            idx is constant across all steps.
+
+        burn_in: int
+            How many steps to run prior to storing frames. This is to handle the fact that the local simulation applies a
+            restraint, and burn in helps equilibrate the local simulation. Running with small numbers of steps (< 10) is not recommended.
```
proteneer, line 283: this is `<10` but above its `<100` - maybe we should be consistent?

timemachine/cpp/src/nonbonded_all_pairs.cu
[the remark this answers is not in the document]
badisa: It is safe as long as we maintain our convention of passing `K_` to the kernels and avoid going beyond that. An example is: https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/kernels/nonbonded_common.cu#L19-L21

timemachine/cpp/src/kernels/k_integrator.cuh
```diff
@@ -17,6 +18,12 @@ __global__ void update_forward_baoab(
...
+        atom_idx = idxs[atom_idx];
```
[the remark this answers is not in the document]
badisa: Fixed in https://github.com/proteneer/timemachine/pull/915/commits/e950fc873e876031ee5090bcb321636472642368

timemachine/cpp/src/wrap_kernels.cpp
badisa: Fixed in https://github.com/proteneer/timemachine/pull/915/commits/829a8df6e428ded5ebceafd4dcb8da1f7502b49b

timemachine/cpp/src/context.cu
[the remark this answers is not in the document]
badisa: For some context: https://github.com/proteneer/timemachine/pull/915#discussion_r1039794884

proteneer (approved):
lgtm

timemachine/cpp/src/context.hpp
maxentile: (for posterity: a consistent alternative to giving VelocityVerlet a temperature is to expose a slightly different local MD interface, as in https://github.com/proteneer/timemachine/pull/1005 )

PR #918, opened 2022-11-23
TM-209: Use larger cut-off for vacuum leg

jkausrelay:
- Ligands in vacuum can move more during minimization step and still be stable
- Increase the cut-off for the vacuum leg to allow these transformations to run


proteneer:
lgtm

proteneer (approved):

PR #917, opened 2022-11-23
Complete rewrite of the MCS algorithm

proteneer:
This PR completely rewrites our MCS algorithm from scratch. There's no meaningful changes from the `bgl` repo, except for the addition of more tests.

PR #919, opened 2022-11-28
Re-enable chiral MCS filter (& disable pyramidal nitrogen SMARTS)

maxentile:
Address TM-187 by re-enabling the chiral filter added in https://github.com/proteneer/timemachine/pull/908 and disabled in https://github.com/proteneer/timemachine/pull/917

TODO:
- [x]  ~Expand tests using an example @proteneer provided, where the approach of filtering `all_cores` returned by `mcgregor.mcs` may not suffice.~ (An example of this behavior is in current tests.) If needed, move this filter into `mcgregor.py`. (Done in https://github.com/proteneer/timemachine/pull/919/commits/15a6327d805610001d31237b87bc824ce0de798e.)
- [x] Update this test: https://github.com/proteneer/timemachine/pull/917#discussion_r1031849102 (Done in https://github.com/proteneer/timemachine/pull/919/commits/5807d8205751c5d53f403e1f02ffa360f91cf236.)

### DIFF
--- a/timemachine/fe/atom_mapping.py
+++ b/timemachine/fe/atom_mapping.py
@@ -5,6 +5,7 @@ import numpy as np
 from rdkit import Chem
 
 from timemachine.fe import mcgregor
+from timemachine.fe.chiral_utils import ChiralRestrIdxSet, has_chiral_atom_flips
 from timemachine.fe.utils import get_romol_bonds, get_romol_conf
 
 # (ytz): Just like how one should never re-write an MD engine, one should never rewrite an MCS library.
@@ -44,7 +45,16 @@ from timemachine.fe.utils import get_romol_bonds, get_romol_conf
 
 
 def get_cores(
-    mol_a, mol_b, ring_cutoff, chain_cutoff, max_visits, connected_core, max_cores, enforce_core_core, complete_rings
+    mol_a,
+    mol_b,
+    ring_cutoff,
+    chain_cutoff,
+    max_visits,
+    connected_core,
+    max_cores,
+    enforce_core_core,
+    complete_rings,
+    enforce_chiral=True,
 ):
     """
     Finds set of cores between two molecules that maximizes the number of common edges.
@@ -92,6 +102,9 @@ def get_cores(
         If we require mapped atoms that are in a ring to be complete.
         If True then connected_core must also be True.
 
+    enforce_chiral: bool
+        Filter out cores that would flip atom chirality
+
     Returns
     -------
     Returns a list of all_cores
@@ -100,36 +113,29 @@ def get_cores(
 
     assert max_cores > 0
 
+    core_kwargs = dict(
+        mol_a=mol_a,
+        mol_b=mol_b,
+        ring_cutoff=ring_cutoff,
+        chain_cutoff=chain_cutoff,
+        max_visits=max_visits,
+        connected_core=connected_core,
+        max_cores=max_cores,
+        enforce_core_core=enforce_core_core,
+        complete_rings=complete_rings,
+        enforce_chiral=enforce_chiral,
+    )
+
     # we require that mol_a.GetNumAtoms() <= mol_b.GetNumAtoms()
     if mol_a.GetNumAtoms() > mol_b.GetNumAtoms():
-        all_cores = _get_cores_impl(
-            mol_b,
-            mol_a,
-            ring_cutoff,
-            chain_cutoff,
-            max_visits,
-            connected_core,
-            max_cores,
-            enforce_core_core,
-            complete_rings,
-        )
+        all_cores = _get_cores_impl(**core_kwargs)
         new_cores = []
         for core in all_cores:
             core = np.array([(x[1], x[0]) for x in core], dtype=core.dtype)
             new_cores.append(core)
         return new_cores
     else:
-        all_cores = _get_cores_impl(
-            mol_a,
-            mol_b,
-            ring_cutoff,
-            chain_cutoff,
-            max_visits,
-            connected_core,
-            max_cores,
-            enforce_core_core,
-            complete_rings,
-        )
+        all_cores = _get_cores_impl(**core_kwargs)
         return all_cores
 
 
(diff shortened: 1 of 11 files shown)


maxentile:
> Expand tests using an example @proteneer provided, where the approach of filtering all_cores returned by mcgregor.mcs may not suffice. If needed, move this filter into mcgregor.py.

Looks like hif2a edge 235 -> 237 in `test_atom_mapping.py::test_all_pairs` already triggers the behavior @proteneer anticipated. (In this case, `all_cores` contains no cores that pass chiral atom filter.)

proteneer:
> Looks like hif2a edge 235 -> 237 in test_atom_mapping.py::test_all_pairs already triggers the behavior @proteneer anticipated. (In this case, all_cores contains no cores that pass chiral atom filter.)

Interesting! Would be good to look at the atom-mapping generated without the chiral filters.

maxentile:
>Interesting! Would be good to look at the atom-mapping generated without the chiral filters.

From looking at this with @proteneer , looks like a case where 2 indistinguishable terminal atoms can only match 1 way due to conformer-aware distance cutoff, and the 1 way they can match does not pass the chiral atom filter.

Possible workarounds:
* Forego this filter or make it optional
* Modify this filter to ignore permutations of indistinguishable atoms
* Tweak distance cutoff -- e.g. adding a third tunable distance cutoff for terminal atoms, and setting it larger than the current non-ring atoms distance cutoff
* Require inputs where the desired atom mapping is more precisely encoded in conformer alignment (so that matchable terminal atoms are always within current distance cutoff)
* Modify MCS following @bp-kelley 's earlier suggestion (remove terminal atoms, run MCS, add terminal atoms) https://github.com/proteneer/timemachine/pull/819#discussion_r966130215
* Modify filter so that it runs on partial solutions in `mcgregor.py` (would cause this case to pass but would de-map more atoms that are part of the conformer-oblivious MCS)
* ...

maxentile:
> If needed, move this filter into mcgregor.py.
> Modify filter so that it runs on partial solutions in mcgregor.py (would cause this case to pass but would de-map more atoms that are part of the conformer-oblivious MCS)

Applied filter to internal nodes of search tree in https://github.com/proteneer/timemachine/pull/919/commits/15a6327d805610001d31237b87bc824ce0de798e

Tests now pass, and this doesn't seem to have much effect on the time required to run the atom mapping tests.

Previously (as of https://github.com/proteneer/timemachine/pull/917/commits/ca7c57b4329f98dbe9ba60b8c36a130e8cba12a2 )
```
2059.52s call     tests/test_atom_mapping.py::test_all_pairs[timemachine/datasets/fep_benchmark/eg5/ligands.sdf]
1968.09s call     tests/test_atom_mapping.py::test_all_pairs[timemachine/datasets/fep_benchmark/hif2a/ligands.sdf]
...
```
Proposed (as of https://github.com/proteneer/timemachine/pull/919/commits/15a6327d805610001d31237b87bc824ce0de798e )
```
2022.33s call     tests/test_atom_mapping.py::test_all_pairs[timemachine/datasets/fep_benchmark/eg5/ligands.sdf]
1956.57s call     tests/test_atom_mapping.py::test_all_pairs[timemachine/datasets/fep_benchmark/hif2a/ligands.sdf]
...
```

timemachine/fe/chiral_utils.py
```diff
@@ -231,12 +231,38 @@ def apply_mapping(c, i, j, k):
     return conflicts
 
 
+def _has_chiral_atom_map_flips_one_direction(
+    core: np.ndarray,
+    chiral_set_a: ChiralRestrIdxSet,
+    chiral_set_b: ChiralRestrIdxSet,
+) -> bool:
+    # _find_atom_map_chiral_conflicts_one_direction, except (1) return bool not set, (2) hard-code mode = FLIP
```
jkausrelay, line 239: nit: Add a comment why you don't just return bool(_find_atom_map_chiral_conflicts_one_direction(...))` 

I assume this is for performance reasons to fail quickly

maxentile: Separate implementation for the case where we only need a bool, in case we want to apply performance optimizations specific to this case, e.g. as suggested in https://github.com/proteneer/timemachine/pull/908#pullrequestreview-1176448612 and https://github.com/proteneer/timemachine/pull/908#issuecomment-1314425524

maxentile:
On an n=47 challenge set provided by @proteneer , the impact of this PR appears to be:
* decrease the overall run-time for the set by a factor of ~2-4x (although it makes some individual transformations take longer)
* often decrease the number of search tree nodes visited -- sometimes by a factor of 100x (although sometimes it increases the number of nodes visited)
* sometimes change the size of the returned cores (most often: no change, worst change: -2 atoms, possibly surprising change: at least one case where this filter _increased_ the size of the returned core by +1 atom)

Used the following settings, toggling `enforce_chiral` = `True` or `False`

```python
all_cores = atom_mapping.get_cores(
                mol_a,
                mol_b,
                ring_cutoff=0.12,
                chain_cutoff=0.2,
                max_visits=1e7,
                connected_core=False,
                max_cores=10_000,
                enforce_core_core=True,
                complete_rings=False,
                enforce_chiral=enforce_chiral,
            )
```

timemachine/fe/atom_mapping.py
```diff
@@ -44,7 +45,16 @@
...
+    mol_a,
+    mol_b,
+    ring_cutoff,
+    chain_cutoff,
+    max_visits,
+    connected_core,
+    max_cores,
+    enforce_core_core,
+    complete_rings,
+    enforce_chiral=True,
```
proteneer, line 57: please disable the default=True, I want to be as explicit as possible here.

proteneer:
Can you add one more test for NH3 -> NH3 (with ring and chain cutoff set to infinity), and check that with `enforce_chiral` we get 3 items in all_cores (with all_cores[0] being the identiy) and without it we get 6 items in all_cores?

timemachine/fe/atom_mapping.py
maxentile: Disabled in https://github.com/proteneer/timemachine/pull/919/commits/2465c43df8a0d87809f47bb42cee740bbdbf97a0

jkausrelay (approved):

maxentile:
>Can you add one more test for NH3 -> NH3 (with ring and chain cutoff set to infinity), and check that with enforce_chiral we get 3 items in all_cores (with all_cores[0] being the identiy) and without it we get 6 items in all_cores?

Note that in master, NH3 currently isn't detected as a chiral atom, see these patterns https://github.com/proteneer/timemachine/blob/5ff0026fc23f848931d4815b36e55aee3113822e/timemachine/fe/chiral_utils.py#L134-L138

```python
from rdkit import Chem
from timemachine.fe.chiral_utils import find_chiral_atoms

nh3 = Chem.AddHs(Chem.MolFromSmiles('N'))
AllChem.EmbedMolecule(nh3)
print(find_chiral_atoms(nh3))  # set()
```

Instead of NH3, added an analogous test with methane in https://github.com/proteneer/timemachine/pull/919/commits/9a23ee50139ebc04f40ae5f696506de32a0f7691

proteneer (approved):
LGTM and nice work!

PR #920, opened 2022-11-29
Fix remove disconnected filter by updating bond core

jkausrelay:
- Need to update marcs matrix between steps of filtering

### DIFF
--- a/timemachine/fe/atom_mapping.py
+++ b/timemachine/fe/atom_mapping.py
@@ -5,6 +5,7 @@ import numpy as np
 from rdkit import Chem
 
 from timemachine.fe import mcgregor
+from timemachine.fe.chiral_utils import ChiralRestrIdxSet, has_chiral_atom_flips
 from timemachine.fe.utils import get_romol_bonds, get_romol_conf
 
 # (ytz): Just like how one should never re-write an MD engine, one should never rewrite an MCS library.
@@ -44,7 +45,16 @@ from timemachine.fe.utils import get_romol_bonds, get_romol_conf
 
 
 def get_cores(
-    mol_a, mol_b, ring_cutoff, chain_cutoff, max_visits, connected_core, max_cores, enforce_core_core, complete_rings
+    mol_a,
+    mol_b,
+    ring_cutoff,
+    chain_cutoff,
+    max_visits,
+    connected_core,
+    max_cores,
+    enforce_core_core,
+    complete_rings,
+    enforce_chiral=True,
 ):
     """
     Finds set of cores between two molecules that maximizes the number of common edges.
@@ -92,6 +102,9 @@ def get_cores(
         If we require mapped atoms that are in a ring to be complete.
         If True then connected_core must also be True.
 
+    enforce_chiral: bool
+        Filter out cores that would flip atom chirality
+
     Returns
     -------
     Returns a list of all_cores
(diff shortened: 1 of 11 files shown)


timemachine/fe/atom_mapping.py
```diff
@@ -271,6 +271,19 @@ def _compute_bond_cores(mol_a, mol_b, marcs):
     return bond_core
 
 
+def _compute_marcs(mol_a, mol_b, bond_core):
+    a_edges = get_edges(mol_a)
+    b_edges = get_edges(mol_b)
+    marcs = np.zeros((len(a_edges), len(b_edges)), dtype=np.byte)
+    for a, b in bond_core.items():
+        e_a = a_edges.index(a)
```
proteneer, line 279: is the ordering here guaranteed?

(i.e. bond [src,dst] and [dst,src] are properly canonicalized to be present in a_edges)

timemachine/fe/atom_mapping.py
```diff
@@ -330,12 +343,12 @@ def _get_cores_impl(
     )
 
     if connected_core:
```
proteneer, line 345: Two notes:


1) I think we should just expand the behavior here explicitly after today's discussion (since we may need a case where complete_rings=True but connected_core=False)

So I think roughly

```

if connected_core and complete_rings:
        # order matters here
        all_cores, all_marcs = remove_incomplete_rings(mol_a, mol_b, all_marcs, all_cores)
        all_cores, all_marcs = remove_disconnected_components(mol_a, mol_b, all_cores, all_marcs)
elif connected_core and not complete_rings:
        all_cores, all_marcs = remove_disconnected_components(mol_a, mol_b, all_cores, all_marcs)
elif completed_rings and not connected_core:
        all_cores, all_marcs = remove_incomplete_rings(mol_a, mol_b, all_marcs, all_cores)
else:
        pass
```

2) Let's make the arguments of `remove_disconnected_components` and `remove_incomplete_rings` consistent, i.e. either have them both be `mol_a, mol_b, all_cores, all_marcs` or  both `mol_a, mol_b, all_marcs, all_cores`

timemachine/fe/atom_mapping.py
jkausrelay: Will take another approach that removed this code.

timemachine/fe/atom_mapping.py
jkausrelay: updated

timemachine/fe/atom_mapping.py
```diff
@@ -202,6 +202,24 @@ def induce_subgraphs_given_marcs(mol_a, mol_b, core, marcs):
...
+def induce_subgraphs_given_bond_core(mol_a, mol_b, core, bond_core):
+    sg_a = nx.Graph()
+    sg_b = nx.Graph()
+    for a in core[:, 0]:
+        sg_a.add_node(a)
+    for b in core[:, 1]:
+        sg_b.add_node(b)
+
+    # edges_a = get_edges(mol_a)
+    # edges_b = get_edges(mol_b)
```
maxentile, line 214: Remove?

timemachine/fe/atom_mapping.py
```diff
@@ -364,13 +373,16 @@ def chiral_filter(trial_core):
...
+    all_bond_cores = [_compute_bond_cores(mol_a, mol_b, marcs) for marcs in all_marcs]
 
-    if complete_rings:
-        assert connected_core
-        all_cores, all_marcs = remove_incomplete_rings(mol_a, mol_b, all_marcs, all_cores)
-        all_cores = remove_disconnected_components(mol_a, mol_b, all_cores, all_marcs)
+    if connected_core and complete_rings:
+        all_cores, all_bond_cores = remove_disconnected_components(mol_a, mol_b, all_cores, all_bond_cores)
+        all_cores, all_bond_cores = remove_incomplete_rings(mol_a, mol_b, all_cores, all_bond_cores)
+        all_cores, all_bond_cores = remove_disconnected_components(mol_a, mol_b, all_cores, all_bond_cores)
```
maxentile, line 381: Non-obvious logic: Comment constraints on filter application order

timemachine/fe/atom_mapping.py
```diff
@@ -202,6 +202,24 @@ def induce_subgraphs_given_marcs(mol_a, mol_b, core, marcs):
...
+        sg_b.add_node(b)
+
+    # edges_a = get_edges(mol_a)
+    # edges_b = get_edges(mol_b)
+    for e1, e2 in bond_core.items():
+        src_a, dst_a = e1
+        src_b, dst_b = e2
+        sg_a.add_edge(src_a, dst_a)
+        sg_b.add_edge(src_b, dst_b)
+    return sg_a, sg_b
```
maxentile, line 220: Style nit: Could be slightly cleaner(?) to write a function `induce_edge_subgraph(mol, edges)` and call it twice, rather than write 2x duplicated instructions for inducing 2 edge subgraphs in the same function

maxentile (approved):

timemachine/fe/atom_mapping.py
maxentile: (Resolved in https://github.com/proteneer/timemachine/pull/920/commits/91d2acbfb08ae22bf1c964a941389c56f2582122 )

PR #922, opened 2022-12-05
Add dG errors by component, plot to RBFE results

mcwitt:
* Adds new fields to `rbfe.SimulationResult`:
  - `dG_errs_by_lambda_by_component`: array of bootstrapped BAR errors for each energy component and pair of adjacent lambda windows
  - `dG_errs_png`: plot analogous to `overlap_summary_png` for bootstrapped BAR errors. Example:

### DIFF
--- a/timemachine/fe/rbfe.py
+++ b/timemachine/fe/rbfe.py
@@ -249,7 +249,7 @@ def plot_BAR(df, df_err, fwd_delta_u, rev_delta_u, title, axes):
     plot_work(fwd_delta_u, rev_delta_u, axes)
 
 
-def dG_err_from_ukln(u_kln):
+def df_err_from_ukln(u_kln):
     k, l, _ = u_kln.shape
     assert k == l == 2
     w_fwd = u_kln[1, 0, :] - u_kln[0, 0, :]
@@ -264,7 +264,7 @@ def plot_dG_errs(ax, components, lambdas, dG_errs):
         ax.plot(lambdas[:-1], ys, marker=".", label=component)
 
     ax.set_xlabel(r"$\lambda_i$")
-    ax.set_ylabel(r"$\Delta G$ error ($\lambda_i$, $\lambda_{i+1}$)")
+    ax.set_ylabel(r"$\Delta G$ error ($\lambda_i$, $\lambda_{i+1}$) / (kJ / mol)")
     ax.legend()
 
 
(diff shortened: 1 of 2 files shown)


timemachine/fe/rbfe.py
```diff
@@ -424,27 +463,22 @@ def estimate_free_energy_given_initial_states(initial_states, protocol, temperat
     ukln_by_lambda_by_component = np.array(ukln_by_component_by_lambda).swapaxes(0, 1)
 
     lambdas = [s.lamb for s in initial_states]
-
-    _, (ax_top, ax_btm) = plt.subplots(2, 1, figsize=(7, 9))
     overlaps_by_lambda = np.array([pair_overlap_from_ukln(u_kln) for u_kln in ukln_by_lambda_by_component.sum(axis=0)])
-    plot_overlap_summary(ax_top, ["Overall"], lambdas, [overlaps_by_lambda])
-
+    dG_errs_by_lambda_by_component = np.array(
+        [[dG_err_from_ukln(u_kln) for u_kln in ukln_by_lambda] for ukln_by_lambda in ukln_by_lambda_by_component]
```
proteneer, line 468: is this in reduced units still? (not sure where the conversion to kJ/mol is happening)

mcwitt: Fixed in 6710e48

proteneer (approved):
lgtm

maxentile (approved):
Looks good!

I continue to have some generic confusion how to interpret these per-energy-component diagnostics (whether per-component overlap as currently reported, or per-component dG_err as proposed). dG_err may be preferable to overlap in some respects since it is a quantity with interpretable units. (Curiosity: Are there examples where the interpretations differ for diagnostics based on dG_err vs. overlap?)

In practice these per-component decomposition are helpfully suggestive of which component to focus attention on, even though in principle (1) neither the estimate nor the error admit a decomposition of this form, (2) this logic could yield arbitrarily bad "false-positive" diagnoses in principle.

A possible short-term enhancement may be to modify these plots with specialized depiction for windows suffering from MD instability. (to preempt questions of form: "was the large error / low overlap due to instability or just inefficiency")

Another possible short-term enhancement may be to add diagnostics for when specific assumptions are violated. (E.g. if a harmonic angle term is being interpolated in a way that assumes the angle marginal is approximately Gaussian, possibly add a diagnostic to flag windows where that feature is highly non-Gaussian.)

A possible long-term enhancement may be to develop a more principled attribution diagnostic (details hazy, but perhaps based on `<du_dparams>_params` and `dparams_dlam`...)

jkausrelay (approved):
lgtm

PR #924, opened 2022-12-08
Modifies atom mapping test to use assert_cores_are_equal

badisa:
* Going from python 3.7 -> 3.8+ changed order and broke the test


timemachine/potentials/jax_utils.py
```diff
@@ -99,7 +99,7 @@ def get_interacting_pair_indices_batch(confs, boxes, pairs, cutoff=1.2):
     assert max_n_neighbors > 0
 
     # sorting in order of [falses, ..., trues]
-    keep_inds = np.argsort(neighbor_masks, axis=1)[:, -max_n_neighbors:]
+    keep_inds = np.argsort(neighbor_masks, kind="stable", axis=1)[:, -max_n_neighbors:]
```
maxentile, line 102: (The uses of this one should be safe w.r.t. sort instability (since it's just used to partition "within cutoff" vs "outside cutoff"), but a good change regardless)

badisa: I agree, just thought we should try and be consistent across the board.

proteneer: +1 for consistency

proteneer (approved):

tests/common.py
```diff
@@ -194,7 +194,7 @@ def hilbert_sort(conf, D):
     for xyz in int_confs.tolist():
         dist = hc.distance_from_coordinates(xyz)
         dists.append(dist)
-    perm = np.argsort(dists)
+    perm = np.argsort(dists, kind="stable")
```
mcwitt, line 197: I guess the unstable sort didn't turn out to be the culprit after all (presumably the instability is deterministic?)

badisa: It did not, but it seems like a nice change to have regardless.

mcwitt (approved):

maxentile (approved):

proteneer (approved):

PR #925, opened 2022-12-09
Reduce default local MD restraint strength

maxentile:
Restraint too strong -> significant discretization error


maxentile:
(Closing as this doesn't fully resolve issue / merits more careful investigation.)

PR #926, opened 2022-12-16
Atom Mapping Modifications

badisa:
When playing around with our atom mapping on large datasets I noticed that sometimes it took a lot longer then expected to map compounds.

Changes:
* Return core if you found one, regardless of timing out
* Adds a test for the max visits
* Remove an unused parameter in the recursion
* Add parameter for specifying the smallest mapping that is acceptable.

### DIFF
--- a/tests/test_atom_mapping.py
+++ b/tests/test_atom_mapping.py
@@ -4,7 +4,7 @@ from rdkit import Chem
 from rdkit.Chem import AllChem
 
 from timemachine.fe import atom_mapping
-from timemachine.fe.mcgregor import MaxVisitsError
+from timemachine.fe.mcgregor import MaxVisitsError, NoMappingError
 from timemachine.fe.utils import plot_atom_mapping_grid
 
 hif2a_set = "timemachine/datasets/fep_benchmark/hif2a/ligands.sdf"
@@ -47,6 +47,7 @@ def test_all_pairs(filepath):
                 enforce_core_core=True,
                 complete_rings=False,
                 enforce_chiral=True,
+                min_threshold=0,
             )
 
             # useful for visualization
@@ -97,6 +98,7 @@ def test_complete_rings_only():
         enforce_core_core=True,
         complete_rings=True,
         enforce_chiral=True,
+        min_threshold=0,
     )
 
     assert len(all_cores) == 1
@@ -232,6 +234,7 @@ $$$$""",
         enforce_core_core=False,
         complete_rings=False,
         enforce_chiral=True,
+        min_threshold=0,
     )
 
     assert len(all_cores) == 1
@@ -252,6 +255,7 @@ $$$$""",
         enforce_core_core=True,
         complete_rings=False,
         enforce_chiral=True,
+        min_threshold=0,
     )
 
     # 2 possible matches, returned core ordering is fully determined
@@ -274,6 +278,7 @@ $$$$""",
         enforce_core_core=True,
         complete_rings=False,
         enforce_chiral=True,
+        min_threshold=0,
     )
 
     # 2 possible matches, if we do not allow for connected_core but do
@@ -404,6 +409,7 @@ def test_hif2a_failure():
         enforce_core_core=True,
         complete_rings=True,
         enforce_chiral=True,
+        min_threshold=0,
     )
 
     expected_core = np.array(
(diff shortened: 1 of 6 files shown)


tests/test_atom_mapping.py
```diff
@@ -90,12 +92,13 @@ def test_complete_rings_only():
         mol_b,
         ring_cutoff=0.1,
         chain_cutoff=0.2,
-        max_visits=1e7,  # 10 million max nodes to visit
+        max_visits=100,  # 100 max nodes to visit
```
badisa, line 95: Previously with this value set to 100 this would raise an exception, despite finding the expected mapping.

timemachine/fe/mcgregor.py
```diff
@@ -172,12 +185,15 @@ def mcs(
...
     priority_idxs = tuple(tuple(x) for x in priority_idxs)
-    start_time = time.time()
+    # Keep start time for debugging purposes below
+    start_time = time.time()  # noqa
 
     # run in reverse by guessing max # of edges to avoid getting stuck in minima.
     max_threshold = _arcs_left(marcs)
     for idx in range(max_threshold):
         cur_threshold = max_threshold - idx
+        if cur_threshold < min_threshold:
```
badisa, line 195: This is useful in the case where you have some idea of a lower bound on the number of atoms needed to map to be well converged.

timemachine/fe/mcgregor.py
```diff
@@ -212,7 +224,12 @@ def mcs(
         # f"==FAILED==[NODES VISITED {mcs_result.nodes_visited} | time taken: {time.time()-start_time} | time out? {mcs_result.timed_out}]====="
         # )
 
-    assert len(mcs_result.all_maps) > 0
+        # If timed out and no maps found, raise exception.
+        if mcs_result.timed_out:
```
badisa, line 228: It seemed like if you had found a map(s) and timed out, was worth returning it rather then raising the error. Perhaps worth having a warning if there is a map and it timed out? But it seems possible to find a map and still end up in a pretty large search space that you might not want to completely explore for performance reasons.

timemachine/fe/mcgregor.py
```diff
@@ -246,7 +263,6 @@ def recursion(
     marcs,
     mcs_result,
     priority_idxs,
-    start_time,
```
badisa, line 249: Just wasn't used anymore.

timemachine/fe/mcgregor.py
jkausrelay: Are you seeing this in practice? It was an assert b/c if the result timed out we're not sure how to interpret the mapping, i.e. it may or may not be optimal. There were no timeouts in all of the test cases that I tried.

tests/test_atom_mapping.py
jkausrelay: Does it make sense to test both 100, 1e7? Not sure how many nodes of the search tree are actually visited in this case.

timemachine/fe/mcgregor.py
badisa: I had not. I would agree it may not be optimal, it wasn't clear on if we want the true optimal every time. Just seemed like you might want to bound the search space and get back a mapping even if it isn't perfect.

tests/test_atom_mapping.py
badisa: I should make this a separate test, was just the quick case I came up with demonstrating a change in max visits failed without the change of returning a mapping even if things had timed out

timemachine/fe/mcgregor.py
```diff
@@ -212,7 +224,12 @@ def mcs(
         # f"==FAILED==[NODES VISITED {mcs_result.nodes_visited} | time taken: {time.time()-start_time} | time out? {mcs_result.timed_out}]====="
         # )
 
-    assert len(mcs_result.all_maps) > 0
+        # If timed out and no maps found, raise exception.
+        if mcs_result.timed_out:
+            raise MaxVisitsError(f"Reached max number of visits: {max_visits}")
+
+    if len(mcs_result.all_maps) <= 0:
```
jkausrelay, line 231: typo: `== 0`

jkausrelay (approved):

PR #927, opened 2022-12-20
Increase max frc norm to 50k

jkausrelay:
- Relative hydration free energy case was failing due to frc cut off during minimization
- Turning off check let simulation run without issue
- 50k comes from https://github.com/proteneer/timemachine/pull/904 (where I lowered the value to 25k to make it consistent with the steric clash check)


maxentile (approved):
Looks good!

(Just need to update https://github.com/proteneer/timemachine/blob/93acab5a523c204422eca9b3655b4a429e5e449f/tests/test_fe_utils.py#L166 from [4, 10, 11] to [10, 11] I think)

PR #928, opened 2022-12-20
Return full paths of results in RBFE driver

mcwitt:
Minor quality of life improvement to return the full path of the result from `run_edge_and_save_results`, e.g. `2022-12-20-14-34-62b78856/success_rbfe_result_254_50.pkl` instead of just `success_rbfe_result_254_50.pkl`. This should remove any guesswork from locating results and allow automating the download.


jkausrelay (approved):

maxentile (approved):
Looks good!

PR #931, opened 2023-01-06
Expand linear basis to support variable LJ sigma parameters

maxentile:
Continuation of https://github.com/proteneer/timemachine/pull/685  -- which only handled variable LJ **epsilon** parameters -- to handle also variable LJ **sigma** parameters.

Uses an approach (from eq. C.1. of Levi Naden's thesis) that:
* Increases the size of the basis expansion (from 1 to 20 numbers per ligand atom)
* Does not scale with the number of environment atom types
* Does not require switching [from arithmetic mixing rule](https://github.com/proteneer/timemachine/blob/9d1c9baa0db7c9313ed998d87e7b56fddfc3a417/timemachine/potentials/nonbonded.py#L31) to geometric mixing rule for sigma parameters

~TODO: update tests~

### DIFF
--- a/tests/test_jax_nonbonded.py
+++ b/tests/test_jax_nonbonded.py
@@ -403,8 +403,9 @@ def test_precomputation():
         return jnp.sum(vdw + es)
 
     @jit
-    def u_batch_ref(eps_ligand, q_ligand):
+    def u_batch_ref(sig_ligand, eps_ligand, q_ligand):
         new_params = jnp.array(params).at[ligand_idx, 0].set(q_ligand)
+        new_params = new_params.at[ligand_idx, 1].set(sig_ligand)
         new_params = new_params.at[ligand_idx, 2].set(eps_ligand)
 
         def f(x, box):
--- a/tests/test_jax_nonbonded.py
+++ b/tests/test_jax_nonbonded.py
@@ -195,7 +195,7 @@ def compare_two_potentials(u_a: NonbondedFxn, u_b: NonbondedFxn, args: Nonbonded
     energy_a, gradients_a = value_and_grads(u_a)(*args)
     energy_b, gradients_b = value_and_grads(u_b)(*args)
 
-    np.testing.assert_almost_equal(energy_a, energy_b)
+    np.testing.assert_allclose(energy_a, energy_b)
     for (g_a, g_b) in zip(gradients_a, gradients_b):
         np.testing.assert_allclose(g_a, g_b)
 
@@ -334,7 +334,7 @@ def test_jax_nonbonded_block():
 
         return jnp.sum(vdw + es)
 
-    np.testing.assert_almost_equal(u_a(conf, box, params), u_b(conf, box, params))
+    np.testing.assert_allclose(u_a(conf, box, params), u_b(conf, box, params))
 
 
 def test_precomputation():
@@ -419,11 +419,11 @@ def test_precomputation():
         actual = u_batch_test(sig_ligand, eps_ligand, q_ligand)
 
         # test array of energies is ~equal to reference
-        np.testing.assert_array_almost_equal(actual, expected)
+        np.testing.assert_allclose(actual, expected)
 
         # test that reweighting estimates and gradients are ~equal to reference
         v_ref, gs_ref = value_and_grad(reweight_ref, argnums=(0, 1, 2))(sig_ligand, eps_ligand, q_ligand)
         v_test, gs_test = value_and_grad(reweight_test, argnums=(0, 1, 2))(sig_ligand, eps_ligand, q_ligand)
 
-        np.testing.assert_almost_equal(v_ref, v_test)
-        np.testing.assert_array_almost_equal(gs_ref, gs_test)
+        np.testing.assert_allclose(v_ref, v_test)
+        np.testing.assert_allclose(gs_ref, gs_test)
--- a/tests/test_jax_nonbonded.py
+++ b/tests/test_jax_nonbonded.py
@@ -13,7 +13,7 @@ from jax import value_and_grad, vmap
 from scipy.optimize import minimize
 from simtk import unit
 
-from timemachine.constants import BOLTZ, DEFAULT_FF
+from timemachine.constants import BOLTZ, DEFAULT_FF, DEFAULT_TEMP
 from timemachine.fe.reweighting import one_sided_exp
 from timemachine.ff import Forcefield
 from timemachine.ff.handlers import openmm_deserializer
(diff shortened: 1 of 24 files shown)


maxentile:
For future reference, this optimization can also be re-discovered by applying the sympy functions `expand` and `factor` to the following function:

```python
from sympy import summation, symbols, IndexedBase, Idx

# variable atom sigma, epsilon
sig = symbols('sigma')
eps = symbols('epsilon')

# environment atoms i = 0 ... K
i = symbols('i', cls=Idx)
K = symbols('K', type=int)

eps_env = IndexedBase('epsilon^env')
sig_env = IndexedBase('sigma^env')

# distance between variable atom and each environment atom
r_env = IndexedBase('r^env')


def u(sig, eps):
    combined_sig = sig + sig_env[i]

    sig_r_12 = (combined_sig / r_env[i])**12
    sig_r_6 = (combined_sig / r_env[i])**6

    energy = eps * eps_env[i] * (sig_r_12 - sig_r_6)

    return summation(energy, (i, 0, K))
```

(more details in this notebook: https://gist.github.com/maxentile/c9cc8d8cff5d5758919c363bdb8744be )

I wasn't quickly able to reduce the number of basis functions from the current (12 + 1) + (6 + 1) = 20 with the help of sympy, but maybe further simplifications can be found in the future.

timemachine/potentials/nonbonded.py
```diff
@@ -565,31 +648,30 @@ def lj_prefactors_on_traj(traj, boxes, sigmas, epsilons, ligand_indices, env_ind
...
     eps_env = epsilons[env_indices]
     sig_env = sigmas[env_indices]
 
     def f_snapshot(coords, box):
         x_ligand = coords[ligand_indices]
         x_env = coords[env_indices]
-        return lj_prefactors_on_snapshot(x_ligand, x_env, sig_ligand, sig_env, eps_env, box, cutoff)
+        return lj_prefactors_on_snapshot(x_ligand, x_env, sig_env, eps_env, box, cutoff)
 
     return vmap(f_snapshot)(traj, boxes)
```
jkausrelay, line 659: @maxentile This is the part that can use a lot of memory. The intermediate computation of `r_env = jax_utils.distance_from_one_to_others(...)` is where most of the memory is allocated. I handled this by breaking up the traj into smaller segments, it could be useful to add this as an option to `lj_prefactors_on_traj`. Same with `coulomb_prefactors_on_traj`.

jkausrelay: 256 should be good (I've been using 200)

jkausrelay (approved):

timemachine/potentials/nonbonded.py
maxentile: Reduced to 200 to match https://github.com/proteneer/timemachine/pull/931/commits/27d0dc9711f7025980c4d907712e795b16b7bf9f

tests/test_jax_nonbonded.py
```diff
@@ -334,7 +337,38 @@ def u_b(x, box, params):
...
+    np.testing.assert_allclose(u_a(conf, box, params), u_b(conf, box, params))
+
+
+def test_lj_basis():
+    """Randomized test that LJ(sig, eps) computed via basis expansion matches reference"""
+
+    np.random.seed(2023)
+
+    n_env = 10_000
+    r_i = np.linspace(1.9, 8.0, n_env)
```
mcwitt, line 349: nit: comment on why these values were chosen?

timemachine/potentials/nonbonded.py
mcwitt: I might be missing something but it's not clear to me why `jax_utils.distance_from_one_to_others` should be allocating a lot of memory -- seems like it should just be linear in `x_others`?

timemachine/potentials/nonbonded.py
```diff
@@ -483,51 +490,133 @@ def coulomb_interaction_group_energy(q_ligand: Array, q_prefactors: Array) -> fl
...
+    h_n : [20] array
+        can be dotted with output of project_lj(sig, eps)
+        to compute energy of one atom interacting with all environment atoms
+    """
+    h_n_12 = _basis_expand_lj_term(sig_env, eps_env, r_env, 12)
+    h_n_6 = -_basis_expand_lj_term(sig_env, eps_env, r_env, 6)
+    return jnp.hstack([h_n_12, h_n_6])
+
+
+def project_lj(sig: float, eps: float) -> Array:
```
mcwitt, line 560: nit: initially I was a bit thrown off by the naming here, since I'm used to thinking of a projection as something that discards information, but I think this mapping is invertible? Is there another reason to call it a projection?

tests/test_jax_utils.py
```diff
@@ -181,3 +182,23 @@ def test_pairwise_distances_periodic_lifting(coords_box_w):
...
+    traj = np.random.randn(T, N, 3)
+    boxes = np.random.randn(T, 3, 3)
+
+    def f_snapshot(x, box):
+        return jnp.sum(x ** 2) + jnp.sum(box ** 3)  # arbitrary fxn
+
+    reference = vmap(f_snapshot)(traj, boxes)
+    actual = process_traj_in_chunks(f_snapshot, traj, boxes)
+
+    np.testing.assert_allclose(actual, reference)
```
mcwitt, line 204: Should this test different `chunk_size`s? (including edge cases `chunk_size == 1`, `chunk_size == T`, and `T % chunk_size != 0`)

tests/test_jax_nonbonded.py
maxentile: Willfix.

(This range was inherited from an initial plot using combined_sigma = fixed_value_near_1 + variable_value_near_1`. The lower end of the range was chosen to be just under 2 to avoid too much clashing, and the upper end of the range was chosen arbitrary and large.)

timemachine/potentials/nonbonded.py
maxentile: This name is not well thought-out -- happy to change to any preferred alternative.

mcwitt: Hmm, would something like

- `basis_expand_lj` -> `basis_expand_lj_env`
- `project_lj` -> `basis_expand_lj_atom`

make sense?

tests/test_jax_utils.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/931/commits/bc584186a026e5859070a70c1d3291bff0f79baf

mcwitt: Could this assert equality?

timemachine/potentials/nonbonded.py
maxentile: Sounds good to me -- applied in https://github.com/proteneer/timemachine/pull/931/commits/8dc364cf0a1d01eaf9230afff773c6795b998619

tests/test_jax_utils.py
maxentile: Yep, strengthened in https://github.com/proteneer/timemachine/pull/931/commits/f06fee7ce1a1f4730c820032350275fb7ee54410

mcwitt (approved):
Looks great to me!

I'm still a little confused about why batching helps significantly with memory usage, but it doesn't seem like it would hurt and I don't think this should block merging.

timemachine/potentials/nonbonded.py
maxentile: I think the size of this intermediate array may be something like `traj_length * n_ligand_atoms * n_environment_atoms * 3` (but I'm not sure). Although linear in `traj_length`, the constant factor is probably large.

Follow-ups could be:
* to provide Jax some clearer hint to "fuse" the computation distance matrix with the sum reduction / avoid this large intermediate array
* to write a "custom op" implementation that uses the neighborlist

I'm leaning towards just doing the second option, since
(1) there would also be utility in computing these on the fly (to avoid ever having to store trajectories),
(2) I don't foresee a need to use any other jax transformations with these preprocessing functions (I don't see a compelling reason to keep the implementation in Jax if doing so is inconvenient / introduces surprising resource requirements)

PR #933, opened 2023-01-10
Fix name-swap typo

maxentile:
Fix typo from https://github.com/proteneer/timemachine/pull/931 -- had swapped the names of `basis_expand_lj_env` and `basis_expand_lj_atom`


mcwitt (approved):
Ah good catch, I'd missed that in review

PR #934, opened 2023-01-17
Updates default FF to OpenFF 2.0.0

badisa:
* Fixes sulfonamide angles (OFF 1.3.1 release)
* Adds 4 periodic torsion patterns (OFF 1.3.* release)
* Adds 2 LJ patterns
* Performance on JACS/Merck is slightly better, though within precision of rerunning simulations

### DIFF
--- a/tests/test_terminal_bond_maps.py
+++ b/tests/test_terminal_bond_maps.py
@@ -162,7 +162,6 @@ def test_on_methane():
     mbar = MBAR(u_kn, N_k)
     estimated_delta_f_mbar = mbar.f_k[1]
 
-    # assert that all estimates are close to precomputed result
     estimates = np.array(
         [
             estimated_delta_f_forward,
@@ -171,4 +170,5 @@ def test_on_methane():
             estimated_delta_f_mbar,
         ]
     )
-    np.testing.assert_allclose(estimates, -2.8, atol=1e-1)
+    # assert that all estimates are in agreement
+    np.testing.assert_allclose(estimates[:3], estimates[-1], atol=1e-1)
--- a/tests/test_terminal_bond_maps.py
+++ b/tests/test_terminal_bond_maps.py
@@ -171,4 +171,4 @@ def test_on_methane():
             estimated_delta_f_mbar,
         ]
     )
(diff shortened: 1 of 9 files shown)


tests/test_terminal_bond_maps.py
```diff
@@ -171,4 +171,4 @@ def test_on_methane():
             estimated_delta_f_mbar,
         ]
     )
-    np.testing.assert_allclose(estimates, -4.2, atol=1e-1)
+    np.testing.assert_allclose(estimates, -2.8, atol=1e-1)
```
badisa, line 174: @maxentile I changed this value to get the test to pass, not sure if the old value had more meaning then an arbitrary value.

maxentile: Not arbitrary (precomputed with previous DEFAULT_FF), but not meaningful --

may just delete outdated comment above: `# assert that all estimates are close to precomputed result`

or modify the test to compare the first 3 estimates to the 4th estimate (and avoid hard-coding a value)

maxentile (approved):

tests/test_vacuum_importance_sampling.py
```diff
@@ -5,13 +5,14 @@
...
 from timemachine import testsystems
+from timemachine.constants import DEFAULT_FF
 from timemachine.ff import Forcefield
 from timemachine.md import enhanced
 from timemachine.potentials import bonded
 
 
 def get_ff_am1ccc():
-    ff = Forcefield.load_from_file("smirnoff_1_1_0_ccc.py")
+    ff = Forcefield.load_from_file(DEFAULT_FF)
```
mcwitt, line 15: random idea for quality-of-life improvement: `Forcefield.load_default()`? (not necessary or related to this PR)

mcwitt (approved):

tests/test_vacuum_importance_sampling.py
badisa: It would be nice to make forcefield loading in the tests more consistent.  `Forcefield.load_default()` seems like a good approach. Perhaps a fixture as well?

PR #935, opened 2023-01-18
Add numerically stable form of harmonic angle potential

mcwitt:
### Changes
- [x] Add `HarmonicAngleStable`, reference implementation, and tests
- [x] ~Add `PeriodicTorsionStable`~ (seems significantly more complicated, plan to reassess and potentially add in a separate PR)

### Motivation
The singularities present in the existing functional forms for harmonic angle and torsion energies, e.g.

1. infinite harmonic angle force when one bond shrinks to zero length
2. infinite torsion force when 3 atoms are collinear

complicate alchemical transformations, since we need to be careful to avoid regions of parameter space where there are potential numerical instabilities (e.g. we currently need to "stagger" turning on bond and angle forces in some bond formation transformations to avoid encountering (1) above).

This PR aims to introduce approximate functional forms for angle and torsion forces with these singularities removed. These have a single additional parameter, $\epsilon$, controlling the approximation, with the property that we recover the original (singular) form as $\epsilon \to 0$.

In alchemical transformations, $\epsilon$ provides an additional knob that we can use to improve efficiency. For example, prior numerical experiments in vacuum showed that we can use a small constant $\epsilon$ value in intermediate states to allow turning on bonds and angle forces simultaneously for bond formation transformations, thus improving efficiency significantly.

### Review notes
- the expressions for du/dx and du/dp were worked out with the help of sympy; notebook here: https://gist.github.com/mcwitt/7c31e296a1292536014e156ffc4b4b58 (see cells 18 and below)


timemachine/cpp/src/kernels/k_harmonic_angle_stable.cuh
```diff
@@ -0,0 +1,91 @@
...
+    if (du_dx) {
+        for (int d = 0; d < 3; d++) {
+            RealType g_i = cij * rij[d] - rkj[d];
+            atomicAdd(du_dx + i_idx * 3 + d, FLOAT_TO_FIXED_BONDED<RealType>(c * g_i));
+
+            // Use __dadd_rd instead of (+) operator to prevent FMA
+            // optimization, which breaks bitwise equivalence for the
+            // symmetry (i, j, k) -> (k, j, i).
+            RealType g_j = __dadd_rd((1 - cij) * rij[d], (1 - ckj) * rkj[d]);
+            atomicAdd(du_dx + j_idx * 3 + d, FLOAT_TO_FIXED_BONDED<RealType>(c * g_j));
```
maxentile, line 69: 1. Is there an alternative that preserves bitwise equivalence but avoids intrinsics? (e.g. converting to fixed-point first? But this may introduce worse precision loss...)

2. If unavoidable, would it be useful to extract float and double versions of this intrinsic to `gpu_utils` : `strict_add(a, b) = __dadd_rn(a, b)` or `__fadd_rn(a, b)`, `strict_mul(a, b) = __fmul_rn(a, b)` or `__dmul_rn(a, b)` (as in https://github.com/proteneer/timemachine/pull/599/files#diff-260d713349dc8e00febdb5463a13b489219f40047075484128fde22bfd65adb6 )

(Just looks odd to template on RealType = float or double, but use an intrinsic specialized to double)

3. Nit: is the rounding mode intentional? (Round-nearest mode is used elsewhere in timemachine when intrinsics are necessary to prevent formation of FMAs, but this function uses round-down)

timemachine/potentials/bonded_stable.py
```diff
@@ -0,0 +1,56 @@
...
+    params: shape [num_angles, 3] np.ndarray
+        force constants, eq angles, epsilons
+        (kas, a0s, epsilons = params.T)
+
+    angle_idxs: shape [num_angles, 3] np.ndarray
+        each element (a, b, c) is a unique angle in the conformation. atom b is defined
+        to be the middle atom.
+
+    cos_angles: True (default)
+        if True, then this instead implements V(t) = k*(cos(t)-cos(t0))^2. This is far more
```
maxentile, line 32: nit: k/2 ?

timemachine/potentials/bonded_stable.py
```diff
@@ -0,0 +1,56 @@
...
+
+    The functional form is identical to :py:func:`potentials.bonded.HarmonicAngle`, except that the following
+    approximation is used for the intermediate computation of :math:`\cos(\theta)`:
+
+    :math::
+
+        \cos(\theta) \approx \frac{r_{ij} \cdot r_{kj} + \epsilon^2}{\sqrt{(r_{ij}^2 + \epsilon^2) (r_{kj}^2 + \epsilon^2)}}
+
+    This reduces to the exact expression when :math:`\epsilon = 0`; When :math:`\epsilon > 0`, this avoids the
+    singularities in the exact expression as :math:`r_{ij}` or :math:`r_{kj}` approach zero.
```
maxentile, line 16: Docs tweak: May clarify that epsilons are used to stabilize both the cos_angles=True and cos_angles=False code paths?

timemachine/cpp/src/kernels/k_harmonic_angle_stable.cuh
mcwitt: 1. Possibly, but I haven't found it after spending some time looking. I don't think converting to fixed-point before the final multiplication is the right approach, since it seems like this would exacerbate precision loss. (I think the situation here is similar to the cross product in the periodic torsion potential, where we resort to intrinsics to ensure bitwise equality: https://github.com/proteneer/timemachine/blob/5c8a0e1cd315cd6f8e4f6481f52efca8389ffa15/timemachine/cpp/src/kernels/k_periodic_torsion.cuh#L13)
2. Good call; I think it's incorrect as-is (not sure why the tests still pass in single-precision mode). Willfix.
3. Willfix

timemachine/potentials/bonded_stable.py
mcwitt: Done in eda6999

timemachine/potentials/bonded_stable.py
mcwitt: Done in https://github.com/proteneer/timemachine/pull/935/commits/eda6999e88d7eb85376c1c5b4cacbdd2ad14d78b

timemachine/cpp/src/kernels/k_harmonic_angle_stable.cuh
mcwitt: > e.g. converting to fixed-point first?

Actually I think this is possible without losing precision if we do something like:

```python
            unsigned long long g_j1 = FLOAT_TO_FIXED_BONDED(c * (1 - cij) * rij[d]);
            unsigned long long g_j2 = FLOAT_TO_FIXED_BONDED(c * (1 - ckj) * rkj[d]);
            atomicAdd(du_dx + j_idx * 3 + d, g_j1 + g_j2);
```

I tested this, but ended up opting for using a wrapper around the intrinsic because

1. (main reason) this version no longer matches the reference implementation with high precision (I think the reference encounters roundoff error in the final floating point addition)
2. this seems a bit less efficient (though haven't benchmarked, likely insignificant)

mcwitt: Fixed (2) and (3) in https://github.com/proteneer/timemachine/pull/935/commits/1b822e48f61f742b0b09cbc97416e82f8e8e90f7

maxentile: Nice!

>ended up opting for using a wrapper around the intrinsic because ...

This looks good to me

maxentile (approved):

jkausrelay (approved):

timemachine/potentials/bonded_stable.py
```diff
@@ -0,0 +1,57 @@
...
+    -----------
+    conf: shape [num_atoms, 3] np.ndarray
+        atomic coordinates
+
+    params: shape [num_angles, 3] np.ndarray
+        force constants, eq angles, epsilons
+        (kas, a0s, epsilons = params.T)
+
+    angle_idxs: shape [num_angles, 3] np.ndarray
+        each element (a, b, c) is a unique angle in the conformation. atom b is defined
```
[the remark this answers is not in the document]
mcwitt: Fixed in https://github.com/proteneer/timemachine/pull/935/commits/ba83ffd9d7c98dc9949d0beffefb411f06873911

PR #937, opened 2023-01-20
NB failure: test_condensed_phase_mtm

jkausrelay:
- See if increasing num_batches helps with test_condensed_phase_mtm failure after switch to Openff 2.0.0
- Increase by 10x, only increases time to run test by a few minutes
- Can't reproduce the error locally, but the torsion plot looks smoother with 400k batches:

num_batches = 40000 
test_condensed_phase_mtm

num_batches = 400000
test_condensed_phase_mtm 400k

### DIFF
--- a/tests/test_mtm_condensed.py
+++ b/tests/test_mtm_condensed.py
@@ -44,7 +44,7 @@ def test_condensed_phase_mtm():
     cache_path = "mtm_condensed_cache.pkl"
     if not os.path.exists(cache_path):
         print("Generating cache")
-        num_batches = 40000
+        num_batches = 400000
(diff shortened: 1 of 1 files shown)


maxentile (approved):
Looks good.

I think @badisa mentioned this wasn't resolved by 3x sampling (maybe the threshold needs to be updated https://github.com/proteneer/timemachine/commit/e2d63a01214c9f8044437b6d49c9cf8cb329ba26 ), but exhaustive sampling seems worth a shot.

Later, we might want to consider again the possibility of relaxing the requirement that good samples must be generated by 1 step of importance sampling. (Which I suspect is contributing to brittleness w.r.t. small changes in FF params, and high variance in the estimated histogram bin heights in general.)

PR #938, opened 2023-01-21
Use stable harmonic angles, simultaneously turn on bonds and angles

mcwitt:
### Summary

Changes single topology transformations to
- interpolate bond and angle force constants simultaneously (instead of staggering)
- use `HarmonicAngleStable` (introduced in #935) instead of `HarmonicAngle`; this is necessary to ensure numerical stability when we turn on bonds and angles simultaneously
  - we use a small, positive $\epsilon$ for intermediate states only. End states have $\epsilon=0$, therefore reducing to the standard functional form

Also makes a small correction/simplification to the `HarmonicAngleStable` functional form: removing $\epsilon^2$ in the numerator, which was not necessary for the goal of numerical stability, and breaks a symmetry by biasing the numerator toward positive values.

### Motivation

Removing the staggering of bonds and angles
- is now safe with the introduction of a numerically stable harmonic angle potential
- decreases BAR errors of hif2a and tyk2 transformations (see [Benchmark results](#bench-results) below)
- simplifies optimal interpolation logic (e.g. makes unnecessary a proposed refinement that would e.g. disable staggering of angle terms that do not span a bond being formed)

#### hif2a

$\Delta \Delta G$ errors relative to reference (9e337214fe2b84fb0fd3c35df3c6fb6b5153a6b2):

|   mean |   std |   min |   25% |   50% |   75% |   max |
|-------:|------:|------:|------:|------:|------:|------:|
|  0.883 | 0.829 |  1.01 | 0.997 | 0.946 | 0.811 | 0.996 |

By transformation type:

| ring_transformation   |   mean |   std |   min |   25% |   50% |   75% |   max |
|:----------------------|-------:|------:|------:|------:|------:|------:|------:|
| closing               |  0.813 | 0.675 | 0.86  | 0.826 | 0.828 | 0.8   | 0.81  |
| none                  |  0.946 | 0.985 | 1.01  | 0.923 | 0.946 | 0.947 | 0.996 |
| opening               |  0.83  | 0.831 | 0.863 | 0.807 | 0.819 | 0.822 | 0.843 |

#### tyk2

|   mean |   std |   min |   25% |   50% |   75% |   max |
|-------:|------:|------:|------:|------:|------:|------:|
|  0.919 | 0.977 |     1 | 0.877 | 0.947 | 0.885 |  1.01 |

### DIFF
--- a/tests/test_velocity_verlet_integrator.py
+++ b/tests/test_velocity_verlet_integrator.py
@@ -98,7 +98,7 @@ def test_reversibility():
     dt = 1.5e-3
 
     # Is not infinitely reversible, will fail after 3000 steps due to accumulation of coords/vels in floating point
-    for n_steps in [1, 10, 100, 500, 1000, 2000]:
+    for n_steps in [1, 10, 100, 500]:
         # Note: reversibility can fail depending on the
         # range of values in the velocities. Setting the seed
         # here keeps the range the same for all n_step values.
--- a/timemachine/fe/single_topology.py
+++ b/timemachine/fe/single_topology.py
@@ -1055,27 +1055,14 @@ class SingleTopology(AtomMapMixin):
 
         return potentials.ChiralBondRestraint(chiral_bond_idxs, chiral_bond_signs).bind(chiral_bond_params)
 
-    def setup_intermediate_state(self, lamb, lambda_angles=0.4, lambda_torsions=0.7):
-        """
+    def setup_intermediate_state(self, lamb):
+        r"""
         Setup intermediate states at some value of lambda.
 
         Parameters
         ----------
         lamb: float
-
-        lambda_angles, lambda_torsions: float
-            For ring opening/closing transformations, alchemical parameter values controlling the intervals over which
-            bonds, angles, and torsions are interpolated. Note that these have no effect on terms not involved in ring
-            opening/closing.
-
-            - Bonds are interpolated in the interval 0 <= lamb <= lambda_angles
-            - Angles are interpolated in the interval lambda_angles <= lamb <= lambda_torsions
-            - Torsions are interpolated in the interval lambda_torsions <= lamb <= 1
-
-            Note: must satisfy 0 < lambda_angles < lambda_torsions < 1
         """
-
-        assert 0.0 < lambda_angles < lambda_torsions < 1.0
 
         src_system = self.src_system
         dst_system = self.dst_system
@@ -1089,7 +1076,7 @@ class SingleTopology(AtomMapMixin):
                 interpolate_harmonic_bond_params,
                 k_min=0.1,  # ~ BOLTZ * (300 K) / (5 nm)^2
                 lambda_min=0.0,
-                lambda_max=lambda_angles,
+                lambda_max=0.7,
             ),
         )
 
@@ -1101,8 +1088,8 @@ class SingleTopology(AtomMapMixin):
             partial(
                 interpolate_harmonic_angle_params,
                 k_min=0.05,  # ~ BOLTZ * (300 K) / (2 * pi)^2
-                lambda_min=lambda_angles,
-                lambda_max=lambda_torsions,
+                lambda_min=0.0,
+                lambda_max=0.7,
             ),
         )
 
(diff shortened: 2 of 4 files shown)


tests/test_velocity_verlet_integrator.py
```diff
@@ -98,7 +98,7 @@ def test_reversibility():
     dt = 1.5e-3
 
     # Is not infinitely reversible, will fail after 3000 steps due to accumulation of coords/vels in floating point
-    for n_steps in [1, 10, 100, 500, 1000, 2000]:
+    for n_steps in [1, 10, 100, 500]:
```
maxentile, line 101: max n_steps = 100 or 500 seems ample to me

(main motivation of this test was to assert suitability for use in HMC, which (1) is not yet incorporated in TM, (2) would not need to use infinitely long proposal trajectories, (3) could use an on-the-fly reversibility check if absolutely needed, as in 4.3 of https://arxiv.org/abs/2003.03950 (and references therein)... the previous large max n_steps may have been derived from tests on the reference implementation, which may be less impacted by precision loss)

timemachine/potentials/bonded_stable.py
```diff
@@ -43,7 +43,7 @@ def harmonic_angle_stable(conf, params, angle_idxs, cos_angles=True):
     vij = ci - cj
     vkj = ck - cj
 
-    top = jnp.sum(vij * vkj, -1) + eps ** 2
+    top = jnp.sum(vij * vkj, -1)
```
mcwitt, line 46: The eps^2 term here was unnecessary for stability and introduced an asymmetry biasing the numerator in the positive direction.

timemachine/fe/single_topology.py
```diff
@@ -1053,28 +1055,15 @@ def _setup_intermediate_chiral_bond_term(self, src_bond, dst_bond, lamb, interpo
...
-        """
+    def setup_intermediate_state(self, lamb):
+        r"""
         Setup intermediate states at some value of lambda.
 
         Parameters
         ----------
         lamb: float
-
-        lambda_angles, lambda_torsions: float
```
badisa, line 1064: Some of the information in this comment seems like useful context, perhaps keeping the bits about how the interpolation works generally would be good.

timemachine/fe/single_topology.py
```diff
@@ -681,7 +681,9 @@ def interpolate_harmonic_angle_params(src_params, dst_params, lamb, k_min, lambd
         lamb,
     )
 
-    return jnp.array([k, phase])
+    eps = jnp.where((lamb == 0.0) | (lamb == 1.0), 0.0, 1e-3)
```
badisa, line 684: Should we have a constant for this epsilon value? Not sure how much giving it a different value matters or not.

badisa:
Was there any impact on the BAR overlap going from the endpoints where `eps == 0` to intermediate simulations at `eps == 1e-3`?

mcwitt:
@badisa there does not seem to be a significant difference in overlap between the end states and the neighboring intermediates when comparing the control and test versions:

imagea141-446c14741ff7.png)

I think this makes sense: under the assumption that end states are stable, we'd expect end state samples to be far away from singularities, in the region where `eps=1e-3` looks basically identical to `eps=0`.

timemachine/fe/single_topology.py
mcwitt: This was chosen somewhat arbitrarily to be large enough that no numerical instabilities were observed, and small enough to minimize the impact on end-state overlaps. Maybe we could add a comment to that effect? (Seemed premature to declare a constant, since I think this part of the code is relatively likely to change and it seemed useful to keep the definition local)

badisa: A comment would work for me

timemachine/fe/single_topology.py
mcwitt: Restored and expanded in https://github.com/proteneer/timemachine/pull/938/commits/8939e562b10f590e032e03be38371f2794fe7db4

timemachine/fe/single_topology.py
mcwitt: Added a comment in https://github.com/proteneer/timemachine/pull/938/commits/e52aa61342ef43b1d913fdfc3941e25713ee08ed

maxentile (approved):

badisa (approved):

jkausrelay (approved):

PR #939, opened 2023-01-25
Strengthen symmetry check to test with log linear interpolation

mcwitt:
Based on discussion, strengthen check for A <-> B symmetry in the case of ring opening/closing to include log linear interpolation.

### DIFF
--- a/tests/test_single_topology.py
+++ b/tests/test_single_topology.py
@@ -6,13 +6,13 @@ import jax
 import jax.numpy as jnp
 import numpy as np
 import pytest
-from hypothesis import given, seed
+from hypothesis import assume, given, seed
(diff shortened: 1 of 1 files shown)


badisa (approved):
🚀

maxentile (approved):
Looks good!

This may already be covered, but is there also a test that shortest-path interpolation for equilibrium angles (in HarmonicAngles https://github.com/proteneer/timemachine/blob/587929765fcf44f11361b65a2c933f498368e8ea/timemachine/fe/single_topology.py#L680 and PeriodicTorsions) is symmetric?

maxentile:
Ahh, makes sense! Thanks

PR #941, opened 2023-01-25
Bootstrap node dg errs

maxentile:


timemachine/fe/mle.py
```diff
@@ -104,3 +104,96 @@ def loss(x):
...
+        edge_idxs,
+        edge_diffs,
+        edge_stddevs,
+        ref_node_idxs,
+        ref_node_vals,
+        ref_node_stddevs,
+        n_bootstrap,
+        seed,
+    )
+    dg = bootstrap_estimates.mean(0)
```
maxentile, line 197: decision: report mean of bootstrap samples, or result of `infer_node_vals` without bootstrap?

timemachine/fe/mle.py
```diff
@@ -104,3 +104,96 @@ def loss(x):
...
+    dg_err = bootstrap_estimates.std(0)
```
maxentile, line 198: decision: print a diagnostic if these estimates are somehow suspicious?

(no diagnostic immediately comes to mind, unlike in https://github.com/proteneer/timemachine/pull/829 )

timemachine/fe/mle.py
```diff
@@ -104,3 +104,96 @@ def loss(x):
...
     return result - result[0]
+
+
+def _bootstrap_node_vals(
+    edge_idxs,
+    edge_diffs,
+    edge_stddevs,
+    ref_node_idxs,
+    ref_node_vals,
+    ref_node_stddevs,
```
maxentile, line 115: note: in contrast to https://github.com/proteneer/timemachine/pull/399 , this function does not support a mix of relative and absolute predictions as input.

(absolute references are assumed to be from experiment, and only influence the result via an additive offset regardless the number of references)

timemachine/fe/mle.py
```diff
@@ -104,3 +104,96 @@ def loss(x):
...
+
+    def estimate_node_val_w_offset(edge_vals, node_vals):
+        estimated_node_vals = infer_node_vals(edge_idxs, edge_vals, edge_stddevs)
+        estimated_node_offset = np.mean(node_vals - estimated_node_vals[ref_node_idxs])
+        return estimated_node_vals + estimated_node_offset
+
+    bootstrap_estimates = np.zeros((n_bootstrap, n_nodes))
+
+    for i in range(n_bootstrap):
+        noisy_edge_diffs = edge_diffs + rng.standard_normal(n_edges) * edge_stddevs
```
maxentile, line 141: note: this part assumes gaussian edge error and is coupled to the definition above `edge_lls = norm.logpdf`

if we switch to heavy-tailed or other edge likelihoods as suggested here https://github.com/proteneer/timemachine/pull/399#issuecomment-826850602 , this line should be updated, e.g. to `rng.standard_t`...

timemachine/fe/mle.py
badisa: Could we test this on some graphs we have already run and see how much the mean and the default returns?

maxentile: Fixed in https://github.com/proteneer/timemachine/pull/941/commits/a50b66faee2a23fc36105d88f6e368be953d0f65

timemachine/fe/mle.py
```diff
@@ -70,7 +70,7 @@ def wrapped(x):
     return wrapped
 
 
-def infer_node_vals(edge_idxs, edge_diffs, edge_stddevs, ref_node_idxs, ref_node_vals):
+def infer_node_vals(edge_idxs, edge_diffs, edge_stddevs, ref_node_idxs=[], ref_node_vals=[]):
```
badisa, line 73: I would avoid mutable data structures as defaults. Empty tuples would be safer

maxentile: Applied in https://github.com/proteneer/timemachine/pull/941/commits/71a963a69329836bdd871b791cf6202b5b97e0f2

timemachine/fe/mle.py
badisa: Not clear what 'definition above' you are referring to. Don't see an `edge_lls`

maxentile: https://github.com/proteneer/timemachine/blob/71a963a69329836bdd871b791cf6202b5b97e0f2/timemachine/fe/mle.py#L41 , introduced in https://github.com/proteneer/timemachine/pull/932

badisa: Thanks for clarifying

timemachine/fe/mle.py
```diff
@@ -70,37 +70,139 @@ def wrapped(x):
...
     Parameters
     ----------
     edge_idxs: [E, 2] array, where the values are in range(K).
     edge_diffs: [E] array, relative values
     edge_stddevs: [E] array, positive
 
+    ref_node_idxs: [N_ref] int array
+    ref_node_vals: [N_ref] array
+
+    TODO: incorporate ref_node_stddevs here?
```
badisa, line 90: This should probably be made an inline comment rather then part of the parameters docstring

badisa (approved):

timemachine/fe/mle.py
jkausrelay: Can you add a comment here and at line 41 to make sure these are synchronized?

jkausrelay (approved):
Minor comment, LGTM.

timemachine/fe/mle.py
maxentile: Applied in https://github.com/proteneer/timemachine/pull/941/commits/d09177cb6e566dc0bd340ae4b9b4cbe0261f5ca2

PR #944, opened 2023-01-26
Skips installation of Custom ops automatically

badisa:
* If not linux, if env variable set or if there is no nvidia-smi
* Don't install cmake on anything but linux

Please try running the following on your linux and OSX boxes to make sure this works everywhere, works on my machine but my machines have known to lie.

```
pip install 'timemachine @ git+https://github.com/proteneer/timemachine@a07a2e3437503d4a03115c39b519a0024764f3a5'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,20 +45,11 @@ conda install openmm=7.5.1 -c conda-forge # only if using openmm from conda
 
 ### Install Time Machine
 
-#### Linux
+The CUDA extension module implementing custom ops is only supported on Linux, but partial functionality is still available on non-Linux OSes.
 
 ```shell
 pip install -r requirements.txt
 pip install .
-```
-
-#### Non-Linux
-
-The CUDA extension module implementing custom ops is only supported on Linux, but partial functionality is still available on non-Linux OSes. To install without the extension:
-
-```shell
-pip install -r requirements.txt
-SKIP_CUSTOM_OPS=1 pip install .
 ```
 
 ## Developing Time Machine
@@ -75,10 +66,7 @@ Possible variants of the last step include
 ```shell
 pip install -e .[dev,test]                 # optionally install dev and test dependencies
 CMAKE_ARGS=-DCUDA_ARCH=86 pip install -e . # override CUDA_ARCH
-SKIP_CUSTOM_OPS=1 pip install -e .         # skip building CUDA extension (non-Linux)
-
-# use parallel CMake build with `nproc` threads
-CMAKE_BUILD_PARALLEL_LEVEL=$(nproc) pip install -e .
+SKIP_CUSTOM_OPS=1 pip install -e .         # skip building CUDA extension
 ```
 
 To rebuild the extension module after making changes to the C++/CUDA code, either rerun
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -15,6 +15,24 @@
...
+
+    If it is not a linux machine and doesn't have at least nvidia-smi we skip it. Can
+    still use the reference platform if needed in such cases.
+    """
+    if os.environ.get("SKIP_CUSTOM_OPS"):
+        return False
+    if "linux" not in sys.platform:
+        return False
+    try:
+        subprocess.check_call(["nvidia-smi"])
```
mcwitt, line 29: nit: should this check for `nvcc`? This seems more accurate in the edge case where the toolkit but not runtime environment is installed.

setup.py
```diff
@@ -57,6 +75,10 @@ def build_extension(self, ext):
 cmdclass = versioneer.get_cmdclass()
 cmdclass.update(build_ext=CMakeBuild)
 
+external_modules = None
```
mcwitt, line 78: nit: naming: I think `ext_modules` stands for "extension modules" (could just use the same name as the kwarg?)

README.md
```diff
@@ -75,10 +66,7 @@ Possible variants of the last step include
 ```shell
 pip install -e .[dev,test]                 # optionally install dev and test dependencies
 CMAKE_ARGS=-DCUDA_ARCH=86 pip install -e . # override CUDA_ARCH
-SKIP_CUSTOM_OPS=1 pip install -e .         # skip building CUDA extension (non-Linux)
-
-# use parallel CMake build with `nproc` threads
-CMAKE_BUILD_PARALLEL_LEVEL=$(nproc) pip install -e .
+SKIP_CUSTOM_OPS=1 pip install -e .         # skip building CUDA extension
```
mcwitt, line 69: nit: maybe add "(no effect on non-Linux)" to comment?

mcwitt (commented):
Nice! This works correctly for me on macOS and Ubuntu. It breaks in my NixOS environment, but I think there's an easy fix: check for `nvcc` instead of `nvidia-smi`. In general it seems like the former is more accurate and should allow building without a GPU available (as long as the CUDA toolkit is installed).

README.md
badisa: Fixed in https://github.com/proteneer/timemachine/pull/944/commits/7d372f6d16783b07cb916a1f350e6efb4a3b32a3

setup.py
badisa: Fixed in https://github.com/proteneer/timemachine/pull/944/commits/7d372f6d16783b07cb916a1f350e6efb4a3b32a3

setup.py
badisa: Good call, fixed in https://github.com/proteneer/timemachine/pull/944/commits/7d372f6d16783b07cb916a1f350e6efb4a3b32a3

setup.py
```diff
@@ -15,6 +15,24 @@
...
+
+    If it is not a linux machine and doesn't have at least nvcc we skip it. Can
+    still use the reference platform if needed in such cases.
+    """
+    if os.environ.get("SKIP_CUSTOM_OPS"):
+        return False
+    if "linux" not in sys.platform:
+        return False
+    try:
+        subprocess.check_call(["nvcc"])
```
mcwitt, line 29: This fails for me locally because `nvcc` returns a failure status when passed no files. `nvcc --version` seems to work for me, though.

mcwitt (approved):

jkausrelay (approved):

maxentile (approved):

PR #947, opened 2023-01-30
DRY, minor fix in RBFE test

mcwitt:
- Removes duplicated `assert solvent_res.overlap_summary_png is not None`
- Adds missing `assert complex_res.overlap_summary_png is not None`
- Strengthens test to assert expected length of `all_dGs`, shape consistency of `dG_errs_by_lambda_by_component`, `overlaps_by_lambda_by_component`
- Fixes incorrect type annotation of `all_dGs` in `SimulationResult`

### DIFF
--- a/tests/test_relative_free_energy.py
+++ b/tests/test_relative_free_energy.py
@@ -94,25 +94,25 @@ def run_triple(mol_a, mol_b, core, forcefield, n_frames, protein_path, n_eq_step
         n_eq_steps=n_eq_steps,
     )
 
-    assert solvent_res.overlap_summary_png is not None
-    assert solvent_res.overlap_detail_png is not None
-    assert np.linalg.norm(solvent_res.all_errs) < 0.1
-    assert len(solvent_res.frames[0]) == n_frames
-    assert len(solvent_res.frames[-1]) == n_frames
-    assert len(solvent_res.boxes[0]) == n_frames
-    assert len(solvent_res.boxes[-1]) == n_frames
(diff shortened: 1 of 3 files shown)


tests/test_relative_free_energy.py
```diff
@@ -94,25 +94,25 @@ def run_triple(mol_a, mol_b, core, forcefield, n_frames, protein_path, n_eq_step
...
-    assert len(solvent_res.boxes[-1]) == n_frames
-    assert [x.lamb for x in solvent_res.initial_states] == lambda_schedule
-    assert solvent_res.protocol.n_frames == n_frames
-    assert solvent_res.protocol.n_eq_steps == n_eq_steps
-
-    def check_overlaps(result: SimulationResult):
+    def check_result(result: SimulationResult):
+        assert result.overlap_summary_png is not None
+        assert result.overlap_detail_png is not None
+        assert np.linalg.norm(result.all_errs) < 0.1
```
badisa, line 100: nit: Would be good to validate that `len(result.all_errs) == n_frames` as `np.linalg.norm(np.array([])) == 0.0`

badisa: nit: Might also be good to check the length of `result.all_dGs` as well

badisa (approved):
Good catch!

tests/test_relative_free_energy.py
mcwitt: Added in https://github.com/proteneer/timemachine/pull/947/commits/d2bc78a076afe717ac434d1ffc9368b5991c2a5e

PR #936, opened 2023-01-18
[WIP] Refactor rbfe.py, partially separating simulation and analysis

maxentile:
Very WIP.

Main goal: Simplify the function `rbfe.estimate_free_energy_given_initial_states`, which is currently in charge of many things. (Sequentially running MD simulations, computing energy decompositions, computing free energy estimates, generating detailed and summarized diagnostic plots, ...)

Motivation: allow on-the-fly scheduling (e.g. determining the next value of lambda based on samples collected from the previous lambda (similar to https://github.com/proteneer/timemachine/pull/442), or based on samples from several previously collected states (using methods prototyped by @mcwitt )).

Introduces misc. opinionated changes in other parts of `rbfe.py` along the way -- some of which I'll revert before requesting review.

PR #949, opened 2023-02-03
Allow different inter and intra molecular charges

jkausrelay:
- Add a `AM1CCCSplitHandler` which allows for different parameters between inter and intra molecular charges
- Note: The implementation required adding **kwargs to the other charge handlers to ignore the parameter specifying to use the intra or inter mol charges
- Update BaseTopology/HostGuestTopology/DualToplogy/DualTopologyMinimization/SingleTopology to use the correct set of charge parameters
- Add unit tests


tests/common.py
```diff
@@ -356,3 +356,18 @@ def params_with_w_coords(w_coords):
...
+        OpenFF 2.0.0 ff,
+        OpenFF 2.0.0 ff with split intramolecular and intermolecular charge terms,
+        OpenFF 2.0.0 ff with all charge terms scaled by 10x,
+        OpenFF 2.0.0 ff with only intermolecular charge terms scaled by 10x.
+    """
+    ff_ref = Forcefield.load_from_file("smirnoff_2_0_0_ccc.py")
+    ff_split = Forcefield.load_from_file("tests/data/smirnoff_2_0_0_ccc_split.py")
+    ff_scaled = Forcefield.load_from_file("tests/data/smirnoff_2_0_0_ccc_scaled.py")
+    ff_inter_scaled = Forcefield.load_from_file("tests/data/smirnoff_2_0_0_ccc_inter_scaled.py")
+    return ff_ref, ff_split, ff_scaled, ff_inter_scaled
```
mcwitt, line 373: nit: worth defining a dataclass for the return value? (seems like it'd be easy to get the order confused)

tests/test_single_topology.py
```diff
@@ -361,6 +362,85 @@ def test_combine_with_host():
...
+        # Compute the expected intermol scaled potential
+        expected_inter_scaled_u = solvent_u_scaled - vacuum_u_scaled + vacuum_u_ref
+
+        # Pad gradients for the solvent
+        vacuum_grad_scaled_padded = np.concatenate([np.zeros(solvent_conf.shape), vacuum_grad_scaled])
+        vacuum_grad_ref_padded = np.concatenate([np.zeros(solvent_conf.shape), vacuum_grad_ref])
+        expected_inter_scaled_grad = solvent_grad_scaled - vacuum_grad_scaled_padded + vacuum_grad_ref_padded
+
+        # They should be equal
+        assert expected_inter_scaled_u == pytest.approx(solvent_u_inter_scaled)
```
mcwitt, line 436: Should this assertion respect the rtol, atol args?

tests/test_topology.py
```diff
@@ -38,3 +46,118 @@ def test_dual_topology_nonbonded_pairlist():
...
+    box_width = 4.0
+    solvent_sys, solvent_conf, solvent_box, solvent_top = build_water_system(box_width, ff_ref.water_ff)
+    solvent_box += np.diag([0.1, 0.1, 0.1])
+    host_bps, host_masses = openmm_deserializer.deserialize_system(solvent_sys, cutoff=1.2)
+
+    with resources.path("timemachine.testsystems.data", "ligands_40.sdf") as path_to_ligand:
+        all_mols = read_sdf(path_to_ligand)
+
+    if ctor == BaseTopology:
+        mol = all_mols[1]
```
mcwitt, line 106: nit: would it be preferable to lookup by mol name here? (maybe not an issue if we're confident ordering is deterministic)

tests/test_topology.py
```diff
@@ -38,3 +46,118 @@ def test_dual_topology_nonbonded_pairlist():
...
+    elif ctor in [DualTopology, DualTopologyMinimization]:
+        mol_a = all_mols[1]
+        mol_b = all_mols[4]
+        ligand_conf = np.concatenate([get_romol_conf(mol_a), get_romol_conf(mol_b)])
+        coords0 = np.concatenate([solvent_conf, ligand_conf])
+        Topology = partial(ctor, mol_a, mol_b)
+    else:
+        raise ValueError(f"Unknown topology class: {ctor}")
+
+    n_lambdas = 3
```
mcwitt, line 119: nit: worth parameterizing/varying this?

jkausrelay: The test needs to run at both endpoints and at least one intermediate. I could increase it but I don't think we need to repeat tests that only change this value.

timemachine/ff/__init__.py
```diff
@@ -114,6 +121,7 @@ def from_handlers(
                 lj_handle = handle
             elif (
                 isinstance(handle, nonbonded.AM1CCCHandler)
+                or isinstance(handle, nonbonded.AM1CCCSplitHandler)
                 or isinstance(handle, nonbonded.AM1BCCHandler)
                 or isinstance(handle, nonbonded.SimpleChargeHandler)
```
maxentile, line 126: Question 1: Is there a convenient way to allow "split" charges but with the other charge models (AM1BCC, SimpleCharge), or a user-specified function that can assign charges to a ligand?

Question 2: Is there any functionality in common between "splitting" charge parameters (this PR) and "splitting" LJ parameters (not this PR)?

maxentile (approved):

timemachine/ff/handlers/nonbonded.py
```diff
@@ -465,6 +471,12 @@ def static_parameterize(mol):
         """
         return oe_assign_charges(mol, "AM1BCCELF10")
 
+    def get_smirks(self, **kwargs):
+        return self.smirks
+
+    def get_params(self, **kwargs):
+        return self.params
```
mcwitt, line 478: If `smirks` and `params` are intended to be private, should they be prefixed with `_`? (Also, `get_smirks` and `get_params` seem unused)

mcwitt:
> Note: The implementation required adding **kwargs to the other charge handlers to ignore the parameter specifying to use the intra or inter mol charges

I might be missing something here, but would it be possible/desirable to split `AM1CCCSplitHandler` into 2 classes for intermolecular and intramolecular respectively? (Seems like this might avoid needing to modify the other handlers with `**kwargs`)

timemachine/ff/__init__.py
jkausrelay: 1) Updated so all of these charge models have a corresponding intramolecular version. It should be simple enough to add additional handlers if needed.
2) The code should follow the same pattern for LJ.

timemachine/ff/handlers/nonbonded.py
jkausrelay: removed these after refactoring.

jkausrelay:
> > Note: The implementation required adding **kwargs to the other charge handlers to ignore the parameter specifying to use the intra or inter mol charges
> 
> I might be missing something here, but would it be possible/desirable to split `AM1CCCSplitHandler` into 2 classes for intermolecular and intramolecular respectively? (Seems like this might avoid needing to modify the other handlers with `**kwargs`)

Modified the PR to split it in the forcefield instead. This means there is a ff.q_handle (for intermolecular terms) and ff.q_handle_intra (for intramolecular terms).

tests/test_single_topology.py
```diff
@@ -361,6 +362,72 @@ def test_combine_with_host():
...
+    core = _get_core_by_mcs(mol_a, mol_b)
+
+    # split forcefield has different parameters for intramol and intermol terms
+    ffs = load_split_forcefields()
+    solvent_sys, solvent_conf, solvent_box, solvent_top = build_water_system(4.0, ffs.ref.water_ff)
+    solvent_bps, _ = openmm_deserializer.deserialize_system(solvent_sys, cutoff=1.2)
+
+    def get_vacuum_solvent_u_grads(ff, lamb):
+        st = SingleTopology(mol_a, mol_b, core, ff)
+        ligand_conf = st.combine_confs(get_romol_conf(mol_a), get_romol_conf(mol_b))  # TODO put lamb here w/Josh's fix
```
jkausrelay, line 380: Fix means: https://github.com/proteneer/timemachine/pull/936/files#diff-0b9bd14a098c7a129a18f14ffb94d84fc8e5c1020d1a55d0e6f53d545c1a4084R899

timemachine/fe/topology.py
```diff
@@ -251,10 +276,9 @@ def parameterize_nonbonded_pairlist(self, ff_q_params, ff_lj_params):
                     q * sf_q,
                     sig,
                     eps * sf_lj,
-                    0.0,  # w offset
+                    0.0,  # w offset for intramolecular term
```
maxentile, line 279: Point of confusion: Meaning of intra molecular w offset. (Does increasing this number make "all intramolecular pairs further away from each other," or something else?)

maxentile (approved):

timemachine/ff/handlers/nonbonded.py
```diff
@@ -353,6 +353,10 @@ class SimpleChargeHandler(NonbondedHandler):
     pass
 
 
+class SimpleChargeIntraHandler(SimpleChargeHandler):
```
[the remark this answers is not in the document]
jkausrelay: The forcefield is stored as a dictionary in the form of {handler_name: (smirks, params)} so the handlers need to have unique names.

timemachine/fe/topology.py
jkausrelay: Yes, it increases the effective distance between pairs.

timemachine/fe/topology.py
```diff
@@ -531,13 +559,34 @@ def parameterize_improper_torsion(self, ff_params):
...
+        params, nb_potential = super().parameterize_nonbonded(
+            ff_q_params, ff_q_params_intra, ff_lj_params, lamb, intramol_params=intramol_params
+        )
         cutoff = nb_potential.get_cutoff()
         params_with_offsets = jnp.asarray(params).at[:, 3].set(lamb * cutoff)
 
         return params_with_offsets, nb_potential
+
+
+def exclude_all_ligand_ligand_ixns(num_host_atoms: int, num_guest_atoms: int) -> Tuple[NDArray, NDArray]:
```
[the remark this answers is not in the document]
jkausrelay: added

timemachine/ff/handlers/nonbonded.py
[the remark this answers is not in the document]
maxentile: Thanks! Makes sense.

I wonder if it would be justifiable later to change the `Forcefield` serialization format to use `hb_handle`, `q_handle`, `intra_q_handle`, etc. as top-level dictionary keys?

`handle_name: (handler_class_name, handler_params_dict)`

e.g.

`{intra_q_handle: (SimpleChargeHandler, {smirks: ..., params: ...}), ...}`

(If we pick a serialization format for `Forcefield` that more closely follows the structure of `Forcefield`, this may avoid the need to reflect `Forcefield`'s top-level keys in the names of classes like `*ChargeHandler` and `*IntraChargeHandler`. I imagine this could also be more flexible if we later wish to explore further elaborations (e.g. splitting L-L, L-P, L-W interactions, or splitting LJ in addition charges). But that's maybe a separate issue -- for the purposes of splitting parameters for L-L charges from L-env charges, the current approach in this PR seems like the smallest code change.)

mcwitt (approved):

timemachine/ff/handlers/nonbonded.py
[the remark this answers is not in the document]
jkausrelay: Yeah, if we start modifying interactions like this we can consider a more flexible serialization format.

PR #950, opened 2023-02-04
Add option to make RBFE memory usage constant in trajectory length

mcwitt:
In the current version, memory usage scales linearly with the number of frames simulated, limiting the maximum possible trajectory length for large systems to a few ns in practice. This PR aims to address the issue with the following changes:

- Introduces `max_buffer_frames` argument to `fe.rbfe.sample` (defaulting to ~1000~ `None`)
- We never store more than `max_buffer_frames` frames in memory; we split longer trajectories into chunks and write to disk
- This is done transparently using `StoredArrays`, which implements the `Sequence[NDArray]` interface but stores data in a temporary directory on disk
- This feature is disabled by default, since it will require changes downstream. A follow-up PR will enable this by default for RBFE: #957

#### TODO
- [x] Update RBFE analysis code to stream saved frames for BAR calculations (not much needed here; `StoredArrays` is a drop-in replacement for list)
- [x] ~Do we want to save all frames in the result? (this might complicate output format & downstream code)~ user is responsible for handling saving of frames; we provide an iterator over frames stored on disk
- [x] Fix unit tests
- [x] Run nightly tests ([successful pipeline run](https://gitlab.com/relaytx/physics-platform/timemachine/-/pipelines/769291768))
- [x] Merge changes from #936 
- [x] ~Compatibility with subprocess/pickling~
   - `StoredArrays` not pickleable; saving results should be handled in application code (example adaptation of `rbfe_edge_list.py` in #957)
   - Adds `StoredArrays.store`, `StoredArrays.load` accepting a `FileClient` to abstract over the storage method (currently a directory of `.npy` files; one file per chunk)

### DIFF
--- a/tests/test_relative_free_energy.py
+++ b/tests/test_relative_free_energy.py
@@ -319,7 +319,7 @@ def test_sample_max_buffer_frames(hif2a_ligand_pair_single_topology_lam0_state, 
     frames_ref, _ = sample(hif2a_ligand_pair_single_topology_lam0_state, protocol, max_buffer_frames=None)
     frames_test, _ = sample(hif2a_ligand_pair_single_topology_lam0_state, protocol, max_buffer_frames=max_buffer_frames)
 
-    assert isinstance(frames_ref, list)
+    assert isinstance(frames_ref, np.ndarray)
     assert isinstance(frames_test, StoredArrays)
     assert len(frames_ref) == len(frames_test)
 
--- a/tests/test_relative_free_energy.py
+++ b/tests/test_relative_free_energy.py
@@ -4,10 +4,12 @@ from importlib import resources
 
 import numpy as np
 import pytest
+from hypothesis import given, seed
+from hypothesis.strategies import integers
 
 from timemachine.constants import DEFAULT_FF
 from timemachine.fe.bar import pair_overlap_from_ukln
-from timemachine.fe.free_energy import HostConfig, SimulationResult, image_frames, sample
+from timemachine.fe.free_energy import HostConfig, SimulationResult, batches, image_frames, sample
 from timemachine.fe.rbfe import estimate_relative_free_energy, run_solvent, run_vacuum
 from timemachine.ff import Forcefield
 from timemachine.md import builders
@@ -279,6 +281,21 @@ def test_pair_overlap_from_ukln():
 
     # overlapping
     assert gaussian_overlap((0, 0.1), (0.5, 0.2)) > 0.1
+
+
+@given(integers(min_value=1))
+@seed(2023)
+def test_batches_of_nothing(batch_size):
+    assert list(batches(0, batch_size)) == []
+
+
+@given(integers(min_value=1, max_value=1000), integers(min_value=1))
+@seed(2023)
+def test_batches(n, batch_size):
+    assert sum(batches(n, batch_size)) == n
+    assert all(batch == batch_size for batch in list(batches(n, batch_size))[:-1])
+    *_, last = batches(n, batch_size)
+    assert 0 < last <= batch_size
 
 
 if __name__ == "__main__":
--- a/tests/test_relative_free_energy.py
+++ b/tests/test_relative_free_energy.py
@@ -7,10 +7,12 @@ import pytest
 from hypothesis import given, seed
 from hypothesis.strategies import integers
 
-from timemachine.constants import DEFAULT_FF
+from timemachine.constants import DEFAULT_FF, DEFAULT_TEMP
 from timemachine.fe.bar import pair_overlap_from_ukln
-from timemachine.fe.free_energy import HostConfig, SimulationResult, batches, image_frames, sample
-from timemachine.fe.rbfe import estimate_relative_free_energy, run_solvent, run_vacuum
+from timemachine.fe.free_energy import HostConfig, SimulationProtocol, SimulationResult, batches, image_frames, sample
+from timemachine.fe.rbfe import estimate_relative_free_energy, run_solvent, run_vacuum, setup_initial_states
+from timemachine.fe.single_topology import SingleTopology
+from timemachine.fe.stored_arrays import StoredArrays
 from timemachine.ff import Forcefield
 from timemachine.md import builders
 from timemachine.md.barostat.utils import compute_box_center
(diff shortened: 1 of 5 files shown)


mcwitt:
Returning to draft status until #936 is merged (since there will likely be conflicts)

tests/test_relative_free_energy.py
```diff
@@ -281,6 +285,48 @@ def u(x):
...
+    *_, last = batches(n, batch_size)
+    assert 0 < last <= batch_size
+
+
+@pytest.fixture(scope="module")
+def hif2a_ligand_pair_single_topology_lam0_state():
+    mol_a, mol_b, core = get_hif2a_ligand_pair_single_topology()
+    forcefield = Forcefield.load_from_file(DEFAULT_FF)
+    st = SingleTopology(mol_a, mol_b, core, forcefield)
+    [state] = setup_initial_states(st, None, DEFAULT_TEMP, [0.0], 2023)
```
badisa, line 308: I find this syntax of setting the variable to a list confusing and hard to read (didn't even know it was possible). Would be nice to make it more explicit.

timemachine/fe/free_energy.py
```diff
@@ -54,38 +56,37 @@ class SimulationResult:
...
+        Boxes to image coordinates into, list of K arrays with shape (3, 3)
 
     Returns
     -------
         imaged_coordinates
     """
-    assert len(frames.shape) == 3, "Must be a 3 dimensional set of frames"
-    assert frames.shape[-1] == 3, "Frame coordinates are not 3D"
-    assert boxes.shape[1:] == (3, 3), "Boxes are not 3x3"
+    assert all(frame.ndim == 2 and frame.shape[-1] == 3 for frame in frames), "All frames must have shape (N, 3)"
```
badisa, line 88: If now every time we do `for x in sequence` if `sequence` is backed by disk, we are going to be incurring a lot more cost. Should we move this assertion to the loop?

timemachine/fe/stored_arrays.py
```diff
@@ -0,0 +1,140 @@
...
+    3
+    >>> len(list(sa._chunks()))
+    2
+    >>> list(sa)
+    [array([1, 2, 3]), array([4, 5, 6]), array([7, 8, 9])]
+    """
+
+    def __init__(self):
+        self._chunk_sizes: List[int] = []
+        self._dir = Path(tempfile.mkdtemp())
```
badisa, line 39: Should use `tempfile.TemporaryDirectory`? That way we don't need the `weakref.finalize`

timemachine/fe/stored_arrays.py
```diff
@@ -0,0 +1,140 @@
...
+        for chunk in self._chunks():
+            for array in chunk:
+                yield array
+
+    def __len__(self) -> int:
+        return sum(self._chunk_sizes)
+
+    @overload
+    def __getitem__(self, key: int) -> NDArray:
+        ...
```
badisa, line 52: Should these raise exceptions rather then return None?

badisa: what exactly are these doing? Seems like `__getitem__` below handles these cases.

timemachine/fe/stored_arrays.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Data is stored in a temporary directory that is cleaned when the `StoredArrays` object is finalized.
+
+    Examples
+    --------
+    >>> sa = StoredArrays()
+    >>> sa.extend([np.array([1, 2, 3]), np.array([4, 5, 6])])
+    >>> len(sa)
+    2
+    >>> len(list(sa._chunks()))
```
badisa, line 26: Should `_chunks` be a non-private function?

timemachine/fe/stored_arrays.py
mcwitt: The `@overload` decorated methods are stubs (purely for type checking; see https://docs.python.org/3/library/typing.html#typing.overload)

timemachine/fe/stored_arrays.py
mcwitt: Hmm, I don't think so -- it's included in the example here for documentation purposes (to give some intuition about how arrays are stored internally). I'd prefer to remove this example if it seems confusing.

badisa: Documenting private functions seem weird to me, would rather just document `_chunks` if the intent is to clarify how things are stored.

timemachine/fe/stored_arrays.py
mcwitt: Good call, that's cleaner (was thinking `TemporaryDirectory` needed to be used as a context manager, but that's not the case). Updated in https://github.com/proteneer/timemachine/pull/950/commits/40b6b1053a19fea968d435b8eb5cefd8d4df8b63

tests/test_relative_free_energy.py
mcwitt: I preferred this method here because it's a little stricter (would fail if the returned list has more than one element), but can use the more familiar method and an extra assertion if preferable

timemachine/fe/free_energy.py
mcwitt: Done in https://github.com/proteneer/timemachine/pull/950/commits/e1d706c82086bca3aeaf1254a812cc3e6168bd3d

tests/test_relative_free_energy.py
badisa: Having declaration be a side effect seems confusing and seems unique to `[]` where as `(x) = "foo"` and `(x,) = "foo"` behave differently.

badisa (approved):

tests/test_relative_free_energy.py
```diff
@@ -166,13 +170,13 @@ def test_steps_per_frames():
     seed = 2022
     frames = 5
     res = run_vacuum(mol_a, mol_b, core, forcefield, None, frames, seed, n_eq_steps=10, steps_per_frame=2, n_windows=2)
-    assert res.frames[0].shape[0] == frames
+    assert len(res.frames[0]) == frames
```
maxentile, line 173: Hmm, to avoid these sorts of changes, I wonder if it make sense to add a `.to_numpy()` method?

(For cases where a `StoredArrays` object can fit entirely in memory, and we would like to interact with it as a single ordinary numpy array)

tests/test_relative_free_energy.py
mcwitt: Changed in https://github.com/proteneer/timemachine/pull/950/commits/39dbf2e5f9a9210325c0be46995cca7a99c613a9

timemachine/fe/stored_arrays.py
mcwitt: Done in https://github.com/proteneer/timemachine/pull/950/commits/d529f3105732308538cbb2d3e6caf4f94d8ce466

tests/test_relative_free_energy.py
mcwitt: Hmm, we could add this, but we'd have to be careful about where we use it to avoid OOM. (Seems like it would be fine here, but I'm not sure I see a reason to prefer `frames.shape[0]` over `len(frames)`?)

mcwitt: (Also, `np.array(frames)` should work)

maxentile: Oh! I overlooked that `np.array(frames)` already worked for this use -- that resolves my comment 👍

maxentile (approved):

jkausrelay (approved):

PR #955, opened 2023-02-08
Add nightly test for rbfe_edge_list

mcwitt:
While working on #950, I realized it would be useful to have a test that exercises running RBFE in a subprocess to catch potential issues with serialization of `SimulationResult`s. This PR

- Adds a (minimal) test for `examples/rbfe_edge_list.py` that checks that the script runs and that frames are accessible in the result pickle
- ~Fixes an issue where `.` in `PYTHONPATH` could be expanded to a directory other than the timemachine source directory (when using `with temporary_working_dir:`)~ don't handle relative paths in `PYTHONPATH` (user should know better :stuck_out_tongue:), update readme example

#### TODO
- [ ] Pass [nightly test pipeline](https://gitlab.com/relaytx/physics-platform/timemachine/-/pipelines/771811950)

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,12 +80,12 @@ make build  # Must have installed dev dependencies for this to work
 
 ### Running Tests
 
-Note that `PYTHONPATH` must be set to include the `timemachine` repo root. To run tests that use `openeye`, ensure that either `OE_LICENSE` or `OE_DIR` are set appropriately.
+To run tests that use `openeye`, ensure that either `OE_LICENSE` or `OE_DIR` are set appropriately.
 
 For example, starting from a clean environment with the openeye license file in `~/.openeye`:
 
 ```shell
-OE_DIR=~/.openeye PYTHONPATH=$PWD:$PYTHONPATH pytest -xsv tests/
+OE_DIR=~/.openeye pytest -xsv tests/
 ```
 
 Note: we currently only support and test on python 3.7, use other versions at your own peril.
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ Note that `PYTHONPATH` must be set to include the `timemachine` repo root. To ru
 For example, starting from a clean environment with the openeye license file in `~/.openeye`:
 
 ```shell
-OE_DIR=~/.openeye PYTHONPATH=. pytest -xsv tests/
+OE_DIR=~/.openeye PYTHONPATH=$PWD:$PYTHONPATH pytest -xsv tests/
 ```
 
 Note: we currently only support and test on python 3.7, use other versions at your own peril.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -85,7 +85,7 @@ Note that `PYTHONPATH` must be set to include the `timemachine` repo root. To ru
 For example, starting from a clean environment with the openeye license file in `~/.openeye`:
 
 ```shell
-OE_DIR=~/.openeye PYTHONPATH=. pytest -xsv tests/
+OE_DIR=~/.openeye PYTHONPATH=$PWD:$PYTHONPATH pytest -xsv tests/
```
badisa, line 88: Thinking about it some more, we should probably remove the reference to `PYTHONPATH`. Users only get `pytest` and other test requirements from doing `pip install .[test]` and not from `pip install requirements`, so it should already be in their path.

mcwitt: Agreed; I think the readme needs some general cleanup. I'm inclined to leave this for now since we mention `make build` as an alternative to `pip install .`, and the user could conceivably install `pytest` separately. Should we open an issue for a general readme cleanup?

mcwitt: Hmm, actually, reading through it again, we do currently bless `pip install .` as the way to install. So I think it makes sense to just remove the `PYTHONPATH` setting as part of this PR.

mcwitt: Removed in https://github.com/proteneer/timemachine/pull/955/commits/0e49d13adba83751f85392c5749efc8ad706ad7d

badisa (approved):

tests/test_examples.py
```diff
@@ -140,3 +142,82 @@ def test_smc_freesolv(smc_free_solv_path):
...
+
+        with temporary_working_dir() as temp_dir:
+            with (Path(temp_dir) / "edges.csv").open("w") as fp:
+                fp.writelines(edges_rows_sample)
+                fp.flush()
+                path = run(fp.name, temp_dir)
+            yield path, base_config, edges
+
+
+def test_rbfe_edge_list_hif2a(rbfe_edge_list_hif2a_path):
```
jkausrelay, line 182: How long does this test take to run?

mcwitt: On the order of 10 minutes with using 1 GPU on my local machine.

jkausrelay (approved):

tests/test_examples.py
jkausrelay: Ok. If you want to make it quicker, you can add n_eq_steps/n_windows as parameters to the script. As a nightly it doesn't matter as much, but this might be quick enough for a unit test.

mcwitt: I tried parameterizing `n_eq_steps` and setting to 2 (I had already been limiting `n_windows` to 3, but had initially missed committing that change). The timing is still a bit long for a unit test (~3 minutes), so I'll probably leave this as a nightly for now unless there are any objections.

PR #956, opened 2023-02-08
Fixes for NB potentials

jkausrelay:
- For molecules with < 4 atoms, the split NB code would fail
- Update the topology classes to handle this corner case and add to tests
- Also update VacuumSystem to be able to run on molecules < 4 atoms
- Also fix DualTopology.parameterize_nonbonded_pairlist() - was using beta instead of cutoff in NonbondedPairListPrecomputed
-- Unclear what the full scope of this issue is
-- potentially affects minimizer.minimize_host_4d and align.align_mols_by_core

### DIFF
--- a/tests/test_topology.py
+++ b/tests/test_topology.py
@@ -5,6 +5,8 @@ import jax.numpy as jnp
 import numpy as np
 import pytest
 from common import load_split_forcefields
+from rdkit import Chem
+from rdkit.Chem import AllChem
 
 from timemachine.fe import topology
 from timemachine.fe.topology import BaseTopology, DualTopology, DualTopologyMinimization
@@ -74,7 +76,8 @@ def parameterize_nonbonded_full(
 
 @pytest.mark.parametrize("precision, rtol, atol", [(np.float64, 1e-8, 1e-8), (np.float32, 1e-4, 5e-4)])
 @pytest.mark.parametrize("ctor", [BaseTopology, DualTopology, DualTopologyMinimization])
-def test_host_guest_nonbonded(ctor, precision, rtol, atol):
+@pytest.mark.parametrize("use_tiny_mol", [True, False])
+def test_host_guest_nonbonded(ctor, precision, rtol, atol, use_tiny_mol):
     def compute_ref_grad_u(ff: Forcefield, precision, x0, lamb):
         # Use the original code to compute the nb grads and potential
         bt = Topology(ff)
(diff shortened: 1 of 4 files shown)


tests/test_topology.py
```diff
@@ -74,7 +76,8 @@ def parameterize_nonbonded_full(
 
 @pytest.mark.parametrize("precision, rtol, atol", [(np.float64, 1e-8, 1e-8), (np.float32, 1e-4, 5e-4)])
 @pytest.mark.parametrize("ctor", [BaseTopology, DualTopology, DualTopologyMinimization])
-def test_host_guest_nonbonded(ctor, precision, rtol, atol):
+@pytest.mark.parametrize("use_tiny_mol", [True, False])
```
maxentile, line 79: Coverage: Would there be benefit to testing other small corner cases, 1, 2, 3, 4, 4+ atoms? (so the molecule may have no bonds, angles, torsions?)

Naming / comment: "mol with no intramolecular NB terms" to indicate the challenge presented by this input

timemachine/fe/topology.py
```diff
@@ -521,7 +531,7 @@ def parameterize_nonbonded_pairlist(self, ff_q_params, ff_q_params_intra, ff_lj_
         assert pairlist_a.get_cutoff() == pairlist_b.get_cutoff()
 
         return params, potentials.NonbondedPairListPrecomputed(
-            inclusion_idxs, pairlist_a.get_beta(), pairlist_a.get_beta()
+            inclusion_idxs, pairlist_a.get_beta(), pairlist_a.get_cutoff()
```
maxentile, line 534: Oof...

Question: Is dual topology in timemachine deprecated? (since https://github.com/proteneer/timemachine/pull/869 ?)

maxentile (approved):

timemachine/fe/system.py
```diff
@@ -176,12 +176,20 @@ def U_fn(x):
...
+        def append_if_params(pot):
+            if len(pot.params) > 0:
+                all_U_fns.append(pot)
+
+        append_if_params(self.bond)
+        append_if_params(self.angle)
+        append_if_params(self.torsion)
+        append_if_params(self.nonbonded)
+
+        return all_U_fns
```
mcwitt, line 192: style nit: consider list comprehension:
```suggestion
        return [p for p in [self.bond, self.angle, self.torsion, self.nonbonded] if len(p.params) > 0]
```

timemachine/fe/topology.py
mcwitt: Good catch!

mcwitt (approved):

timemachine/fe/topology.py
jkausrelay: It's still used in rbfe for minimization. It seems to produce the same minimized conformations though for the system I tested.

maxentile: Ahh, thanks! Impact is likely small if it only affects minimization.

(Other factors: I think default `beta = 2.0`, default `cutoff = 1.2`, so this had the the effect of increasing the cutoff relative to its default. The behavior of "lambda" in 4D decoupling does depend on the value of `cutoff`, so fixing this might make the lambda schedule used for `minimize_host_4d` more efficient / more similar to its assumptions...)

tests/test_topology.py
jkausrelay: > Coverage: Would there be benefit to testing other small corner cases, 1, 2, 3, 4, 4+ atoms? (so the molecule may have no bonds, angles, torsions?)

Molecules smaller than 3 atoms have other issues (atom mapping can fail, AM1BCC seems to fail). I'll file a case for this but it's not as high a priority.

> Naming / comment: "mol with no intramolecular NB terms" to indicate the challenge presented by this input

Added comment to tests

PR #958, opened 2023-02-10
Rename SimulationProtocol -> MDParams

mcwitt:
Rationale:
- "Protocol" is an overloaded term and we generally use it to refer to the spacing of lambda windows / set of intermediate states
- As we move to introduce new methods of generating intermediate states (e.g. bisection), we might want to reserve `protocol` to mean the latter to avoid confusion

### DIFF
--- a/tests/test_fe_absolute_hydration.py
+++ b/tests/test_fe_absolute_hydration.py
@@ -26,8 +26,8 @@ def test_run_solvent():
     assert len(res.frames[-1]) == n_frames
     assert len(res.boxes[0]) == n_frames
     assert len(res.boxes[-1]) == n_frames
-    assert res.protocol.n_frames == n_frames
-    assert res.protocol.n_eq_steps == n_eq_steps
+    assert res.md_params.n_frames == n_frames
+    assert res.md_params.n_eq_steps == n_eq_steps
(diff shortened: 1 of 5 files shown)


jkausrelay (approved):

maxentile (approved):
Renaming from `SimulationProtocol` seems good, since `protocol` is overloaded and is most closely associated with `lambda_schedule`.

Hesitation: I would expect something named `MDParams` to contain settings like `timestep`, `barostat_frequency`, `enable_hmr`, etc. -- do we want to pick a name for that bundle of settings as well, distinct from the name for the current bundle of settings `(n_frames, n_eq_steps, steps_per_frame)` ?

An ideal pair of names doesn't immediately come to mind...

maxentile:
(Above comment is non-blocking -- I think the proposal is an improvement, and I can't quickly think of a further improvement...)

mcwitt:
> Hesitation: I would expect something named MDParams to contain settings like timestep, barostat_frequency, enable_hmr, etc.

Good point, I don't immediately have a good solution to this either, other than maybe to consider a more generic name in place of `MDParams` (`SimulationParams`, `SimulationConfig`, etc.). My main motivation was to remove the word "protocol".

PR #964, opened 2023-03-01
Support slicing StoredArrays

mcwitt:
Implements `__getitem__(self, slice)` for `StoredArrays`. This will be useful e.g. for subsampling frames in downstream code once #959 or #957 are merged.

### DIFF
--- a/timemachine/fe/stored_arrays.py
+++ b/timemachine/fe/stored_arrays.py
@@ -1,5 +1,6 @@
 import io
 import tempfile
+from functools import lru_cache, partial
 from itertools import count
 from pathlib import Path
 from typing import Collection, Iterator, List, Sequence, overload
@@ -32,6 +33,9 @@ class StoredArrays(Sequence[NDArray]):
         self._chunk_sizes: List[int] = []
         self._dir = tempfile.TemporaryDirectory()
 
+        # Each instance caches a single chunk. This simplifies the efficient implementation of slicing.
+        self._load_chunk = lru_cache(maxsize=1)(partial(load_chunk, self._path()))
+
     def __iter__(self) -> Iterator[NDArray]:
         for chunk in self._chunks():
             for array in chunk:
@@ -56,7 +60,7 @@ class StoredArrays(Sequence[NDArray]):
                 key += len(self)
             for idx, size in enumerate(self._chunk_sizes):
                 if key < size:
-                    return np.load(self._get_chunk_path(idx))[key]
+                    return self._load_chunk(idx)[key]
                 key -= size
             assert False, "should not get here"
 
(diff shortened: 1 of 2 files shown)


badisa (approved):

jkausrelay (approved):

mcwitt:
I think there is one optimization worth making: in the current implementation, the extreme case of slicing with a stride of one (e.g. `stored_arrays[:]`) loads each `.npy` file many times (equal to the number of elements in the array). Added a failing test to demonstrate this in https://github.com/proteneer/timemachine/pull/964/commits/dee102dc580a48a9510e5799329b4e6fa6c961f5.

I think a simple solution would be to use `lru_cache(maxsize=1)` to cache the last-read chunk, so traversing in order shouldn't result in multiple reads of the same file regardless of the stride. Will make this modification in an upcoming commit.

timemachine/fe/stored_arrays.py
```diff
@@ -78,13 +82,13 @@ def __eq__(self, other) -> bool:
             np.array_equal(a, b, equal_nan=True) for a, b in zip(self, other)
         )
 
-    def _chunks(self) -> Iterator[List[NDArray]]:
+    def _chunks(self) -> Iterator[NDArray]:
```
mcwitt, line 85: this type was incorrect before

badisa:
> I think there is one optimization worth making: in the current implementation, the extreme case of slicing with a stride of one (e.g. `stored_arrays[:]`) loads each `.npy` file many times (equal to the number of elements in the array). Added a failing test to demonstrate this in [dee102d](https://github.com/proteneer/timemachine/commit/dee102dc580a48a9510e5799329b4e6fa6c961f5).
> 
> I think a simple solution would be to use `lru_cache(maxsize=1)` to cache the last-read chunk, so traversing in order shouldn't result in multiple reads of the same file regardless of the stride. Will make this modification in an upcoming commit.

Is there a way to avoid this? We currently have O(1) memory usage here, and adding an lru_cache switches it to O(N) for a case that seems like we can typically avoid in practice.

badisa:
It seems like it should be possible to group the indices produced by the slice by their relevant chunk and doing that rather then using an LRU cache. That way we still get one read per chunk, but don't add memory overhead

mcwitt:
> We currently have O(1) memory usage here, and adding an lru_cache switches it to O(N)

(Note that `N` here is the size of a single chunk, and most operations defined here already do use O(N) memory. To be clear, this would add an O(N) overhead for the lifetime of the object.)

> It seems like it should be possible to group the indices produced by the slice by their relevant chunk and doing that rather then using an LRU cache

This is possible, but implementing this correctly seems tricky (e.g. to support the slice interface need to handle arbitrary combinations possibly missing or negative start, end, and step).

One option would be to remove the slice capability and add a limited interface that only supports subsampling via changing the stride.

Another (maybe preferable?) option is to move the cache into the scope of the `__getitem__`. This seems like the best of both worlds.

badisa:
> To be clear, this would add an O(N) overhead for the lifetime of the object.)

Sorry that was imprecise, yes its that it is for the lifetime of the object.

> One option would be to remove the slice capability and add a limited interface that only supports subsampling via changing the stride.

I would probably advocate for this, as it requires more explicit usage of the class. And the current implementation of the slicing in the case of `stored_arrays[:]` will store a single file per frame rather then the original of a thousand frames per file, which has its own potential pain points (in the typical case calling sample). I don't know how much value there is in having StoredArray be used generically as a `Sequence` when we have pretty specific use cases and will need to handle its usage differently already.

mcwitt:
> And the current implementation of the slicing in the case of stored_arrays[:] will store a single file per frame rather then the original of a thousand frames per file

This is true. I'm not sure of a good way around this with the current interface. In general, it seems like supporting slicing with the current interface leads to suboptimal characteristics. Agreed that it might make more sense to deal with e.g. subsampling explicitly in downstream code.

I'm now thinking that this PR was a bit premature. Will close for now and potentially resurrect later if I can think of a better approach.

PR #965, opened 2023-03-02
Draft: Update angle k to better match original angle fxnl form

jkausrelay:
- Update smirnoff_converter.py to rescale the angle force constants
- This better approximates the original angle potential when using the cos angle form
- Add smirnoff_X_Y_Z_ccc_cos_angle.py forcefields
- Add unit test using propyne
- Change the default forcefield to use the cos_angle version


timemachine/ff/smirnoff_converter.py
```diff
@@ -84,6 +90,106 @@ def parse_quantity(number_string):
...
+
+    Based on jfass's optimization code.
+    """
+    # Vacuum simulation of propyne is unstable
+    # if the angle k is above 12k, set below for some buffer.
+    MAX_ANGLE_K = 10000.0  # (kJ/mol)/rad**2
+    x_grid = np.linspace(-1, +1, 10000, dtype=np.float64)
+    loss_fxn = make_cos_angle_loss_fxn(x_grid, k, a0)
+    optimize_result = minimize_scalar(loss_fxn, method="bounded", bounds=(0, MAX_ANGLE_K))
+    return float(optimize_result.x)
```
maxentile, line 190: Question: How should this change be validated?

In the version of this script I wrote on GitLab last year, I had just observed that the JS divergence on this 1D surrogate was highest for equilibrium angles near pi

image

and observed that optimizing this loss increased many of the angle force constants

image

but (1) some of the minimizations NaN'd (I think this is addressed in the current PR by the addition of bounds), and (2) I did not know how to validate the changes.

Other questions:
* Would it make sense to apply this conversion only to angle parameters where `theta_0` is near `pi`? Currently applied to all angle parameters.
* Would it make sense to add a penalty that keeps the adjusted parameter close to its original value? Currently no penalty.
* Any function of the 1D profiles would be equally convenient to optimize here -- JS divergence was picked intuitively. Is JS divergence a good reflection of our intent in adjusting these parameters?

maxentile: Thanks for walking me through the validation offline!

(It's a compelling large improvement, esp. at theta_0 = pi, and doesn't seem to worsen any other cases. This looks good to me to be merged as-is. However, was surprised that there's still a difference between cos_angles=True and cos_angles=False in some cases where theta_0 < pi. Will postpone further review while you're looking into other possible approaches for those cases, possibly including an alternative that doesn't require converting the parameters.)

jkausrelay:
Close for now until we have a better understanding of the impact of this issue

PR #966, opened 2023-03-03
Fix performance regression in single topology conformation optimization

mcwitt:
#959 inadvertently changed the behavior of `setup_initial_states` to rerun host minimization for each lambda window. Host optimization is independent of lambda and should be run only once.

This regression was breaking nightly tests on master:

- Last good run on master (2h 41m): https://gitlab.com/relaytx/physics-platform/timemachine/-/pipelines/793714466
- Failed run on current master (timed out at 6h): https://gitlab.com/relaytx/physics-platform/timemachine/-/pipelines/794987570
- Successful run on this branch (2h 51m): https://gitlab.com/relaytx/physics-platform/timemachine/-/pipelines/795620590

The time is increased by ~10 minutes relative to last passing, but this is expected since we run RBFE tests with bisection in addition to sequentially as of #959.

### DIFF
--- a/timemachine/fe/rbfe.py
+++ b/timemachine/fe/rbfe.py
@@ -189,7 +189,7 @@ def setup_initial_states(
         A single topology object
 
     host: Host or None
-        Pre-optimized host configuration. If None, then a vacuum state will be setup.
+        Pre-optimized host configuration, generated using `setup_optimized_host`. If None, return vacuum states.
 
     temperature: float
         Temperature to run the simulation at.
--- a/timemachine/fe/rbfe.py
+++ b/timemachine/fe/rbfe.py
@@ -1,5 +1,6 @@
 import pickle
 import traceback
+from dataclasses import dataclass
 from functools import partial
 from typing import Any, Dict, Iterable, List, NamedTuple, Optional, Sequence, Tuple, Union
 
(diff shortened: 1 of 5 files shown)


timemachine/fe/rbfe.py
```diff
@@ -174,8 +188,8 @@ def setup_initial_states(
     st: SingleTopology
         A single topology object
 
-    host_config: HostConfig or None
-        Configurations of the host. If None, then a vacuum state will be setup.
```
jkausrelay, line 178: nit: Use `setup_optimized_host` to generate the host.

jkausrelay (approved):

timemachine/fe/rbfe.py
mcwitt: Clarified in https://github.com/proteneer/timemachine/pull/966/commits/2ea7cabb92785738d8e6199b94d1f8545f271a5e

timemachine/fe/rbfe.py
```diff
@@ -98,10 +99,18 @@ def assert_all_states_have_same_masses(initial_states: List[InitialState]):
     np.testing.assert_array_almost_equal(deviation_among_windows, 0, err_msg="masses assumed constant w.r.t. lambda")
 
 
+@dataclass
+class Host:
+    system: VacuumSystem
+    masses: List[float]
```
maxentile, line 105: [nonblocking, unrelated to fixing test]

Possibly rename to `physical_masses`, or have 2 fields `physical_masses` and `hmr_masses`? (Current usages look correct, but good to preempt accidentally applying HMR 0 or 2 times.)

maxentile (approved):
Good catch -- fix looks good!

timemachine/fe/rbfe.py
mcwitt: Ah, I thought we currently disambiguated by prefixing HMR masses with `hmr_`. (I'm not immediately seeing where this field is used for HMR masses). Is the idea just to be extra explicit?

maxentile: Sorry, this is tangential -- will double-check this separately!

>Ah, I thought we currently disambiguated by prefixing HMR masses with hmr_. (I'm not immediately seeing where this field is used for HMR masses)

Good point -- the current convention `masses` vs `hmr_masses` is unambiguous already!

mcwitt: Renamed in https://github.com/proteneer/timemachine/pull/966/commits/271007e6c96ba80734743eda7d2a96e1d1309bfe

PR #967, opened 2023-03-03
Fixes inconcistency in ref nonbonded

badisa:
This was not an issue due to https://github.com/proteneer/timemachine/blob/8f97ebcf11b82bd315444f581e0aa6fc327f09db/tests/common.py#L286 which will take w=`1.2` and convert it to `1.20000005`.

If you remove the conversion to 32 and then to 64 some tests failed, so as a first pass I didn't remove it. Though we might want to update tests with new tolerances to pass, as in practice we don't do this conversion when running the C++ potentials.

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -62,7 +62,7 @@ def prepare_system_params(x: NDArray, cutoff: float, sigma_scale: float = 5.0) -
 
     Returns
     -------
-    (N, 3) np.ndarray containing charges, sigmas and epsilons respectively.
+    (N, 4) np.ndarray containing charges, sigmas, epsilons and w coordinates respectively.
     """
     assert x.ndim == 2
     N = x.shape[0]
--- a/tests/common.py
+++ b/tests/common.py
@@ -307,11 +307,11 @@ class GradientTest(unittest.TestCase):
                     x, params, box, compute_du_dx, compute_du_dp, compute_u
                 )
 
-            np.testing.assert_array_equal(test_du_dx, test_du_dx_2)
-            np.testing.assert_array_equal(test_u, test_u_2)
+                np.testing.assert_array_equal(test_du_dx, test_du_dx_2)
+                np.testing.assert_array_equal(test_u, test_u_2)
 
-            if isinstance(test_potential, potentials.Nonbonded):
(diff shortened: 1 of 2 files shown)


timemachine/potentials/nonbonded.py
```diff
@@ -235,7 +235,7 @@ def nonbonded(
     # funny enough lim_{x->0} erfc(x)/x = 0
     eij_charge = jnp.where(keep_mask, qij * erfc(beta * dij) * inv_dij, 0)  # zero out diagonals
     if cutoff is not None:
-        eij_charge = jnp.where(dij > cutoff, 0, eij_charge)
+        eij_charge = jnp.where(dij < cutoff, eij_charge, 0)
```
badisa, line 238: This is the change in question. Per https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/kernels/k_nonbonded.cuh#L367-L369 this should be strictly less than. 

Other changes are just some minor things I noticed while investigated the nonbonded potentials.

maxentile:
Good catch!

Pair potential should be close to 0 at distance == cutoff, but good to fix.

(Btw, looks like > cutoff vs. >= cutoff may also need to be slightly adjusted in the `nonbonded_block` function https://github.com/proteneer/timemachine/blob/8f97ebcf11b82bd315444f581e0aa6fc327f09db/timemachine/potentials/nonbonded.py#L113 )

tests/common.py
```diff
@@ -307,11 +307,11 @@ def compare_forces(
...
 
-            np.testing.assert_array_equal(test_du_dx, test_du_dx_2)
-            np.testing.assert_array_equal(test_u, test_u_2)
+                np.testing.assert_array_equal(test_du_dx, test_du_dx_2)
+                np.testing.assert_array_equal(test_u, test_u_2)
 
-            if isinstance(test_potential, potentials.Nonbonded):
-                np.testing.assert_array_equal(test_du_dp, test_du_dp_2)
+                if isinstance(test_potential, potentials.Nonbonded):
+                    np.testing.assert_array_equal(test_du_dp, test_du_dp_2)
```
mcwitt, line 314: :+1:

mcwitt (approved):

maxentile (approved):
Good catch!

PR #969, opened 2023-03-06
[WIP] Update reaction field

maxentile:
Draft: Consider replacing `erfc(2 * d_ij)` reaction field with the reaction field from:

[Kubincová, Riniker, Hünenberger, 2020] "Reaction-field electrostatics in molecular dynamics simulations: development of a conservative scheme compatible with an atomic cutoff" https://pubs.rsc.org/en/content/articlelanding/2020/CP/D0CP03835K 

influenced by suggestions from @jchodera , a jax implementation from @dominicrufa , and discussion with @jkausrelay .

Completed:
- [x] LBF-compatible Jax implementation -- Jax implementation compatible with linear basis function trick for fast reweighting (increases storage requirement for ligand charge reweighting from 1 to 5 numbers per ligand atom compared with https://github.com/proteneer/timemachine/pull/685 )
- [x] Draft custom ops implementation -- MD doesn't blow up, but this is otherwise untested. (Also, hard-codes constants and needs refinement.)
- [x] Preliminary performance benchmark -- Force computation requires fewer arithmetic operations than erfc reaction field, and appears to give an overall ~5-7% speed increase in `tests/test_benchmark.py` compared with master. (Compare [current](https://github.com/proteneer/timemachine/blob/1d5b9f00d4a75662d9e459325cb06006a4ff5be2/timemachine/cpp/src/kernels/nonbonded_common.cuh#L33-L43) and [proposed](https://github.com/proteneer/timemachine/blob/1d5b9f00d4a75662d9e459325cb06006a4ff5be2/timemachine/cpp/src/kernels/nonbonded_common.cuh#L79-L93).)

To-do:
- [ ] Update parameter derivatives
- [ ] Update tests
- [ ] Validate on FreeSolv
- [ ] Decide whether to add a toggle between the two reaction field treatments, or to just switch over to a different reaction field
- [ ] Expose global parameters
- [ ] Document


maxentile:
Opened this prematurely, appears buggy -- will re-open when closer to review!

PR #971, opened 2023-03-07
Add utility to view atom mapping in 3D

mcwitt:
Adds `view_atom_mapping_3d` to `timemachine.fe.utils`. This is similar to the existing `plot_atom_mapping_grid`, but instead of producing static images at fixed rotations, this 
displays a rotatable 3D view of both molecules, using [py3Dmol](https://pypi.org/project/py3Dmol/). When an atom mapping is passed into the `core` argument, this displays a second row of rotatable views where the atoms have been color-coded according to the atom mapping. Dummy atoms and bonds are colored white.

### Example from hif2a
image


timemachine/fe/utils.py
```diff
@@ -225,6 +227,80 @@ def plot_atom_mapping_grid(mol_a, mol_b, core, num_rotations=5):
...
+
+    Returns
+    -------
+    py3Dmol.view
+    """
+
+    try:
+        import py3Dmol
+    except ImportError:
+        raise ValueError("requires py3Dmol to be installed")
```
maxentile, line 267: nit: ValueError -> ImportError?

maxentile (approved):
Nice! Rotating with the 3D widget is pretty intuitive (relative to multiple static views).

timemachine/fe/utils.py
mcwitt: I'm unclear on a good convention for handling missing optional dependencies, but reraising an `ImportError` didn't seem quite right to me. Changed to raise a `RuntimeError` chained with the original `ImportError` in https://github.com/proteneer/timemachine/pull/971/commits/0df7735f6635814d0c8ae5af879c7596e925f240. Does this seem OK to you?

maxentile: Looks good!

(Any of these options seems fine to me: ValueError, ImportError, or RuntimeError. I don't know of a convention. Elsewhere in timemachine, import errors from missing optional dependencies are not explicitly handled, I think -- see e.g. https://github.com/proteneer/timemachine/blob/f70586d95af0231b8a663749c268e440457f0d8c/timemachine/ff/handlers/nonbonded.py#L32-L33  . In any case, the required action to fix the error will be clear!)

PR #974, opened 2023-03-10
Avoid recomputation of pair BAR results in simulations with bisection

mcwitt:
Longer-term fix for the issue patched in #973. This adds back the computation of intermediate results, storing the results of BAR computations performed in the bisection phase to avoid recomputing them at the end.

* Performance improvements:
  * Caches BAR computations done during the bisection phase, avoiding the need for recomputation to construct the final `SimulationResult`. This improves the current $O(L^2)$ computation of intermediate results to $O(L)$, and further removes a constant factor of 2, since estimates computed during bisection are reused. ($L$ is number of lambda windows.)
  * Switches to using the $\Delta G$ error estimate from `pymbar.BAR` for componentwise analysis (we still bootstrap overall errors)
* Cleanups:
  * Renames/changes `estimate_free_energy_pair_bar` -> `estimate_free_energy_bar` to compute the result for a single pair of states. This simplifies the implementation of caching.
  * Removes duplicated code for computing BAR estimates given a $u_{kln}$ matrix
  * Renames `IntermediateResult` -> `PairBarResult`, reuses for final result
  * Uses consistent naming convention for "BAR" vs. "Pair BAR":
    *  "BAR" refers to BAR on a single pair of states
    * "Pair BAR" refers to estimator that accepts a list of K states and applies BAR to the K-1 adjacent pairs

Profile visualizations for hif2a 338 -> 165 complex leg, before and after changes:
<details>
<h3>Profile before #973</h3>
NOTE: absolute times not comparable with the other results (this was run locally, the other 2 on AWS)

image

<h3>Profile before this PR</h3>

image

<h3>Profile after this PR</h3>
Comparable/slightly better than before (but we are now storing intermediate results)

image
</details>

### DIFF
--- a/timemachine/fe/free_energy.py
+++ b/timemachine/fe/free_energy.py
@@ -59,12 +59,12 @@ class InitialState:
 
 @dataclass
 class PairBarResult:
-    all_dGs: List[float]  # L - 1
-    all_errs: List[float]  # L - 1
-    dG_errs_by_lambda_by_component: NDArray  # (len(U_names), L - 1)
-    overlaps_by_lambda: List[float]  # L - 1
-    overlaps_by_lambda_by_component: NDArray  # (len(U_names), L - 1)
-    u_kln_by_lambda_by_component: NDArray
+    dG: float
+    dG_err: float
+    dG_err_by_component: NDArray  # (len(U_names),)
+    overlap: float
+    overlap_by_component: NDArray  # (len(U_names),)
+    u_kln_by_component: NDArray  # (len(U_names), 2, 2, N)
 
 
 @dataclass
@@ -77,13 +77,13 @@ class PairBarPlots:
 @dataclass
 class IntermediateResult:
     initial_states: List[InitialState]
-    result: PairBarResult
+    pair_bar_results: List[PairBarResult]
 
 
 @dataclass
 class SimulationResult:
     initial_states: List[InitialState]
-    result: PairBarResult
+    pair_bar_results: List[PairBarResult]
     plots: PairBarPlots
     frames: List[NDArray]  # (len(keep_idxs), n_frames, N, 3)
     boxes: List[NDArray]
@@ -343,7 +343,7 @@ def sample(initial_state: InitialState, md_params: MDParams, max_buffer_frames: 
     return all_coords, all_boxes
 
 
-def estimate_free_energy_pair_bar(u_kln_by_component_by_lambda: NDArray, temperature: float) -> PairBarResult:
+def estimate_free_energy_pair_bar(u_kln_by_component_by_lambda: NDArray, temperature: float) -> List[PairBarResult]:
     """
     Estimate free energies given pre-generated samples. This implements the pair-BAR method, where
     windows assumed to be ordered with good overlap, with the final free energy being a sum
@@ -368,9 +368,8 @@ def estimate_free_energy_pair_bar(u_kln_by_component_by_lambda: NDArray, tempera
     kBT = BOLTZ * temperature
     beta = 1 / kBT
 
-    all_dGs = []
-    all_errs = []
-    for lamb_idx, u_kln_by_component in enumerate(u_kln_by_component_by_lambda):
+    results = []
+    for u_kln_by_component in u_kln_by_component_by_lambda:
         # pair BAR
         u_kln = u_kln_by_component.sum(0)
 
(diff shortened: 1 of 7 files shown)


timemachine/fe/free_energy.py
```diff
@@ -379,9 +380,6 @@ def estimate_free_energy_pair_bar(
         df, df_err = bar_with_bootstrapped_uncertainty(w_fwd, w_rev)  # reduced units
         dG, dG_err = df / beta, df_err / beta  # kJ/mol
 
-        message = f"{prefix} BAR: lambda {lamb_idx} -> {lamb_idx + 1} dG: {dG:.3f} +- {dG_err:.3f} kJ/mol"
-        print(message, flush=True)
```
mcwitt, line 383: Thinking to remove this to simplify the code; it doesn't seem very useful in the current state, since diagnostics are only printed in the final analysis loop after the simulations have run (for both sequential and bisection methods).

timemachine/fe/free_energy.py
```diff
@@ -74,15 +74,21 @@ class PairBarPlots:
...
+class IntermediateResult:
+    initial_states: List[InitialState]
+    pair_bar_results: List[PairBarResult]
+
+
 @dataclass
 class SimulationResult:
     initial_states: List[InitialState]
-    result: PairBarResult
+    pair_bar_results: List[PairBarResult]
```
mcwitt, line 86: This change to make `PairBarResult` represent the result for a single pair of states simplifies the implementation of caching. If we want to avoid downstream changes, we could introduce a new internal type and avoid modifying this.

mcwitt, line 86: This change to make `PairBarResult` represent the result for a single pair of states simplifies the implementation of caching. If we want to avoid downstream changes, we could introduce a new internal type and avoid modifying this.

maxentile: This change make sense to me.
-------
Nit: Noting a small name shift / collision, since current state is a bit confusing:

Current:
* "BAR" -- estimator that accepts 2 states
* "Pair BAR" -- estimator that accepts a list of K states, and applies BAR to the K-1 adjacent pairs
* "MBAR" -- estimator that generalizes BAR to K states

Proposed:
* "Pair BAR" -- synonymous with "BAR"

(not sure if there's a less ambiguous name for what we're currently calling "pair BAR", to avoid collision with "BAR on a single pair"...)

mcwitt: I think that makes sense. I think the convention in this PR (i.e., calling the result for a pair of states `PairBarResult`, instead of the latter referring to $L-1$ adjacent pairs) matches the proposed; is that correct?

maxentile: >I think that makes sense. I think the convention in this PR ... matches the proposed; is that correct?

Correct.

(Sorry for unclarity -- not advocating for any change relative to your PR, just observing a name shift between "current" (master) and "proposed" (this PR).)

mcwitt: > not sure if there's a less ambiguous name for what we're currently calling "pair BAR", to avoid collision with "BAR on a single pair"...

Ah, I think I might have misunderstood this comment before (I assumed you were proposing the convention "Pair BAR" = "BAR", rather than noting that this seems to be implied by the renaming in this PR).

I'm now leaning toward keeping the current convention of "Pair BAR" referring to the estimator that accepts a list of K states, and applies BAR to the K-1 adjacent pairs, and renaming the type representing the result for a single pair to just `BARResult` or similar.

mcwitt: Updated the naming convention in https://github.com/proteneer/timemachine/pull/974/commits/ebcfe5d0f40c9e6460eb7de50548b812d01d874a:

* Rename IntermediateResult -> PairBarResult
* Use "pair BAR" to refer to estimator that accepts a list of K states, and applies BAR to the K-1 adjacent pairs
* Use "BAR" to refer to BAR on a single pair of states

timemachine/fe/free_energy.py
```diff
@@ -75,14 +75,19 @@ class PairBarPlots:
 
 
 @dataclass
-class SimulationResult:
+class PairBarResult:
     initial_states: List[InitialState]
-    result: PairBarResult
+    bar_results: List[BarResult]
```
maxentile, line 80: Nit: shape documentation: would it be valid / helpful to
* annotate the lengths `len(initial_states) == K`, `len(bar_results) == K - 1` in line comments
* possibly assert `len(initial_states) == len(bar_results) + 1` in a `__post_init__`?

maxentile (approved):
Good fix, and nice changes all around!

timemachine/fe/free_energy.py
mcwitt: Added in https://github.com/proteneer/timemachine/pull/974/commits/a767963466b0e11820af8a60649dab0acaef8e78

badisa (approved):

timemachine/fe/rbfe.py
```diff
@@ -793,10 +787,10 @@ def run_edge_and_save_results(
     pkl_obj = (mol_a, mol_b, edge.metadata, core, solvent_res, solvent_top, complex_res, complex_top)
     file_client.store(path, pickle.dumps(pkl_obj))
 
-    solvent_ddg = np.sum(solvent_res.result.all_dGs)
-    solvent_ddg_err = np.linalg.norm(solvent_res.result.all_errs)
-    complex_ddg = np.sum(complex_res.result.all_dGs)
-    complex_ddg_err = np.linalg.norm(complex_res.result.all_errs)
+    solvent_ddg = sum(r.dG for r in solvent_res.final_result.bar_results)
```
jkausrelay, line 790: Could `PairBarResult` add properties to return all_dGs, all_errs, etc to make it simpler to read the values by lambda?

mcwitt: Added in https://github.com/proteneer/timemachine/pull/974/commits/f4a479e837d37c1b41a1d1c7efbac88809c4824f

jkausrelay (approved):

PR #975, opened 2023-03-13
Modifies Local MD to use Ixn Group

badisa:
* 1.5x speed of global, .6x if `multiple_steps_local` called repeatedly
* Moves local md logic to separate file


timemachine/cpp/src/context.hpp
```diff
@@ -64,7 +64,6 @@ class Context {
     double *d_v_t_;   // velocities
     double *d_box_t_; // box vectors
 
-    unsigned long long *d_du_dx_t_;  // du/dx [N,3]
```
badisa, line 67: This is no longer used by the context

timemachine/cpp/src/local_md_utils.hpp
```diff
@@ -0,0 +1,150 @@
...
+            new NonbondedInteractionGroup<double>(N, dummy_idxs, nb_pot->get_beta(), nb_pot->get_cutoff()));
+        return std::shared_ptr<BoundPotential>(new BoundPotential(ixn_group, shape, &h_params[0]));
+    } else {
+        throw std::runtime_error("unable to cast potential to NonbondedAllPairs");
+    }
+}
+
+// Recursively flatten the potentials. Important to find specific NonbondedAllPairs potentials for multiple_steps_local which
+// can be wrapped in FanoutSummedPotential or SummedPotential objects. Creates copies of bound potentials, for simplicity and safety.
+void flatten_potentials(
```
badisa, line 95: This makes copies of all of the nested potentials, which is a lot of overhead.

Likely need to add constructors for bound potential to take in gpu memory

timemachine/potentials/generic.py
```diff
@@ -439,6 +439,12 @@ def from_gpu(p: gpu.CustomOpWrapper) -> Optional[Potential]:
     if isinstance(p, gpu.Nonbonded):
         return Nonbonded.from_gpu(p)
 
+    if isinstance(p, gpu.NonbondedAllPairs):
```
badisa, line 442: In some exploration I noticed that these two were missing from the generic and added them. Can easily remove these, just left them as they seemed like nice to haves.

timemachine/cpp/src/local_md_utils.hpp
```diff
@@ -0,0 +1,138 @@
...
+#include "nonbonded_interaction_group.hpp"
+
+#include "fanout_summed_potential.hpp"
+#include "summed_potential.hpp"
+
+#include <typeinfo>
+
+namespace timemachine {
+
+bool is_nonbonded_all_pairs_potential(std::shared_ptr<Potential> pot) {
```
mcwitt, line 12: Should the implementations be in a `.cpp` or `.cu` file? (seems like we'd run into multiple definitions issues if `local_md_utils.hpp` is included in multiple places)

timemachine/cpp/src/local_md_utils.hpp
mcwitt: I'm not sure I understand the necessity of creating copies of the nested potentials. Does this have to do with the fact that local MD modifies the neighborlist used by `NonbondedAllPairs`? Would it be possible to at least avoid copying the other potentials here?

timemachine/cpp/src/local_md_utils.hpp
badisa: Fixed in [24a5d72](https://github.com/proteneer/timemachine/pull/975/commits/24a5d72de80e2e0b3814f90b005ad00aed907a05)

Along with moving a function that was defined here that was in the wrong place

timemachine/potentials/generic.py
mcwitt: LGTM (I'd just been adding things to `generic` as I found them useful)

timemachine/cpp/src/context.cu
```diff
@@ -263,7 +195,8 @@ std::array<std::vector<double>, 2> Context::multiple_steps_local(
         int num_row_indices = p_num_selected.data[0];
         int num_col_indices = N_ - num_row_indices;
         if (num_row_indices == 0 || num_col_indices == 0) {
-            throw std::runtime_error("local md no longer stable, check system");
+            throw std::runtime_error(
+                "entire system selected, reduce radius, increase k, check system or use Context::multiple_steps");
```
mcwitt, line 199: nit: for clarity
```suggestion
                "entire system selected; reduce radius, increase k, check system or use Context::multiple_steps");
```

timemachine/cpp/src/local_md_utils.hpp
badisa: This is primarily due to the simplicity of this approach. Had played around with several other versions that avoided copying, but seemed overly complex for a first pass that is likely to change.

The goals are to be able determine the following:
* Find `NonbondedAllPairs` Potentials
* Determine that is only one `NonbondedAllPairs`, hard to verify the case where you have multiple
* Have the parameters to construct the `NonbondedInteractionGroup`

Will try modifying it to return a list of NonBondedAllPairs wrapped as BoundPotentials and only reconstruct SummedPotentials/NonbondedAllPairs.

badisa: Changed as the approach of flattening only the nonbonded all pairs is simple and gets more directly at the problem.

Does add a bit more nuance in that the BoundPotential of the NonbondedAllPairs can't be modified, but only the underlying potential is modified which is shared with the potential in the FanoutSummedPotential.

badisa: Done in [5b41c5d](https://github.com/proteneer/timemachine/pull/975/commits/5b41c5d7d354686cf213b781ecd8a7cccf8b912e)

mcwitt (approved):

timemachine/cpp/src/context.cu
```diff
@@ -263,7 +193,8 @@ std::array<std::vector<double>, 2> Context::multiple_steps_local(
         int num_row_indices = p_num_selected.data[0];
         int num_col_indices = N_ - num_row_indices;
         if (num_row_indices == 0 || num_col_indices == 0) {
-            throw std::runtime_error("local md no longer stable, check system");
+            throw std::runtime_error(
+                "entire system selected; reduce radius, increase k, check system or use Context::multiple_steps");
```
maxentile, line 197: Missed this in previous PR, but is it more appropriate for this to be a warning rather than an error? (Since there's a nonzero chance of this happening under normal operation, and selecting the whole system would cause no problems in principle.)

If we need to throw an error here because it is difficult to throw an error closer to the assumption this would violate (which may be independent of local MD, maybe in the neighborlist?), please expand this error message accordingly.

timemachine/cpp/src/local_md_utils.cu
```diff
@@ -0,0 +1,147 @@
...
+                i++;
+            }
+            get_nonbonded_all_pair_potentials(flattened_bps, nb_pots);
+            continue;
+        }
+        if (is_nonbonded_all_pairs_potential(pot->potential)) {
+            nb_pots.push_back(pot);
+        }
+    }
+}
```
maxentile, line 145: I can't review this confidently, but it looks reasonable.

timemachine/cpp/src/context.cu
```diff
@@ -272,77 +203,35 @@ std::array<std::vector<double>, 2> Context::multiple_steps_local(
...
-        gpuErrchk(cudaPeekAtLastError());
-
-        k_unique_indices<<<ceil_divide(max_interactions, tpb), tpb, 0, stream>>>(
-            max_interactions, N_, nblist.get_ixn_atoms(), d_col_idxs.data);
-        gpuErrchk(cudaPeekAtLastError());
-
-        // Add the inner indices to the outer indices
-        k_unique_indices<<<ceil_divide(num_row_indices, tpb), tpb, 0, stream>>>(
-            num_row_indices, N_, d_row_idxs.data, d_col_idxs.data);
-        gpuErrchk(cudaPeekAtLastError());
```
maxentile, line 313: This looks like an improvement in safety -- I did not have a good grasp of when the "inner and outer shell" optimization was correct.

maxentile (approved):

timemachine/cpp/src/context.cu
```diff
@@ -209,49 +137,51 @@ std::array<std::vector<double>, 2> Context::multiple_steps_local(
...
         gpuErrchk(cudaPeekAtLastError());
         // Generate values between (0, 1.0]
         curandErrchk(curandGenerateUniform(cr_rng, probability_buffer.data, round_up_even(N_)));
 
         unsigned int reference_idx = local_idxs[random_dist(rng)];
 
-        k_log_probability_selection<double><<<ceil_divide(N_, tpb), tpb, 0, stream>>>(
-            N_, kBT, radius, k, reference_idx, d_x_t_, d_box_t_, probability_buffer.data, d_shell_idxs_inner.data);
+        // Select all of the particles that will be free
+        k_log_probability_selection<float><<<ceil_divide(N_, tpb), tpb, 0, stream>>>(
```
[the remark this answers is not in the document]
badisa: It was, but a pretty insignificant.

timemachine/cpp/src/local_md_utils.cu
badisa: If you want to walk through this together offline let me know.

timemachine/cpp/src/context.cu
badisa: Definitely much cleaner with the ixn group code

timemachine/cpp/src/context.cu
badisa: We could have it fall back on global MD in this case.

Though in the case of falling back to global do we still want the restraint?

maxentile: >We could have it fall back on global MD in this case.

Possibly -- but still warn of possibly unintended configuration?

>Though in the case of falling back to global do we still want the restraint?

Yes.

Let's say we have a nonzero chance of selecting each particle in the system independently (which is true with all settings of the currently implemented variant). Then on any given iteration, there's also a nonzero chance of selecting all the particles at once. We might have a branch that switches to a more efficient implementation on those iterations, but omitting the restraint when that branch is taken would not be correct.

There are special cases where omitting the restraint would be equivalent to including the restraint. (E.g. if the flat-bottom radius is very large, so that there's always a 100% chance of selecting every particle.) But in general, I think we want the implementation to be equivalent to including a restraint function that is [consistent with the selection probability function](https://github.com/proteneer/timemachine/blob/2779aef8c3a30f2e28b1c0aae9f77cf27f816f55/timemachine/md/local_resampling.py#L23-L25).

maxentile: ^ Non-blocking -- can be considered further in a later PR 👍

PR #976, opened 2023-03-15
Fix HMR typo

maxentile:
While prototyping a simpler equilibration method, noticed a small typo in `equilibrate_host` (HMR was being applied to a mass array that was never referenced)


mcwitt:
Good catch! Do we expect this to have significant correctness implications? I'm unclear on how important HMR is for the optimization done here.

maxentile:
>Do we expect this to have significant correctness implications? I'm unclear on how important HMR is for the optimization done here.

I don't think so -- possibly it could allow stable equilibration with a faster annealing schedule or fewer MD steps? (But I don't think this has been causing any problems currently.)

mcwitt (approved):

PR #978, opened 2023-03-15
Robust proposals, applied to pre-equilibration

maxentile:
Adds a reference implementation of the Barker proposal, based on the paper:

[Livingstone, Zanella, 2020] The Barker proposal: combining robustness and efficiency in gradient-based MCMC https://arxiv.org/abs/1908.11812

This method is very appealing to me, and could likely find several interesting FEP-related applications -- I've recently prototyped some of these for my own curiosity, and discussed some possible future directions with @mcwitt and others.

----------------

One immediate application may be to avoid some arbitrariness in methods for pre-equilibration.
* Currently: "4D annealing to the desired state, followed by FIRE minimization at that state" ([`minimize_host_4d`](https://github.com/proteneer/timemachine/blob/1d9eff40bcd28d395dad20d8cb9049a16088efe0/timemachine/md/minimizer.py#L98-L101))
* Option added in this PR: "just run a robust sampler at the desired state" ([`equilibrate_host_barker`](https://github.com/proteneer/timemachine/blob/1d9eff40bcd28d395dad20d8cb9049a16088efe0/timemachine/md/minimizer.py#L267-L269)) (clashy initial conditions pose no problem, and it even appears stable when the Metropolis check is omitted...)

This seems like an improvement in all the aspects I can think to care about for an initial minimizer (finds a configuration with lower force norm, in less wall time, requiring a smaller change to initial coords, using fewer lines of simpler code, ...). But so far I've only tested on the ⁠[hif2a example](https://github.com/proteneer/timemachine/blob/5479c2c2922a7327cb32a37323162c962cdcb570/tests/test_minimizer.py#L12-L28) in test_minimizer.py and on made-up toy examples locally (e.g. initializing a bunch of LJ particles on top of each other)...

Comparison to minimize_host_4d on the hif2a example:

```text
using unadjusted Barker proposal @ temperature = 0.001 K...
        force norm after low-temperature equilibration: 498.011 kJ/mol / nm
        max distance traveled = 0.085 nm
        done in 4.504 s
using 4D FIRE annealing
        force norm after 4D FIRE annealing: 673.744 kJ/mol / nm
        max distance traveled = 0.445 nm
        done in 38.187 s
```

Also not sure if we can further simplify from "equilibrate at temp ~ 0 K, then equilibrate at temp ~ 300 K" to "just equilibrate at temp ~ 300 K". This possibility is successfully exercised in the tests, but I may be overlooking other requirements that rule out this simplification.

```text
using unadjusted Barker proposal @ temperature = 300.0 K...
        force norm after room-temperature equilibration: 7372.178 kJ/mol / nm
        max distance traveled = 0.032 nm
        done in 6.255 s
```

### DIFF
--- a/tests/test_barker.py
+++ b/tests/test_barker.py
@@ -109,7 +109,8 @@ def test_accurate_mcmc(threshold=1e-4):
 
 @pytest.mark.nogpu
 @pytest.mark.parametrize("proposal_sig", [0.1, 1.0])
-def test_proposal_magnitude_independent_of_force_magnitude(proposal_sig):
+@pytest.mark.parametrize("shared_seed", range(5))
+def test_proposal_magnitude_independent_of_force_magnitude(proposal_sig, shared_seed):
     """Generate Lennard-Jones-informed proposals from clashy vs. relaxed starting points
         (where |force(x_clash)| ~= +inf and |force(x_relaxed)| ~ 0).
 
@@ -126,14 +127,17 @@ def test_proposal_magnitude_independent_of_force_magnitude(proposal_sig):
 
     barker_prop = BarkerProposal(grad_log_q, proposal_sig=proposal_sig)
     expected_sq_distance = barker_prop.proposal_sig ** 2
+    n_samples = 100_000
+    rel_tol = 1e-2
+    abs_tol = 1e-2
 
     # ---------------------
-    np.random.seed(0)
+    np.random.seed(shared_seed)
 
     # sample many proposals from a clashy initial condition
-    x_clash = np.array([1e-3])
+    x_clash = 1e-3 * np.ones(n_samples)
     assert np.linalg.norm(grad_log_q(x_clash)) > 1e10
-    ys_clash = np.array([barker_prop.sample(x_clash) for _ in range(10_000)])
+    ys_clash = barker_prop.sample(x_clash)
 
     # assert that the gradient-informed proposals
     # are the same avg. sq. distance from starting point
@@ -141,28 +145,28 @@ def test_proposal_magnitude_independent_of_force_magnitude(proposal_sig):
     disp_clash = (ys_clash - x_clash).flatten()
     mean_sq_distance_clash = (disp_clash ** 2).mean()
 
-    assert pytest.approx(mean_sq_distance_clash, rel=1e-1) == expected_sq_distance
+    assert pytest.approx(mean_sq_distance_clash, rel=rel_tol) == expected_sq_distance
 
     # assert that the proposals are skewed in the expected direction
     skew = np.sign(disp_clash).mean()
-    assert pytest.approx(skew, abs=1e-2) == 1.0
+    assert pytest.approx(skew, abs=abs_tol) == 1.0
 
     # ---------------------
-    np.random.seed(0)
+    np.random.seed(shared_seed)
 
     # sample many proposals from a relaxed initial condition
-    x_relaxed = np.array([1e3])
+    x_relaxed = 1e3 * np.ones(n_samples)
     assert np.linalg.norm(grad_log_q(x_relaxed)) < 1e-10
-    ys = np.array([barker_prop.sample(x_relaxed) for _ in range(10_000)])
+    ys = barker_prop.sample(x_relaxed)
 
     # again, assert proposals are the expected avg. sq. distance from starting point
     disp = (ys - x_relaxed).flatten()
     mean_sq_distance = (disp ** 2).mean()
-    assert pytest.approx(mean_sq_distance, rel=1e-1) == expected_sq_distance
+    assert pytest.approx(mean_sq_distance, rel=rel_tol) == expected_sq_distance
 
     # assert that the proposals are not skewed much
     skew = np.sign(disp).mean()
-    assert pytest.approx(skew, abs=1e-2) == 0.0
+    assert pytest.approx(skew, abs=abs_tol) == 0.0
 
     # ---------------------
 
--- a/tests/test_barker.py
+++ b/tests/test_barker.py
@@ -61,8 +61,6 @@ def test_proposal_normalization(x0, proposal_sig):
 
 @pytest.mark.nogpu
 def test_accurate_mcmc(threshold=1e-4):
-    np.random.seed(0)
-
     def log_q(x):
         return np.sum(-(x ** 4))
 
@@ -72,7 +70,7 @@ def test_accurate_mcmc(threshold=1e-4):
     # system with a large number of quartic oscillators
     x = np.zeros(1_000)
 
-    prop = BarkerProposal(grad_log_q, proposal_sig=0.1)
+    prop = BarkerProposal(grad_log_q, proposal_sig=0.1, seed=0)
 
     def mcmc_move(x):
         y = prop.sample(x)
@@ -109,8 +107,8 @@ def test_accurate_mcmc(threshold=1e-4):
 
 @pytest.mark.nogpu
 @pytest.mark.parametrize("proposal_sig", [0.1, 1.0])
-@pytest.mark.parametrize("shared_seed", range(5))
-def test_proposal_magnitude_independent_of_force_magnitude(proposal_sig, shared_seed):
+@pytest.mark.parametrize("seed", range(5))
+def test_proposal_magnitude_independent_of_force_magnitude(proposal_sig, seed):
     """Generate Lennard-Jones-informed proposals from clashy vs. relaxed starting points
         (where |force(x_clash)| ~= +inf and |force(x_relaxed)| ~ 0).
 
@@ -125,14 +123,13 @@ def test_proposal_magnitude_independent_of_force_magnitude(proposal_sig, shared_
 
     grad_log_q = jit(grad(log_q))
 
-    barker_prop = BarkerProposal(grad_log_q, proposal_sig=proposal_sig)
-    expected_sq_distance = barker_prop.proposal_sig ** 2
+    expected_sq_distance = proposal_sig ** 2
     n_samples = 100_000
     rel_tol = 1e-2
     abs_tol = 1e-2
 
     # ---------------------
-    np.random.seed(shared_seed)
+    barker_prop = BarkerProposal(grad_log_q, proposal_sig=proposal_sig, seed=seed)
 
     # sample many proposals from a clashy initial condition
     x_clash = 1e-3 * np.ones(n_samples)
@@ -152,7 +149,7 @@ def test_proposal_magnitude_independent_of_force_magnitude(proposal_sig, shared_
     assert pytest.approx(skew, abs=abs_tol) == 1.0
 
     # ---------------------
-    np.random.seed(shared_seed)
+    barker_prop = BarkerProposal(grad_log_q, proposal_sig=proposal_sig, seed=seed)
 
     # sample many proposals from a relaxed initial condition
     x_relaxed = 1e3 * np.ones(n_samples)
--- a/tests/test_barker.py
+++ b/tests/test_barker.py
@@ -1,6 +1,4 @@
 """
-Ideas for tests:
-
 Assert that the move:
 
     Is correctly implemented:
@@ -11,11 +9,13 @@ Assert that the move:
     Has robust behavior:
     - [x] Resolves clashes without blowing up or requiring alchemical intermediates
         (Added to test_minimizer.py)
-    - [ ] Avg. norm(proposal - x) remains <= avg. norm(gaussian(0, sig)), even when norm(grad_log_q(x)) >> 1
+    - [x] Avg. norm(proposal - x) remains ~= avg. norm(gaussian(0, sig)), even when norm(grad_log_q(x)) >> 1
 """
 
 import numpy as np
 import pytest
+from jax import grad, jit
+from jax import numpy as jnp
 
 from timemachine.md.barker import BarkerProposal
 
@@ -105,3 +105,67 @@ def test_accurate_mcmc(threshold=1e-4):
     histogram_mse = np.mean((y_ref - y_empirical) ** 2)
 
     assert histogram_mse < threshold
+
+
+@pytest.mark.nogpu
+@pytest.mark.parametrize("proposal_sig", [0.1, 1.0])
+def test_proposal_magnitude_independent_of_force_magnitude(proposal_sig):
+    """Generate Lennard-Jones-informed proposals from clashy vs. relaxed starting points
+        (where |force(x_clash)| ~= +inf and |force(x_relaxed)| ~ 0).
+
+    Assert that:
+    * the avg. squared distance between proposal and starting point is the same in both cases
+    * the proposal skew is ~ 100% in the |force| ~= +inf case, and ~ 0% in the |force| ~= 0 case
+    """
+
+    def log_q(r):
+        sig, eps = 1.0, 1.0
+        return jnp.sum(-4 * eps * ((sig / r) ** 12 - (sig / r) ** 6))
+
+    grad_log_q = jit(grad(log_q))
+
+    barker_prop = BarkerProposal(grad_log_q, proposal_sig=proposal_sig)
+    expected_sq_distance = barker_prop.proposal_sig ** 2
+
+    # ---------------------
+    np.random.seed(0)
+
+    # sample many proposals from a clashy initial condition
+    x_clash = np.array([1e-3])
+    assert np.linalg.norm(grad_log_q(x_clash)) > 1e10
+    ys_clash = np.array([barker_prop.sample(x_clash) for _ in range(10_000)])
+
+    # assert that the gradient-informed proposals
+    # are the same avg. sq. distance from starting point
+    # as if proposed from the base kernel Normal(mu=x_clash, sig=proposal_sig)
+    disp_clash = (ys_clash - x_clash).flatten()
+    mean_sq_distance_clash = (disp_clash ** 2).mean()
+
+    assert pytest.approx(mean_sq_distance_clash, rel=1e-1) == expected_sq_distance
+
+    # assert that the proposals are skewed in the expected direction
+    skew = np.sign(disp_clash).mean()
+    assert pytest.approx(skew, abs=1e-2) == 1.0
+
+    # ---------------------
+    np.random.seed(0)
+
+    # sample many proposals from a relaxed initial condition
+    x_relaxed = np.array([1e3])
+    assert np.linalg.norm(grad_log_q(x_relaxed)) < 1e-10
+    ys = np.array([barker_prop.sample(x_relaxed) for _ in range(10_000)])
+
+    # again, assert proposals are the expected avg. sq. distance from starting point
+    disp = (ys - x_relaxed).flatten()
+    mean_sq_distance = (disp ** 2).mean()
+    assert pytest.approx(mean_sq_distance, rel=1e-1) == expected_sq_distance
+
+    # assert that the proposals are not skewed much
+    skew = np.sign(disp).mean()
+    assert pytest.approx(skew, abs=1e-2) == 0.0
+
+    # ---------------------
+
+    # shared random seed, so can assert with tighter tolerance that (A == B)
+    # than we could assert ((A == expectation) and (B == expectation)) above
+    assert pytest.approx(mean_sq_distance, rel=1e-100) == mean_sq_distance_clash
--- a/tests/test_barker.py
+++ b/tests/test_barker.py
@@ -6,7 +6,7 @@ Assert that the move:
     Is correctly implemented:
     - [x] Methods sample(x) and log_density(x, y) support expected shapes
     - [x] Proposal is normalized
-    - [ ] An MCMC move constructed from this proposal accurately samples a simple target
+    - [x] An MCMC move constructed from this proposal accurately samples a simple target
 
     Has robust behavior:
     - [x] Resolves clashes without blowing up or requiring alchemical intermediates
@@ -20,6 +20,7 @@ import pytest
 from timemachine.md.barker import BarkerProposal
 
 
+@pytest.mark.nogpu
 def test_barker_shapes():
     def grad_log_q(x):
         return np.ones_like(x)
@@ -38,6 +39,7 @@ def test_barker_shapes():
         assert np.isscalar(logpdf)
 
 
+@pytest.mark.nogpu
 @pytest.mark.parametrize("x0", [-1, 0, +1])
 @pytest.mark.parametrize("proposal_sig", [0.1, 1.0])
 def test_proposal_normalization(x0, proposal_sig):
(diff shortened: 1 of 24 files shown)


tests/test_barker.py
```diff
@@ -0,0 +1,171 @@
...
+    skew = np.sign(disp_clash).mean()
+    assert pytest.approx(skew, abs=1e-2) == 1.0
+
+    # ---------------------
+    np.random.seed(0)
+
+    # sample many proposals from a relaxed initial condition
+    x_relaxed = np.array([1e3])
+    assert np.linalg.norm(grad_log_q(x_relaxed)) < 1e-10
+    ys = np.array([barker_prop.sample(x_relaxed) for _ in range(10_000)])
```
mcwitt, line 156: nit: this is ~1000x faster in my testing (probably not important, but does reduce test duration from 15s to ~0.1s)
```suggestion
    ys = barker_prop.sample(np.tile(x_relaxed, 10_000))
```

mcwitt (commented):
This looks awesome! I left some initial comments, but am planning to take another pass on the minimization-specific stuff

tests/test_barker.py
maxentile: Ahh, good catch -- in this case these two should be identical (1 proposal on 10000D state vs. 10000 independent proposals on 1D state...)

tests/test_barker.py
```diff
@@ -0,0 +1,171 @@
...
+    mean_sq_distance_clash = (disp_clash ** 2).mean()
+
+    assert pytest.approx(mean_sq_distance_clash, rel=1e-1) == expected_sq_distance
+
+    # assert that the proposals are skewed in the expected direction
+    skew = np.sign(disp_clash).mean()
+    assert pytest.approx(skew, abs=1e-2) == 1.0
+
+    # ---------------------
+    np.random.seed(0)
```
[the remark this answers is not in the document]
maxentile: Good catch -- will add `pytest.parameterize("shared_seed", [0,1,2,3,4])` or similar, and adjust test thresholds accordingly...

tests/test_barker.py
maxentile: >Hmm, is it ever the case that they would be different?

Depends how we define `log_q` / `grad_log_q`.

(In the test, defined in such a way that `grad_log_q(xs) == [grad_log_q(x) for x in xs]`, but this may not be the case for other user-defined functions `grad_log_q`.)

timemachine/md/minimizer.py
```diff
@@ -172,15 +183,104 @@ def minimize_host_4d(mols, host_system, host_coords, ff, box, mol_coords=None) -
...
+    grad_log_q = lambda x_host: -du_dx_host_fxn(x_host) / (BOLTZ * temperature)
+
+    # TODO: arbitrary step size -- adjust?
+    # TODO: if needed, revisit choice to omit Metropolis correction
+    barker_prop = BarkerProposal(grad_log_q, 0.0001)
+
+    x_host = np.array(host_coords)
+
+    max_iters = 1000
+    early_stop_threshold = MAX_FORCE_NORM / 100  # TODO: arbitrary threshold -- adjust?
```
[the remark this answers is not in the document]
maxentile: Decided to simplify: remove `early_stop_threshold`, rename `max_iters` -> `n_steps`, and expose `n_steps` as optional argument in https://github.com/proteneer/timemachine/pull/978/commits/811557f73b7e303a9eebd88d108f30093e566752

timemachine/md/minimizer.py
```diff
@@ -172,15 +183,104 @@ def minimize_host_4d(mols, host_system, host_coords, ff, box, mol_coords=None) -
...
+def equilibrate_host_barker(mols, host_system, host_coords, ff, box, mol_coords=None, temperature=300.0):
+    """possibly simpler alternative to minimize_host_4d, for purposes of
+    clash resolution and initial pre-equilibration"""
+
+    du_dx_host_fxn = make_host_du_dx_fxn(mols, host_system, host_coords, ff, box, mol_coords)
+    grad_log_q = lambda x_host: -du_dx_host_fxn(x_host) / (BOLTZ * temperature)
+
+    # TODO: arbitrary step size -- adjust?
+    # TODO: if needed, revisit choice to omit Metropolis correction
+    barker_prop = BarkerProposal(grad_log_q, 0.0001)
```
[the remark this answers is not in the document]
maxentile: Yes, I think so -- added in https://github.com/proteneer/timemachine/pull/978/commits/811557f73b7e303a9eebd88d108f30093e566752

tests/test_barker.py
maxentile: Applied in https://github.com/proteneer/timemachine/pull/978/commits/96e035df2074527faf9013dfd99284ddd38fba45

tests/test_barker.py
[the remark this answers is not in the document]
maxentile: Applied in https://github.com/proteneer/timemachine/pull/978/commits/96e035df2074527faf9013dfd99284ddd38fba45

timemachine/md/minimizer.py
```diff
@@ -26,6 +28,15 @@ class MinimizationError(Exception):
     pass
 
 
+def check_force_norm(du_dx):
+    """raise Minimization error if the force on any atom exceeds MAX_FORCE_NORM"""
+    norm = np.linalg.norm(du_dx, axis=-1)
+    if not np.all(norm < MAX_FORCE_NORM):
```
mcwitt, line 34: nit: seems a bit odd to me that this refers to a global constant (vs. optional argument). I see that the constant is used in one other place (https://github.com/proteneer/timemachine/blob/195ce1c0f9cebc7f184b5322dac2faba0345ef7e/timemachine/fe/model_utils.py#L43), but it's a bit unclear to me that it makes sense to reuse the same value in both contexts.

maxentile: Intent was just to reduce repetition of this check within `minimizer.py` -- the suggestion of an optional argument makes sense, applied in https://github.com/proteneer/timemachine/pull/978/commits/743706d4972fe3adddc1c4a62b1950986b97589b

timemachine/md/minimizer.py
```diff
@@ -172,15 +183,94 @@ def minimize_host_4d(mols, host_system, host_coords, ff, box, mol_coords=None) -
...
 
 
+def make_gpu_impl(potentials):
+    """return bound impl of a SummedPotential constructed from potentials"""
+
+    params = [p.params for p in potentials]
+    flat_params = np.concatenate([param.reshape(-1) for param in params])
+
+    generic_potentials = [generic.from_gpu(p) for p in potentials]
+    summed_potential = generic.SummedPotential(generic_potentials, params)
```
mcwitt, line 198: nit: could this be simplified by using `timemachine.lib.potentials.SummedPotential` directly, rather than converting from GPU to generic and back to GPU?

timemachine/md/minimizer.py
```diff
@@ -172,15 +183,94 @@ def minimize_host_4d(mols, host_system, host_coords, ff, box, mol_coords=None) -
...
+def equilibrate_host_barker(
+    mols, host_system, host_coords, ff, box, mol_coords=None, temperature=300.0, proposal_stddev=0.0001, n_steps=1000
+):
+    """possibly simpler alternative to minimize_host_4d, for purposes of
+    clash resolution and initial pre-equilibration"""
+
+    du_dx_host_fxn = make_host_du_dx_fxn(mols, host_system, host_coords, ff, box, mol_coords)
+    grad_log_q = lambda x_host: -du_dx_host_fxn(x_host) / (BOLTZ * temperature)
+
+    # TODO: if needed, revisit choice to omit Metropolis correction
```
mcwitt, line 260: Can you comment on why it seems valid to ignore the Metropolis correction? Do we expect this to approximately correctly sample from the finite temperature distribution without the correction?

timemachine/md/minimizer.py
mcwitt, line 1: nit, long-term organization: it feels a bit weird to add this to `md.minimizer`, since it's not really minimization. Would it make sense to put in a different file, or generalize the name of this submodule? (don't immediately have ideas that I like for this..)

timemachine/md/minimizer.py
maxentile: Ahh, good call! Will fix, no reason for the round-trip -- just this is the code snippet I had most readily at-hand.

tests/test_barker.py
```diff
@@ -0,0 +1,175 @@
...
+
+    y_grid = np.linspace(-10, +10, 1_000)
+
+    prop = BarkerProposal(grad_log_q, proposal_sig=proposal_sig)
+    logpdf_grid = np.array([prop.log_density(x0, y) for y in y_grid])
+    pdf_grid = np.exp(logpdf_grid)
+
+    Z = np.trapz(pdf_grid, y_grid)
+
+    assert pytest.approx(Z) == 1
```
badisa, line 59: nit/question: Should we prefer np over pytest for numerical comparisons?

timemachine/md/minimizer.py
maxentile: >Can you comment on why it seems valid to ignore the Metropolis correction? Do we expect this to approximately correctly sample from the finite temperature distribution without the correction?

In the limit of `proposal_stddev` approaching 0, I think this proposal process should sample from the finite temperature distribution even without Metropolis correction. (Following the "local balancing" argument in the reference, and based on numerical experiments.)

Omitting the Metropolis correction when `proposal_stddev > 0` is not something that I think was intended by the authors, and I don't think it's mentioned / recommended in the paper.

Anecdotally, I would not expect this to outperform our default approximate sampler that omits a Metropolis correction (BAOAB) once equilibrated.

(I would expect the sampling bias would be worse than BAOAB's at a comparable "step size". I'd also expect this move to be much less productive per gradient evaluation, since it's not inertial -- similar to how BAOAB with `friction=+np.inf` is less efficient than BAOAB with `friction=1.0`.)

The main advantage is robustness w.r.t. clashy initialization.

Anecdotally, the unadjusted Barker proposal with proposal_stddev = 0.0001 nm appears to be good enough to resolve Lennard-Jones clashes, and can make progress even when `|force| = +inf`.

...

I'll flesh out some comments in the docstring, add optional argument to use Metropolis correction, and add some other guardrails now that `proposal_stddev` is exposed to a user of `host_equilibrate_barker`.

timemachine/md/minimizer.py
maxentile: Wanted to put near the most similar existing functions `minimize_host_4d`, `equilibrate_host` , which are currently in minimizer.py -- may be better to split this into multiple files, possibly adding a new file `equilibrate.py`

timemachine/md/minimizer.py
```diff
@@ -172,15 +183,94 @@ def minimize_host_4d(mols, host_system, host_coords, ff, box, mol_coords=None) -
...
+    """construct function to compute du_dx w.r.t. host coords, given fixed mols and box"""
+
+    # openmm host_system -> timemachine host_bps
+    host_bps, _ = openmm_deserializer.deserialize_system(host_system, cutoff=1.2)
+
+    # construct appropriate topology from (mols, ff)
+    if len(mols) == 1:
+        top = topology.BaseTopology(mols[0], ff)
+    elif len(mols) == 2:
+        top = topology.DualTopologyMinimization(mols[0], mols[1], ff)
```
badisa, line 213: Should this be `DualTopology` rather than `DualTopologyMinimization`? The behavior looks identical (at lamb == 0, no difference if I am reading it right), but conceptual BaseTopology and DualTopologyMinimization seem different.

timemachine/md/minimizer.py
```diff
@@ -26,6 +28,15 @@ class MinimizationError(Exception):
     pass
 
 
+def check_force_norm(du_dx, threshold=MAX_FORCE_NORM):
+    """raise Minimization error if the force on any atom exceeds MAX_FORCE_NORM"""
+    norm = np.linalg.norm(du_dx, axis=-1)
+    if not np.all(norm < threshold):
+        raise MinimizationError(
+            f"Minimization failed to reduce large forces below threshold: |frc| = {norm} > {threshold}"
```
badisa, line 36: nit: for readability might be nice to either do the max norm or sort them. When the number of particles is large you may not be able to see the force that is large.

On line 415 in the previous it would print the last.

timemachine/md/minimizer.py
```diff
@@ -172,15 +183,94 @@ def minimize_host_4d(mols, host_system, host_coords, ff, box, mol_coords=None) -
...
+
+        du_dx, _ = gpu_impl.execute(x, box)
+        du_dx_host = du_dx[:num_host_atoms]
+        return du_dx_host
+
+    return du_dx_host_fxn
+
+
+def equilibrate_host_barker(
+    mols, host_system, host_coords, ff, box, mol_coords=None, temperature=300.0, proposal_stddev=0.0001, n_steps=1000
```
badisa, line 252:
```suggestion
    mols, host_system, host_coords, ff, box, mol_coords=None, temperature=constants.DEFAULT_TEMP, proposal_stddev=0.0001, n_steps=1000
```

timemachine/md/minimizer.py
maxentile: Ahh, good call -- this was derived from the version in `minimize_host_4d` https://github.com/proteneer/timemachine/blob/195ce1c0f9cebc7f184b5322dac2faba0345ef7e/timemachine/md/minimizer.py#L175-L179
but the version in `local_minimize` is more informative https://github.com/proteneer/timemachine/blob/195ce1c0f9cebc7f184b5322dac2faba0345ef7e/timemachine/md/minimizer.py#L414-L419

timemachine/md/minimizer.py
```diff
@@ -172,15 +183,94 @@ def minimize_host_4d(mols, host_system, host_coords, ff, box, mol_coords=None) -
...
+    # bound impls of potentials @ lam=0 (fully coupled) endstate
+    params_potential_pairs = parameterize_system(hgt, ff, 0.0)
+    bound_potentials = [potential.bind(params) for params, potential in params_potential_pairs]
+    gpu_impl = make_gpu_impl(bound_potentials)
+
+    # read conformers from mol_coords if given, or each mol's conf0 otherwise
+    conf_list = [np.array(host_coords)]
+
+    if mol_coords is not None:
+        for mc in mol_coords:
```
badisa, line 228: Know this is duplicated from else where, but would be good to validate `len(mol_coords) == len(mols)`

timemachine/md/minimizer.py
```diff
@@ -172,15 +183,94 @@ def minimize_host_4d(mols, host_system, host_coords, ff, box, mol_coords=None) -
...
+            conf_list.append(mc)
+    else:
+        for mol in mols:
+            conf_list.append(get_romol_conf(mol))
+
+    combined_coords = np.concatenate(conf_list)
+
+    # wrap gpu_impl, partially applying box, mol coords
+    num_host_atoms = host_coords.shape[0]
+    assert box.shape == (3, 3)
```
badisa, line 238: nit: validate this at the start of the function before constructing potentials

timemachine/md/barker.py
```diff
@@ -0,0 +1,61 @@
...
+        ----------
+        [Livingstone, Zanella, 2020]
+            The Barker proposal: combining robustness and efficiency in gradient-based MCMC
+            https://arxiv.org/abs/1908.11812
+        """
+        self.grad_log_q = grad_log_q
+        assert proposal_sig > 0
+        self.proposal_sig = proposal_sig
+
+    def _sample(self, x, gaussian_rvs, uniform_rvs):
```
badisa, line 24: To clarify, `uniform_rvs` is expected to be a normal distribution? 

In `sample()` the gaussian_rvs and uniform_rvs both come from `randn` (https://numpy.org/doc/stable/reference/random/generated/numpy.random.randn.html) which are both normal.

maxentile: >To clarify, uniform_rvs is expected to be a normal distribution?

No -- expected to be uniform

>In sample() the gaussian_rvs and uniform_rvs both come from randn (https://numpy.org/doc/stable/reference/random/generated/numpy.random.randn.html) which are both normal.

No -- uniform_rvs come from `rand`, which is uniform in [0,1)

https://github.com/proteneer/timemachine/blob/743706d4972fe3adddc1c4a62b1950986b97589b/timemachine/md/barker.py#L42-L47

badisa: 🤦 Just can't read

maxentile: (Reason for splitting it up is so that `_sample` can be a deterministic function. Isolating the RNG calls seems cleaner in my opinion (more easily swap global numpy RNG vs. jax RNG if desired, the deterministc function may be more easily tested...), and could be useful for batching, which could look like the following if `grad_log_q` is amenable to jax transformations)

```python
    def batch_sample(self, xs):
        """[y_i ~ p(. | x_i) for x_i in xs], using vmap"""
        gauss = np.random.randn(*xs.shape)
        unif = np.random.rand(*xs.shape)

        return vmap(self._sample, (0, 0, 0))(xs, gauss, unif)
```

(batching was reverted in https://github.com/proteneer/timemachine/pull/978/commits/e2773b00dea061dc9305f5e1adf2001ee3efa9ff to keep scope smaller, and to avoid making `barker.py` depend on jax initially)

timemachine/md/minimizer.py
mcwitt: Sounds good. For this PR I think just a note in the docstring would be sufficient.

timemachine/md/barker.py
```diff
@@ -0,0 +1,61 @@
...
+        b_xz = np.sign(p_xz - uniform_rvs)
+
+        y = x + b_xz * z
+
+        return y
+
+    def sample(self, x):
+        """y ~ p(. | x)"""
+        gauss = np.random.randn(*x.shape)
+        unif = np.random.rand(*x.shape)
```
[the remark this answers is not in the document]
mcwitt: Just to note another alternative, I personally like the explicit pattern that JAX uses, where all random functions must take in a random state as an additional argument. (Here it could look like adding an argument `rng: numpy.random.Generator`). I think it's a little more flexible than objects maintaining random state (e.g. simplifies running independent chains or parallelizing sampling) and more explicit (downside is there's a little more boilerplate required to pass the rng around).

mcwitt (approved):
LGTM! (pending addition to docstring explaining assumptions around step size and finite-temperature sampling bias)

timemachine/md/barker.py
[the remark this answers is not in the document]
maxentile: >Here it could look like adding an argument rng: numpy.random.Generator to the sample method)

That would also match https://github.com/proteneer/timemachine/blob/80b748034c63918e4b28027848d10ac1a1ded8dd/timemachine/integrator.py#L145-L158

>I think it's a little more flexible than objects maintaining random state (e.g. simplifies running independent chains or parallelizing sampling) and more explicit 

Good points

timemachine/md/minimizer.py
maxentile: Round-trip avoided in https://github.com/proteneer/timemachine/pull/978/commits/11ab17763e24a875e1711e40ef3e9010a77deac4

badisa (approved):

timemachine/md/minimizer.py
maxentile: Added these notes, and modified test to exercise the case of temperature == 0 (aka grad_log_q is all +/-infs) https://github.com/proteneer/timemachine/pull/978/commits/38e76d141b789103e78d4eeb7e92f0e6902aa506

Did not add metropolis correction, but added assertion preventing user from setting proposal_stddev larger than max tested value.

timemachine/md/barker.py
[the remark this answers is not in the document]
maxentile: Thanks for both suggestions.

In the interest of keeping the signature simple, applied @badisa 's suggestion of having the object maintain random state https://github.com/proteneer/timemachine/pull/978/commits/b8dc43b21bc110f8e080167a88f988e0659b0ca5

@mcwitt : more control is still available by providing own source of gaussian and uniform r.v.'s to `_sample`

timemachine/md/minimizer.py
```diff
@@ -26,6 +28,15 @@ class MinimizationError(Exception):
     pass
 
 
+def check_force_norm(du_dx, threshold=MAX_FORCE_NORM):
```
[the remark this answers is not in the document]
maxentile: Made consistent in https://github.com/proteneer/timemachine/pull/978/commits/c516cbfec71782ad70267017cd2cf0a4d41eb752

timemachine/md/minimizer.py
maxentile: Expanded error message in https://github.com/proteneer/timemachine/pull/978/commits/c516cbfec71782ad70267017cd2cf0a4d41eb752 -- report max force norm, as well as the number of atoms that exceed threshold

timemachine/md/minimizer.py
maxentile: Applied in https://github.com/proteneer/timemachine/pull/978/commits/97196fb589c4b32c54cd2be832f7c1d9bf665058

timemachine/md/minimizer.py
maxentile: Input validation expanded in https://github.com/proteneer/timemachine/pull/978/commits/757323d85b85eb3e644f9dba9f0e1707d7f67612

timemachine/md/minimizer.py
maxentile: Moved to start of function in https://github.com/proteneer/timemachine/pull/978/commits/757323d85b85eb3e644f9dba9f0e1707d7f67612

timemachine/md/barker.py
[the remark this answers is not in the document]
mcwitt: Thinking about this a bit more, having random states scattered around in different objects vs. a single global random state seems potentially a bit dangerous; for example, a user might reasonably expect that setting `np.random.seed(123)` would make the following code deterministic. But with hidden random state, we have to be sure to track down all objects that maintain their own random state and be sure to reset/reconstruct them too. Don't mean to derail this PR, but it seems to me worth returning to this in the future.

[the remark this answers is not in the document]
maxentile: >for example, a user might reasonably expect that setting np.random.seed(123) would make the following code deterministic.

Agreed that would be an easy mistake to make. But it would also be relatively easy to detect. (User could call the "deterministic" code twice and usually observe that the outputs aren't the same.)

A possible benefit of having the object maintain random state is that certain other kinds of subtler mistakes become harder to make. (E.g. it's ~impossible to accidentally reuse the same random seed on each iteration of a simulation loop `for _ in range(T): x = prop.sample(x)`.)

>Don't mean to derail this PR, but it seems to me worth returning to this in the future.

This sounds like a good area to consider more carefully in the future. Maybe we can take a more unified approach to random state in timemachine's python layer.

(I don't think we currently have a unified approach here: some objects maintain their own random state, some have random functions that accept numpy `rng`'s or jax rng `key`'s, and some sites reference global numpy random state.)

(And a user who is being extra careful about random state can still supply draws to the deterministic function `_sample(x, gaussian_rvs, uniform_rvs)`.)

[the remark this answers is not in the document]
mcwitt: > E.g. it's ~impossible to accidentally reuse the same random seed on each iteration of a simulation loop

That's a good point, this does seem like a benefit of the current approach.

> I don't think we currently have a unified approach here

Agreed that we currently have mixed conventions. (Just wanted to express some reservations over "objects that use randomness maintain their own implicit random state" becoming the convention without further discussion.)

Sounds good to me to continue this discussion in another thread to not distract from this PR.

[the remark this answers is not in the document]
maxentile: >(Just wanted to express some reservations over "objects that use randomness maintain their own implicit random state" becoming the convention without further discussion.)

Agreed -- not obvious to me what's the best approach overall.

>Sounds good to me to continue this discussion in another thread to not distract from this PR.

Agreed! Migrated to https://github.com/proteneer/timemachine/issues/980

PR #981, opened 2023-03-17
Handle edge cases in local md

badisa:


timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -219,26 +220,20 @@ void declare_context(py::module &m) {
                 if (burn_in < 0) {
                     throw std::runtime_error("burn in steps must be greater than zero");
```
maxentile, line 221: nit[edge-case] `burn_in <= 0` to match error message, or update error message to match `burn_in < 0`

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -219,26 +220,20 @@ void declare_context(py::module &m) {
...
                 }
+                if (radius < 0.0) {
+                    throw std::runtime_error("radius must be greater or equal to zero");
+                }
+                if (k <= 0.0) {
+                    throw std::runtime_error("k must be greater than zero");
+                }
+
```
maxentile, line 229: behavior: is there a maximum value of `k` to enforce as well? (limit may be imposed by MD stability)

maxentile (approved):
Looks good!

timemachine/cpp/src/wrap_kernels.cpp
badisa: I would guess that it is imposed by MD stability, but haven't come across a value I would codify. Though we could say 100k, no strong reason to think we would want to go that high.

maxentile: Maybe 10x or 100x the max value of `k` you've tested, with a reminder to determine a threshold more precisely in the future?

(Confident we want to rule out `k == +inf`, but haven't attempted to determine a finite threshold.)

tests/test_md.py
```diff
@@ -345,24 +345,27 @@ def test_multiple_steps_local_validation(self):
 
         # Verify that indices are correctly checked
         ctxt = custom_ops.Context(coords, v0, box, intg_impl, bps)
-        with pytest.raises(RuntimeError, match="number of idxs must be at least 1"):
+        with pytest.raises(RuntimeError, match="indices can't be empty"):
             ctxt.multiple_steps_local(100, np.array([], dtype=np.int32), radius=radius)
 
-        with pytest.raises(RuntimeError, match="indices values must be less than N"):
+        with pytest.raises(RuntimeError, match="indice values must be less than N"):
```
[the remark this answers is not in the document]
badisa: Fixed in [1738d57](https://github.com/proteneer/timemachine/pull/981/commits/1738d577ff2a502a0a40b30f3a88a45511307907)

timemachine/cpp/src/wrap_kernels.cpp
badisa: Good catch, fixed in [6a8044f](https://github.com/proteneer/timemachine/pull/981/commits/6a8044fb14740fd3aa406915d5278136d11e3f12)

tests/test_md.py
```diff
@@ -532,6 +538,81 @@ def test_multiple_steps_local_consistency(self):
...
+        intg = LangevinIntegrator(temperature, dt, friction, masses, seed)
+
+        intg_impl = intg.impl()
+
+        ctxt = custom_ops.Context(coords, v0, box, intg_impl, bps)
+
+        xs, boxes = ctxt.multiple_steps_local(num_steps, local_idxs, radius=radius)
+
+        identical_positions = (xs == coords).sum(axis=1)
+        assert len(identical_positions) == 1, "Expected only a single atom to be stationary"
```
mcwitt, line 576: This assertion seems trivially true (i.e. independent of the equality check).

mcwitt: I mean that `assert len(identical_positions) == 1` is just checking that the first element of the shape is 1, but that will be true regardless of the values in `xs` and `coords`, right?

badisa: Talking offline, this was wrong. There was a mismatch in the shapes of the two arrays anyways which made it more confusing.

Fixed in [2cb977f](https://github.com/proteneer/timemachine/pull/981/commits/2cb977fa681004b8470e5a99273e681d1305a238)

tests/test_md.py
```diff
@@ -532,6 +538,81 @@ def test_multiple_steps_local_consistency(self):
         np.testing.assert_array_equal(summed_pot_xs, xs)
         np.testing.assert_array_equal(summed_pot_boxes, boxes)
 
+    def test_multiple_steps_local_entire_system(self):
+        """Verify that running multiple_steps_local is valid even when consuming the entire system, IE radius ~= inf.
+
+        - Only a single particle, the reference, should not move
+        """
+        mol, _ = get_biphenyl()
+        ff = Forcefield.load_from_file("smirnoff_1_1_0_sc.py")
```
mcwitt, line 547: nit/curiosity: why not use the default forcefield here?

badisa: The simple charge forcefield doesn't require openeye, which allows us to test the implementation free of needing a license.

mcwitt: Ah, makes sense. I missed that it was not just the older version of the default forcefield.

timemachine/cpp/src/context.cu
```diff
@@ -192,9 +192,14 @@ std::array<std::vector<double>, 2> Context::multiple_steps_local(
 
         int num_row_indices = p_num_selected.data[0];
         int num_col_indices = N_ - num_row_indices;
-        if (num_row_indices == 0 || num_col_indices == 0) {
-            throw std::runtime_error(
-                "entire system selected; reduce radius, increase k, check system or use Context::multiple_steps");
+
+        if (num_row_indices == 0) {
+            throw std::runtime_error("Context::multiple_steps_local(): no free particles selected");
```
mcwitt, line 197: Related to this [previous discussion](https://github.com/proteneer/timemachine/pull/975#discussion_r1134732982), wondering whether there's any concern about this being triggered in long simulations (even for reasonable parameter values, there is still some small probability, correct?). Would it make sense for this to be converted to a warning, similarly to the "entire system selected" case?

badisa: At the moment this crashes if the exception isn't raise as the code currently tries to run a kernel with 0 blocks. Though  we could make it work if this seems like a case to handle

badisa: Would also need to change NonbondedAllPairs (https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/nonbonded_all_pairs.cu#L153-L155)

We could also do something like the following to avoid running any GPU code for this case
```
if (num_row_indices == 0) {
    copy_current_x_and_box_to_output_buffer()
    return std::array<std::vector<double>, 2>({h_x_buffer, h_box_buffer});
}
```

badisa: > Makes sense to raise an error if this is not currently a case we handle. Would it make sense to raise the error from the site where we make an assumption about the number of blocks being nonzero (neighborlist?). That way it'll be obvious what needs to be changed if this ever does become an issue.

Since this is currently coming out of a kernel launch failure, don't think we want it to fail from the site, or if we do we would probably want an exception and not a call to exit. Already seems like multiple layers to the failure (kernel launch, NonbondedAllPairs, probably another) so the digging will have to occur regardless

badisa: This also makes it clear that we hit a case we don't support, rather than potentially hitting a bug due to an unexpected case.

mcwitt: I think it's fine as-is for now. Seems potentially useful to raise exceptions nearer to where assumptions are made in addition, looking ahead to when we want to enable this and remove the check in this PR. But I don't think it's necessary for this PR.

badisa: Will look at `radius >= reasonable_radius and k > reasonable_k`, initial thought is `reasonable_radius = 0.1` and `reasonable_k = 1`

badisa: Made changes to enforce 'reasonable' values here [354dfbe](https://github.com/proteneer/timemachine/pull/981/commits/354dfbedf02cebaf800a36fffc1e6c845e0b20df)

Had to change test to be on pathological case. Should be easy to use to make this case work in the future if we need it.

PR #983, opened 2023-03-23
Make random seed deterministic

jkausrelay:
- hash() is only deterministic for a given python invocation (see https://docs.python.org/3.3/reference/datamodel.html#object.__hash__)
- This means the random seed was being changed every time we run rbfe
- Use hashlib instead to make this deterministic
- Update test to explicitly check the seeds


maxentile (approved):
Oof, good fix.

(For tracking purposes, I introduced this bug in https://github.com/proteneer/timemachine/pull/936#discussion_r1091288793 )

PR #985, opened 2023-03-28
Draft: Fix test failure

jkausrelay:


mcwitt (approved):
I'm not sure I understand why bisection uses a different seed at lambda = 0.5, but this looks reasonable to me as a quick fix.

mcwitt:
> I'm not sure I understand why the initial state differs between the two methods at lambda = 0.5, but this looks reasonable to me as a quick fix.

Ah, I think I see what's going on:

```
np.linspace(0.01, 0.03, 3)[1] != (0.01 + 0.03) / 2.0
```
so https://github.com/mcwitt/timemachine/blob/47e020baab3d75122c8a197e436a2a96ac3c83b7/timemachine/fe/rbfe.py#L252 fails and we rerun the minimization in the case of bisection. Might be worth changing this to approximate equality in a future PR?

jkausrelay:
> 

Yes, there is little reason to re-minimize for such a small change in lambda.

PR #986, opened 2023-03-28
Remove deceptive warning around FF loading

badisa:
* Path("some_path") / "/bar" returns "Path(/bar)" which resulted in warning when reading an absolute path file

### DIFF
--- a/timemachine/ff/__init__.py
+++ b/timemachine/ff/__init__.py
@@ -84,7 +84,7 @@ class Forcefield:
         path = Path(path_or_str)  # Safe to construct a Path object from another Path object
 
         with resources.files("timemachine.ff.params") as params_path:
-            built_in_path = params_path / original_path
+            built_in_path = params_path / path.name
             if built_in_path.is_file():
                 if path.is_file():
                     warn(f"Provided path {original_path} shares name with builtin forcefield, falling back to builtin")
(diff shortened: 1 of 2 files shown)


jkausrelay (approved):

timemachine/ff/__init__.py
```diff
@@ -84,7 +84,7 @@ def load_from_file(cls, path_or_str: Union[str, Path]) -> "Forcefield":
         path = Path(path_or_str)  # Safe to construct a Path object from another Path object
 
         with resources.files("timemachine.ff.params") as params_path:
-            built_in_path = params_path / original_path
+            built_in_path = params_path / path.name
```
mcwitt, line 87: Question: would this change allow `/foo/bar/smirnoff_2_0_0_ccc.py` to match a built-in forcefield?

badisa: Yes, which was the expectation we set at the time. 

We could perhaps change that expectation, though I would probably err on the side of not having multiple ffs that have the same name but different parameters?

mcwitt: Ah, I see. I was under the impression that we would not fall back to a builtin FF if an absolute path is passed (even if the basename is the same as a builtin FF). The comment added in https://github.com/proteneer/timemachine/pull/986/commits/0f879aa3aa44445f0453a6a7246cdc151ef09e63 clarifies for me.

mcwitt (approved):

PR #987, opened 2023-03-30
Add version of infer_node_vals_and_errs accepting a networkx graph

mcwitt:
Quality of life improvement adding a utility function wrapping `infer_node_vals_and_errs` to accept a networkx graph.

* Adds `timemachine.fe.mle.infer_node_vals_and_errs_networkx`
* Adds test


tests/test_mle.py
```diff
@@ -177,6 +177,46 @@ def test_infer_node_dgs_w_error():
         assert res.rvalue > 0.9
 
 
+def test_infer_node_vals_and_errs_networkx():
+    edge_noise_stddev = np.random.rand()
+    g = generate_random_valid_regular_graph()
+    g = nx.convert_node_labels_to_integers(g)
```
badisa, line 183: Is this important to the test? In practice the nodes will be string labels

timemachine/fe/mle.py
```diff
@@ -212,3 +215,60 @@ def infer_node_vals_and_errs(
...
+    nx_graph: nx.Graph,
+    edge_diff_prop: str,
+    edge_stddev_prop: str,
+    ref_nodes: Sequence[Any],
+    ref_node_val_prop: str,
+    ref_node_stddev_prop: Optional[str] = None,
+    n_bootstrap: int = 100,
+    seed: int = 0,
+) -> Tuple[NDArray, NDArray]:
+    """Version of :py:func:`fe.mle.infer_node_vals_and_errs` that accepts a networkx graph.
```
badisa, line 230: Be explicit in the parent namespace of this function

```suggestion
    """Version of :py:func:`timemachine.fe.mle.infer_node_vals_and_errs` that accepts a networkx graph.
```

tests/test_mle.py
```diff
@@ -177,6 +177,46 @@ def test_infer_node_dgs_w_error():
...
+    n_nodes = g.number_of_nodes()
+
+    node_vals, edge_idxs, obs_edge_diffs, edge_stddevs = generate_instance(g, edge_noise_stddev)
```
badisa, line 186: This looks like it only tests the case where all nodes/edges have values. Should strengthen test to verify that if some nodes/edges don't have values, this still behaves as expected.

timemachine/fe/mle.py
```diff
@@ -212,3 +215,60 @@ def infer_node_vals_and_errs(
...
+    edge_stddev_prop: str
+        Edge property to use for standard deviations
+    ref_nodes: sequence
+        Reference nodes (must have properties ref_node_val_prop and ref_node_stddev_prop)
+    ref_node_val_prop: str
+        Node property to use for reference values
+    ref_node_stddev_prop: str or None, optional
+        Node property to use for reference standard deviations. If None, assume zero uncertainty in reference values.
+    n_bootstrap, seed:
+        See documentation for :py:func:`fe.mle.infer_node_vals_and_errs`
```
badisa, line 247:
```suggestion
        See documentation for :py:func:`timemachine.fe.mle.infer_node_vals_and_errs`
```

timemachine/fe/mle.py
```diff
@@ -212,3 +215,60 @@ def infer_node_vals_and_errs(
...
+
+    Returns
+    -------
+    dg: [K] array
+        inferred absolute values
+    dg_err : [K] array
+    """
+
+    node_to_idx = {n: idx for idx, n in enumerate(nx_graph.nodes)}
```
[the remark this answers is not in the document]
badisa: Looking at the [nx.relabel_nodes](https://networkx.org/documentation/stable/reference/generated/networkx.relabel.relabel_nodes.html#networkx.relabel.relabel_nodes) documentation, looks like it can change the ordering of the nodes. Which probably makes @maxentile suggestion important to ensure users know exactly which node the dG is assigned to.

tests/test_mle.py
mcwitt: This is a bit misleading since the test cases already have integer labels (the reference function does not support string labels). The call to `nx.convert_node_labels_to_integers` is used to make the node labels coincide with their indices (i.e. number them consecutively), which simplifies the implementation of the test.

tests/test_mle.py
mcwitt: Good call. Currently this does not support the case where some edges are missing values, but I think it can be generalized in tandem with implementing https://github.com/proteneer/timemachine/pull/987#discussion_r1154493745. Will make this change and strengthen the test.

timemachine/fe/mle.py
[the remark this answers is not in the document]
mcwitt: Nice, I like these suggestions. Agreed that (2) seems particularly nice. Implemented in https://github.com/proteneer/timemachine/pull/987/commits/9a5f0defe9b59cb04ce34feee4184bdd1665b351

tests/test_mle.py
badisa: Should we convert the labels to strings and handle string -> index conversion in the function then?

mcwitt: Sorry, not sure I understand. This test checks for consistency with the existing `infer_node_values_and_errs`, which doesn't accept string labels. We could add a separate test to verify that string labels are handled correctly (but doing conversion from string labels to indices in the test function seems like it would replicate a large part of the functionality we're testing in the test implementation).

badisa: In practice the nodes will identified by strings rather than integers. Want to make sure that this is tested, or explicit that the node labels need to be integers.

badisa (approved):

tests/test_mle.py
mcwitt: OK, I think I've come up with a good way to test this (check that arbitrary relabeling of the nodes doesn't affect the result) in https://github.com/proteneer/timemachine/pull/987/commits/8f1df0e971e29f9b9e3ab51abe29bfe2616db5c0

timemachine/fe/mle.py
```diff
@@ -212,3 +214,64 @@ def infer_node_vals_and_errs(
...
+    edge_diff_prop: str
+        Edge property to use for differences
+    edge_stddev_prop: str
+        Edge property to use for standard deviations
+    node_val_prop: str
+        Node property to use for absolute values
+    node_stddev_prop: str
+        Node property to use for standard deviations. If missing in reference nodes, assume zero.
+    ref_nodes: sequence
+        Reference nodes (must have properties ref_node_val_prop and ref_node_stddev_prop)
```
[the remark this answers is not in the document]
mcwitt: Ah, thanks. Fixed in https://github.com/proteneer/timemachine/pull/987/commits/51c8097c344ad957a14845df40c912b255b1cba1

> Uncertainty: Is it preferable to overwrite `node_val_prop` on `ref_nodes` (as currently implemented), or to keep these properties separate (i.e. read some `ref_node_*_props` from `ref_nodes`, but write MLE results to different property names `node_*_props`)?

I'm inclined to overwrite in the result for simplicity and since we're returning a copy, so the original values will still be accessible. Don't have a terribly strong preference, though.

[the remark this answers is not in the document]
mcwitt: As discussed offline, we have some use cases where we want to avoid overwriting reference properties due to potential for causing confusion. Added separate arguments for `node_*_prop` and `ref_node_*_prop` in https://github.com/proteneer/timemachine/pull/987/commits/1c02d73eb7db1b9b170d04bb64f7fcb42565b861

tests/test_mle.py
mcwitt: Added test for missing edge values in https://github.com/proteneer/timemachine/pull/987/commits/dbeafa92ec3a7313e1b6265fb8aa33d33272f37c and implementation in https://github.com/proteneer/timemachine/pull/987/commits/be601a2ce5e5db904e90eb81b9c5e23992946e77

maxentile (approved):

badisa (approved):

timemachine/fe/mle.py
```diff
@@ -212,3 +214,103 @@ def infer_node_vals_and_errs(
...
+        See documentation for :py:func:`timemachine.fe.mle.infer_node_vals_and_errs`
+
+    Returns
+    -------
+    networkx.Graph
+        Graph where all nodes have been labeled with the inferred value of `node_val_prop` and `node_stddev_prop`.
+    """
+
+    def get_valid_subgraph(g: nx.Graph) -> nx.Graph:
+        "Remove edges with missing edge_diff_prop or edge_stddev_prop; then, remove any isolated nodes"
```
maxentile, line 261: super nit: triple quotes?

timemachine/fe/mle.py
```diff
@@ -212,3 +214,103 @@ def infer_node_vals_and_errs(
     dg_err = bootstrap_estimates.std(0)
 
     return dg, dg_err
+
+
+def infer_node_vals_and_errs_networkx(
+    nx_graph: nx.Graph,
```
badisa, line 220: Does that work expectedly with a DiGraph vs nx.Graph? I believe the subsetting might behave differently with those two cases.

timemachine/fe/mle.py
```diff
@@ -212,3 +214,103 @@ def infer_node_vals_and_errs(
...
+
+    def get_valid_subgraph(g: nx.Graph) -> nx.Graph:
+        "Remove edges with missing edge_diff_prop or edge_stddev_prop; then, remove any isolated nodes"
+        sg = g.copy()
+
+        missing_edges = [e for e, v in sg.edges.items() if not v.get(edge_diff_prop) or not v.get(edge_stddev_prop)]
+        sg.remove_edges_from(missing_edges)
+
+        isolated_nodes = list(nx.isolates(sg))
+        sg.remove_nodes_from(isolated_nodes)
```
maxentile, line 268: Uncertainty: Is "nodes with zero degree (`nx.isolates`) the intended concept?

(May want to remove nodes that aren't in the largest connected component, even if they have degree > 0)

timemachine/fe/mle.py
```diff
@@ -212,3 +214,103 @@ def infer_node_vals_and_errs(
...
+
+def infer_node_vals_and_errs_networkx(
+    nx_graph: nx.Graph,
+    edge_diff_prop: str,
+    edge_stddev_prop: str,
+    node_val_prop: str,
+    node_stddev_prop: str,
+    ref_nodes: Sequence[Any],
+    ref_node_val_prop: str,
+    ref_node_stddev_prop: str,
```
maxentile, line 227: opinionated super nit: may consider reordering these so that "read only" things are grouped and appear first `edge_diff_prop, edge_stddev_prop, ref_nodes, ref_node_val_prop, ref_node_stddev_prop`, and then the writable properties appear second, possibly with defaults `node_val_prop = "mle_dg", node_stddev_prop = "mle_dg_err"` or similar. (haven't thought carefully about the specific default names)

timemachine/fe/mle.py
maxentile: Handling missing edges seems more subtle -- would it be reasonable to split this into a separate PR?

I'm uncertain about the right behavior when edges are missing:
* keep all nodes with degree > 0 vs. keep largest connected component vs. keep component that contains all ref_nodes
* throw error vs. not if some ref_nodes are in different disconnected components
* return copy of full graph where some nodes may not have mle attributes vs. return copy of subgraph where all nodes have mle attributes...

For now, perhaps this function can assert as a pre-condition that all edges in the graph have valid values for `edge_diff_prop`?

mcwitt: > Uncertainty: Is "nodes with degree zero" (nx.isolates) the intended concept?

It was intended; this is to prevent the assertion https://github.com/proteneer/timemachine/blob/9ac44c207b21750f15c5fc3b8156e5516e131438/timemachine/fe/mle.py#L51 from failing in the case where removing edges results in some nodes that don't appear in the edge list. It seemed more appropriate to (eventually) handle the case of multiple connected components in `infer_node_vals_and_errs`.

maxentile: https://github.com/proteneer/timemachine/pull/987#discussion_r1154830927

Ahh, sorry, I forgot most of these are already ruled out by https://github.com/proteneer/timemachine/blob/9cb5204566c06461653dc83aec031d56ddc47d1d/timemachine/fe/mle.py#L52-L54  -- thanks for the reminder @mcwitt .

I think this looks good.

maxentile: (The "disconnected components but node degree > 0" case is already ruled out -- the current PR only needs to handle the new case now covered by `nx.isolates`, as implemented.)

timemachine/fe/mle.py
mcwitt: I like this suggestion. Applied in https://github.com/proteneer/timemachine/pull/987/commits/6c2a41add7f35017eb05520e82679172b502acf7, using `"inferred_dg"` and `"inferred_dg_stddev"` (open to alternatives).

timemachine/fe/mle.py
mcwitt: Changed in https://github.com/proteneer/timemachine/pull/987/commits/47c8325bcebb9e218c7118a6444e2a234c728cfb

timemachine/fe/mle.py
mcwitt: You're right, it should be `DiGraph`. Changed in https://github.com/proteneer/timemachine/pull/987/commits/0e30af40ed9ce764a75df5ac27289d54ab09c6d3

PR #988, opened 2023-04-03
Deboggle: Update python/library versions

jkausrelay:
- Python 3.10.10 (3.11 has dependency issues)
- Jax/jaxlib 0.4.8
- numpy 1.23.5 (1.24 causes issue w/pymbar 3)
- scipy 1.10.1
- openmm 8.0.0
- openeye-toolkits 2022.2.2
- removed importlib-resources, typing-extensions
- mypy to 1.1.1 to fix check in numpy (https://github.com/python/mypy/issues/13627)


tests/test_jax_nonbonded.py
```diff
@@ -120,6 +120,7 @@ def generate_waterbox_nb_args() -> NonbondedArgs:
     conf = positions.value_in_unit(unit.nanometer)
 
     N = conf.shape[0]
+    print()
```
badisa, line 123: Remove?

timemachine/ff/handlers/openmm_deserializer.py
```diff
@@ -200,8 +204,17 @@ def deserialize_system(system: mm.System, cutoff: float) -> Tuple[List[potential
...
-            bps.append(potentials.Nonbonded(N, exclusion_idxs, scale_factors, beta, cutoff).bind(nb_params))
+            bps_dict["Nonbonded"].append(
+                potentials.Nonbonded(N, exclusion_idxs, scale_factors, beta, cutoff).bind(nb_params)
+            )
 
             # nrg_fns.append(('Exclusions', (exclusion_idxs, scale_factors, es_scale_factors)))
 
+    # ugh, ... various parts of our code assume the bps are in a certain order
+    # so put them back in that order here
+    bps = []
```
badisa, line 215: What changed the ordering? Did a data structure change its ordering between python 3.7 and 3.11?

badisa: Oh its probably the openmm update??

badisa (approved):
🚀

environment.yml
```diff
@@ -3,7 +3,7 @@ channels:
   - conda-forge
   - openeye
 dependencies:
-  - python=3.7
-  - jax=0.3.14
-  - jaxlib=0.3.14=cpu*
-  - openeye-toolkits=2020.2.0
+  - python=3.10.10
+  - jax=0.4.7
```
mcwitt, line 7: nit: any reason not to upgrade jax to latest ([0.4.8](https://github.com/google/jax/releases/tag/jax-v0.4.8))?

timemachine/ff/handlers/openmm_deserializer.py
mcwitt: I think this makes sense for this PR. Longer term, seems like we might want to use a dataclass instead of a list?

mcwitt (commented):
> removed importlib-resources, typing-extensions

Looks like setup.py might need to be updated?

mcwitt (changes requested):
Pending update of setup.py to remove importlib-resources, typing-extensions

timemachine/ff/handlers/openmm_deserializer.py
jkausrelay: Yeah, it's the openmm update. I agree, a better data structure could be used in the future.

mcwitt (approved):

environment.yml
jkausrelay: 0.4.8 drops 11.7 CUDA support and I didn't want to also require a cuda update with this change.

setup.py
```diff
@@ -96,18 +96,16 @@ def build_extension(self, ext):
     keywords="molecular dynamics",
     ext_modules=ext_modules,
     packages=find_packages(),
-    python_requires=">=3.7",
+    python_requires=">=3.10",
     install_requires=[
-        "importlib-resources",
         "jax",
-        "jaxlib>0.3.13",
+        "jaxlib>0.4.6",
```
mcwitt, line 102: nit: is there a known issue with 0.4.6 and before?

jkausrelay: No known issue, reverted change.

setup.py
```diff
@@ -96,9 +96,8 @@ def build_extension(self, ext):
     keywords="molecular dynamics",
     ext_modules=ext_modules,
     packages=find_packages(),
-    python_requires=">=3.7",
+    python_requires=">=3.10",
     install_requires=[
-        "importlib-resources",
         "jax",
         "jaxlib>0.3.13",
```
badisa, line 102:
```suggestion
        "jaxlib>=0.4.1",
```

One of the reasons for upgrading was to use new features like the jax.Array type, so presumably we will want at least the version that introduced it?

mcwitt: Agreed, but would it make sense to add the constraint in the PR that introduces usage of `jax.Array`?

environment.yml
mcwitt: Oh, but we use CPU-only jax though, right? Does it still somehow require CUDA to be present?

jkausrelay: Good point, updated to 0.4.8

jkausrelay: Err, 0.4.8 is not yet in conda, back to 0.4.7

mcwitt: Hmm, I don't think jaxlib 0.4.8 has been released yet? https://github.com/google/jax/releases

(IIUC jax and jaxlib do not need to be the same version, as long as jaxlib satisfies jax's lower bound)

timemachine/ff/__init__.py
```diff
@@ -87,21 +87,18 @@ def load_from_file(cls, path_or_str: Union[str, Path]) -> "Forcefield":
...
-                raise ValueError(f"Unable to find {original_path} in file system or built-in forcefields")
-            with open(path, "r") as ifs:
-                handlers, protein_ff, water_ff = deserialize_handlers(ifs.read())
-
+        # Look for builtin ff
+        for resource in resources.files("timemachine.ff.params").iterdir():
+            if resource.name == path.name:
+                warn(f"Provided path {original_path} shares name with built-in forcefield, falling back to built-in")
+                handlers, protein_ff, water_ff = deserialize_handlers(resource.read_text())
+                return cls.from_handlers(handlers, protein_ff=protein_ff, water_ff=water_ff)
```
mcwitt, line 95: with resources.as_file(resources.files("timemachine.ff.params") / path.name) as rpath:
            if rpath.exists():
                warn(f"Provided path {original_path} shares name with built-in forcefield, falling back to built-in")
                handlers, protein_ff, water_ff = deserialize_handlers(rpath.read_text())
                return cls.from_handlers(handlers, protein_ff=protein_ff, water_ff=water_ff)

timemachine/ff/__init__.py
```diff
@@ -88,20 +87,18 @@ def load_from_file(cls, path_or_str: Union[str, Path]) -> "Forcefield":
...
-            if not path.is_file():
-                raise ValueError(f"Unable to find {original_path} in file system or built-in forcefields")
-            with open(path, "r") as ifs:
-                handlers, protein_ff, water_ff = deserialize_handlers(ifs.read())
+        # Look for builtin ff
+        for resource in resources.files("timemachine.ff.params").iterdir():
+            if resource.name == path.name:
+                warn(f"Provided path {original_path} shares name with built-in forcefield, falling back to built-in")
+                handlers, protein_ff, water_ff = deserialize_handlers(resource.read_text())
+                return cls.from_handlers(handlers, protein_ff=protein_ff, water_ff=water_ff)
```
mcwitt, line 95:
```suggestion
        with resources.as_file(resources.files("timemachine.ff.params") / path.name) as rpath:
            if rpath.exists():
                warn(f"Provided path {original_path} shares name with built-in forcefield, falling back to built-in")
                handlers, protein_ff, water_ff = deserialize_handlers(rpath.read_text())
                return cls.from_handlers(handlers, protein_ff=protein_ff, water_ff=water_ff)
```

nit: use `resources.as_file` to avoid iteration?

PR #989, opened 2023-04-05
Add Forcefield.load_default()

maxentile:
Add (and test) alias for `Forcefield.load_from_file(DEFAULT_FF)` (suggested by @mcwitt  in https://github.com/proteneer/timemachine/pull/934#discussion_r1072752247 ), updating all usages in timemachine/ and tests/ .

### DIFF
--- a/tests/test_forcefields.py
+++ b/tests/test_forcefields.py
@@ -63,3 +63,21 @@ def test_loading_forcefield_from_file():
             with pytest.raises(ValueError, match="Unsupported charge handler"):
                 Forcefield.load_from_file(bad_ff)
         assert len(w) == 0
+
+
+def test_load_default():
+    """assert that load_default() is an alias for load_from_file(DEFAULT_FF)"""
+
+    ref = Forcefield.load_from_file(constants.DEFAULT_FF)
+    test = Forcefield.load_default()
+
+    # ref == test evaluates to false, so assert equality of smirks lists and parameter arrays manually
+
+    ref_handles = ref.get_ordered_handles()
+    test_handles = test.get_ordered_handles()
+
+    assert len(ref_handles) == len(test_handles)
+
+    for (ref_handle, test_handle) in zip(ref.get_ordered_handles(), test.get_ordered_handles()):
+        assert ref_handle.smirks == test_handle.smirks
+        assert (ref_handle.params == test_handle.params).all()
--- a/tests/test_forcefields.py
+++ b/tests/test_forcefields.py
@@ -3,6 +3,7 @@ from glob import glob
 from pathlib import Path
 from warnings import catch_warnings
 
+import numpy as np
 import pytest
 from common import temporary_working_dir
 
@@ -80,4 +81,4 @@ def test_load_default():
 
     for (ref_handle, test_handle) in zip(ref.get_ordered_handles(), test.get_ordered_handles()):
         assert ref_handle.smirks == test_handle.smirks
-        assert (ref_handle.params == test_handle.params).all()
+        np.testing.assert_array_equal(ref_handle.params, test_handle.params)
--- a/timemachine/ff/__init__.py
+++ b/timemachine/ff/__init__.py
@@ -5,7 +5,7 @@ from warnings import warn
 
 import importlib_resources as resources
 
-from timemachine.constants import DEFAULT_PROTEIN_FF, DEFAULT_WATER_FF
+from timemachine.constants import DEFAULT_FF, DEFAULT_PROTEIN_FF, DEFAULT_WATER_FF
 from timemachine.ff.handlers import bonded, nonbonded
 from timemachine.ff.handlers.deserialize import deserialize_handlers
 from timemachine.ff.handlers.serialize import serialize_handlers
@@ -103,6 +103,11 @@ class Forcefield:
                 handlers, protein_ff, water_ff = deserialize_handlers(ifs.read())
 
         return cls.from_handlers(handlers, protein_ff=protein_ff, water_ff=water_ff)
+
+    @classmethod
+    def load_default(cls) -> "Forcefield":
+        """alias for load_from_file(DEFAULT_FF)"""
+        return cls.load_from_file(DEFAULT_FF)
 
     @classmethod
     def from_handlers(
--- a/timemachine/ff/__init__.py
+++ b/timemachine/ff/__init__.py
@@ -72,13 +72,13 @@ class Forcefield:
         Parameters
         ----------
 
-        path: string or pathlib.Path
+        path_or_str: string or pathlib.Path
             Either the filename of a built in ff (smirnoff_1_1_0_sc.py) or a path to a new forcefield file
 
         Returns
         -------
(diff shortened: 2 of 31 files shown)


tests/test_forcefields.py
```diff
@@ -63,3 +63,21 @@ def test_loading_forcefield_from_file():
...
+    # ref == test evaluates to false, so assert equality of smirks lists and parameter arrays manually
+
+    ref_handles = ref.get_ordered_handles()
+    test_handles = test.get_ordered_handles()
+
+    assert len(ref_handles) == len(test_handles)
+
+    for (ref_handle, test_handle) in zip(ref.get_ordered_handles(), test.get_ordered_handles()):
+        assert ref_handle.smirks == test_handle.smirks
+        assert (ref_handle.params == test_handle.params).all()
```
mcwitt, line 83: nit: should we prefer `np.testing.assert_array_equal`? (mainly for more helpful message on failure)

timemachine/ff/__init__.py
```diff
@@ -5,7 +5,7 @@
 
 import importlib_resources as resources
 
-from timemachine.constants import DEFAULT_PROTEIN_FF, DEFAULT_WATER_FF
+from timemachine.constants import DEFAULT_FF, DEFAULT_PROTEIN_FF, DEFAULT_WATER_FF
```
mcwitt, line 8: super-nit, arguable: would it make sense for `DEFAULT_FF` to be defined here?

tests/test_forcefields.py
maxentile: Yes! Fixed in https://github.com/proteneer/timemachine/pull/989/commits/d5476858efc993ca96e7ac581c1e5583fc0a03a8

timemachine/ff/__init__.py
maxentile: Possibly.

I think it also makes sense as-is, and it is nice to have the top-level collection of defaults in `constants.py`... It doesn't seem too-cluttered as-is.

maxentile: Reviewed the contents of `constants.py`, and I think it still seems reasonable to me to leave DEFAULT_FF there https://github.com/proteneer/timemachine/pull/989/commits/4213c4cb1a47e7ad7cabb0d1e7f16869806d1919

However, what does look out-of-place in constants.py is these tags https://github.com/proteneer/timemachine/pull/989/commits/4213c4cb1a47e7ad7cabb0d1e7f16869806d1919#diff-58fa561ba38bbef66f3498a7dc546f317f49a1fb35ec5d5b667aebe9b8d74891R24-R26 , which look like an internal detail that doesn't need to be referenced much globally...

maxentile: These constants are only referenced in two files, `ff/handlers/serialize.py` and `ff/handlers/deserialize.py`.

(Moved these from top-level `constants.py` to a new file `ff/handlers/serialization_format.py` in https://github.com/proteneer/timemachine/pull/989/commits/075438b358637eaa88b43182b2b0c5cb789d58cb)

timemachine/ff/handlers/deserialize.py
```diff
@@ -21,8 +21,8 @@ def deserialize_handlers(obj):
 
     handlers = []
 
-    protein_ff = obj_dict.pop(constants.PROTEIN_FF_TAG, constants.DEFAULT_PROTEIN_FF)
-    water_ff = obj_dict.pop(constants.WATER_FF_TAG, constants.DEFAULT_WATER_FF)
+    protein_ff = obj_dict.pop(serialization_format.PROTEIN_FF_TAG, constants.DEFAULT_PROTEIN_FF)
+    water_ff = obj_dict.pop(serialization_format.WATER_FF_TAG, constants.DEFAULT_WATER_FF)
```
maxentile, line 25: Small surprise / possible code-smell: `deserialize_handlers()` hard-codes dependence on DEFAULT_PROTEIN_FF and DEFAULT_WATER_FF, but `serialize_handlers(..., protein_ff, water_ff)` exposes these as parameters.

badisa: Could add the default values when there are no protein/water specified in the handlers in the signature? Otherwise this seems reasonable to me, if there is no water/protein ff specified fall back to the defaults. We could also put explicit water/protein ffs in our default FFs

badisa (approved):

timemachine/ff/handlers/deserialize.py
maxentile: Added in https://github.com/proteneer/timemachine/pull/989/commits/171c17ce29f30f26a6e839046fe49c68eaa13055

timemachine/ff/handlers/deserialize.py
```diff
@@ -5,12 +5,13 @@
 from timemachine.ff.handlers.suffix import _SUFFIX
 
 
-def deserialize_handlers(obj):
+def deserialize_handlers(obj, protein_ff_name=constants.DEFAULT_PROTEIN_FF, water_ff_name=constants.DEFAULT_WATER_FF):
```
badisa, line 8: You should make it clear that these are the ffs that this will fall back to if they are not specified in the object being deserialized.

maxentile: Sorry, I don't understand how this code works well enough to be confident in the most recent change https://github.com/proteneer/timemachine/pull/989/commits/171c17ce29f30f26a6e839046fe49c68eaa13055 or write an accurate description of this fallback logic -- will revert for now, and possibly revisit.

timemachine/ff/handlers/deserialize.py
maxentile: Reverted in https://github.com/proteneer/timemachine/pull/989/commits/be35ee1d02559ed6680f1246d308ef05b2268d39 to avoid messing things up, following https://github.com/proteneer/timemachine/pull/989#discussion_r1158931631

timemachine/ff/handlers/deserialize.py
badisa: For posterity.

When a forcefield (smirnoff_*.py) file is deserialized, this code will look for the relevant tag protein/water in the dictionary object. If the tag is not there, it will fall back to the values specified by the defaults.

Currently you can't control what the FF should default to, besides modifying the file format or the Forcefield object fields. The change you had reverted would have allowed users to specify a default of their own, though if the file being deserialized had defined a water/protein ff already it would have done nothing.

badisa: It seems like we should prefer `Forcefield.serialize` or modify `serialize_handlers` to take a ff rather than handlers + other parameters, as the Forcefield is really what users care about. Is a bit confusing as it stands.

maxentile: That makes sense -- thanks for explaining!

Sounds like a useful change to be picked up in a separate PR, but want to do this carefully to avoid introducing new possibilities for mismatches.

maxentile:
Thanks for the reviews! Merging

PR #992, opened 2023-04-06
Remove unused params in Protein Builder

badisa:


badisa:
Didn't update the attic references, not sure if we want to do that.

jkausrelay:
> Didn't update the attic references, not sure if we want to do that.

We shouldn't update attic. Otherwise it's just code we need to maintain that's untested.

jkausrelay (approved):

PR #995, opened 2023-04-07
Unify potential interface

mcwitt:
Overhaul of the potentials API, unifying `timemachine.lib.potentials` and `timemachine.potentials.generic`.

### `Potential` objects are callable; directly calling them invokes the Python reference implementation:

- Before:

    ```python
    from timemachine.potentials.generic import HarmonicBond

    u_ref = HarmonicBond(bond_idxs).to_reference()
    energy = u_ref(conf, params, box)
    du_dx, du_dp = jax.grad(u_ref, argnums=(0, 1))(coords, params, box)
    ```

- Now:

    ```python
    from timemachine.potentials import HarmonicBond

    u_ref = HarmonicBond(bond_idxs)
    energy = u_ref(conf, params, box)
    du_dx, du_dp = jax.grad(u_ref, argnums=(0, 1))(coords, params, box)
    ```

### New `BoundPotential` type for potentials partially applied to parameters:

- Before:

    ```python
    from timemachine.lib.potentials import HarmonicBond

    HarmonicBond(bond_idxs)               # type: HarmonicBond
    HarmonicBond(bond_idxs).bind(params)  # type: HarmonicBond
    ```

- Now:

    ```python
    from timemachine.potentials import HarmonicBond

    HarmonicBond(bond_idxs)               # type: HarmonicBond
    HarmonicBond(bond_idxs).bind(params)  # type: BoundPotential[HarmonicBond]
    ```

### GPU implementations are wrapped to be JAX-differentiable:

- Before:

    ```python
    from timemachine.potentials.generic import HarmonicBond

    gpu_impl = HarmonicBond(bond_idxs).to_gpu().unbound_impl(np.float32)
    du_dx, du_dp, u = gpu_impl.unbound_impl.execute_selective(x, params, box, True, True, True)
    ```

- Now:

    ```python
    from timemachine.potentials import HarmonicBond

    u_fn = HarmonicBond(bond_idxs).to_gpu(np.float32)
    u, (du_dx, du_dp) = jax.value_and_grad(u_fn, argnums=(0, 1))(coords, params, box)
    ```
    
Notes:
  - This also makes the differentiable interface composable, i.e. JAX will handle the chain rule for us if `u_fn` is a complicated expression involving multiple potentials
  - The unwrapped versions are still exposed, so the former usage is possible in situations where we want precise control over what is evaluated (see below)
    
### `construct_differentiable_interface_fast` is no longer necessary

- Before:

    ```python
    from timemachine.fe.functional import construct_differentiable_interface_fast
    
    potentials: List[CustomOpWrapper]
    u_fn = construct_differentiable_interface_fast(potentials, params, np.float32)
    u, (du_dx, du_dp) = jax.value_and_grad(u_fn, argnums=(0, 1))(coords, params, box)
    ```

- Now:

    ```python
    from timemachine.potentials import SummedPotential
    
    potentials: List[Potential]

    # Note: call_with_params_list provides an interface that automatically handles flattening lists of parameter sets to interface with custom_ops.SummedPotential
    u_fn = SummedPotential(potentials, params).to_gpu(np.float32).call_with_params_list

    u, (du_dx, du_dp) = jax.value_and_grad(u_fn, argnums=(0, 1))(coords, params, box)
    ```
    
Alternatively, we can do the summation on the Python side (analogous to `construct_differentiable_interface`) and still get derivatives:

```python
gpu_impls = [p.to_gpu(np.float32) for p in potentials]

def u_fn(potentials, params):
    return jnp.sum(jnp.array([f(conf, params_, box) for f, params_ in zip(gpu_impls, params)]))

u, (du_dx, du_dp) = jax.value_and_grad(u_fn, argnums=(0, 1))(coords, params, box)
```
    
### Access to the bare C++ objects is available via `GpuImplWrapper.unbound_impl` or `BoundGpuImplWrapper.bound_impl`:

- Before:

    ```python
    from timemachine.potentials.generic import HarmonicBond

    unbound_gpu_impl = HarmonicBond(bond_idxs).to_gpu().unbound_impl(np.float32)
    bound_gpu_impl = HarmonicBond(bond_idxs).to_gpu().bind(params).bound_impl(np.float32)
    ```
    
- Now:

    ```python
    from timemachine.potentials import HarmonicBond

    unbound_gpu_impl = HarmonicBond(bond_idxs).to_gpu(np.float32).unbound_impl

    # the following are equivalent
    bound_gpu_impl = HarmonicBond(bond_idxs).bind(params).to_gpu(np.float32).bound_impl
    bound_gpu_impl = HarmonicBond(bond_idxs).to_gpu(np.float32).bind(params).bound_impl
    ```

## Other changes

- Removes `disable_hilbert_sort`, `set_nblist_padding` methods from C++ nonbonded potentials. Adds corresponding constructor arguments (https://github.com/proteneer/timemachine/pull/995/commits/b5e632b407d7e9a3b74271ce8aacdb0593183197).

- Fixes inconsistency in `VacuumSystem.get_U_fn` where we'd always been using `HarmonicAngle` even when the system was parameterized with `HarmonicAngleStable` 

## Todos

- [ ] Update nightly tests


tests/common.py
```diff
@@ -138,56 +138,11 @@ def prepare_nb_system(
...
-    harmonic_bond_fn = functools.partial(bonded.harmonic_bond, box=None, bond_idxs=bond_idxs)
-
-    # custom_angles = potentials.HarmonicAngle(angle_idxs, angle_param_idxs, precision=precision)
-    # harmonic_angle_fn = functools.partial(bonded.harmonic_angle, box=None, angle_idxs=angle_idxs, param_idxs=angle_param_idxs)
-
-    # custom_torsions = potentials.PeriodicTorsion(torsion_idxs, torsion_param_idxs, precision=precision)
-    # periodic_torsion_fn = functools.partial(bonded.periodic_torsion, box=None, torsion_idxs=torsion_idxs, param_idxs=torsion_param_idxs)
-
-    return (bond_params, harmonic_bond_fn), custom_bonded
-    # return params, [harmonic_bond_fn, harmonic_angle_fn, periodic_torsion_fn], [custom_bonded, custom_angles, custom_torsions]
```
mcwitt, line 188: Unused and apparently using a very old API; didn't seem worth moving to attic

tests/common.py
```diff
@@ -253,27 +208,14 @@ def assert_equal_vectors(self, truth, test, rtol):
     def compare_forces(
```
mcwitt, line 208: Simplified `compare_forces` to accept an unbound GPU impl (rather than constructing it in the function), removing the need to loop over parameter sets for efficiency.

tests/nonbonded/test_nonbonded_jit.py
```diff
@@ -2,23 +2,22 @@
...
         N,
         exclusion_idxs=jnp.zeros((0,)),
         scale_factors=jnp.zeros((0, 1)),
         beta=1.0,
         cutoff=0.1,
     )
 
-    U_ref = potential.to_reference()
-    U_ref_jit = jax.jit(U_ref)
+    U_ref_jit = jax.jit(U_ref.__call__)
```
mcwitt, line 20: Slightly unfortunate; `jit` does not seem to work on `U_ref` by itself (one workaround is to add explicit `__call__`)

timemachine/potentials/potentials.py
```diff
@@ -0,0 +1,260 @@
...
+    idxs: NDArray[np.int32]
+    rescale_mask: NDArray[np.float64]
+    beta: float
+    cutoff: float
+
+    def __call__(self, conf: Conf, params: Params, box: Optional[Box]) -> float | Array:
+        vdW, electrostatics = nonbonded.nonbonded_on_specific_pairs(
+            conf, params, box, self.idxs, self.beta, self.cutoff, self.rescale_mask
+        )
+        return -jnp.sum(vdW) - jnp.sum(electrostatics)
```
maxentile, line 198: Opinionated super-nit: apply negation once `U = jnp.sum(vdW) + jnp.sum(electrostatics); return -U`

timemachine/lib/potentials.py
```diff
@@ -1,324 +0,0 @@
...
-
-
-class NonbondedPairListPrecomputed(CustomOpWrapper):
-    """
-    This implements a pairlist with precomputed parameters. It differs from the regular NonbondedPairlist in that it
-    expects params of the form s0*q_ij, s_ij, s1*e_ij, and w_offsets_ij, where s are the scaling factors and combining
-    rules have already been applied.
-
-    Note that you should not use this class to implement exclusions (that are later cancelled out by AllPairs)
-    since the floating point operations are different in python vs C++.
```
maxentile, line 298: Possibly lossy step: I didn't immediately see if the content of this docstring was retained in the code that supersedes this.

tests/test_md.py
```diff
@@ -661,7 +661,7 @@ def build_context(barostat_interval):
             weak_refs = []
             bps = []
             for p, bp in zip(sys_params, unbound_potentials):
-                bound_impl = bp.bind(p).bound_impl(np.float32)
+                bound_impl = bp.bind(p).to_gpu(np.float32).bound_impl
```
maxentile, line 664: Small surprise, not immediately clarified from inspecting `to_gpu` definition -- does `to_gpu(precision)` eagerly instantiate both bound and unbound impls? (so that the `.bound_impl` attribute is populated / ready to be referenced in this way?)

maxentile (commented):
This looks like a great simplification, and the before/after usage examples are 🔥.

(I've left a few small drive-by comments, but haven't completed review, and haven't come up with any major feedback yet.)

timemachine/lib/potentials.py
mcwitt: Good catch. I'll go back over this file and move over any relevant comments.

timemachine/fe/system.py
```diff
@@ -72,23 +82,27 @@ def simulate_system(U_fn, x0, num_samples=20000, steps_per_batch=500, num_worker
...
+            torsion = bp
+        elif isinstance(bp.potential, potentials.Nonbonded):
+            nonbonded = bp
         else:
             assert 0, "Unknown potential"
 
-    return system
+    assert bond
+    assert angle
+    assert nonbonded
```
badisa, line 103: assert torsion?

tests/test_md.py
mcwitt: > does to_gpu(precision) eagerly instantiate both bound and unbound impls?

No; there are two different methods: `Potential.to_gpu` and `BoundPotential.to_gpu` (the `.bind(p)` call turns a Potential into a BoundPotential. But the naming here seems confusing (using `bp` to refer to an (unbound) Potential).

Stepping back, it seems like there might be a mixed convention where "bound potential" sometimes refers to (1) a potential (reference or GPU) that has been partially applied to parameters, and sometimes to (2) a GPU implementation of a potential, whether or not it's an instance of `Potential` or `BoundPotential`. In this PR I changed some of the cases of (2) to (1), but there is still more work to do to make it consistent (first step is probably to agree on a convention).

timemachine/potentials/potentials.py
mcwitt: I prefer your suggestion too. Willfix.

timemachine/potentials/potentials.py
```diff
@@ -0,0 +1,260 @@
...
+        sizes = [ps.size for ps in self.params_init]
+        return SummedPotentialGpuImplWrapper(custom_ops.SummedPotential(impls, sizes))
+
+
+@dataclass
+class SummedPotentialGpuImplWrapper(GpuImplWrapper):
+    """Handles flattening parameters before passing to kernel to provide a nicer interface"""
+
+    def call_with_params_list(self, conf: Conf, params: Sequence[Params], box: Box) -> float:
+        params_flat = jnp.concatenate([ps.reshape(-1) for ps in params])
```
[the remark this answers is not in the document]
mcwitt: > Question: If differentiated w.r.t. params, will it return Sequence[Params] or a flat Params?

It returns a `Sequence[Params]` (I think this is a consequence of the type signature of `jax.grad`). It's tested here: https://github.com/proteneer/timemachine/blob/24fe528245040c8ed2c81f1df8485ffccd102eb1/tests/test_potentials_differentiable_interface.py#L12)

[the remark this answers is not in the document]
mcwitt: Impression: Useful to provide this option, and syntax looks intuitive, but initial reaction is surprise at lack of parentheses in usages like `u = sp.to_gpu(prec).call_with_params_list; ... ; u(conf, params, box)`

Hmm, I can sympathize with this impression, though I'm not sure how to improve it. Note that this is equivalent to

```python
u = sp.to_gpu(prec).call_with_params_list(conf, params, box)
```

timemachine/potentials/summed.py
```diff
@@ -41,3 +38,28 @@ def summed_potential(
...
+    params: array (P,)
+        flattened array of parameters shared by each potential term
+
+    box: array (3, 3)
+        periodic box
+
+    U_fns: list of functions with signature (conf, params, box) -> energy
+        potential terms
+    """
+    return sum(U_fn(conf, ps, box) for U_fn, ps in zip(U_fns, jnp.array(params)))
```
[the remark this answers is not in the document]
mcwitt: I'm not aware of any significant difference. I guess I was thinking that since this is just summing floats (vs. arrays), it would be simpler to use the stdlib function. On second thought though, I wonder if there might be a difference when we jax-transform the output. Might change to `jnp.sum(jnp.array(...))` for better consistency with existing code.

timemachine/potentials/jax_utils.py
```diff
@@ -210,7 +210,7 @@ def bernoulli_logpdf(log_p_i, z_i) -> float:
     b = jnp.array([jnp.ones(n), -jnp.ones(n)])
     log_1_minus_p_i = logsumexp(a=a, b=b, axis=0)
 
-    return jnp.sum(jnp.where(z_i, log_p_i, log_1_minus_p_i))
+    return cast(float, jnp.sum(jnp.where(z_i, log_p_i, log_1_minus_p_i)))
```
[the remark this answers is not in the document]
mcwitt: This was causing a mypy error, since the return type of `jnp.sum` is `jax.Array | float`. I was thinking it was a result of the jax upgrade, but am not sure now. Will double check.

timemachine/fe/system.py
mcwitt: Torsion is optional in `VacuumSystem`, which is why it isn't asserted here (we have some code that constructs `VacuumSystem`s with `torsion=None`; I'm not sure if it's necessary, will double-check)

tests/common.py
```diff
@@ -1,15 +1,14 @@
...
-import functools
 import itertools
 import os
 import unittest
+from collections.abc import Iterator
 from dataclasses import dataclass
 from importlib import resources
-from tempfile import NamedTemporaryFile, TemporaryDirectory
-from typing import Iterable, Optional
+from tempfile import NamedTemporaryFile, TemporaryDirectory, _TemporaryFileWrapper
```
badisa, line 8: Is there a reason to import use private class rather than NamedTemporaryFile?

mcwitt: This was to work around a mypy error:

> error: Function "tempfile.NamedTemporaryFile" is not valid as a type  [valid-type]

(relevant: https://stackoverflow.com/questions/64429113/how-should-a-namedtemporaryfile-be-annotated)

We don't run mypy on tests/ in CI, so could revert this or just add `# type: ignore`

badisa: `# type: ignore` would be my preference. Just don't want to rely on private variables

timemachine/potentials/potentials.py
mcwitt: Changed in https://github.com/proteneer/timemachine/pull/995/commits/df34ffeea9136d3d7081ac472f59407a492c03ae

timemachine/potentials/summed.py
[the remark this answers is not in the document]
mcwitt: Changed to `jnp.sum(jnp.array(...))` for consistency in https://github.com/proteneer/timemachine/pull/995/commits/97526fafcbce8ef1fc4e18464024ba185cb8232a

tests/common.py
mcwitt: Fixed in https://github.com/proteneer/timemachine/pull/995/commits/811f290ac54985e41fde4950201d01c96723a358

timemachine/lib/potentials.py
mcwitt: Restored docstring in the new location in https://github.com/proteneer/timemachine/pull/995/commits/b74413f7f885973f1eb87c26fb84589e2d903105

badisa (approved):

tests/test_md.py
maxentile: Thanks for clarifying! This diagnosis is helpful.

Regarding (2), I think previously we could use `bp.bound_impl()` or `bp.unbound_impl()`, but now bound potentials only support `bp.bound_impl`, unbound potentials only support `ubp.unbound_impl`, if I'm understanding correctly.

I think that makes sense as-is, but in a future refinement we might want to consider
(A) adding a method to `.unbind()` (`ubp = bp.unbind()`, so that the previous behavior can be reproduced `bp.to_gpu(np.float32).bound_impl`, `bp.unbind().to_gpu(np.float32).unbound_impl`,
(B) revising both names to `.impl`?

mcwitt: > (A) adding a method to .unbind() (ubp = bp.unbind(), so that the previous behavior can be reproduced

The way to achieve this currently is `bp.potential` (i.e. just accessing the potential attribute). E.g. `bp.potential.to_gpu(np.float32).unbound_impl`

maxentile (approved):
Great changes throughout! Improvements in usability, consistency, and simplicity

tests/test_md.py
maxentile: Ahh, got it!

maxentile: Sorry, missed second part of response:

>I considered this, but overloading impl seemed a bit less explicit than what we had been doing, so didn't seem like an improvement. (Maybe I'm missing an argument as to why this would be preferable?)

No argument, just a reflex against repetition `bound_potential.bound_impl`, `unbound_potential.unbound_impl`.

On further thought, I agree with preference for current `.bound_impl`, `.unbound_impl` over `.impl` since it is more explicit.

timemachine/potentials/potentials.py
```diff
@@ -0,0 +1,270 @@
...
+        res = jax_interface.call_unbound_impl(self.unbound_impl, conf, params_flat, box)
+        return cast(float, res)
+
+    def bind_params_list(self, params: Params) -> "BoundGpuImplWrapper":
+        params_flat = np.concatenate([ps.reshape(-1) for ps in params])
```
jkausrelay, line 257: np.concatenate or jnp.concatenate?

mcwitt: Good catch. Using `np` instead of `jnp` here was actually intentional, to emphasize that it doesn't make sense to differentiate `bind_params_list` with respect to `params`.

timemachine/potentials/jax_utils.py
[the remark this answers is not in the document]
mcwitt: As discussed offline, it looks like this was not caught before because the pre-commit mypy check did not declare JAX as a dependency. Planning to leave this fix in for now and address the mypy issue in a later PR.

PR #998, opened 2023-04-11
Fix DuplicateAlignmentKeysError in torsion interpolation

mcwitt:
The current default forcefield has some patterns that contain duplicate period values, e.g. https://github.com/proteneer/timemachine/blob/c88b42c8aeadaeb7979558f04cae961502fd2917/timemachine/ff/params/smirnoff_2_0_0_ccc.py#L480. Parameter interpolation fails when such a pattern is matched, since we align source and destination parameters using (idxs, period).

This adds phase as an additional alignment key for torsions, so we now align on (idxs, period, phase). This is safe because there should not be duplicates in this key (if there were, the corresponding force constants could be combined).


tests/test_interpolate.py
```diff
@@ -83,25 +84,25 @@ def test_align_torsion():
     We expect that decoupled terms have their force constants turned set to zero,
     while maintaining the same equilibrium angles for the *same* period.
     """
-    a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p = np.random.rand(16)
+    a, b, c, d, e, f, g, h, i, j, j, k, l, m, n = np.random.rand(15)
```
maxentile, line 87: Surprise: repeated random assignment to j

```suggestion
    a, b, c, d, e, f, g, h, i, j, k, l, m, n = np.random.rand(14)
```

maxentile (approved):
Looks good!

tests/test_interpolate.py
```diff
@@ -261,3 +264,38 @@ def test_intermediate_states(num_pairs_to_setup=10):
...
+    mol_a = Chem.AddHs(Chem.MolFromSmiles("CNC(=O)c1cc2cc(Nc3nccc(-c4cn(C)cn4)n3)cc(Cl)c2[nH]1"))
+    AllChem.EmbedMolecule(mol_a)
+
+    # CHEMBL3668838
+    # has a subgroup matching a pattern with duplicate periods:
+    # https://github.com/proteneer/timemachine/blob/c88b42c8aeadaeb7979558f04cae961502fd2917/timemachine/ff/params/smirnoff_2_0_0_ccc.py#L480
+    mol_b = Chem.AddHs(Chem.MolFromSmiles("Cc1cc(Nc2nccc(-c3cn(C)cn3)n2)cc2cc(C(=O)NCc3nccs3)[nH]c12"))
+    AllChem.EmbedMolecule(mol_b)
+
+    core = atom_mapping.get_cores(
```
jkausrelay, line 283: Should the core be fixed (in case we change the atom mapping code later?)

mcwitt: Switched to fixed core in https://github.com/proteneer/timemachine/pull/998/commits/f43c120ec90066eca4b415f293bb66d1cd1f38fb

jkausrelay (approved):

PR #999, opened 2023-04-12
Increase vacuum min_cut to prevent false positives

jkausrelay:
- For larger ligands, minimization can cause it to rotate in vacuum leg
- This can be > 15 Å displacement 
- Resulting vacuum free energies seem consistent (with run that has large displacement and similar run that doesn't)

### DIFF
--- a/timemachine/fe/rbfe.py
+++ b/timemachine/fe/rbfe.py
@@ -628,9 +628,9 @@ def run_vacuum(
     n_eq_steps=10000,
     steps_per_frame=400,
     n_windows=None,
-    min_cutoff=1.5,
+    min_cutoff=2.5,
 ):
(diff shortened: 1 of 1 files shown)


proteneer:
This makes sense to me, but out of curiosity, which commit ended up triggering this behavior?

jkausrelay:
> This makes sense to me, but out of curiosity, which commit ended up triggering this behavior?

This was always present. I increased this value before (arbitrarily to 15 Å).

This is the case I showed you yesterday, where the molecule folds on itself. Note that if we account for the w-coord, the rings are not really on top of each other. Running the full vacuum leg and comparing to a run with a different atom mapping (where the minimization induces a smaller change in the structure) gives comparable results.

proteneer:
lgtm

PR #1000, opened 2023-04-12
Fix for HMR and dummy atoms

jkausrelay:
- Apply HMR to each molecule separately
- Combine core by picking the larger mass (similar to what we already are doing)
- Use the HMR value for the dummy atoms
- Fixes case where you have partially mapped methyl group, HMR was being applied to all dummy atoms causing the C mass to become too low

### DIFF
--- a/tests/test_single_topology.py
+++ b/tests/test_single_topology.py
@@ -310,6 +310,72 @@ def test_combine_masses():
     np.testing.assert_almost_equal(test_masses, ref_masses)
 
 
+def test_combine_masses_hmr():
+    C_mass = Chem.MolFromSmiles("C").GetAtomWithIdx(0).GetMass()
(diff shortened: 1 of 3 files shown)


maxentile (approved):

proteneer:
Congrats on hitting PR number 1000

proteneer (approved):
LGTM this makes total sense, also makes me wonder how we didn't get bitten by this before...

tests/test_single_topology.py
```diff
@@ -437,6 +503,7 @@ def get_vacuum_solvent_u_grads(ff, lamb):
 def ligand_from_smiles(smiles):
     mol = Chem.AddHs(Chem.MolFromSmiles(smiles))
     AllChem.Compute2DCoords(mol)
+    mol.SetProp("_Name", smiles)
```
proteneer, line 506: intentional? (probably doesn't hurt)

PR #1003, opened 2023-04-14
Fix for random seeds changed by HMR fix

jkausrelay:
- HMR Fix #1000 caused nightly test to fail
- Update test to check for consistency without looking a specific seed values

### DIFF
--- a/tests/test_relative_free_energy.py
+++ b/tests/test_relative_free_energy.py
@@ -19,12 +19,18 @@ from timemachine.md.barostat.utils import compute_box_center
 from timemachine.testsystems.relative import get_hif2a_ligand_pair_single_topology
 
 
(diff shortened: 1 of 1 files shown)


tests/test_relative_free_energy.py
```diff
@@ -187,6 +197,7 @@ def test_run_hif2a_test_system(estimate_relative_free_energy_fn):
         forcefield,
         n_frames=100,
         estimate_relative_free_energy_fn=estimate_relative_free_energy_fn,
+        sim_results_by_leg=sim_results_by_leg,
```
mcwitt, line 200: I might be misunderstanding, but wasn't the issue in https://github.com/proteneer/timemachine/pull/983 that seeds were not consistent between runs in different processes? Would that issue be caught by this test?

(We do now have a consistency check between runs in separate processes in `test_examples.py`.)

mcwitt (approved):

PR #1004, opened 2023-04-14
Refactor Local MD to speed up multiple_steps_local

badisa:
Trying to reorganize local MD to avoid the overhead of reconstructing potentials.

Has identical behavior to previous implementation, however when calling `multiple_steps_local` multiple times, it is significantly faster.

## Benchmark
Current performance
image

Performance with these changes
image

Plot has steps on the x axis and time in seconds on the y axis.

Note that in the second plot, the context had previously called `multiple_steps_local`, without it, would show a ~0.4 spike at the start from the cost of initialization. 

## Next Steps

* Add a `multiple_steps_selection` that allows user to provide a reference idx and a selection mask up front.

### DIFF
--- a/timemachine/cpp/src/local_md_config.cu
+++ b/timemachine/cpp/src/local_md_config.cu
@@ -21,7 +21,7 @@ struct LessThan {
 
 LocalMDConfig::LocalMDConfig(const int N, const std::vector<std::shared_ptr<BoundPotential>> bps)
     : N_(N), temp_storage_bytes_(0), all_potentials_(bps), restraints_(N_ * 2), bond_params_(N_ * 3),
-      probability_buffer_(round_up_even(N_)), d_free_indices_(N_), d_row_idxs_(N_), d_col_idxs_(N_), p_num_selected_(1),
+      probability_buffer_(round_up_even(N_)), d_free_idxs_(N_), d_row_idxs_(N_), d_col_idxs_(N_), p_num_selected_(1),
       num_selected_buffer_(1) {
 
     std::vector<std::shared_ptr<BoundPotential>> nonbonded_pots;
@@ -58,7 +58,7 @@ LocalMDConfig::LocalMDConfig(const int N, const std::vector<std::shared_ptr<Boun
     cub::DevicePartition::If(
         nullptr,
         temp_storage_bytes_,
-        d_free_indices_.data,
+        d_free_idxs_.data,
         d_row_idxs_.data,
         num_selected_buffer_.data,
         N_,
@@ -72,7 +72,7 @@ LocalMDConfig::LocalMDConfig(const int N, const std::vector<std::shared_ptr<Boun
 
 LocalMDConfig::~LocalMDConfig() { curandErrchk(curandDestroyGenerator(cr_rng_)); }
 
-// setup_from_idxs takes a set of indices, a temperature and a seed to determine the free particles. Fix the local_idxs to length
+// setup_from_idxs takes a set of idxs, a temperature and a seed to determine the free particles. Fix the local_idxs to length
 // one to ensure the same reference everytime, though the seed also handles the probabilities of selecting particles, and it is suggested
 // to provide a new seed at each step.
 void LocalMDConfig::setup_from_idxs(
@@ -93,7 +93,7 @@ void LocalMDConfig::setup_from_idxs(
 
     // Set the array to all N, which indicates to ignore that idx
     k_initialize_array<unsigned int>
-        <<<ceil_divide(N_, warp_size), warp_size, 0, stream>>>(N_, d_free_indices_.data, N_);
+        <<<ceil_divide(N_, warp_size), warp_size, 0, stream>>>(N_, d_free_idxs_.data, N_);
     gpuErrchk(cudaPeekAtLastError());
 
     // Generate values between (0, 1.0]
@@ -108,7 +108,7 @@ void LocalMDConfig::setup_from_idxs(
     const double kBT = BOLTZ * temperature;
     // Select all of the particles that will be free
     k_log_probability_selection<float><<<ceil_divide(N_, warp_size), warp_size, 0, stream>>>(
-        N_, kBT, radius, k, reference_idx, d_x_t, d_box_t, probability_buffer_.data, d_free_indices_.data);
+        N_, kBT, radius, k, reference_idx, d_x_t, d_box_t, probability_buffer_.data, d_free_idxs_.data);
     gpuErrchk(cudaPeekAtLastError());
 
     this->_setup_free(reference_idx, radius, k, stream);
@@ -120,11 +120,11 @@ void LocalMDConfig::_setup_free(
 
     LessThan select_op(N_);
 
-    // Partition the free indices into the row indices
+    // Partition the free idxs into the row idxs
     gpuErrchk(cub::DevicePartition::If(
         d_temp_storage_buffer_->data,
         temp_storage_bytes_,
-        d_free_indices_.data,
+        d_free_idxs_.data,
         d_row_idxs_.data,
         num_selected_buffer_.data,
         N_,
@@ -139,55 +139,55 @@ void LocalMDConfig::_setup_free(
         stream));
     gpuErrchk(cudaStreamSynchronize(stream));
 
-    const int num_row_indices = p_num_selected_.data[0];
-    const int num_col_indices = N_ - num_row_indices;
+    const int num_row_idxs = p_num_selected_.data[0];
+    const int num_col_idxs = N_ - num_row_idxs;
 
-    if (num_row_indices == 0) {
+    if (num_row_idxs == 0) {
         throw std::runtime_error("LocalMDConfig setup has no free particles selected");
     }
 
-    // The reference particle will always be in the column indices
-    if (num_row_indices == N_ - 1) {
+    // The reference particle will always be in the column idxs
+    if (num_row_idxs == N_ - 1) {
         fprintf(stderr, "LocalMDConfig setup has entire system selected\n");
     }
 
-    k_construct_bonded_params<<<ceil_divide(num_row_indices, tpb), tpb, 0, stream>>>(
-        num_row_indices, N_, reference_idx, k, 0.0, radius, d_row_idxs_.data, restraints_.data, bond_params_.data);
+    k_construct_bonded_params<<<ceil_divide(num_row_idxs, tpb), tpb, 0, stream>>>(
+        num_row_idxs, N_, reference_idx, k, 0.0, radius, d_row_idxs_.data, restraints_.data, bond_params_.data);
     gpuErrchk(cudaPeekAtLastError());
 
     // Setup the flat bottom restraints
-    bound_restraint_->set_params_device(std::vector<int>({num_row_indices, 3}), bond_params_.data, stream);
-    restraint_->set_bonds_device(num_row_indices, restraints_.data, stream);
+    bound_restraint_->set_params_device(std::vector<int>({num_row_idxs, 3}), bond_params_.data, stream);
+    restraint_->set_bonds_device(num_row_idxs, restraints_.data, stream);
 
     // Set the nonbonded potential to compute forces of free particles
-    set_nonbonded_potential_idxs(nonbonded_bp_->potential, num_row_indices, d_row_idxs_.data, stream);
+    set_nonbonded_potential_idxs(nonbonded_bp_->potential, num_row_idxs, d_row_idxs_.data, stream);
 
-    // Invert to get column indices
-    k_invert_indices<<<ceil_divide(N_, tpb), tpb, 0, stream>>>(N_, d_free_indices_.data);
+    // Invert to get column idxs
+    k_invert_indices<<<ceil_divide(N_, tpb), tpb, 0, stream>>>(N_, d_free_idxs_.data);
     gpuErrchk(cudaPeekAtLastError());
 
-    // Partition the column indices to the column buffer to setup the interaction group
+    // Partition the column idxs to the column buffer to setup the interaction group
     gpuErrchk(cub::DevicePartition::If(
         d_temp_storage_buffer_->data,
         temp_storage_bytes_,
-        d_free_indices_.data,
+        d_free_idxs_.data,
         d_col_idxs_.data,
         num_selected_buffer_.data,
         N_,
         select_op,
         stream));
 
-    // Free particles should be in the row indices
+    // Free particles should be in the row idxs
     set_nonbonded_ixn_potential_idxs(
-        ixn_group_->potential, num_col_indices, num_row_indices, d_col_idxs_.data, d_row_idxs_.data, stream);
+        ixn_group_->potential, num_col_idxs, num_row_idxs, d_col_idxs_.data, d_row_idxs_.data, stream);
 }
 
 std::vector<std::shared_ptr<BoundPotential>> LocalMDConfig::get_potentials() { return all_potentials_; }
 
-DeviceBuffer<unsigned int> *LocalMDConfig::get_free_indices() { return &d_row_idxs_; }
+DeviceBuffer<unsigned int> *LocalMDConfig::get_free_idxs() { return &d_row_idxs_; }
 
 void LocalMDConfig::reset(const cudaStream_t stream) {
-    // Set the row indices back to the identity.
+    // Set the row idxs back to the identity.
     k_arange<<<ceil_divide(N_, warp_size), warp_size, 0, stream>>>(N_, d_row_idxs_.data);
     gpuErrchk(cudaPeekAtLastError());
     // Set back to the full system
--- a/timemachine/cpp/src/local_md_config.cu
+++ b/timemachine/cpp/src/local_md_config.cu
@@ -1,0 +1,198 @@
+#include "constants.hpp"
+#include "gpu_utils.cuh"
+#include "kernel_utils.cuh"
+#include "kernels/k_flat_bottom_bond.cuh"
+#include "kernels/k_indices.cuh"
+#include "kernels/k_local_md.cuh"
+#include "local_md_config.hpp"
+#include "math_utils.cuh"
+#include <cub/cub.cuh>
+#include <random>
+#include <vector>
+
+namespace timemachine {
+
+// Struct to as a CUB <= operation
+struct LessThan {
+    int compare;
+    CUB_RUNTIME_FUNCTION __device__ __forceinline__ explicit LessThan(int compare) : compare(compare) {}
+    CUB_RUNTIME_FUNCTION __device__ __forceinline__ bool operator()(const int &a) const { return (a < compare); }
+};
+
+LocalMDConfig::LocalMDConfig(const int N, const std::vector<std::shared_ptr<BoundPotential>> bps)
+    : N_(N), temp_storage_bytes_(0), all_potentials_(bps), restraints_(N_ * 2), bond_params_(N_ * 3),
+      probability_buffer_(round_up_even(N_)), d_free_indices_(N_), d_row_idxs_(N_), d_col_idxs_(N_), p_num_selected_(1),
+      num_selected_buffer_(1) {
+
+    std::vector<std::shared_ptr<BoundPotential>> nonbonded_pots;
+    get_nonbonded_all_pair_potentials(bps, nonbonded_pots);
+
+    if (nonbonded_pots.size() > 1) {
+        throw std::runtime_error("found multiple NonbondedAllPairs potentials");
+    }
+    if (nonbonded_pots.size() != 1) {
+        throw std::runtime_error("unable to find a NonbondedAllPairs potential");
+    }
+
+    // Only used to reference shared_ptr to potential and for Nonbonded parameters
+    // modifications to the BoundPotential has no impact
+    nonbonded_bp_ = nonbonded_pots[0];
+
+    // Ensure that we allocate enough space for all potential bonds
+    std::vector<int> default_bonds(2 * N_);
+    for (int i = 0; i < N_; i++) {
+        default_bonds[i * 2 + 0] = 0;
+        default_bonds[i * 2 + 1] = i + 1;
+    }
+    restraint_ = std::shared_ptr<FlatBottomBond<double>>(new FlatBottomBond<double>(default_bonds));
+    // Construct a bound potential with 0 params
+    bound_restraint_ = std::shared_ptr<BoundPotential>(new BoundPotential(restraint_, std::vector<int>({0}), nullptr));
+
+    ixn_group_ =
+        construct_ixn_group_potential(N_, nonbonded_bp_->potential, nonbonded_bp_->size(), nonbonded_bp_->d_p->data);
+
+    // Add the restraint potential and ixn group potential
+    all_potentials_.push_back(bound_restraint_);
+    all_potentials_.push_back(ixn_group_);
+
+    cub::DevicePartition::If(
+        nullptr,
+        temp_storage_bytes_,
+        d_free_indices_.data,
+        d_row_idxs_.data,
+        num_selected_buffer_.data,
+        N_,
+        LessThan(N_));
+    // Allocate char as temp_storage_bytes_ is in raw bytes and the type doesn't matter in practice.
+    // Equivalent to DeviceBuffer<int> buf(temp_storage_bytes_ / sizeof(int))
+    d_temp_storage_buffer_.reset(new DeviceBuffer<char>(temp_storage_bytes_));
+
+    curandErrchk(curandCreateGenerator(&cr_rng_, CURAND_RNG_PSEUDO_DEFAULT));
+};
+
+LocalMDConfig::~LocalMDConfig() { curandErrchk(curandDestroyGenerator(cr_rng_)); }
+
+// setup_from_idxs takes a set of indices, a temperature and a seed to determine the free particles. Fix the local_idxs to length
+// one to ensure the same reference everytime, though the seed also handles the probabilities of selecting particles, and it is suggested
+// to provide a new seed at each step.
+void LocalMDConfig::setup_from_idxs(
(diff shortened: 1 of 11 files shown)


timemachine/cpp/src/local_config.cu
```diff
@@ -0,0 +1,190 @@
...
+            throw std::runtime_error("LocalMDConfig setup has no free particles selected");
+        }
+
+        // The reference particle will always be in the column indices
+        if (num_row_indices == N_ - 1) {
+            fprintf(stderr, "LocalMDConfig setup has entire system selected\n");
+        }
+
+        k_construct_bonded_params<<<ceil_divide(num_row_indices, tpb), tpb, 0, stream>>>(
+            num_row_indices, N_, reference_idx, k, 0.0, radius, d_row_idxs_.data, restraints_.data, bond_params_.data);
```
maxentile, line 149: From looking at this together offline with @badisa , I think this may have been part of the explanation for a discrepancy relative to `master`.

(If random # of restraints in current iteration is higher than random # of restraints in previous iteration, then these may not be zeroed out in `restraints_`, `bond_params_` due to this check https://github.com/proteneer/timemachine/blob/84442e30b4f62ea305722f969b6a983690d282e2/timemachine/cpp/src/kernels/k_local_md.cuh#L15-L17  .

Previously, `k_construct_bonded_params` was only invoked once immediately after construction, rather than repeatedly on the same `restraints_`, `bond_params_` arrays, which can now contain stale non-zero values at `idxs` > current `num_idxs`.)

mcwitt:
In the plots, I'm assuming that the blue line is unchanged? (it's a bit hard to tell because of the log scale in the first plot)

badisa:
> In the plots, I'm assuming that the blue line is unchanged? (it's a bit hard to tell because of the log scale in the first plot)

The blue line is unchanged on my machine, will verify with the CI tests. The first plot was generated by @maxentile and the second was generated by me. Guessing we have different GPUs, which is why the time is different.

maxentile:
(Note: The plots in the benchmark above may be complicated due to different devices: the "current performance" graph was on a 2080 Ti, not sure which device the proposed graph was run on.)

badisa:
> (Note: The plots in the benchmark above may be complicated due to different devices: the "current performance" graph was on a 2080 Ti, not sure which device the proposed graph was run on.)

Was run on an A4000. And confirmed that CI shows no difference. Which we would expect, as the object should never be instantiated in the global md case.

maxentile:
Nice work @badisa !

This change looks like a major improvement for cases where the overhead of calling multiple_steps_local was previously significant, and the refactoring is helpful for clarity as well.

timemachine/cpp/src/local_md_config.cu
```diff
@@ -0,0 +1,197 @@
...
+    k_construct_bonded_params<<<ceil_divide(num_row_indices, tpb), tpb, 0, stream>>>(
+        num_row_indices, N_, reference_idx, k, 0.0, radius, d_row_idxs_.data, restraints_.data, bond_params_.data);
+    gpuErrchk(cudaPeekAtLastError());
+
+    // Setup the flat bottom restraints
+    bound_restraint_->set_params_device(std::vector<int>({num_row_indices, 3}), bond_params_.data, stream);
+    restraint_->set_bonds_device(num_row_indices, restraints_.data, stream);
+
+    // Set the nonbonded potential to compute forces of free particles
+    set_nonbonded_potential_idxs(nonbonded_bp_->potential, num_row_indices, d_row_idxs_.data, stream);
```
mcwitt, line 163: tiny nit: mixed convention with "indices" and "idxs" (might have been present before)

timemachine/cpp/src/local_md_config.cu
```diff
@@ -0,0 +1,197 @@
...
+namespace timemachine {
+
+// Struct to as a CUB <= operation
+struct LessThan {
+    int compare;
+    CUB_RUNTIME_FUNCTION __device__ __forceinline__ explicit LessThan(int compare) : compare(compare) {}
+    CUB_RUNTIME_FUNCTION __device__ __forceinline__ bool operator()(const int &a) const { return (a < compare); }
+};
+
+LocalMDConfig::LocalMDConfig(const int N, const std::vector<std::shared_ptr<BoundPotential>> bps)
```
mcwitt, line 22: naming nit: to me `LocalMDConfig` suggests an object that holds configuration values, but this seems like more of a wrapper for instantiated potentials? Don't immediately have a good suggestion, but maybe something like `LocalMDPotentials` would be preferable?

mcwitt (approved):
Looks great!

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -272,6 +272,8 @@ void declare_context(py::module &m) {
 
         F = iterations / store_x_interval
 
+        The first call to `multiple_steps_local` takes longer than subsequent calls, potentials need to be configured internally.
```
maxentile, line 275: Usage nit (not for this PR!) -- may be convenient to add a function `ctxt.initialize_local_md()` or similar, that calls these constructors but doesn't perform any steps / doesn't update (x, v, box) state.

(So that this can be done without the small amount of inconsistency that could be introduced by making 1 initial call.)

(And the first call to `multiple_steps_local()` would invoke this, if it hasn't been invoked before.)

timemachine/cpp/src/local_md_config.hpp
```diff
@@ -0,0 +1,66 @@
...
+    DeviceBuffer<double> bond_params_;
+
+    DeviceBuffer<float> probability_buffer_;
+    DeviceBuffer<unsigned int> d_free_indices_;
+    std::unique_ptr<DeviceBuffer<char>> d_temp_storage_buffer_;
+
+    DeviceBuffer<unsigned int> d_row_idxs_;
+    DeviceBuffer<unsigned int> d_col_idxs_;
+
+    // Pinned memory for getting lengths of indice arrays
```
maxentile:
```suggestion
    // Pinned memory for getting lengths of index arrays
```

maxentile (approved):
Looks great!

Thanks for walking me through these changes offline.

timemachine/cpp/src/wrap_kernels.cpp
badisa: Definitely. This change would also allow us to avoid the need for the user to pass in a seed every step. Which makes things a bit messy.

We could switch to a more explicit style of calling `ctxt.initialize_local_md` before they can run local md and do all of the parameterization then rather than having to pass in 'good' configurations in each call.

maxentile:
(For future reference, reproducing the benchmark in https://github.com/proteneer/timemachine/pull/1004#issue-1668556315 on the same device and settings produces the following plot:

image

and with log y-scale, for better resolution of the very small timings:

image

(settings: burn_in = 0, store_x_interval = 0, radius = 1.0, k = 10000.0 -- resulting in ~25% of the n = 2573 particles in the benzene + waterbox system to be selected per local move)

So even for the smallest systems we'd consider running, even at very small n_steps, the version @ https://github.com/proteneer/timemachine/pull/1004/commits/8265dc02fac4f6996f108aa5b81fb653df0adecb is at least as fast as `multiple_steps`!

And we would of course expect larger gains for larger systems and with settings that select a smaller region, as the ratio of selected interactions / total interactions decreases. Nicely done!)

timemachine/cpp/src/local_md_potentials.cu
```diff
@@ -0,0 +1,190 @@
...
+#include "kernels/k_local_md.cuh"
+#include "local_md_potentials.hpp"
+#include "math_utils.cuh"
+#include <cub/cub.cuh>
+#include <random>
+#include <vector>
+
+namespace timemachine {
+
+// Struct to as a CUB <= operation
```
proteneer, line 15: <= or < ?

timemachine/cpp/src/local_md_potentials.hpp
```diff
@@ -0,0 +1,66 @@
...
+    const int N_;
+    std::size_t temp_storage_bytes_;
+
+    std::vector<std::shared_ptr<BoundPotential>> all_potentials_;
+    std::shared_ptr<BoundPotential> ixn_group_;
+    std::shared_ptr<BoundPotential> nonbonded_bp_;
+    std::shared_ptr<FlatBottomBond<double>> restraint_;
+    std::shared_ptr<BoundPotential> bound_restraint_;
+
+    DeviceBuffer<int> restraints_;
```
proteneer, line 47: Rename to `restraint_idxs_` or `restraint_pairs_`?

proteneer: (also currently very easy to typo into the shared_ptr `restraint_`)

timemachine/cpp/src/local_md_potentials.cu
```diff
@@ -0,0 +1,190 @@
...
+
+    // Only used to reference shared_ptr to potential and for Nonbonded parameters
+    // modifications to the BoundPotential has no impact
+    nonbonded_bp_ = nonbonded_pots[0];
+
+    // Ensure that we allocate enough space for all potential bonds
+    std::vector<int> default_bonds(2 * N_);
+    for (int i = 0; i < N_; i++) {
+        default_bonds[i * 2 + 0] = 0;
+        default_bonds[i * 2 + 1] = i + 1;
```
proteneer, line 45: why is this set to `i+1` if we only want to do initialization? did we want to create bonds between `(0,1), (0,3), (0,5)`?

proteneer: (is N_ == K here?)

timemachine/cpp/src/local_md_potentials.cu
```diff
@@ -0,0 +1,190 @@
...
+    }
+    restraint_ = std::shared_ptr<FlatBottomBond<double>>(new FlatBottomBond<double>(default_bonds));
+    // Construct a bound potential with 0 params
```
proteneer, line 48: did we not want to allocate some default params (similar to the above for bond idxs)?

timemachine/cpp/src/local_md_potentials.cu
```diff
@@ -0,0 +1,190 @@
...
+    const double kBT = BOLTZ * temperature;
+    // Select all of the particles that will be free
+    k_log_probability_selection<float><<<ceil_divide(N_, warp_size), warp_size, 0, stream>>>(
+        N_, kBT, radius, k, reference_idx, d_x_t, d_box_t, probability_buffer_.data, d_free_idxs_.data);
+    gpuErrchk(cudaPeekAtLastError());
+
+    this->_setup_free(reference_idx, radius, k, stream);
+}
+
+void LocalMDPotentials::_setup_free(
```
proteneer, line 110: maybe rename to `_setup_free_idxs()`?

timemachine/cpp/src/local_md_potentials.cu
```diff
@@ -0,0 +1,190 @@
...
+    if (num_row_idxs == N_ - 1) {
+        fprintf(stderr, "LocalMDPotentials setup has entire system selected\n");
+    }
+
+    k_construct_bonded_params<<<ceil_divide(num_row_idxs, tpb), tpb, 0, stream>>>(
+        num_row_idxs, N_, reference_idx, k, 0.0, radius, d_row_idxs_.data, restraints_.data, bond_params_.data);
+    gpuErrchk(cudaPeekAtLastError());
+
+    // Setup the flat bottom restraints
+    bound_restraint_->set_params_device(std::vector<int>({num_row_idxs, 3}), bond_params_.data, stream);
```
proteneer, line 152: Note to self, pinged @badisa offline about this:

I think setting new parameters is fine (overriding existing ones but maintaining shape), but setting new shapes is going to be hell to debug. Not sure why we ever supported this.

proteneer: (maybe @mcwitt has an idea)

timemachine/cpp/src/local_md_potentials.cu
badisa: What we want is for `default_bonds[i * 2 + 0] != default_bonds[i * 2 + 1]` as that is verified in the constructor. This was a way to ensure that, could do it pretty much any way we want.

Will add comment for now, and we can determine a different approach later

> (is N_ == K here?)

Yes

timemachine/cpp/src/local_md_potentials.cu
badisa: This was originally done to try and avoid using more memory than we had to. With this approach and the fact that these contexts will be few in number, we can remove it.

We can allocate the potential parameters like you asked here: https://github.com/proteneer/timemachine/pull/1004#discussion_r1169039369 instead. I would be inclined to make this change in a separate PR and combined it with the @mcwitt  suggestion of removing `shape` as a concept from the BoundPotentials, as it always ends up converted to a shape in practice.

timemachine/cpp/src/local_md_potentials.cu
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1004/commits/57f21ba6cf5a201ba2f7a99f3580474e81a82570

timemachine/cpp/src/local_md_potentials.cu
badisa: Added comment here: https://github.com/proteneer/timemachine/pull/1004/commits/5621c2dc9fbd1ac77b6205c62de668ed233c6a29

timemachine/cpp/src/local_md_potentials.cu
badisa: Renamed in https://github.com/proteneer/timemachine/pull/1004/commits/23af6f2ae599ddc303f8887155882c3966d6287a

timemachine/cpp/src/local_md_potentials.hpp
badisa: Good catch, fixed here: https://github.com/proteneer/timemachine/pull/1004/commits/59ead18e83f111b194321725b0d92782388c468b

timemachine/cpp/src/local_md_potentials.cu
```diff
@@ -0,0 +1,191 @@
...
+#include "kernels/k_local_md.cuh"
+#include "local_md_potentials.hpp"
+#include "math_utils.cuh"
+#include <cub/cub.cuh>
+#include <random>
+#include <vector>
+
+namespace timemachine {
+
+// Struct to as a CUB < operation
```
proteneer, line 15: nit: struct to ?

timemachine/cpp/src/local_md_potentials.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include "curand.h"
+#include "flat_bottom_bond.hpp"
+#include "local_md_utils.hpp"
+
+namespace timemachine {
+
+class LocalMDPotentials {
+
+public:
+    LocalMDPotentials(const int N, const std::vector<std::shared_ptr<BoundPotential>> bps);
```
proteneer, line 17: why not `const ... &bps`?

timemachine/cpp/src/local_md_potentials.hpp
```diff
@@ -0,0 +1,67 @@
...
+    DeviceBuffer<float> probability_buffer_;
+    DeviceBuffer<unsigned int> d_free_idxs_;
+    std::unique_ptr<DeviceBuffer<char>> d_temp_storage_buffer_;
+
+    DeviceBuffer<unsigned int> d_row_idxs_;
+    DeviceBuffer<unsigned int> d_col_idxs_;
+
+    // Pinned memory for getting lengths of index arrays
+    PinnedHostBuffer<int> p_num_selected_;
+    DeviceBuffer<int> num_selected_buffer_;
```
proteneer, line 59: Add the `d_` prefix for consistency

timemachine/cpp/src/local_md_potentials.hpp
```diff
@@ -0,0 +1,67 @@
...
+    const int N_;
+    std::size_t temp_storage_bytes_;
+
+    std::vector<std::shared_ptr<BoundPotential>> all_potentials_;
+    std::shared_ptr<BoundPotential> ixn_group_;
+    std::shared_ptr<BoundPotential> nonbonded_bp_;
+    std::shared_ptr<FlatBottomBond<double>> restraint_;
+    std::shared_ptr<BoundPotential> bound_restraint_;
+
+    DeviceBuffer<int> restraint_pairs_;
```
proteneer, line 47: Add the `d_` prefix for consistency

timemachine/cpp/src/local_md_potentials.hpp
```diff
@@ -0,0 +1,67 @@
...
+    DeviceBuffer<double> bond_params_;
```
proteneer, line 48: Add the `d_` prefix for consistency

timemachine/cpp/src/local_md_potentials.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+    DeviceBuffer<float> probability_buffer_;
```
proteneer, line 50: Add the `d_` prefix for consistency

timemachine/cpp/src/local_md_potentials.cu
```diff
@@ -0,0 +1,191 @@
...
+    const double kBT = BOLTZ * temperature;
+    // Select all of the particles that will be free
+    k_log_probability_selection<float><<<ceil_divide(N_, warp_size), warp_size, 0, stream>>>(
+        N_, kBT, radius, k, reference_idx, d_x_t, d_box_t, probability_buffer_.data, d_free_idxs_.data);
+    gpuErrchk(cudaPeekAtLastError());
+
+    this->_setup_free_idxs_given_reference_idx(reference_idx, radius, k, stream);
+}
+
+void LocalMDPotentials::_setup_free_idxs_given_reference_idx(
```
proteneer, line 111: was there actually a performance reason to do this on the GPU? I imagine we reset  `reference_idxs` and `free_idxs` only once say every 1000 steps in practice?

timemachine/cpp/src/local_md_potentials.cu
proteneer: sgtm!

timemachine/cpp/src/local_md_potentials.hpp
```diff
@@ -0,0 +1,67 @@
...
+namespace timemachine {
+
+class LocalMDPotentials {
+
+public:
+    LocalMDPotentials(const int N, const std::vector<std::shared_ptr<BoundPotential>> bps);
+
+    ~LocalMDPotentials();
+
+    DeviceBuffer<unsigned int> *get_free_idxs();
```
proteneer, line 21: ugh - this is a bit gnarly, why do we return a raw pointer here? Do we not support returning by const reference? (eg.

```const DeviceBuffer<unsigned int> &get_free_idxs();```

File: timemachine/cpp/src/local_md_potentials.cu [hunk 20]
```diff
@@ -0,0 +1,191 @@
...
+    set_nonbonded_ixn_potential_idxs(
+        ixn_group_->potential, num_col_idxs, num_row_idxs, d_col_idxs_.data, d_row_idxs_.data, stream);
+}
+
+std::vector<std::shared_ptr<BoundPotential>> LocalMDPotentials::get_potentials() { return all_potentials_; }
+
+DeviceBuffer<unsigned int> *LocalMDPotentials::get_free_idxs() { return &d_row_idxs_; }
+
+void LocalMDPotentials::reset(const cudaStream_t stream) {
+    // Set the row idxs back to the identity.
```
proteneer, line 184: we don't need to do anything with `d_col_idxs_` or other buffers here?

proteneer: (usually `reset` semantically resets all the internal state, here it looks like we're just resetting the row_idxs)

timemachine/cpp/src/local_md_potentials.cu
```diff
@@ -0,0 +1,191 @@
...
+    // Free particles should be in the row idxs
+    set_nonbonded_ixn_potential_idxs(
+        ixn_group_->potential, num_col_idxs, num_row_idxs, d_col_idxs_.data, d_row_idxs_.data, stream);
+}
+
+std::vector<std::shared_ptr<BoundPotential>> LocalMDPotentials::get_potentials() { return all_potentials_; }
+
+DeviceBuffer<unsigned int> *LocalMDPotentials::get_free_idxs() { return &d_row_idxs_; }
+
+void LocalMDPotentials::reset(const cudaStream_t stream) {
```
proteneer, line 183: `const cudaStream_t & tream`?

timemachine/cpp/src/local_md_potentials.cu
proteneer: nm - we're deferring this until we remove the ability to modify shapes in bound potentials?

timemachine/cpp/src/local_md_potentials.cu
```diff
@@ -0,0 +1,191 @@
...
+
+    gpuErrchk(cudaMemcpyAsync(
+        p_num_selected_.data,
+        num_selected_buffer_.data,
+        1 * sizeof(*p_num_selected_.data),
+        cudaMemcpyDeviceToHost,
+        stream));
+    gpuErrchk(cudaStreamSynchronize(stream));
+
+    const int num_row_idxs = p_num_selected_.data[0];
```
proteneer, line 136: can you add a comment indicating what the `row_idxs` and `col_idxs` correspond to?

proteneer: (discussing offline: row_idxs: free_idxs but no reference, col_idxs: reference + everything else.)

timemachine/cpp/src/local_md_potentials.cu
badisa: Do we want to pass the stream by reference? A quick search through the repo seems to indicate that we pass it by value.

timemachine/cpp/src/local_md_potentials.cu
badisa: Will rename to `reset_potentials` as really it is here to reset the potentials that it had modified for local MD back to their original state. At the moment the only potential modified is the NonbondedAllPairs potential to run on the free idxs.

timemachine/cpp/src/local_md_potentials.hpp
badisa: If we pass the const reference, the underlying memory gets cleaned up and triggers an error. Could return raw pointer like our other APIs for consistency

timemachine/cpp/src/local_md_potentials.cu
proteneer: hm - then maybe make the signature just `(cudaStream_t stream)` - it's a convention thing to prefer `const` references as oppose to bare `const`, or pure values:

either is probably fine:

`cudaStream_t stream`
or
`const  cudaStream_t &stream`

timemachine/cpp/src/local_md_potentials.cu
badisa: This decision was made based on wanting to keep everything on the GPU. Could do it on the CPU, but would have to copy the coords from the GPU to the host in some cases which seemed undesirable.

timemachine/cpp/src/local_md_potentials.hpp
badisa: Done in https://github.com/proteneer/timemachine/pull/1004/commits/31ce4bd377349ede34c0e2ec60bf01411b1055b2

timemachine/cpp/src/local_md_potentials.hpp
badisa: Done in https://github.com/proteneer/timemachine/pull/1004/commits/31ce4bd377349ede34c0e2ec60bf01411b1055b2

timemachine/cpp/src/local_md_potentials.hpp
badisa: Done in https://github.com/proteneer/timemachine/pull/1004/commits/31ce4bd377349ede34c0e2ec60bf01411b1055b2

timemachine/cpp/src/local_md_potentials.hpp
badisa: Done in https://github.com/proteneer/timemachine/pull/1004/commits/31ce4bd377349ede34c0e2ec60bf01411b1055b2

timemachine/cpp/src/local_md_potentials.hpp
badisa: Done in https://github.com/proteneer/timemachine/pull/1004/commits/ff9ded78bfea7a3cfc9f7a74bba3f1b8992ddb99

timemachine/cpp/src/local_md_potentials.cu
badisa: Clarified in https://github.com/proteneer/timemachine/pull/1004/commits/83a4e4d6ab9197bccc5f725ea659d656484a3dc6

timemachine/cpp/src/local_md_potentials.cu
badisa: Done in https://github.com/proteneer/timemachine/pull/1004/commits/c18081743ee1457373fec505d09a1b6f4fe85ac8

timemachine/cpp/src/local_md_potentials.cu
badisa: Added comment in https://github.com/proteneer/timemachine/pull/1004/commits/4e35e5e1a439746ae038507419311b924a0cd309

timemachine/cpp/src/local_md_potentials.cu
badisa: Renamed and added doc string https://github.com/proteneer/timemachine/pull/1004/commits/ebdbd0e0581ac1b15718e1c3497869c7ae7aa3d4

proteneer:
lgtm!

proteneer (approved):

PR #1005, opened 2023-04-17
Local MD by Selection Mask

badisa:
Adds a method to select particles outside of C++ to apply Local MD to.

Allows a constant selection of particles to avoid the bias of turning on and off restraints.


timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -308,6 +309,107 @@ void declare_context(py::module &m) {
...
+            py::arg("k") = 10000.0,
+            R"pbdoc(
+        Take multiple steps using a selection of free particles restrainted to a reference particle. Useful for avoiding the bias
+        introduced by switching on and off the restraint on different particles as is done with multiple_steps_local.
+
+        Running a barostat and local MD at the same time are not currently supported. If a barostat is
+        assigned to the context, the barostat won't run.
+
+        Note: Running this multiple times with small number of steps (< 100) may result in a vacuum around the local idxs due to
+        discretization error caused by switching on the restraint after a particle has moved beyond the radius.
```
maxentile, line 377: Is this note from a different version, or observed in this version?

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -308,6 +309,107 @@ void declare_context(py::module &m) {
...
+                    throw std::runtime_error("radius must be greater or equal to " + std::to_string(min_radius));
+                }
+                if (k < 1.0) {
+                    throw std::runtime_error("k must be at least one");
+                }
+                // TBD determine a more precise threshold, currently 10x what has been tested
+                const double max_k = 1000000.0;
+                if (k > max_k) {
+                    throw std::runtime_error("k must be less than than " + std::to_string(max_k));
+                }
```
maxentile, line 342: todo [ future ] -- consolidate `validate_local_md_settings()` that can be used by this function and `multiple_steps_local`

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -308,6 +309,107 @@ void declare_context(py::module &m) {
...
+
+        The first call to `multiple_steps_local_selection` takes longer than subsequent calls, if initialize_local_md has not been called previously,
+        initializes potentials needed for local md.
+
+        Parameters
+        ----------
+        n_steps: int
+            Number of steps to run.
+
+        selection_mask: np.array of int32
```
maxentile, line 389: nit [ future / naming ] -- `selection_mask` -> `selection_idxs`.

(To standardize on `selection_mask` = length-N boolean mask, `selection_mask[i] = True` if `i` is selected, vs. `selection_idxs` = variable length int list, `i in selection_idxs` if `i` is selected.)

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -308,6 +309,107 @@ void declare_context(py::module &m) {
...
+            },
+            py::arg("n_steps"),
+            py::arg("reference_idx"),
+            py::arg("selection_mask"),
+            py::arg("burn_in") = 500, // This is arbitrarily selected as a default, TODO make informed choice
+            py::arg("store_x_interval") = 0,
+            py::arg("radius") = 1.2,
+            py::arg("k") = 10000.0,
+            R"pbdoc(
+        Take multiple steps using a selection of free particles restrainted to a reference particle. Useful for avoiding the bias
```
maxentile, line 370: restrainted -> restrained (also L402)

tests/test_md.py
```diff
@@ -640,6 +640,127 @@ def test_multiple_steps_local_no_free_particles(self):
...
+            reference_values.append(bp.execute(coords, box))
+
+        intg = LangevinIntegrator(temperature, dt, friction, masses, seed)
+
+        idxs = np.arange(0, len(coords))
+
+        reference_idx = rng.choice(idxs)
+        # TODO, make it so you can't have the reference in the free
+        free_particles = rng.choice(idxs, size=len(coords) // 2)
+        free_particles = np.delete(free_particles, free_particles == reference_idx)
```
maxentile, line 751: 👍 -- should assert 

Aside: I think `reference_idx` in `free_particles` would technically be permissible if `r0 == 0`.

Problem being avoided: A discontinuity in `log(1 - exp(-u_flat_bottom_restraint(r; r0; k))` if `reference_idx` moves in such a way that its distance `r` to any frozen particle that was initially `r > r0` away from `x[reference_idx]` becomes `r < r0` away. We prevent local updates from crossing this discontinuity by preventing local updates from moving `reference_idx`.

(Bonus aside: We also have an assertion in the wrappers that `r0 > 0.1` whose motivation is less clear to me -- if we relax this assertion, then we could technically allow `reference_idx` in `free_particles` as long as `r0 == 0.0`. But I see no practical motivation to allow this.)

maxentile (approved):
Awesome!

Should increase control, and allow users to:

(1) make repeated calls to `multiple_steps_local_*` with the same selection but different other parameters (as in noneq switching / SMC / local augmentation / other uses).

(2) experiment with different ways to choose reference idx from Python, other than uniform from a pre-defined set of idxs.

(E.g. weighted sample from a pre-defined set of idxs (which requires no additional bookkeeping), or configuration-dependent selections (if we do pursue this, we should remember to derive / perform the relevant additional bookkeeping).)

----------

(Left cosmetic nits for future reference, but substance looks good!)

timemachine/cpp/src/context.cu
```diff
@@ -135,6 +139,83 @@ std::array<std::vector<double>, 2> Context::multiple_steps_local(
...
+    const int store_x_interval,
+    const double radius,
+    const double k) {
+    if (store_x_interval <= 0) {
+        throw std::runtime_error("store_x_interval <= 0");
+    }
+
+    const int x_buffer_size = n_steps / store_x_interval;
+
+    const int box_buffer_size = x_buffer_size * 3 * 3;
```
proteneer, line 156: Isn't the box constant during local MD? (i.e. NVT)

timemachine/cpp/src/wrap_kernels.cpp
proteneer: +1 - I got confused by this too. I like the convention that `masks` are boolean arrays, and `idxs` are list of ints.

proteneer: (note that what's more confusing, is that we often times avoid using `std::vector<bool>` to denote boolean flags of size `N` particles, i.e. we still prefer `std::vector<int>`, due to `sizeof(bool)` being a bit janky)

proteneer: In either case, I think documenting the size of `selection_mask` would be useful, i.e. either being of `size K` or something else

timemachine/cpp/src/context.cu
```diff
@@ -55,6 +55,12 @@ double Context::_get_temperature() {
     }
 }
 
+void Context::initialize_local_md() {
+    if (!this->local_md_pots_) {
```
proteneer, line 59: nit: I prefer to be a bit more verbose in these types of situations, and not rely on operator overloads of `!`

```
if(this->local_md_pots_.get() == nullptr) {}
```

timemachine/cpp/src/context.cu
maxentile: Yes, this should be a constant trajectory as currently implemented.

----------------

(Something we haven't discussed, but I think would be technically permissible if motivated, is to interleave MC barostat moves and local MD moves. Right now the context only allows to interleave MC barostat moves with global MD moves, but I do not think this is a real requirement.)

(Also, I don't currently foresee a motivation to apply MC barostat moves _during_ a single local MD move, but that could also be feasible, with some not-yet-worked-out bookkeeping to account for `p(selection | x_scaled, box_scaled) / p(selection| x, box)`. (Haven't investigated yet, but guess would be: incorporate this factor into MC barostat accept/reject step.))

proteneer:
General note: the `Context` class is starting to get quite unwieldy (esp. with a lot of extra state dependent on calls to things like `local_md`). We should consider a refactor of this class in the next deboggling week.

timemachine/cpp/src/wrap_kernels.cpp
badisa: Done in https://github.com/proteneer/timemachine/pull/1005/commits/d9d511a09a7e3a09ecb0de50ccba1b85e2704609

timemachine/cpp/src/wrap_kernels.cpp
badisa: Done in https://github.com/proteneer/timemachine/pull/1005/commits/7eee6dec2226057d4513b3770313fa89d2e6d521

Though didn't document size, will expand that

timemachine/cpp/src/wrap_kernels.cpp
badisa: Good catch, fixed in https://github.com/proteneer/timemachine/pull/1005/commits/c4b945361803938a86533a84cc6b441e666a025a

tests/test_md.py
badisa: Asserted in  https://github.com/proteneer/timemachine/pull/1005/commits/37cd439f94c34fd532d9e286434259287b201144

Comment removed in https://github.com/proteneer/timemachine/pull/1005/commits/801bb6c51982b1c566726fd1d1062e2506d7a72f

timemachine/cpp/src/context.cu
badisa: It is, though seemed more consistent to return boxes rather than having to have the user handle tracking the boxes that go along with local md.

timemachine/cpp/src/wrap_kernels.cpp
badisa: Added in https://github.com/proteneer/timemachine/pull/1005/commits/1b098ad9cacf81bf0b7c5d0ad4bf664a92901b59

timemachine/cpp/src/context.cu
proteneer: okay - I'm in favor of keeping things consistent then. (but maybe a comment that for NVT boxes are the same)

badisa: Added comment: https://github.com/proteneer/timemachine/pull/1005/commits/770649816066f30a07ccf1bba6a1dee45f907337

proteneer (approved):
lgtm

timemachine/cpp/src/local_md_potentials.cu
```diff
@@ -114,6 +113,37 @@ void LocalMDPotentials::setup_from_idxs(
     this->_setup_free_idxs_given_reference_idx(reference_idx, radius, k, stream);
 }
 
+// setup_from_idxs takes a set of idxs, a temperature and a seed to determine the free particles. Fix the local_idxs to length
+// one to ensure the same reference everytime, though the seed also handles the probabilities of selecting particles, and it is suggested
+// to provide a new seed at each step.
+void LocalMDPotentials::setup_from_idxs(
```
maxentile, line 119:
```suggestion
// setup_from_idxs takes a set of idxs, flat-bottom restraint parameters (radius, k)
// assumes selection_idxs are sampled based on exp(-beta U_flat_bottom(distance_to_reference, radius, k))
// (or that the user is otherwise accounting for selection probabilities)
void LocalMDPotentials::setup_from_idxs(
```

tests/test_md.py
```diff
@@ -640,6 +692,127 @@ def test_multiple_steps_local_no_free_particles(self):
...
+        ref_xs, ref_boxes = ctxt.multiple_steps(steps)
+
+        ctxt = custom_ops.Context(coords, v0, box, intg.impl(), bps)
+        ctxt.initialize_local_md()
+        comp_xs, comp_boxes = ctxt.multiple_steps(steps)
+
+        np.testing.assert_array_equal(ref_xs, comp_xs)
+        np.testing.assert_array_equal(ref_boxes, comp_boxes)
+
+        # Verify that initializing local md doesn't modify global md behavior
```
maxentile, line 752:
```suggestion
        # Verify that initializing local md doesn't modify local md behavior
```

tests/test_md.py
```diff
@@ -640,6 +692,127 @@ def test_multiple_steps_local_no_free_particles(self):
...
+
+        reference_idx = rng.choice(idxs)
+
+        free_particles = rng.choice(idxs, size=len(coords) // 2, replace=False)
+        free_particles = np.delete(free_particles, free_particles == reference_idx)
+        frozen_particles = np.delete(idxs, free_particles)
+
+        steps = 10
+
+        # Verify that initializing local md doesn't modify global md behavior
```
maxentile, line 805:
```suggestion
        # Verify that initializing local md doesn't modify local md behavior
```

tests/test_md.py
```diff
@@ -391,6 +377,76 @@ def test_multiple_steps_local_validation(self):
...
+
+        dt = 1.5e-3
+        radius = 1.2
+
+        v0 = np.zeros_like(coords)
+        bps = [nb_pot.bind(params).bound_impl]
+
+        # Use verlet since multiple_steps_local_selection doesn't depend on temperature
+        verlet = VelocityVerletIntegrator(dt, masses)
+        intg_impl = verlet.impl()
```
maxentile, line 409:
```suggestion
        # Compatible with local NVE since multiple_steps_local_selection doesn't depend on temperature
        verlet = VelocityVerletIntegrator(dt, masses)
        intg_impl = verlet.impl()
```

Nice -- this interface also resolves an issue that was bugging me in https://github.com/proteneer/timemachine/pull/915#discussion_r1039794884

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -308,6 +298,103 @@ void declare_context(py::module &m) {
...
+
+        The first call to `multiple_steps_local_selection` takes longer than subsequent calls, if initialize_local_md has not been called previously,
+        initializes potentials needed for local md.
+
+        Parameters
+        ----------
+        n_steps: int
+            Number of steps to run.
+
+        selection_idxs: np.array of int32
```
maxentile, line 373:
```suggestion
        reference_idxs, selection_idxs: int, np.array of int32
```

maxentile (approved):
Looks great!

(Documentation nits only.)

tests/test_md.py
badisa: This line is copy-pasta, should be removed

timemachine/cpp/src/wrap_kernels.cpp
badisa: Fixed in [e5f18c6](https://github.com/proteneer/timemachine/pull/1005/commits/e5f18c6d278dd64575877e9a8212a502847f91f3)

tests/test_md.py
```diff
@@ -391,6 +377,76 @@ def test_multiple_steps_local_validation(self):
...
+                100, reference_idx, np.array([1], dtype=np.int32), radius=radius, burn_in=-5
+            )
+
+        with pytest.raises(RuntimeError, match="radius must be greater or equal to 0.1"):
+            ctxt.multiple_steps_local_selection(100, reference_idx, np.array([1], dtype=np.int32), radius=0.01)
+
+        with pytest.raises(RuntimeError, match="k must be at least one"):
+            ctxt.multiple_steps_local_selection(100, reference_idx, np.array([1], dtype=np.int32), k=0.0)
+
+        with pytest.raises(RuntimeError, match="k must be less than than 1000000.0"):
```
mcwitt, line 438: nit: readability:
```suggestion
        with pytest.raises(RuntimeError, match="k must be less than than 1e6"):
```

mcwitt, line 438: nit: readability:
```suggestion
        with pytest.raises(RuntimeError, match="k must be less than than 1e6"):
```

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -308,6 +298,106 @@ void declare_context(py::module &m) {
...
+                verify_atom_idxs(N, vec_selection_idxs);
+                std::set<int> selection_set(vec_selection_idxs.begin(), vec_selection_idxs.end());
+                if (selection_set.find(reference_idx) != selection_set.end()) {
+                    throw std::runtime_error("reference idx must not be in selection idxs");
+                }
+
+                std::array<std::vector<double>, 2> result = ctxt.multiple_steps_local_selection(
+                    n_steps, reference_idx, vec_selection_idxs, burn_in, x_interval, radius, k);
+                const int D = 3;
+                const int F = result[0].size() / (N * D);
```
mcwitt, line 338: nit: could this be `n_free_particles` or similar?

mcwitt (approved):
Looks good! Only minor nits.

timemachine/cpp/src/context.cu
```diff
@@ -135,6 +139,83 @@ std::array<std::vector<double>, 2> Context::multiple_steps_local(
     return std::array<std::vector<double>, 2>({h_x_buffer, h_box_buffer});
 }
 
+std::array<std::vector<double>, 2> Context::multiple_steps_local_selection(
```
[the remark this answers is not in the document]
badisa: Yes, think this should be unified in a follow up. As @proteneer mentioned in https://github.com/proteneer/timemachine/pull/1005#issuecomment-1511522091 this class is getting a bit unwieldy.

timemachine/cpp/src/wrap_kernels.cpp
badisa: If I understand correctly, you are asking if we could only return the coordinates that actually change? If so we could, not sure the performance would be worth it, but would have to investigate more thoroughly.

mcwitt: No, sorry, I was just suggesting a more descriptive name instead of `F` (but maybe I misunderstood what `F` represents?)

tests/test_md.py
```diff
@@ -640,6 +692,127 @@ def test_multiple_steps_local_no_free_particles(self):
...
+        ctxt.initialize_local_md()
+        comp_local_xs, comp_local_boxes = ctxt.multiple_steps_local(steps, local_idxs)
+
+        np.testing.assert_array_equal(ref_local_xs, comp_local_xs)
+        np.testing.assert_array_equal(ref_local_boxes, comp_local_boxes)
+
+    def test_local_md_with_selection_mask(self):
+        """Verify that running local md with a selection mask works as expected"""
+        seed = 2023
+        np.random.seed(seed)
```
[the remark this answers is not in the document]
badisa: Yes, it is. Needed because of `prepare_nb_system` which uses the global rng. 

Will switch to the global rng.

timemachine/cpp/src/wrap_kernels.cpp
mcwitt: Ah, I see, `F` is frames and we use it elsewhere. No action needed.

badisa: F = # number of frames, but can switch it to `num_frames`

tests/test_md.py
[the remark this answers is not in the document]
badisa: Was changed in a subsequent commit: https://github.com/proteneer/timemachine/blob/e5f18c6d278dd64575877e9a8212a502847f91f3/tests/test_md.py#L752

tests/test_md.py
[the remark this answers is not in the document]
badisa: Removed in a subsequent commit: https://github.com/proteneer/timemachine/blob/e5f18c6d278dd64575877e9a8212a502847f91f3/tests/test_md.py#L805

tests/test_md.py
badisa: Since this is matching the value that comes out of C++ from https://github.com/proteneer/timemachine/blob/feature/local-selection-mask-md/timemachine/cpp/src/local_md_utils.cu#L160 would need to change that as well. Will look at converting the double to scientific notation.

mcwitt: Ah, I see. Sorry, this comment was a bit too hasty. If it's easy to change, that would be great, but I don't think it's a big deal.

timemachine/cpp/src/context.cu
```diff
@@ -79,9 +85,7 @@ std::array<std::vector<double>, 2> Context::multiple_steps_local(
         d_box_traj.reset(new DeviceBuffer<double>(box_buffer_size));
     }
 
-    if (!this->local_md_pots_) {
-        this->local_md_pots_.reset(new LocalMDPotentials(N_, bps_));
-    }
+    this->initialize_local_md();
```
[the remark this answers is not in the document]
badisa: Changed in https://github.com/proteneer/timemachine/pull/1005/commits/e3b2d8afd8c086a95163c71257ec66548615b26c

tests/test_md.py
badisa: Changed in [dbe3593](https://github.com/proteneer/timemachine/pull/1005/commits/dbe359392358e7c4893c838f9c40ff338b995655)

PR #1006, opened 2023-04-18
Revert to openeye-toolkit 2020.2.0

jkausrelay:
- Revert to openeye toolkit 2020.2.0 to deal with issue where certain ligands fail to have charges assigned


requirements.txt
```diff
@@ -5,3 +5,5 @@ pyyaml==5.4.1
 networkx==2.8.8
 matplotlib==3.7.1
 rdkit==2022.3.5
+--extra-index-url https://pypi.anaconda.org/OpenEye/simple
+openeye-toolkits==2020.2.0
```
badisa, line 9: Should we add `openeye-toolkits` to the setup.py and mention to use the custom pypi in the README?

badisa (approved):

requirements.txt
jkausrelay: For simplicity, let's leave it as if. If we need to separate the toolkit install we can do this later.

mcwitt (approved):

PR #1007, opened 2023-04-24
Fixes title of bar plots

badisa:
* Adds test to verify titles are as expected


mcwitt:
Good catch! I found one more place that might need an update: https://github.com/proteneer/timemachine/blob/984f48ee76e2364e29cb78819a3fdfc3cdb9b15d/tests/test_benchmark.py#L543

PR #1009, opened 2023-04-24
Log Flat Bottom Bond

proteneer:
This PR implements a log flat bottom bond and its associated derivatives:

```
-1/beta * log(1-exp(-beta*U_fb(x)))
```

(To be integrated later into Local MD)


tests/test_bonded.py
```diff
@@ -167,6 +168,58 @@ def test_flat_bottom_bond(self, n_particles=64, n_bonds=35, dim=3):
             np.testing.assert_array_equal(test_du_dx, test_du_dx_rev)
             np.testing.assert_array_equal(test_du_dp, test_du_dp_rev)
 
+    def test_log_flat_bottom_bond(self, n_particles=64, n_bonds=35, dim=3):
+        """Randomly connect pairs of particles, then validate the resulting LogFlatBottomBond force"""
+        np.random.seed(2022)
+
+        # TODO(deboggle) : reduce code duplication between HarmonicBond and LogFlatBottomBond
```
maxentile, line 175:
```suggestion
        # TODO(deboggle) : reduce code duplication between HarmonicBond, FlatBottomBond, and LogFlatBottomBond
```

(implementation more closely follows https://github.com/proteneer/timemachine/pull/764 than HarmonicBond -- both introduce a little duplication / variation, could possibly refactor `GenericBondForce` or similar to template on u(r, p), du_dr(r, p), du_dp(r, p) or something in the future)

tests/test_bonded.py
```diff
@@ -167,6 +168,58 @@ def test_flat_bottom_bond(self, n_particles=64, n_bonds=35, dim=3):
...
+        box = np.eye(3) * 100
+        x = self.get_random_coords(n_particles, dim)
+
+        atom_idxs = np.arange(n_particles)
+
+        k = np.random.rand(n_bonds) * 10  # k not too large to avoid exp(-inf)
```
maxentile, line 181: Is this important? How big can k be?

Applying the assumption that these terms will keep the central particle nearly fixed (consequence of crowded environment), we have been we have been considering values of k not constrained by this consideration, and have typically been setting k a few orders of magnitude larger than this.

Also, worth noting that exp(-inf) for all r < r_max, for any k > 0, so there is no "safe" value of k in this regard...

timemachine/cpp/src/kernels/k_log_flat_bottom_bond.cuh
```diff
@@ -0,0 +1,112 @@
+#include "../fixed_point.hpp"
+#include "k_fixed_point.cuh"
+#include "k_flat_bottom_bond.cuh"
+
+// branchless implementation of piecewise function
```
maxentile, line 5:
```suggestion
```

timemachine/cpp/src/kernels/k_log_flat_bottom_bond.cuh
```diff
@@ -0,0 +1,112 @@
...
+template <typename RealType>
+RealType __device__ __forceinline__
+compute_log_flat_bottom_energy(RealType k, RealType r, RealType rmin, RealType rmax, RealType beta) {
+    RealType r_gt_rmax = static_cast<RealType>(r > rmax);
+    RealType r_lt_rmin = static_cast<RealType>(r < rmin);
```
maxentile, line 10: Internal to FlatBottomBond -- not needed here?

```suggestion
```

timemachine/potentials/bonded.py
```diff
@@ -231,5 +231,25 @@ def flat_bottom_bond(conf, params, box, bond_idxs):
...
+    return jnp.sum(bond_nrgs)
+
+
+def log_flat_bottom_bond(conf, params, box, bond_idxs, beta):
+    """
+    Implements the log inverse of the flat bottom potential
+    """
+    nrgs = _flat_bottom_bond_impl(conf, params, box, bond_idxs)
+    log_nrgs = -jnp.log(1 - jnp.exp(-beta * nrgs))
+    # note the extra 1/beta is to be consistent with the flat bottom bond
```
maxentile, line 254: Clarify?

(Is this to be consistent with some arbitrary convention, or is this to have units of energy?)

timemachine/cpp/src/log_flat_bottom_bond.cu
```diff
@@ -0,0 +1,79 @@
...
+        auto src = bond_idxs[b * 2 + 0];
+        auto dst = bond_idxs[b * 2 + 1];
+        if (src == dst) {
+            throw std::runtime_error("src == dst");
+        }
+
+        if ((src < 0) or (dst < 0)) {
+            throw std::runtime_error("idxs must be non-negative");
+        }
+    }
```
maxentile, line 29: (deboggle / future) duplication of index validation across all 3 bond forces

timemachine/cpp/src/log_flat_bottom_bond.cu
```diff
@@ -0,0 +1,79 @@
...
+
```
maxentile, line 30: should we also validate `beta`?

(assert non-negative, warn if not within 10% or so of room temperature in expected units?)

timemachine/cpp/src/kernels/k_log_flat_bottom_bond.cuh
```diff
@@ -0,0 +1,112 @@
...
+    RealType nrg = compute_flat_bottom_energy(k, r, rmin, rmax);
+    RealType log_nrg = -log(1 - exp(-beta * nrg)) / beta;
```
maxentile, line 12: Haven't thought about this carefully regarding overflow:
(1) do any precautions need to be taken, e.g. https://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf
(2) roughly what range of `nrg` will cause overflow at our default fixed-point settings?

-------

Also, nit: naming -- `log_nrg` could be misleading, does this not have units of energy (kJ/mol)?

maxentile (approved):
Looks good!

tests/test_bonded.py
proteneer: Turns out this test still passes for ```k=np.random.rand(n_bonds)*1000``` - just wanted to conservative here. Will adjust to k =

timemachine/cpp/src/kernels/k_log_flat_bottom_bond.cuh
```diff
@@ -0,0 +1,112 @@
+#include "../fixed_point.hpp"
+#include "k_fixed_point.cuh"
+#include "k_flat_bottom_bond.cuh"
+
+// branchless implementation of piecewise function
+template <typename RealType>
+RealType __device__ __forceinline__
+compute_log_flat_bottom_energy(RealType k, RealType r, RealType rmin, RealType rmax, RealType beta) {
```
badisa, line 8: Is this function used? Appears that the kernel calls `compute_flat_bottom_energy`?

timemachine/potentials/bonded.py
proteneer: Correct - we need the energies to have units of energy (so we can write the deselection term to also have a beta pre-pended to the log). This is also to ensure so that energies, forces, du/dps, etc. are all consistent with the other potentials.

timemachine/cpp/src/kernels/k_log_flat_bottom_bond.cuh
proteneer: I've implemented equation 7 - I think the paper raises good points, since we do have cases where x=0 (when we're at the flat bottom), and where x is large-ish (where we are hitting the repulsive barrier).

tests/test_bonded.py
badisa: Not for this PR, but we will need to adjust the limit for local MD if we think there is an upper bound on the stability of this (https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/local_md_utils.cu#L159)

timemachine/cpp/src/log_flat_bottom_bond.cu
maxentile: Fair enough

(didn't mean to suggest an assertion for temperature, but just a warning. I think it is surprising for a potential energy function to accept `beta` (so it might be easier than other places to introduce unit errors or mistake what the parameter does). We use warnings elsewhere to detect possible temperature / pressure unit bugs

https://github.com/proteneer/timemachine/blob/88da6f3e57dc55b23f582787ba44f67599f6ee11/timemachine/cpp/src/barostat.cu#L28-L35
)

badisa (approved):

timemachine/cpp/src/log_flat_bottom_bond.cu
maxentile: > surprising for a potential energy function to accept beta

Sorry, to clarify: it is required and makes perfect sense for this function to accept `beta`, just it may be surprising to a user of the function (so may merit an extra warning).

(There's a note in the confluence page about whether we want to specify temperature in one place or another, I think the convention @badisa and I selected is preferable to the alternative, but neither alternative considered was perfect.)

timemachine/cpp/src/kernels/k_log_flat_bottom_bond.cuh
```diff
@@ -0,0 +1,112 @@
...
+    RealType r = sqrt(r2);
+
+    // branches -> masks
+    RealType r_gt_rmax = static_cast<RealType>(r > rmax);
+    RealType r_lt_rmin = static_cast<RealType>(r < rmin);
+
+    RealType nrg = compute_flat_bottom_energy(k, r, rmin, rmax);
+
+    if (u) {
+        RealType u_real = -log(1 - exp(-beta * nrg)) / beta;
```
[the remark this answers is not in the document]
proteneer: this will be replaced by the stable log exp

timemachine/potentials/bonded.py
proteneer: added a comment

timemachine/cpp/src/kernels/k_log_flat_bottom_bond.cuh
proteneer: To address #2) I don't think we have to worry about overflows for large energies, but forces may be problematic for very small values of energy:

``` cpp

#include <iostream>
#include <cmath>

using namespace std;

#define FIXED_EXPONENT 0x1000000000


template <typename RealType> unsigned long long FLOAT_TO_FIXED_BONDED(RealType v) {
    return static_cast<unsigned long long>(llrintf(v * FIXED_EXPONENT));
}

template <typename RealType> RealType FIXED_TO_FLOAT(unsigned long long v) {
    return static_cast<RealType>(static_cast<long long>(v)) / FIXED_EXPONENT;
}

template <typename RealType> RealType identity(RealType v) {
    return FIXED_TO_FLOAT<RealType>(FLOAT_TO_FIXED_BONDED<RealType>(v));
}



template <typename RealType> RealType stable_log_1_exp_neg(RealType x) {
    const RealType LOG_2 = 0.693147180559945309417232121;
    return x < LOG_2 ? log(-expm1(-x)) : log1p(-exp(-x));
}


template <typename RealType> RealType fn(RealType nrg) {
    RealType beta = 1.0;
    // return -log(1 - exp(-beta * nrg)) / beta;
    return identity(-stable_log_1_exp_neg(beta*nrg)/beta);
}


template <typename RealType> RealType fn_grad_exact(RealType nrg) {
    RealType beta = 1.0;
    return -exp(-beta * nrg) / (1 - exp(-beta * nrg));
}


template <typename RealType> RealType fn_grad_fixed(RealType nrg) {
    return identity(fn_grad_exact(nrg));
}


int main() {
    std::cout << fn(0.0) << " " << fn_grad_exact(0.0) << " " << fn_grad_fixed(0.0) << std::endl;
    std::cout << fn(1e-10) << " " << fn_grad_exact(1e-10) << " " << fn_grad_fixed(1e-10) << std::endl;
    std::cout << fn(1e-9) << " " << fn_grad_exact(1e-9) << " " << fn_grad_fixed(1e-9) << std::endl;
    std::cout << fn(1e-8) << " " << fn_grad_exact(1e-8) << " " << fn_grad_fixed(1e-8) << std::endl;    
    std::cout << fn(1e-7) << " " << fn_grad_exact(1e-7) << " " << fn_grad_fixed(1e-7) << std::endl;    
    std::cout << fn(1e-6) << " " << fn_grad_exact(1e-6) << " " << fn_grad_fixed(1e-6) << std::endl;    
    std::cout << fn(1e-5) << " " << fn_grad_exact(1e-5) << " " << fn_grad_fixed(1e-5) << std::endl;
    std::cout << fn(1e-4) << " " << fn_grad_exact(1e-4) << " " << fn_grad_fixed(1e-4) << std::endl;
    std::cout << fn(1e-3) << " " << fn_grad_exact(1e-3) << " " << fn_grad_fixed(1e-3) << std::endl;
    std::cout << fn(1e-2) << " " << fn_grad_exact(1e-2) << " " << fn_grad_fixed(1e-2) << std::endl;
    std::cout << fn(1e-1) << " " << fn_grad_exact(1e-1) << " " << fn_grad_fixed(1e-1) << std::endl;
    std::cout << fn(1.0) << " " << fn_grad_exact(1.0) << " " << fn_grad_fixed(1.0) << std::endl;
    std::cout << fn(10.0) << " " << fn_grad_exact(10.0) << " " << fn_grad_fixed(10.0) << std::endl;
    std::cout << fn(100.0) << " " << fn_grad_exact(100.0) << " " << fn_grad_fixed(100.0) << std::endl;
    std::cout << fn(1000.0) << " " << fn_grad_exact(1000.0) << " " << fn_grad_fixed(1000.0) << std::endl;
}
```

Results in:

```
-1.34218e+08 -inf -1.34218e+08
23.0259 -1e+10 -1.34218e+08
20.7233 -1e+09 -1.34218e+08 <-- overflow
18.4207 -1e+08 -1e+08
16.1181 -1e+07 -1e+07
13.8155 -1e+06 -1e+06
11.5129 -99999.5 -99999.5
9.21039 -9999.5 -9999.5
6.90826 -999.5 -999.5
4.61017 -99.5008 -99.5008
2.35217 -9.50833 -9.50833
0.458675 -0.581977 -0.581977
4.5401e-05 -4.5402e-05 -4.5402e-05
0 -3.72008e-44 0
0 -0 0
```

timemachine/cpp/src/kernels/k_log_flat_bottom_bond.cuh
```diff
@@ -0,0 +1,107 @@
...
+    const RealType LOG_2 = 0.693147180559945309417232121;
+    return x < LOG_2 ? log(-expm1(-x)) : log1p(-exp(-x));
+}
+
+template <typename RealType>
+void __global__ k_log_flat_bottom_bond(
+    const int B, // number of bonds
+    const double *__restrict__ coords,
+    const double *__restrict__ box,
+    const double *__restrict__ params, // [P, 3]
```
badisa, line 15: No P defined here, is params actually [B, 3]?

timemachine/cpp/src/kernels/k_log_flat_bottom_bond.cuh
maxentile: Thanks!

>I don't think we have to worry about overflows for large energies, but forces may be problematic for very small values of energy:

That looks perfect to me, even at small input values.

(Spot-on down to 1e-8 inputs. At even smaller inputs, gets the right sign, big magnitude. I can't think of an application that would be sensitive to the difference between -inf and -1e+8.)

proteneer: > gets the right sign

I can't guarantee this for *all* inputs, just the ones I've checked. But I'm not as worried here, because we're not evaluates samples generated without the restraints using hamiltonians *including* the restraint (unlike the nonbonded kernels). If this becomes a problem later on I'll be happy to do more sensitivity analysis, I think this is sufficient for now.

timemachine/cpp/src/kernels/k_log_flat_bottom_bond.cuh
proteneer: good catch - I'll fix this for both the flat_bottom and the harmonic_bond.

tests/test_bonded.py
proteneer: +1 - good call

PR #1011, opened 2023-04-26
Allow user to set col_atom_idxs for NonbondedInteractionGroup

jkausrelay:
- Update NonbondedInteractionGroup to accept col_atom_idxs
- If not set, default to current behavior of using all non-row idxs as the col idxs
- Update tests


timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -1077,24 +1077,37 @@ template <typename RealType> void declare_nonbonded_interaction_group(py::module
     std::string pyclass_name = std::string("NonbondedInteractionGroup_") + typestr;
     py::class_<Class, std::shared_ptr<Class>, timemachine::Potential>(
         m, pyclass_name.c_str(), py::buffer_protocol(), py::dynamic_attr())
-        .def("set_atom_idxs", &timemachine::NonbondedInteractionGroup<RealType>::set_atom_idxs, py::arg("atom_idxs"))
+        .def(
+            "set_atom_idxs",
+            &timemachine::NonbondedInteractionGroup<RealType>::set_atom_idxs,
+            py::arg("atom_idxs"),
+            py::arg("col_atom_idxs") = py::none())
```
proteneer, line 1084: maybe just break API and rename `atom_idxs` to `row_atom_idxs`?

proteneer: I'm actually in favor of `row` and `col` here still, because the underlying implementation is literally a square symmetric `NxN` matrix with distinct `rows` and `cols`, everything internally (including neighborlists), explicitly refer to `rows` and `cols` with different semantics, eg. `row` idxs are contiguous in ordering and `col` idxs. We've never used `group_1` or `group_2` in any other context before, and I'd be hesitant to introduce new notation.

proteneer: Anyways - I can see pros and cons of each @jkausrelay I'll let you decide.

jkausrelay: - Decided to go with row/col as that is consistent with the rest of the code (NC/NR, etc)

timemachine/cpp/src/nonbonded_interaction_group.hpp
```diff
@@ -69,13 +70,14 @@ template <typename RealType> class NonbondedInteractionGroup : public Potential
     void set_atom_idxs_device(
         const int NC, const int NR, unsigned int *d_column_idxs, unsigned int *d_row_idxs, const cudaStream_t stream);
 
-    void set_atom_idxs(const std::vector<int> &atom_idxs);
+    void set_atom_idxs(const std::vector<int> &row_atom_idxs, const std::optional<std::set<int>> &col_atom_idxs);
```
mcwitt, line 73: Should this file `#include <optional>`? (I get compilation errors locally without this)

timemachine/potentials/potentials.py
```diff
@@ -161,18 +161,24 @@ class NonbondedInteractionGroup(Potential):
...
     disable_hilbert_sort: bool = False
     nblist_padding: float = 0.1
 
     def __call__(self, conf: Conf, params: Params, box: Optional[Box]) -> float | Array:
         num_atoms, _ = jnp.array(conf).shape
 
+        # If not set, col_atom_idxs are all others not in the row
+        col_atom_idxs = self.col_atom_idxs
+        if col_atom_idxs is None:
+            col_atom_idxs = np.setdiff1d(jnp.arange(num_atoms), self.row_atom_idxs)
```
mcwitt, line 174: nit: could this logic be moved to the reference potential implementation?

timemachine/cpp/generate_stubs
```diff
@@ -8,3 +8,4 @@ stubgen --output . --module "$MODULE"
 # Workaround for
 # > error: "ndarray" expects 2 type arguments, but 1 given
 sed -i 's/numpy\.ndarray\[\([^]]*\)\]/numpy.typing.NDArray[\1]/g' "${MODULE}.pyi"
+sed -i 's/from typing import List/from typing import List, Set/g' "${MODULE}.pyi"
```
mcwitt, line 11: Nice workaround! Unfortunate that this seems necessary.

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -41,17 +42,33 @@ NonbondedInteractionGroup<RealType>::NonbondedInteractionGroup(
...
         throw std::runtime_error("row_atom_idxs must be nonempty");
     }
     if (row_atom_idxs.size() == static_cast<long unsigned int>(N)) {
-        throw std::runtime_error("must be less then N(" + std::to_string(N) + ") indices");
+        throw std::runtime_error("must be less then N(" + std::to_string(N) + ") row indices");
     }
     verify_atom_idxs(N_, row_atom_idxs);
 
+    if (col_atom_idxs) {
+        std::vector<int> col_atom_idxs_h(col_atom_idxs->begin(), col_atom_idxs->end());
```
mcwitt, line 57: nit: I think in other places we `h` (or `d`) is a prefix instead of a suffix?

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -220,11 +237,16 @@ void NonbondedInteractionGroup<RealType>::execute_device(
 
     const int tpb = warp_size;
     const int B = ceil_divide(N_, tpb);
+    const int K_B = ceil_divide(K_, tpb);
 
     // (ytz) see if we need to rebuild the neighborlist.
-    k_check_rebuild_coords_and_box<RealType>
-        <<<B, tpb, 0, stream>>>(N_, d_x, d_nblist_x_, d_box, d_nblist_box_, nblist_padding_, d_rebuild_nblist_);
 
+    // TODO: Is this more efficient as a single call (K_, d_atom_idxs_, ...)?
```
mcwitt, line 244: Seems like it might be possible to avoid the overhead of an extra kernel launch by making `d_row_indices` and `d_col_indices` contiguous. But from what I understand the overhead is pretty small (~20 us if I'm reading [this](https://developer.nvidia.com/blog/understanding-the-visualization-of-overhead-and-latency-in-nsight-systems/) correctly), so it may not be worth it. @badisa or @proteneer might know better.

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -220,11 +237,16 @@ void NonbondedInteractionGroup<RealType>::execute_device(
 
     const int tpb = warp_size;
     const int B = ceil_divide(N_, tpb);
+    const int K_B = ceil_divide(K_, tpb);
```
mcwitt, line 240: nit: the naming of `K_B` seems a bit confusing to me. It's really a number of blocks, right? (would `B_K` make more sense?)

timemachine/cpp/src/nonbonded_interaction_group.cu
badisa: Would be nice to avoid this, though as @mcwitt said its not a big deal.

We could put the row indices at the start of an array and the column in the rest of array and use NC/NR to point at them specifically. Since the number of total indices can't exceed N. Then you could have a single call here. Would require re-working more of the potential.

Though not sure I entirely see a case where these two kernel calls produces significantly fewer NBlist rebuilds, so using ` k_check_rebuild_coords_and_box`  might also be reasonable.

timemachine/cpp/src/nonbonded_interaction_group.hpp
```diff
@@ -69,13 +70,14 @@ template <typename RealType> class NonbondedInteractionGroup : public Potential
...
 
     NonbondedInteractionGroup(
         const int N,
         const std::vector<int> &row_atom_idxs,
         const double beta,
         const double cutoff,
+        const std::optional<std::set<int>> &col_atom_idxs,
```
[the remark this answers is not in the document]
badisa: I would argue for `vectors` in the C++. The set behavior is really for validation of indices which is handled by `verify_atom_indices`

tests/nonbonded/test_consistency.py
```diff
@@ -48,6 +51,14 @@ def test_nonbonded_consistency(
     ligand_idxs = rng.choice(num_atoms, size=(num_atoms_ligand,), replace=False).astype(np.int32)
     host_idxs = np.setdiff1d(np.arange(num_atoms), ligand_idxs).astype(np.int32)
 
+    # Partition host_idxs into num_ixn_groups
+    if num_ixn_groups:
+        group_idxs = rng.choice(num_ixn_groups, size=(len(host_idxs),), replace=True).astype(np.int32)
+        col_atom_idxss = [host_idxs[group_idxs == i] for i in range(num_ixn_groups)]
```
proteneer, line 57: for my own sanity, maybe rename to `all_col_atom_idxs` to indicate that its a list of `col_atom_idxs`

tests/nonbonded/test_nonbonded_interaction_group.py
```diff
@@ -23,7 +23,15 @@ def test_nonbonded_interaction_group_invalid_indices():
...
         NonbondedInteractionGroup(3, [0, 1, 2], 1.0, 1.0).to_gpu(np.float64).unbound_impl
-    assert "must be less then N(3) indices" == str(e.value)
+    assert "must be less then N(3) row indices" == str(e.value)
+
+    with pytest.raises(RuntimeError) as e:
+        NonbondedInteractionGroup(3, [0, 1], 1.0, 1.0, col_atom_idxs=[2, 3, 4]).to_gpu(np.float64).unbound_impl
+    assert "must be less then N(3) col indices" == str(e.value)
+
+    with pytest.raises(RuntimeError) as e:
+        NonbondedInteractionGroup(3, [0, 1], 1.0, 1.0, col_atom_idxs=[0, 1]).to_gpu(np.float64).unbound_impl
```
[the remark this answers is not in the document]
mcwitt: Ah sorry, I meant to request a test for `row_atom_idxs`, `col_atom_idxs` that are different but overlapping (e.g. have one common index). I didn't see that this is tested?

timemachine/cpp/generate_stubs
proteneer: I have no idea what's going on here

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -23,9 +23,10 @@ NonbondedInteractionGroup<RealType>::NonbondedInteractionGroup(
     const std::vector<int> &row_atom_idxs,
     const double beta,
     const double cutoff,
+    const std::optional<std::set<int>> &col_atom_idxs,
     const bool disable_hilbert_sort,
     const double nblist_padding)
-    : N_(N), NR_(row_atom_idxs.size()), NC_(N_ - NR_),
+    : N_(N), NR_(row_atom_idxs.size()), NC_(col_atom_idxs ? col_atom_idxs->size() : N_ - NR_), K_(NR_ + NC_),
```
proteneer, line 29: does `std::optional<std::set<int> >col_atom_idxs` overload the boolean operator so you can evaluate it like this? This is starting to hurt my brain.

timemachine/cpp/generate_stubs
mcwitt: I think this is to deal with the fact that mypy `stubgen` doesn't seem to add the required imports to `.pyi` files. (Might be unnecessary if we change the type of `col_atom_idxs` from `set` to `vector`?)

timemachine/cpp/src/nonbonded_interaction_group.hpp
```diff
@@ -4,16 +4,18 @@
...
 namespace timemachine {
 
 template <typename RealType> class NonbondedInteractionGroup : public Potential {
 
 private:
-    const int N_; // N_ = NC_ + NR_
+    const int N_; // total number of atoms, i.e. first dimension of input coords, params
     int NR_;      // number of row atoms
     int NC_;      // number of column atoms
+    int K_;       // number of interacting atoms, K_ = NC_ + NR_
```
proteneer, line 18: Can we turn this into a method? i.e. `get_num_interacting_atoms()` - generally we'd like to avoid creating another member that can be computed from two other members.

proteneer: (especially since `NR_` and `NC_` can change I think? and we'd like to avoid bookkeeping of keeping it synced s.t. `K_=NR_+NC_` for all values)

timemachine/cpp/src/nonbonded_interaction_group.hpp
```diff
@@ -69,13 +71,14 @@ template <typename RealType> class NonbondedInteractionGroup : public Potential
...
 
-    void set_atom_idxs(const std::vector<int> &atom_idxs);
+    void set_atom_idxs(const std::vector<int> &row_atom_idxs, const std::optional<std::vector<int>> &col_atom_idxs);
 
     NonbondedInteractionGroup(
         const int N,
         const std::vector<int> &row_atom_idxs,
         const double beta,
         const double cutoff,
+        const std::optional<std::set<int>> &col_atom_idxs,
```
proteneer, line 81: I really don't want an `optional` for something this low-level, I'd rather the signature just be:
``` c++
NonbondedInteractionGroup(
        const int N,
        const std::vector<int> &row_atom_idxs,
        const std::vector<int> &col_atom_idxs,
        const double beta,
        const double cutoff
  )```

File: tests/nonbonded/test_nonbonded_interaction_group.py [hunk 14]
```diff
@@ -54,7 +65,9 @@ def test_nonbonded_interaction_group_zero_interactions(rng: np.random.Generator)
 @pytest.mark.parametrize("precision,rtol,atol", [(np.float64, 1e-8, 1e-8), (np.float32, 1e-4, 5e-4)])
 @pytest.mark.parametrize("num_atoms_ligand", [1, 15])
 @pytest.mark.parametrize("num_atoms", [33, 231])
+@pytest.mark.parametrize("num_col_atoms", [0, 1, 10, -1])
```
proteneer, line 68: What does -1 here do?

proteneer, line 68: Add a test for num_atoms > 32 just to test for multi block / multi warp correctness

timemachine/cpp/src/nonbonded_interaction_group.cu
jkausrelay: Yes it does, but removed in the next commit

timemachine/cpp/generate_stubs
jkausrelay: Right. This was needed due to a missing import, removed since we are using vector now.

timemachine/cpp/src/local_md_utils.cu
```diff
@@ -62,17 +62,18 @@ construct_ixn_group_potential(const int N, std::shared_ptr<Potential> pot, const
     gpuErrchk(cudaMemcpy(&h_params[0], d_params, P * sizeof(*d_params), cudaMemcpyDeviceToHost));
     std::vector<int> dummy_idxs{0};
     std::vector<int> shape{P};
+    std::optional<std::set<int>> dummy_col_idxs(std::nullopt);
```
badisa, line 65: No longer used

timemachine/cpp/src/local_md_utils.cu
```diff
@@ -62,17 +62,18 @@ construct_ixn_group_potential(const int N, std::shared_ptr<Potential> pot, const
...
 
     if (std::shared_ptr<NonbondedAllPairs<float>> nb_pot = std::dynamic_pointer_cast<NonbondedAllPairs<float>>(pot);
         nb_pot) {
         std::shared_ptr<Potential> ixn_group(
-            new NonbondedInteractionGroup<float>(N, dummy_idxs, nb_pot->get_beta(), nb_pot->get_cutoff()));
+            new NonbondedInteractionGroup<float>(N, dummy_idxs, dummy_idxs, nb_pot->get_beta(), nb_pot->get_cutoff()));
```
badisa, line 69: It looks like this should fail due to this check
https://github.com/proteneer/timemachine/blob/split_ff_solv_ixn_gpu/timemachine/cpp/src/nonbonded_interaction_group.cu#L63-L68

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -41,17 +42,31 @@ NonbondedInteractionGroup<RealType>::NonbondedInteractionGroup(
                     &k_nonbonded_unified<RealType, 1, 1, 0>,
                     &k_nonbonded_unified<RealType, 1, 1, 1>}),
 
-      beta_(beta), cutoff_(cutoff), nblist_(N_), nblist_padding_(nblist_padding), d_sort_storage_(nullptr),
+      beta_(beta), cutoff_(cutoff), nblist_(NR_ + NC_), nblist_padding_(nblist_padding), d_sort_storage_(nullptr),
```
badisa, line 45: For Local MD this change of `NR_ + NC_` will likely cause failures, as it reuse the potential with different sizes of indices, I would probably keep it `N_`. Otherwise, just need to modify https://github.com/proteneer/timemachine/blob/split_ff_solv_ixn_gpu/timemachine/cpp/src/local_md_utils.cu#L63 to construct ixn group that has size N


```suggestion
      beta_(beta), cutoff_(cutoff), nblist_(N_), nblist_padding_(nblist_padding), d_sort_storage_(nullptr),
```

proteneer (approved):
overall lgtm, pending tests

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -119,13 +119,7 @@ NonbondedInteractionGroup<RealType>::NonbondedInteractionGroup(
...
     // estimate size needed to do radix sorting, this can use uninitialized data.
     cub::DeviceRadixSort::SortPairs(
-        d_sort_storage_,
-        d_sort_storage_bytes_,
-        d_sort_keys_in_,
-        d_sort_keys_out_,
-        d_sort_vals_in_,
-        d_perm_,
-        std::max(NC_, NR_));
+        d_sort_storage_, d_sort_storage_bytes_, d_sort_keys_in_, d_sort_keys_out_, d_sort_vals_in_, d_perm_, N_);
```
jkausrelay, line 122: @badisa I also changed this to use N_, otherwise this would overflow w/local MD when initialized only with the single dummy row/col idxs.

timemachine/cpp/src/local_md_utils.cu
```diff
@@ -3,6 +3,7 @@
 #include "nonbonded_interaction_group.hpp"
 
 #include "fanout_summed_potential.hpp"
+#include "set_utils.hpp"
```
badisa, line 6: nit: Unused import

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -346,15 +362,17 @@ void NonbondedInteractionGroup<RealType>::execute_device(
...
-    std::set<unsigned int> unique_row_atom_idxs(unique_idxs(unsigned_idxs));
-    // compute set of column atoms as set difference
-    std::vector<unsigned int> col_atom_idxs_v = get_indices_difference(N_, unique_row_atom_idxs);
+void NonbondedInteractionGroup<RealType>::set_atom_idxs(
+    const std::vector<int> &row_atom_idxs, const std::vector<int> &col_atom_idxs) {
+    verify_atom_idxs(N_, row_atom_idxs, true);
+    verify_atom_idxs(N_, col_atom_idxs, true);
+
+    std::vector<unsigned int> unsigned_row_idxs = std::vector<unsigned int>(row_atom_idxs.begin(), row_atom_idxs.end());
+    std::set<unsigned int> unique_row_atom_idxs(unique_idxs(unsigned_row_idxs));
```
badisa, line 371: Don't do the same validation here that the constructor does, to make sure they are disjoint.

Also doesn't seem to ensure that the column indices are unique.

jkausrelay: Updated in https://github.com/proteneer/timemachine/pull/1011/commits/4c1a68a05f88abc2836f5719ba2a28c8db0db076 to use the same code for validation.

Fo local MD, is it correct to say the initializer should have non-zero NR/NC but set_atom_idxs can have 0 NR or NC? Want to make sure to check the right things here.

badisa: Correct. It is an inconsistency, but for Local MD it is an expectation that it can handle at least N particles, but can also be run on nothing.

badisa (approved):
🚀

mcwitt (approved):

PR #1014, opened 2023-04-28
Adds log restraint to reference for Local MD

badisa:
- Fixes the naming of a function that was accidentally overloaded
- Switches md tests to be pytest to parameterize them cleanly and match convention 
- Enables the reference particle to move
- Repulsive force between all frozen and the reference

- [x] Clean up IxnGroup changes and test
- [x] Add benchmark plots between frozen and unfrozen

All the plots below are with calling `setup_local_md` prior to timings, so no overhead is included in the timings.

image
image
image


timemachine/cpp/src/context.cu
```diff
@@ -22,8 +22,16 @@ Context::Context(
     const double *box_0,
     std::shared_ptr<Integrator> intg,
     std::vector<std::shared_ptr<BoundPotential>> bps,
-    std::shared_ptr<MonteCarloBarostat> barostat)
-    : N_(N), barostat_(barostat), step_(0), d_sum_storage_(nullptr), d_sum_storage_bytes_(0), intg_(intg), bps_(bps) {
+    std::shared_ptr<MonteCarloBarostat> barostat,
+    bool freeze_reference)
```
badisa, line 26: Chose this as a separate parameter so we could test it with and without, not really happy with it especially since if you freeze the reference right now you have to provide a temperature. I rely on the integrator for the temperature, though it reintroduces the issue that the selection idxs PR has fixed for @maxentile

timemachine/cpp/src/nonbonded_interaction_group.cu
```diff
@@ -370,29 +375,30 @@ void NonbondedInteractionGroup<RealType>::set_atom_idxs_device(
     if (NC + NR != N_) {
         throw std::runtime_error("Total of indices must equal N");
     }
-    const size_t tpb = warp_size;
+    if (NR > 0 && NC > 0) {
+        const size_t tpb = warp_size;
```
badisa, line 379: Don't like this, but either we can add and remove the ixn group to the `all_potentials_` of the LocalMDPotentials class, which seems like more complicated bookkeeping though would avoid a potential being a no-op.

badisa: These changes are also likely to conflict with https://github.com/proteneer/timemachine/pull/1011

timemachine/cpp/src/context.cu
maxentile: Is there a way to set context temperature if the integrator is VelocityVerlet?

timemachine/cpp/src/kernels/k_local_md.cuh
```diff
@@ -26,3 +26,7 @@ void __global__ k_construct_bonded_params(
     params[idx * 3 + 1] = r_min;
     params[idx * 3 + 2] = r_max;
 }
+
+template <typename T> void __global__ k_update_indice(T *d_array, std::size_t idx, unsigned int val) {
+    d_array[idx] = val;
+}
```
maxentile, line 32: nit: surprise template -- is it intended to allow T other than unsigned int?

consider

```suggestion
void __global__ k_update_index(unsigned int *d_array, std::size_t idx, unsigned int val) {
    d_array[idx] = val;
}
```

or 

```suggestion
template <typename T> void __global__ k_update_index(T *d_array, std::size_t idx, T val) {
    d_array[idx] = val;
}
```

timemachine/cpp/src/context.cu
badisa: Not at the moment, just wasn't sure the best way to implement this.

Could have it so you can call `ensure_md_is_initialized` with a temperature which is honored if temperature > 0.0, otherwise tries to use the integrator's temperature.

proteneer, line 26: Do we want this to be a parameter in `Context` itself? Since for non-local MD use cases this parameter would be ignored?

timemachine/cpp/src/local_md_potentials.cu
```diff
@@ -199,6 +213,17 @@ void LocalMDPotentials::_setup_free_idxs_given_reference_idx(
     bound_restraint_->set_params_device(std::vector<int>({num_row_idxs, 3}), d_bond_params_.data, stream);
     restraint_->set_bonds_device(num_row_idxs, d_restraint_pairs_.data, stream);
 
+    if (!freeze_reference_) {
+        // Remove the reference idx from the column indices, d_free_idxs gets inverted to construct column idxs,
+        // and add to the row indices
```
proteneer, line 218: This is probably pretty inefficient (launching two kernels with 1 thread 1 block).

Would it be simpler to just modify the partition code above (lines 168-170) itself to simplify always include reference in the `free_idxs`?

timemachine/cpp/src/local_md_potentials.cu
```diff
@@ -55,6 +61,13 @@ LocalMDPotentials::LocalMDPotentials(const int N, const std::vector<std::shared_
     // Add the restraint potential and ixn group potential
     all_potentials_.push_back(bound_restraint_);
```
proteneer, line 62: I would probably do a quick rename here:

`bound_restraint_` -> `bound_free_restraint_`
`reference_bound_restraint_` -> `bound_frozen_restraint_`

timemachine/cpp/src/local_md_potentials.cu
badisa: Had originally done it this way to avoid the reference -> reference restraint thatt would be setup. But since r is always 0 in that case, might actually be okay.

timemachine/cpp/src/kernels/k_local_md.cuh
badisa: Removed in https://github.com/proteneer/timemachine/pull/1014/commits/d8b0b8f5022d0f37036ca8b19d6519b8fd73f412

timemachine/cpp/src/local_md_potentials.cu
badisa: Renamed in https://github.com/proteneer/timemachine/pull/1014/commits/19c74bbd25d29ccc608a25086f44b51dd404fdfb

timemachine/cpp/src/local_md_potentials.cu
badisa: Yup, works. Though maybe we should have some worry about being able to construct a restraint here that would be invalid to construct from scratch (we verify src_idx != dst_idx).

Added in https://github.com/proteneer/timemachine/pull/1014/commits/37ea591777808702a5ec9f3657b017554fd09b80

timemachine/cpp/src/context.cu
badisa: It defaults to true, so currently it behaves the same way as it use to.

The idea here is that we want to test these two cases (frozen/free reference) and get a sense of how the two behave. In the longer term, we will likely want to settle on a single implementation and this will go away for better version.

We could also add a separate function such as `ctxt.setup_local_md(temperature, frozen)` that can be called after construction to configure local md a specific way. And as I type this, I am inclined to think that is the way to go. And remove the `ensure_md_is_intialized` and only allow the user to call it once. Seems like a better way and could allow us to maintain both versions (if we wanted to) without additional jank.

maxentile: >We could also add a separate function such as ctxt.setup_local_md(temperature, frozen) that can be called after construction to configure local md a specific way. And as I type this, I am inclined to think that is the way to go. And remove the ensure_md_is_intialized and only allow the user to call it once. Seems like a better way and could allow us to maintain both versions (if we wanted to) without additional jank.

💯

badisa: Reworked in https://github.com/proteneer/timemachine/pull/1014/commits/39087381d71a0f57369c6b08ac785e0442eb19b0

badisa: Reworked so that you can control local md behavior explicitly in https://github.com/proteneer/timemachine/pull/1014/commits/39087381d71a0f57369c6b08ac785e0442eb19b0

tests/test_local_move.py
```diff
@@ -64,6 +64,14 @@ def assert_no_drift(
 
     expected_selection_fraction_traj = np.array([observable_fxn(x) for x in traj])
 
+    # A sanity check that the early samples don't have a massive jump within
+    # in the case of unstable local MD this test can pass neighboring atoms go from ~10% of the system to ~100% of the system
+    # in the first step
+    differences_early = np.diff(expected_selection_fraction_traj[:n_samples]) / expected_selection_fraction_traj[0]
+    assert (
+        np.any(differences_early) < 2.0
```
mcwitt, line 72: This doesn't seem correct?

mcwitt:
> Switches md tests to be pytest to parameterize them cleanly and match convention

:tada:

timemachine/cpp/src/context.hpp
```diff
@@ -72,6 +72,8 @@ class Context {
 
     double _get_temperature();
 
+    void _ensure_local_md_intialized();
```
mcwitt, line 75: nit: should this be `private`?

timemachine/cpp/src/local_md_potentials.cu
```diff
@@ -170,9 +189,9 @@ void LocalMDPotentials::_setup_free_idxs_given_reference_idx(
     gpuErrchk(cudaStreamSynchronize(stream));
 
     // The row indices is all of the free indices, which excludes the reference
-    const int num_row_idxs = p_num_selected_.data[0];
+    int num_row_idxs = p_num_selected_.data[0];
     // The col indices is all indices, except for the free, including the reference which is frozen.
-    const int num_col_idxs = N_ - num_row_idxs;
+    int num_col_idxs = N_ - num_row_idxs;
```
mcwitt, line 194: nit: were these changes intentional? (don't immediately see why these can't be `const`)

timemachine/cpp/src/local_md_potentials.cu
```diff
@@ -220,6 +239,26 @@ void LocalMDPotentials::_setup_free_idxs_given_reference_idx(
     // Free particles should be in the row idxs
     set_nonbonded_ixn_potential_idxs(
         ixn_group_->potential, num_col_idxs, num_row_idxs, d_col_idxs_.data, d_row_idxs_.data, stream);
+
+    if (!freeze_reference_) {
+        // If there are no frozen, don't attach any
```
mcwitt, line 244: nit: clarify comment (should contain "indices"?)

timemachine/cpp/src/context.hpp
badisa: Yes, `setup_local_md` is the preferred way to do this. It makes it explicit and allows modification to the behavior.

`_ensure_local_md_initialized` is just for internal usage, though should it should call `setup_local_md`

timemachine/cpp/src/local_md_potentials.cu
badisa: Was done in a previous implementation, put back in [2a2a934](https://github.com/proteneer/timemachine/pull/1014/commits/2a2a9344ad5eee81fd1d7ae5bf95efeb003d4121)

timemachine/cpp/src/local_md_potentials.cu
badisa: Good catch, fixed up the comment [e0a0819](https://github.com/proteneer/timemachine/pull/1014/commits/e0a0819e25379d3549ca2bc4dfd40b2c71bd127e)

timemachine/cpp/src/context.hpp
badisa: `_ensure_local_md_initialized` now calls `setup_local_md` in [addad67](https://github.com/proteneer/timemachine/pull/1014/commits/addad672327d50ab8b9454af0681fbd80a033d57)

tests/test_local_move.py
badisa: Fixed in [282a845](https://github.com/proteneer/timemachine/pull/1014/commits/282a84539b4b34e2168eb87c0fcaaef4edfe8e75)

proteneer:
One more comment: can we add a test to ensure that with `freeze_reference=True`, exactly 1 particle in `selection_idxs` maintains its position (bitwise), and with `freeze_reference=False`, no particles maintain their position.

tests/test_local_move.py
```diff
@@ -64,6 +64,14 @@ def assert_no_drift(
...
     expected_selection_fraction_traj = np.array([observable_fxn(x) for x in traj])
 
+    # A sanity check that the early samples don't have a massive jump within
+    # in the case of unstable local MD this test can pass neighboring atoms go from ~10% of the system to ~100% of the system
+    # in the first step
+    differences_early = np.diff(expected_selection_fraction_traj[:n_samples]) / expected_selection_fraction_traj[0]
+    assert (
+        np.all(differences_early) < 2.0
+    ), "Difference between first and last sample greater than 200%, likely unstable"
+
```
maxentile, line 74: I don't quite understand the motivation for this assertion -- (1) did this come up when testing including this term, (2) can this assertion be combined with the next assertion?

(this one: check that (f[t] - f[t-1] / f[0]) < 2 for all t. next assertion: compare first and second half of trajectory, check that these don't differ by more than 50% up or down. could maybe modify the second assertion to average over a smaller initial portion of the trajectory.)

tests/test_md.py
```diff
@@ -20,866 +19,891 @@
...
+    np.testing.assert_array_equal(summed_pot_boxes, boxes)
 
-        xs, boxes = ctxt.multiple_steps_local(num_steps, local_idxs, radius=radius)
 
-        assert xs.shape[0] == 1
+@pytest.mark.parametrize("freeze_reference", [True, False])
+def test_multiple_steps_local_entire_system(freeze_reference):
+    """Verify that running multiple_steps_local is valid even when consuming the entire system, IE radius ~= inf.
+
+    - Only a single particle, the reference, should not move
```
maxentile, line 631: "If `freeze_reference, ..."

maxentile (approved):

tests/test_local_move.py
badisa: This came up in testing, but I had seen similar behavior much earlier.

The issue is that occasionally it will be stable for a 'few' steps, but then suddenly explode, resulting in the number of selected particles going from say 1000 to 6000.

Since assertion is looking at the beginning as it verifies that you aren't introducing big differences by switching to local moves. If you are, it seems like it warrants more investigation.

Perhaps doing a running average on the start to make sure the values agree would be better.

badisa:
> One more comment: can we add a test to ensure that with `freeze_reference=True`, exactly 1 particle in `selection_idxs` maintains its position (bitwise), and with `freeze_reference=False`, no particles maintain their position.

Already added here: https://github.com/proteneer/timemachine/blob/addad672327d50ab8b9454af0681fbd80a033d57/tests/test_md.py#L556-L559

proteneer (approved):
lgtm!

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -363,8 +362,8 @@ void declare_context(py::module &m) {
 
         F = iterations / store_x_interval
 
-        The first call to `multiple_steps_local_selection` takes longer than subsequent calls, if ensure_local_md_intialized has not been called previously,
-        initializes potentials needed for local md.
+        The first call to `multiple_steps_local_selection` takes longer than subsequent calls, if setup_local_md has not been local MD called are for local MD.         the
+        integrator, which must the integrator, which must be a LangevinIntegrator. be a LangevinIntegrator.
```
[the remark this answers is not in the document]
badisa: Good catch, fixed in https://github.com/proteneer/timemachine/pull/1014/commits/02b7ca5a5d0bfa46cfbf2632a4dfd99d69c4b695

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -299,7 +298,7 @@ void declare_context(py::module &m) {
             Coordinates have shape (F, N, 3)
             Boxes have shape (F, 3, 3)
 
-        Note: All boxes returned will be identical as local md only runs under NVT.
+        Note: All boxes returned will be identical as local MD only runs under NVT.
```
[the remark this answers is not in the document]
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1014/commits/31b0d157d9c91f9462810fc176e377de95739439

Correct, this is simply to document the current behavior.

tests/test_md.py
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1014/commits/2ad3c38a2a5e185ed36eae984cff7a674d03252a

tests/test_local_move.py
maxentile: Ahh, thanks for the additional context!

This does indicate a limitation of the existing assertion (averaging over too large of an initial window) -- perhaps in the future this could be made more sensitive (similar to the new assertion in this PR), by averaging over a smaller initial window.

For now there's value in having both.

badisa: Minor improvement to the comment and adds abs: https://github.com/proteneer/timemachine/pull/1014/commits/aa47b6aa5f9f4da2769e280d80400c9c730017b2

PR #1015, opened 2023-04-28
Simplify anchor group selection

mcwitt:
Currently, we check whether a bond (anchor, neighbor) is present in mol_a when selecting a suitable anchor group for a given mol_b dummy group. The original motivation was to ensure numerical stability near $\lambda = 0$: since we might have an angle term like (dummy_b, anchor, neighbor), stability of the standard harmonic angle force in particular requires the bond (anchor, neighbor) be present in mol_a.

Since we have moved to using a stable form of the harmonic angle force (#935), this check should no longer be necessary; For  $0 < \lambda < 1$, angle terms are stable even when interatomic distances shrink to zero; at $\lambda = 0$, where we use the standard harmonic angle force, angles involving dummy atoms have zero force constant.

This PR:

* removes logic ensuring that the (anchor, neighbor) bond is present in mol_a when setting up an anchor group for mol_b
* removes now unused `mol_a`, `core_a` arguments from the signature of `find_dummy_groups_and_anchors`
* removes `CoreBondChangeWarning`, which was previously raised when (anchor, neighbor) was missing in mol_a
* updates test

Note: Related to #1016 (but does not yet resolve). The proposed fix for the latter will not perform the check removed here, so testing and merging this first should make the eventual fix simpler to validate.

### DIFF
--- a/tests/test_single_topology.py
+++ b/tests/test_single_topology.py
@@ -92,16 +92,16 @@ def test_find_dummy_groups_and_anchors():
     AllChem.EmbedMolecule(mol_a, randomSeed=2022)
     AllChem.EmbedMolecule(mol_b, randomSeed=2022)
 
-    core_pairs = np.array([[1, 2], [2, 1], [3, 0]])
+    core_atoms = np.array([2, 1, 0])
 
-    dgs, jks = single_topology.find_dummy_groups_and_anchors(mol_b, core_pairs[:, 1])
+    dgs, jks = single_topology.find_dummy_groups_and_anchors(mol_b, core_atoms)
     assert dgs == [{3}]
     assert jks == [(2, 1)]
 
     # angle should swap
-    core_pairs = np.array([[1, 2], [2, 0], [3, 1]])
+    core_atoms = np.array([2, 0, 1])
 
-    dgs, jks = single_topology.find_dummy_groups_and_anchors(mol_b, core_pairs[:, 1])
+    dgs, jks = single_topology.find_dummy_groups_and_anchors(mol_b, core_atoms)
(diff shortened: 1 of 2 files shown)


tests/test_single_topology.py
```diff
@@ -95,17 +94,16 @@ def test_find_dummy_groups_and_anchors():
...
     assert dgs == [{3}]
     assert jks == [(2, 1)]
 
     # angle should swap
     core_pairs = np.array([[1, 2], [2, 0], [3, 1]])
 
-    with pytest.warns(CoreBondChangeWarning):
-        dgs, jks = single_topology.find_dummy_groups_and_anchors(mol_a, mol_b, core_pairs[:, 0], core_pairs[:, 1])
-        assert dgs == [{3}]
-        assert jks == [(2, None)]
```
mcwitt, line 108: This is testing the following case:
test-atom-map

mcwitt: Currently, the angle 3-2-1 is disallowed because the blue-green bond is not present in mol_a. With these changes, it is allowed.

proteneer (approved):
lgtm. Nice simplifications.

tests/test_single_topology.py
proteneer: Hm after staring at this a bit more - we will leave this angle term on at all values of lambda - wouldn't it flop around considerably? (The dummy anchor terms are not alchemically turned off).

mcwitt:
After discussion with @proteneer , closing this in favor of an alternate approach. The issue is that angle terms involving dummy atoms are left on even at $\lambda=0$, so there is no guarantee of numerical stability with this approach.

PR #1017, opened 2023-05-01
Move unused single topology code to attic

### LINKED ISSUE #1016: Disconnected dummy atoms in some cases with complete_rings=False
E.g. methane → cyclobutadiene

With `complete_rings=False`, we generate a mapping like:
methane-cyclobutane-atom-map

Due to the heuristic [here](https://github.com/proteneer/timemachine/blob/03b9dd1e8bb86f06291c941840563a66287475f2/timemachine/fe/dummy.py#L253), we only generate a single dummy group for B containing all 4 dummy atoms. Then, turning off all but one of the anchor bonds to maintain factorizability results in a disconnected dummy atom.

Potential fix: modify to use two passes:

1. Initialize dummy groups as connected components of the subgraph induced by dummy atoms, then 

2. merge groups with a common anchor. In the above example, this would result in 2 dummy groups with no disconnected atoms.

mcwitt:
This code was mostly introduced in [#633](https://github.com/proteneer/timemachine/pull/633) and refined in [#656](https://github.com/proteneer/timemachine/pull/656).

Simplified single topology, introduced in [#756](https://github.com/proteneer/timemachine/pull/756), utilizes only a subset of the original functionality; the remainder is moved to `attic/` in this PR to simplify maintenance.


proteneer (approved):
lgtm! glad to see this attic'd

maxentile (approved):
Looks good, and thanks for adding the context.

(I had missed in reviewing #756 that these functions became unused!)

PR #1018, opened 2023-05-02
Expose nonbonded cutoff in setup_dhfr

maxentile:
Default is lower than the default used in the rest of the project (I think for compatibility with OpenMM DHFR Explicit-RF benchmark), but this introduces a more severe truncation artifact.

(Possibly corrupting some numerical results that assume continuous energies, such as HMC.)


maxentile:
( applied the change directly to https://github.com/proteneer/timemachine/compare/master...local_hmc , probably not needed elsewhere )

PR #1024, opened 2023-05-16
Runs potentials in parallel

badisa:
* Adds a StreamPotentialRunner for handling the logic of running potentials in parallel
* Adds a StreamManager to handle syncing and stream creation/destruction
* Appears to improve performance of local MD by 30%, closer to 5-10% faster for global MD
* Potentials have to increment buffers atomically to work

# Benchmarks
A10 GPU
Cuda Arch 8.6

image
Streams are overlapping as we expect

# Master
```
dhfr-apo: N=23558 speed: 341.40ns/day dt: 1.5fs (ran 100000 steps in 37.96s)
dhfr-apo-barostat-interval-25: N=23558 speed: 313.44ns/day dt: 1.5fs (ran 100000 steps in 41.35s)
dhfr-apo-hmr-barostat-interval-25: N=23558 speed: 521.17ns/day dt: 2.5fs (ran 100000 steps in 41.45s)
hif2a-apo: N=8805 speed: 579.45ns/day dt: 1.5fs (ran 100000 steps in 22.37s)
hif2a-apo-barostat-interval-25: N=8805 speed: 499.71ns/day dt: 1.5fs (ran 100000 steps in 25.94s)
hif2a-rbfe: N=8840 speed: 560.64ns/day dt: 1.5fs (ran 100000 steps in 23.12s)
hif2a-rbfe-local: N=8840 speed: 673.33ns/day dt: 1.5fs (ran 100000 steps in 19.25s)
solvent-apo: N=6282 speed: 725.43ns/day dt: 1.5fs (ran 100000 steps in 17.87s)
solvent-apo-barostat-interval-25: N=6282 speed: 649.74ns/day dt: 1.5fs (ran 100000 steps in 19.95s)
solvent-rbfe: N=6317 speed: 687.67ns/day dt: 1.5fs (ran 100000 steps in 18.85s)
solvent-rbfe-local: N=6317 speed: 683.04ns/day dt: 1.5fs (ran 100000 steps in 18.98s)
```

# PR Changes without parallelizing Summed Potentials
```
dhfr-apo: N=23558 speed: 350.31ns/day dt: 1.5fs (ran 100000 steps in 37.00s)
dhfr-apo-barostat-interval-25: N=23558 speed: 319.06ns/day dt: 1.5fs (ran 100000 steps in 40.62s)
dhfr-apo-hmr-barostat-interval-25: N=23558 speed: 530.35ns/day dt: 2.5fs (ran 100000 steps in 40.73s)
hif2a-apo: N=8805 speed: 591.88ns/day dt: 1.5fs (ran 100000 steps in 21.90s)
hif2a-apo-barostat-interval-25: N=8805 speed: 508.00ns/day dt: 1.5fs (ran 100000 steps in 25.51s)
hif2a-rbfe: N=8840 speed: 625.76ns/day dt: 1.5fs (ran 100000 steps in 20.71s)
hif2a-rbfe-local: N=8840 speed: 907.55ns/day dt: 1.5fs (ran 100000 steps in 14.28s)
solvent-apo: N=6282 speed: 744.57ns/day dt: 1.5fs (ran 100000 steps in 17.41s)
solvent-apo-barostat-interval-25: N=6282 speed: 645.83ns/day dt: 1.5fs (ran 100000 steps in 20.07s)
solvent-rbfe: N=6317 speed: 780.84ns/day dt: 1.5fs (ran 100000 steps in 16.60s)
solvent-rbfe-local: N=6317 speed: 856.33ns/day dt: 1.5fs (ran 100000 steps in 15.14s)
```
# PR Changes with parallelizing Summed Potentials
```
dhfr-apo: N=23558 speed: 352.65ns/day dt: 1.5fs (ran 100000 steps in 36.75s)
dhfr-apo-barostat-interval-25: N=23558 speed: 322.43ns/day dt: 1.5fs (ran 100000 steps in 40.20s)
dhfr-apo-hmr-barostat-interval-25: N=23558 speed: 537.94ns/day dt: 2.5fs (ran 100000 steps in 40.16s)
hif2a-apo: N=8805 speed: 610.21ns/day dt: 1.5fs (ran 100000 steps in 21.24s)
hif2a-apo-barostat-interval-25: N=8805 speed: 521.51ns/day dt: 1.5fs (ran 100000 steps in 24.85s)
hif2a-rbfe: N=8840 speed: 651.41ns/day dt: 1.5fs (ran 100000 steps in 19.90s)
hif2a-rbfe-local: N=8840 speed: 888.21ns/day dt: 1.5fs (ran 100000 steps in 14.60s)
solvent-apo: N=6282 speed: 757.39ns/day dt: 1.5fs (ran 100000 steps in 17.11s)
solvent-apo-barostat-interval-25: N=6282 speed: 665.12ns/day dt: 1.5fs (ran 100000 steps in 19.49s)
solvent-rbfe: N=6317 speed: 801.27ns/day dt: 1.5fs (ran 100000 steps in 16.18s)
solvent-rbfe-local: N=6317 speed: 844.12ns/day dt: 1.5fs (ran 100000 steps in 15.36s)
```


timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -187,7 +187,7 @@ void __global__ k_scatter_accum(
         return;
     }
 
-    array[unique_idxs[idx] * stride + stride_idx] += gathered_array[idx * stride + stride_idx];
+    atomicAdd(array + (unique_idxs[idx] * stride + stride_idx), gathered_array[idx * stride + stride_idx]);
```
badisa, line 190: All potentials need to do atomic adds to the buffers, which would make this a more fragile test. Lots of tests fail without this change.

badisa:
> Question: Is the problem this addresses that kernel launch overhead is no longer negligible compared with kernel execution?

This doesn't address launch overhead, this simply allows multiple potentials to run in parallel. The overhead of launches is going to be addressed in subsequent PRs.

> (If so, docs suggest this may be a problem that CUDA graphs https://developer.nvidia.com/blog/cuda-graphs/ https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#cuda-graphs may be suited for. Could you comment on the stream approach vs. other approaches?)

Based on https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#cross-stream-dependencies-and-events it looks like this could be added to a cuda graph. Think there would need to be some change to our potentials for this to work, as there is work done on the CPU (number of ixns in nonbonded) that would not be represented in the cuda graph, at least without more invasive changes to potentials.

timemachine/cpp/src/kernels/k_nonbonded.cuh
mcwitt: Have you tried running this with the racecheck tool yet? https://docs.nvidia.com/compute-sanitizer/ComputeSanitizer/index.html#racecheck-tool

(not sure if it'd be able to detect this specific issue, but seems like a good idea when introducing parallelism)

mcwitt:
> Potentials wrapped into SummedPotentials will not be parallelized

Could you comment on why this is the case? Is this just because we can't be sure that the components are safe to parallelize? (if that's the case, I wonder if it'd be worth eventually introducing a parallel summed potential)

timemachine/cpp/src/streamed_potential_runner.cu
```diff
@@ -0,0 +1,76 @@
...
+    // Expect stream to be the next increment
+    if (num_streams != index) {
+        throw std::runtime_error("Asked for new index out of order");
+    }
+    cudaStream_t new_stream;
+    // Create stream that doesn't block with the null stream
+    gpuErrchk(cudaStreamCreateWithFlags(&new_stream, cudaStreamNonBlocking));
+
+    cudaEvent_t new_event;
+    // Create stream with timings disabled
```
mcwitt, line 36: nit: expand comment (why are timings disabled?)

badisa:
> > Potentials wrapped into SummedPotentials will not be parallelized
> 
> Could you comment on why this is the case? Is this just because we can't be sure that the components are safe to parallelize? (if that's the case, I wonder if it'd be worth eventually introducing a parallel summed potential)

Simply because from the integrator/barostat where these changes have been implemented see it as a single potential. Definitely could introduce the same logic into the summed potential or a parallel version, I initially started on implementing it but decided to see how this change was viewed before getting too far along. Its also less of a win, since the only SummedPotential commonly constructed is the Nonbonded All Pairs + Pair List

timemachine/cpp/src/streamed_potential_runner.cu
badisa: Will expand, but its because timings make events slower

timemachine/cpp/src/kernels/k_nonbonded.cuh
badisa: Good question, I hadn't tried. Will give it a shot. I imagine it would pick it up

timemachine/cpp/src/streamed_potential_runner.cu
```diff
@@ -0,0 +1,76 @@
...
+        cudaStream_t pot_stream = this->_get_potential_stream(i);
+
+        gpuErrchk(cudaStreamWaitEvent(pot_stream, sync_event_));
+        bps[i]->execute_device(N, d_x, d_box, d_du_dx, d_du_dp, d_u, pot_stream);
+        cudaEvent_t event = this->_get_potential_event(i);
+        // Tell the main stream to synchronize on all of the potential streams
+        gpuErrchk(cudaEventRecord(event, pot_stream));
+        gpuErrchk(cudaStreamWaitEvent(stream, event));
+    }
+};
```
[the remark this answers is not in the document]
badisa: This is not a potential, this is just a way to run a series of bound potentials. 

Maybe StreamedBoundPotentialRunner?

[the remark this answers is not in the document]
mcwitt: Ah, I guess one difference is that this only accepts bound potentials. Otherwise, though, isn't it basically a summed potential (i.e. takes a list of potentials and computes their sum given (coords, box) arguments)?

Maybe this suggestion would make more sense if the interface allowed passing unbound potentials and parameters, similar to the existing bound potential. As it is, is it  basically a `ParallelizedBoundSummedPotential`? (that's a mouthful)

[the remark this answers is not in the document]
badisa: It is similar to a summed potential in the outcome, though this doesn't maintain any parameters. This API acts on any arbitrary set of BoundPotentials.

Having to construct this class around a set of bound potentials would result in creating and destroying many of these objects as the number of potentials passed to the integrator/barostat changes (local vs global md). Or some clever way of tracking which would seem like unnecessary complexity.

The solution that comes to mind is to split this runner up into two APIs:
* Syncing class that provides streams, events with APIs for syncing less verbosely
* ParallelizedBoundPotentialRunner for running a series of bound potentials in separate streams that relies on the syncing class

Then use the syncing class to duplicate the parallelism in the Summed/Fanout potential. How does that sound?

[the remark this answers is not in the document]
mcwitt: > Having to construct this class around a set of bound potentials would result in creating and destroying many of these objects as the number of potentials passed to the integrator/barostat changes

Makes sense; I hadn't considered that the set of potentials passed to the integrator changes dynamically in local MD.

> The solution that comes to mind is to split this runner up into two APIs

This sounds reasonable to me.

> Then use the syncing class to duplicate the parallelism in the Summed/Fanout potential

My only concern is if we want to maintain sequential versions of these (in case there are potentials we want to explicitly avoid running in parallel), this would require added classes and complexity. Maybe better to split this up into a separate PR?

jkausrelay:
>>> Potentials wrapped into SummedPotentials will not be parallelized

>> Could you comment on why this is the case? Is this just because we can't be sure that the components are safe to parallelize? (if that's the case, I wonder if it'd be worth eventually introducing a parallel summed potential)

> Simply because from the integrator/barostat where these changes have been implemented see it as a single potential. Definitely could introduce the same logic into the summed potential or a parallel version, I initially started on implementing it but decided to see how this change was viewed before getting too far along. Its also less of a win, since the only SummedPotential commonly constructed is the Nonbonded All Pairs + Pair List

The SummedPotential is how I'm implementing the split interaction groups, so this will be SummedPotential ( IxnGroup(P-L) + IxnGroup(L-W) + Nonbonded + Pair list ) soon. Not sure how much of a difference this will make, but something to keep in mind.

maxentile:
>I initially started on implementing it but decided to see how this change was viewed before getting too far along. Its also less of a win, since the only SummedPotential commonly constructed is the Nonbonded All Pairs + Pair List

>The SummedPotential is how I'm implementing the split interaction groups

(For any application where we call potentials from Python, I have also been using a single SummedPotential, for efficiency.)

timemachine/cpp/src/streamed_potential_runner.cu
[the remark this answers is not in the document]
badisa: > My only concern is if we want to maintain sequential versions of these (in case there are potentials we want to explicitly avoid running in parallel), this would require added classes and complexity. Maybe better to split this up into a separate PR?

I was going to add a `serial` flag to the syncing class, so parallelism could be disabled. Though I think all potentials should be capable of being run in parallel, but might be missing something.

timemachine/cpp/src/kernels/k_nonbonded.cuh
badisa: Using `compute-sanitizer --tool racecheck` it did not catch the du_dp bug I ran into. Think the profiling changes the runtime enough to not hit the parallel write, as I could get the test to pass 2/3 of the time locally.

Going through the rest of the potentials don't see any non-atomic updates left.

timemachine/cpp/src/streamed_potential_runner.cu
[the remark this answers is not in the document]
badisa: Modified Summed potentials to be parallelized in https://github.com/proteneer/timemachine/pull/1024/commits/1ee3b0f64d0f5801675a078ae4852e88476edd07

timemachine/cpp/src/stream_manager.cu
```diff
@@ -0,0 +1,60 @@
...
+    streams_[key] = new_stream;
+    return new_stream;
+};
+
+cudaEvent_t StreamManager::get_stream_event(int key) {
+    if (events_.count(key) == 1) {
+        return events_[key];
+    }
+    cudaEvent_t new_event;
+    // Create stream with timings disabled as timings slow down events
```
mcwitt, line 37: Should this say "create event"?
```suggestion
    // Create event with timings disabled as timings slow down events
```

timemachine/cpp/src/stream_manager.cu
```diff
@@ -0,0 +1,60 @@
...
+    }
+    cudaStream_t new_stream;
+    // Create stream that doesn't block with the null stream
+    gpuErrchk(cudaStreamCreateWithFlags(&new_stream, cudaStreamNonBlocking));
+
+    streams_[key] = new_stream;
+    return new_stream;
+};
+
+cudaEvent_t StreamManager::get_stream_event(int key) {
```
mcwitt, line 32: Clarification: is the purpose of `get_stream` and `get_stream_event` just to cache streams and events and avoid any overhead of creating new streams and events, or is there a more fundamental reason? Might be worth documenting (naively it seems like we could simplify the code by removing this and making `sync_to` and `sync_from` simple functions)

timemachine/cpp/src/summed_potential.cu
```diff
@@ -48,9 +57,12 @@ void SummedPotential::execute_device(
             d_du_dx,
             d_du_dp == nullptr ? nullptr : d_du_dp + offset,
             d_u,
-            stream);
+            pot_stream);
 
         offset += params_sizes_[i];
+        if (!serial_) {
```
mcwitt, line 63: nit: would it be cleaner to call the argument `parallel`? (seems like `serial` mostly appears as `!serial`)

timemachine/cpp/src/stream_manager.cu
badisa: The overhead of creating events and streams are negligible, destruction of streams is not. cudaEventDestroy is async while cudaStreamDestroy is synchronous. Mostly to simply the management of streams and events to avoid having to manage creation and destruction.

Will add some documentation.

timemachine/cpp/src/summed_potential.cu
badisa: Changed to parallel in [78658e3](https://github.com/proteneer/timemachine/pull/1024/commits/78658e3a9a9c8e7d1826020615c9390c7111cfad)

mcwitt (approved):
LGTM, pending added documentation mentioned in https://github.com/proteneer/timemachine/pull/1024#discussion_r1199613446

timemachine/cpp/src/stream_manager.cu
badisa: Added in [9665423](https://github.com/proteneer/timemachine/pull/1024/commits/96654233a86cf5630ed02634c1f13f3ddb0079ce)

badisa:
Kicked off nightly tests as this PR has been open for a while and is a pretty significant change

https://gitlab.com/relaytx/physics-platform/timemachine/-/pipelines/883282741

badisa:
> LGTM after discussing offline
> 
> 1. This is useful for Global MD
> 2. This may be less useful for parallelized versions of Local MD
> 3. `SummedPotential` variants add a lot of extra complexity, we should deboggle later so that `Potential` subclasses do not create additional streams, and/or modify `SummedPotential` so it no longer subclasses `Potential`

@mcwitt @maxentile @jkausrelay From discussions with @proteneer offline today relating to the SummedPotential detail of this PR

Further details regarding 3.
* Makes implementations that modify/construct potentials in C++ much more difficult as it is recursive, see Local MD code
* Requires a separate form of parallelism that make things more error prone (some nuance to parallelizing with streams) and modifies performance based on how you construct the system

Will likely merge this PR as is so we can play around with this and follow up with PRs that do the following

1. Remove Nonbonded potential (that creates a summed potential with allpairs and exclusions)
2. Remove SummedPotential implementations and replace with a method/class to run a list of bound potentials all at once. Which will allow the same performance benefits of SummedPotentials without the polymorphism.

badisa:
> Kicked off nightly tests as this PR has been open for a while and is a pretty significant change
> 
> https://gitlab.com/relaytx/physics-platform/timemachine/-/pipelines/883282741

Looks to reduce the wall clock time of the nightly tests from 4 hours 20 minutes down to 3 hours 20 minutes.

badisa:
> > Kicked off nightly tests as this PR has been open for a while and is a pretty significant change
> > https://gitlab.com/relaytx/physics-platform/timemachine/-/pipelines/883282741
> 
> Looks to reduce the wall clock time of the nightly tests from 4 hours 20 minutes down to 3 hours 20 minutes.

Nevermind, that reduction was due to https://github.com/proteneer/timemachine/pull/1039

PR #1028, opened 2023-05-18
Miscellaneous improvements to view_atom_mapping_3d

mcwitt:
* use 12-color map generated by ColorBrewer2
* allow passing PRNG seed, used to determine color ordering
* change color assignment method when core > colors to avoid neighbors having the same color in a way that can't be fixed by varying the seed
* allow passing multiple cores; display an additional row for each core
* allow passing core as list of lists
* add type annotations
* add test

Note: this is a breaking change to the API; the `core` argument is removed and replaced with `cores`, which expects a list of cores


proteneer:
nice! might be good to note that this breaks API, since `core` is removed and the new `cores` is now a list of cores. I'm okay with this though, since its just a utility function

mcwitt:
> might be good to note that this breaks API, since core is removed and the new cores is now a list of cores

Added an assertion that argument is a list of cores in https://github.com/proteneer/timemachine/pull/1028/commits/02ea3e065f7a47afe923fa4a213cf7dd35722aba

setup.py
```diff
@@ -122,6 +122,7 @@ def build_extension(self, ext):
             "pytest-cov",
             "hilbertcurve==1.0.5",
             "hypothesis[numpy]==6.54.6",
+            "py3Dmol",
```
jkausrelay, line 125: pin the version?

mcwitt: Good call, done in https://github.com/proteneer/timemachine/pull/1028/commits/07b26d68578938cb17b5e4c9a816fd323195d387

maxentile (approved):
Nice improvements!

jkausrelay (approved):

PR #1030, opened 2023-05-19
Fixes CIFWriter

badisa:
- OpenMM doesn't write an ending # that is needed to be read
- Adds back the close method, also makes the writer a context
- Correctly strip out units, as the PDBxFile writer is units away

### DIFF
--- a/tests/test_cif_writer.py
+++ b/tests/test_cif_writer.py
@@ -76,6 +76,7 @@ def test_cif_writer(n_frames):
         cif = PDBxFile(temp.name)
         assert cif.getNumFrames() == n_frames
         assert cif.getPositions(asNumpy=True).shape == good_coords.shape
+        np.testing.assert_allclose(cif.getPositions(asNumpy=True), good_coords)
 
     _, solvent_coords, _, solvent_top = builders.build_water_system(4.0, ff.water_ff)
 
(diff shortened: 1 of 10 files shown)


tests/test_cif_writer.py
```diff
@@ -35,34 +36,71 @@ def test_write_single_topology_frame():
...
+
+    # test vacuum
+    with NamedTemporaryFile(suffix=".cif") as temp:
+        good_coords = np.concatenate([get_romol_conf(mol_a), get_romol_conf(mol_b)], axis=0)
+        with CIFWriter([mol_a, mol_b], temp.name) as writer:
+            for _ in range(n_frames):
+                writer.write_frame(good_coords * 10)
+        cif = PDBxFile(temp.name)
+        assert cif.getNumFrames() == n_frames
+        assert cif.getPositions(asNumpy=True).shape == good_coords.shape
```
proteneer, line 58: can we sanity check that coordinates are close as well? we probably can't do bitwise checks but an approximate check to .01 Angstroms is probably good

proteneer (approved):
lgtm

tests/test_cif_writer.py
badisa: Added in https://github.com/proteneer/timemachine/pull/1030/commits/68959d9b05f5d2fe704f9e37ee90aa58bc9fd8e8

jkausrelay (approved):

PR #1032, opened 2023-05-22
Improve support for duplicate atom ids in CIFWriter

proteneer:
Previously we hashed a kv using the IDs in atoms, which turns out to be not unique.

Instead, this we now directly hash the atom objects themselves.


badisa:
Is there a way to add a test for this?

badisa (approved):

proteneer:
The assertions in the CIF code now asserts that the generated IDs are unique. There is no way (or is it appropriate) to require that input topologies have identical IDs.

PR #1033, opened 2023-05-22
Make test deadline setting consistent

mcwitt:
We currently set a custom test timeout deadline for `test_pairwise_distances_periodic`, which overrides the default setting of `None` in CI.

We should probably either remove this and run tests locally with `--hypothesis-profile=ci`, or override this specific test with a value of `None` for consistency.

(This is causing failures due to timeout locally when running tests in parallel.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ MKFILE_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 CPP_DIR := $(MKFILE_DIR)timemachine/cpp/
 INSTALL_PREFIX := $(MKFILE_DIR)timemachine/
 # Conditionally set pytest args, to be able to override in CI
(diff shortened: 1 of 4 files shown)


badisa:
If we have to run with a specific flag, we should just add it to the `pytest.ini` file.

mcwitt:
> If we have to run with a specific flag, we should just add it to the pytest.ini file.

Hmm, my only hesitation is that I'm not sure how we could then run with timeouts if desired (i.e. when developing locally). Will try to think of a good solution.

mcwitt:
Disabled deadlines entirely in `pytest.ini` in https://github.com/proteneer/timemachine/pull/1033/commits/bb96eb7b2e926c6db534ce47b2e28b76292442d3

badisa (approved):

PR #1036, opened 2023-05-23
Allow passing empty list of cores

mcwitt:
Fixes regressions in `view_atom_mapping_3d` introduced by #1028: 

- omitting the `cores` argument (or passing an empty list) results in `AssertionError`
- passing list of cores with different sizes results in runtime error
- passing list of ndarrays results in runtime error

### DIFF
--- a/timemachine/fe/utils.py
+++ b/timemachine/fe/utils.py
@@ -287,8 +287,8 @@ def view_atom_mapping_3d(
     except ImportError as e:
         raise RuntimeError("requires py3Dmol to be installed") from e
 
-    if len(cores) > 0:
-        assert np.asarray(cores[0]).ndim == 2, "expect a list of cores"
+    for core in cores:
+        assert np.asarray(core).ndim == 2, "expect a list of cores"
 
     make_style = lambda props: {"stick": props}
     atom_style = lambda color: make_style({"color": color})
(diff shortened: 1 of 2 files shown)


timemachine/fe/utils.py
```diff
@@ -287,9 +287,8 @@ def view_atom_mapping_3d(
     except ImportError as e:
         raise RuntimeError("requires py3Dmol to be installed") from e
 
-    cores_ = np.asarray(cores)
-    assert cores_.ndim == 3, "expect a list of cores"
-    cores = cores_.tolist()
+    if len(cores) > 0:
+        assert np.asarray(cores[0]).ndim == 2, "expect a list of cores"
```
maxentile, line 291: Nit: could be strengthened to `for c in cores: assert np.assarray(c).ndim == 2`?

Usage question: is there an upper limit on `len(cores)`?

maxentile (approved):
Good catch!

timemachine/fe/utils.py
mcwitt: > Nit: could be strengthened

Good call, your suggestion is an improvement. Changed in https://github.com/proteneer/timemachine/pull/1036/commits/a5623cf379b8c9413803fff369c6776fbe418891

> is there an upper limit on len(cores)?

Only that imposed by `py3Dmol.view`, if any. I'm inclined to let py3Dmol handle raising errors for exceeding the maximum grid size (if there is one).

maxentile: Thanks!

PR #1034, opened 2023-05-22
Split solvent-ligand / solvent-protein ixns pt2

jkausrelay:
- Split interactions into ligand-water and ligand-other (protein, ions, ...) terms
TODO:
- [x] NBAllPairs - skip all ligand terms (need to support atom_idxs_i argument in potentials)
- [x] Interface with Local MD
- [x] Determine performance impact (see below)
- [x] Improve performance

```
- (all from test_benchmark.py compared with the latest master w/the parallel potential implementation)

- RBFE complex leg - 30% slower
- (master) hif2a-rbfe: N=8840 speed: 693.75ns/day dt: 1.5fs (ran 200 steps in 0.04s)
- (PR) hif2a-rbfe: N=8840 speed: 494.50ns/day dt: 1.5fs (ran 200 steps in 0.05s)

- RBFE complex local MD - 40% slower
- (master) hif2a-rbfe-local: N=8840 speed: 1019.04ns/day dt: 1.5fs (ran 200 steps in 0.03s)
- (PR) hif2a-rbfe-local: N=8840 speed: 644.70ns/day dt: 1.5fs (ran 200 steps in 0.04s)

- RBFE solvent leg - 20% slower
- (master) solvent-rbfe: N=6317 speed: 845.57ns/day dt: 1.5fs (ran 200 steps in 0.03s)
- (PR) solvent-rbfe: N=6317 speed: 694.83ns/day dt: 1.5fs (ran 200 steps in 0.04s)

- RBFE solvent leg local MD - 25% slower
- (master) solvent-rbfe-local: N=6317 speed: 912.76ns/day dt: 1.5fs (ran 200 steps in 0.03s)
- (PR) solvent-rbfe-local: N=6317 speed: 682.05ns/day dt: 1.5fs (ran 200 steps in 0.04s)
```

PR #1037, opened 2023-05-23
Update potentials.Nonbonded to accept atom_idxs arg

jkausrelay:
- Will be used in #1034 
- Filter to allow Nonbonded all pairs for subset of system without needing exclusions

### DIFF
--- a/tests/nonbonded/test_nonbonded_jit.py
+++ b/tests/nonbonded/test_nonbonded_jit.py
@@ -1,11 +1,13 @@
 import jax
 import jax.numpy as jnp
 import numpy as np
+import pytest
 
 from timemachine.potentials import Nonbonded
 
 
-def test_nonbonded_reference_jittable(rng: np.random.Generator):
+@pytest.mark.parametrize("num_atom_idxs", [None, 15])
+def test_nonbonded_reference_jittable(num_atom_idxs, rng: np.random.Generator):
 
     N = 30
 
@@ -15,6 +17,7 @@ def test_nonbonded_reference_jittable(rng: np.random.Generator):
         scale_factors=jnp.zeros((0, 2)),
         beta=1.0,
         cutoff=0.1,
+        atom_idxs=np.arange(num_atom_idxs) if num_atom_idxs is not None else None,
     )
 
     U_ref_jit = jax.jit(U_ref.__call__)
@@ -23,4 +26,9 @@ def test_nonbonded_reference_jittable(rng: np.random.Generator):
         conf=rng.uniform(0, 1, size=(N, 3)),
         params=rng.uniform(0, 1, size=(N, 3)),
         box=10.0 * np.eye(3),
+    )
+    _ = jax.value_and_grad(U_ref_jit, argnums=(0, 1))(
+        rng.uniform(0, 1, size=(N, 3)),
+        rng.uniform(0, 1, size=(N, 3)),
+        10.0 * np.eye(3),
     )
--- a/tests/nonbonded/test_nonbonded_jit.py
+++ b/tests/nonbonded/test_nonbonded_jit.py
@@ -12,7 +12,7 @@ def test_nonbonded_reference_jittable(rng: np.random.Generator):
     U_ref = Nonbonded(
         N,
         exclusion_idxs=jnp.zeros((0,)),
-        scale_factors=jnp.zeros((0, 1)),
+        scale_factors=jnp.zeros((0, 2)),
         beta=1.0,
         cutoff=0.1,
     )
--- a/tests/nonbonded/test_nonbonded_jit.py
+++ b/tests/nonbonded/test_nonbonded_jit.py
@@ -23,12 +23,12 @@ def test_nonbonded_reference_jittable(num_atom_idxs, rng: np.random.Generator):
     U_ref_jit = jax.jit(U_ref.__call__)
 
     _ = U_ref_jit(
-        conf=rng.uniform(0, 1, size=(N, 3)),
-        params=rng.uniform(0, 1, size=(N, 3)),
+        conf=jnp.array(rng.uniform(0, 1, size=(N, 3))),
+        params=jnp.array(rng.uniform(0, 1, size=(N, 3))),
         box=10.0 * np.eye(3),
     )
     _ = jax.value_and_grad(U_ref_jit, argnums=(0, 1))(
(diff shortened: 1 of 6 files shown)


tests/nonbonded/test_nonbonded_jit.py
```diff
@@ -12,7 +12,7 @@ def test_nonbonded_reference_jittable(rng: np.random.Generator):
     U_ref = Nonbonded(
         N,
         exclusion_idxs=jnp.zeros((0,)),
-        scale_factors=jnp.zeros((0, 1)),
+        scale_factors=jnp.zeros((0, 2)),
```
badisa, line 15: Why did this change?

jkausrelay: Scale factors should have a shape of (N, 2), otherwise the test fails at https://github.com/proteneer/timemachine/blob/726028fb0446f6f90c3bae93528ead31e860f74a/timemachine/potentials/nonbonded.py#L127

badisa (approved):

timemachine/potentials/potentials.py
```diff
@@ -97,17 +97,21 @@ class Nonbonded(Potential):
...
     disable_hilbert_sort: bool = False
     nblist_padding: float = 0.1
 
     def __call__(self, conf: Conf, params: Params, box: Optional[Box]) -> float | Array:
+        s = self.atom_idxs if self.atom_idxs is not None else slice(None)
+        num_atoms = len(self.atom_idxs) if self.atom_idxs is not None else self.num_atoms
+        exclusion_idxs, scale_factors = self._filter_exclusions(update_idxs=True)
         charge_rescale_mask, lj_rescale_mask = nonbonded.convert_exclusions_to_rescale_masks(
-            self.exclusion_idxs, self.scale_factors, self.num_atoms
+            exclusion_idxs, scale_factors, num_atoms
```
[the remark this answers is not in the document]
jkausrelay: I've refactored this code to move most of the implementation to nonbonded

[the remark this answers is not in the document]
jkausrelay: No, this doesn't work. The problem is nonbonded.nonbonded needs to jit'able and neither filter_exclusions/convert_exclusions_to_rescale_masks are.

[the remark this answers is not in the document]
mcwitt: Hmm, but ideally we'd like `Nonbonded.__call__` to be jittable as well, right? Is it jittable using the `static_argnums` argument to `jit` to indicate that `atom_idxs` is fixed?

timemachine/potentials/nonbonded.py
```diff
@@ -131,6 +132,51 @@ def convert_exclusions_to_rescale_masks(exclusion_idxs, scales, N):
...
+    """
+    atom_idxs_set = set(atom_idxs)
+    map_to_filtered = {j: i for i, j in enumerate(atom_idxs)}
+    filtered_exclusion_idxs_ = []
+    filtered_scale_factors_ = []
+    for (i, j), sf in zip(exclusion_idxs, scale_factors):
+        if i not in atom_idxs_set or j not in atom_idxs_set:
+            continue
+        if update_idxs:
+            i, j = map_to_filtered[i], map_to_filtered[j]
```
badisa, line 169: Can you explain this line a bit more? Since these exclusions go into a NonbondedPairList, shouldn't the indices stay the same, and only the filtering matter?

timemachine/potentials/nonbonded.py
```diff
@@ -131,6 +132,51 @@ def convert_exclusions_to_rescale_masks(exclusion_idxs, scales, N):
     return charge_rescale_mask, lj_rescale_mask
 
 
+def filter_exclusions(
+    atom_idxs: NDArray[np.int32],
+    exclusion_idxs: NDArray[np.int32],
+    scale_factors: NDArray[np.float64],
+    update_idxs=False,
```
badisa, line 139:
```suggestion
    update_idxs: bool = False,
```

timemachine/potentials/potentials.py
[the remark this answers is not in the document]
jkausrelay: Nonbonded.__call__ is jit'able w/atom_idxs (as long as the atom_idxs are np.arrays and not jnp arrays).

I'm not sure why the test is failing if nonbonded is called directly, I'm still investigating this.

timemachine/potentials/nonbonded.py
badisa: Nevermind, had misread the code

jkausrelay: That is correct for the GPU implementation.

In the reference jax implementation, the conformations/parameters are filtered by remapping (removing) the idx not in atom_idxs. I.e. `conf = jnp.array(conf)[atom_idxs, :]`

timemachine/potentials/potentials.py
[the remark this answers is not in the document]
jkausrelay: Ok, I figured out the issue. `Nonbonded.\_\_call\_\_` is jittable. `nonbonded.nonbonded` can be if it is wrapped in a function (which is why `Nonbonded.\_\_call\_\_` works).  We can expose the conf, params, box args but not exclusion_idxs/scale_factors. It's the difference of:

```
import jax
import jax.numpy as jnp

import numpy as np

def test_a(a, b, c):
    zero_arr = np.ones((5, 5))
    for (i, j), val in zip(b, c):
        zero_arr[i][j] = val
    return zero_arr

a = jnp.array([1.0])
b = np.array([(1,2), (2,3)])
c = np.array([0.1, 0.2])

# ok
# b, c are implicit so jax only traces 'a'
print("OK:")
def test_a_wrap(a):
    return test_a(a, b, c)

jit_a = jax.jit(test_a_wrap)
jit_a(a)

# error
# b, c are explicit and can't be traced
print("ERROR:")
jit_a = jax.jit(test_a)
jit_a(a, b, c)
```

So this is mostly an artifact of how the test is running, where it calls `nonbonded.nonbonded` directly. Also, I can't just use static_argnums b/c we are passing in np.arrays and these are not hashable.

I'm going to modify the test to wrap the calls. In practice, we should be using `Nonbonded`.

jkausrelay:
@mcwitt Any additional comments?

mcwitt (approved):
LGTM

PR #1038, opened 2023-05-25
Task/support local md with nonbonded all pairs subset

badisa:


timemachine/cpp/src/local_md_potentials.cu
```diff
@@ -190,18 +214,20 @@ void LocalMDPotentials::_setup_free_idxs_given_reference_idx(
...
-    const int num_col_idxs = N_ - num_row_idxs;
+    const int num_col_idxs = N_ - num_row_idxs - indices_to_remove;
 
     if (num_row_idxs == 0) {
         throw std::runtime_error("LocalMDPotentials setup has no free particles selected");
     }
 
-    // The reference particle will always be in the column idxs
-    if (num_row_idxs == N_ - 1) {
+    // The reference particle will always be in the column idxs if the referenc is frozen
```
jkausrelay, line 223: typo: reference

timemachine/cpp/src/local_md_potentials.hpp
```diff
@@ -68,6 +69,9 @@ class LocalMDPotentials {
     DeviceBuffer<float> d_probability_buffer_;
 
     DeviceBuffer<unsigned int> d_free_idxs_;
+    DeviceBuffer<unsigned int>
+        d_all_pairs_idxs_;                  // Where intersection of free/frozen and initial atom indices are stored
```
jkausrelay, line 73: nit: The naming here is unexpected, d_all_pairs_idxs_ is what I'd think the 'initial atom indices, of NonbondedAllPairs, are stored'

tests/test_md.py
```diff
@@ -907,3 +915,100 @@ def build_context(barostat_interval):
...
+    np.testing.assert_array_equal(all_pairs_bound.get_potential().get_atom_idxs(), non_ligand_idxs)
+
+    # Global MD should be identical, verify that these match before moving onto local where potentials are modified
+    np.testing.assert_array_equal(ref_xs, test_xs)
+    np.testing.assert_array_equal(ref_boxes, test_boxes)
+
+    np.testing.assert_array_equal(ref_local_xs, test_local_xs)
+    np.testing.assert_array_equal(ref_local_boxes, test_local_boxes)
+
+    # Verify that the all pairs is set back to the original atom indices set
```
jkausrelay, line 1014: todo?

jkausrelay (approved):
Minor comments, lgtm

tests/test_md.py
badisa: Fixed up in a rebase. Hope that doesn't confuse things, hadn't notice your review.

timemachine/cpp/src/local_md_potentials.hpp
badisa: Changed in https://github.com/proteneer/timemachine/pull/1038/commits/7dbeabcffb9e5b0cc94b1ccebbf69ebaf8eefb1a

timemachine/cpp/src/local_md_potentials.cu
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1038/commits/7dbeabcffb9e5b0cc94b1ccebbf69ebaf8eefb1a

tests/nonbonded/test_nonbonded_all_pairs.py
```diff
@@ -35,6 +35,27 @@ def test_nonbonded_all_pairs_invalid_num_params():
...
+    gpu_pot = potential.to_gpu(np.float32).unbound_impl
+    ref_atom_idxs = gpu_pot.get_atom_idxs()
+    assert len(ref_atom_idxs) == num_atoms
+    assert num_atoms == gpu_pot.get_num_atom_idxs()
+    assert len(set(ref_atom_idxs)) == len(ref_atom_idxs)
+    identity_idxs = np.arange(0, num_atoms, dtype=np.int32)
+    updated_idxs = rng.choice(identity_idxs, size=100, replace=False)
+    gpu_pot.set_atom_idxs(updated_idxs)
+    assert len(updated_idxs) == gpu_pot.get_num_atom_idxs()
+    np.testing.assert_array_equal(gpu_pot.get_atom_idxs(), updated_idxs)
```
mcwitt, line 56: Should we test edge cases here? (`atom_idxs` empty, singleton, having duplicates, etc.?)

tests/test_md.py
```diff
@@ -907,3 +915,99 @@ def build_context(barostat_interval):
...
+    identity_idxs = np.arange(0, len(coords), dtype=np.int32)
+    ligand_idxs = np.arange(len(coords) - mol.GetNumAtoms(), len(coords), dtype=np.int32)
+    non_ligand_idxs = np.delete(identity_idxs, ligand_idxs)
+
+    bound_pots = [bp.bind(params) for bp, params in zip(unbound_potentials, sys_params)]
+
+    summed_bound = next(fn for fn in bound_pots if isinstance(fn.potential, SummedPotential))
+    nb_pot_idx = next(i for i, pot in enumerate(summed_bound.potential.potentials) if isinstance(pot, Nonbonded))
+    nb_pot = summed_bound.potential.potentials[nb_pot_idx]
+    nb_bound = nb_pot.bind(summed_bound.potential.params_init[nb_pot_idx])
```
mcwitt, line 945: nit: maybe slightly cleaner to use `zip` here:

```suggestion
    nb_bound = next(
        pot.bind(params)
        for pot, params in zip(summed_bound.potential.potentials, summed_bound.potential.params_init)
        if isinstance(pot, Nonbonded)
    )
```

tests/nonbonded/test_nonbonded_all_pairs.py
badisa: All of those edge cases, if I understand what you mean, are verified by the atom idxs that we have for the NonbondedAllPairs. You aren't able to configure the potential with those indices and if local MD somehow did provide bad indices, the coordinates wouldn't agree. 

If you mean to test that local MD works under all of these conditions, I think that becomes hairy because the potentials need to be setup with care. At least I convinced myself that you couldn't just randomly select particles as part of the all pairs (notably you can't have the ixn group have any ligand atoms else double counts forces with precomputed). 

This care required to setup the system is definitely a weakness of how local MD functions and I would like to determine a solution long term, but in the short term this unblocks https://github.com/proteneer/timemachine/pull/1034

tests/test_md.py
badisa: Need nb_pot to construct the other potential and this pushes more logic than I like in a single line.

tests/nonbonded/test_nonbonded_all_pairs.py
mcwitt: I meant specifically for the `set_atom_idxs` method, but looking closer, it doesn't seem like there is any runtime validation to test here (maybe this is deliberate, since this is intended to run in a hot loop?)

badisa: Good catch, will expand the `set_atom_idxs` tests to ensure there is validation.

tests/test_md.py
mcwitt: Ah, fair enough (I updated the suggestion to expose both `nb_pot` and `nb_params`, but agreed that it's a matter of preference)

tests/test_md.py
```diff
@@ -907,3 +915,99 @@ def build_context(barostat_interval):
...
+    ctxt.setup_local_md(temperature, True)
+    ref_xs, ref_boxes = ctxt.multiple_steps(steps)
+    ref_local_xs, ref_local_boxes = ctxt.multiple_steps_local(steps, ligand_idxs, burn_in=0)
+
+    ctxt = custom_ops.Context(coords, v0, box, intg.impl(), comp_bps)
+    ctxt.setup_local_md(temperature, True)
+    test_xs, test_boxes = ctxt.multiple_steps(steps)
+    test_local_xs, test_local_boxes = ctxt.multiple_steps_local(steps, ligand_idxs, burn_in=0)
+
+    # Verify that the all pairs is set back to the original atom indices set
```
mcwitt, line 1005: Not sure I understand what "set back" means here. It looks like `all_pairs` is initialized with `non_ligand_idxs` on L958. Are the indices expected to change?

mcwitt: Oh, is this because `multiple_steps_local` modifies the atom indices?

timemachine/cpp/src/kernels/k_indices.cu
```diff
@@ -1,7 +1,7 @@
 #include "k_indices.cuh"
 
-// Takes a source and destination array. Assumes K <= N with values in the src are less than or equal
-// to K. The value of the src is used as the indice and the value in the destination array. Allows combining
+// Takes a source and destination array.
+// The value of the src is used as the indice and the value in the destination array. Allows combining
```
mcwitt, line 4: nit:
```suggestion
// The value of the src is used as the index and the value in the destination array. Allows combining
```

tests/test_md.py
mcwitt: Never mind -- was confused at first but I think it makes sense.

mcwitt (approved):

PR #1026, opened 2023-05-17
Adds Local MD to BFE methods

badisa:
- Fixes two tests that reliably failed locally
- Maintains the BFE defaults of 400 steps per frame, 200k equil, 2000 frames
- Modifies BFE methods to take MDParams to allow more precise control as well as to enable local MD moves.

PR #1039, opened 2023-05-26
Use n_windows in run_edge_and_save_results

maxentile:
`n_windows` was unused, may have been defaulting to 30


mcwitt:
Good catch; looks like `n_windows` inadvertently got dropped from the `run_complex` and `run_solvent` calls in https://github.com/proteneer/timemachine/pull/1026/files#diff-49ef8469d9fb5c9d39e05a0cef32710703096c53f657e55eff49d5c3ddd47b13L766-L771

mcwitt:
nit: could we add a test for the expected number of windows, e.g. [here](https://github.com/proteneer/timemachine/blob/d1af042c671a12338473f28efd8a1fb2bbf82559/tests/test_examples.py#L224)?

maxentile:
>nit: could we add a test for the expected number of windows, e.g. [here](https://github.com/proteneer/timemachine/blob/d1af042c671a12338473f28efd8a1fb2bbf82559/tests/test_examples.py#L224)?

Good call -- added assertion in https://github.com/proteneer/timemachine/pull/1039/commits/a59c92118ef76ab74e5983cb92b93d98d97daf38

mcwitt (approved):

PR #1041, opened 2023-05-30
Adds method to generate fwd and reverse ddg plot

badisa:
* Based on https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4420631/ Figure 5

Here are some plots of some Hif2a edges.
## Example of good convergence
image

## Example of poor convergence
image

Not sure what the right way to test plotting code is, beyond verifying that it doesn't fail and that the plots look reasonable.


timemachine/fe/bar.py
```diff
@@ -217,3 +263,52 @@ def pair_overlap_from_ukln(u_kln: np.ndarray) -> float:
...
+
+    Returns
+    -------
+        fwd_df: [chunks] np.ndarray
+            numpy array of dF for each chunk
+        fwd_df_err: [chunks] np.ndarray
+            numpy array of dF errors for each chunk
+        rev_df: [chunks] np.ndarray
+            numpy array of dF for each chunk, from reversed ukln
+        ref_df_err: [chunks] np.ndarray
```
mcwitt, line 292:
```suggestion
        rev_df_err: [chunks] np.ndarray
```

maxentile:
Nice!!

>Not sure what the right way to test plotting code is, beyond verifying that it doesn't fail and that the plots look reasonable.

Not sure. 

Ideas that come to mind would involve modifying this to return some non-graphical diagnostic info (such as a timeseries of how much the blue and orange intervals are overlapping), and then asserting:
* for "unconverged" input, the intervals don't overlap until the very last step
* for "converged" input, (1) the intervals start overlapping at an early step, (2) the "overlap over time indicator timeseries" looks like [False, False, False, ... , True, True, True] but not [True, False, True, False, True, False, False], for example

timemachine/fe/plots.py
```diff
@@ -169,10 +171,73 @@ def make_overlap_detail_figure(
...
+
+    plt.figure(figsize=(6, 6))
+    plt.title("∆∆G Convergence Over Time")
+    plt.plot(xs, fwd, label="Forward ∆∆G", marker="o")
+    plt.fill_between(xs, fwd - fwd_err, fwd + fwd_err, alpha=0.25)
+    plt.plot(xs, rev, label="Reverse ∆∆G", marker="o")
+    plt.fill_between(xs, rev - rev_err, rev + rev_err, alpha=0.25)
+    plt.axhline(fwd[-1], linestyle="--")
+    plt.xlabel("Fraction of simulation time")
+    plt.ylabel("∆∆G (kcal/mol)")
```
mcwitt, line 235: Should state in the docstring that expected units are kcal/mol?

maxentile (approved):

mcwitt (approved):
Looks good!

mcwitt:
> Not sure what the right way to test plotting code is, beyond verifying that it doesn't fail and that the plots look reasonable.

I don't know about the plotting code, but should we test `compute_fwd_and_reverse_df_over_time`?

timemachine/fe/plots.py
badisa: Will add mention that they are in kBT

timemachine/fe/plots.py
```diff
@@ -169,10 +171,73 @@ def make_overlap_detail_figure(
...
+    chunks: int
+        Number of chunks to split uklns by.
+
+    Returns
+    -------
+    ddg_convergence_plot_bytes: bytes
+    """
+
+    solvent_fwd, solvent_fwd_err, solvent_rev, solvent_rev_err = compute_fwd_and_reverse_df_over_time(
+        solvent_ukln_by_component_by_lambda.sum(0)  # Combine by component
```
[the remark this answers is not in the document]
badisa: Correct, will fix. Seems confusing that by component by lambda means [lambda, component, ...] is there a convention we are following with this naming?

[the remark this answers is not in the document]
mcwitt: I probably introduced this, and I don't think there's necessarily an established convention. I mentally parse it with parentheses associating left, e.g. "((X by foo) by bar) by baz", so "baz" would be the first dimension. Don't have strong feelings about convention but consistency seems useful.

timemachine/fe/plots.py
```diff
@@ -169,10 +171,73 @@ def make_overlap_detail_figure(
...
+    not be drifting with all samples.
+
+    Refer to figure 5 of https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4420631/ for more details.
+
+    Parameters
+    ----------
+    solvent_ukln_by_component_by_lambda: [n_component, n_lambdas, 2, 2, N] array
+        Solvent ukln broken up by components and lambdas
+    complex_ukln_by_component_by_lambda: [n_component, n_lambdas, 2, 2, N] array
+        Complex ukln broken up by components and lambdas
```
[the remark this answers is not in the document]
badisa: Changed in https://github.com/proteneer/timemachine/pull/1041/commits/0341c3d94bb84556c04b93208ac25674f1b32aae

badisa:
> Nice!!
> 
> > Not sure what the right way to test plotting code is, beyond verifying that it doesn't fail and that the plots look reasonable.
> 
> Not sure.
> 
> Ideas that come to mind would involve modifying this to return some non-graphical diagnostic info (such as a timeseries of how much the blue and orange intervals are overlapping), and then asserting:
> 
>     * for "unconverged" input, the intervals don't overlap until the very last step
> 
>     * for "converged" input, (1) the intervals start overlapping at an early step, (2) the "overlap indicator timeseries" looks like [False, False, False, ... , True, True, True] but not [True, False, True, False, True, False, False], for example

Added validation that the values match at the end in https://github.com/proteneer/timemachine/pull/1041/commits/efad350c24938f202a1375f2d5a7830c41e1c4fb

timemachine/fe/plots.py
```diff
@@ -169,10 +171,73 @@ def make_overlap_detail_figure(
...
     buffer.seek(0)
     overlap_detail_png = buffer.read()
 
     return overlap_detail_png
+
+
+def plot_forward_and_reverse_ddg(
+    solvent_ukln_by_component_by_lambda: NDArray,
+    complex_ukln_by_component_by_lambda: NDArray,
+    temperature: float,
```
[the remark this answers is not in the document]
badisa: Changed in https://github.com/proteneer/timemachine/pull/1041/commits/d220ed4ec7009d88573a8934cf8f7f33b5a6253f

badisa:
> > Not sure what the right way to test plotting code is, beyond verifying that it doesn't fail and that the plots look reasonable.
> 
> I don't know about the plotting code, but should we test `compute_fwd_and_reverse_df_over_time`?

Added test in https://github.com/proteneer/timemachine/pull/1041/commits/cd655721884cfd35bd83d35f263711c84d375426

timemachine/fe/plots.py
badisa: Added in https://github.com/proteneer/timemachine/pull/1041/commits/0341c3d94bb84556c04b93208ac25674f1b32aae that they are unitless

timemachine/fe/bar.py
```diff
@@ -217,3 +263,52 @@ def pair_overlap_from_ukln(u_kln: np.ndarray) -> float:
...
+        rev_df_err: [chunks] np.ndarray
+            numpy array of dF errors for each chunk, from reversed ukln
+    """
+    assert len(ukln_by_lambda.shape) == 4
+    assert ukln_by_lambda.shape[1] == 2
+    forward_predictions_ = []
+    reverse_predictions_ = []
+    total_frames = ukln_by_lambda.shape[-1]
+    assert total_frames >= chunks, "fewer samples than chunks"
+    frame_step_size = total_frames // chunks
```
jkausrelay, line 301: Expose this instead of chunks? It might be more intuitive to think in terms of frame step size vs chunks.

timemachine/fe/plots.py
```diff
@@ -169,10 +171,77 @@ def make_overlap_detail_figure(
...
+    chunks: int
+        Number of chunks to split uklns by.
+
+    Returns
+    -------
+    ddg_convergence_plot_bytes: bytes
+    """
+
+    solvent_fwd, solvent_fwd_err, solvent_rev, solvent_rev_err = compute_fwd_and_reverse_df_over_time(
+        solvent_ukln_by_lambda
```
jkausrelay, line 211: pass in chunks

timemachine/fe/plots.py
```diff
@@ -169,10 +171,77 @@ def make_overlap_detail_figure(
...
+    assert np.allclose(fwd_err[-1], rev_err[-1])
+
+    xs = np.linspace(1.0 / chunks, 1.0, chunks)
+
+    plt.figure(figsize=(6, 6))
+    plt.title("∆∆G Convergence Over Time")
+    plt.plot(xs, fwd, label="Forward ∆∆G", marker="o")
+    plt.fill_between(xs, fwd - fwd_err, fwd + fwd_err, alpha=0.25)
+    plt.plot(xs, rev, label="Reverse ∆∆G", marker="o")
+    plt.fill_between(xs, rev - rev_err, rev + rev_err, alpha=0.25)
```
jkausrelay, line 236: nit: Separate plot specific code to it's own function? It could be reused if we want to do a similar analysis at the leg instead of the edge level.

jkausrelay (approved):

PR #1048, opened 2023-06-13
Simplifies Cmake Cuda Architecture using 'native'

badisa:
* As of CMake 3.24 (https://cmake.org/cmake/help/latest/prop_tgt/CUDA_ARCHITECTURES.html) can build for the native architecture rather than having to handle with a call to nvidia-smi


mcwitt:
Looks good to me!

Could this be further simplified / made more flexible by simply setting [`CMAKE_CUDA_ARCHITECTURES`](https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_ARCHITECTURES.html#variable:CMAKE_CUDA_ARCHITECTURES) to `"native"`? Based on my reading, this could then be overridden using the built-in [`CUDAARCHS`](https://cmake.org/cmake/help/latest/envvar/CUDAARCHS.html#envvar:CUDAARCHS) environment variable. (This would also allow setting, e.g. `CUDAARCHS=all-major` to target multiple architectures.)

badisa:
> Could this be further simplified / made more flexible by simply setting [`CMAKE_CUDA_ARCHITECTURES`](https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_ARCHITECTURES.html#variable:CMAKE_CUDA_ARCHITECTURES) to `"native"`? Based on my reading, this could then be overridden using the built-in [`CUDAARCHS`](https://cmake.org/cmake/help/latest/envvar/CUDAARCHS.html#envvar:CUDAARCHS) environment variable. (This would also allow setting, e.g. `CUDAARCHS=all-major` to target multiple architectures.)

If we do that, you will only be able to set `CMAKE_CUDA_ARCHITECTURES` on a fresh build of timemachine using the `CUDAARCHS` variable, as it is only honored on `cmake ..`. Right now, we can change the Cuda arch on the fly with `make CUDA_ARCH=all-major build`.  The latter version seems preferable as it is faster to develop without having to get pybind/eigen every time.

mcwitt:
> If we do that, you will only be able to set CMAKE_CUDA_ARCHITECTURES on a fresh build of timemachine using the CUDAARCHS variable

I was confused by this and looked into it a bit - seems like the core issue is that CMake caches the `CMAKE_CUDA_ARCHITECTURES` variable (in `CMakeCache.txt`) and from that point on ignores the setting of `CUDAARCHS`. Otherwise I would've expected `make CUDAARCHS=all-major` to work (after adding `export CUDAARCHS` to the Makefile), since the default target calls CMake.

I'm not sure how to get around this caching, so it seems like my suggestion isn't worth it if we want the ability to switch arches without cleaning up the build.

proteneer (approved):
lgtm!

PR #1051, opened 2023-06-14
Add `ring_matches_ring` flag to `atom_mapping.get_cores`

mcwitt:
When `ring_matches_ring_only=True`, do not allow ring atoms to map to non-ring atoms and vice-versa.

Implemented by excluding from the codomain of each atom in A (specified in `priority_idxs`) indices in B where one atom is in a ring and the other isn't.

* Adds `ring_matches_ring_only` argument to `atom_mapping.get_cores`
* Sets `ring_matches_ring_only=True` by default for RBFE
* Adds test
* Updates existing tests to pass a value for `ring_matches_ring_only` matching the value for `complete_rings`


timemachine/fe/atom_mapping.py
```diff
@@ -347,10 +351,18 @@ def chiral_filter(trial_core):
...
+            if idx_b != mcgregor.UNMAPPED
+        )
+
+    filter_fxn = lambda _: True
     if enforce_chiral:
-        filter_fxn = chiral_filter
-    else:
-        filter_fxn = lambda core: True
+        filter_fxn = compose_predicates(filter_fxn, chiral_filter)
+    if ring_matches_ring:
```
badisa, line 364: Rather than doing it during the mcgregor algorithm, would it make more sense to just set up the priorities idxs to exclude two atoms that aren't both in a ring?

https://github.com/proteneer/timemachine/blob/master/timemachine/fe/atom_mapping.py#L312-L336

mcwitt: Ah, yeah, that seems better. Will update.

mcwitt: Simplified to enforce in the codomains in https://github.com/proteneer/timemachine/pull/1051/commits/6d6a0cddea5eaebc5cb29a2b7d937e594c20b194

timemachine/fe/atom_mapping.py
```diff
@@ -347,10 +350,7 @@ def chiral_filter(trial_core):
         passed = not has_chiral_atom_flips(np_core, chiral_set_a, chiral_set_b)
         return passed
 
-    if enforce_chiral:
-        filter_fxn = chiral_filter
-    else:
-        filter_fxn = lambda core: True
+    filter_fxn = chiral_filter if enforce_chiral else lambda _: True
```
badisa, line 353: Would prefer we keep theis as two lines, this is harder for me to read.

examples/relative_free_energy.py
```diff
@@ -118,6 +118,7 @@ def read_from_args():
         connected_core=True,
         max_cores=1e6,
         enforce_core_core=True,
+        ring_matches_ring=True,
```
badisa, line 121: Not for this PR: Would be nice to just have default keyword values rather than having to change this in all these different files.

badisa (approved):

timemachine/fe/atom_mapping.py
mcwitt: Will revert, this diff is an unintentional artifact of an earlier revert.

examples/relative_free_energy.py
mcwitt: Yes, agreed. Was tempted to do this, but agree that it seems like a task for another PR.

timemachine/fe/atom_mapping.py
maxentile: Nice -- restoring `ring_matches_ring_only` option is important, and applying this in predicates matrix / codomain structure is my implementation preference as well https://github.com/proteneer/timemachine/pull/921 (not sure why I didn't finish this one)

examples/relative_free_energy.py
maxentile: (Preference for explicitness was cited in previous similar suggestions for this function https://github.com/proteneer/timemachine/pull/919#discussion_r1037477106 )

timemachine/fe/atom_mapping.py
```diff
@@ -53,6 +53,7 @@ def get_cores(
     connected_core,
     max_cores,
     enforce_core_core,
+    ring_matches_ring,
```
maxentile, line 56: nit: keep docstring in sync

maxentile (approved):
Looks good!

timemachine/fe/atom_mapping.py
mcwitt: Thanks, I forgot about this.

timemachine/fe/atom_mapping.py
mcwitt: Oh wow, I hadn't seen #921. I cherry-picked the docstring and some further improvements from there; the implementation in this PR is nearly identical now.

PR #1053, opened 2023-06-15
Remove ixn count sync

badisa:
* Speeds up MD by ~1% with fewer registers in nonbonded kernel and more threads per block
* Removes sync that reduces parallelism.

# Benchmark
Nvidia A10 Cuda arch 8.6

## Current
```
dhfr-apo: N=23558 speed: 412.06ns/day dt: 1.5fs (ran 100000 steps in 31.46s)
dhfr-apo-barostat-interval-25: N=23558 speed: 378.96ns/day dt: 1.5fs (ran 100000 steps in 34.20s)
dhfr-apo-hmr-barostat-interval-25: N=23558 speed: 632.83ns/day dt: 2.5fs (ran 100000 steps in 34.14s)
building a protein system with 1758 protein atoms and 7047 water atoms
hif2a-apo: N=8805 speed: 703.61ns/day dt: 1.5fs (ran 100000 steps in 18.42s)
hif2a-apo-barostat-interval-25: N=8805 speed: 606.01ns/day dt: 1.5fs (ran 100000 steps in 21.39s)
hif2a-rbfe: N=8840 speed: 729.92ns/day dt: 1.5fs (ran 100000 steps in 17.76s)
hif2a-rbfe-local: N=8840 speed: 1116.33ns/day dt: 1.5fs (ran 100000 steps in 11.61s)
solvent-apo: N=6282 speed: 861.06ns/day dt: 1.5fs (ran 100000 steps in 15.05s)
solvent-apo-barostat-interval-25: N=6282 speed: 776.06ns/day dt: 1.5fs (ran 100000 steps in 16.70s)
solvent-rbfe: N=6317 speed: 904.13ns/day dt: 1.5fs (ran 100000 steps in 14.34s)
solvent-rbfe-local: N=6317 speed: 1078.32ns/day dt: 1.5fs (ran 100000 steps in 12.02s)
```

## Without sync
Re-run after https://github.com/proteneer/timemachine/pull/1053/commits/545feb4882d7173427b7b7de71b1fe0b5c7864d4 which reduces the number of registers in the nonbonded kernel
```
dhfr-apo: N=23558 speed: 416.59ns/day dt: 1.5fs (ran 100000 steps in 31.11s)
dhfr-apo-barostat-interval-25: N=23558 speed: 382.46ns/day dt: 1.5fs (ran 100000 steps in 33.89s)
dhfr-apo-hmr-barostat-interval-25: N=23558 speed: 639.22ns/day dt: 2.5fs (ran 100000 steps in 33.79s)
hif2a-apo: N=8805 speed: 705.58ns/day dt: 1.5fs (ran 100000 steps in 18.37s)
hif2a-apo-barostat-interval-25: N=8805 speed: 610.04ns/day dt: 1.5fs (ran 100000 steps in 21.25s)
hif2a-rbfe: N=8840 speed: 731.90ns/day dt: 1.5fs (ran 100000 steps in 17.71s)
hif2a-rbfe-local: N=8840 speed: 1180.43ns/day dt: 1.5fs (ran 100000 steps in 10.98s)
solvent-apo: N=6282 speed: 856.24ns/day dt: 1.5fs (ran 100000 steps in 15.14s)
solvent-apo-barostat-interval-25: N=6282 speed: 780.73ns/day dt: 1.5fs (ran 100000 steps in 16.60s)
solvent-rbfe: N=6317 speed: 889.52ns/day dt: 1.5fs (ran 100000 steps in 14.57s)
solvent-rbfe-local: N=6317 speed: 1078.35ns/day dt: 1.5fs (ran 100000 steps in 12.02s)
```


tests/test_md.py
```diff
@@ -1011,3 +1011,96 @@ def test_local_md_nonbonded_all_pairs_subset():
...
+        _, boxes, _ = ctxt.multiple_steps_U(steps, 0, 0)
+        assert len(boxes) == 1
+    with pytest.raises(RuntimeError, match=err_msg):
+        _, boxes = ctxt.multiple_steps_local(steps, ligand_idxs, burn_in=0)
+        assert len(boxes) == 1
+    with pytest.raises(RuntimeError, match=err_msg):
+        _, boxes = ctxt.multiple_steps_local_selection(steps, reference_idx, selection, burn_in=0)
+        assert len(boxes) == 1
+
+    # Without returning boxes no check will be performed
```
mcwitt, line 1064: I'm not sure I understand why this behavior is desirable? Wouldn't we want to perform this check in all cases?

badisa: Was trying to avoid having to pay the cost of copying if the user doesn't request the boxes. Probably doesn't make much difference, would probably only matter where lots of `multiple_steps` are called with few steps and no coords/boxes are returned.

timemachine/cpp/src/nonbonded_all_pairs.cu
```diff
@@ -335,7 +305,7 @@ void NonbondedAllPairs<RealType>::execute_device(
     kernel_idx |= d_du_dx ? 1 << 1 : 0;
     kernel_idx |= d_u ? 1 << 2 : 0;
 
-    kernel_ptrs_[kernel_idx]<<<ceil_divide(p_ixn_count_[0], tpb / warp_size), tpb, 0, stream>>>(
+    kernel_ptrs_[kernel_idx]<<<NONBONDED_KERNEL_BLOCKS, NONBONDED_KERNEL_THREADS_PER_BLOCK, 0, stream>>>(
```
mcwitt, line 308: nit: should we use a standard case for constants? (`warp_size` vs. `NONBONDED_KERNEL_BLOCKS`)

timemachine/cpp/src/nonbonded_all_pairs.hpp
```diff
@@ -91,6 +89,8 @@ template <typename RealType> class NonbondedAllPairs : public Potential {
 
     double get_cutoff() const { return cutoff_; };
 
+    double get_padding() const { return nblist_padding_; };
```
mcwitt, line 92: nit: prefer `get_nblist_padding`?

timemachine/cpp/src/nonbonded_all_pairs.cu
badisa: That makes sense to me. I prefer all caps, currently we have a mix.

timemachine/cpp/src/nonbonded_common.cu
```diff
@@ -0,0 +1,127 @@
...
+#include "nonbonded_common.cuh"
+
+#include "fanout_summed_potential.hpp"
+#include "summed_potential.hpp"
+
+#include "set_utils.hpp"
+
+namespace timemachine {
+
+void verify_atom_idxs(int N, const std::vector<int> &atom_idxs, const bool allow_empty) {
```
mcwitt, line 15: nit:
```suggestion
void verify_atom_idxs(const int N, const std::vector<int> &atom_idxs, const bool allow_empty) {
```

timemachine/cpp/src/nonbonded_common.cu
```diff
@@ -0,0 +1,127 @@
...
+}
+
+bool is_summed_potential(std::shared_ptr<Potential> pot) {
+    if (std::shared_ptr<FanoutSummedPotential> fanned_potential = std::dynamic_pointer_cast<FanoutSummedPotential>(pot);
+        fanned_potential != nullptr) {
+        return true;
+    } else if (std::shared_ptr<SummedPotential> summed_potential = std::dynamic_pointer_cast<SummedPotential>(pot);
+               summed_potential != nullptr) {
+        return true;
+    }
```
mcwitt, line 42: unrelated to this PR, but could these runtime type checks be simplified using the [`typeid` operator](https://en.cppreference.com/w/cpp/language/typeid)?

mcwitt:
> Removes sync that reduces parallelism

Would you mind adding a brief description of the changes that were necessary to allow removing the sync? I'm having a bit of trouble following the PR (it appears there might be some unrelated reorganization happening too?)

badisa:
> > Removes sync that reduces parallelism
> 
> Would you mind adding a brief description of the changes that were necessary to allow removing the sync? I'm having a bit of trouble following the PR (it appears there might be some unrelated reorganization happening too?)
The following two commits remove the sync, which is simply deletion of some code and setting the threads per block and number of blocks as constants. This requires no reorganization, but it does remove the checks that the nonbonded potentials previously did for the boxes.
a55243a71836919cd9518d560d8d00a67a13cabc
cb4abe26c9df5b73258cf83f6d077b4253e6dc7c

The change to verify the boxes in the context required a reorganization, moving what had been previously local md specific logic (retrieving nonbonded all pairs potentials) to the nonbonded.cuh file. The rename of `nonbonded.hpp` to `nonbonded.cuh` was to be able to access Cuda symbols (cudaStream_t specifically) through its imports. This will also be helpful going forward as we can't currently use CUDA intrinsics in `k_nonbonded_common.cuh` where we might want to optimize some of our kernel specific functions.

timemachine/cpp/src/nonbonded_all_pairs.cu
badisa: Going to open a separate PR that converts all of our non caps constants to caps

timemachine/cpp/src/nonbonded_common.cu
badisa: Added in [9fa2580](https://github.com/proteneer/timemachine/pull/1053/commits/9fa2580d4e51e70d55ead2eb10fccab9af5567be)

tests/test_md.py
mcwitt: OK, I think this makes sense. The only situation I can think of where we might run a large number of steps without storing any frames is during equilibration/burn-in, but maybe this is not an issue in practice. Should we make a note in the docstring for `multiple_steps_*` that checks are only performed when storing frames?

mcwitt (approved):

tests/test_md.py
badisa: Will update docstrings

badisa: Added in https://github.com/proteneer/timemachine/pull/1053/commits/a77d8249e0107b4773cc35d50ef5b25aeb72ff4c

mcwitt: Is it worth also adding an explicit warning like "note: runtime validation of box size is only performed when collecting frames, and thus will not be performed (for example) when `store_x_interval > n_steps`.", maybe to the docstring for `store_x_interval`?

badisa: Expanded in https://github.com/proteneer/timemachine/pull/1053/commits/5b1fe2ae7866b4ea182958baef753fc7b764e81a

timemachine/cpp/src/context.cu
```diff
@@ -37,6 +39,8 @@ Context::Context(
     cub::DeviceReduce::Sum(d_sum_storage_, d_sum_storage_bytes_, d_in_tmp, d_out_tmp, N_);
     gpuErrchk(cudaPeekAtLastError());
     cudaSafeMalloc(&d_sum_storage_, d_sum_storage_bytes_);
+
+    get_nonbonded_all_pair_potentials(bps, nonbonded_pots_);
```
proteneer, line 43: (maybe add a note that this is a no-op if we're running vacuum or if the nonbonded all pairs are missing).

timemachine/cpp/src/context.cu
```diff
@@ -47,6 +51,26 @@ Context::~Context() {
...
+    for (auto boundpot : nonbonded_pots_) {
+        double cutoff = get_nonbonded_all_pair_cutoff_with_padding(boundpot->potential);
+        double db_cutoff = 2 * cutoff;
+        for (int i = 0; i < 3; i++) {
+            if (h_box[i * 3 + i] < db_cutoff) {
+                throw std::runtime_error(
+                    "cutoff with padding is more than half of the box width, neighborlist is no longer reliable");
+            }
+        }
+    }
```
proteneer, line 71: I wonder if we can move some of this logic back into the kernel GPU code for the neighborlist, since CUDA code can now assert:

https://stackoverflow.com/questions/50755717/triggering-a-runtime-error-within-a-cuda-kernel

timemachine/cpp/src/neighborlist.hpp
```diff
@@ -55,15 +55,15 @@ template <typename RealType> class Neighborlist {
     void compute_block_bounds_host(
         const int N, const double *h_coords, const double *h_box, double *h_bb_ctrs, double *h_bb_exts);
 
-    unsigned int *get_ixn_atoms() { return d_ixn_atoms_; }
+    unsigned int *get_ixn_atoms() { return d_ixn_atoms_; };
```
proteneer, line 58: were these extra semi colons required? (not sure they're needed if we're defining the function, as opposed to just declaring them)

timemachine/cpp/src/nonbonded_common.cu
```diff
@@ -0,0 +1,127 @@
+#include <set>
```
proteneer, line 1: why is this a `.cu` file?

timemachine/cpp/src/neighborlist.hpp
badisa: I had run into some compilation issues that seemed like they were due to missing semi-colons and I noticed these trying to find the issue. I left them to be consistent with our other code, but not actually of any importance.

timemachine/cpp/src/nonbonded_common.cuh
```diff
@@ -0,0 +1,35 @@
+#pragma once
```
proteneer, line 1: why is this a `.cuh` file?

timemachine/cpp/src/context.cu
badisa: Based on the documentation (https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#assertion) it looks like that will return a cudaError code which we will then call exit() inside of cudaErrchk.

Seems like an interesting idea, but would require more careful usage to avoid exiting which I find a bit less usable when working with jupyter notebooks.

proteneer (approved):
lgtm - just minor nits.

timemachine/cpp/src/nonbonded_common.cu
proteneer: // includes CUDA Runtime
#include <cuda_runtime.h>

// maybe you need also helpers
#include <helper_cuda.h>
#include <helper_functions.h> // helper utility functions

proteneer: ^ does this work? (Sorry to nit, but trying to keep scope of .cu files somewhat minimal), and I probably didn't do this consistently for previous files either.

badisa: Will give that a shot and get back to you.

timemachine/cpp/src/context.cu
proteneer: Hm - yeah. I guess we should probably clean-up cudaErrchk to throw first to be consistent. I would much prefer the box checks etc. to be done in the GPU code if possible (to keep the scope local to neighborlist code) so it doesn't pollute nblist code.

timemachine/cpp/src/nonbonded_common.cu
badisa: Included `<cuda_runtime.h>` in the neighborlist header in https://github.com/proteneer/timemachine/pull/1053/commits/0e617ed927eb592767a67f75c127ca6b322b6af0

Was that we were ending up at the neighborlist with gcc rather than nvcc. Good call to do this

timemachine/cpp/src/nonbonded_common.cuh
badisa: Done in https://github.com/proteneer/timemachine/pull/1053/commits/0e617ed927eb592767a67f75c127ca6b322b6af0

timemachine/cpp/src/context.cu
badisa: Added check in https://github.com/proteneer/timemachine/pull/1053/commits/968c2a8d954cd3e3a851a6703bcc64ee46543f04

PR #1043, opened 2023-06-06
Upgrade to pymbar 4

mcwitt:

PR #1054, opened 2023-06-15
Handle pymbar convergence failure during bisection

mcwitt:
#### Background
* During the initial iterations of bisection, we compute BAR $\Delta G$ estimates between states that typically have very little overlap
* pymbar 4.0.1 raises `ConvergenceError` exceptions when estimators fail to converge
  * pymbar <4.0.1 had a bug preventing `ConvergenceError`s from being raised, fixed in https://github.com/choderalab/pymbar/commit/66156f888a55a3c15b492e08fb91ae076784587a
* Since upgrading to pymbar 4.0.1 (#1043), we occasionally see failures due to unhandled `ConvergenceError` exceptions early in bisection

#### This PR
* Handles `ConvergenceError` exceptions raised by `pymbar.bar` during bisection
  * when convergence fails for a pair of states, the cost used for the next bisection decision (typically the BAR $\Delta G$ error estimate) is set to infinity
* **Modifies the result schema** to allow representing BAR computations that failed (as `None`), while retaining $u_{kln}$

### DIFF
--- a/timemachine/fe/absolute_hydration.py
+++ b/timemachine/fe/absolute_hydration.py
@@ -2,10 +2,10 @@
 
 import pickle
 from functools import partial
-from typing import List, Sequence, Tuple
+from typing import List, Optional, Sequence, Tuple
 
 import numpy as np
-from numpy.typing import NDArray as Array
+from numpy.typing import NDArray
 from openmm import app
 
 from timemachine import potentials
@@ -13,6 +13,7 @@ from timemachine.constants import BOLTZ, DEFAULT_TEMP
 from timemachine.fe import model_utils
 from timemachine.fe.free_energy import (
     AbsoluteFreeEnergy,
+    BarResult,
     HostConfig,
     InitialState,
     MDParams,
@@ -40,7 +41,7 @@ def generate_endstate_samples(
     num_samples: int,
     solvent_samples: Sequence[CoordsVelBox],
     ligand_samples: Sequence,
-    ligand_log_weights: Array,
+    ligand_log_weights: NDArray,
     num_ligand_atoms: int,
 ) -> List[CoordsVelBox]:
     """solvent + (noninteracting ligand) sample --> solvent + (vacuum ligand) sample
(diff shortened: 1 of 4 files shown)


timemachine/fe/absolute_hydration.py
```diff
@@ -2,17 +2,18 @@
 
 import pickle
 from functools import partial
-from typing import List, Sequence, Tuple
+from typing import List, Optional, Sequence, Tuple
 
 import numpy as np
-from numpy.typing import NDArray as Array
```
mcwitt, line 8: Prefer to remove this alias due of potential confusion with `jax.Array`

timemachine/fe/free_energy.py
```diff
@@ -95,34 +95,40 @@ class PairBarResult:
     """Results of BAR analysis on L-1 adjacent pairs of states given a sequence of L states."""
 
     initial_states: List[InitialState]  # length L
-    bar_results: List[BarResult]  # length L - 1
+
+    # length L - 1 pairs of u_kln_by_component, BAR result
+    # NOTE: BAR result may be None if estimator fails to converge
+    bar_results: List[Tuple[NDArray, Optional[BarResult]]]
```
mcwitt, line 101: Note: this modification to the output schema is to allow for returning partial results (i.e. the $u_{kln}$) if the BAR computation fails to converge. But it will require downstream modifications if we keep it.

badisa: What is preferable about changing the output schema over populating the BarResult objects with Nan values? Seems like these change require modifying a lot of code beyond just `esimate_free_energy_bar`

badisa, line 101: If the BarResult object and the uklns are going to be separated, would prefer that there are two fields on PairBarResult rather than a `List[Tuple[NDArray, Optional[BarResult]]`

mcwitt:
After offline discussion, closing in favor of reverting to pymbar 3 until

1. convergence check can be disabled in pymbar 4, or
2. we determine that self-consistent iteration is appropriate (vs. the default `false-position` method)

PR #1056, opened 2023-06-15
Simplify failing test

badisa:
* The previous version of this test was too restrictive that relied on converting one set of potentials to another. 
* This verifies that running local MD with a different set of potentials doesn't impact the potential and leaves the other tests to verify the correctness of Local MD.

### DIFF
--- a/timemachine/cpp/src/nonbonded_pair_list.cu
+++ b/timemachine/cpp/src/nonbonded_pair_list.cu
@@ -59,13 +59,15 @@ void NonbondedPairList<RealType, Negated>::execute_device(
     unsigned long long *d_u,
     cudaStream_t stream) {
 
-    const int tpb = default_threads_per_block;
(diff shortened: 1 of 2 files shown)


badisa:
Had to force push to remove some commits that are now on master

timemachine/cpp/src/nonbonded_pair_list.cu
```diff
@@ -59,13 +59,15 @@ void NonbondedPairList<RealType, Negated>::execute_device(
     unsigned long long *d_u,
     cudaStream_t stream) {
 
-    const int tpb = default_threads_per_block;
-    const int num_blocks_pairs = ceil_divide(M_, tpb);
+    if (M_ > 0) {
```
badisa, line 62: Without this, there is a failure to launch the kernel in the test. Probably an uncommon case

badisa:
The test that fails is fixed on master

jkausrelay (approved):

PR #1057, opened 2023-06-16
Revert "Upgrade to pymbar 4 (#1043)"

mcwitt:
pymbar 4.0.1 fixed a bug that prevented `ConvergenceErrors` from being raised by `pymbar.bar` (fixed in this [commit](https://github.com/choderalab/pymbar/commit/66156f888a55a3c15b492e08fb91ae076784587a); see #1054 for more context). As a result, there is currently no way to implement our preferred behavior of returning the final iterate (vs. raising an error) if we reach `maximum_iterations` without convergence.

This PR temporarily rolls back to pymbar 3. In the long run, we'll probably want to either

1. upgrade to pymbar 4 if an option is added to skip the convergence check, or
2. use `iterated_solution=False` if we determine that the "self-consistent iteration" solution method is acceptable for our use case

This reverts commit 5bb97e9955a55a8fad5e00b40dd81d8f950b4c7a.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 numpy==1.23.5
 scipy==1.10.1
-pymbar==4.0.1
+pymbar==3.0.5
 pyyaml==5.4.1
 networkx==2.8.8
 matplotlib==3.7.1
--- a/setup.py
+++ b/setup.py
@@ -102,7 +102,7 @@ setup(
         "jaxlib>0.4.1",
         "networkx",
         "numpy",
-        "pymbar>=4",
+        "pymbar>3.0.4,<4",
         "pyyaml",
         "rdkit",
         "scipy",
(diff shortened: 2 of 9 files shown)


badisa:
Do we have a test case we can add as part of this to demonstrate how exactly Pymbar 4 fail? Might not be trivial to create, just might be nice to have a test that we can use to validate upgrading to pymbar 4.

mcwitt:
> Do we have a test case we can add as part of this to demonstrate how exactly Pymbar 4 fail?

Do you mean a test verifying that `pymbar.bar` does not raise `ConvergenceError`s? I think that should be relatively easy, e.g. by setting `maximum_iterations=1` with a small relative tolerance.

mcwitt:
@badisa I'd prefer to add this in a separate PR to keep this as a clean revert (so it can be reverted in the future when we decide to upgrade)

jkausrelay (approved):

maxentile (approved):

PR #1062, opened 2023-06-21
Fix typo in Barostat

badisa:
* The barostat is looking for atom idx 0 rather than idx 0. This is almost always fine, but if we were to not include the first atom idx this would break.


badisa:
The name of this branch broke mirroring, closing the PR and reopenning with a valid name

PR #1063, opened 2023-06-21
Removes support for resizing BoundPotential parameter vectors

badisa:
* Makes debugging more complicated, now it behaves like the neighborlist
* In a future PR we should remove the `shape` of BoundPotentials


mcwitt (approved):
lgtm, nice simplification

proteneer (approved):
lgtm

PR #1064, opened 2023-06-21
Modify infer_node_vals_and_errs_networkx to support nx.MultiDiGraph

jkausrelay:
- Update infer_node_vals_and_errs_networkx to allow a MultiDiGraph to be passed
- Can be used to infer dgs across replicates (where the replicates have the same nodes, but multiple edges with different pred_ddgs)
- Update test

### DIFF
--- a/tests/test_mle.py
+++ b/tests/test_mle.py
@@ -299,6 +299,38 @@ def test_infer_node_vals_and_errs_networkx(nx_graph_with_reference_mle_instance)
         # assert g_res.nodes[n][node_stddev_prop] == pytest.approx(ref_dg_err)
 
 
+def test_infer_node_vals_and_errs_networkx_multi(nx_graph_with_reference_mle_instance):
+    g, seed, ref_dgs, ref_dg_errs = nx_graph_with_reference_mle_instance
+
+    # MultiDiGraph should result in same values
+    mg = nx.MultiDiGraph()
+    for n, d in g.nodes.items():
+        mg.add_node(n, **d)
+    for e, d in g.edges.items():
+        mg.add_edge(*e, **d)
+
+    g_res = infer_node_vals_and_errs_networkx_partial(mg, seed=seed)
+    for n, (ref_dg, ref_dg_err) in enumerate(zip(ref_dgs, ref_dg_errs)):
+        assert g_res.nodes[n][node_val_prop] == pytest.approx(ref_dg, rel=1e-5)
+        assert g_res.nodes[n][node_stddev_prop] == pytest.approx(ref_dg_err, rel=1e-5)
+
+    # multiple edges should result in similar values
+    for e, d in g.edges.items():
+        assert mg.add_edge(*e, **d) == 1
+
+    g_res = infer_node_vals_and_errs_networkx_partial(mg, seed=seed)
+    for n, (ref_dg, ref_dg_err) in enumerate(zip(ref_dgs, ref_dg_errs)):
+        assert g_res.nodes[n][node_val_prop] == pytest.approx(ref_dg, abs=1e-5)
+
+    # adding all 0 edges should shift the dgs toward 0
+    for e, d in g.edges.items():
+        assert mg.add_edge(*e, **{edge_diff_prop: 0.0, edge_stddev_prop: 0.1}) == 2
+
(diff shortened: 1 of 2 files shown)


timemachine/fe/mle.py
```diff
@@ -224,7 +226,7 @@ def infer_node_vals_and_errs_networkx(
     node_stddev_prop: str = "inferred_dg_stddev",
     n_bootstrap: int = 100,
     seed: int = 0,
-) -> nx.DiGraph:
+) -> Union[nx.DiGraph, nx.MultiDiGraph]:
```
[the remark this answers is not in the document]
badisa: I would push back against this. I find these TypeVars to be less readable than a raw Union, and I would prioritize readability over precision of typing that I don't see the clear value of.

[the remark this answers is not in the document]
jkausrelay: Added `NxDiGraph = Union[nx.DiGraph, nx.MultiDiGraph]` as a type. There is a bit of nuance lost with this, but I think that it's more readable.

tests/test_mle.py
```diff
@@ -299,6 +299,38 @@ def test_infer_node_vals_and_errs_networkx(nx_graph_with_reference_mle_instance)
         # assert g_res.nodes[n][node_stddev_prop] == pytest.approx(ref_dg_err)
 
 
+def test_infer_node_vals_and_errs_networkx_multi(nx_graph_with_reference_mle_instance):
```
maxentile, line 302: Not necessary, but another testing strategy could be to summarize all parallel edges as a single edge. (Since product of Gaussians is a Gaussian https://github.com/proteneer/timemachine/pull/399#issuecomment-824893476 )

timemachine/fe/mle.py
```diff
@@ -279,7 +283,7 @@ def infer_node_vals_and_errs_networkx(
     edge_idxs = np.array(sg_relabeled.edges)
 
     dgs, dg_errs = infer_node_vals_and_errs(
-        edge_idxs,
+        np.array([e[:2] for e in edge_idxs]),
```
maxentile, line 286: I found `e[:2]` surprising. Is `len(e) != 2`?

timemachine/fe/mle.py
```diff
@@ -224,7 +228,7 @@ def infer_node_vals_and_errs_networkx(
     node_stddev_prop: str = "inferred_dg_stddev",
     n_bootstrap: int = 100,
     seed: int = 0,
-) -> nx.DiGraph:
+) -> NxDiGraph:
     """Version of :py:func:`timemachine.fe.mle.infer_node_vals_and_errs` that accepts a directed networkx graph.
```
maxentile, line 233: nit: docstring falling further out of sync

maxentile (approved):
Looks good!

timemachine/fe/mle.py
jkausrelay: Will add a comment. With MultiDiGraph it's (from_node, to_node, edge_key) where edge_key = 0, 1, ... for each duplicate edge.

timemachine/fe/mle.py
jkausrelay: Updated

timemachine/fe/mle.py
maxentile: Ahh, then might rename this from `edge_idxs` to `edges` (since depending on input type it may or may not just contain idxs)

jkausrelay: Updated in [65e7c57](https://github.com/proteneer/timemachine/pull/1064/commits/65e7c5797f2794ddae05eb632ed29678221ffc49)

timemachine/fe/mle.py
```diff
@@ -276,12 +281,12 @@ def infer_node_vals_and_errs_networkx(
             ref_node_vals.append(d[ref_node_val_prop])
             ref_node_stddevs.append(d.get(ref_node_stddev_prop, 0.0))
 
-    edge_idxs = np.array(sg_relabeled.edges)
-
+    edges = np.array(sg_relabeled.edges)
+    edge_idxs = np.array([e[:2] for e in edges])  # remove edge key in MultiDiGraph if present
```
mcwitt, line 285: nit: avoid array -> list -> array
```suggestion
    edge_idxs = edges[:, :2]  # remove edge key in MultiDiGraph if present
```

PR #1066, opened 2023-06-21
Draft: Split LJ ixn

jkausrelay:
- Similar to qs, split the LJ parameters
- Needs to be rebased once #1034 is merged


timemachine/potentials/nonbonded.py
```diff
@@ -539,12 +539,12 @@ def coulomb_interaction_group_energy(q_ligand: Array, q_prefactors: Array) -> fl
...
+    q_ligand: [N_frames, N_lig] array
+    q_prefactors: [N_frames, N_lig] array
 
     Returns
     -------
-    energy: float
+    energy: [N_frames] array
     """
 
     return jnp.dot(q_prefactors, q_ligand)
```
maxentile, line 550: docstring was previously correct -- this function applies to a single frame / snapshot.

It can be mapped over multiple frames / snapshots (e.g. using a for loop or vmap or `process_traj_in_chunks`), but for clarity, it is implemented for a single snapshot.

timemachine/potentials/nonbonded.py
```diff
@@ -740,13 +740,13 @@ def lj_interaction_group_energy(sig_ligand, eps_ligand, lj_prefactors):
...
 
     Returns
     -------
-    energy: float
+    energy: [N_frames] array
     """
 
     projection = vmap(basis_expand_lj_atom)(sig_ligand, eps_ligand)
-    return jnp.sum(projection * lj_prefactors)
+    return jnp.sum(jnp.sum(projection * lj_prefactors, axis=2), axis=1)
```
maxentile, line 752: Preference: leave this function as defined for a single snapshot, and add a function to apply it to a sequence of snapshots.

(Or update "Assuming lj_prefactors = lj_prefactors_on_snapshot" from docstring to match revised interface.)

timemachine/ff/__init__.py
```diff
@@ -69,6 +73,8 @@ class Forcefield:
         ]
     ]
     lj_handle: Optional[nonbonded.LennardJonesHandler]
+    lj_handle_solv: Optional[nonbonded.LennardJonesSolventHandler]
+    lj_handle_intra: Optional[nonbonded.LennardJonesIntraHandler]
```
maxentile, line 77: nit: is there a canonical order for `lj_handle, lj_handle_solv, lj_handle_intra`?

`parameterize_nonbonded` uses a different order `lj_handle, lj_handle_intra, lj_handle_solv`, which seems error-prone.

tests/test_jax_nonbonded.py
```diff
@@ -435,7 +435,7 @@ def f(x, box):
 
     @jit
     def u_batch_test(sig_ligand, eps_ligand, q_ligand):
-        vdw = vmap(lj_interaction_group_energy, (None, None, 0))(sig_ligand, eps_ligand, lj_prefactors)
+        vdw = lj_interaction_group_energy(sig_ligand, eps_ligand, lj_prefactors)
```
badisa, line 438: Why is this no longer vmap?

maxentile (approved):

badisa (approved):

tests/test_jax_nonbonded.py
jkausrelay: reverted

timemachine/potentials/nonbonded.py
jkausrelay: reverted

timemachine/potentials/nonbonded.py
jkausrelay: reverted

I see. I was relying on a feature of jnp.dot which made coulomb_interaction_group_energy happen to work on a trajectory (If a is an N-D array and b is a 1-D array, it is a sum product over the last axis of a and b). I'll update the calling code and keep this the same.

PR #1067, opened 2023-06-22
Return suboptimal atom mapping with warning on timeout

mcwitt:
Currently, `fe.mcgregor.mcs` raises a `MaxVisitsError` exception if either `max_visits` or `max_cores` is exceeded in the backtracking search. This guarantees that successful termination of `mcs` implies that the returned cores are optimal (i.e. largest possible number of atoms).

In some examples, an exhaustive search is impractical, but we would still like to return a "good enough" solution if the specified "timeout" (i.e. `max_visits` or `max_cores`) is reached.

This PR modifies the behavior to warn (using `MaxVisitsWarning`) and return the current (possibly suboptimal) set of cores under the same conditions where we would previously raise `MaxVisitsError`.


badisa (approved):

proteneer:
lgtm!

proteneer (approved):

proteneer:
I'm okay with this PR as-is, but I want to be clear, reaching max visits and reaching max cores have different implications.

Reaching max cores means that we've found a couple of optimal subgraphs with the greatest size K possible, but there are many many cores that satisfy this possibility.

Reaching max visits means that we can't even guarantee that we've found that best size K, it may be some number much less than K still.

mcwitt:
> I'm okay with this PR as-is, but I want to be clear, reaching max visits and reaching max cores have different implications.
> 
> Reaching max cores means that we've found a couple of optimal subgraphs with the greatest size K possible, but there are many many cores that satisfy this possibility. (Warning is okay)
> 
> Reaching max visits means that we can't even guarantee that we've found that best size K for _any_ of the graphs, and it may be a graph of some size much less than K still. (I still think this should throw an exception/failure).

Makes sense, thanks for the clarification.

PR #1070, opened 2023-06-27
Cleans up Benchmarks

badisa:
* Default to 2.5fs since that is the more common standard at this point
* Increase number of steps between collected frames for potentials to ensure NBlist rebuilds
* Remove extranenous barostat benchmark (1.5fs vs 2.5fs isn't informative)
* Reuse more minimization code


proteneer:
lgtm but where's the "Default to 2.5fs since that is the more common standard at this point" happening? I wanted to make sure that it was done only with HMR but couldn't find the change.

badisa:
> lgtm but where's the "Default to 2.5fs since that is the more common standard at this point" happening? I wanted to make sure that it was done only with HMR but couldn't find the change.

That was merely a comment that at this stage we run everything under 2.5fs in practice. The HMR flag sets the step size to 2.5fs here: https://github.com/proteneer/timemachine/blob/master/tests/test_benchmark.py#L170-L172 and I switched hmr to be on by default which changes it from 1.5fs to 2.5fs.

proteneer (approved):
Thanks - lgtm!

PR #1071, opened 2023-06-28
Optimizes initial states using SummedPotential

badisa:
* About 10% faster for some tests (locally, does not seem to be the case in CI)

Running the potentials previously shows to take 4ms vs 1ms using nsys, but doesn't seem to translate to the tests. Unsure whether or not we should merge this, but opening up for consideration.
# Current
Takes 4ms to run the bonded potentials plus the nonbonded potentials
old_version

# PR
Takes ~1ms to run all of the potentials
new


jkausrelay (approved):

proteneer (approved):
lgtm

PR #1072, opened 2023-06-28
Fixes MTM tests to use the correct parameters

badisa:
* The ligand parameters were set up to be zeros

Nightly Pipeline: https://gitlab.com/relaytx/physics-platform/timemachine/-/pipelines/915106937


jkausrelay (approved):

badisa:
Just waiting on the nightly tests to pass before merging, but did already pass the `test_condensed_mtm` test that failed previously.

PR #1074, opened 2023-07-05
Adds Smirnoff 2.1.0 Forcefield

badisa:
* XML taken from https://github.com/openforcefield/openff-forcefields/blob/19d29a8c39836bfdad675a76caa4c78f09ec22fe/openforcefields/offxml/openff-2.1.0.offxml


badisa:
Seemed like something worth having in the mix of FFs we support. Doesn't change any logic, just adds the converted ff to the repo.

maxentile (approved):

jkausrelay (approved):

PR #1075, opened 2023-07-06
Handle inhomogenous parameter dimensions in reference SummedPotential

mcwitt:
The reference summed potential currently fails when the component potentials have parameters with different numbers of dimensions. For example, in

```python
 nested_sp = SummedPotential([harmonic_bond.potential], [harmonic_bond.params])
 nested_sp_params = harmonic_bond.params.flatten()
 sp = SummedPotential([nested_sp, harmonic_bond.potential], [nested_sp_params, harmonic_bond.params])
```

(where `nested_sp_params.ndim == 1` and `harmonic_bond.params.ndim == 2`), `sp` will fail when called, raising

> ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part.

This PR adds a test for the above case, and modifies the reference summed potential implementation to handle inhomogenous numbers of dimensions.

### DIFF
--- a/timemachine/potentials/summed.py
+++ b/timemachine/potentials/summed.py
@@ -34,6 +34,9 @@ def summed_potential(
     """
     assert len(U_fns) == len(shapes)
     sizes = [np.prod(shape) for shape in shapes]
+    assert params.ndim == 1
+    (num_params,) = params.shape
+    assert sum(sizes) == num_params
     # np.split expects indices, must increment sizes to be indices
     split_indices = np.cumsum(sizes)
     paramss = [ps.reshape(shape) for ps, shape in zip(np.split(params, split_indices[:-1]), shapes)]
--- a/timemachine/potentials/summed.py
+++ b/timemachine/potentials/summed.py
@@ -1,4 +1,3 @@
-from functools import reduce
 from typing import Optional, Sequence, Tuple
 
 import jax.numpy as jnp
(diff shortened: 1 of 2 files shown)


timemachine/potentials/summed.py
```diff
@@ -33,7 +33,7 @@ def summed_potential(
         shapes of the parameter array input for each potential term (must be same length as U_fns)
     """
     assert len(U_fns) == len(shapes)
-    sizes = np.prod(shapes, axis=1)
+    sizes = [np.prod(shape) for shape in shapes]
     # np.split expects indices, must increment sizes to be indices
     split_indices = np.cumsum(sizes)
```
maxentile, line 38: Question: would it be redundant to `assert params.shape == (sum(sizes),)`?

(May not add anything, since a stronger check is already implicit in np.split(...).reshape(...) below.)

jkausrelay (approved):

timemachine/potentials/summed.py
mcwitt: I think this is a good suggestion. As you say it fails if the condition is not met, but the issue is immediately obvious. Added assertions on input parameters in https://github.com/proteneer/timemachine/pull/1075/commits/d9dc6d2fec7c17d4605f58cc2d6c6d0529ad878c

PR #1077, opened 2023-07-07
Moves test minimizer test to nightly

badisa:
When working on #1076 noticed that the slow minimization test we have has no assertions, and seems to mostly be an informative test. Seems like we could move it to nightly and save ~5 minutes of CI test time.

We don't appear to use the barker equilibration anywhere and we already have a lot of code that calls out to `minimize_host_4d` which has the same internal assertion that this test could raise.

### DIFF
--- a/tests/test_minimizer.py
+++ b/tests/test_minimizer.py
@@ -2,6 +2,7 @@ from importlib import resources
 from time import time
 
 import numpy as np
+import pytest
 
 from timemachine.fe.free_energy import HostConfig
 from timemachine.fe.utils import read_sdf
@@ -11,6 +12,7 @@ from timemachine.md import builders, minimizer
 from timemachine.md.minimizer import equilibrate_host_barker, make_host_du_dx_fxn
 
 
+@pytest.mark.nightly(reason="Slow and no assertions")
 def test_minimizer():
     ff = Forcefield.load_default()
 
--- a/tests/test_minimizer.py
+++ b/tests/test_minimizer.py
@@ -12,8 +12,7 @@ from timemachine.md import builders, minimizer
 from timemachine.md.minimizer import equilibrate_host_barker, make_host_du_dx_fxn
 
 
-@pytest.mark.nightly(reason="Slow and no assertions")
-def test_minimizer():
+def test_minimize_host_4d():
     ff = Forcefield.load_default()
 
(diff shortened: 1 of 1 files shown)


maxentile (approved):

maxentile:
Question: Rather than moving the _whole_ test to nightly, would you prefer to revert this test to how it was prior to https://github.com/proteneer/timemachine/pull/978 so that CI continues to run the parts of test_minimizer.py::test_minimizer that exercise minimize_host_4d (without assertions), and just the Barker-specific parts of the test are removed?

badisa:
> Question: Rather than moving the _whole_ test to nightly, would you prefer to revert this test to how it was prior to #978 so that CI continues to run the parts of test_minimizer.py::test_minimizer that exercise minimize_host_4d (without assertions), and just the Barker-specific parts of the test are removed?

That is probably the way to go, would keep the minimization tests and add the barker tests as nightly.

tests/test_minimizer.py
```diff
@@ -26,6 +27,26 @@ def test_minimizer():
     mol_a = all_mols[1]
     mol_b = all_mols[4]
 
+    for mols in [[mol_a], [mol_b], [mol_a, mol_b]]:
+        x_host = minimizer.minimize_host_4d(mols, host_config, ff)
+        assert x_host.shape == complex_coords.shape
+
+
+@pytest.mark.nightly(reason="Currently not used in practice")
```
maxentile, line 35: We may eventually write down functional requirements for pre-equilibration in timemachine, which might allow alternate implementations of this step to be considered for use in practice.

(Based on inferred requirements, current test coverage of `minimize_host_4d`, and initial investigation in https://github.com/proteneer/timemachine/pull/978 , the `equilibrate_host_barker` option implemented in timemachine appeared to offer a uniform improvement over the method currently used. But I am likely missing requirements.)

maxentile (approved):
This seems better, to avoid reducing CI coverage of the pre-equilibration method currently used in practice in timemachine.

tests/test_minimizer.py
badisa: Based on what spawned the original `minimize_host_4d` implementation, those two molecules caused failures due to clashes in the protein.

Not sure, there is much of a requirement beyond 'leads to a stable simulation'. From the barker tests it seems like it would be a fine drop-in replacement, would just need to test over a larger set of compounds to make sure there isn't some edge case.

Issue #714, opened 2022-04-11
Some lambda schedules fail validation when num_windows < 5


maxentile:
Migrating from this thread: https://github.com/proteneer/timemachine/pull/712#discussion_r847399620

Problem:

[`validate_lambda_schedule`](https://github.com/proteneer/timemachine/blob/56d9933d208b07bfa0524376086f27a26a468f82/timemachine/fe/lambda_schedule.py#L32-L37) fails assertion that endpoints exactly == 0.0 or 1.0, for some lambda schedules generated with num_windows < 5 .

Extent:

Locally it looks like this is a problem with the methods that use `linspace` and `concatenate`:
* construct_absolute_lambda_schedule_solvent
* construct_absolute_lambda_schedule_complex
* construct_relative_lambda_schedule

but not with the method that uses `interp`
* construct_pre_optimized_absolute_lambda_schedule_solvent

PR #1078, opened 2023-07-07
Move unused lambda schedule to attic

maxentile:
Was optimized for earlier iterations of RBFE and RABFE protocol, and is currently unused.

(closes https://github.com/proteneer/timemachine/issues/714 )


mcwitt:
Curious why this was closed? To me it seems reasonable to attic this, since we aren't using it and it and I would not expect it to give good results with the current version of the code, which already incorporates a nonlinear $w(\lambda)$.

proteneer:
+1 to moving it to the attic (Sorry I just saw that I was added as a reviewer, my bad. I'm 100% okay with moving this).

proteneer (approved):

maxentile:
Sorry, no good reason, just closed since the change is trivial and unrelated to any priorities, and the benefit of merging seemed small relative to the communication overhead of a review.

Re-opened.

mcwitt (approved):

PR #1081, opened 2023-07-11
Add option to allow early termination of bisection if overlap threshold is reached

mcwitt:


timemachine/fe/free_energy.py
```diff
@@ -659,7 +676,14 @@ def compute_intermediate_result(lambdas: Sequence[float]) -> PairBarResult:
             print(f"Sampling new state at λ={lamb_new:.3g}…")
 
         lambdas = lambdas_new
-        results.append(compute_intermediate_result(lambdas))
+        result = compute_intermediate_result(lambdas)
+        results.append(result)
+    else:
+        warn(
```
jkausrelay, line 682: nit: Will this show a warning for min_overlap = 1?

mcwitt: Yes. I was on the fence about making `min_overlap: Optional[float]`, in which case we could not warn if set to `None`. Maybe that is the preferable approach.

mcwitt: Updated to make `min_overlap` optional, and not warn with the default setting of `None` in https://github.com/proteneer/timemachine/pull/1081/commits/d368e05afec6db35834abfd27a54c4ec694c8ce6

timemachine/fe/free_energy.py
```diff
@@ -591,7 +596,10 @@ def run_sims_with_greedy_bisection(
     temperature: float
         Temperature in K
 
-    verbose: bool
+    min_overlap: float or None, optional
+        If not None, return early when the BAR overlap between all neighboring pairs of states exceeds this value
```
maxentile, line 600: Clarification: Is the threshold defined in terms of BAR overlap or [2x](https://github.com/proteneer/timemachine/blob/deccde794e82079ca55ced20baa85f1a42f3b725/timemachine/fe/bar.py#L265) BAR overlap? (i.e. normalized to [0, 0.5] or [0, 1]?)

maxentile (approved):

timemachine/fe/free_energy.py
mcwitt: This uses the definition here: https://github.com/proteneer/timemachine/blob/deccde794e82079ca55ced20baa85f1a42f3b725/timemachine/fe/bar.py#L265

My understanding is that this is the standard definition of "scalar overlap". We might consider updating this line to

```python
return pymbar.MBAR(u_kn, N_k).computeOverlap()["scalar"]
```

to minimize future confusion?

maxentile: Ahh, sorry, I was confused. Correct as-is!

"Scalar overlap" sounds more self-documenting

PR #1083, opened 2023-07-13
[Draft] Water exchange

proteneer:
This Draft PR implements a very basic prototype of instantaneous water exchange in the NPT ensemble. Currently, the move is a simple, uniform translation move, with very low acceptance probabilities. It also provides three useful files:

1) `bb_centered.sdf` - a centered buckyball system
2) `bb_0_waters.pdb` - a water system with no waters inside the buckyball.
3) `bb_6_waters.pdb` - a water system with 6 waters inside the buckyball.

These are all pre-minimized for convenience.


timemachine/exchange/exchange_mover.py
```diff
@@ -0,0 +1,354 @@
...
+        b_idxs = np.delete(np.arange(n_atoms), a_idxs)
+
+        # compute delta_U of insertion
+        trial_chosen_coords = coords[chosen_water_atoms]
+        trial_translation = np.diag(box) * np.random.rand(3)
+        # tbd - what should we do  with velocities?
+
+        # this has a higher acceptance probability than if we allowed for rotations
+        # (probably because we have a much higher chance of a useless move)
+        moved_coords = randomly_translate(trial_chosen_coords, trial_translation)
```
maxentile, line 166: Good observation.

(Suggests that biasing the distribution of proposed translations away from (0,0,0) might be a useful optimization in the future.)

timemachine/exchange/exchange_mover.py
```diff
@@ -0,0 +1,354 @@
...
+    # TBD: cache the minimized and equilibrated initial structure later on to iterate faster.
+
+    cur_x_t = image_frames(initial_state, [cur_x_t], [cur_box])[0]
+    writer.write_frame(cur_x_t * 10)
+
+    md_steps_per_batch = 1000
+    for idx in range(100000):
+        density = compute_density(nwm, cur_box)
+        cur_x_t = image_frames(initial_state, [cur_x_t], [cur_box])[0]
+        occ = compute_occupancy(cur_x_t, cur_box, initial_state.ligand_idxs, threshold=0.46)
```
maxentile, line 326: (note to self: will just need to adapt this `threshold` if using a different spherical cage (C60 fullerene vs. C90 fullerene vs. ...), or modify signature (e.g. to accept `cage_idxs` and compute `threshold = median(distance(centroid, x[cage_idxs], box))` or similar).)

timemachine/exchange/exchange_mover.py
```diff
@@ -0,0 +1,354 @@
...
+    centered_coords = coords - centroid
+    rot_mat = special_ortho_group.rvs(3)
+    rotated_coords = np.matmul(centered_coords, rot_mat)
+    return rotated_coords + offset
+
+
+def randomly_translate(coords, offset):
+    centroid = np.mean(coords, axis=0, keepdims=True)
+    centered_coords = coords - centroid
+    return centered_coords + offset
```
maxentile, line 71: Translating to a uniform point in the box lends to a robust implementation, but may be a contributing factor if a lower acceptance rate is observed with this approach compared with prior approaches.

Prior approaches for targeting water displacement moves include:
* Decomposing box into small (0.2A) grid cells, and focusing proposals on sterically unoccupied cells (as in 3.1.1. of https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6456390/ , and GPU implementation in https://pubmed.ncbi.nlm.nih.gov/34762421/ )
* Restricting attention to a spherical region containing both the site to be hydrated and some bulk water (e.g. https://pubmed.ncbi.nlm.nih.gov/32968887/, https://pubs.acs.org/doi/10.1021/acs.jctc.2c00823 , and https://github.com/MobleyLab/blues-water-hopping-paper/blob/master/SI/water-hopping-supplementary-documents/scripts/WaterTranslation.py )

maxentile (commented):
Thanks for the good testsystems and implementation. I'll adapt parts of this implementation for gradual or parallel variants of water displacement MC moves, as discussed offline.

timemachine/exchange/exchange_mover.py
proteneer: Great points.

I was under the impression that that grid approach of 1) was mainly to address the performance issue when PME is used. I'm not sure if it actually modifies the proposal probability in the Metropolis-Hastings evaluation. A spherical selection region could probably be implemented as a Gaussian sphere (whose reverse probability can be computed exactly).

timemachine/exchange/exchange_mover.py
proteneer: Yep - though we'd want to quantify also the productivity of the accepted moves, i.e. a change in the occupancy

timemachine/exchange/exchange_mover.py
proteneer: Yes - this was computed using a heuristic. A C90 is an ellipsoid as it turns out, so this was computed by measuring the diameter along the widest dimension of the ellipsoid, and halving it into the radius.

timemachine/exchange/exchange_mover.py
maxentile: >I was under the impression that that grid approach of 1) was mainly to address the performance issue when PME is used. I'm not sure if it actually modifies the proposal probability in the Metropolis-Hastings evaluation.

Hmm, I'm not sure.

(My read of 3.1.1-3.1.2 of https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6456390/ was that the steric grid was different from the grid used for coarse energy evaluations, but I'm not 100% sure, and I didn't catch what the typical sparsities of these two grids were.)

timemachine/exchange/exchange_mover.py
```diff
@@ -170,21 +170,44 @@ def exchange_move(self, coords, box):
...
+                b_idxs=b_idxs,
+            )
             print("MC overflow detected", delta_U_delete, delta_U_insert)
-            assert 0
+
+        # convert to inf if we get a nan
+        if np.isnan(delta_U_total):
+            delta_U_total = np.inf
 
         p_accept = min(1, np.exp(-self.beta * delta_U_total))
```
maxentile, line 199: patched elsewhere, but note that python `min` and `np.minimum` differ slightly in semantics, and we should favor `np.minimum` here for comparison to possibly nan values

```python
>>> min(1, np.nan)
1
>>> min(np.nan, 1)
nan
>>> np.minimum(np.nan, 1)
nan
>>> np.minimum(1, np.nan)
nan
```

proteneer:
Closing in favor of #1135

PR #1084, opened 2023-07-13
Detect Particle Clashes

badisa:
- Detects individual particle clashes, does *not* detect overflows when combining multiple energies (Likely the case that @maxentile reported with a buckyball, though needs more investigation)
- Removes `multiple_steps_U` as it is unused and seemed cleaner to remove than to have the context handle energies if we don't have a use for it.
- Clean up some of the barostat tests, they were blowing up silently (lambda = 1.0 and no minimization has instant clashes)
- Marks some tests as `memcheck` which seem like they should be evaluated for memory leaks

## How the clash detection works
We have `NonbondedAllPairs`, `NonbondedInteractionGroup` and `NonbondedPairList`. The first two are unaware of exclusions and will compute the energy naively. These values tend to overflow (either nan/inf or > MAX_LONG_LONG / FIXED_EXPONENT) and in such as case will atomically increment the `d_u_overflow_count` buffer. The `NonbondedPairList`, when negated, handles exclusions. In this case when we come across an energy that overflows we atomically decrement the `d_u_overflow_count`. This way if the only overflows are caused by exclusions the final `d_u_overflow_count == 0`. If the overflow count is non-zero, we instead set the energy to `+inf` and return it to the user.

## Next steps
- ~Put together a reliable failing test for the case where combining a set of energies, which individually are fine, that when summed overflow~ @maxentile  provided test, added in https://github.com/proteneer/timemachine/commit/2f43d5f4dc1f4a42f86873bda3b9940cd1122312
- ~Switch to accumulating energies deterministically in floating point to avoid having to explicitly watch for overflows~ Accumulate in long long and keep track of overflows


tests/test_energy_overflows.py
```diff
@@ -0,0 +1,126 @@
...
+
+@pytest.mark.parametrize("precision,rtol,atol", [(np.float64, 1e-8, 1e-8), (np.float32, 1e-4, 5e-4)])
+def test_energy_overflow_cancelled_by_exclusions(precision, rtol, atol):
+    """This test validates that if two particles overlap, the Nonbonded potential will return an inf energy"""
+    np.random.seed(2023)
+
+    N = 9  # Multiple of three for waters
+    D = 3
+
+    x0 = GradientTest().get_water_coords(D)[:N]
```
proteneer, line 64: Does this actually return a clashy system?

proteneer:
Two additional sanity tests:

1) Add a test case where two overlapping particles both with LJ eps==0 and, one with charge +1 and charge -1. What should the expected overflow count be? Do we actually see a negative overflow count with this change?
2) Can we test exactly when the energy overflows? Is it actually equal to MAX_LONG_LONG/FIXED_EXPONENT?

timemachine/cpp/src/fixed_point.hpp
```diff
@@ -16,3 +16,11 @@ RealType __host__ __device__ __forceinline__ FIXED_TO_FLOAT_DU_DP(unsigned long
 template <typename RealType> RealType __host__ __device__ __forceinline__ FIXED_TO_FLOAT(unsigned long long v) {
     return static_cast<RealType>(static_cast<long long>(v)) / FIXED_EXPONENT;
 }
+
+template <typename RealType> bool inline __device__ __host__ energy_overflowed(RealType u, int &overflow_count) {
+    if (!isfinite(u) || u > static_cast<RealType>(LLONG_MAX / static_cast<long long>(FIXED_EXPONENT))) {
```
proteneer, line 21: We need to be very precise here. Let's just make this bulletproof for both overflow and :

1) do `u* FIXED_EXPONENT > LLONG_MAX` since that's what actually happens in the casting code.
2) We need to be precise, because we're comparing u * FIXED_EXPONENT (`Real`) against a LLONG_MAX that's casted into a (`Real`), so we should use `nextafter`

```
RealType test_u = u * FIXED_EXPONENT;
if (isnan(u) || is_positive_infinity(u) || is_overflow(u)) {
   overflow_count++;
   return true;
} elseif (is_negative_infinity(u) || is_underflow(u)) {
   overflow_count--;
   return true;
}
return false;

bool is_overflow(x) {
    return x > nextafter(LLONG_MAX, 0);
}

bool is_underflow(x) {
    return x < nextafter(LLONG_MIN, 0);
}

```

tests/test_energy_overflows.py
badisa: It does not, the intention with this test is to verify that the 'clashes' in the AllPairs and PairListNegated cancel out.

maxentile:
>does not detect overflows when combining multiple energies
>Put together a reliable failing test for the case where combining a set of energies, which individually are fine, that when summed overflow

I think I have found a simple example of this case, which may be useful for testing.

(Make a collection of NB particles evenly spaced along a line, and vary the spacing and number of particles.)

```python
import numpy as np
from timemachine.potentials import Nonbonded

num_atoms = 1000

# all particles evenly spaced along x-axis, in a very large box
spacing = 0.75
x = spacing * np.array([np.arange(num_atoms), np.zeros(num_atoms), np.zeros(num_atoms)]).T
box = np.eye(3) * 10000.0
assert((np.diag(box) > spacing * num_atoms).all())

nb_params = np.ones((num_atoms, 4))
nb_params[:,-1] = 0.0

U = Nonbonded(
    num_atoms=num_atoms,
    exclusion_idxs=np.array([(0, num_atoms-1)], dtype=np.int32),
    scale_factors=np.zeros((1, 2)),
    beta=2.0,
    cutoff=1.2,
)
U_gpu = U.to_gpu(np.float32)

print(U)
print("x = ", x)
print("box =", box)
print("nb_params = ", nb_params)
print("U_ref(x) = ", U(x, nb_params, box)) # + 515277630.0
print("U_gpu(x) = ", U_gpu(x, nb_params, box)) # -21594110.84375
```

badisa:
> > does not detect overflows when combining multiple energies
> > Put together a reliable failing test for the case where combining a set of energies, which individually are fine, that when summed overflow
> 
> I think I have found a simple example of this case, which may be useful for testing.
> 
> (Make a collection of NB particles evenly spaced along a line, and vary the spacing and number of particles.)
> 
> ```python
> import numpy as np
> from timemachine.potentials import Nonbonded
> 
> num_atoms = 1000
> 
> # all particles evenly spaced along x-axis, in a very large box
> spacing = 0.75
> x = spacing * np.array([np.arange(num_atoms), np.zeros(num_atoms), np.zeros(num_atoms)]).T
> box = np.eye(3) * 10000.0
> assert((np.diag(box) > spacing * num_atoms).all())
> 
> nb_params = np.ones((num_atoms, 4))
> nb_params[:,-1] = 0.0
> 
> U = Nonbonded(
>     num_atoms=num_atoms,
>     exclusion_idxs=np.array([(0, num_atoms-1)], dtype=np.int32),
>     scale_factors=np.zeros((1, 2)),
>     beta=2.0,
>     cutoff=1.2,
> )
> U_gpu = U.to_gpu(np.float32)
> 
> print(U)
> print("x = ", x)
> print("box =", box)
> print("nb_params = ", nb_params)
> print("U_ref(x) = ", U(x, nb_params, box)) # + 515277630.0
> print("U_gpu(x) = ", U_gpu(x, nb_params, box)) # -21594110.84375
> ```

Added a failing test based off of this, thanks for it! In the future this should return a `nan` energy.

badisa:
> b8de871

Added in https://github.com/proteneer/timemachine/pull/1084/commits/b8de87133fcbf98ab28a1101ffb876eba3b5255a and https://github.com/proteneer/timemachine/pull/1084/commits/3856aed64feb461aee22fade2edbfe55a1e71b0c

timemachine/cpp/src/fixed_point.hpp
badisa: Added in https://github.com/proteneer/timemachine/pull/1084/commits/8fd09fcb86325602dd48175ba077a9349a262a6e and https://github.com/proteneer/timemachine/pull/1084/commits/251de0b36f9742781cb5c1771bc55dceb9dfdbaf

tests/test_energy_overflows.py
```diff
@@ -0,0 +1,282 @@
...
+
+pytestmark = [pytest.mark.memcheck]
+
+
+def verify_energies(a, b):
+    """The energies can be nan, and can't assert nan == nan. This function handles that correctly"""
+    if np.isfinite(a) or np.isfinite(b):
+        assert a == b
+    else:
+        assert not (np.isfinite(a) or np.isfinite(b))
```
maxentile, line 18: Could you document intent of this function a bit more?

I think it defines a special case "-inf == +inf == nan", but otherwise uses `a.__eq__`

tests/test_energy_overflows.py
```diff
@@ -0,0 +1,282 @@
...
+    nonbonded_gpu = nonbonded.to_gpu(precision)
+
+    # TBD: Remove the try/finally to clean up the GPU potential, else calling reset_device
+    # will fail
+    try:
+        # The reference should return a large energy value
+        assert nonbonded(x, nb_params, box) > 0.0
+
+        # The GPU potentials will overflow, resulting in a nan value
+        assert not np.isfinite(nonbonded_gpu(x, nb_params, box))
```
maxentile, line 280: [not immediately actionable]

(If needed in the future, can tweak / weaken this condition slightly: the main problem in the context of https://github.com/proteneer/timemachine/issues/481 is if `sign(nonbonded_gpu(...)) < 0` when `nonbonded_ref(...) >> 0`. Speculatively, it would also be reasonable for `nonbonded_gpu(...)` to return a number close to `nonbonded_ref(...)`, which on this input is large / maybe not representable with current fixed-point accumulator parameters, but does not _have_ to be defined as +inf...)

timemachine/cpp/src/kernels/k_nonbonded_pair_list.cuh
```diff
@@ -154,8 +153,16 @@ void __global__ k_nonbonded_pair_list(
...
         gj_z += FLOAT_TO_FIXED_NONBONDED(-delta_prefactor * delta_z);
 
-        // energy is size extensive so this may not be a good idea
-        energy += FLOAT_TO_FIXED_NONBONDED(u);
+        if (u_buffer) {
+            // Don't add energy if it overflows, but always add the overflow count
+            int overflow_count = 0;
+            if (!energy_overflowed<RealType>(u, overflow_count)) {
+                // energy is size extensive so this may not be a good idea
+                unsigned long long energy = FLOAT_TO_FIXED_NONBONDED(u);
```
maxentile, line 161: Could you elaborate on this comment? 

(Do we use the same `FIXED_EXPONENT` for the accumulator (which represents a size-extensive) as for the individual pair interactions (which are not size-extensive)?)

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -61,6 +62,20 @@ void verify_coords_and_box(
     }
 }
 
+// accumlate_energies handles the combining of energies, summing them up deterministically
+// and returning +inf if there are overflows
+double accumlate_energies(int N, unsigned long long *fixed_u, int overflows) {
```
maxentile, line 67: nit: "+inf" -> "nan", "accumlate" -> "accumulate"

timemachine/cpp/src/kernels/k_nonbonded_pair_list.cuh
```diff
@@ -154,8 +153,16 @@ void __global__ k_nonbonded_pair_list(
...
+                accumulate<Negated>(u_buffer + atom_i_idx, energy);
+            }
+            atomicAdd(u_overflow_count, Negated ? -overflow_count : overflow_count);
```
maxentile, line 164: This logic is in the PR description, but it would be useful to include a line comment (in nonbonded_common? not sure where...) that the overflow_count can either be incremented or decremented when overflows are encountered.

badisa: Currently, it is implemented as specific to Nonbonded, but this will eventually apply to the other potentials. What about adding the documentation to the `accumulate_energies` function in the wrap kernels? It is where we implement the behavior that a non-zero overflow_count results in a nan.

timemachine/cpp/src/kernels/k_nonbonded_pair_list.cuh
badisa: The comment is probably stale/copy-pasta. It doesn't apply to the pair interactions specifically, rather to the`u_buffer` which accumulates N energies into a final value. Inclined to remove as this seems confusing to mention in the pair list when it really applies more broadly to how we accumulate energies. 

We do use the same `FIXED_EXPONENT` in the final accumulation for converting back to double. https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/fixed_point.hpp#L16-L18 https://github.com/proteneer/timemachine/blob/feature/detect-nb-clashes/timemachine/cpp/src/wrap_kernels.cpp#L74

timemachine/cpp/src/kernels/k_nonbonded_pair_list.cuh
maxentile: In `wrap_kernels.cpp::accumulate_energies` sounds appropriate!

timemachine/cpp/src/wrap_kernels.cpp
badisa: Fixed in [b82bbf2](https://github.com/proteneer/timemachine/pull/1084/commits/b82bbf2a3cfebcf3604858c3c4d170d37c036bed)

timemachine/cpp/src/kernels/k_nonbonded_pair_list.cuh
badisa: Expanded the docstring to include mention about incrementing/decrementing overflows in [9c8c748](https://github.com/proteneer/timemachine/pull/1084/commits/9c8c7488b72de7c8d1b9380b9f322512e4017999)

tests/test_energy_overflows.py
badisa: Fixed bug I had introduced in the last commit in [b8e7703](https://github.com/proteneer/timemachine/pull/1084/commits/b8e77037d5478bd1f1cd0f2edf7decd87929d773)

tests/test_energy_overflows.py
badisa: For the foreseeable future, we will stick with the fixed-point accumulator, despite an earlier version of the PR description indicating that we would move to float accumulation. Which means this likely continue to be a `nan`. Does this raise potential issues for your use cases?

maxentile: Thanks -- I think that makes sense!

(Just wasn't sure if the parameters of the fixed-point accumulator might vary in the foreseeable future, e.g. to change the max or min representable energy value. If so, the energy of this input wouldn't necessarily have to be asserted to be not finite -- nan, +inf, or any large positive number would be usable. I think the main behavior that needs to be ruled out on inputs like this is to return a negative number.)

timemachine/md/minimizer.py
```diff
@@ -482,7 +482,7 @@ def val_and_grad_fn_local(x_local):
         u_full, grad_full = val_and_grad_fn(x_prime)
         # avoid being trapped when overflows spuriously appear as large negative numbers
         # remove after resolution of https://github.com/proteneer/timemachine/issues/481
-        if u_0 - u_full > guard_threshold:
+        if not np.isfinite(u_full) or u_0 - u_full > guard_threshold:
```
mcwitt, line 485: Is `guard_threshold` still relevant here, with the added overflow handling?

badisa: Yes, you can still add N energies of `MAX_LONG_LONG - 1` that will overflow in a way that is not account for. There will be a follow-up PR that handles that case. There is a failing test of this latter case: https://github.com/proteneer/timemachine/blob/feature/detect-nb-clashes/tests/test_energy_overflows.py#L249-L282

timemachine/cpp/src/kernels/k_nonbonded_pair_list.cuh
mcwitt: It might be worth noting that the (negated version of) `NonbondedPairList` is specialized to handling exclusions. I'm not sure of other use cases, but it seems like this logic could lead to incorrect results in the case that the negated pair list isn't used together with a nonbonded potential with the same interactions (since overflows in different terms could erroneously "cancel").

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -61,6 +62,23 @@ void verify_coords_and_box(
     }
 }
 
+// accumulate_energies handles the combining of energies, summing them up deterministically
+// and returning nan if there are overflows.
+// The energies are collected in unsigned long long and overflows are incremented or decremented
+// by the potentials. Currently only the negated NonbondedAllPairs decrements the overflows which
+// are expected to exactly cancel out with other Nonbonded potentials if there are no clashes.
+double accumulate_energies(int N, unsigned long long *fixed_u, int overflows) {
```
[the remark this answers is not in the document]
badisa: Mostly for simplicity. If we simply return a single value (+inf/nan), the exclusions can decrement and the nonbonded can increment. Otherwise, they would need to both increment/decrement according to the sign of the overflow, which is more difficult to debug. Though the `+inf/-inf` distinction would be useful to the end user? 

Is there a use case for distinguishing between for +/- inf?

tests/test_energy_overflows.py
```diff
@@ -0,0 +1,282 @@
...
+    assert not np.isfinite(test_energy) and np.isnan(test_energy)
+
+    # Verify that there was an overflow that wasn't cancelled out
+    _, overflows = gpu_bound.execute_fixed(x0, box)
+    assert overflows != 0
+
+
+@pytest.mark.parametrize("precision,rtol,atol", [(np.float64, 1e-8, 1e-8), (np.float32, 1e-4, 5e-4)])
+def test_energy_overflow_max_representation(precision, rtol, atol):
+    """Validates that the max representation that the GPU platform is less than that of the reference platform as the GPU
```
mcwitt, line 68: nit: fix, maybe
```suggestion
    """Validates that the max representable fixed-point value on the GPU platform is less than that of the reference platform
```
?

tests/test_energy_overflows.py
```diff
@@ -0,0 +1,282 @@
...
+    nonbonded = Nonbonded(
+        num_atoms=num_atoms,
+        exclusion_idxs=np.array([(0, num_atoms - 1)], dtype=np.int32),
+        scale_factors=np.zeros((1, 2)),
+        beta=2.0,
+        cutoff=1.2,
+    )
+    nonbonded_gpu = nonbonded.to_gpu(precision)
+
+    # TBD: Remove the try/finally to clean up the GPU potential, else calling reset_device
```
mcwitt, line 273: Is this TBD for this PR, or future? (not sure I understand why explicit cleanup is needed in this case)

badisa: For the future. If you don't have this, the `device_reset` will clear the memory, which will result in a failure of compute sanitizer due to it thinking the memory is lost.

I chock this up to python garbage collection not playing well with C++ objects. There is a call to the gc in the tests, but doesn't clean up this object. Doing it explicitly fixes it, and when this test starts passing we can remove the try/fnally

timemachine/cpp/src/kernels/k_nonbonded_pair_list.cuh
badisa: Added docstring to the NonbondedPairList being more explicit about its use case [6221fad](https://github.com/proteneer/timemachine/pull/1084/commits/6221fad25f36f3b55785db0fd570da8f6f05ec43)

tests/test_energy_overflows.py
```diff
@@ -0,0 +1,282 @@
...
+    assert overflows == 0
+
+    # Set the two particles to be overlapping
+    x0[0, :] = x0[1, :]
+
+    # Note that the reference potential returns an inf, while the unbound returns a nan
+    reference_energy = potential(x0, params, box)
+    assert not np.isfinite(reference_energy)
+    _, _, test_energy = gpu_unbound.execute_selective(x0, params, box, False, False, True)
+    assert not np.isfinite(test_energy) and np.isnan(test_energy)
```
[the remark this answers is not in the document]
badisa: Removed in [d90fb8e](https://github.com/proteneer/timemachine/pull/1084/commits/d90fb8eedf6f2988d142b98cb9179ad1390c58ae)

timemachine/cpp/src/potential.hpp
```diff
@@ -22,7 +22,8 @@ class Potential {
         const double *h_box,
         unsigned long long *h_du_dx,
         unsigned long long *h_du_dp,
-        unsigned long long *h_u);
+        unsigned long long *h_u,
+        int *h_overflow_count);
```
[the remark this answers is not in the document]
badisa: Renamed in [11154da](https://github.com/proteneer/timemachine/pull/1084/commits/11154dad9904f3037bb4c9721bdf8b5a9eb41a8b)

mcwitt (approved):

maxentile (approved):
Nice work!

Handles one of the main cases that can lead to undetected numerical overflow in the nonbonded implementation.

timemachine/cpp/src/fixed_point.hpp
```diff
@@ -16,3 +16,17 @@ RealType __host__ __device__ __forceinline__ FIXED_TO_FLOAT_DU_DP(unsigned long
 template <typename RealType> RealType __host__ __device__ __forceinline__ FIXED_TO_FLOAT(unsigned long long v) {
     return static_cast<RealType>(static_cast<long long>(v)) / FIXED_EXPONENT;
 }
+
+template <typename RealType> bool inline __device__ is_overflow(RealType u) {
+    return u > nextafter(static_cast<RealType>(LLONG_MAX), static_cast<RealType>(0.0)) ||
+           u < nextafter(static_cast<RealType>(LLONG_MIN), static_cast<RealType>(0.0));
+}
+
+template <typename RealType> bool inline __device__ energy_overflowed(RealType u_orig, int &overflow_count) {
```
proteneer, line 25: nice! this looks right to me now.

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -61,6 +62,23 @@ void verify_coords_and_box(
...
+    double res = NAN;
```
proteneer, line 71: is `NAN` the same thing as `std::nan`? not sure where this is defined

proteneer (approved):
lgtm - looking forward to Pt2!

proteneer:
One last question: how much is the slowdown for barostat benchmarks?

badisa:
> One last question: how much is the slowdown for benchmarks that use a barostat?

Doesn't appear to make any noticeable difference:
```
# Current
dhfr-apo-barostat-interval-25: N=23558 speed: 635.75ns/day dt: 2.5fs (ran 100000 steps in 33.98s)
hif2a-apo-barostat-interval-25: N=8805 speed: 1022.51ns/day dt: 2.5fs (ran 100000 steps in 21.13s)
solvent-apo-barostat-interval-25: N=6282 speed: 1290.74ns/day dt: 2.5fs (ran 100000 steps in 16.74s)

# Changes
dhfr-apo-barostat-interval-25: N=23558 speed: 633.08ns/day dt: 2.5fs (ran 100000 steps in 34.12s)
hif2a-apo-barostat-interval-25: N=8805 speed: 1021.63ns/day dt: 2.5fs (ran 100000 steps in 21.15s)
solvent-apo-barostat-interval-25: N=6282 speed: 1305.74ns/day dt: 2.5fs (ran 100000 steps in 16.55s)
```

PR #1085, opened 2023-07-14
[wip] Return inf not nan from jax potentials

maxentile:
Possibly cleans up a case @proteneer encountered in https://github.com/proteneer/timemachine/pull/1083

Options: 
* convert nans to infs only in `lennard_jones`
* convert nans to infs in other jax potentials (unsure)
* convert nans to infs in downstream code (such as Metropolis checks https://github.com/proteneer/timemachine/pull/1083/commits/36d5d1fd0ac74054d9852c50302401439be1d60e )

### DIFF
--- a/tests/test_jax_nonbonded.py
+++ b/tests/test_jax_nonbonded.py
@@ -477,3 +477,15 @@ def test_precomputation():
 
         np.testing.assert_allclose(v_ref, v_test)
         np.testing.assert_allclose(gs_ref, gs_test)
+
+
+def test_lj_not_nan():
+    distances = [0.0, 0.00000001, 0.00001, 0.001, np.inf]
+
+    def U(r):
+        return lennard_jones(jnp.array(r), 0.3, 0.1)
+
+    for r in distances:
+        nrg = U(r)
+        assert not np.isnan(nrg)
+        assert np.sign(nrg) > 0
--- a/timemachine/potentials/jax_utils.py
+++ b/timemachine/potentials/jax_utils.py
@@ -12,6 +12,10 @@ Array: TypeAlias = NDArray
 DEFAULT_CHUNK_SIZE = 200
 
 
+def nan_to_inf(x):
+    return jnp.nan_to_num(x, nan=+jnp.inf)
+
+
 def get_all_pairs_indices(n: int) -> Array:
     """all indices i, j such that i < j < n"""
     n_interactions = n * (n - 1) / 2
--- a/timemachine/potentials/nonbonded.py
+++ b/timemachine/potentials/nonbonded.py
@@ -12,6 +12,7 @@ from timemachine.potentials.jax_utils import (
     DEFAULT_CHUNK_SIZE,
     delta_r,
     distance_on_pairs,
+    nan_to_inf,
     pairs_from_interaction_groups,
     pairwise_distances,
     process_traj_in_chunks,
@@ -48,7 +49,8 @@ def lennard_jones(dij, sig_ij, eps_ij):
     sig6 = (sig_ij / dij) ** 6
     sig12 = sig6 ** 2
 
-    return 4 * eps_ij * (sig12 - sig6)
+    nrg = 4 * eps_ij * (sig12 - sig6)  # +inf - +inf -> nan
+    return nan_to_inf(nrg)  # nrg can be nan if dij == 0
 
 
 def direct_space_pme(dij, qij, beta):


tests/test_jax_nonbonded.py
```diff
@@ -477,3 +477,15 @@ def reweight(sig_ligand, eps_ligand, q_ligand):
 
         np.testing.assert_allclose(v_ref, v_test)
         np.testing.assert_allclose(gs_ref, gs_test)
+
+
+def test_lj_not_nan():
+    distances = [0.0, 0.00000001, 0.00001, 0.001, np.inf]
```
proteneer, line 483: I think we want to be in 32-bit mode when we test this - 64-bit has significantly more precision. Does `0.00000001` result in a `nan` in 64-bit as well?

proteneer: (I thought all the `test_jax_nonbonded.py` files ran in 64bit-jax, but I can't actually find the config line at the top anymore)

proteneer:
+1 to putting the fix in the `lennard_jones` function.

tests/test_jax_nonbonded.py
maxentile: >I think we want to be in 32-bit mode when we test this - 64-bit has significantly more precision. Does 0.00000001 result in a nan in 64-bit as well?

Ahh, good call -- 0.0 results in a nan in 64-bit, but 0.00000001 does not. 

Updated test to exercise both 32-bit and 64-bit mode: https://github.com/proteneer/timemachine/pull/1085/commits/f0f8bf5a56b0711303208584ecdfe4b98e32af12 ( using https://jax.readthedocs.io/en/latest/_autosummary/jax.experimental.enable_x64.html#jax.experimental.enable_x64 )

proteneer (approved):
lgtm!

proteneer:
One more note, I think the `direct_space_pme` may also return a `nan` when `dij` is very close to 0. This due to `qij * erfc(beta * dij) / dij`. We can correct this since the limiting behavior for small `dij` result in either`+inf` or `-inf`

However, if we sum up `lennard_jones` with `direct_space_pme`. A `nan` can still arises if the `direct_space_pme` returns `-inf` and `lj` returns `+inf`. I think application specific code probably still needs to check for `nan` when this happens.

maxentile:
>if we sum up lennard_jones with direct_space_pme, a nan can still arises if the direct_space_pme returns -inf and lj returns +inf

Thank you  -- good point!

Proposal: The overall potential energy function should not be defined as "nan" at any point x that we can evaluate. For the special case of two oppositely charged (LJ + ES) particles at distance 0 from each other, it would be reasonable to define U(x) to be +inf rather than nan. Added another check in https://github.com/proteneer/timemachine/pull/1085/commits/bb0b24ffcb84a9dcfab51c4cb8e9e7da7dce1e9d .

(It may still be necessary to be defensive and check for nans in other cases (I guess, at least, any case where oppositely charged particles are collocated?), but this might reduce the frequency of those cases.)

proteneer:
Whether or not the combined expression `lj+es` numerically evaluates to `+inf` or `-inf` instead of `nan` is dependent on the parameters `sig,eps,q` and the distance `r`. You can have a system that evalutes to `+inf` at `r=0` but `-inf` at `r=0.01` if `sig` is small:

Screenshot from 2023-07-18 13-22-14

Note that the black line (small vdw radius) sharply drops to `-inf` before returning to `+inf`. Whether or not the resulting combined `lj+es` result actually should be a `+inf` or `-inf` depends on `r` and the parameters in a very sensitive way.

> Proposal: The overall potential energy function should not be defined as "nan" at any point x that we can evaluate. 

We should decide if:
1) we should leave it up to the caller of the function to disambiguate if a returned `nan` (under a given set of parameters) should be interpreted as `+inf` or `-inf` or,
2) we should make the decision on behalf of the caller by introspection of the parameters within these functions (which we currently do not do).

I'm okay with either solution, but we should be careful if we implement 2).

maxentile:
>We should decide if: ... 

I'm not sure.

* It should probably be invalid to define a system where U(x) = -inf at any point x. (Since it'll make `p(x) \propto exp(-beta U(x)) -> +inf`, making `p(x)` unnormalizable .) (i.e. it should be invalid to define a system with "naked charges")

* It should probably also be invalid to define a system where U(x) = nan at any point x. (what's the normalizing constant \int dx p(x) if p(x) is nan in some places?)

* It seems practical to define p(x) = 0 at all points x where the implementation for any reason cannot compute the energy. (For example, by converting U(x) = nan to U(x) = +inf.)

maxentile:
>It should probably be invalid to define a system where U(x) = -inf at any point x. (Since it'll make p(x) \propto exp(-beta U(x)) -> +inf, making p(x) unnormalizable .) (i.e. it should be invalid to define a system with "naked charges")

Complication I had overlooked (thanks for comments from @proteneer and @badisa last week):
* TIP3P assigns LJ eps = 0 to hydrogens ([ref in openff-forcefields](https://github.com/openforcefield/openff-forcefields/blob/19d29a8c39836bfdad675a76caa4c78f09ec22fe/openforcefields/offxml/tip3p.offxml#L5))
* Consequently, there are configurations of **flexible** TIP3P where the energy goes to -inf, as distances between oppositely charged particles approach 0.

Example with 2 waters: (but note that this configuration is not possible with rigid-body constraints -- bonds are stretched!)

```python
from jax import config
config.update("jax_enable_x64", True)
from jax import jit

import numpy as np
from timemachine.md.builders import build_water_system
from timemachine.ff.handlers import openmm_deserializer

def make_water_cluster_potentials():
    """get bound potentials for an n=2 water cluster in effectively open boundary conditions"""
    system, positions, _, _ = build_water_system(
        0.50, # very small box width --> just a couple waters
        water_ff=ff.water_ff
    )
    num_waters = len(positions) // 3
    assert num_waters == 2

    bps, _ = openmm_deserializer.deserialize_system(system, cutoff=1.2)

    open_boundary_conditions_box = 100.0 * np.eye(3)

    @jit
    def U_ref(x):
        return sum(bp(x, open_boundary_conditions_box) for bp in bps)
    
    gpu_bps = []
    for bp in bps:
        gpu_bps.append(bp.to_gpu(np.float32))
        
    def U_gpu(x):
        return sum(gpu_bp(x, open_boundary_conditions_box) for gpu_bp in gpu_bps)
    
    return U_ref, U_gpu


U_ref, U_gpu = make_water_cluster_potentials()

x = np.array([[ 0.                   ,  0.                   ,
         0.                   ],
       [ 0.23468019182676886  , -0.27501762075242653  ,
         0.20330731275297725  ],
       [ 0.2348330960655381   , -0.2742542244781153   ,
         0.20311446995569982  ],
       [ 0.23468022555779056  , -0.2750175955775044   ,
         0.20330733589134378  ],
       [-0.000466390775090475 , -0.0041670468725262255,
         0.0009657274517013703],
       [ 0.2619482156903966   , -0.3504305740722373   ,
         0.13028281682847767  ]])
    
print(U_ref(x))  # -1005997184.9894862
print(U_gpu(x))  # +134217716.60884753 prior to PR1084, now nan
```

On this configuration of 2 (stretched) waters, U_ref is a large **negative** number.
(Prior to https://github.com/proteneer/timemachine/pull/1084 , U_gpu(x) was a large **positive** number, and since https://github.com/proteneer/timemachine/pull/1084 , U_gpu(x) is **nan**.)

This example shows that all systems containing flexible TIP3P would currently be "invalid", if we were to apply the standard "p(x) must not be defined as exp(+inf) at any point x". This may be something to patch by giving the water hydrogens LJ parameters (non-zero epsilon and a small radius, as in Charmm etc.), so that p(x) goes to exp(-inf) instead at these points.

Configurations like this are effectively never seen with MD at typical step sizes. However, configurations like this could be generated by MC proposals (and spuriously accepted) in principle.

maxentile:
Closing for now -- I think it makes more sense to handle the causes in system setup (e.g. ruling out "naked charges") or convert nan_to_inf in specific applications (as in https://github.com/proteneer/timemachine/pull/1098 ), rather than attempt to add special cases like this in the definition of the energy function.

PR #1087, opened 2023-07-19
Separate host-host nonbonded interactions in HostGuestSystem

mcwitt:
Currently, the single topology code sets up a single summed potential (`nonbonded_host_guest`) containing both host-host and host-guest nonbonded interactions. Host-host parameters are independent of lambda, while host-guest parameters depend on lambda (via 4-d decoupling).

This PR gives the host-guest component a separate handle in `HostGuestSystem`, enabling more efficient computation of energy differences with constant conformation and varying lambda (since we can skip recomputing the expensive host-host interactions). Specifically, `nonbonded_host_guest` is split into

- `nonbonded_host`: nonbonded host-host interactions
- `nonbonded_host_guest`: nonbonded host-guest interactions (i.e. not including host-host as before)


tests/test_benchmark.py
```diff
@@ -455,18 +454,11 @@ def test_hif2a():
...
-                    return pot, params
-    return None, None
-
-
 def test_nonbonded_interaction_group_potential(hi2fa_test_frames):
     bps, frames, boxes, ligand_idxs = hi2fa_test_frames
-    nonbonded_potential, nonbonded_params = get_nonbonded_pot_params(bps)
+    nonbonded_potential, nonbonded_params = next(
+        (bp.potential, bp.params) for bp in bps if isinstance(bp.potential, Nonbonded)
+    )
```
jkausrelay, line 461: nit: This still could be a function to hide the complexity

timemachine/fe/single_topology.py
```diff
@@ -1235,6 +1232,19 @@ def _parameterize_host_guest_nonbonded(
             atom_idxs=np.arange(num_host_atoms, dtype=np.int32),
         )
 
+        return combined_nonbonded.bind(hg_nb_params)
+
+    def _parameterize_host_guest_nonbonded(
```
jkausrelay, line 1237: nit: _parameterize_host_guest_nonbonded_ixn to emphasize this is now the ixn only and not the full host-guest potential

timemachine/fe/system.py
```diff
@@ -158,15 +158,15 @@ def get_U_fns(self) -> List[BoundPotential[Potential]]:
...
+class HostGuestSystem:
     bond: BoundPotential[HarmonicBond]
     angle: BoundPotential[HarmonicAngleStable]
     torsion: BoundPotential[PeriodicTorsion]
     chiral_atom: BoundPotential[ChiralAtomRestraint]
     chiral_bond: BoundPotential[ChiralBondRestraint]
     nonbonded_guest_pairs: BoundPotential[NonbondedPairListPrecomputed]
-    nonbonded_host_guest: BoundPotential[_Nonbonded]
+    nonbonded_host: BoundPotential[Nonbonded]
+    nonbonded_host_guest: BoundPotential[SummedPotential]
```
jkausrelay, line 169: nit: nonbonded_host_guest_ixn

timemachine/fe/single_topology.py
```diff
@@ -1251,21 +1261,19 @@ def get_other_idxs():
...
             guest_ixn_water_params,
             guest_ixn_other_params,
-            beta=beta,
+            beta=host_nonbonded.potential.beta,
             cutoff=cutoff,
         )
 
-        hg_total_pot = [combined_nonbonded] + ixn_pots
-        hg_total_params = [hg_nb_params] + ixn_params
-        sum_pot = SummedPotential(hg_total_pot, hg_total_params)
```
jkausrelay, line 1263: Are there any performance considerations for this change?

timemachine/ff/handlers/nonbonded.py
```diff
@@ -259,7 +259,7 @@ def compute_or_load_bond_smirks_matches(mol, smirks_list):
     return np.array(bond_idxs), np.array(type_idxs)
 
 
-def apply_bond_charge_corrections(initial_charges, bond_idxs, deltas):
+def apply_bond_charge_corrections(initial_charges, bond_idxs, deltas, runtime_validate=True):
```
maxentile, line 262: Nice -- this corrects an oversight on my part from https://github.com/proteneer/timemachine/pull/571

maxentile (commented):
Tangent: Will this introduce any considerations for implementing "gradually turn a selected water molecule off and on" in a SingleTopology calculation?

(I guess, may just have to index into one more place to update the selected host NB parameters: host-host and host-guest... Or maybe the alchemical water could be added to guest_idxs.)

mcwitt:
> Tangent: Will this introduce any considerations for implementing "gradually turn a selected water molecule off and on" in a SingleTopology calculation?
> 
> (I guess, may just have to index into one more place to update the selected host NB parameters: host-host and host-guest... Or maybe the alchemical water could be added to guest_idxs.)

I think I would probably implement this by updating the nonbonded parameters in `host_system` before calling `SingleTopology.combine_with_host`; this should be agnostic to the changes here. Does that seem reasonable?

timemachine/fe/single_topology.py
mcwitt: Renamed in https://github.com/proteneer/timemachine/pull/1087/commits/f38038adf130f0516b958232480abc0f66dbfd1e

tests/test_benchmark.py
mcwitt: Changed in https://github.com/proteneer/timemachine/pull/1087/commits/c4309ed4b843f9201b16e8d71c3e712af7c16893

maxentile:
>I don't think this change really affects the difficulty of updating host atom parameters, since this would currently have to be done in 2 places anyway (in the host-host and ligand-water components of the nonbonded_host_guest summed potential). With this change, this gets slightly easier (if anything), since it would require unwrapping one fewer summed potential to modify the parameters inside.

Thanks @mcwitt !

timemachine/fe/single_topology.py
mcwitt: I've verified by profiling `test_relative_free_energy` that there is no significant performance impact.

jkausrelay (approved):

timemachine/fe/bar.py
```diff
@@ -235,7 +204,8 @@ def df_from_ukln_by_lambda(ukln_by_lambda: NDArray) -> Tuple[float, float]:
     for lambda_idx in range(ukln_by_lambda.shape[0]):
         window_ukln = ukln_by_lambda[lambda_idx]
 
-        dF, dF_err = df_from_ukln(window_ukln)
+        w_fwd, w_rev = works_from_ukln(window_ukln)
+        dF, dF_err = pymbar.BAR(w_fwd, w_rev)
```
maxentile, line 208: nit: units: are the outputs assumed in kBT or kJ/mol?

tests/test_benchmark.py
```diff
@@ -455,20 +454,17 @@ def test_hif2a():
...
-    for bp in bps:
-        if isinstance(bp.potential, SummedPotential):
-            for pot, params in zip(bp.potential.potentials, bp.potential.params_init):
-                if isinstance(pot, Nonbonded):
-                    return pot, params
-    return None, None
+def get_nonbonded_pot_params(bps):
+    try:
+        pot, params = next((bp.potential, bp.params) for bp in bps if isinstance(bp.potential, Nonbonded))
+        return pot, params
```
maxentile, line 460: Could you add a quick comment about this change? (Is it no longer nested?)

maxentile (approved):
Looks good!

timemachine/fe/bar.py
mcwitt: Outputs are in kT. (The only change here was to pull out a helper function to compute forward and reverse works from the reduced energies $u_{k l n}$, and then eliminate `df_from_ukln` and `df_err_from_ukln`, which just become 2-liners). It didn't seem necessary to state the output units explicitly since this seems to be the convention for other functions in this module. Do you think we should update this and the others?

tests/test_benchmark.py
mcwitt: With this PR, the `Nonbonded` potential is no longer nested inside of a `SummedPotential`, so we don't need to search inside of `SummedPotential`s. Can add a quick docstring if it would be helpful.

mcwitt: Actually, was your question more about what this function does, or why the change was needed? Hopefully the above comment addresses why it was needed, and the simplification in https://github.com/proteneer/timemachine/pull/1087/commits/49e5755c9c13ffc046bae81059d20d8239d4d0e3 should make the code pretty self-explanatory

timemachine/fe/bar.py
maxentile: Sorry, this was a tangent -- not a change introduced by this PR, but just showed up in the diff.

(convention: I prefer to use lower case for reduced quantities `u`, `w`, `df` (in kBT), uppercase `U`, `W`, `dF` (in kJ/mol). The code near this diff already mixed these conventions.)

tests/test_benchmark.py
maxentile: Thanks for clarifying -- makes sense, and good additional simplification!

maxentile (approved):

timemachine/fe/bar.py
mcwitt: Ah, I see. I agree with the preference for lower-case here, but would prefer to change it in a separate PR as it would introduce a fair amount of noise to the diff here.

PR #1089, opened 2023-07-19
Refactor using the existing MonteCarloMove class.

proteneer:
Refactors the water_exchange branch to use existing movers:

1) Re-use existing CoordsVelBox structure to represent state. 
2) ExchangeMove now subclasses `MonteCarloMove` and implements the abstract `proposal` method
3) Propagate moves replaced by NPTMove.

### DIFF
--- a/timemachine/exchange/exchange_mover.py
+++ b/timemachine/exchange/exchange_mover.py
@@ -5,6 +5,7 @@
 
 import argparse
 import os
+from typing import Tuple
 
 import jax
 import jax.numpy as jnp
@@ -21,9 +22,11 @@ from timemachine.fe.topology import BaseTopology
 from timemachine.fe.utils import get_romol_conf
 from timemachine.ff import Forcefield
 from timemachine.ff.handlers.nonbonded import PrecomputedChargeHandler
-from timemachine.lib import LangevinIntegrator, MonteCarloBarostat, custom_ops
+from timemachine.lib import LangevinIntegrator, MonteCarloBarostat
+from timemachine.md import moves
 from timemachine.md.barostat.utils import get_bond_list, get_group_indices
 from timemachine.md.builders import strip_units
+from timemachine.md.states import CoordsVelBox
 from timemachine.potentials import nonbonded
 
 # uncomment if we want to re-enable minimization
@@ -125,7 +128,7 @@ def get_initial_state(water_pdb, mol, ff, seed, nb_cutoff):
     return initial_state, num_water_mols, solvent_topology
 
 
-class ExchangeMover:
+class ExchangeMove(moves.MonteCarloMove):
     def __init__(self, nb_params, nb_beta, nb_cutoff, water_idxs):
         self.nb_beta = nb_beta
         self.nb_cutoff = nb_cutoff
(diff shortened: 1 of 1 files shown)


timemachine/exchange/exchange_mover.py
```diff
@@ -125,7 +128,7 @@ def get_initial_state(water_pdb, mol, ff, seed, nb_cutoff):
     return initial_state, num_water_mols, solvent_topology
 
 
-class ExchangeMover:
+class ExchangeMove(moves.MonteCarloMove):
```
maxentile, line 131: Nice -- can now remove unused `a_count`, `t_count` fields.

(should be handled by `n_accepted`, `n_proposed` counters inherited from https://github.com/proteneer/timemachine/blob/68040c238553a210164cdb1dc01d0cb71cd394a1/timemachine/md/moves.py#L20-L22 )

maxentile: resolved in https://github.com/proteneer/timemachine/pull/1089/commits/8f127e1a5d185e55f2c1da99743876d1bf74a354

timemachine/exchange/exchange_mover.py
```diff
@@ -173,68 +174,17 @@ def exchange_move(self, coords, box):
...
-                nb_params=self.nb_params,
-                coords=coords,
-                a_idxs=a_idxs,
-                b_idxs=b_idxs,
-            )
-            print("MC overflow detected", delta_U_delete, delta_U_insert)
-
         # convert to inf if we get a nan
         if np.isnan(delta_U_total):
             delta_U_total = np.inf
```
maxentile, line 182: Side note: this conversion is no longer necessary, since the implementation will never accept a proposal whose `log_acceptance_probability` is nan
https://github.com/proteneer/timemachine/blob/68040c238553a210164cdb1dc01d0cb71cd394a1/timemachine/md/moves.py#L32-L34.

i.e.
```python
alpha < np.exp(np.nan)
```
will always evaluate to `False` for any realizable value of `alpha`.

Side-side note: Maybe we want to log / warn when `log_acceptance_probability` is nan?

timemachine/exchange/exchange_mover.py
```diff
@@ -173,68 +174,17 @@ def exchange_move(self, coords, box):
...
-    ctxt = custom_ops.Context(
-        initial_state.x0,
-        initial_state.v0,
-        initial_state.box0,
-        intg_impl,
-        bp_impls,
-        baro_impl,
-    )
-    return ctxt
+        log_p_accept = min(0, -self.beta * delta_U_total)
```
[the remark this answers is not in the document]
proteneer: I will test `np.minimum` more carefully in a separate PR, just trying to keep the logic identical and only attempt the refactor.

timemachine/exchange/exchange_mover.py
proteneer: Ah good point

proteneer: Note that this does require us to propagate the delta_U `nan` into a log_acceptance_probability `nan`, i.e. I think this can only be removed after we change `min` to `np.min` below?

proteneer:
Let me do the nan clean-up logic in a separate, isolated pass!

PR #1090, opened 2023-07-19
Detect Energy Overflows in Summation

badisa:
- [x] Fixes a memory leak in Barostat and adds test that verifies leak is resolved
- [x] Swap all potentials to store their own energy buffers, as atomicAdds not possible on int128
- [x] Replace unsigned long long and overflow counter with int128 and if final value is >= LLONG_MAX or <= LLONG_MIN return nan


mcwitt:
Could you comment on the motivation for moving to int128? E.g. does it seem likely that we would overflow `unsigned long long` for non-clashy systems? Also, is there any noticeable performance impact of this change?

tests/common.py
```diff
@@ -230,6 +231,9 @@ def compare_forces(
         params = (params.astype(np.float32)).astype(np.float64)
         assert params.dtype == np.float64
         ref_u = ref_potential(x, params, box)
+        assert (
+            np.abs(ref_u) < np.iinfo(np.int64).max / FIXED_EXPONENT
+        ), "System is invalid, GPU platform unable to represent energies"
```
mcwitt, line 236: nit: seems like it would be nice to import `FIXED_EXPONENT` from `custom_ops`, to avoid having to keep Python and C++ definitions in sync. Would it be possible to export this from the wrapper?

tests/test_energy_overflows.py
```diff
@@ -18,6 +18,10 @@ def verify_energies(a, b):
         assert np.isnan(a) and np.isnan(b)
 
 
+def fixed_overflowed(a):
+    return a <= np.iinfo(np.int64).min or a >= np.iinfo(np.int64).max
```
mcwitt, line 22: Not sure I understand why this uses `np.int64`, when I thought we were moving to `int128`? Docstring here seems useful.

mcwitt: Sorry, premature comment. I think I understand this now. (I think docstring would still be useful)

proteneer:
Some background:

1) Currently, MAX_INT64 is 2^63-1, MAX_INT128 is 2^127-1, and our FIXED_EXPONENT for energies 68,719,476,736
2) We want to avoid using `unsigned long long`s for the energy. Adding two large positive signed numbers can overflow into a negative signed number.
3) We need to decide if an interaction `u` is `nan`, `+inf` or `-inf` (eg. exclusions), or a number that can overflow the signed integer representation (eg. clashing state when proposed by a MC move) when a certain GUARD threshold is set. we do `np.clip(u, -GUARD, +GUARD)` on every interaction even if `u` is not finite. 
4) In fixed point using INT64, the maximum energy total that we can represent is MAX_INT64/FIXED_EXPONENT=134,217,728 kJ/mol. If the energy is greater than this we will overflow the INT64 buffer in an undefined way. This is what we want to avoid. 
5) In fixed point using INT128, the maximum energy total that we can represent is MAX_INT128/FIXED_EXPONENT ~= 2^64 kJ/mol. This is very very unlikely if we set a reasonable GUARD threshold.

Suppose we set a GUARD threshold of +-1,000,000 kJ/mol. i.e. If we do this, then we can only accommodate 134 interactions of magnitude GUARD (exclusions + clashy MC moves). A water box with 10,000 waters typically has 20,000 exclusions. So we'd overflow. So the larger the system, the lower the GUARD. 

If we used INT128 instead, we can set a GUARD threshold of MAX_INT64, and still have room for 2^63 GUARD interactions until we overflow the INT128 representation. This is an extremely high value (it would take us 292 years to compute the interactions if each interaction took 1 nanosecond).

timemachine/cpp/src/centroid_restraint.cu
```diff
@@ -20,7 +20,7 @@ CentroidRestraint<RealType>::CentroidRestraint(
     cudaSafeMalloc(&d_centroid_a_, 3 * sizeof(*d_centroid_a_));
     cudaSafeMalloc(&d_centroid_b_, 3 * sizeof(*d_centroid_b_));
 
-    cudaSafeMalloc(&d_u_buffer_, (N_A_ + N_B_) * sizeof(d_u_buffer_));
```
badisa: Previously I had read the kernels incorrectly, so this was `N_A + N_B`, but really it is only a single energy value. 

Also notice `sizeof(d_u_buffer)` rather than `sizeof(*d_u_buffer)`, which previously worked because the size of the pointer and the value was identical. Not the case with int128

timemachine/cpp/src/nonbonded_all_pairs.cu
```diff
@@ -71,9 +71,9 @@ NonbondedAllPairs<RealType>::NonbondedAllPairs(
     cudaSafeMalloc(&d_rebuild_nblist_, 1 * sizeof(*d_rebuild_nblist_));
     gpuErrchk(cudaMallocHost(&p_rebuild_nblist_, 1 * sizeof(*p_rebuild_nblist_)));
 
-    cudaSafeMalloc(&d_sort_keys_in_, N_ * sizeof(d_sort_keys_in_));
-    cudaSafeMalloc(&d_sort_keys_out_, N_ * sizeof(d_sort_keys_out_));
-    cudaSafeMalloc(&d_sort_vals_in_, N_ * sizeof(d_sort_vals_in_));
+    cudaSafeMalloc(&d_sort_keys_in_, N_ * sizeof(*d_sort_keys_in_));
```
badisa, line 74: With the previous commit, found a couple places where we were allocating more memory than we needed

tests/common.py
badisa: Will make a case to do this in a follow-up PR

tests/test_energy_overflows.py
badisa: Added comment point the user to read the C++ docs : [2cc4964](https://github.com/proteneer/timemachine/pull/1090/commits/2cc4964310da1dbc9dac9d2e7b08c1b323f2e5ad)

badisa:
> Also, is there any noticeable performance impact of this change?


Significant performance hit when it comes to running potentials for energy and barostat. There are definitely some improvements to make to the barostat, in the process of this PR found a lot to improve. Though looks like the Nonbonded potentials got a lot slower, of which there are several things we could try to fix that.  

Benchmarks from A10
```
# Master
dhfr-apo: N=23558 speed: 694.74ns/day dt: 2.5fs (ran 100000 steps in 31.09s)
dhfr-apo-barostat-interval-25: N=23558 speed: 634.76ns/day dt: 2.5fs (ran 100000 steps in 34.03s)

hif2a-apo: N=8805 speed: 1170.99ns/day dt: 2.5fs (ran 100000 steps in 18.45s)
hif2a-apo-barostat-interval-25: N=8805 speed: 1020.28ns/day dt: 2.5fs (ran 100000 steps in 21.17s)
hif2a-rbfe: N=8840 speed: 895.51ns/day dt: 2.5fs (ran 100000 steps in 24.12s)
hif2a-rbfe-local: N=8840 speed: 1310.05ns/day dt: 2.5fs (ran 100000 steps in 16.49s)
solvent-apo: N=6282 speed: 1452.91ns/day dt: 2.5fs (ran 100000 steps in 14.87s)
solvent-apo-barostat-interval-25: N=6282 speed: 1285.22ns/day dt: 2.5fs (ran 100000 steps in 16.81s)
solvent-rbfe: N=6317 speed: 1173.01ns/day dt: 2.5fs (ran 100000 steps in 18.42s)
solvent-rbfe-local: N=6317 speed: 1415.63ns/day dt: 2.5fs (ran 100000 steps in 15.26s)

NonbondedInteractionGroup_f32: N=8840 Frames=1000 Params=5 speed: 1181.91 executions/seconds (ran 25000 potentials in 21.15s) du_dp=True, du_dx=True, u=True
NonbondedInteractionGroup_f64: N=8840 Frames=1000 Params=5 speed: 614.60 executions/seconds (ran 25000 potentials in 40.68s) du_dp=True, du_dx=True, u=True
Nonbonded_f32: N=8840 Frames=1000 Params=5 speed: 953.04 executions/seconds (ran 25000 potentials in 26.23s) du_dp=True, du_dx=True, u=True
Nonbonded_f64: N=8840 Frames=1000 Params=5 speed: 128.94 executions/seconds (ran 25000 potentials in 193.88s) du_dp=True, du_dx=True, u=True
HarmonicBond_f32: N=8840 Frames=1000 Params=5 speed: 1510.20 executions/seconds (ran 25000 potentials in 16.55s) du_dp=True, du_dx=True, u=True
HarmonicBond_f64: N=8840 Frames=1000 Params=5 speed: 1507.59 executions/seconds (ran 25000 potentials in 16.58s) du_dp=True, du_dx=True, u=True
HarmonicAngleStable_f32: N=8840 Frames=1000 Params=5 speed: 1464.99 executions/seconds (ran 25000 potentials in 17.07s) du_dp=True, du_dx=True, u=True
HarmonicAngleStable_f64: N=8840 Frames=1000 Params=5 speed: 1457.21 executions/seconds (ran 25000 potentials in 17.16s) du_dp=True, du_dx=True, u=True
PeriodicTorsion_f32: N=8840 Frames=1000 Params=5 speed: 1470.02 executions/seconds (ran 25000 potentials in 17.01s) du_dp=True, du_dx=True, u=True
PeriodicTorsion_f64: N=8840 Frames=1000 Params=5 speed: 1444.91 executions/seconds (ran 25000 potentials in 17.30s) du_dp=True, du_dx=True, u=True
NonbondedPairListPrecomputed_f32: N=8840 Frames=1000 Params=5 speed: 1610.37 executions/seconds (ran 25000 potentials in 15.52s) du_dp=True, du_dx=True, u=True
NonbondedPairListPrecomputed_f64: N=8840 Frames=1000 Params=5 speed: 1589.88 executions/seconds (ran 25000 potentials in 15.72s) du_dp=True, du_dx=True, u=True

NonbondedInteractionGroup_f32: N=8840 Frames=1000 Params=5 speed: 8952.13 executions/seconds (ran 25000 potentials in 2.79s) du_dp=False, du_dx=False, u=True
NonbondedInteractionGroup_f64: N=8840 Frames=1000 Params=5 speed: 1532.75 executions/seconds (ran 25000 potentials in 16.31s) du_dp=False, du_dx=False, u=True
Nonbonded_f32: N=8840 Frames=1000 Params=5 speed: 5330.28 executions/seconds (ran 25000 potentials in 4.69s) du_dp=False, du_dx=False, u=True
Nonbonded_f64: N=8840 Frames=1000 Params=5 speed: 222.66 executions/seconds (ran 25000 potentials in 112.28s) du_dp=False, du_dx=False, u=True
HarmonicBond_f32: N=8840 Frames=1000 Params=5 speed: 16255.94 executions/seconds (ran 25000 potentials in 1.54s) du_dp=False, du_dx=False, u=True
HarmonicBond_f64: N=8840 Frames=1000 Params=5 speed: 16145.25 executions/seconds (ran 25000 potentials in 1.55s) du_dp=False, du_dx=False, u=True
HarmonicAngleStable_f32: N=8840 Frames=1000 Params=5 speed: 16120.73 executions/seconds (ran 25000 potentials in 1.55s) du_dp=False, du_dx=False, u=True
HarmonicAngleStable_f64: N=8840 Frames=1000 Params=5 speed: 15446.97 executions/seconds (ran 25000 potentials in 1.62s) du_dp=False, du_dx=False, u=True
PeriodicTorsion_f32: N=8840 Frames=1000 Params=5 speed: 16003.03 executions/seconds (ran 25000 potentials in 1.56s) du_dp=False, du_dx=False, u=True
PeriodicTorsion_f64: N=8840 Frames=1000 Params=5 speed: 14335.13 executions/seconds (ran 25000 potentials in 1.74s) du_dp=False, du_dx=False, u=True
NonbondedPairListPrecomputed_f32: N=8840 Frames=1000 Params=5 speed: 13755.53 executions/seconds (ran 25000 potentials in 1.82s) du_dp=False, du_dx=False, u=True
NonbondedPairListPrecomputed_f64: N=8840 Frames=1000 Params=5 speed: 12774.41 executions/seconds (ran 25000 potentials in 1.96s) du_dp=False, du_dx=False, u=True

# Changes
dhfr-apo: N=23558 speed: 694.33ns/day dt: 2.5fs (ran 100000 steps in 31.11s)
dhfr-apo-barostat-interval-25: N=23558 speed: 287.39ns/day dt: 2.5fs (ran 100000 steps in 75.16s)

hif2a-apo: N=8805 speed: 1179.47ns/day dt: 2.5fs (ran 100000 steps in 18.32s)
hif2a-apo-barostat-interval-25: N=8805 speed: 633.94ns/day dt: 2.5fs (ran 100000 steps in 34.08s)
hif2a-rbfe: N=8840 speed: 890.39ns/day dt: 2.5fs (ran 100000 steps in 24.26s)
hif2a-rbfe-local: N=8840 speed: 1322.05ns/day dt: 2.5fs (ran 100000 steps in 16.34s)
solvent-apo: N=6282 speed: 1425.93ns/day dt: 2.5fs (ran 100000 steps in 15.15s)
solvent-apo-barostat-interval-25: N=6282 speed: 1036.45ns/day dt: 2.5fs (ran 100000 steps in 20.84s)
solvent-rbfe: N=6317 speed: 1167.07ns/day dt: 2.5fs (ran 100000 steps in 18.51s)
solvent-rbfe-local: N=6317 speed: 1420.29ns/day dt: 2.5fs (ran 100000 steps in 15.21s)

NonbondedInteractionGroup_f32: N=8840 Frames=1000 Params=5 speed: 991.61 executions/seconds (ran 25000 potentials in 25.21s) du_dp=True, du_dx=True, u=True
NonbondedInteractionGroup_f64: N=8840 Frames=1000 Params=5 speed: 563.09 executions/seconds (ran 25000 potentials in 44.40s) du_dp=True, du_dx=True, u=True
Nonbonded_f32: N=8840 Frames=1000 Params=5 speed: 370.11 executions/seconds (ran 25000 potentials in 67.55s) du_dp=True, du_dx=True, u=True
Nonbonded_f64: N=8840 Frames=1000 Params=5 speed: 111.94 executions/seconds (ran 25000 potentials in 223.33s) du_dp=True, du_dx=True, u=True
HarmonicBond_f32: N=8840 Frames=1000 Params=5 speed: 780.91 executions/seconds (ran 25000 potentials in 32.01s) du_dp=True, du_dx=True, u=True
HarmonicBond_f64: N=8840 Frames=1000 Params=5 speed: 780.16 executions/seconds (ran 25000 potentials in 32.05s) du_dp=True, du_dx=True, u=True
HarmonicAngleStable_f32: N=8840 Frames=1000 Params=5 speed: 828.92 executions/seconds (ran 25000 potentials in 30.16s) du_dp=True, du_dx=True, u=True
HarmonicAngleStable_f64: N=8840 Frames=1000 Params=5 speed: 825.85 executions/seconds (ran 25000 potentials in 30.27s) du_dp=True, du_dx=True, u=True
PeriodicTorsion_f32: N=8840 Frames=1000 Params=5 speed: 845.41 executions/seconds (ran 25000 potentials in 29.57s) du_dp=True, du_dx=True, u=True
PeriodicTorsion_f64: N=8840 Frames=1000 Params=5 speed: 836.97 executions/seconds (ran 25000 potentials in 29.87s) du_dp=True, du_dx=True, u=True
NonbondedPairListPrecomputed_f32: N=8840 Frames=1000 Params=5 speed: 1614.46 executions/seconds (ran 25000 potentials in 15.49s) du_dp=True, du_dx=True, u=True
NonbondedPairListPrecomputed_f64: N=8840 Frames=1000 Params=5 speed: 1594.94 executions/seconds (ran 25000 potentials in 15.68s) du_dp=True, du_dx=True, u=True

NonbondedInteractionGroup_f32: N=8840 Frames=1000 Params=5 speed: 3585.82 executions/seconds (ran 25000 potentials in 6.97s) du_dp=False, du_dx=False, u=True
NonbondedInteractionGroup_f64: N=8840 Frames=1000 Params=5 speed: 1250.11 executions/seconds (ran 25000 potentials in 20.00s) du_dp=False, du_dx=False, u=True
Nonbonded_f32: N=8840 Frames=1000 Params=5 speed: 539.16 executions/seconds (ran 25000 potentials in 46.37s) du_dp=False, du_dx=False, u=True
Nonbonded_f64: N=8840 Frames=1000 Params=5 speed: 177.22 executions/seconds (ran 25000 potentials in 141.07s) du_dp=False, du_dx=False, u=True
HarmonicBond_f32: N=8840 Frames=1000 Params=5 speed: 1468.25 executions/seconds (ran 25000 potentials in 17.03s) du_dp=False, du_dx=False, u=True
HarmonicBond_f64: N=8840 Frames=1000 Params=5 speed: 1466.10 executions/seconds (ran 25000 potentials in 17.05s) du_dp=False, du_dx=False, u=True
HarmonicAngleStable_f32: N=8840 Frames=1000 Params=5 speed: 1702.76 executions/seconds (ran 25000 potentials in 14.68s) du_dp=False, du_dx=False, u=True
HarmonicAngleStable_f64: N=8840 Frames=1000 Params=5 speed: 1694.07 executions/seconds (ran 25000 potentials in 14.76s) du_dp=False, du_dx=False, u=True
PeriodicTorsion_f32: N=8840 Frames=1000 Params=5 speed: 1765.49 executions/seconds (ran 25000 potentials in 14.16s) du_dp=False, du_dx=False, u=True
PeriodicTorsion_f64: N=8840 Frames=1000 Params=5 speed: 1743.09 executions/seconds (ran 25000 potentials in 14.34s) du_dp=False, du_dx=False, u=True
NonbondedPairListPrecomputed_f32: N=8840 Frames=1000 Params=5 speed: 13943.34 executions/seconds (ran 25000 potentials in 1.79s) du_dp=False, du_dx=False, u=True
NonbondedPairListPrecomputed_f64: N=8840 Frames=1000 Params=5 speed: 13069.05 executions/seconds (ran 25000 potentials in 1.91s) du_dp=False, du_dx=False, u=True

```

proteneer:
before: dhfr-apo-barostat-interval-25: N=23558 speed: 634.76ns/day dt: 2.5fs (ran 100000 steps in 34.03s)
after: dhfr-apo-barostat-interval-25: N=23558 speed: 287.39ns/day dt: 2.5fs (ran 100000 steps in 75.16s)

This seems crazy - 50% slow down with our default barostat settings. Let's see if there's any low hanging fruits can do to improve perf before merging this.

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -427,16 +417,26 @@ void __global__ k_nonbonded_unified(
...
-                u_buffer,
-                u_overflow_count);
+                du_dp);
         };
         tile_idx += stride;
     }
+    if (COMPUTE_U) {
+        // Sync to ensure the shared buffers are populated
+        __syncthreads();
+        if (threadIdx.x == 0) {
```
proteneer, line 434: I suspect this line might be the cause of the slowdown. Let's talk on Monday to see if we can speed this up.

timemachine/cpp/src/kernels/k_fixed_point.cuh
```diff
@@ -67,3 +67,30 @@ template <typename RealType> unsigned long long __device__ __forceinline__ FLOAT
...
+* the energy is beyond the bounds and only exclusion cancellations (only done between NonbondedAllPairs/IxnGroups and NonbondedPairList<..., true>) do not trigger invalid energies.
+* In the case where all interactions are within the bounds, we can still overflows due to summation which int128 allows us to detect
+*
+* Example of Summation overflow
+* -----------------------------
+* accumulated_energy = sum([LLONG_MAX - 1, LLONG_MAX - 1])
+* (__int128)accumulated_energy > LLONG_MAX  - Correctly detects that energies are beyond valid range
+* (long long)accumulated_energy > LLONG_MAX - Overflows and results in seemingly valid energies
+*/
+template <typename RealType> __int128 __device__ __forceinline__ FLOAT_TO_FIXED_ENERGY(RealType u_orig) {
```
proteneer, line 86: Let's also try to optimize this a bit - chat offline for ideas.

timemachine/cpp/src/kernels/k_fixed_point.cuh
```diff
@@ -67,3 +67,30 @@ template <typename RealType> unsigned long long __device__ __forceinline__ FLOAT
...
+    RealType u = u_orig * FIXED_EXPONENT;
+    // All clashes (beyond representation of long long) are treated as LLONG_MAX, to avoid clashes of different signs but non-identical values
+    // cancelling out.
+    if (!isfinite(u) || static_cast<__int128>(u) >= static_cast<__int128>(LLONG_MAX) ||
+        static_cast<__int128>(u) <= static_cast<__int128>(LLONG_MIN)) {
```
proteneer, line 91: I hope that some of these are compile time constants

timemachine/cpp/src/kernels/k_fixed_point.cuh
```diff
@@ -67,3 +67,30 @@ template <typename RealType> unsigned long long __device__ __forceinline__ FLOAT
...
+        return static_cast<__int128>(LLONG_MAX);
+    } else {
+        return static_cast<__int128>(real_to_int64(u));
```
proteneer, line 94: Maybe we can optimize the 64bit -> 128bit int conversion

proteneer:
Benchmarks on my 4080:

```
before: dhfr-apo-barostat-interval-25: N=23558 speed: 424.64ns/day dt: 2.5fs (ran 10000 steps in 5.09s)
after: dhfr-apo-barostat-interval-25: N=23558 speed: 304.22ns/day dt: 2.5fs (ran 10000 steps in 7.11s)
```

Still a pretty significant drop in performance

timemachine/cpp/src/kernels/k_fixed_point.cuh
proteneer: Also, can we try and do the comparison in floating point (using `nextafter` on LLONG_MAX/LLONG_MIN, generated at compile time, to avoid the extra casts to `__int128`)

timemachine/cpp/src/gpu_utils.cu
```diff
@@ -10,3 +10,24 @@ curandStatus_t
...
 }
+
+void __global__ k_accumulate_energy(
+    int N,
+    const __int128 *__restrict__ input_buffer, // [N]
+    __int128 *__restrict u_buffer              // [1]
+) {
+    int idx = blockIdx.x * blockDim.x + threadIdx.x;
+    // Should only have 1 idx, could have a warp that does a parallel reduce in the future
+    if (idx >= 1) {
```
proteneer, line 21: doesn't `cub::reduce` support `__int128`?

timemachine/cpp/src/kernels/k_nonbonded.cuh
badisa: As of commit https://github.com/proteneer/timemachine/pull/1090/commits/db125b53b086bc8bf14ea337e997a391f8db1d69

```
dhfr-apo: N=23558 speed: 694.65ns/day dt: 2.5fs (ran 100000 steps in 31.10s)
dhfr-apo-barostat-interval-25: N=23558 speed: 622.25ns/day dt: 2.5fs (ran 100000 steps in 34.72s)
building a protein system with 1758 protein atoms and 7047 water atoms
hif2a-apo: N=8805 speed: 1179.01ns/day dt: 2.5fs (ran 100000 steps in 18.32s)
hif2a-apo-barostat-interval-25: N=8805 speed: 1002.58ns/day dt: 2.5fs (ran 100000 steps in 21.55s)
hif2a-rbfe: N=8840 speed: 892.35ns/day dt: 2.5fs (ran 100000 steps in 24.21s)
hif2a-rbfe-local: N=8840 speed: 1313.10ns/day dt: 2.5fs (ran 100000 steps in 16.45s)
solvent-apo: N=6282 speed: 1452.43ns/day dt: 2.5fs (ran 100000 steps in 14.87s)
solvent-apo-barostat-interval-25: N=6282 speed: 1261.00ns/day dt: 2.5fs (ran 100000 steps in 17.13s)
solvent-rbfe: N=6317 speed: 1175.46ns/day dt: 2.5fs (ran 100000 steps in 18.38s)
solvent-rbfe-local: N=6317 speed: 1408.38ns/day dt: 2.5fs (ran 100000 steps in 15.34s)
```

timemachine/cpp/src/gpu_utils.cu
badisa: It does, I had avoided cub initially as had imagined the cost of the setup of CUB would be larger than the potential speed up. While I was wrong about that, the parallel reduction that was added gets us within 1% of the previous benchmark. I think cub is still overkill here as it does inter-block reductions while we should be able to get away with block reductions.

The changes that I would like to follow up on after this PR:

- Make the amount of memory allocated for energy buffers constant relative to the system size. This will amortize more of the accumulation in the kernels that generate the energies as well as reduce memory
- Improve the barostat code, its janky as I didn't know what I was doing and doing more work than I think it needs to.

tests/common.py
proteneer: +1 (later pass)

tests/common.py
```diff
@@ -41,6 +42,12 @@ def get_110_ccc_ff():
     return forcefield
 
 
+def fixed_overflowed(a):
+    """Refer to timemachine/cpp/src/kernels/k_fixed_point.hpp::FLOAT_TO_FIXED_ENERGY for documentation onOne how we handle energies and overflows"""
```
proteneer, line 46: typo `onOne`

tests/common.py
```diff
@@ -41,6 +42,12 @@ def get_110_ccc_ff():
...
+    converted_a = np.int64(np.uint64(a))
```
proteneer, line 47: We should make this function a bit more precise.

Doesn't this always generate a number in the closed interval `[np.iinfo(np.int64).min, np.iinfo(np.int64).max]`?  I.e. I don't think it's possible to have `<=`, or `>=` simply testing for equality should suffice:

`converted_a == np.iinfo(np.int64).min or converted_a == np.iinfo(np.int64).max`

Alternatively, if `u_fixed` is a `int128`, then there shouldn't be a need to cast

`a <= np.iinfo(np.int64).min or a >= np.iinfo(np.int64).max`


Another nit:

When do we actually trigger the the `min` case, I thought we always converted to `MAX_INT64` in the case of `nans/-inf/+inf` etc. Should we just throw an assertion if we did hit `MIN_INT64`?

tests/test_energy_overflows.py
```diff
@@ -227,28 +227,27 @@ def compute_potential_energy(pot):
...
+    fixed_all_pairs_energy = all_pairs_bound.execute_fixed(x0, box)
+    fixed_pair_list_energy = pair_list_bound.execute_fixed(x0, box)
+    assert fixed_overflowed(fixed_all_pairs_energy)
+    assert fixed_overflowed(fixed_pair_list_energy)
 
-    # The sum of the fixed point energies should be bitwise identical to the test energy
-    assert FIXED_TO_FLOAT(fixed_all_pairs_energy + fixed_pair_list_energy) == test_energy
+    # The values are identical because of overflow, as its not clear how to go from
+    # C++ __int128 to python integer.
+    assert fixed_pair_list_energy == fixed_pair_list_energy
```
proteneer, line 248: Isn't this just testing to see if `fixed_pair_list_energy` is a `nan`? maybe just replace with `isnan(fixed_pair_list_energy)`

timemachine/cpp/src/gpu_utils.cuh
```diff
@@ -83,3 +84,81 @@ double __device__ __forceinline__ rmul_rn(double a, double b) { return __dmul_rn
...
+    shared_mem[tid] = 0;
+    while (i < N) {
+        // Only add two values if the input buffer has enough input
+        shared_mem[tid] += input_buffer[i] + (i + BLOCK_SIZE < N ? input_buffer[i + BLOCK_SIZE] : 0);
+        i += stride;
+    }
+    //
+    __syncthreads();
+
+    // For larger blocks larger than a warp
```
proteneer, line 126: 1) Since `BLOCK_SIZE` is a compile time constant, can we `static_assert` that `BLOCK SIZE <= 1024?`
2) Does this code break if `BLOCK_SIZE` is not a power of 2?

timemachine/cpp/src/gpu_utils.cuh
```diff
@@ -83,3 +84,81 @@ double __device__ __forceinline__ rmul_rn(double a, double b) { return __dmul_rn
...
+    if (BLOCK_SIZE >= 512) {
+        if (tid < 256) {
+            shared_mem[tid] += shared_mem[tid + 256];
```
proteneer, line 129: does `+=` here work with a single `__syncthreads`?  Wasn't sure if we had to separate the reads from the write with an extra `__syncthreads()` in between.

timemachine/cpp/src/gpu_utils.cuh
```diff
@@ -83,3 +84,81 @@ double __device__ __forceinline__ rmul_rn(double a, double b) { return __dmul_rn
...
+
+// Taken originally from https://developer.download.nvidia.com/assets/cuda/files/reduction.pdf
+// but modified to account for post Cuda 9.0 which no longer implicitly syncs warps and the
+// fact that we don't need to launch recursive blocks as we can simply reduce the number of energies
+// that are accumulated.
+template <unsigned int BLOCK_SIZE>
+void __global__ k_accumulate_energy(
+    int N,
+    const __int128 *__restrict__ input_buffer, // [N]
+    __int128 *__restrict u_buffer              // [1]
```
proteneer, line 107: `__restrict` or  `__restrict__` ?

timemachine/cpp/src/kernels/k_fixed_point.cuh
```diff
@@ -67,3 +67,30 @@ template <typename RealType> unsigned long long __device__ __forceinline__ FLOAT
 template <typename RealType> unsigned long long __device__ __forceinline__ FLOAT_TO_FIXED_BONDED(RealType v) {
     return static_cast<unsigned long long>(real_to_int64(v * FIXED_EXPONENT));
 }
+
+/* FLOAT_TO_FIXED_ENERGY converts floating point energies into fixed point. For infinite values or values beyond LLONG_MAX/LLONG_MIN
+* the value will be capped. This is accumulated into __int128 which handles the over and underflows, allowing us to be able to account
```
proteneer, line 72: explicitly state `capped to LLONG_MAX/LLONG_MIN`

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -279,9 +275,7 @@ void __device__ v_nonbonded_unified(
 
             if (COMPUTE_U) {
                 // If the energy is overflowed, don't bother adding it to the buffer
```
proteneer, line 277: Is this comment still relevant?

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -427,16 +417,26 @@ void __global__ k_nonbonded_unified(
...
+                du_dp);
         };
         tile_idx += stride;
     }
+    if (COMPUTE_U) {
+        // Sync to ensure the shared buffers are populated
+        __syncthreads();
+        if (threadIdx.x == 0) {
+            __int128 accumulated_energy = 0;
+            for (int i = 0; i < THREADS; i++) {
```
proteneer, line 436: We should definitely parallelize this as well using the same strategy.

timemachine/cpp/src/kernels/k_nonbonded_pair_list.cuh
```diff
@@ -119,10 +118,10 @@ void __global__ k_nonbonded_pair_list(
     RealType delta_w = w_i - w_j;
     RealType d2ij = delta_x * delta_x + delta_y * delta_y + delta_z * delta_z + delta_w * delta_w;
 
+    RealType u = static_cast<RealType>(0.0);
     // see note: this must be strictly less than
     if (d2ij < cutoff_squared) {
 
-        RealType u;
```
proteneer, line 125: I was curious about this change, did something break that require us to do explicit initialization? Before this was uninitialized, because:

1) `compute_electrostatics` runs `u = charge_scale * qij * inv_dij * ebd;` (initialization)
2) `compute_lennard_jones` runs `u += lj_scale * 4 * eps_ij * (sig6_inv_d6ij - 1) * sig6_inv_d6ij;`

timemachine/cpp/src/nonbonded_all_pairs.cu
```diff
@@ -31,18 +31,18 @@ NonbondedAllPairs<RealType>::NonbondedAllPairs(
       d_sort_storage_bytes_(0), disable_hilbert_(disable_hilbert_sort),
 
       kernel_ptrs_({// enumerate over every possible kernel combination
+                    // Set threads to 1 if not computing energy to reduced unused shared memory
```
proteneer, line 34: Can we set this to zero?

timemachine/cpp/src/nonbonded_all_pairs.cu
proteneer: Good call

proteneer (changes requested):

timemachine/cpp/src/gpu_utils.cuh
```diff
@@ -83,3 +84,81 @@ double __device__ __forceinline__ rmul_rn(double a, double b) { return __dmul_rn
...
+        __syncwarp();
+    }
+}
+
+// Taken originally from https://developer.download.nvidia.com/assets/cuda/files/reduction.pdf
+// but modified to account for post Cuda 9.0 which no longer implicitly syncs warps and the
+// fact that we don't need to launch recursive blocks as we can simply reduce the number of energies
+// that are accumulated.
+template <unsigned int BLOCK_SIZE>
+void __global__ k_accumulate_energy(
```
proteneer, line 104: Can we add a unit test just for this kernel? (Either in C++ or Python)

tests/test_energy_overflows.py
badisa: This is wrong because it is comparing against itself, but it supposed to check that `fixed_pair_list_energy == fixed_all_pair_energy` because this is in fixed point and should be both represented by a LLONG_MAX

timemachine/cpp/src/gpu_utils.cuh
badisa: Yes, because of the `tid < 256` there is no risk that the threads being read in from will change.

timemachine/cpp/src/kernels/k_nonbonded_pair_list.cuh
badisa: Because we always set the values in the energy buffer, need to initialize it to zero.

timemachine/cpp/src/nonbonded_all_pairs.cu
badisa: Sadly we can not

tests/test_energy_overflows.py
```diff
@@ -272,13 +271,8 @@ def test_energy_overflows_with_summation_of_energies(precision):
     )
     nonbonded_gpu = nonbonded.to_gpu(precision)
 
-    # TBD: Remove the try/finally to clean up the GPU potential, else calling reset_device
-    # will fail
-    try:
-        # The reference should return a large energy value
-        assert nonbonded(x, nb_params, box) > 0.0
+    # The reference should return a large energy value
+    assert nonbonded(x, nb_params, box) > 0.0
```
mcwitt, line 275: Can this assertion be strengthened? (comment says "large", but only tests "positive")

badisa: Strengthened in https://github.com/proteneer/timemachine/pull/1090/commits/13988bcd6633fa0d7391a657052b3f1effc7e114

timemachine/cpp/src/gpu_utils.cuh
badisa: Working on this, but a bit more involved to expose the kernels to the warp_kernels due to where the logic is currently and what is kernel only code (float -> fixed) and such. 

Want to investigate gtest at some point, but would be best in a separate pass

timemachine/cpp/src/nonbonded_all_pairs.cu
badisa: Offline discussion, this gets removed by the compiler if its unused, so made consistent for visual purposes.

timemachine/cpp/src/kernels/k_nonbonded.cuh
badisa: Refactored to use the same logic here: https://github.com/proteneer/timemachine/pull/1090/commits/f335db89264349c37d13f05af03223c5b033bef4

timemachine/cpp/src/gpu_utils.cuh
badisa: Fixed in 78c4f4b52b7ab817f8a0bdadaaf8bb064f3d6e46 to be `__restrict__` to be consistent

timemachine/cpp/src/kernels/k_fixed_point.cuh
badisa: Made more explicit in 742757945f3dc64bf5a38879e908507e86bad7a7

tests/test_energy_overflows.py
badisa: Fixed the assertion here b6e7dc54317b3b0e60ad13417876516c8af5aae5

tests/common.py
badisa: Changed to check equality here 080e54a60c1eefc46080c7c1b0cfa8bd1ea8c2e2

Still need to add assertion

timemachine/cpp/src/gpu_utils.cuh
badisa: Added static assertions. You are correct, need to be powers of two. BLOCK_SIZE=512 seems to be the best, but added support for 1024 in f335db89264349c37d13f05af03223c5b033bef4

tests/common.py
badisa: Fixed typo in 080e54a60c1eefc46080c7c1b0cfa8bd1ea8c2e2

timemachine/cpp/src/kernels/k_nonbonded.cuh
badisa: Removed comment in 894b0b62bcf99a2891b80a4414e5150e0f67a5bc

tests/common.py
badisa: Added assertion for MIN_INT64 in https://github.com/proteneer/timemachine/pull/1090/commits/0665bf77d57027ae164519872759648835dbbc14

timemachine/cpp/src/gpu_utils.cu
proteneer: sgtm!

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -869,18 +856,21 @@ void declare_bound_potential(py::module &m) {
             "execute_fixed",
             [](timemachine::BoundPotential &bp,
                const py::array_t<double, py::array::c_style> &coords,
-               const py::array_t<double, py::array::c_style> &box) -> py::tuple {
+               const py::array_t<double, py::array::c_style> &box) -> const py::array_t<uint64_t, py::array::c_style> {
```
proteneer, line 859: I can't remember why we chose for this to return a numpy array of size 1 as opposed a scalar - might be good to standardize on this later.

timemachine/cpp/src/gpu_utils.cuh
proteneer: I do think we need to test some of this logic in isolation (like how we do with the neighborlist code, even though the NB code uses them):

1) so we can sleep better at night
2) so we can iterate quickly on it when we add support for more specialized shuffles later on

I think we can just add a wrapper function and expose it to `wrap_kernels.cpp` , similar to:

https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/wrap_kernels.cpp#L1252

timemachine/cpp/src/wrap_kernels.cpp
badisa: I think the reason for this, as I tried to just return `uint64` is that it returns a python int, which seemed to break the conversion. This way it is a numpy array with a fixed type that is maintained through the wrapping layer.

timemachine/cpp/src/gpu_utils.cuh
badisa: Added test in https://github.com/proteneer/timemachine/pull/1090/commits/adf5700639c9631f7e7d865af2782e6732b95266

timemachine/cpp/src/wrap_kernels.cpp
proteneer: Ah good call, thanks.

tests/common.py
```diff
@@ -41,6 +42,13 @@ def get_110_ccc_ff():
     return forcefield
 
 
+def fixed_overflowed(a):
+    """Refer to timemachine/cpp/src/kernels/k_fixed_point.hpp::FLOAT_TO_FIXED_ENERGY for documentation on how we handle energies and overflows"""
```
maxentile, line 46:
```suggestion
    """Refer to timemachine/cpp/src/kernels/k_fixed_point.cuh::FLOAT_TO_FIXED_ENERGY for documentation on how we handle energies and overflows"""
```

tests/test_energy_overflows.py
```diff
@@ -272,13 +272,22 @@ def test_energy_overflows_with_summation_of_energies(precision):
...
+
+
+def test_energy_accumulation():
+    """Test the the logic used to accumulate energy in int128.
+
+    It relies on doing a block level parallel reduce for performance.
+    """
+    rng = np.random.default_rng(2023)
+
+    vals = rng.integers(-10000, 10000, size=10000, dtype=np.int64)
```
proteneer, line 289: Might be good to test some different values of size as well?  (eg. 0, <32/warp, < threads per block, etc.)

tests/test_md.py
```diff
@@ -789,7 +789,7 @@ def test_setup_context_with_references():
     seed = 2022
     pressure = constants.DEFAULT_PRESSURE
 
-    unbound_potentials, sys_params, masses, coords, box = get_solvent_phase_system(mol, ff, 0.0, minimize_energy=False)
+    unbound_potentials, sys_params, masses, coords, box = get_solvent_phase_system(mol, ff, 1.0, minimize_energy=False)
```
maxentile, line 792: nit: surprising diff -- is this related to the main changes?

timemachine/cpp/src/gpu_utils.cuh
```diff
@@ -83,3 +84,85 @@ double __device__ __forceinline__ rmul_rn(double a, double b) { return __dmul_rn
...
+        i += stride;
+    }
+    //
+    __syncthreads();
+
+    block_energy_reduce<BLOCK_SIZE>(shared_mem, threadIdx.x);
+    if (tid == 0) {
+        // This could be a race condition if multiple `k_accumulate_energy` are running
+        // on the same u_buffer
+        u_buffer[0] = shared_mem[0];
```
maxentile, line 166: Is this comment actionable?

timemachine/cpp/src/kernels/k_fixed_point.cuh
```diff
@@ -67,3 +67,30 @@ template <typename RealType> unsigned long long __device__ __forceinline__ FLOAT
 template <typename RealType> unsigned long long __device__ __forceinline__ FLOAT_TO_FIXED_BONDED(RealType v) {
     return static_cast<unsigned long long>(real_to_int64(v * FIXED_EXPONENT));
 }
+
+/* FLOAT_TO_FIXED_ENERGY converts floating point energies into fixed point. Values beyond LLONG_MAX/LLONG_MIN
+* (or nonfinite) will be capped to LLONG_MAX. This is accumulated into __int128 which handles the over and underflows,
```
maxentile, line 72: nit: does this detect underflow?

(my understanding is that it detects positive or negative overflows)

timemachine/cpp/src/kernels/k_fixed_point.cuh
```diff
@@ -67,3 +67,30 @@ template <typename RealType> unsigned long long __device__ __forceinline__ FLOAT
...
+* allowing us to be able to account or overflows triggered by the summation of energies.
```
maxentile, line 73:
```suggestion
* allowing us to be able to account for overflows triggered by the summation of energies.
```

maxentile (commented):
Nice!

Among many other improvements, this PR reverts the overflow counting patch added in https://github.com/proteneer/timemachine/pull/1084 , and instead uses a superaccumulator (doubling the bit length from 64 to 128). This sounds like a simpler and safer approach to avoiding undetected energy overflows https://github.com/proteneer/timemachine/issues/481 , which could previously arise during fixed-point summation.

(As a consequence of relying on bitwise cancellation to handle exclusions https://github.com/proteneer/timemachine/pull/273 , other approaches to detecting overflow during fixed-point summation (such as https://github.com/rapidsai/cudf/blob/ad64c66add770ad1d69419b65b301fe97980174e/cpp/include/cudf/fixed_point/fixed_point.hpp#L635-L649 ) cannot be directly applied.)

timemachine/cpp/src/kernels/k_fixed_point.cuh
badisa: Will fix to say 'negative overflow'

tests/test_md.py
badisa: Yes, it's because having the ligand at lambda 0.0 results in a clash which breaks the test while previously the barostat wouldn't see a clash and accept bad moves

maxentile: Woah -- thanks for clarifying!

timemachine/cpp/src/barostat.hpp
```diff
@@ -56,13 +56,11 @@ class MonteCarloBarostat {
     int *d_num_attempted_;
     int *d_num_accepted_;
 
-    unsigned long long *d_u_buffer_;
-    unsigned long long *d_u_after_buffer_;
+    __int128 *d_u_buffer_;
+    __int128 *d_u_after_buffer_;
```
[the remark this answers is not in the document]
badisa: That would make sense, as I know @proteneer and I had talked about 96 bit instead of 128.

If we want to do this, I would prefer doing it in a separate pass as this has gotten pretty large already.

timemachine/cpp/src/gpu_utils.cuh
badisa: I believe so? You could easily write the code in a way that would result in different potentials clobbering the energy buffers

timemachine/cpp/src/kernels/k_fixed_point.cuh
badisa: Fixed in [6da81a0](https://github.com/proteneer/timemachine/pull/1090/commits/6da81a094c6812b11b64a39b78e8e3726948074e)

tests/test_energy_overflows.py
badisa: Added additional tests in [d59a706](https://github.com/proteneer/timemachine/pull/1090/commits/d59a70604e18418c9875dda5e1fb27250f196f60)

timemachine/cpp/src/gpu_utils.cuh
maxentile: >You could easily write the code in a way that would result in different potentials clobbering the energy buffers

Sorry, I don't have a good sense of context or severity.

(Can this already happen (even if rarely or non-deterministically) for any of the potential configurations in the tests? Or is this just a possible future problem, which could only occur if, say, the implementation of `StreamedPotentialRunner` is modified unsafely? If a current problem, would atomically incrementing `u_buffer[0]` avoid it?)

tests/common.py
```diff
@@ -19,6 +19,7 @@
 from timemachine.fe.utils import read_sdf
 from timemachine.ff import Forcefield
 from timemachine.ff.handlers import openmm_deserializer
+from timemachine.integrator import FIXED_EXPONENT
```
mcwitt, line 22: nit/deboggle: surprising that this is defined in the integrator module

tests/common.py
```diff
@@ -41,6 +42,13 @@ def get_110_ccc_ff():
     return forcefield
 
 
+def fixed_overflowed(a):
+    """Refer to timemachine/cpp/src/kernels/k_fixed_point.cuh::FLOAT_TO_FIXED_ENERGY for documentation on how we handle energies and overflows"""
+    converted_a = np.int64(np.uint64(a))
+    assert converted_a != np.iinfo(np.int64).min, "Unexpected value for fixed point energy"
```
mcwitt, line 48: Worth expanding the message or adding a comment here? (i.e. why can negative overflow not happen?)

badisa: I think the documentation for `FLOAT_TO_FIXED_ENERGY` should cover this. This is just an indicator that the expectations are off.

timemachine/cpp/src/gpu_utils.cuh
```diff
@@ -83,3 +84,85 @@ double __device__ __forceinline__ rmul_rn(double a, double b) { return __dmul_rn
...
+    for (int i = 32; i > 0; i /= 2) {
+        if (thread_idx < i) {
+            shared_data[thread_idx] = shared_data[thread_idx] + shared_data[thread_idx + i];
+        }
+        __syncwarp();
+    }
+}
+
+template <unsigned int THREADS_PER_BLOCK>
+__device__ __forceinline__ void block_energy_reduce(volatile __int128 *shared_data, int tid) {
```
mcwitt, line 100: Worth adding a comment here? (many "magic numbers", and not clear to me what's going on here)

timemachine/cpp/src/centroid_restraint.hpp
```diff
@@ -14,6 +14,8 @@ template <typename RealType> class CentroidRestraint : public Potential {
     unsigned long long *d_centroid_a_;
     unsigned long long *d_centroid_b_;
 
+    __int128 *d_u_buffer_;
```
mcwitt, line 17: nit/deboggle: it looks like `d_u_buffer_` is added as a member variable to most/all of the `Potential` subclasses? Would it make sense to declare in `Potential`?

timemachine/cpp/src/kernels/k_flat_bottom_bond.cuh
```diff
@@ -110,11 +110,8 @@ void __global__ k_flat_bottom_bond(
     if (u) {
         RealType u_real = compute_flat_bottom_energy<RealType>(k, r, rmin, rmax);
 
-        // cast float -> fixed
-        auto sum_u = FLOAT_TO_FIXED_BONDED<RealType>(u_real);
-
-        // atomic add to u array
-        atomicAdd(u + src_idx, sum_u);
+        // Set idx to force
```
mcwitt, line 113: nit: comment not clear to me. What is "force" here?

timemachine/cpp/src/kernels/k_chiral_restraint.cuh
```diff
@@ -50,6 +50,11 @@ void __global__ k_chiral_atom_restraint(
     pyramidal_vol_and_grad(xc, x1, x2, x3, vol, xc_grad, x1_grad, x2_grad, x3_grad);
     RealType k_restr = params[r_idx];
 
+    // Always set the energies to avoid getting meaningless values
```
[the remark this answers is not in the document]
badisa: Clarified in https://github.com/proteneer/timemachine/pull/1090/commits/47c11b5f2eec237eb82c01300b56b8a8ff24feda

timemachine/cpp/src/kernels/k_flat_bottom_bond.cuh
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1090/commits/f9e29f743413a8d51736473e77c0fc2395d585d9

timemachine/cpp/src/gpu_utils.cuh
badisa: Added docstring here https://github.com/proteneer/timemachine/pull/1090/commits/b0dc781532902ab665cf012727add165fb56fd1f

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -121,18 +122,17 @@ void __device__ v_nonbonded_unified(
     const int tile_idx,
     const int N,
     const int NR,
-    const double *__restrict__ coords,
-    const double *__restrict__ params, // [N]
+    const double *__restrict__ coords, // [N * 3]
+    const double *__restrict__ params, // [N * 4]
```
mcwitt, line 126: nit: 
```suggestion
    const double *__restrict__ params, // [N * PARAMS_PER_ATOM]
```,
as below

File: timemachine/cpp/src/kernels/k_nonbonded_pair_list.cuh [hunk 38]
```diff
@@ -193,4 +182,10 @@ void __global__ k_nonbonded_pair_list(
             accumulate<Negated>(du_dp + w_param_idx_j, g_wj);
         }
     }
+    // Always accumulate into the energy buffers even if there is no interaction to ensure
+    // buffer is zeroed out, avoids having to memset every call
+    if (u_buffer) {
+        // Do not do `FLOAT_TO_FIXED_ENERGY(Negated ? -u : u)` as that can produce a positive fixed point value
+        u_buffer[pair_idx] = Negated ? -FLOAT_TO_FIXED_ENERGY<RealType>(u) : FLOAT_TO_FIXED_ENERGY<RealType>(u);
+    }
```
mcwitt, line 190: track for future PR: consider renaming `NonbondedPairListNegated` in the Python API to `NonbondedExclusions` or similar, to emphasize that it is specialized for exclusions and not valid on its own

timemachine/cpp/src/gpu_utils.cuh
```diff
@@ -83,3 +84,85 @@ double __device__ __forceinline__ rmul_rn(double a, double b) { return __dmul_rn
 float __device__ __forceinline__ radd_rn(float a, float b) { return __fadd_rn(a, b); }
 
 double __device__ __forceinline__ radd_rn(double a, double b) { return __dadd_rn(a, b); }
+
+// If this were not int128s, we could do __shfl_down_sync, but only supports up to 64 values
+__device__ __forceinline__ void energy_warp_reduce(volatile __int128 *shared_data, int thread_idx) {
+#pragma unroll 6
```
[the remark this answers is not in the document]
badisa: Added docstring here reference PDF https://github.com/proteneer/timemachine/pull/1090/commits/b0dc781532902ab665cf012727add165fb56fd1f

mcwitt (approved):
LGTM!

tests/test_md.py
proteneer: wtf really?

badisa: Yes, that is because the minimization is disabled. For the purposes of the test, minimization just slows it down and doesn't impact the correctness. We could also set `minimize_energy=True` and use `lambda=0.0`, at the price of speed.

We could also disable the check that the box volume changed, but that seems useful to test when we are checking if shared pointers are doing what they should.

badisa:
As the benchmarks had been previous run on an A10 (or a datacenter gpu) with better 64 bit performance. The following is benchmarks on a RTX 2080 for posteriority

```
RTX 2080 - Pre Overflow checking (https://github.com/proteneer/timemachine/commit/4e81078f41a22697e78a0f0a4fec5e4ebf1e29d9)

dhfr-apo: N=23558 speed: 559.59ns/day dt: 2.5fs (ran 100000 steps in 38.60s)
dhfr-apo-barostat-interval-25: N=23558 speed: 508.01ns/day dt: 2.5fs (ran 100000 steps in 42.52s)
hif2a-apo: N=8805 speed: 1001.51ns/day dt: 2.5fs (ran 100000 steps in 21.57s)
hif2a-apo-barostat-interval-25: N=8805 speed: 848.86ns/day dt: 2.5fs (ran 100000 steps in 25.45s)
hif2a-rbfe: N=8840 speed: 822.54ns/day dt: 2.5fs (ran 100000 steps in 26.26s)
hif2a-rbfe-local: N=8840 speed: 1332.04ns/day dt: 2.5fs (ran 100000 steps in 16.22s)
solvent-apo: N=6282 speed: 1295.52ns/day dt: 2.5fs (ran 100000 steps in 16.68s)
solvent-apo-barostat-interval-25: N=6282 speed: 1153.82ns/day dt: 2.5fs (ran 100000 steps in 18.72s)
solvent-rbfe: N=6317 speed: 1094.48ns/day dt: 2.5fs (ran 100000 steps in 19.74s)
solvent-rbfe-local: N=6317 speed: 1558.64ns/day dt: 2.5fs (ran 100000 steps in 13.86s)
NonbondedInteractionGroup_f32: N=8840 Frames=1000 Params=5 speed: 720.58 executions/seconds (ran 25000 potentials in 34.69s) du_dp=True, du_dx=True, u=True
NonbondedInteractionGroup_f64: N=8840 Frames=1000 Params=5 speed: 476.94 executions/seconds (ran 25000 potentials in 52.42s) du_dp=True, du_dx=True, u=True
Nonbonded_f32: N=8840 Frames=1000 Params=5 speed: 611.24 executions/seconds (ran 25000 potentials in 40.90s) du_dp=True, du_dx=True, u=True
Nonbonded_f64: N=8840 Frames=1000 Params=5 speed: 113.35 executions/seconds (ran 25000 potentials in 220.56s) du_dp=True, du_dx=True, u=True
HarmonicBond_f32: N=8840 Frames=1000 Params=5 speed: 929.52 executions/seconds (ran 25000 potentials in 26.90s) du_dp=True, du_dx=True, u=True
HarmonicBond_f64: N=8840 Frames=1000 Params=5 speed: 915.65 executions/seconds (ran 25000 potentials in 27.30s) du_dp=True, du_dx=True, u=True
HarmonicAngleStable_f32: N=8840 Frames=1000 Params=5 speed: 899.21 executions/seconds (ran 25000 potentials in 27.80s) du_dp=True, du_dx=True, u=True
HarmonicAngleStable_f64: N=8840 Frames=1000 Params=5 speed: 885.13 executions/seconds (ran 25000 potentials in 28.24s) du_dp=True, du_dx=True, u=True
PeriodicTorsion_f32: N=8840 Frames=1000 Params=5 speed: 894.44 executions/seconds (ran 25000 potentials in 27.95s) du_dp=True, du_dx=True, u=True
PeriodicTorsion_f64: N=8840 Frames=1000 Params=5 speed: 895.81 executions/seconds (ran 25000 potentials in 27.91s) du_dp=True, du_dx=True, u=True
NonbondedPairListPrecomputed_f32: N=8840 Frames=1000 Params=5 speed: 1031.98 executions/seconds (ran 25000 potentials in 24.23s) du_dp=True, du_dx=True, u=True
NonbondedPairListPrecomputed_f64: N=8840 Frames=1000 Params=5 speed: 995.68 executions/seconds (ran 25000 potentials in 25.11s) du_dp=True, du_dx=True, u=True


RTX 2080 - int 128 accumulation

dhfr-apo: N=23558 speed: 557.16ns/day dt: 2.5fs (ran 100000 steps in 38.77s)
dhfr-apo-barostat-interval-25: N=23558 speed: 496.91ns/day dt: 2.5fs (ran 100000 steps in 43.47s)
hif2a-apo: N=8805 speed: 1009.30ns/day dt: 2.5fs (ran 100000 steps in 21.40s)
hif2a-apo-barostat-interval-25: N=8805 speed: 834.15ns/day dt: 2.5fs (ran 100000 steps in 25.90s)
hif2a-rbfe: N=8840 speed: 829.09ns/day dt: 2.5fs (ran 100000 steps in 26.06s)
hif2a-rbfe-local: N=8840 speed: 1337.60ns/day dt: 2.5fs (ran 100000 steps in 16.15s)
solvent-apo: N=6282 speed: 1315.97ns/day dt: 2.5fs (ran 100000 steps in 16.42s)
solvent-apo-barostat-interval-25: N=6282 speed: 1119.20ns/day dt: 2.5fs (ran 100000 steps in 19.30s)
solvent-rbfe: N=6317 speed: 1101.90ns/day dt: 2.5fs (ran 100000 steps in 19.61s)
solvent-rbfe-local: N=6317 speed: 1560.43ns/day dt: 2.5fs (ran 100000 steps in 13.85s)
NonbondedInteractionGroup_f32: N=8840 Frames=1000 Params=5 speed: 746.10 executions/seconds (ran 25000 potentials in 33.51s) du_dp=True, du_dx=True, u=True
NonbondedInteractionGroup_f64: N=8840 Frames=1000 Params=5 speed: 490.57 executions/seconds (ran 25000 potentials in 50.96s) du_dp=True, du_dx=True, u=True
Nonbonded_f32: N=8840 Frames=1000 Params=5 speed: 609.30 executions/seconds (ran 25000 potentials in 41.03s) du_dp=True, du_dx=True, u=True
Nonbonded_f64: N=8840 Frames=1000 Params=5 speed: 114.81 executions/seconds (ran 25000 potentials in 217.75s) du_dp=True, du_dx=True, u=True
HarmonicBond_f32: N=8840 Frames=1000 Params=5 speed: 980.83 executions/seconds (ran 25000 potentials in 25.49s) du_dp=True, du_dx=True, u=True
HarmonicBond_f64: N=8840 Frames=1000 Params=5 speed: 989.89 executions/seconds (ran 25000 potentials in 25.26s) du_dp=True, du_dx=True, u=True
HarmonicAngleStable_f32: N=8840 Frames=1000 Params=5 speed: 941.58 executions/seconds (ran 25000 potentials in 26.55s) du_dp=True, du_dx=True, u=True
HarmonicAngleStable_f64: N=8840 Frames=1000 Params=5 speed: 934.62 executions/seconds (ran 25000 potentials in 26.75s) du_dp=True, du_dx=True, u=True
PeriodicTorsion_f32: N=8840 Frames=1000 Params=5 speed: 937.50 executions/seconds (ran 25000 potentials in 26.67s) du_dp=True, du_dx=True, u=True
PeriodicTorsion_f64: N=8840 Frames=1000 Params=5 speed: 940.97 executions/seconds (ran 25000 potentials in 26.57s) du_dp=True, du_dx=True, u=True
NonbondedPairListPrecomputed_f32: N=8840 Frames=1000 Params=5 speed: 1083.00 executions/seconds (ran 25000 potentials in 23.08s) du_dp=True, du_dx=True, u=True
NonbondedPairListPrecomputed_f64: N=8840 Frames=1000 Params=5 speed: 1050.50 executions/seconds (ran 25000 potentials in 23.80s) du_dp=True, du_dx=True, u=True
```

tests/test_md.py
proteneer: Thanks for the clarification - so the previous barostat test was just broken?

badisa: Exactly, the test was previously broken.

proteneer:
> (As a consequence of relying on bitwise cancellation to handle exclusions https://github.com/proteneer/timemachine/pull/273 , other approaches to detecting overflow during fixed-point summation (such as https://github.com/rapidsai/cudf/blob/ad64c66add770ad1d69419b65b301fe97980174e/cpp/include/cudf/fixed_point/fixed_point.hpp#L635-L649 ) cannot be directly applied.)

If we implemented exclusions correctly (i.e. so we don't have any overflows due to exclusions), we can probably just use float32s for the energies, but we still cannot use atomicAdds on said floats since we need to guarantee determinism (we'd need to still do a parallel reduction). There's other difficulties with implementing exclusions correctly (i.e. splitting them into the appropriate interaction groups) that we haven't looked into, but this is something worth considering in a future pass.

maxentile (approved):
Looks great!

PR #1091, opened 2023-07-20
[wip] Gradual water moves

maxentile:
A couple special cases of NCMC being investigated for application to sampling buried waters.

Motivation: Equilibrate water between buried sites and bulk.

Intuition:
* When proposing to move a water from point a to point b in a crowded simulation box, instantaneous moves may have very low acceptance rates.
* Several mitigations are possible (including brute force), but one class of mitigation is to increase the acceptance rate by making the proposed change happen "gradually" -- i.e. give the water's previous environment and new environment at least a few steps to accommodate the proposed change. There is some design flexibility about how to do this correctly.

A few recent references in this area:
* [Ben-Shalom+, 2019] Simulating water exchange to buried binding sites https://pubmed.ncbi.nlm.nih.gov/30835999/
* [Bergazin+, 2021] Enhancing water sampling of buried binding sites using NCMC https://pubmed.ncbi.nlm.nih.gov/32968887/
* [Melling+, 2023] Enhanced grand canonical sampling of occluded water sites using NCMC https://pubmed.ncbi.nlm.nih.gov/36692215/

Implementation context: Partially picks up from a draft implementation in timemachine a couple summers ago: https://github.com/proteneer/timemachine/blob/061ec7e325f876f6bfa69528a1d3b629e4b783a6/md/noneq/moves.py , with a couple twists: (1) includes a step to make propagation more "robust" i.e. less susceptible to blowing up when changes are introduced very abruptly (compared with MD alone), (2) considers a couple options for computing acceptance probabilities of proposal paths.

Status[2023-07-20]: Very WIP, and I have yet to update previous draft utilities for making selected waters "alchemical" https://github.com/proteneer/timemachine/blob/061ec7e325f876f6bfa69528a1d3b629e4b783a6/md/noneq/utils.py


examples/bulk_water_langevin_ncmc.py
```diff
@@ -0,0 +1,315 @@
+# gradually turn off a single water, displace, then turn on
```
maxentile, line 1: note: although this script used NVT rather than NPT, I don't think it was impacted by the bug fixed in https://github.com/proteneer/timemachine/pull/1168 (since it didn't use ixngroup)

maxentile:
May revisit after instantaneous moves.

PR #1092, opened 2023-07-20
Run CI linting without a GPU

badisa:


badisa:
Closing as it seems we have to hardcode a cuda arch in to the precommit script to do this and with more CI agents lower priority

PR #1096, opened 2023-07-25
Renames NonbondedPairListNegated to NonbondedExclusions

badisa:
* Simplifies and makes it clear what the intention of it is

One of the follow ups to #1090


maxentile (approved):

mcwitt:
This makes sense to me.

For the future, is there any benefit to maintaining this as a separate potential, given that it's only use case is exclusions? (e.g., it seems maybe simpler to revert to calling the `k_nonbonded_pair_list` kernel directly from the `NonbondedAllPairs` potential)

mcwitt (approved):

badisa:
> This makes sense to me.
> 
> For the future, is there any benefit to maintaining this as a separate potential, given that its only use case is exclusions? E.g., it seems maybe simpler to revert to calling the `k_nonbonded_pair_list` kernel directly from the `NonbondedAllPairs` potential. This would also reduce the surface area of a potential refactor implementing exclusions by omission rather than subtraction.

At the moment, the benefits of having it separate that I can think of are the following:

- Allows us to test that these cancel out (energy overflow tests)
- Simplifies local MD as you just always run exclusions, though with some work could be made to work if merged back in to the Nonbonded
- Parallelize over the two potentials, easily made to be parallel within the potential

We could work around all of these cases. I would like to move to just not computing the exclusions, but that is a non-trivial undertaking, especially for local MD.

PR #1098, opened 2023-07-25
Handle undetermined energies in BAR calculations

mcwitt:
Since https://github.com/proteneer/timemachine/pull/1084, we detect numerical overflows during the evaluation of potentials and represent undetermined energies as NaNs. This has exposed some cases where we previously returned invalid energies (for example, overflows appear to occur frequently during the initial evaluation of the BAR df error and overlap between the end states, before the first iteration of bisection).

Now, when we encounter an energy overflow, the resulting NaN in the `u_kln` matrix causes `pymbar.MBAR` (used to compute overlap) to fail with a `LinAlgError`, crashing the simulation. Also, `pymbar.BAR` warns and returns zeros for the $\Delta f$ and uncertainty estimates when there are NaN work values.

This PR addresses the latter issues by:

1. Detecting NaNs in `u_kln`. When we detect a NaN, we raise a warning and replace NaNs with `np.inf` (representing a configuration with zero probability).
1. Replacing usage of the `pymbar.BAR` estimator with `pymbar.MBAR` (on a 2-state `u_kn` matrix). The latter interprets `np.inf` as a configuration with zero weight (as desired), while the former returns `(0.0, 0.0)` if there are any infs or NaNs.
1. Switching to a cost function for bisection based on MBAR overlap (rather than bootstrapped $\Delta f$ error). See https://github.com/proteneer/timemachine/pull/1098#issuecomment-1652375914

This uncovered a related issue where the `timeout` parameter in `bootstrap_bar`, intended to cap computational cost, can lead to nondeterminism. This was addressed by

1. Removing timeout logic and the `timeout` parameter from `bootstrap_bar` and `bar_with_bootstrapped_uncertainty`
1. Modifying MBAR `relative_tolerance` and `maximum_iterations` to reduce cost
1. Upgrading `pymbar` from `3.0.5` to `3.1.0` (`3.0.6` [fixed](https://github.com/choderalab/pymbar/pull/425) a bug that prevented `maximum_iterations` from being respected)

Todo:
- [x] Add test for case with many NaNs, e.g. completely non-overlapping states
- [x] Is bootstrapped $\Delta f$ error still reliable for bisection?
  - ~seems robust for non-overlapping states in testing: uncertainty estimate produced by MBAR is finite and large~
  - EDIT: point above was wrong; MBAR-estimated $\Delta f$ errors are finite and large for pairs with zero overlap, but bootstrapped errors are zero in this case (since MBAR estimates $\Delta f \approx 0$ for each bootstrap sample). Addressed by switching to bisection cost function based on MBAR overlap.


tests/test_bar.py
```diff
@@ -76,16 +104,48 @@ def test_df_from_u_kln_compare_with_pymbar_bar(sigma):
...
+    w_F, w_R = works_from_ukln(u_kln)
+
+    # this example has some infinite work values
+    assert np.any(np.isinf(w_F)) or np.any(np.isinf(w_R))
+
+    # pymbar.BAR warns and returns zero for df and uncertainty if inf is present in either input
+    assert pymbar.BAR(w_F, w_R) == (0.0, 0.0)
+
+    df, df_err = df_and_err_from_u_kln(u_kln)
+    assert np.isfinite(df) and df != 0.0
```
maxentile, line 129: nit: could strengthen this assertion about `df`, as df_ref is I think in this example

timemachine/fe/bar.py
```diff
@@ -99,15 +99,37 @@ def dG_dw(w):
     return dG_dw
 
 
-def bootstrap_bar(w_F, w_R, n_bootstrap=1000, timeout=10) -> Tuple[float, NDArray]:
-    """Subsample w_F, w_R with replacement and re-run BAR many times
+def mbar_from_u_kln(u_kln: NDArray, **kwargs):
```
maxentile, line 102: weakly held style opinion: suspicious of forwarding **kwargs 

unsure if it's reasonable to make this more explicit (maybe expose only a solver tolerance option and initial guess option)?

timemachine/fe/bar.py
```diff
@@ -142,27 +166,22 @@ def bootstrap_bar(w_F, w_R, n_bootstrap=1000, timeout=10) -> Tuple[float, NDArra
...
 
-        bar_result = pymbar.BAR(
-            w_F=w_F_sample,
-            w_R=w_R_sample,
-            DeltaF=full_bar_result,  # warm start
-            compute_uncertainty=False,
+        bar_result = df_from_u_kln(
+            u_kln_sample,
+            initial_f_k=mbar.f_k,  # warm start
             relative_tolerance=1e-6,  # reduce cost
```
maxentile, line 174: may need to tweak this (or add a TODO to tweak this in the future) -- I picked this in context of pymbar.BAR, not sure if it'll have similar behavior in context of pymbar.MBAR

tests/test_bar.py
```diff
@@ -1,67 +1,164 @@
...
+
+    rng = np.random.default_rng(2023)
+
+    x_a = rng.uniform(0, 1, (1000,))
+    x_b = rng.uniform(0.5, 1.5, (1000,))
+
+    assert np.isfinite(u_a(x_a)).all()
+    assert np.isfinite(u_b(x_b)).all()
+
+    u_kln = np.array([[u_a(x_a), u_a(x_b)], [u_b(x_a), u_b(x_b)]])
```
maxentile, line 69: unsure of ordering of first two dimensions, don't remember convention

timemachine/fe/bar.py
mcwitt: Could also make this just return `u_kln` and `N_k` and leave the construction of the `MBAR` instance up to the caller. I was on the fence about doing this.

timemachine/fe/bar.py
badisa: Could we use the built-in bootstrapping (even in 3.1.1) for MBAR? https://github.com/choderalab/pymbar/blob/3.1.1/pymbar/mbar.py#L316-L339

Maybe not for consideration for this PR, especially since the bootstrapping in pymbar3 was done in parallel with the bootstrapping for pymbar4 (https://github.com/choderalab/pymbar/pull/302 vs https://github.com/choderalab/pymbar/pull/448) so maybe they differ?

tests/test_bar.py
mcwitt: Done in https://github.com/proteneer/timemachine/pull/1098/commits/d3120b8e1bdb80bf4dbe7ff32aa86ae12ef9243b

tests/test_bar.py
mcwitt: Added a note on convention in https://github.com/proteneer/timemachine/pull/1098/commits/0ed9719e5d2fad979457649857a2e2ec16d32d67

timemachine/fe/bar.py
badisa: Nevermind, ignore me. We are on 3.0.5 which doesn't have the bootstrapping.

timemachine/fe/bar.py
mcwitt: Should be addressed by https://github.com/proteneer/timemachine/pull/1098/commits/75da78a05c3929666ffa7c725f881ea34621aef0

tests/test_bar.py
maxentile: >based on the convention used in timemachine (k indexes evaluation state, l indexes sampling state); however, it seems like pymbar might use the opposite convention

Ahh, I think usages in timemachine may be mixed: another usage I'm aware of is consistent with the one in pymbar.utils.kln_to_kn https://github.com/proteneer/timemachine/blob/c4369bdd3accbb88d48cabd24c9dc209d19d94f6/timemachine/maps/estimators.py#L54-L61

Doesn't matter which convention is used if used uniformly (just a transpose), but mixed conventions could lead to bugs. (May be improved in the future by picking less ambiguous named indices, can afford more than a single letter per dimension.)

mcwitt: Created an issue to track this: https://github.com/proteneer/timemachine/issues/1100

timemachine/fe/bar.py
mcwitt: > Nevermind, ignore me. We are on 3.0.5 which doesn't have the bootstrapping.

Would like to return to this when we're able to upgrade to pymbar4 (this might actually be possible with the switch to MBAR, but haven't confirmed).

mcwitt: > I picked this in context of pymbar.BAR, not sure if it'll have similar behavior in context of pymbar.MBAR

@maxentile can you elaborate on the behavior you'd like to verify here? Had you run some specific tests for the previous variant that we could repeat here?

maxentile: @mcwitt : Sorry, I don't have well-stated requirements, and I don't have specific tests available.

I think had eyeballed for a favorable trade-off between speeding things up and not changing the results too much,  relative to default solver tolerance https://github.com/proteneer/timemachine/pull/829#issuecomment-1247266763 . I think on instances of increasing difficulty of a gaussian work test system, but I did not retain good notes unfortunately.

Since the errors will be severely underestimated anyway (since sample autocorrelation is large and ignored), preserving the behavior of the default solver tolerance doesn't seem critical. Don't know if there's a big speed difference between 1e-6 and the default tolerance (1e-7).

In the pymbar references @badisa posted, I think the bootstrap solver tolerance in 3.1.1. may have been set to 1e-12(?), and in 4 it may have been left at 1e-7 but other solver settings were changed.

mcwitt:
Even after preventing BAR calculations from failing when there are undetermined energies, there is a separate issue that bootstrapped $\Delta G$ errors are no longer useful for bisection: in the extreme case of zero overlap between states, pymbar's MBAR returns a $\Delta f$ estimate of 0.0 for every bootstrap sample, so the resulting bootstrapped error is 0. (See the failing test added in https://github.com/proteneer/timemachine/pull/1098/commits/749506386f744e8f3f65535863d0a5353f63068f)

In https://github.com/proteneer/timemachine/pull/1098/commits/54e4b5c007bcdf5bcf2c54fe47912bd449ca24ae I modified the bisection cost function to bisect states with lowest MBAR overlap, rather than maximum bootstrapped $\Delta G$ error. This works in the case of non-overlapping distributions, where the overlap is zero. We've also observed the MBAR overlap estimate to be more robust numerically compared with MBAR's $\Delta G$ error estimate.

timemachine/fe/bar.py
mcwitt: @maxentile thanks! I'd actually misread the comment location and thought you were referring to the `initial_f_k` setting. Starting with the default tolerance of `1e-7` when we don't have reason to choose something different seems reasonable to me.

mcwitt: Removed the override of the default tolerance setting in https://github.com/proteneer/timemachine/pull/1098/commits/5f93cf64ab8bc384b1d217902ae19a80a0219501

maxentile: >Starting with the default tolerance of 1e-7 when we don't have reason to choose something different seems reasonable to me.
>Removed the override of the default tolerance setting in https://github.com/proteneer/timemachine/commit/5f93cf64ab8bc384b1d217902ae19a80a0219501

Nice -- agreed!

(Can tweak later if we find a need to.)

timemachine/fe/bar.py
```diff
@@ -99,15 +98,64 @@ def dG_dw(w):
...
     ----------
-    w_F : array
-        forward works
-    w_R : array
-        reverse works
+    u_kln : array
+        2-state u_kln matrix
     n_bootstrap : int
         # bootstrap samples
     timeout : int
```
maxentile, line 161: nit: remove documentation of timeout here and in the description of return types

timemachine/fe/bar.py
```diff
@@ -99,15 +98,64 @@ def dG_dw(w):
...
+) -> float:
+    """Compute free energy difference given a 2-state u_kln matrix."""
+    u_kn, N_k = ukln_to_ukn(u_kln)
+    mbar = pymbar.MBAR(u_kn, N_k, initial_f_k=initial_f_k, maximum_iterations=maximum_iterations)
+    df = mbar.getFreeEnergyDifferences(compute_uncertainty=False)[0]
+    return df[0, 1]
+
+
+def bootstrap_bar(u_kln: NDArray, n_bootstrap=100) -> Tuple[float, NDArray]:
+    """Given a 2-state u_kln matrix, subsample u_kln with replacement and re-run bar_from_u_kln many times
```
maxentile, line 153: nit: name of function being called many times

timemachine/fe/bar.py
```diff
@@ -126,43 +174,34 @@ def bootstrap_bar(w_F, w_R, n_bootstrap=1000, timeout=10) -> Tuple[float, NDArra
...
-            w_R=w_R_sample,
-            DeltaF=full_bar_result,  # warm start
-            compute_uncertainty=False,
-            relative_tolerance=1e-6,  # reduce cost
-        )
+        u_kln_sample = rng.choice(u_kln, size=(n,), replace=True, axis=2)
 
+        bar_result = df_from_u_kln(
+            u_kln_sample,
+            initial_f_k=mbar.f_k,  # warm start
```
maxentile, line 194: q: would it make sense to add bootstrap_maximum_iterations to signature of bootstrap_bar, then forward it here?

maxentile (approved):

timemachine/fe/bar.py
mcwitt: Done in https://github.com/proteneer/timemachine/pull/1098/commits/0ddf9a0ed7007c38f2e809cd7935ec816bfd3160

timemachine/fe/bar.py
mcwitt: Fixed in https://github.com/proteneer/timemachine/pull/1098/commits/f0f34140697bbeada47f1fa7177a2402ef2a62c3

setup.py
```diff
@@ -102,7 +102,7 @@ def build_extension(self, ext):
         "jaxlib>0.4.1",
         "networkx",
         "numpy",
-        "pymbar>3.0.4,<4",
+        "pymbar>=3.0.6,<4",
```
jkausrelay, line 105: nit: Should this be 3.1.0 or higher? Not sure why this is different from requirements

timemachine/fe/bar.py
mcwitt: Unsure whether it would be useful to be able to specify max iterations separately for the point estimate and the bootstrap samples, but this can probably be added later if useful.

tests/test_bar.py
```diff
@@ -1,70 +1,192 @@
...
-        print(f"stddev(w_F) = {sigma_F}, bootstrap uncertainty = {bootstrap_sigma}, pymbar.BAR uncertainty = {ddf_ref}")
-        assert df_0 == df_ref
-        assert df_1 == df_ref
-        assert len(bootstrap_samples) == n_bootstrap, "timed out on default problem size!"
-        np.testing.assert_approx_equal(bootstrap_sigma, ddf_ref, significant=1)
+    # assert estimates identical, uncertainties comparable
+    print(f"bootstrap uncertainty = {bootstrap_sigma}, pymbar.MBAR uncertainty = {df_err_ref}")
+    assert df_0 == df_ref
+    assert df_1 == df_ref
+    assert len(bootstrap_samples) == n_bootstrap, "timed out on default problem size!"
```
jkausrelay, line 93: nit: Is this still relevant?

tests/test_free_energy.py
```diff
@@ -351,3 +355,33 @@ def result_with_overlap(overlap):
...
+    with pytest.warns(IndeterminateEnergyWarning, match="NaN"):
+        result_with_nan = estimate_free_energy_bar(np.array([u_kln_with_nan]), DEFAULT_TEMP)
+
+    u_kln_with_inf = np.array(u_kln)
+    u_kln_with_inf[0, 1, 10] = np.inf
+
+    # should give the same result with inf
+    result_with_inf = estimate_free_energy_bar(np.array([u_kln_with_inf]), DEFAULT_TEMP)
+    assert result_with_nan.dG == result_with_inf.dG
+    assert result_with_nan.dG_err == result_with_inf.dG_err
```
jkausrelay, line 385: nit: assert finite

timemachine/fe/bar.py
maxentile: >Unsure whether it would be useful to be able to specify max iterations separately for the point estimate and the bootstrap samples, but this can probably be added later if useful.

Separately makes sense to me (to control expense), but can be added later if needed.

maxentile (approved):

timemachine/fe/bar.py
```diff
@@ -99,70 +98,113 @@ def dG_dw(w):
...
+    u_kn, N_k = ukln_to_ukn(u_kln)
+    mbar = pymbar.MBAR(u_kn, N_k, maximum_iterations=maximum_iterations)
+    try:
+        df, ddf = mbar.getFreeEnergyDifferences()
+        return df[0, 1], ddf[0, 1]
+    except pymbar.utils.ParameterError:
+        # As of pymbar 3.1.0, computation of the covariance matrix can raise an exception on incomplete convergence.
+        # In this case, return the unconverged estimate with NaN as uncertainty.
+        df = mbar.getFreeEnergyDifferences(compute_uncertainty=False)[0]
+        return df[0, 1], np.nan
```
jkausrelay, line 139: nit: nan or inf?

timemachine/fe/bar.py
```diff
@@ -99,70 +98,113 @@ def dG_dw(w):
...
+
+    Parameters
+    ----------
+    u_kln : array
+        2-state u_kln matrix
     n_bootstrap : int
-        # bootstrap samples
-    timeout : int
-        in seconds
+        number o bootstrap samples
```
jkausrelay, line 162: typo: o -> of

jkausrelay (approved):
Minor comments, overall LGTM.

timemachine/fe/bar.py
mcwitt: Leaning toward NaN as more appropriate here, since we can't be sure of the failure reason. The choice here doesn't affect bisection, since we now use overlap.

setup.py
mcwitt: `<3.0.6` definitely won't work because of https://github.com/choderalab/pymbar/pull/425, which was merged in `3.0.6`. In general I prefer to use version constraints in `setup.py` only to exclude known-incompatible versions.

When upgrading, I opted to use the latest release <4.

badisa (approved):

timemachine/fe/bar.py
jkausrelay: Makes sense.

timemachine/fe/bar.py
mcwitt: Fixed in https://github.com/proteneer/timemachine/pull/1098/commits/6950d82b1d3b12b8fb1951bdd6020c009b826caa

tests/test_bar.py
mcwitt: Kept assertion but removed misleading message in https://github.com/proteneer/timemachine/pull/1098/commits/28c6d9ffac52398d2c3f0d0509a1d590c3a1a9d8

PR #1104, opened 2023-07-27
Raise InvalidHardware exception if no gpu

badisa:
* Will allow us to handle hardware failures as special cases


jkausrelay (approved):

timemachine/cpp/src/gpu_utils.cuh
```diff
@@ -21,8 +22,22 @@ templateCurandNormal(curandGenerator_t generator, double *outputPtr, size_t n, d
...
+            // If the GPU is invalid or missing for some reason, raise an exception so we can handle that
+            // Error codes can be found here: https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1g3f51e3575c2178246db0a94a430e0038
+            switch (code) {
+            case cudaErrorInvalidDevice:
+            case cudaErrorInsufficientDriver:
+            case cudaErrorNoDevice:
+            case cudaErrorStartupFailure:
+            case cudaErrorInvalidPtx:
+            case cudaErrorUnsupportedPtxVersion:
+                throw InvalidHardware();
```
mcwitt, line 35: Would it be useful to include the error code in the exception?

timemachine/cpp/src/exceptions.hpp
```diff
@@ -0,0 +1,7 @@
+#pragma once
+#include <stdexcept>
+
+class InvalidHardware : public std::exception {
+public:
+    const char *what() const throw() { return "Either no GPU or the GPU is acting up"; }
```
mcwitt, line 6: This formatting seems weird - could `what` and `throw` be defined on separate lines?

timemachine/cpp/src/gpu_utils.cuh
badisa: Would the error string work as the thing to return? As the codes won't be accessible from Python (though we could expose them)

mcwitt: Not sure, just seems like it could be useful if there was a way to tell from the Python exception (possibly by referencing the CUDA API docs) which of these conditions we hit.

timemachine/cpp/src/exceptions.hpp
badisa: Switched to `const noexcept` as that is the more modern style of `const throw()` which indicates that the function doesn't raise an exception. https://github.com/proteneer/timemachine/pull/1104/commits/b0723ab266a82782e07ab114063769b08ce6e3e0

timemachine/cpp/src/gpu_utils.cuh
badisa: Added code along with the message as a string in https://github.com/proteneer/timemachine/pull/1104/commits/b0723ab266a82782e07ab114063769b08ce6e3e0

Didn't add the int on the exception as it isn't clear how to do that and don't think it would be of much utility.

timemachine/cpp/src/exceptions.hpp
mcwitt: Ah I see, was misinterpreting that as separate definitions. Makes sense.

mcwitt (approved):

PR #1107, opened 2023-07-31
[wip] Utilities for per-atom decoupling

maxentile:
Adds and tests basic utilities for per-atom decoupling:
* extracting atom ranks based on shortest path lengths to selected atoms
* decoupling all atoms with `(rank == i)` before decoupling all atoms with `(rank == (i+1))`

(A similar goal was considered here in https://github.com/proteneer/timemachine/pull/420 , and these utilities derived from the `sifting.py` file in @proteneer 's branch https://github.com/proteneer/timemachine/blob/acccb46c3ed7eaf614fd79b42e42db597a546891/examples/sifting.py#L212-L328 (One notable difference relative to `sifting.py` is that this PR considers the ranks in the opposite order, i.e. decoupling of rank i precedes decoupling of rank i+1, but it might make sense to reverse this.)))

The implementation is currently only applicable to the "absolute" case (i.e. "deletions only"), but see also related functions in transformato for scheduling one-atom-at-a-time mutations that might involve both additions and deletions https://github.com/wiederm/transformato/blob/77d5f0fd2e3e9a0d67c0ecc817bfe8389295a3aa/transformato/annihilation.py#L10 .

### DIFF
--- a/tests/test_experimental.py
+++ b/tests/test_experimental.py
@@ -1,0 +1,42 @@
+import numpy as np
+from jax import vmap
+
+from timemachine.datasets import fetch_freesolv
+from timemachine.fe.experimental import DecoupleByAtomRank, rank_atoms_by_path_length_to_src
+
+
+def test_decouple_by_atom_rank():
+    """for a few random molecules, make basic assertions about per-atom decoupling schedule"""
+
+    mols = fetch_freesolv()
+
+    num_test_mols = 10
+    np.random.seed(1234)
+    lambdas = np.linspace(0, 1, 100)
+    for mol in mols[:num_test_mols]:
+        n_atoms = mol.GetNumAtoms()
+        n_src_idxs = np.random.randint(1, min(5, n_atoms))  # arbitrary threshold: between 1 and 5 src_idxs
+        src_idxs = list(set(np.random.randint(0, n_atoms, n_src_idxs)))
+
+        atom_idxs = np.arange(n_atoms)
+        atom_ranks = rank_atoms_by_path_length_to_src(mol, src_idxs)
+        assert atom_ranks.shape == atom_idxs.shape
+
+        decoupler = DecoupleByAtomRank(atom_idxs, atom_ranks)
+        atom_lams = vmap(decoupler.atom_lams_from_global_lam)(lambdas)
+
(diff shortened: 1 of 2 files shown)


proteneer:
Nice! Would be also neat at a future point to add weights (propto sigma/epsilon or something)

maxentile:
>Nice! Would be also neat at a future point to add weights (propto sigma/epsilon or something)
>Caveat: with some something like Bisection in equilibrium methods or ESS in non-equilibrium methods, maybe weights are less important. (Presumably we'd naturally spend more fractional lambdas on atoms that have large sigmas).

That sounds useful.

Clarification: Do you mean adding weights to
* edges (e.g. add a weighted path length heuristic for determining atom ranks) or
* nodes (e.g. "budget 1.5x as long to decoupling atom i compared with atom j")?

If the second, I'd be inclined to defer this to the other methods you mentioned for adapting the spacing.

proteneer:
I was thinking about nodes initially. I agree that just letting ESS/Bisection deal with per node spacing.

Not sure how to weight on edges yet - good point, and will need to think about it a bit more.

maxentile:
>Weighted path length (where weights are prop to bond lengths) probably makes sense too, but I don't know how variation there is in the bond-lengths.

I don't have any good ideas for edge weight heuristic yet. Bond lengths don't seem very informative as weights, initially.

For now, I think I'll keep the example atom ranks heuristic unweighted, but if we have other compelling heuristics we can add them.

>I was thinking about nodes initially. I agree that just letting ESS/Bisection deal with per node spacing is better.

To give the adaptive spacing method more flexibility here, I've changed the behavior of the default atom ranks heuristic in https://github.com/proteneer/timemachine/pull/1107/commits/7ca9a50baee45b3a6455739fb9d59d97171daaac to argsort.(i.e. give every atom a unique rank).

(Bisection could then allocate many steps to handle a single difficult atom, or handle several atoms in a single large step, as appropriate.)

maxentile:
Abandoned, and didn't turn out to be necessary: https://github.com/proteneer/timemachine/pull/1158/commits/2d24f12a0c41cc13ce707b6acbe7b7865b2f91cf 

(Although it might be interesting to revisit in other contexts)

PR #1108, opened 2023-08-01
Add `nocuda` mark for tests that should run without CUDA available

mcwitt:
The current `nogpu` mark is overloaded with at least 2 distinct intents:

1. Save cost by running tests that don't require GPU on cheaper CPU instances
2. Ensure that certain code runs without the CUDA extension module being built (e.g. on a macbook)

As of https://github.com/proteneer/timemachine/pull/1079, we build the extension module in CI before running `nogpu` tests, making this designation less effective for use (2).

This PR introduces a new mark, `nocuda`, intended for use with tests that should run without the extension module being built. The proposed intended meanings are then:

- `nocuda`: this should run without the CUDA extension module available, e.g. on a user's macbook
- `nogpu`: this requires the CUDA extension module, but does not actually use the GPU (so we can run it on a CPU instance)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -47,6 +47,27 @@ lint:
   script:
     - make verify
 
+nocuda-tests:
+  stage: test
+  image: $DOCKER_TAG
+  needs: ["lint"]
+  tags:
+    - timemachine
+    - cpu
+  rules:
+    - if: $CI_EXTERNAL_PULL_REQUEST_IID
+  variables:
+    CMAKE_ARGS: -DCUDA_ARCH=75
+  script:
+    - SKIP_CUSTOM_OPS=1 pip install .[test]
+    - make nocuda_tests
+  artifacts:
+    name: "$CI_JOB_NAME-$CI_COMMIT_REF_NAME-nocuda-tests"
+    paths:
+        - coverage/
+    when: on_success
+    expire_in: 1 week
+
 nogpu-tests:
   stage: test
   image: $DOCKER_TAG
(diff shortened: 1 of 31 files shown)


tests/test_vacuum_importance_sampling.py
```diff
@@ -15,7 +15,7 @@ def get_ff_am1ccc():
     return ff
 
 
-@pytest.mark.nogpu
```
badisa, line 18: It doesn't look like there is a `nightly_no_cuda` CI stage, does this run?

.gitlab-ci.yml
```diff
@@ -47,6 +47,27 @@ lint:
...
+  stage: test
+  image: $DOCKER_TAG
+  needs: ["lint"]
+  tags:
+    - timemachine
+    - cpu
+  rules:
+    - if: $CI_EXTERNAL_PULL_REQUEST_IID
+  variables:
+    CMAKE_ARGS: -DCUDA_ARCH=75
```
badisa, line 60: Don't need the Cuda arch

tests/test_vacuum_importance_sampling.py
mcwitt: Good catch. I suppose this could be changed back to a `nogpu` test, but for consistency I opted to introduce a `nocuda_nightly_tests` target in https://github.com/proteneer/timemachine/pull/1108/commits/5e798c90b3d5aab08a4ab099d2c02e1c144f2f47.

.gitlab-ci.yml
mcwitt: Removed in https://github.com/proteneer/timemachine/pull/1108/commits/3e45db9e90882bddfc0a7b3271d1e0728879d4cd.

jkausrelay (approved):

PR #1112, opened 2023-08-08
Removes local MD burn in parameter

badisa:
* Didn't end up necessary
* Ended up not being used in https://arxiv.org/abs/2305.05475 as it didn't play as an important role as we had expected. 
* You can replicate the behavior by calling `multiple_steps_local` twice if we determine that we need this again.


jkausrelay (approved):

maxentile (approved):
Looks good!

Clarification: is this comment still applicable https://github.com/proteneer/timemachine/pull/1112/files#diff-083440281823a69425cd546f57c1bf3e60feceea58a481bb073bca90c6110456L279-L280 ?

badisa:
> Looks good!
> 
> Clarification: is this comment still applicable https://github.com/proteneer/timemachine/pull/1112/files#diff-083440281823a69425cd546f57c1bf3e60feceea58a481bb073bca90c6110456L279-L280 ?

Yes, that comment is still applicable. The burn in was partly to address that comment, but really the number of steps overall should be > 100 and seems gross to control it with two parameters rather than one.

PR #1105, opened 2023-07-29
Speed up k_find_block_bounds

badisa:
- Speeds up the k_find_block_bounds by ~3x (24k ns to 7.5k ns) by using shuffle syncs
- [x] Exposed that we have a gap in the tests, as another implementation passed all the tests but resulted in a 30% slow down due to certain cases of box bounds being much larger. ~Still investigating~

# Why bad implementations of k_find_block_bounds could pass when generating bad boxes
The issue was that the python was actually imaging the coordinates prior to passing them into the Neighborlist (https://github.com/proteneer/timemachine/blob/master/tests/test_nblist.py#L41-L59) which resulted in never actually testing that PBCs were handled correctly in the GPU kernel.

# Benchmarks
A10 Cuda Arch 86

Ranges from a <1% speed up to a 3% speed up. The cases where it is most significant is where there are more neighborlists which is going to be the non-apo cases where there are all pairs + ixn groups which result in more neighborlist evaluations

## Master
```
dhfr-apo: N=23558 speed: 694.34ns/day dt: 2.5fs (ran 100000 steps in 31.11s)
dhfr-apo-barostat-interval-25: N=23558 speed: 622.48ns/day dt: 2.5fs (ran 100000 steps in 34.70s)
building a protein system with 1758 protein atoms and 7047 water atoms
hif2a-apo: N=8805 speed: 1177.04ns/day dt: 2.5fs (ran 100000 steps in 18.35s)
hif2a-apo-barostat-interval-25: N=8805 speed: 1005.22ns/day dt: 2.5fs (ran 100000 steps in 21.49s)
hif2a-rbfe: N=8840 speed: 901.40ns/day dt: 2.5fs (ran 100000 steps in 23.97s)
hif2a-rbfe-local: N=8840 speed: 1298.07ns/day dt: 2.5fs (ran 100000 steps in 16.64s)
solvent-apo: N=6282 speed: 1459.73ns/day dt: 2.5fs (ran 100000 steps in 14.80s)
solvent-apo-barostat-interval-25: N=6282 speed: 1286.07ns/day dt: 2.5fs (ran 100000 steps in 16.80s)
solvent-rbfe: N=6317 speed: 1156.37ns/day dt: 2.5fs (ran 100000 steps in 18.68s)
solvent-rbfe-local: N=6317 speed: 1413.71ns/day dt: 2.5fs (ran 100000 steps in 15.28s)
```

## Shuffle syncing across threads
```
[...]

PR #1113, opened 2023-08-08
Refactor hilbert sorting into HilbertCurve class

badisa:
* Could allow us to use larger grid sizes by sharing static buffer between instances of classes without incurring N^3 memory usage
* Reduces the amount of duplicated logic between AllPairs and Ixn group code
* More of a deboggling task, and was written as part of #1105 to determine the source of the test passing incorrectly. Don't need to merge, but already written so I wanted to open up to review

### DIFF
--- a/timemachine/cpp/src/hilbert_curve.cu
+++ b/timemachine/cpp/src/hilbert_curve.cu
@@ -1,0 +1,85 @@
+#include "gpu_utils.cuh"
+#include "hilbert_curve.hpp"
+#include "kernels/k_hilbert.cuh"
+#include "vendored/hilbert.h"
+#include <cub/cub.cuh>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+namespace timemachine {
+
+HilbertCurve::HilbertCurve(int N)
+    : N_(N), d_bin_to_idx_(HILBERT_GRID_DIM * HILBERT_GRID_DIM * HILBERT_GRID_DIM), d_sort_keys_in_(N),
+      d_sort_keys_out_(N), d_sort_vals_in_(N), d_sort_storage_(nullptr), d_sort_storage_bytes_(0) {
+    // initialize hilbert curve
+    std::vector<unsigned int> bin_to_idx(HILBERT_GRID_DIM * HILBERT_GRID_DIM * HILBERT_GRID_DIM);
+    for (int i = 0; i < HILBERT_GRID_DIM; i++) {
+        for (int j = 0; j < HILBERT_GRID_DIM; j++) {
+            for (int k = 0; k < HILBERT_GRID_DIM; k++) {
+
+                bitmask_t hilbert_coords[3];
+                hilbert_coords[0] = i;
+                hilbert_coords[1] = j;
+                hilbert_coords[2] = k;
+
+                unsigned int bin = static_cast<unsigned int>(hilbert_c2i(3, HILBERT_N_BITS, hilbert_coords));
+                bin_to_idx[i * HILBERT_GRID_DIM * HILBERT_GRID_DIM + j * HILBERT_GRID_DIM + k] = bin;
+            }
+        }
+    }
+
+    d_bin_to_idx_.copy_from(&bin_to_idx[0]);
+
+    // estimate size needed to do radix sorting
+    // reuse d_sort_keys_in_ rather than constructing a dummy output idxs buffer
+    cub::DeviceRadixSort::SortPairs(
+        nullptr,
+        d_sort_storage_bytes_,
+        d_sort_keys_in_.data,
+        d_sort_keys_out_.data,
+        d_sort_vals_in_.data,
+        d_sort_keys_in_.data,
+        N_);
+    gpuErrchk(cudaPeekAtLastError());
+
+    d_sort_storage_.reset(new DeviceBuffer<char>(d_sort_storage_bytes_));
+}
+
+HilbertCurve::~HilbertCurve(){};
+
+void HilbertCurve::sort_device(
+    const int N,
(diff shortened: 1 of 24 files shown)


timemachine/cpp/src/hilbert_curve.cu
```diff
@@ -0,0 +1,85 @@
...
+                hilbert_coords[1] = j;
+                hilbert_coords[2] = k;
+
+                unsigned int bin = static_cast<unsigned int>(hilbert_c2i(3, HILBERT_N_BITS, hilbert_coords));
+                bin_to_idx[i * HILBERT_GRID_DIM * HILBERT_GRID_DIM + j * HILBERT_GRID_DIM + k] = bin;
+            }
+        }
+    }
+
+    d_bin_to_idx_.copy_from(&bin_to_idx[0]);
```
badisa, line 32: Not done in this PR, but we could make this buffer static across multiple instances of HilbertCurve to remove the N^3 memory usage per curve and investigate using larger grid sizes (reduced in #692)

timemachine/cpp/src/nonbonded_all_pairs.hpp
```diff
@@ -46,16 +48,10 @@ template <typename RealType> class NonbondedAllPairs : public Potential {
...
-    unsigned int *d_bin_to_idx_; // mapping from 256x256x256 grid to hilbert curve index
-    unsigned int *d_sort_keys_in_;
-    unsigned int *d_sort_keys_out_;
-    unsigned int *d_sort_vals_in_;
-    unsigned int *d_sort_storage_;
-    size_t d_sort_storage_bytes_;
-
     cudaEvent_t nblist_flag_sync_event_; // Event to synchronize on
 
+    std::unique_ptr<HilbertCurve> hilbert_sort_;
```
mcwitt, line 53: nit: could the name of the class be `HilbertSort` too? (seems more descriptive of its purpose vs. just representing an abstract hilbert curve)

timemachine/cpp/src/hilbert_curve.cu
```diff
@@ -0,0 +1,85 @@
...
+#include "kernels/k_hilbert.cuh"
+#include "vendored/hilbert.h"
+#include <cub/cub.cuh>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+namespace timemachine {
+
+HilbertCurve::HilbertCurve(int N)
```
mcwitt, line 12: Would it make sense for `N` to be a template parameter to enable potential optimizations?

timemachine/cpp/src/nonbonded_all_pairs.cu
```diff
@@ -238,7 +171,7 @@ void NonbondedAllPairs<RealType>::execute_device(
 
     // (ytz) the nonbonded algorithm proceeds as follows:
 
-    // (done in constructor), construct a hilbert curve mapping each of the HILBERT_GRID_DIM x HILBERT_GRID_DIM x HILBERT_GRID_DIM cells into an index.
+    // (done in constructor), construct a hilbert curve sorting.
```
mcwitt, line 174: I found the original comment useful. Is this description replicated somewhere else?

timemachine/cpp/src/hilbert_curve.cu
mcwitt: Disregard, sorry. I had originally misinterpreted `N` as the size of the grid.

badisa: N is dependent on the system size, so we couldn't make it a template parameter without some weirdness. By constant do you simply mean `const int N` as the input parameter? That is doable, though don't think it would make much difference since `N_` is `const`.

mcwitt (approved):
Looks great, nice DRYing!

timemachine/cpp/src/nonbonded_all_pairs.cu
badisa: Will add back in the comment in the hilbert constructor

timemachine/cpp/src/nonbonded_all_pairs.hpp
badisa: Seems reasonable to me, had started with that but switched to imitate the python library we use.

timemachine/cpp/src/nonbonded_all_pairs.cu
badisa: Added back in [3e0341c](https://github.com/proteneer/timemachine/pull/1113/commits/3e0341cba746d8cf46706d888db67e8c28823f5b)

timemachine/cpp/src/hilbert_curve.cu
badisa: Made the param `const` to be consistent with the `.hpp` definition of the constructor.

timemachine/cpp/src/nonbonded_all_pairs.hpp
badisa: Renamed in [fbe6ca8](https://github.com/proteneer/timemachine/pull/1113/commits/fbe6ca82469b404246fa0d943e5050e15ec7fcfe)

proteneer (approved):
lgtm! looks pretty clerical. Do we have any intention of ever exposing the hilbert sorter into python? (maybe add a test? not sure if the ordering is consistent with pypi's `hilbertcurve` package.)

badisa:
> lgtm! looks pretty clerical. Do we have any intention of ever exposing the hilbert sorter into python? (maybe add a test? not sure if the ordering is consistent with pypi's `hilbertcurve` package.)

I had thought about it, but didn't see a clear use case. Will follow up with a PR to expose this to python and add a test.

PR #1115, opened 2023-08-09
Misc fixes and improvements to RBFE code

mcwitt:
This PR collects some miscellaneous fixes and improvements, spun out of the HREX development branch to simplify review of the latter:

- Renames some internal RBFE methods that previously had unwieldy names, e.g. `estimate_relative_free_energy_via_greedy_bisection` -> `estimate_relative_free_energy_bisection` (https://github.com/proteneer/timemachine/pull/1115/commits/d5cbe0a36f9f6c29452ab45876e6f7a8738a37b8)
- Fixes overly-restrictive argument validation in `estimate_relative_free_energy_bisection`. E.g. should allow empty `keep_idxs`. (https://github.com/proteneer/timemachine/pull/1115/commits/c5b48fabcec270018a5de3f05e2626052ae43477)
- Simplifies trajectory accumulation in `run_production_local_steps`, avoids copying current coordinates and boxes after each batch (https://github.com/proteneer/timemachine/pull/1115/commits/82c7cbe2c7f00d9f368ee455fd698dcb50ca2b69)
- Makes `MonteCarloMove` generic in the state type, so it can be used with types other than `CoordsVelBox`. This enables reusing this code for HREX, where the state type is a mapping from states to replicas. (https://github.com/proteneer/timemachine/pull/1115/commits/355b9835a1b01245e214122fb32af697e6853d3a)

### DIFF
--- a/timemachine/fe/rbfe.py
+++ b/timemachine/fe/rbfe.py
@@ -585,9 +585,7 @@ def estimate_relative_free_energy_bisection(
                 stored_frames.append(np.array(frames[i]))
                 stored_boxes.append(boxes[i])
             except IndexError:
-                warnings.warn(
-                    f"Invalid index in keep_idxs: {i}. Bisection terminated with only {len(results)} windows."
-                )
+                warnings.warn(f"Invalid index in keep_idxs: {i}. Bisection terminated with only {len(frames)} windows.")
 
         return SimulationResult(
             final_result,
--- a/timemachine/fe/rbfe.py
+++ b/timemachine/fe/rbfe.py
@@ -537,10 +537,8 @@ def estimate_relative_free_energy_bisection(
 
     assert len(keep_idxs) <= n_windows
 
-    if min_overlap is not None and keep_idxs not in [None, [0, -1], list(range(n_windows))]:
-        raise ValueError(
-            "when min_overlap is not None, keep_idxs must be equal to one of None, [0, -1], or list(range(n_windows))"
-        )
+    if min_overlap is not None and keep_idxs is not None and not all(i in {0, -1} for i in keep_idxs):
+        raise ValueError("when min_overlap is not None, keep_idxs must be None or a list with all elements in {0, -1}")
 
     single_topology = SingleTopology(mol_a, mol_b, core, ff)
 
--- a/timemachine/fe/rbfe.py
+++ b/timemachine/fe/rbfe.py
@@ -20,8 +20,8 @@ from timemachine.fe.free_energy import (
     SimulationResult,
     estimate_free_energy_bar,
     make_pair_bar_plots,
+    run_sims_bisection,
     run_sims_sequential,
-    run_sims_with_greedy_bisection,
 )
 from timemachine.fe.single_topology import SingleTopology
 from timemachine.fe.system import VacuumSystem, convert_omm_system
@@ -458,7 +458,7 @@ def estimate_relative_free_energy(
         raise err
 
 
-def estimate_relative_free_energy_via_greedy_bisection(
+def estimate_relative_free_energy_bisection(
     mol_a: Chem.rdchem.Mol,
     mol_b: Chem.rdchem.Mol,
     core: NDArray,
@@ -568,7 +568,7 @@ def estimate_relative_free_energy_via_greedy_bisection(
     combined_prefix = get_mol_name(mol_a) + "_" + get_mol_name(mol_b) + "_" + prefix
 
     try:
-        results, frames, boxes = run_sims_with_greedy_bisection(
+        results, frames, boxes = run_sims_bisection(
             [lambda_min, lambda_max],
             make_optimized_initial_state,
             md_params,
@@ -615,7 +615,7 @@ def run_vacuum(
         md_params = replace(md_params, local_steps=0)
         warnings.warn("Vacuum simulations don't support local steps, will use all global steps")
     # min_cutoff defaults to None since there is no environment to prevent conformational changes in the ligand
-    return estimate_relative_free_energy_via_greedy_bisection(
+    return estimate_relative_free_energy_bisection(
         mol_a,
         mol_b,
         core,
@@ -646,7 +646,7 @@ def run_solvent(
     solvent_sys, solvent_conf, solvent_box, solvent_top = builders.build_water_system(box_width, forcefield.water_ff)
     solvent_box += np.diag([0.1, 0.1, 0.1])  # remove any possible clashes, deboggle later
     solvent_host_config = HostConfig(solvent_sys, solvent_conf, solvent_box, solvent_conf.shape[0])
-    solvent_res = estimate_relative_free_energy_via_greedy_bisection(
+    solvent_res = estimate_relative_free_energy_bisection(
         mol_a,
         mol_b,
         core,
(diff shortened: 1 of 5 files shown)


timemachine/fe/rbfe.py
```diff
@@ -537,10 +537,8 @@ def estimate_relative_free_energy_via_greedy_bisection(
 
     assert len(keep_idxs) <= n_windows
 
-    if min_overlap is not None and keep_idxs not in [None, [0, -1], list(range(n_windows))]:
-        raise ValueError(
-            "when min_overlap is not None, keep_idxs must be equal to one of None, [0, -1], or list(range(n_windows))"
-        )
+    if min_overlap is not None and keep_idxs is not None and not all(i in {0, -1} for i in keep_idxs):
```
badisa, line 540:
```suggestion
    if min_overlap is not None and not all(i in {0, -1} for i in keep_idxs):
```
keep_idxs is already a list at this line, since we check len(keep_idxs) which would raise an exception if it was None.

Doesn't this remove the ability to have `keep_idxs` be anything besides `[0, -1]`? Should `[3, 5]` be invalid? Looks like the old code was similar as well, only allowing None (which we set to `[0, -1]` above https://github.com/proteneer/timemachine/pull/1115/files#diff-49ef8469d9fb5c9d39e05a0cef32710703096c53f657e55eff49d5c3ddd47b13R536), `list(range(windows))` and `[0, -1]`. Think I am missing why `keep_idxs` is dependent on the state of `min_overlap`, doesn't the min overlap get evaluated on intermediate windows that we don't necessarily store?

mcwitt: > keep_idxs is already a list at this line, since we check len(keep_idxs) which would raise an exception if it was None.

Good catch, I'd missed that we already check for `None` above.

> Doesn't this remove the ability to have keep_idxs be anything besides [0, -1]? Should [3, 5] be valid?

I don't think `[3, 5]` or any indices except `0` and `-1` (first and last) are valid when adaptive bisection is enabled (i.e. `min_overlap` is not `None`), because we don't know ahead of time how many windows there will be. This was changed in #1093 as part of a fix for an issue where we'd been passing something like `keep_idxs=[0, n_windows - 1]`, which caused an `IndexError` when bisection resulted in fewer than `n_windows` windows.

This could still be relaxed a bit (e.g. we will always have at least 2 windows, so `1` might be allowed), but in general arbitrary indices aren't safe here.

badisa: Does that then mean we are unable to retrieve frames for an intermediate window during adaptive bisection (without modifying the code)? Storing intermediate frames seems like a very useful feature for debugging if something goes awry in an intermediate lambda window.

mcwitt: That's true, it would currently require changing the code. We should probably rethink this logic, as this had been intended as a quick patch for production, where we typically only keep first and last.

Could do something like ignore indices that don't exist at the end of a run, silently or with a warning? There should also probably be an option for "keep all frames", though that would be achievable with the proposed behavior by passing `range(n_windows)`.

badisa: Adding back in the support for `range(n_windows)` would probably be good for now, that way frames can get gotten back out at the expense of disk space.

timemachine/fe/rbfe.py
```diff
@@ -537,10 +537,8 @@ def estimate_relative_free_energy_via_greedy_bisection(
 
     assert len(keep_idxs) <= n_windows
 
-    if min_overlap is not None and keep_idxs not in [None, [0, -1], list(range(n_windows))]:
-        raise ValueError(
-            "when min_overlap is not None, keep_idxs must be equal to one of None, [0, -1], or list(range(n_windows))"
-        )
+    if min_overlap is not None and not all(i in {0, -1} for i in keep_idxs):
```
badisa, line 540: nit: Might be worth while to assert `len(keep_idxs) == 2` here as well since the current check allows `keep_idxs = [0, -1, -1]`

timemachine/fe/rbfe.py
badisa: Woops had misunderstood and thought we could just add back in the previous check for `list(range(n_windows))`. For now this is fine as it prevents an issue that already exists.

Would be good to follow up on this, as seems important to be able to get out intermediate frames if we need them.

badisa (approved):

timemachine/fe/rbfe.py
mcwitt: Tried to improve this in https://github.com/proteneer/timemachine/pull/1115/commits/3388ef2cc181e973e049cacb0e0922d57d592a88.

This reverts to the old behavior w.r.t `keep_idxs`, except that we'll now skip storing (with warning) indices that are invalid after bisection is complete. So e.g. `range(n_windows)` is valid as an input and will result in storing frames for at most `n_windows` windows.

timemachine/fe/rbfe.py
mcwitt: Added an assertion of unique indices in https://github.com/proteneer/timemachine/pull/1115/commits/3388ef2cc181e973e049cacb0e0922d57d592a88

timemachine/fe/rbfe.py
```diff
@@ -579,8 +577,17 @@ def estimate_relative_free_energy_bisection(
...
         plots = make_pair_bar_plots(final_result, temperature, combined_prefix)
 
-        stored_frames = [np.array(frames[i]) for i in keep_idxs]
-        stored_boxes = [boxes[i] for i in keep_idxs]
+        assert len(results) == len(frames) == len(boxes)
+        stored_frames = []
+        stored_boxes = []
+        for i in keep_idxs:
+            try:
+                stored_frames.append(np.array(frames[i]))
```
badisa, line 585: Does this convert the StoredFrames to np.array? Looks like we were doing this previously, but means we could run out of memory trying to get all the frames.

mcwitt: Yes - so we definitely wouldn't want to set `keep_idxs=range(n_windows)`, at least for typical system sizes. I don't think the code here has ever handled this in a robust way. Should probably be a task for deboggling?

badisa: Sounds good, just had thought we were returning StoredFrames in SimulationResults.

timemachine/md/moves.py
```diff
@@ -16,16 +16,18 @@
...
     n_proposed: int = 0
     n_accepted: int = 0
 
-    def propose(self, x: CoordsVelBox) -> Tuple[CoordsVelBox, float]:
+    def propose(self, x: _State) -> Tuple[_State, float]:
         """return proposed state and log acceptance probability"""
         raise NotImplementedError
 
-    def move(self, x: CoordsVelBox) -> CoordsVelBox:
+    def move(self, x: _State) -> _State:
```
maxentile, line 30: Nice -- this is useful.

In addition to the motivating example of HREX, other cases that indicate "state = (x,v,b)" was prematurely specialized include:
* cases where the state omits velocities (instantaneous displacement moves)
* cases where the move cannot update box (NVT)
* cases where the move is allowed to update other auxiliary variables (`(x, box, selection_mask)`, `(x, v, box, selection_mask)` , etc. in localized moves)
* cases where the move only updates a subset of coordinates

maxentile (approved):

PR #1117, opened 2023-08-10
Small adjustments in test_nblist.py

maxentile:
Random inconsequential clean-up in tests, encountered while measuring some characteristics of the spatial sort.

* Move 3x repeated `coords -> non_negative_coords` logic from individual tests into common `hilbert_sort` fxn
* Implement `coords -> non_negative_coords` by subtracting minimum element, rather than adding argmin index
* Name and use HILBERT_GRID_DIM = 128 rather than 64, to match current implementation


badisa (approved):
Nice! Much cleaner

PR #1119, opened 2023-08-11
Add set_params method to BoundPotential

mcwitt:
Allows updating parameters in GPU memory without reconstructing potential objects. Useful for efficiently evaluating a constant set of functional forms with many different parameter sets.


tests/test_potentials.py
```diff
@@ -37,6 +37,20 @@ def test_bound_potential_get_potential(harmonic_bond):
     assert unbound_impl is bound_impl.get_potential()
 
 
+def test_bound_potential_set_params(harmonic_bond):
```
maxentile, line 40: Test expansion request: To increase peace of mind for future changes, can we increase coverage here?

(Expansion along 2 axes: (1) Cover more Potential classes and signatures (esp. corner cases in Nonbonded such as changing exception parameters, and SummedPotential such as composition with `.call_with_params_list`). (2) Test on more random instances -- may be faster to assert self-consistency with unbound potential (to avoid calling `.to_gpu(np.float32)` for every test instance).)

maxentile (approved):
Nice -- this should be very useful in similar situations to OpenMM's setParametersInContext methods.

(Compared with the two alternative approaches currently possible in timemachine: Reduces copy overhead associated with calling unbound_potentials repeatedly with the same parameters, and avoids constructor overhead associated with creating a new bound_potential for every update to the parameters.)

timemachine/cpp/src/bound_potential.cu
```diff
@@ -58,12 +55,20 @@ void BoundPotential::execute_host(
     }
 };
 
+void BoundPotential::set_params(const std::vector<double> params) {
+    this->size = params.size();
+    if (this->size > 0) {
```
badisa, line 60: This should call down to `set_params_device` for consistency with our other APIs and should have the same behavior of not allowing you to resize/reallocate the underlying parameters buffer (We removed this option in https://github.com/proteneer/timemachine/pull/1063)

timemachine/cpp/src/bound_potential.hpp
```diff
@@ -11,12 +11,13 @@ namespace timemachine {
 // a potential bounded to a set of parameters with some shape
 struct BoundPotential {
 
-    BoundPotential(std::shared_ptr<Potential> potential, const int size, const double *h_p);
+    BoundPotential(std::shared_ptr<Potential> potential, const std::vector<double> params);
```
proteneer, line 14: `const std::vector<double> &params` (note the missing &)

timemachine/cpp/src/bound_potential.hpp
```diff
@@ -11,12 +11,13 @@ namespace timemachine {
...
 
     int size;
     std::unique_ptr<DeviceBuffer<double>> d_p;
     std::shared_ptr<Potential> potential;
-    const int max_size_;
+
+    void set_params(const std::vector<double> params);
```
proteneer, line 20: const std::vector<double> &params (note the missing &)

timemachine/cpp/src/bound_potential.hpp
proteneer: (this is mostly a C++ convention, if you're passing in a constant vector, just do it by reference to avoid a potential copy)

timemachine/cpp/src/bound_potential.cu
badisa: Simply so that calling `set_params_device` and `set_params` has identical behavior. Which is consistent with our other APIs that have a device and host method (https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/nonbonded_all_pairs.cu#L106-L114) that way it has exactly the same behavior called from device code or host code.

mcwitt: OK. Consistent behavior for `set_params` and `set_params_device` makes sense to me. Was just confused by

> This should call down to set_params_device

mcwitt: Changed the behavior to allocate a static buffer for the lifetime of the `BoundPotential` object and raise an exception if `set_params` is called with parameters larger than the buffer in https://github.com/proteneer/timemachine/pull/1119/commits/c0eaa8de677f2f60526fd25856b6d03dd85deeb9.

timemachine/cpp/src/bound_potential.hpp
mcwitt: Changed to pass by reference in https://github.com/proteneer/timemachine/pull/1119/commits/96a22f13e34205baea99ae5a31d769897ecb1416

timemachine/cpp/src/bound_potential.hpp
mcwitt: Fixed in https://github.com/proteneer/timemachine/pull/1119/commits/96a22f13e34205baea99ae5a31d769897ecb1416

tests/test_potentials.py
mcwitt: > esp. corner cases in Nonbonded such as changing exception parameters

Could you clarify what you mean here? (assuming you mean exclusion parameters, these are attributes of the underlying `Potential` objects and are not settable using `set_params`)

timemachine/cpp/src/bound_potential.cu
```diff
@@ -58,15 +55,28 @@ void BoundPotential::execute_host(
...
 };
 
+void BoundPotential::set_params(const std::vector<double> &params) {
+    if (params.size() > 0) {
+        if (params.size() * sizeof(*d_p.data) > d_p.size) {
+            throw std::runtime_error(
+                "parameter size is greater than device buffer size: " + std::to_string(params.size()) + " > " +
+                std::to_string(d_p.size / sizeof(*d_p.data)));
+        }
+        // NOTE: can't use d_p.copy_from here because new parameters size may be less than buffer size
```
proteneer, line 65: Shouldn't they always be the same size?

mcwitt: Not necessarily, since we currently allow the updated parameters to be <= the buffer size. This is the behavior of `set_params_device` (which I assume is useful for Local MD), and I kept the same behavior for `set_params` for consistency.

tests/test_potentials.py
maxentile: >Could you clarify what you mean here? (assuming you mean exclusion parameters, these are attributes of the underlying Potential objects and are not settable using set_params)

Ahh, sorry -- yes, meant exclusion parameters, but got confused about whether these would be settable or not. This comment is not actionable, since exclusion_scales wouldn't be settable using set_params.

mcwitt: OK, thanks - I think the other suggestions for strengthening the tests in the original comment are actionable, though.

timemachine/cpp/src/bound_potential.cu
proteneer: Hm - outside of local MD I'm not sure why this would be allowed, the vast majority of the time this would likely indicate a user error?

timemachine/cpp/src/bound_potential.cu
```diff
@@ -58,15 +55,28 @@ void BoundPotential::execute_host(
...
+        gpuErrchk(cudaMemcpyAsync(d_p.data, params.data(), params.size() * sizeof(*d_p.data), cudaMemcpyHostToDevice));
```
proteneer, line 66: Also - why is this calling the Async version?

timemachine/cpp/src/bound_potential.cu
badisa: It is useful for Local MD, though I was probably premature in recommending the exact same behavior as the device method. As when we do it in Local MD we also modify the underlying potential (https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/local_md_potentials.cu#L246-L247) to match the shift in parameter size. Perhaps we should simply have `set_params` only allow the size to match exactly, which avoids an intermediate buffer which might allow us to speed up minimization.

timemachine/cpp/src/bound_potential.cu
```diff
@@ -58,15 +55,28 @@ void BoundPotential::execute_host(
...
+    }
+    this->size = params.size();
+}
+
 void BoundPotential::set_params_device(const int new_size, const double *d_new_params, const cudaStream_t stream) {
     if (new_size > 0) {
-        if (new_size > max_size_) {
+        if (new_size * sizeof(*d_p.data) > d_p.size) {
```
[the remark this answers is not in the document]
badisa: This has been on my mind to do in a deboggling pass, would like `size` to match with the `vector().size()` convention. 

I would say put back `max_size_`, for now, as it is easier to think about than multiplying/dividing by `sizeof(*d_p.data)`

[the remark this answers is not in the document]
mcwitt: Reverted in https://github.com/proteneer/timemachine/pull/1119/commits/3e9edb332b269ce989e086cb60b3da9f97719d29.

timemachine/cpp/src/bound_potential.cu
mcwitt: Changed to require `set_params` argument to have the same size as the buffer in https://github.com/proteneer/timemachine/pull/1119/commits/4f75acadfd7324d4416d6efac0ad1d59c3b8a8de.

timemachine/cpp/src/bound_potential.cu
mcwitt: Fixed as a side-effect of https://github.com/proteneer/timemachine/pull/1119/commits/4f75acadfd7324d4416d6efac0ad1d59c3b8a8de

timemachine/cpp/src/bound_potential.cu
```diff
@@ -3,12 +3,9 @@
...
 namespace timemachine {
 
-BoundPotential::BoundPotential(std::shared_ptr<Potential> potential, const int size, const double *h_p)
-    : size(size), d_p(nullptr), potential(potential), max_size_(size) {
-    if (this->size > 0) {
-        d_p.reset(new DeviceBuffer<double>(this->size));
-        d_p->copy_from(h_p);
-    }
+BoundPotential::BoundPotential(std::shared_ptr<Potential> potential, const std::vector<double> &params)
+    : size(params.size()), buffer_size_(size), d_p(buffer_size_), potential(potential) {
```
badisa, line 7: Isn't it invalid to construct a DeviceBuffer with size 0? Isn't that why previously this was a `unique_pointer<DeviceBuffer>`?

timemachine/cpp/src/bound_potential.cu
```diff
@@ -58,15 +55,27 @@ void BoundPotential::execute_host(
     }
 };
 
+void BoundPotential::set_params(const std::vector<double> &params) {
+    if (params.size() > 0) {
```
badisa, line 59: For the host code, don't think we want to preserve this logic of handling 0 like this. Instead, it should just do the check of `params.size() != buffer_size`.

tests/test_potentials.py
mcwitt: Thinking about this more, testing with different potentials (even `SummedPotential`) feels redundant, as `set_params` (and `BoundPotential` in general) is agnostic to / polymorphic in `potential`. Will think about it a bit more in case I'm missing something.

timemachine/cpp/src/bound_potential.cu
mcwitt: Hmm, what is the use case for passing empty parameters here? IIRC the restriction against initializing `DeviceBuffer(0)` is artificial and it could just be a no-op?

timemachine/cpp/src/bound_potential.cu
mcwitt: Unless there is a reason to pass empty parameters, it seems like this branch could be removed in both the host and device code. For device code, it avoids a trivial kernel launch, but I'm not sure we would ever trigger this intentionally anyway?

timemachine/cpp/src/bound_potential.cu
mcwitt: Removed the `size > 0` assertion from `DeviceBuffer` in https://github.com/proteneer/timemachine/pull/1119/commits/1c276be74ba253c8a11e75cc1bf245d3b3d305e8. Admittedly I'm not totally sure why we have it, but removing this simplifies `BoundPotential` and possibly other code as well where we allow passing empty arrays. (This also feels more consistent with other array/buffer APIs)

badisa: > Hmm, what is the use case for passing empty parameters here? IIRC the restriction against initializing `DeviceBuffer(0)` is ad-hoc and it could just be a no-op?

I don't have a clear use case, but the old code supported it and didn't want to remove support for zero length params unless it is clear that we don't require it. @proteneer is there a potential we might want that has no parameters?

proteneer: It's certainly plausible to have potentials that are devoid of parameters, but we wouldn't need to call `set_params` on them anyways.

badisa: Could we add a test for constructing a BoundPotential with zero length params? Just to make sure this edge case is covered by our tests

badisa: The `CentroidRestraint` is actually a potential we have that doesn't use any parameters

tests/test_potentials.py
maxentile: >Thinking about this more, testing with different potentials (even SummedPotential) feels redundant, as set_params (and BoundPotential in general) is agnostic to / polymorphic in potential. Will think about it a bit more in case I'm missing something.

I don't think it would expose any errors in the current PR, but increased coverage would just increase peace of mind for future changes.

(For example, I could imagine possible complications in how this function interacts with foreseeable optimizations in the nonbonded implementation, and these complications wouldn't necessarily show up in testing the harmonic bond implementation. If (in a future optimization) the neighborlist were computed in a way that was sensitive to both `coords` and `w_coords`, then using `set_params` to update `w_coords` should invalidate the neighborlist. If the test suite exercised only updating a harmonic bond potential's parameters but didn't exercise updating a nonbonded potential's parameters, we might miss if, say, https://github.com/proteneer/timemachine/blob/aecade69b90e6c13aaa82855a50e6210abd724db/timemachine/cpp/src/nonbonded_all_pairs.cu#L201-L208 didn't handle this corner case correctly.)

timemachine/cpp/src/bound_potential.cu
mcwitt: Added a test verifying that the `BoundPotential::BoundPotential` and `BoundPotential::set_params` work as expected when the parameters are an empty array.

tests/test_potentials.py
mcwitt: Would it be OK open a separate issue with the above example for follow up?

It seems like any issue that could be exposed by calling `set_params` on a `BoundPotential` would also be observable when calling the unbound potential successively with different parameters. I think adding tests along these lines would be useful, but I would tend to prefer them to be independent of `BoundPotential`.

proteneer (approved):

badisa (approved):

tests/test_potentials.py
maxentile: Migrated to https://github.com/proteneer/timemachine/issues/1120 !

(Not intended to be blocking for this PR -- looks good to me as-is.)

PR #1123, opened 2023-08-15
Water sampling improvements

proteneer:
This PR modifies the existing `ExchangeMove` class to bias deletions using the water's interaction energy, and implements a targetted `TIBDExchangeMove` class to also target insertions in a region of interest (in addition to biasing the deletions). This method preserves water distributions in bulk:

image
image

In addition, for the case of tricky charged buckyballs, it's capable of sampling N=1 and N=2 occupancies overnight (compared to 2+ weeks before):

MicrosoftTeams-image (19)


timemachine/potentials/nonbonded.py
```diff
@@ -60,6 +60,61 @@ def direct_space_pme(dij, qij, beta):
     return qij * erfc(beta * dij) / dij
 
 
+def nonbonded_block_unsummed(xi, xj, box, params_i, params_j, beta, cutoff):
```
proteneer, line 63: This was just me being lazy - I will deduplicate this later.

proteneer:
(CI is failing until I rebase from master)

proteneer:
@maxentile I think this is ready for review - I'll take care of rebasing off of @mcwitt 's HREX code in a second pass.

timemachine/exchange/exchange_mover.py
```diff
@@ -61,12 +63,15 @@ def build_system(host_pdbfile: str, water_ff: str, padding: float):
...
 def randomly_rotate_and_translate(coords, offset):
+    # coords: shape 3(OHH) x 3(xyz), water coordinates
     centroid = np.mean(coords, axis=0, keepdims=True)
     centered_coords = coords - centroid
     rot_mat = special_ortho_group.rvs(3)
-    rotated_coords = np.matmul(centered_coords, rot_mat)
+
+    # equivalent to standard form where R is first:
+    # (R @ A.T).T = A @ R.T
+    rotated_coords = centered_coords @ rot_mat.T
```
maxentile, line 74: could shrink slightly using [scipy spatial Rotation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html), something like

```python
rotation = Rotation.random()
rotation.apply(centered_coords)
```

depending on how much reminder of low-level details we want in this function

timemachine/exchange/exchange_mover.py
```diff
@@ -61,12 +63,15 @@ def build_system(host_pdbfile: str, water_ff: str, padding: float):
...
     return rotated_coords + offset
```
maxentile, line 75: question: should this be `+ offset + centroid`?

(aside: may be good to break this into a simpler function `randomly_rotate(coords)`, rather than passing through `randomly_rotate_and_translate(coords, offset)`)

timemachine/exchange/exchange_mover.py
```diff
@@ -61,12 +63,15 @@ def build_system(host_pdbfile: str, water_ff: str, padding: float):
     return solvated_host_system, solvated_host_coords, box, solvated_topology, nwa
 
 
-# currently un-used.
 def randomly_rotate_and_translate(coords, offset):
+    # coords: shape 3(OHH) x 3(xyz), water coordinates
     centroid = np.mean(coords, axis=0, keepdims=True)
     centered_coords = coords - centroid
```
maxentile, line 69: point of uncertainty: sensitivity of MC acceptance rate to choice of pivot

(rotate around oxygen, vs. rotate around centroid, vs. rotate around center of mass)

timemachine/exchange/exchange_mover.py
```diff
@@ -87,108 +95,385 @@ def get_initial_state(water_pdb, mol, ff, seed, nb_cutoff):
...
+        group_idxs = get_group_indices(bond_list, len(host_masses))
+        barostat_interval = 25
+        barostat = MonteCarloBarostat(
+            len(host_masses), DEFAULT_PRESSURE, temperature, group_idxs, barostat_interval, seed + 1
+        )
+
+        ligand_idxs = [0, 1, 2]  # pick first water molecule to be a "ligand" for targetting purposes
+        initial_state = InitialState(
+            host_bps, integrator, barostat, solvent_conf, np.zeros_like(solvent_conf), solvent_box, 0.0, ligand_idxs
+        )
```
maxentile, line 148: nit: clarity -- some partially redundant definitions before, after, and within this branch, might move common definitions outside of branch where possible.
* within branch: define potentials, masses,  `ligand_idxs`
* after branch: construct integrator, barostat, initial state
* remove: second definition of `barostat_interval = 25`

timemachine/exchange/exchange_mover.py
```diff
@@ -87,108 +95,385 @@ def get_initial_state(water_pdb, mol, ff, seed, nb_cutoff):
...
 
     return initial_state, num_water_mols, solvent_topology
 
 
-class ExchangeMove(moves.MonteCarloMove):
-    def __init__(self, nb_params, nb_beta, nb_cutoff, water_idxs):
+from scipy.special import logsumexp
+
+
+class BDExchangeMove(moves.MonteCarloMove):
```
maxentile, line 158: nit: expand new acronym "biased deletion" in docstring

timemachine/exchange/exchange_mover.py
```diff
@@ -87,108 +95,385 @@ def get_initial_state(water_pdb, mol, ff, seed, nb_cutoff):
...
+    def __init__(
+        self,
+        nb_beta: float,
+        nb_cutoff: float,
+        nb_params: NDArray,
+        water_idxs: NDArray,
+        beta: float,
+    ):
```
maxentile, line 166: question: would it make sense to parameterize the **amount of biasing**?

i.e. expose a constructor parameter that when set to 0.0 gives "unbiased" deletion, 1.0 gives the current behavior, and other real values give different behavior. might be interesting to see a curve of MC performance vs. this parameter, and see if it's roughly maximized at 1.0.

timemachine/exchange/exchange_mover.py
```diff
@@ -87,108 +95,385 @@ def get_initial_state(water_pdb, mol, ff, seed, nb_cutoff):
...
+# numpy version of delta_r
+def delta_r_np(ri, rj, box):
+    diff = ri - rj  # this can be either N,N,3 or B,3
+    if box is not None:
+        box_diag = np.diag(box)
+        diff -= box_diag * np.floor(diff / box_diag + 0.5)
+    return diff
+
+
+def v1_insertion(radius, center, box):
```
maxentile, line 304: nit: name, perhaps `vol1_insertion`, `vol2_insertion` or `sample_within_sphere`, `sample_outside_sphere`?

(initially misread as "version 1 insertion", "version 2 insertion")

timemachine/exchange/exchange_mover.py
```diff
@@ -87,108 +95,385 @@ def get_initial_state(water_pdb, mol, ff, seed, nb_cutoff):
...
+    c = np.cbrt(np.random.rand())
+    new_xyz = xyz * c * radius + center
+    assert np.linalg.norm(delta_r_np(new_xyz, center, box)) < radius
+
+    return new_xyz
+
+
+def v2_insertion(radius, center, box):
+    # generate random proposals outside of the sphere but inside the box
+    while True:
```
maxentile, line 321: preference: avoid possible infinite loop if radius is large relative to box -- perhaps a for loop with an error if this doesn't succeed after a large number of tries?

timemachine/exchange/exchange_mover.py
```diff
@@ -87,108 +95,385 @@ def get_initial_state(water_pdb, mol, ff, seed, nb_cutoff):
...
+        water_idxs: NDArray (W,3)
+            W is the total number of water molecules in the system.
+            Each element is a 3-tuple denoting the index for oxygen, hydrogen, hydrogen (or whatever is
+            consistent with the nb_params)
+
+        ligand_idxs: List[int]
+            Indices corresponding to the atoms that should be used to compute the centroid
+
+        beta: float
+            Thermodynamic beta, 1/kT (in kJ/mol)
```
maxentile, line 369: nit: `(kJ/mol)^-1`

(or just accept temperature, in units kelvin, and compute beta internally... also avoids name collision with nb_beta)

timemachine/exchange/exchange_mover.py
```diff
@@ -87,108 +95,385 @@ def get_initial_state(water_pdb, mol, ff, seed, nb_cutoff):
...
+        self.ligand_idxs = np.array(ligand_idxs)  # used to determine center of sphere
+        self.radius = radius
+
+    def get_water_groups(self, coords, box, center):
+        """
+        Partition water molecules into two groups, depending if it's inside the sphere or outside the sphere.
+        """
+        dijs = np.linalg.norm(delta_r_np(np.mean(coords[self.water_idxs_np], axis=1), center, box), axis=1)
+        v1_mols = np.argwhere(dijs < self.radius).reshape(-1)
+        v2_mols = np.argwhere(dijs >= self.radius).reshape(-1)
```
maxentile, line 386: nit: readability: move `np.mean` onto its own line `mol_centroids = np.mean(coords[self.water_idxs_np], axis=1)`

(was easy to misread as applying distance test to each atom, rather than each molecule.)

timemachine/exchange/exchange_mover.py
```diff
@@ -87,108 +95,385 @@ def get_initial_state(water_pdb, mol, ff, seed, nb_cutoff):
...
 
-        # this has a higher acceptance probability than if we allowed for rotations
-        # (probably because we have a much higher chance of a useless move)
-        moved_coords = randomly_translate(trial_chosen_coords, trial_translation)
-        trial_coords = coords.copy()  # can optimize this later if needed
-        trial_coords[chosen_water_atoms] = moved_coords
+        moved_coords = randomly_rotate_and_translate(trial_chosen_coords, trial_translation)
+        # moved_coords = randomly_translate(trial_chosen_coords, trial_translation)
+
+        # optimized version using double transposition
```
maxentile, line 279: doc request: add description of "the transposition trick" and "double transposition" in PR or code?

timemachine/exchange/exchange_mover.py
```diff
@@ -87,108 +95,385 @@ def get_initial_state(water_pdb, mol, ff, seed, nb_cutoff):
...
+        log_weights_after, trial_coords = self.batch_log_weights_incremental(
+            coords, box, chosen_water, moved_coords, log_weights_before
+        )
+
+        # reference version
+        # trial_coords = coords.copy()  # can optimize this later if needed
+        # trial_coords[chosen_water_atoms] = moved_coords
+        # log_weights_after = self.batch_log_weights(trial_coords, box)
```
maxentile, line 287: test request: move this from comment into `test_water_exchange.py` or similar

maxentile (approved):
Looks great!

Mostly nits in first pass: only thing that looked like a possible bug was handling of computed centroid vs. input offset in `randomly_rotate_and_translate`. (Might impact acceptance rate.)

Reviewing made me curious about possibly varying the biasing strength in the future.

timemachine/exchange/exchange_mover.py
proteneer: The intent is probably not well-documented. Like `randomly_translate`, the intent is to set the *centroid* of the displaced water to the location of "offset", a better name should be `new_loc` as opposed to `offset`.

proteneer: (will clarify in documentation and rename `offset`->`new_loc`)

timemachine/exchange/exchange_mover.py
proteneer: I'm going to keep it as-is for now, since I'll be porting this function directly to C++ (where I care about low-level details).

timemachine/exchange/exchange_mover.py
proteneer: That's a good question - I haven't investigated this yet. Mainly for simplicity, since keeping it at 1.0 gives you the exact cancellation with the Metropolis component. FWIW, most of the time, the same one or two waters are selected (since the water energies follow a gaussian-like distribution), so the high energy tail is typically only occupied by a very few number of waters.

timemachine/exchange/exchange_mover.py
proteneer: I'll try and clarify this - TL;DR allows for computation of the denominator (comprised of summed exponentials) to be done significantly faster by computing only 3 rows and transposing it.

timemachine/exchange/exchange_mover.py
proteneer: will do!

timemachine/exchange/exchange_mover.py
proteneer: willfix

timemachine/exchange/exchange_mover.py
proteneer: good idea - willfix.

timemachine/exchange/exchange_mover.py
proteneer: willfix

timemachine/exchange/exchange_mover.py
proteneer: willfix

timemachine/exchange/exchange_mover.py
proteneer: not sure how much this matters in practice

timemachine/exchange/exchange_mover.py
proteneer: willfix

timemachine/exchange/exchange_mover.py
proteneer: renamed to `inner_insertion` and `outer_insertion`

proteneer: (also renamed `v1_mols` to `inner_mols` and `v2_mols` to `outer_mols`)

timemachine/exchange/exchange_mover.py
maxentile: Would guess not very sensitive, but I don't know -- I don't remember if there was a big difference between "translate only" vs. "translate + randomly rotate"? If so, might also be sensitive to little details like this of how exactly the proposal is prepared.

I guess, if uncertain, we could also randomize the choice of pivot (weighted centroid, random weights).

proteneer: For biased deletions you need to randomly rotate, this is because you're only picking 1 or 2 unique waters with fixed orientations. In order to successfully insert into the buckyball, it's unlikely the original orientation is "aligned" with whatever environment is in the buckyball.

timemachine/exchange/exchange_mover.py
proteneer: Willfix

timemachine/exchange/exchange_mover.py
maxentile: Ahh, thanks for clarifying this -- makes sense!

maxentile (approved):

PR #1125, opened 2023-08-15
Use StoredArrays for frames in SimulationResult

mcwitt:
We currently load stored frames into memory to construct `SimulationResult` objects. This is unnecessary; loading can be deferred until we're ready to save the frames to disk or analyze them; at that point, we can process individual chunks separately to avoid using too much memory.

This PR:

* Modifies `SimulationResult` to store frames as `StoredArrays` instead of `NDArray`, avoiding the need to load the frames into memory to construct the result object
* Makes some minor modifications to make the interface of `run_sims_sequential` more consistent with `run_sims_bisection`
* Adds support for pickling `StoredArrays` objects. This allows existing code that pickles `SimulationResult` objects (e.g. `examples/rbfe_edge_list.py`) to continue working without changes.

### DIFF
--- a/timemachine/fe/free_energy.py
+++ b/timemachine/fe/free_energy.py
@@ -135,7 +135,7 @@ class PairBarResult:
 class SimulationResult:
     final_result: PairBarResult
     plots: PairBarPlots
-    frames: List[NDArray]  # (len(keep_idxs), n_frames, N, 3)
+    frames: List[StoredArrays]  # (len(keep_idxs), n_frames, N, 3)
     boxes: List[NDArray]
     md_params: MDParams
     intermediate_results: List[PairBarResult]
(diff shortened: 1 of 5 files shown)


timemachine/fe/free_energy.py
```diff
@@ -557,7 +557,7 @@ def run_sims_sequential(
     for lamb_idx, initial_state in enumerate(initial_states):
 
         # run simulation
-        cur_frames, cur_boxes = sample(initial_state, md_params)
+        cur_frames, cur_boxes = sample(initial_state, md_params, max_buffer_frames=100)
```
mcwitt, line 560: Same setting of `max_buffer_frames` we use for `run_sims_bisection`; prevents OOM for large systems.

badisa (approved):
🚀

jkausrelay (approved):

PR #1126, opened 2023-08-16
Adds test for HilbertSort class

badisa:
* Basic test that verifies that the distance between the particles within a block are reduced by sorting
* Modifies tests/common.py::hilbert_sort to image particles the same way that the C++ does for consistency, rather than just putting things above zero


tests/common.py
```diff
@@ -155,11 +155,12 @@ def prepare_nb_system(
     return params, potential
 
 
-def hilbert_sort(conf, D):
-    hc = HilbertCurve(HILBERT_GRID_DIM, D)
+def hilbert_sort(conf, box):
```
badisa, line 158: @maxentile I changed this to more closely match the behavior in the C++

https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/kernels/k_hilbert.cu#L33-L38

tests/test_hilbert_sort.py
```diff
@@ -0,0 +1,43 @@
...
+    distances = get_max_block_distances(coords, box, block_size)
+    unsorted_mean_dist = np.mean(distances)
+
+    sorter = custom_ops.HilbertSort(coords.shape[0])
+    perm = sorter.sort(coords, box)
+    sorted_coords = coords[perm]
+    sorted_dists = get_max_block_distances(sorted_coords, box, block_size)
+    sorted_mean_dist = np.mean(sorted_dists)
+    # The mean maximum distance within hilbert curve sorted blocks should be lower than the unsorted coords
+    assert sorted_mean_dist < unsorted_mean_dist
```
mcwitt, line 43: Could this be strengthened? At present it seems like this would pass 1/8 of the time even if the sorted and unsorted mean distances have the same distribution.

mcwitt (approved):

tests/test_hilbert_sort.py
badisa: Good question, I couldn't think of something obvious when I did this. Will take a look through our hilbert curve python library to see how they verify their code.

proteneer (approved):

tests/test_hilbert_sort.py
badisa: Looked through the tests here https://github.com/galtay/hilbertcurve/blob/main/tests/test_hilbertcurve.py nothing stood out. Actually makes me trust the library less, though we don't use it in practice. 

Any suggestions?

mcwitt: I guess I was just thinking to empirically choose a factor by which the hilbert-sorted distances should be smaller for this test. Testing that the code is generating a correct hilbert curve seems more difficult; don't immediately have good ideas for the latter.

badisa: Strengthened to be less than half the distance of the unsorted: https://github.com/proteneer/timemachine/pull/1126/commits/938610b38b13764b7875ebe38b74efb629be079b

PR #1127, opened 2023-08-16
Use set_params to speed up minimization

badisa:
* Made possible thanks to the addition of `set_params` in https://github.com/proteneer/timemachine/pull/1119
* Appears 3x faster to minimize (Reduces CI time by almost 20 minutes)
* Also avoid running needless minimization at the endpoints

### DIFF
--- a/timemachine/md/minimizer.py
+++ b/timemachine/md/minimizer.py
@@ -59,7 +59,7 @@ def parameterize_system(topo, ff: Forcefield, lamb: float) -> Tuple[List[Potenti
             lamb,
         ),
     ]
-    return [p for _, p in params_potential_pairs], [params for params, _ in params_potential_pairs]
+    return [pot for (_, pot) in params_potential_pairs], [params for (params, _) in params_potential_pairs]
 
 
 def flatten_params(params: List[NDArray]) -> NDArray:
--- a/timemachine/md/minimizer.py
+++ b/timemachine/md/minimizer.py
@@ -181,16 +181,17 @@ def minimize_host_4d(mols, host_config: HostConfig, ff, mol_coords=None) -> np.n
     x0 = combined_coords
     v0 = np.zeros_like(x0)
 
-    u_impl = summed_potential_from_potentials_and_params(*parameterize_system(hgt, ff, 1.0))
+    potentials, params = parameterize_system(hgt, ff, 1.0)
+    u_impl = summed_potential_from_potentials_and_params(potentials, params)
     bound_impls = [u_impl]
     x = fire_minimize(x0, bound_impls, box, 50)
 
+    # No need to reconstruct the context, just change the bound potential params. Allows
+    # for preserving the velocities between windows
+    ctxt = custom_ops.Context(x, v0, box, intg, bound_impls)
     for lamb in np.linspace(1.0, 0, 50, endpoint=False):
         _, params = parameterize_system(hgt, ff, lamb)
         u_impl.set_params(flatten_params(params))
-        # NOTE: we don't save velocities between trajectories at different lambda windows; empirically this seems to
-        # reduce the efficiency of the optimization, with more windows being required to achieve an equivalent result
-        ctxt = custom_ops.Context(x, v0, box, intg, bound_impls)
         xs, _ = ctxt.multiple_steps(50)
         x = xs[-1]
 
--- a/timemachine/md/minimizer.py
+++ b/timemachine/md/minimizer.py
@@ -66,7 +66,7 @@ def flatten_params(params: List[NDArray]) -> NDArray:
     return np.concatenate([p.reshape(-1) for p in params])
 
 
-def summed_potential_from_potentials_and_params(
+def summed_potential_bound_impl_from_potentials_and_params(
     potentials: List[Potential], params: List[NDArray]
 ) -> custom_ops.BoundPotential:
     flat_params = flatten_params(params)
@@ -182,7 +182,7 @@ def minimize_host_4d(mols, host_config: HostConfig, ff, mol_coords=None) -> np.n
     v0 = np.zeros_like(x0)
 
     potentials, params = parameterize_system(hgt, ff, 1.0)
-    u_impl = summed_potential_from_potentials_and_params(potentials, params)
+    u_impl = summed_potential_bound_impl_from_potentials_and_params(potentials, params)
     bound_impls = [u_impl]
     x = fire_minimize(x0, bound_impls, box, 50)
 
@@ -225,7 +225,7 @@ def make_host_du_dx_fxn(mols, host_config, ff, mol_coords=None):
 
     # bound impls of potentials @ lam=0 (fully coupled) endstate
     potentials, params = parameterize_system(hgt, ff, 0.0)
-    gpu_impl = summed_potential_from_potentials_and_params(potentials, params)
+    gpu_impl = summed_potential_bound_impl_from_potentials_and_params(potentials, params)
 
     # read conformers from mol_coords if given, or each mol's conf0 otherwise
     conf_list = [np.array(host_config.conf)]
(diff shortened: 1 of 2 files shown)


timemachine/md/minimizer.py
```diff
@@ -178,38 +181,30 @@ def minimize_host_4d(mols, host_config: HostConfig, ff, mol_coords=None) -> np.n
...
-        ctxt = custom_ops.Context(x, v0, box, intg, u_impls)
+    potentials, params = parameterize_system(hgt, ff, 1.0)
+    u_impl = summed_potential_from_potentials_and_params(potentials, params)
+    bound_impls = [u_impl]
+    x = fire_minimize(x0, bound_impls, box, 50)
+
+    # No need to reconstruct the context, just change the bound potential params. Allows
+    # for preserving the velocities between windows
+    ctxt = custom_ops.Context(x, v0, box, intg, bound_impls)
+    for lamb in np.linspace(1.0, 0, 50, endpoint=False):
```
badisa, line 192: Previously this had the endpoints included, don't think that is necessary since we minimize at each end, but could mistaken. And probably negligible compared to the speed up of using `set_params`

timemachine/md/minimizer.py
```diff
@@ -59,15 +59,18 @@ def parameterize_system(topo, ff: Forcefield, lamb: float):
...
-    pots = [pot for _, pot in params_potential_pairs]
-    params = [p for p, _ in params_potential_pairs]
-    flat_params = np.concatenate([p.reshape(-1) for p in params])
-    u_impls = [SummedPotential(pots, params).bind(flat_params).to_gpu(precision=np.float32).bound_impl]
-    return u_impls
+def flatten_params(params: List[NDArray]) -> NDArray:
+    return np.concatenate([p.reshape(-1) for p in params])
+
+
+def summed_potential_from_potentials_and_params(
```
mcwitt, line 69: Could be another PR, but it seems like this would be nice as a general utility. Maybe it could be a class method like `SummedPotential.from_potentials_and_params` or `SummedPotential.from_bound_potentials`?

badisa: Will follow up with another PR, agree would be a nice utility to have.

timemachine/md/minimizer.py
badisa: Combined with the change to accumulate velocities this breaks the local MD test, though that threshold is pretty arbitrary. Put it back in as the performance didn't make a difference.

timemachine/md/minimizer.py
```diff
@@ -178,38 +181,30 @@ def minimize_host_4d(mols, host_config: HostConfig, ff, mol_coords=None) -> np.n
...
+        u_impl.set_params(flatten_params(params))
         xs, _ = ctxt.multiple_steps(50)
         x = xs[-1]
 
-    u_impls = bind_potentials(parameterize_system(hgt, ff, 0.0))
-    final_coords = fire_minimize(x, u_impls, box, 50)
-    for impl in u_impls:
+    _, params = parameterize_system(hgt, ff, 0.0)
+    u_impl.set_params(flatten_params(params))
+    final_coords = fire_minimize(x, bound_impls, box, 50)
```
maxentile: nit: should we expose lambda schedule (which happens to be length-50 and linear) and n_steps_per_lambda window (which happens to be 50)?

timemachine/md/minimizer.py
```diff
@@ -395,22 +389,23 @@ def equilibrate_host(
...
+    ctxt = custom_ops.Context(x0, v0, host_config.box, integrator, [u_impl], barostat)
 
-    ctxt.multiple_steps(n_steps)
+    xs, boxes = ctxt.multiple_steps(n_steps)
 
-    return ctxt.get_x_t(), ctxt.get_box()
+    return xs[-1], boxes[-1]
 
 
 def get_val_and_grad_fn(bps: Iterable[BoundPotential], box: NDArray, precision=np.float32):
```
maxentile, line 411: possibly later: consider consolidating `summed_potential_from_potentials_and_params`, `make_host_du_dx_fxn`, and `get_val_and_grad_fn` in this file.

(Maybe body of `get_val_and_grad_fn` can be slightly simplified using `summed_potential_from_potentials_and_params`.)

timemachine/md/minimizer.py
```diff
@@ -395,22 +389,23 @@ def equilibrate_host(
...
 
     # context components: positions, velocities, box, integrator, energy fxns
-    ctxt = custom_ops.Context(x0, v0, host_config.box, integrator, u_impls, barostat)
+    ctxt = custom_ops.Context(x0, v0, host_config.box, integrator, [u_impl], barostat)
 
-    ctxt.multiple_steps(n_steps)
+    xs, boxes = ctxt.multiple_steps(n_steps)
 
-    return ctxt.get_x_t(), ctxt.get_box()
+    return xs[-1], boxes[-1]
```
maxentile, line 408: nit, memory: can adjust `store_x_interval` so that `len(xs) == len(boxes) == 1` or so

timemachine/md/minimizer.py
```diff
@@ -59,15 +59,18 @@ def parameterize_system(topo, ff: Forcefield, lamb: float):
...
-    return u_impls
+def flatten_params(params: List[NDArray]) -> NDArray:
+    return np.concatenate([p.reshape(-1) for p in params])
+
+
+def summed_potential_from_potentials_and_params(
+    potentials: List[Potential], params: List[NDArray]
+) -> custom_ops.BoundPotential:
+    flat_params = flatten_params(params)
+    return SummedPotential(potentials, params).bind(flat_params).to_gpu(precision=np.float32).bound_impl
```
maxentile, line 73: nit: naming inconsistency -- rename to clunkier `summed_potential_bound_impl_from_potentials_and_params` or remove `.bound_impl` here?

timemachine/md/minimizer.py
```diff
@@ -59,15 +59,18 @@ def parameterize_system(topo, ff: Forcefield, lamb: float):
             lamb,
         ),
     ]
-    return params_potential_pairs
+    return [p for _, p in params_potential_pairs], [params for params, _ in params_potential_pairs]
```
maxentile, line 62: nit, readability: single-letter abbreviation `p` for words that both start with `p`

maybe 
```suggestion
    return [pot for (_, pot) in params_potential_pairs], [params for (params, _) in params_potential_pairs]
```
or
```suggestion
    potentials = [pot for (_, pot) in params_potential_pairs]
    params = [theta for (theta, _) in params_potential_pairs]
    return potentials, params
```

maxentile (approved):
Nice improvements and speed-up!

(Just nits.)

timemachine/md/minimizer.py
badisa: Are you asking for assertion? Because the default `store_x_interval` returns `len(xs) == len(boxes) == 1`

badisa: Not entirely clear what you meant, but added assertions here: https://github.com/proteneer/timemachine/pull/1127/commits/badefc34e319083df88473e63cef240ee1af98c3

timemachine/md/minimizer.py
badisa: Exposed number of windows and steps per window here: https://github.com/proteneer/timemachine/pull/1127/commits/a354028fb0b11276d18a8a35c2e3a1ab2fd17c22

timemachine/md/minimizer.py
badisa: Adopted the clunkier name to be explicit

timemachine/md/minimizer.py
maxentile: Oh, sorry, my mistake  -- wasn't asking for an assertion, but I had misremembered and thought the default store_x_interval would result in `len(xs) == n_steps`

timemachine/md/minimizer.py
badisa: Playing around with this a bit, `SummedPotential.from_bound_potentials` might be useful and `SummedPotential.from_potentials_and_params` would be identical to `SummedPotential(potentials, params)`.  With this function taking care of the additional need to flatten the params at this stage.

mcwitt: > SummedPotential.from_potentials_and_params would be identical to SummedPotential(potentials, params)

Ah, I guess the analog of the function here would return a `BoundPotential[SummedPotential]`, so may not make sense as a class method of `SummedPotential`. Would have to think about this more, but could imagine a class extending `BoundPotential[SummedPotential]` with a `from_potentials_and_params` class method and possibly other utilities for handling flattening/unflattening (in a future PR)

Issue #980, opened 2023-03-17
Consider unifying approach to PRNG state


maxentile:
Thanks to @mcwitt for thoughtful comments: migrating from https://github.com/proteneer/timemachine/pull/978#discussion_r1140414877 . Would be good to discuss and adopt project-wide conventions, if possible.

Some approaches currently used:
* some objects store and update their own random state (integrator, barostat, context, etc.)
* some objects have random functions that accept numpy rng's, jax rng key's, or integer random seeds
* some sites reference global numpy random state

Some possible trade-offs:
* ease of making application code deterministic
* ease of making certain kinds of mistakes
* explicitness
* simplicity

See also: https://jax.readthedocs.io/en/latest/jax-101/05-random-numbers.html

mcwitt:
Attempting to map out a decision tree:

1. Nondeterministic functions accept PRNG state. This is the approach taken by JAX.
   1. Use numpy API (`numpy.random.Generator`)
   1. Use JAX API (`jax.random.PRNGKey`)
1. Instances with nondeterministic methods maintain their own PRNG state. Seed or initial PRNG state is passed to the constructor.
1. Use global PRNG state. This is the approach taken by the original `numpy.random` API.

maxentile:
Thanks for mapping these out.

By default I lean towards (2) due to familiarity, imposing looser requirements (maybe one class uses cuRAND, a different class only uses numpy, ...), and since it seems harder to make certain kinds of errors (forgetting to split keys etc.). But (1) does seem cleaner.

An additional practice that may be compatible with all of the above options is to implement a random function `random_f(x)` by composing a deterministic function `f(x, gaussian_noise)` (whose implementation is RNG-agnostic) with random generation `gaussian_noise = rng.normal(0, 1, x.shape)` or `gaussian_noise = np.random.randn(*x.shape)` or ...

mcwitt:
> and since it seems harder to make certain kinds of errors (forgetting to split keys etc.)

That's a good point that forgetting to split keys would be a class of error unique to (1b) (added to "disadvantages" above).

maxentile:
> It seems like it does introduce some additional room for error, though. E.g. the caller must be careful to ensure that the input randomness has the expected distribution, and because generation is decoupled from transformation, it might be harder to keep in sync.

[...]

PR #1128, opened 2023-08-18
Hamiltonian replica exchange

mcwitt:
Adds a generic implementation of Hamiltonian replica exchange (HREX) and a specialized version for use with RBFE implemented using the former.

### Notes

* This implements a variant of nearest-neighbor HREX (described in [1][]) that performs many consecutive nearest-neighbor swap attempts per iteration to generate effectively uncorrelated permutations. The matrix of energies by replica and state is computed only once per iteration, so the cost is amortized over many neighbor swap attempts.
* We use the heuristic $K^3$ for number of neighbor swap attempts per iteration, where $K$ is the number of states. This is derived from the empirical bounds of $K^3$ to $K^5$ in [1][].
* This **does not** yet switch to using HREX as the default method for production RBFE calculations (i.e. `run_leg` functions).

### To do

- [x] Add benchmark comparing RBFE performance with and without HREX enabled
- [ ] Avoid modifying global PRNG state
- [x] Verify / tune added nightly tests

### Example plots

<details>

Test system: evenly-spaced overlapping Gaussians

transition_matrix
accept_rates_convergence
replica_state_counts
replica_state_counts_heatmap
replica_state_counts_convergence

</details>

[1]: http://dx.doi.org/10.1063/1.3660669


timemachine/md/moves.py
```diff
@@ -103,6 +108,31 @@ def n_proposed_by_move(self) -> List[int]:
         return [m._n_proposed for m in self.moves]
 
 
+class Choice(CompoundMove[_State]):
```
maxentile, line 111: nit: naming -- `MixtureOfMoves` or similar?

timemachine/md/moves.py
```diff
@@ -103,6 +108,31 @@ def n_proposed_by_move(self) -> List[int]:
...
+        self.moves = moves
+
+    def move(self, x: _State) -> _State:
+        idx = np.random.choice(len(self.moves))
+        chosen_move = self.moves[idx]
+        x = chosen_move.move(x)
+        return x
+
+
+class Sequential(CompoundMove[_State]):
```
maxentile, line 124: nit: naming -- `SequenceOfMoves` or similar?

timemachine/md/hrex.py
```diff
@@ -0,0 +1,206 @@
...
+        neighbor_pairs: Sequence[Tuple[StateIdx, StateIdx]],
+        log_q: Callable[[ReplicaIdx, StateIdx], float],
+        n_swap_attempts: int,
+    ) -> Tuple["Hrex[_Replica]", List[Tuple[int, int]]]:
+
+        # For each swap attempt, we uniformly choose between the neighboring pairs and an "identity" move that does
+        # nothing. The latter is included as a workaround for slow mixing when neighboring pairs have a near 100% swap
+        # acceptance rate.
+        identity: MonteCarloMove[List[ReplicaIdx]] = Identity()
+        move = Choice([NeighborSwapMove(log_q, s_a, s_b) for s_a, s_b in neighbor_pairs] + [identity])
```
maxentile, line 77: rather than mixing in this move, could we augment `neighbor_pairs` with `(i,i) for i in range(n)`?

maxentile (commented):
🚀 -- nice!

Only nits in this pass, will attempt some tire-kicking before completing review.

timemachine/md/hrex.py
mcwitt: I considered something like this, but wouldn't this mean that ~half of the swap attempts would be no-ops, even if the nearest-neighbor acceptance probabilities are all near 100%?

timemachine/md/moves.py
mcwitt: Nice, I like `MixtureOfMoves`. Will update.

mcwitt: Renamed in https://github.com/proteneer/timemachine/pull/1128/commits/543cac9b411362b37946bf85c0c251cc4e7d7711

timemachine/md/moves.py
mcwitt: Renamed in https://github.com/proteneer/timemachine/pull/1128/commits/543cac9b411362b37946bf85c0c251cc4e7d7711

timemachine/md/hrex.py
maxentile: >I considered something like this, but wouldn't this mean that ~half of the swap attempts would be no-ops, even if the nearest-neighbor acceptance probabilities are all near 100%?

Ahh, sorry, good point. That would be a large amount of overhead when `pairs` is "all pairs (i, j) where abs(i - j) == 1"

"no-op" swap attempts would be a smaller fraction of total swap attempts if `pairs` were something like "all pairs (i, j)", or "all pairs (i, j) where abs(i - j) < 10" or something. Possibly something to consider later, but the way you've implemented this for nearest neighbor swaps makes sense.

maxentile: Actually, sorry, I was over-complicating this. Including a single pair `(0,0)` should suffice.

mcwitt: > Actually, sorry, I was over-complicating this. Including a single pair (0,0) should suffice.

Nice, I like this suggestion. I was also just thinking that implementing the work-around here breaks the generality of this function, which is supposed to be agnostic to the shape of the graph (e.g. `neighbor_pairs` *could* be all pairs). Will update to remove the mixing with the identity move and just include `(0, 0)` in neighbor pairs in downstream code where we do assume a linear graph.

mcwitt: I suppose we could also just remove this, as it's only relevant when K=2, which is probably an uncommon scenario?

mcwitt: I opted to just remove this for now, since K=2 seems like an uncommon scenario, and even so I think I'd need more evidence that this is actually an issue. The slow mixing I'd observed that inspired adding this actually turned out to be due to a earlier bug in the code.

maxentile: >I suppose we could also just remove this, as it's only relevant when K=2, which is probably an uncommon scenario?

I'd be in favor of just adding `(0,0)` (if it's not already in the list of neighbor pairs), to ensure that the move is aperiodic even for K=2, and to avoid having to worry about whether the move might be periodic in some unchecked situations for K > 2.

Supporting K = 2 might be useful. (E.g. in cases where we currently jump in one direction from a softened state to a physical state, as in enhanced.py.)

maxentile: >I opted to just remove this for now, since K=2 seems like an uncommon scenario, and even so I think I'd need more evidence that this is actually an issue. The slow mixing I'd observed that inspired this actually turned out to be due to a bug in the code.

Fair enough! I think it should be aperiodic already for K > 2. It might currently be periodic in certain K=2 situations. (E.g. the case you mentioned where accept prob = 100%.)

(Might assert K > 2 or warn if K = 2.)

mcwitt: oops, sorry for the crossed wires! I went ahead with an intermediate solution I'm reasonably happy with of manually adding `(0, 0)` to the neighbor pairs in `fe.run_sims_hrex`, which is at a higher level of abstraction and constructs the neighbor pairs given a list of states.

For the lower-level functions, it's currently still up to the caller to ensure that `neighbor_pairs` is constructed in a way to avoid periodic updates when swap acceptance probabilities approach 100% (though could add a warning)

maxentile:
Test expansion request: can we add an HREX workload to the performance benchmarking script `test_benchmark.py`?

Not a full free energy calculation, but running a solvent leg and complex leg with a couple settings of n_windows (maybe 10, 20, 50) for about 30 seconds - 1 minute each? Would be helpful to keep an eye on relative costs of swap attempts vs. MD.

maxentile (approved):
Nice work -- looks good to me! Excited to see this integrated.

tests/test_relative_free_energy.py
```diff
@@ -169,6 +174,29 @@ def test_run_hif2a_test_system(estimate_relative_free_energy_fn):
...
+    "estimate_relative_free_energy_fn",
+    [
+        estimate_relative_free_energy,
+        estimate_relative_free_energy_bisection,
+        pytest.param(
+            estimate_relative_free_energy_bisection_hrex,
+            marks=pytest.mark.xfail(
+                reason="lambda window trajectories are not individually reproducible given InitialState due to mixing"
+            ),
+        ),
```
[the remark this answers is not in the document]
mcwitt: The latter; this test simply samples from each window individually and compares the resulting frames against the results returned by `estimate_relative_free_energy_fn`. For HREX, the results are expected to be different because configurations are (deterministically, given a fixed seed) mixed between states during sampling.

The XFAIL here would fail if there is actually no mixing and the results end up being the same between HREX and individually sampling the states.

timemachine/fe/plots.py
```diff
@@ -322,3 +322,83 @@ def plot_fwd_reverse_predictions(
     plot_png = buffer.read()
 
     return plot_png
+
+
+def plot_hrex_swap_acceptance_rates(swap_acceptance_rates: NDArray):
```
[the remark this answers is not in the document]
mcwitt: > One plot that may be "missing" is an all-pairs transition matrix.

By "all-pairs transition matrix" I'm assuming you mean a $K \times K$ matrix where the $(i,\,j)$ element is transition rate between state $i$ and state $j$? If so, I didn't think to include this because I'd been focused only on nearest-neighbor swaps in a 1-d chain, but agreed that this seems useful for more general situations. Will add.

tests/test_relative_free_energy.py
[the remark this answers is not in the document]
maxentile: 👍 , thanks for clarifying!

>The XFAIL here would fail if there is actually no mixing and the results end up being the same between HREX and individually sampling the states.

Nice -- that's an added bonus.

timemachine/fe/plots.py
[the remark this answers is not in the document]
maxentile: Yes, exactly. 

>I didn't think to include this because I'd been focused only on nearest-neighbor swaps in a 1-d chain, but agreed that this seems useful for more general situations. Will add.

If I'm understanding correctly, that would still be informative even if the "inner loop" swaps are nearest neighbor only. (Since, in the "outer loop" (K**4 inner loop steps), `i` can swap with `j` that's not immediately adjacent.)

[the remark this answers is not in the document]
mcwitt: > If I'm understanding correctly, that would still be informative even if the "inner loop" swaps are nearest neighbor only. (Since, in the "outer loop" (K**4 inner loop steps), i can swap with j that's not immediately adjacent.)

Ah, interesting. As it is we don't separately track the transition rates for the "outer loop" moves, and I don't think this is recoverable from the individual swap move accepted/proposed counts. I think this could be inferred from the replica-to-state mappings that we currently do track, though. Might need to think about this a little more.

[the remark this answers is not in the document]
mcwitt: Added the transition matrix for "outer loop", i.e. permutation moves to `HrexDiagnostics` and added a plot in https://github.com/proteneer/timemachine/pull/1128/commits/341c286d853c0d014955d8cc93d77525eac60feb.

Also added an estimate of the relaxation time computed from the second-largest eigenvalue to the diagnostics in https://github.com/proteneer/timemachine/pull/1128/commits/e43369992207c5cccbef1c97cc5cf65f9abd7b8b.

timemachine/fe/free_energy.py
```diff
@@ -725,3 +722,189 @@ def compute_intermediate_result(lambdas: Sequence[float]) -> PairBarResult:
...
+            frames, boxes = sample(
+                replace(initial_states[state_idx], x0=xb.coords, box0=xb.box),
+                replace(
+                    md_params,
+                    n_frames=n_frames_iter,
+                    n_eq_steps=0,
+                    seed=np.random.randint(np.iinfo(np.int32).max),
+                ),
+                max_buffer_frames=100,
+            )
```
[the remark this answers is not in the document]
mcwitt: Applied in https://github.com/proteneer/timemachine/pull/1128/commits/b4bc816bb179eacbeed6cee0dd6581079815c7ae

timemachine/fe/plots.py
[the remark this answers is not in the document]
maxentile: Sweet! These will be highly informative about how effectively we're sampling `p(state_permutations | xs)`.

Tangent 1: I don't expect this to be a bottleneck, but in case it is, I've started playing around a little with a possibly faster-mixing alternative to the uniform mixture of nearest-neighbor swap moves. (Apply [locally-balanced discrete MCMC](https://gist.github.com/maxentile/5497a1a87cf51fb132962ca26462dc0a) using some definition of a permutation's "neighborhood", e.g. considering in one proposal step [the set of all permutations that could have been accessed in T successive neighbor swap moves](https://gist.github.com/maxentile/9a90704b6df4f8cd2e255a033ee25e10).)

Tangent 2: In the future, we might consider going all the way to the "infinite swap limit." (An approach introduced by [Dupuis+ (2012)](https://epubs.siam.org/doi/abs/10.1137/110853145), reformulated and adapted to MD context by [Lu+ (2013)](https://arxiv.org/pdf/1212.0931.pdf) , [Plattner+ (2013)](https://pubs.acs.org/doi/abs/10.1021/ct400355g), …), and investigated briefly in the K=2 case last year by @proteneer .)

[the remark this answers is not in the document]
mcwitt: > Doc request: 1-liner descriptions of each plot.

Added in https://github.com/proteneer/timemachine/pull/1128/commits/be0ef78bddb064b667df991dd04e412451cde6de

timemachine/fe/free_energy.py
```diff
@@ -725,3 +722,184 @@ def compute_intermediate_result(lambdas: Sequence[float]) -> PairBarResult:
...
+    boxes_by_state_: List[List[NDArray]] = [[] for _ in lambdas]
+    for samples_by_state in samples_by_state_by_iter:
+        for samples, frames, boxes in zip(samples_by_state, frames_by_state, boxes_by_state_):
+            frames_i, boxes_i = samples
+            frames.extend(frames_i)
+            boxes.extend(boxes_i)
+
+    boxes_by_state: List[NDArray] = [np.array(boxes) for boxes in boxes_by_state_]
+
+    energy_decomposed_states = [
```
badisa, line 889: It looks like here you construct all the potentials in memory which could cause an OOM. Seems better to do it iteratively like we do in bisection

timemachine/fe/plots.py
```diff
@@ -322,3 +322,107 @@ def plot_fwd_reverse_predictions(
     plot_png = buffer.read()
 
     return plot_png
+
+
+def plot_hrex_transition_matrix(transition_rate: NDArray):
```
badisa, line 327: Consistency question: Should these return bytes like the other plotting functions in this file? No opinion really, but would be good to be consistent

timemachine/md/hrex.py
```diff
@@ -0,0 +1,247 @@
...
+        super().__init__()
+        self.log_q = log_q
+        self.s_a = s_a
+        self.s_b = s_b
+
+    def propose_with_log_q_diff(self, state: List[_Replica]) -> Tuple[List[_Replica], float]:
+        s_a = self.s_a
+        s_b = self.s_b
+        state_ = list(state)
+        state_[s_a], state_[s_b] = state_[s_b], state_[s_a]
```
badisa, line 28: Would be cleaner if not modifying the state of the object that state is coming from. This assumes that this swap is not sequential (load values s_b and s_a before assigning the values) which I assume it does, but non-obvious reading the code.

```suggestion
        state_[s_a], state_[s_b] = state[s_b], state[s_a]
```

timemachine/md/hrex.py
```diff
@@ -0,0 +1,247 @@
...
+    def relaxation_time(self):
+        return estimate_relaxation_time(self.transition_matrix)
+
+
+def get_swap_attempts_per_iter_heuristic(n_states: int) -> int:
+    """Heuristic for number of swap attempts per iteration derived from [1].
+
+    References
+    ----------
+    [1]: http://dx.doi.org/10.1063/1.3660669
```
badisa, line 161: Should we point at the arvix (https://arxiv.org/abs/1105.5749) rather than the doi? Not sure about this, just nice to be able to get the paper immediately without having to go through the process of getting the paper through the journal.

timemachine/md/hrex.py
```diff
@@ -0,0 +1,247 @@
...
+        n_swap_attempts_per_iter = get_swap_attempts_per_iter_heuristic(n_replicas)
+
+    hrex = Hrex.from_replicas(replicas)
+
+    samples_by_state_by_iter: List[List[_Samples]] = []
+    replica_idx_by_state_by_iter: List[List[ReplicaIdx]] = []
+    fraction_accepted_by_pair_by_iter: List[List[Tuple[int, int]]] = []
+
+    for n_samples_batch in batches(n_samples, n_samples_per_iter):
+        sample_replica_ = lambda replica, state_idx: sample_replica(replica, state_idx, n_samples_batch)
```
badisa, line 238: Is there somewhere that asserts that the number of samples returned by `sample_replica` equals `n_samples_batch`?

timemachine/md/moves.py
```diff
@@ -103,6 +96,31 @@ def n_proposed_by_move(self) -> List[int]:
...
 
 
+class MixtureOfMoves(CompoundMove[_State]):
+    """Apply a single move uniformly selected from a list"""
+
+    def __init__(self, moves: Sequence[MonteCarloMove[_State]]):
+        self.moves = moves
+
+    def move(self, x: _State) -> _State:
+        idx = np.random.choice(len(self.moves))
```
badisa, line 106: Seems preferable to be able to seed this

timemachine/fe/free_energy.py
```diff
@@ -725,3 +722,184 @@ def compute_intermediate_result(lambdas: Sequence[float]) -> PairBarResult:
...
+        box trajectories
+
+    HrexDiagnostics
+        HREX statistics (e.g. swap rates, replica-state distribution)
+    """
+
+    if n_swap_attempts_per_iter is None:
+        n_swap_attempts_per_iter = get_swap_attempts_per_iter_heuristic(len(initial_states))
+
+    np.random.seed(md_params.seed)
```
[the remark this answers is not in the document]
mcwitt: Agreed on the preference for no side effects. Unfortunately I think this is required for now because `timemachine.md.moves.MonteCarloMove` relies on the global numpy RNG state. This could potentially be refactored to accept a seed or generator argument, but this seemed like a task for a separate PR. We might track this in #980.

tests/test_relative_free_energy.py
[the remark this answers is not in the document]
badisa: A test that verifies that calling `HREX` is deterministic/bitwise deterministic is important, but can't be replicated with the logic that is in this test.

`xfail` seems confusing here, as we know this shouldn't work and there are presumably better ways to assert that we get swapping, such as checking that the acceptance rate is at least > 0. I agree, the time to run this test that we know will fail seems bad.

tests/hrex/test_hrex_alchemical.py
```diff
@@ -0,0 +1,73 @@
...
+)
+from timemachine.fe.rbfe import estimate_relative_free_energy_bisection_hrex
+from timemachine.ff import Forcefield
+from timemachine.md import builders
+from timemachine.testsystems.relative import get_hif2a_ligand_pair_single_topology
+
+
+@pytest.mark.nightly(reason="Slow")
+@pytest.mark.parametrize("host", [None, "complex", "solvent"])
+def test_hrex_rbfe_hif2a(host: Optional[str]):
```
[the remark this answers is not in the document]
mcwitt: oops, I hadn't actually run the nightly tests through to completion yet. Will check this and reduce the cost if it's unreasonable.

timemachine/fe/free_energy.py
```diff
@@ -725,3 +722,184 @@ def compute_intermediate_result(lambdas: Sequence[float]) -> PairBarResult:
...
+
+    lambdas = [s.lamb for s in initial_states]
+    replicas = [CoordsBox(s.x0, s.box0) for s in initial_states]
```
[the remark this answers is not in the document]
mcwitt: Hmm, not sure I understand. I don't think this line actually copies of the underlying numpy arrays?

timemachine/fe/free_energy.py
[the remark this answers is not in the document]
badisa: I have a preference for doing this now, as this just seems to risk changing behavior that we don't intend to. If is to be done in a separate PR I would say that this shouldn't be used in the `run_leg` methods until we remove the side effect.

timemachine/fe/free_energy.py
```diff
@@ -725,3 +722,184 @@ def compute_intermediate_result(lambdas: Sequence[float]) -> PairBarResult:
...
+
+    bps = initial_states[0].potentials  # TODO: assert initial states have compatible potentials?
+    sp = SummedPotential([bp.potential for bp in bps], [bp.params for bp in bps])
+    ubp = sp.to_gpu(np.float32)
+
+    def compute_energy_offsets(xbs: List[CoordsBox], params: List[NDArray]):
+        coords = np.array([xb.coords for xb in xbs])
+        boxes = np.array([xb.box for xb in xbs])
+        _, _, U = ubp.unbound_impl.execute_selective_batch(
+            np.array(coords), np.array(params), np.array(boxes), False, False, True
```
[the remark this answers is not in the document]
mcwitt: Good catch. Fixed in https://github.com/proteneer/timemachine/pull/1128/commits/142ecc2e18ce96c0b8019e7895bf9bff8d9c6149

timemachine/fe/free_energy.py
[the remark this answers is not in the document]
badisa: Was unclear, it's that later on you take a `List[CoordsBox] -> NDArray]`. But realizing now there isn't a way to combine the coords without a copy, ignore.

timemachine/fe/rbfe.py
```diff
@@ -476,7 +477,11 @@ def estimate_relative_free_energy_bisection(
     host_config: HostConfig or None
         Configuration for the host system. If None, then the vacuum leg is run.
 
-    prefix: str
+    md_params: MDParams, optional
+        Parameters for the equilibration and production MD. Defaults to 400 global steps per frame, 1000 frames and 10k
```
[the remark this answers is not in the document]
mcwitt: Agreed. This was just copied from `estimate_relative_free_energy`, but will just update both places.

timemachine/md/hrex.py
mcwitt: Swap assignments in Python are "simultaneous", e.g.

```python
>>> x = 1
>>> y = 2
>>> x, y = y, x
>>> x, y
(2, 1)
```
but will apply the change to disambiguate.

timemachine/fe/free_energy.py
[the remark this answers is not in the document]
maxentile: >Unfortunately I think this is required for now because timemachine.md.moves.MonteCarloMove relies on the global numpy RNG state. This could potentially be refactored to accept a seed or generator argument, but this seemed like a task for a separate PR.

The dependence on global numpy RNG state in `MonteCarloMove` was not well-considered and is not something I think we need to preserve.

Including a seed in the constructor `HREX(..., seed=1234)` ... `HREX::move(x) = ... behavior that depends on and updates self.rng state ...` may be one way to allow to seed the move without depending on / modifying global numpy RNG state and without refactoring / modifying signatures anywhere.

There is not currently a fully consistent convention in timemachine, and I think all 3 approaches are used in various places (global rng state, instance rng state, functions accept rng state as input).

timemachine/md/hrex.py
mcwitt: Not sure. I see precedent for both in the repo currently. For published papers it seemed appropriate to list the DOI, but it does make it harder to quickly get a PDF. @maxentile do you have an opinion here?

maxentile: No preference -- I don't think I've used a consistent style.

(When adding references to code I like to include [author, year] title - description of relevance, e.g. https://github.com/proteneer/timemachine/blob/3984fe517d92904986614515810ba392dc60f5a4/timemachine/md/smc.py#L148-L149 )

timemachine/fe/free_energy.py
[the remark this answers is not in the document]
mcwitt: > without refactoring / modifying signatures anywhere.

If I'm not mistaken, this would still require updating `MonteCarloMove` and subclasses, correct? (Unless we remove the inheritance for now and just have `NeighborSwapMove` reimplement `move`)

timemachine/md/hrex.py
mcwitt: Added arxiv URL additionally in https://github.com/proteneer/timemachine/pull/1128/commits/df187e38c69505b79471a5730fd0037ce509ea2b

tests/test_relative_free_energy.py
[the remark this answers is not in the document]
mcwitt: > A test that verifies that calling HREX is deterministic/bitwise deterministic is important

Yes, agreed we should have a test for bitwise determinism of HREX. Sorry, was stuck on thinking of something with the structure of this test, which verifies that the results of running individual simulations starting from each initial state match those returned by the `estimate_relative_free_energy` function. Something like this doesn't make sense for HREX, but we should still test end-to-end determinism of HREX, e.g. by running it twice in succession with the same seed.

timemachine/md/hrex.py
mcwitt: Not currently. I'm not sure that makes sense to add here, since this function is generic in `_Samples` (e.g. it could be a list or a tuple of lists, as is the case in RBFE simulations), so it doesn't know how to query the length of the returned samples.

timemachine/md/moves.py
mcwitt: Will address along with https://github.com/proteneer/timemachine/pull/1128#discussion_r1310314713

timemachine/fe/free_energy.py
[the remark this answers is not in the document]
maxentile: Ahh, sorry, yes -- this would require changing a few more places than I realized.

Would require:
* modifying signatures and implementations of constructors (to accept `seed=1234` and add `self.rng = np.random.default_rng(seed)`)
* modifying the use of the unif(0,1) r.v. for acceptance test https://github.com/proteneer/timemachine/blob/3984fe517d92904986614515810ba392dc60f5a4/timemachine/md/moves.py#L51-L53 (to use `self.rng.uniform(0,1)` instead of `np.random.rand()`)

Just wouldn't require modifying places that call `.move(x)`.

Changing the places that call `.move(x)` to be `.move(x, key)` or similar may be the about the same difficulty.

>(Unless we remove the inheritance for now and just have NeighborSwapMove reimplement move)

That also seems fine to me -- the only things MonteCarloMove defines are simple, generic, and easily replicated (like an acceptance counter)

tests/test_relative_free_energy.py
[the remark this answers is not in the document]
badisa: > nit: this is a strict xfail, so the test itself will fail if the assertions pass. So I would argue it does provide some signal, but maybe not worth the time required to run it.

Thinking about this some more, this will always fail. Even if there are no swaps. Because of the way you run the equilibration in the hrex it is `eq_steps + 1` rather than `eq_steps` so even calling `sample()` with the original MDParams will produce a difference. I get your point, this simply seems to be too abstract of a test that requires thinking about running HREX in an atypical way.

timemachine/fe/plots.py
mcwitt: It seems like we have a mix of existing conventions in this file. Personally I think I'd prefer to decouple the drawing of the plots with serialization to bytes, since the former would be nicer to use in notebooks and scripts.

timemachine/md/hrex.py
mcwitt: The `n_samples_batch` argument is currently used to handle cases where the total number of samples requested is not a multiple of `n_samples_per_iter`. In this case, the last batch will have fewer samples. We could consider just "rounding up" `n_samples` to the nearest multiple of `n_samples_per_iter` (which would allow removing this argument), but I wasn't immediately sure of the best approach here.

tests/test_relative_free_energy.py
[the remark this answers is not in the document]
mcwitt: > Because of the way you run the equilibration in the hrex it is eq_steps + 1 rather than eq_steps so even calling sample() with the original MDParams will produce a difference.

Yes, good point. Will just remove this test.

timemachine/md/hrex.py
badisa: Right, missed that. Thanks for the clarification

tests/test_relative_free_energy.py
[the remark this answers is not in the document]
mcwitt: Changed from `xfail` to `skip` in https://github.com/proteneer/timemachine/pull/1128/commits/ccbcb619501c32e7eca6adc39e612c6335dfbf88.

(Skipped instead of removed for documentation purposes)

timemachine/md/hrex.py
```diff
@@ -0,0 +1,247 @@
...
+from timemachine.md.moves import MetropolisHastingsMove, MixtureOfMoves
+from timemachine.utils import batches
+
+_Replica = TypeVar("_Replica")
+
+StateIdx = NewType("StateIdx", int)
+ReplicaIdx = NewType("ReplicaIdx", int)
+
+
+@dataclass
```
proteneer, line 16: nit: why is this a dataclass?  (it doesn't seem to declare any internals in the usual dataclass style)

timemachine/md/hrex.py
```diff
@@ -0,0 +1,247 @@
...
+    replica_idx_by_state_by_iter_ = np.array(replica_idx_by_state_by_iter)  # (iter, state) -> replica
+    _, n_states = replica_idx_by_state_by_iter_.shape
+    states = np.arange(n_states)
+    replica_in_state = replica_idx_by_state_by_iter_[:, :, None] == states  # (iter, replica, state) -> bool
+    cumulative_count = np.cumsum(replica_in_state.astype(int), axis=0)  # (iter, replica, state) -> int
+    return cumulative_count
+
+
+def estimate_transition_matrix(replica_idx_by_state_by_iter: Sequence[Sequence[ReplicaIdx]]) -> NDArray:
+    """Given a mapping of state index to replica index by iteration, returns an estimate of the transition matrix.
```
proteneer, line 105: Add documentation:

1) Square matrix but not expected to be symmetric
2) Markovian, but along which dimension? (i.e. reduction along which axis sums to 1?)

timemachine/md/hrex.py
```diff
@@ -0,0 +1,247 @@
...
+    transition_rate = transition_count / n_iters  # (to state, from state) -> float
+
+    return transition_rate
+
+
+def estimate_relaxation_time(transition_matrix: NDArray) -> float:
+    """Estimate the relaxation time of permutation moves (in number of iterations) as a function of the second-largest
+    eigenvalue of the transition matrix."""
+
+    eigvals_ascending = np.linalg.eigvals(transition_matrix)
```
proteneer, line 128: Any assumptions about the underlying structure of the markov chain? Maybe assert that sum of column (or row) vectors sum to 1 before proceeding?

timemachine/md/hrex.py
```diff
@@ -0,0 +1,247 @@
...
+        log_q_diff = self.log_q(r_a, s_b) + self.log_q(r_b, s_a) - self.log_q(r_a, s_a) - self.log_q(r_b, s_b)
+
+        return proposed_state, log_q_diff
+
+
+_Samples = TypeVar("_Samples")
+
+
+@dataclass
+class Hrex(Generic[_Replica]):
```
proteneer, line 42: we use HREX (in full upper case) a lot, maybe we should just use all-caps here?

timemachine/fe/free_energy.py
```diff
@@ -725,3 +722,185 @@ def compute_intermediate_result(lambdas: Sequence[float]) -> PairBarResult:
     boxes = [get_state(lamb).boxes for lamb in lambdas]
 
     return results, frames, boxes
+
+
+def run_sims_hrex(
```
proteneer, line 727: nice! seems like I can use this pretty easily!

timemachine/fe/free_energy.py
```diff
@@ -725,3 +722,185 @@ def compute_intermediate_result(lambdas: Sequence[float]) -> PairBarResult:
...
+    initial_states: Sequence[InitialState],
+    md_params: MDParams,
+    n_frames_per_iter: int,
+    temperature: float,
+    n_swap_attempts_per_iter: Optional[int] = None,
+    verbose: bool = True,
+) -> Tuple[PairBarResult, List[StoredArrays], List[NDArray], HrexDiagnostics]:
+    r"""Sample from a sequence of states using nearest-neighbor Hamiltonian Replica EXchange (HREX).
```
proteneer, line 735: General question: does the algorithm in this PR still assume some sort of order dependence in the lambda schedule / set of initial states to be effective? (i.e. nearby neighbors have to have good overlap?). At first glance, since we rely on near-by-neighbor swaps (as opposed to global swaps from all possible permutations), if we say accidentally randomly shuffled the initial states, then HREX as currently implemented would fail? 

(This is not a blocker for the PR or anything, just wanted to double check)

proteneer (commented):
(Wasn't tagged as a reviewer, but wanted to take a look at this awesome work anyways!)

tests/test_relative_free_energy.py
```diff
@@ -391,3 +415,4 @@ def test_estimate_free_energy_bisection_invalid_args():
     # toggling the pytest marker
     test_run_hif2a_test_system(estimate_relative_free_energy)
     test_run_hif2a_test_system(estimate_relative_free_energy_bisection)
+    test_run_hif2a_test_system(estimate_relative_free_energy_bisection_hrex)
```
badisa, line 418: Do we still want this since we know it will fail? Could be confusing in the future

timemachine/md/hrex.py
mcwitt: No strong preference; I've tended to prefer dataclasses for classes with a trivial `__init__` method. Switching to a regular class here would allow preventing mutation of `replicas` and `replica_idx_by_state`, which might be preferable

timemachine/md/hrex.py
mcwitt: Could do; I'd chosen this convention because multi-word identifiers can look awkward with the all-caps style (e.g. `HREXDiagnostics`), but have no strong preference.

timemachine/md/hrex.py
mcwitt: Ah, sorry, I misread the location of this comment and thought it applied to the `Hrex` class. I didn't intend `NeighborSwapMove` to be a dataclass; this is left over from a previous design. Will remove.

mcwitt: Removed in https://github.com/proteneer/timemachine/pull/1128/commits/46707c9c63b3c38ae2f594dba1f0e40dbaf3911d

timemachine/md/hrex.py
mcwitt: Renamed in https://github.com/proteneer/timemachine/pull/1128/commits/2878505cfe1d8b0de5cdffc430203def182a72e8

badisa (approved):

timemachine/md/hrex.py
mcwitt: Addressed in https://github.com/proteneer/timemachine/pull/1128/commits/8406c2164737ccc232d084094acd9fdb9b494af5 and https://github.com/proteneer/timemachine/pull/1128/commits/7994a93275dd5504af0dc901442fa664d73e56aa

proteneer (approved):

timemachine/md/hrex.py
mcwitt: Added assertion in https://github.com/proteneer/timemachine/pull/1128/commits/01755c6c64f74facea06228042af2ebfb7928470

timemachine/fe/free_energy.py
mcwitt: Added mention of this to the docstring in https://github.com/proteneer/timemachine/pull/1128/commits/814a8c825e353f63ddc9480b64e7e752e5f8f5c8

timemachine/fe/free_energy.py
```diff
@@ -725,3 +722,184 @@ def compute_intermediate_result(lambdas: Sequence[float]) -> PairBarResult:
...
+    state_idxs = [StateIdx(i) for i, _ in enumerate(lambdas)]
+    neighbor_pairs = list(zip(state_idxs, state_idxs[1:]))
+
+    # Add (0, 0) to the list of neighbor pairs considered for swap moves to ensure that performing a fixed number of
+    # neighbor swaps is aperiodic in cases where swap acceptance rates approach 100%
+    neighbor_pairs = [(StateIdx(0), StateIdx(0))] + neighbor_pairs
+
+    hrex = Hrex.from_replicas(replicas)
+
+    def get_equilibrated_initial_state(initial_state: InitialState, seed: int) -> InitialState:
```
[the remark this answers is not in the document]
mcwitt: Attempted to "deboggle" this in https://github.com/proteneer/timemachine/pull/1128/commits/966d565f87cbb2448683a33dd395348e81935e81

timemachine/md/hrex.py
mcwitt: In https://github.com/proteneer/timemachine/pull/1128/commits/6a31ac42c0dc57a6b73eef44b19ea22fa062840f, modified to add an identity move if and only if K=2, since I don't think there is a potential issue with aperiodicity for K>2.

maxentile: >I don't think there is a potential issue with aperiodicity for K>2.

👍 , I think it should be aperiodic already for K > 2.

timemachine/fe/free_energy.py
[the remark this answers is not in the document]
mcwitt: After some time attempting various fixes, I wasn't able to find a satisfactory solution to removing the dependence on global random state that seemed reasonable to implement in this PR. The approaches I tried were:

1. **Add random state as instance variable to `Move` and subclasses, add `seed` to constructor.** this becomes quite tedious (we have to initialize each neighbor swap move and the mixture move with a random seed and ensure that they are all different) and I found that it obscures the mathematical definitions
1. **Augment signature of `propose` and `move` methods to accept a `numpy.random.Generator` instance.** This seems to have better compositionality than (1), e.g. it's not necessary to initialize separate PRNGs for each move in a mixture, but is incompatible with `DeterministicMTMMove`, which uses the JAX random API.
1. **Augment signature of `propose` and `move` methods to accept a `jax.random.PRNGKeyArray`. This approach has the compositionality of (2) and works with the existing `DeterministicMTMMove`, but generating random numbers with JAX is *much* [slower](https://github.com/google/jax/issues/968) than with numpy unless we're able to use `jax.jit` (in which case it's blazingly fast). This would probably be my preferred approach in the future, but it would require some refactoring (I think mainly just in the HREX code) due to the restriction that `State` must be a PyTree (e.g. immutable collection of primitive types or `jax.Array`s)

For this PR, I would propose to remove the call to `np.random.seed` in `run_sims_hrex` and rely on the caller to set the numpy random state (with the understanding that this is _not_ controlled by `md_params.seed`, which I will emphasize in the docstring). I think this aligns well with existing convention (i.e. places where we currently use `MonteCarloMove`), and will allow us time to design a good solution to PRNG state. @badisa @maxentile what do you think?

mcwitt:
> Test expansion request: can we add an HREX workload to the performance benchmarking script `test_benchmark.py`?
> 
> Not a full free energy calculation, but running a solvent leg and complex leg with a couple settings of n_windows (maybe 10, 20, 50) for about 30 seconds - 1 minute each? Would be helpful to keep an eye on relative costs of swap attempts vs. MD.

Added a nightly benchmark in https://github.com/proteneer/timemachine/pull/1128/commits/33f0e72ee0e37a8b03fe626157530252598cf213.

timemachine/fe/free_energy.py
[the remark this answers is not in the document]
maxentile: Thanks for investigating and reporting on these.

Any of these 3 described refactors sounds appropriate to me for later, and for this PR either the currently implemented approach (which resets global numpy random state within `run_sims_hrex`) or the proposed approach (which has `run_sims_hrex` depend on and modify global numpy random state) sounds appropriate.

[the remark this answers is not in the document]
mcwitt: In the end, I opted to keep setting the numpy random state in `run_sims_hrex`, but added a warning and todo to address this in https://github.com/proteneer/timemachine/pull/1128/commits/2db9fd794d8a434f3ad8ef37945ca3c8f98c235f

timemachine/fe/free_energy.py
mcwitt: Was fixed in https://github.com/proteneer/timemachine/pull/1128/commits/001202da338e20c48fcad0c9499da3da82cca19a and https://github.com/proteneer/timemachine/pull/1128/commits/6ff38e4d7bf6d66595a8c5b451f1ac4abde8404b

tests/test_relative_free_energy.py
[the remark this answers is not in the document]
mcwitt: Added a test for reproducibility in https://github.com/proteneer/timemachine/pull/1128/commits/d0f5a030e49958ef0af726341835278cfbefcfe4

tests/hrex/test_hrex_alchemical.py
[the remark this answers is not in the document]
mcwitt: After tuning the parameters, running all 3 stages ends up taking about 15 minutes, which seems reasonable for a nightly test.

tests/test_benchmark_free_energy.py
```diff
@@ -0,0 +1,98 @@
...
+    print("hrex:", enable_hrex)
+    print("host:", host_name)
+    print("n_windows:", n_windows)
+    print("n_frames:", n_frames)
+
+    dt_ns = initial_states[0].integrator.dt / 1e3
+    ns = md_params.n_frames * md_params.steps_per_frame * dt_ns
+    elapsed_days = elapsed_ns / 1e9 / 60 / 60 / 24
+
+    print(f"total ns/day: {n_windows * ns / elapsed_days:.1f}")
```
maxentile, line 98: nit: locally, this is printing ns/day numbers far in excess of what I'd expect, e.g.

```
...
hrex: False
host: None
n_windows: 20
n_frames: 50
total ns/day: 6320.9
...
hrex: True
host: None
n_windows: 20
n_frames: 50
total ns/day: 3668.2
...
```

where I'd expect at most ~800 ns/day (based on timings from `test_benchmark.py`, `hif2a-rbfe-barostat-interval-15`)

but I don't immediately see an error. Is this expected to exceed the speed of running MD in a single window (i.e. without overheads from multiple contexts + replica-mixing)?

mcwitt: Is this in excess of what we'd expect for a vacuum simulation? (Note `host: None` here)

maxentile: doh! sorry, overlooked that. not in excess of expectations for vacuum simulation, was mistakenly comparing to hif2a-rbfe benchmark script (with n_atoms=8840)

PR #1129, opened 2023-08-18
Skip nightly condensed mtm test

badisa:
* Modified to make it easier to test multiple seeds for variation
* Modifies the cache to write out files based on the parameters to avoid reuse when changing parameters
* Fixes num workers to 4 to be consistent across CI and other machines

### DIFF
--- a/tests/test_mtm_condensed.py
+++ b/tests/test_mtm_condensed.py
@@ -19,13 +19,14 @@ from timemachine.md.states import CoordsVelBox
 from timemachine.potentials import NonbondedInteractionGroup, bonded, nonbonded
 
 
+@pytest.mark.skip("Has shown to be flaky, needs further investigation. Condensed MTM not currently used")
 @pytest.mark.nightly(reason="Takes a long time to run")
-def test_condensed_phase_mtm():
+@pytest.mark.parametrize("seed", [2021])
+def test_condensed_phase_mtm(seed):
     """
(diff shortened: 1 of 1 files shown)


jkausrelay (approved):

proteneer (approved):

badisa:
Some context from my investigation of this:

In https://github.com/proteneer/timemachine/commit/7ae44c382edce5439d3d9995817b4f46ca8b3e98 we changed from 40k batches to 400k batches to resolve this line failing https://github.com/proteneer/timemachine/blob/b853c2d287da0d1c1babb963eaec8fda41539b90/tests/test_mtm_condensed.py#L101 when I tried going back to 40k, I saw it vary between 0.06 and 0.23 over the course of 10 random seeds

 Looking at the line that failed today https://github.com/proteneer/timemachine/blob/master/tests/test_mtm_condensed.py#L162 it appears that can go up to 0.09. Due to time it took to run with 400k batches, I reduced to 40k batches and saw a range of 5e-3 and 0.15 across 10 seeds.
 
 There is still something resulting in different results on different machines. Have another run going now that I hope will reproduce it, and will update this comment if I have correctly reproduced the failure.
 
 Kicked off nightly tests to verify that there isn't another nightly test that will fail: https://gitlab.com/relaytx/physics-platform/timemachine/-/pipelines/972748703

PR #1133, opened 2023-08-23
Clean up Benchmarks script

badisa:
- Separate out solvent for benchmarking as sometimes it isn't of interest
- Refactor the code to be less verbose
- Adds a configuration object which removes an old TODO
- Avoid minimizing host twice, which reduces the noise in the nvidia profiles we get out
- Avoids duplicate benchmarking case that had arisen when we did the FF split

### DIFF
--- a/tests/test_benchmark.py
+++ b/tests/test_benchmark.py
@@ -317,8 +317,68 @@ def benchmark_local(
     )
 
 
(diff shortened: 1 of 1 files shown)


mcwitt (approved):
Looks great!

PR #1134, opened 2023-08-23
Refactor and organize interface for Monte Carlo moves

mcwitt:
Collection of miscellaneous improvements spun out of #1128. Highlights:

* Adds `MetropolisHastingsMove`, which extends `MonteCarloMove` to compute acceptance probabilities using the Metropolis convention
* Adds a base class `Move` to work around inconsistency between definition of `n_proposed`, `n_accepted` in `MonteCarloMove` vs. `CompoundMove`, and to clarify whether subclasses implement `propose` or `move`

Note: commits are mostly orthogonal changes

### DIFF
--- a/tests/test_md_moves.py
+++ b/tests/test_md_moves.py
@@ -32,13 +32,13 @@ def test_random_walk_metropolis_hastings(dist, seed):
     tau = round(1 / dx ** 2)
     n_independent_samples = n_samples // tau - 1
 
-    d_log_q = np.random.uniform(-1.0, 1.0)  # arbitrary offset added to log_q
+    log_q_offset = np.random.uniform(-1.0, 1.0)  # arbitrary offset added to log_q
 
     if dist == "normal":
-        log_q = lambda x: -(x ** 2) / 2 + d_log_q
+        log_q = lambda x: -(x ** 2) / 2 + log_q_offset
         target_samples = np.random.normal(0, 1, size=(n_independent_samples,))
     else:
-        log_q = lambda x: d_log_q if -1 < x < 1 else -float("inf")
+        log_q = lambda x: log_q_offset if -1 < x < 1 else -float("inf")
         target_samples = np.random.uniform(-1, 1, size=(n_independent_samples,))
 
     sampler = RWMH1D(log_q, dx)
--- a/tests/test_md_moves.py
+++ b/tests/test_md_moves.py
@@ -1,37 +1,50 @@
-from typing import Tuple
+from typing import Callable, Tuple
 
 import numpy as np
+import pytest
 from scipy.stats import ks_2samp
 
 from timemachine.md.moves import MetropolisHastingsMove
 
 
-def test_random_walk_metropolis_hastings():
+class RWMH1D(MetropolisHastingsMove[float]):
+    def __init__(self, log_q: Callable[[float], float], proposal_radius: float):
+        super().__init__()
+        self.log_q = log_q
+        self.proposal_radius = proposal_radius
+
+    def propose_with_log_q_diff(self, x: float) -> Tuple[float, float]:
+        dx = self.proposal_radius
+        x_prop = x + np.random.uniform(-dx, dx)
+        log_q_diff = self.log_q(x_prop) - self.log_q(x)
+        return x_prop, log_q_diff
+
+
+@pytest.mark.parametrize("dist", ["normal", "uniform"])
+@pytest.mark.parametrize("seed", [2023, 2024, 2025])
+def test_random_walk_metropolis_hastings(dist, seed):
     """Test sampling from a 1-d normal distribution using Random Walk Metropolis-Hastings"""
 
-    np.random.seed(2023)
-    n_samples = 100_000
+    np.random.seed(seed)
+    n_samples = 200_000
     dx = 0.1
 
-    class RWMH1D(MetropolisHastingsMove[float]):
-        def propose_with_log_q_diff(self, x: float) -> Tuple[float, float]:
-            x_prop = x + np.random.uniform(-dx, dx)
+    if dist == "normal":
+        log_q = lambda x: -(x ** 2) / 2
+        sample_target = lambda n: np.random.normal(0, 1, size=(n,))
+    else:
+        log_q = lambda x: 0.0 if -1 < x < 1 else -float("inf")
+        sample_target = lambda n: np.random.uniform(-1, 1, size=(n,))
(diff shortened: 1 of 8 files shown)


timemachine/md/moves.py
```diff
@@ -12,69 +13,85 @@
...
 
-class CompoundMove(MonteCarloMove[_State]):
-    def __init__(self, moves: List[MonteCarloMove]):
-        """Apply each of a list of moves in sequence"""
-        self.moves = moves
+
+class MetropolisHastingsMove(MonteCarloMove[_State], ABC):
+    @abstractmethod
+    def propose_with_dlogp(self, x: _State) -> Tuple[_State, float]:
+        "return proposed state and its unnormalized log probability"
```
maxentile, line 66: documentation: `dlogp = `
* `logp(proposed) - logp(x)`, or 
* unnormalized `logp(proposed)`?

(adding a quick usage example in tests would be ideal, e.g. RWMH in 1D)

maxentile (approved):
Nice improvements!

timemachine/md/moves.py
mcwitt: Added documentation and renamed to `dlogq` (`q` to indicate unnormalized by convention) in https://github.com/proteneer/timemachine/pull/1134/commits/03373e08f9f1e4c1d4f61501690c44f4fb49b0bf.

Added 1D RWMH test in https://github.com/proteneer/timemachine/pull/1134/commits/b8aebff39f3b1f814c373f79ee598e46fab96082.

tests/test_md_moves.py
```diff
@@ -0,0 +1,33 @@
...
+
+    np.random.seed(2023)
+    n_samples = 100_000
+    dx = 0.1
+
+    class RWMH1D(MetropolisHastingsMove[float]):
+        def propose_with_dlogq(self, x: float) -> Tuple[float, float]:
+            x_prop = x + np.random.uniform(-dx, dx)
+            log_q = -(x_prop ** 2) / 2
+            return x_prop, log_q
```
maxentile, line 20: should this be 
```suggestion
            log_q_0 = -(x ** 2) / 2
            log_q_prop = -(x_prop ** 2) / 2
            
            log_q_diff = log_q_prop - log_q_0
            
            return x_prop, log_q_diff
```
?

mcwitt: Yes :facepalm: 

Fixed in https://github.com/proteneer/timemachine/pull/1134/commits/83205ba792c8f2ee7cbaa3a8557225a4794a827a

tests/test_md_moves.py
```diff
@@ -0,0 +1,38 @@
+from typing import Tuple
+
+import numpy as np
+from scipy.stats import ks_2samp
+
+from timemachine.md.moves import MetropolisHastingsMove
+
+
+def test_rwmh():
```
badisa, line 9: nit: For clarity when running tests
```suggestion
def test_random_walk_metropolis_hastings():
```

timemachine/md/moves.py
```diff
@@ -12,69 +14,96 @@
...
 
-class MonteCarloMove(Generic[_State]):
-    n_proposed: int = 0
-    n_accepted: int = 0
+class Move(Generic[_State], ABC):
+    @abstractmethod
+    def move(self, _: _State) -> _State:
+        ...
+
+    def sample_chain_iter(self, x: _State) -> Iterator[_State]:
```
badisa, line 28: Maybe this is a convention already, but seems to have an infinite iterator here. Would be good to have a docstring to mention this, as it would be dangerous to run this in production

timemachine/md/moves.py
```diff
@@ -12,69 +14,96 @@
...
+        while True:
+            x = self.move(x)
+            yield x
+
+    def sample_chain(self, x: _State, n_samples: int) -> List[_State]:
+        return list(islice(self.sample_chain_iter(x), n_samples))
+
 
+@dataclass
```
badisa, line 37: What does dataclass get us in this case? This seems less like a class that contains data than a collection of functionality.

badisa (approved):

timemachine/md/moves.py
```diff
@@ -12,69 +14,96 @@
...
+        proposal, log_q_diff = self.propose_with_log_q_diff(x)
+        log_acceptance_probability = np.minimum(log_q_diff, 0.0)
+        return proposal, log_acceptance_probability
+
+
+@dataclass
+class CompoundMove(Move[_State]):
+    """Apply each of a list of moves in sequence"""
+
+    moves: Sequence[MonteCarloMove[_State]]
```
badisa, line 90: Should this be `Sequence[Move[_State]]` else the doc string should clarify that it is only `MonteCarloMove`s

timemachine/md/moves.py
mcwitt: Not much in retrospect. I wanted to be able to annotate the types of `n_proposed` and `n_accepted`, but it appears that's possible to do without making it a dataclass. Will revert.

badisa: Before you revert it, it looks like it does make `help(MonteCarloMove)` much cleaner, so it might be worth it?  Find it a bit of an odd use of the `dataclass` abstraction, but looks like it might make the use/documentation of it cleaner.

timemachine/md/moves.py
mcwitt: The annotation is as intended, since `n_accepted_by_move` etc. only make sense for `MonteCarloMove`s. Updated the docstring in https://github.com/proteneer/timemachine/pull/1134/commits/d634b45c8e93ecfb33159dbb2d29120e480904e0 to hopefully clarify.

timemachine/md/moves.py
mcwitt: Hmm, they seem [about the same](https://gist.github.com/mcwitt/2c47ff750cd19bc1c2b4a0435f5666f3) (i.e. not that clean) to me? With the plain class, help includes "Methods inherited from Move", which seems to be missing with the dataclass.

tests/test_md_moves.py
mcwitt: A bit concerning that the test was still passing with this error. Will try to strengthen before merging.

mcwitt: Strengthened the test in https://github.com/proteneer/timemachine/pull/1134/commits/dae47fbeb0f8052e54133588532bf878e125085f, adding a uniform sampling case and multiple seeds.

PR #1136, opened 2023-08-25
Clean up Future Interface

badisa:
* Adds an AbstractFuture class to ensure all futures have the same interface
* Adds a method to iterate through futures that are completed for use when futures return at different rates (important to handle certain types of failures)
* Fixes up the id/name field of the MockFutures

### DIFF
--- a/tests/test_parallel.py
+++ b/tests/test_parallel.py
@@ -2,6 +2,7 @@
 import io
 import os
 import pickle
+import time
 import unittest
 from pathlib import Path
 from unittest.mock import patch
@@ -31,6 +32,37 @@ def mult(x, y):
 
 def sum(*args, **kwargs):
     return args[0] + kwargs["key"]
+
+
+class DummyFuture(client.BaseFuture):
+    """A future that if you don't call done several times, the result() method will sleep"""
+
+    def __init__(self, iterations: int):
+        self.count = 0
+        self.iterations = iterations
+
+    def _ready(self):
+        return self.count > self.iterations
+
+    def result(self) -> str:
+        if not self._ready():
+            time.sleep(2)
+        return "finished"
+
+    def done(self) -> bool:
+        if not self._ready():
+            self.count += 1
+            return False
+        else:
+            return True
+
+    @property
+    def id(self) -> str:
+        return "a"
+
+    @property
+    def name(self) -> str:
+        return "a"
 
 
 class TestProcessPool(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


tests/test_parallel.py
```diff
@@ -202,3 +234,21 @@ def test_save_results(tmpdir):
...
+
+def test_iterate_completed_futures():
+    iterations = 5
+    fut = DummyFuture(iterations)
+    start = time.time()
+    fut.result()
+    # If we call result directly, will take ~2 seconds
+    assert time.time() - start > 1
+
+    futures = [DummyFuture(iterations) for _ in range(iterations)]
```
mcwitt, line 247: Rather than a time-based test, would it be simpler and more robust to test this with mock futures? Thinking of something like:

```suggestion
    futures = [Mock(done=Mock(side_effect=[False] * n + [True])) for n in range(iterations)]
```

timemachine/parallel/client.py
```diff
@@ -2,21 +2,43 @@
...
 # multiprocessing (typically for local cluster use) and gRPC (distributed and multi-node).
 
 
+class BaseFuture(ABC):
+    @abstractmethod
+    def done(self) -> bool:
+        ...
+
+    @abstractmethod
+    def result(self) -> Any:
```
mcwitt, line 23: nit: suggest using a `TypeVar` rather than returning `Any` here. This could help catch errors where we accidentally mix futures returning different types of results.

mcwitt (approved):

timemachine/parallel/client.py
badisa: It is actually desirable in my mind to be able to have a mix of futures. They might be doing completely different tasks

tests/test_parallel.py
badisa: Will do

badisa: Removed time dependence in [6527eaa](https://github.com/proteneer/timemachine/pull/1136/commits/6527eaacb30017b935f5b611716383b08bd65b65) doesn't use Mock though as while that can assert the iteration, I wanted to verify that `result` is 'unsafe' to call without the iterating and seems cleaner to define a new class than to create construct an equivalent mock object.

timemachine/parallel/client.py
mcwitt: To clarify, I just mean that this could prevent errors of the form:

```
def foo(fut: BaseFuture[Optional[int]]):
    return fut.result() + 1
```

With the type parameter, this sort of thing would fail static checks (because `1 + None` isn't allowed), but as-is it wouldn't.

(non-blocking)

PR #1138, opened 2023-09-06
Fix split ixn

jkausrelay:
- Ordering of ff.get_params() was flipped leading to the wrong set of parameters being used
- Add unit test to check this
- [x] add test for combine_params


tests/test_forcefields.py
```diff
@@ -84,3 +84,51 @@ def test_load_default():
     for (ref_handle, test_handle) in zip(ref.get_ordered_handles(), test.get_ordered_handles()):
         assert ref_handle.smirks == test_handle.smirks
         np.testing.assert_array_equal(ref_handle.params, test_handle.params)
+
+
+def test_split():
+    ffs = load_split_forcefields()
```
maxentile, line 90: Test expansion request: Would it be informative also to assert self-consistency with the order of handles assumed by `parameterize_nonbonded` methods in `fe` module?

maxentile (approved):
Good fix!

tests/test_forcefields.py
jkausrelay: The parameters are referenced by name (https://github.com/proteneer/timemachine/blob/master/timemachine/fe/free_energy.py#L196), unless you were thinking of something else.

maxentile: Ahh, sorry, I see. I guess it's possible to pass handles in a wrong order to 
https://github.com/proteneer/timemachine/blob/0628c0e8465805c94bad0d70d25a0abbde3d522e/timemachine/fe/free_energy.py#L196-L203

but that wouldn't have anything to do with consistency of `ff_params` and `topology.parameterize_nonbonded`.

Don't have great ideas for an integration test here -- and maybe this is already tested thoroughly enough? -- but some test to check that the generated `sys_params` (not just the `ff_params`) are correct in the presence of IG splitting may be helpful.

timemachine/ff/__init__.py
```diff
@@ -33,11 +33,11 @@ def combine_params(a: ForcefieldParams[_T], b: ForcefieldParams[_T]) -> Forcefie
...
         (a.it_params, b.it_params),
         (a.q_params, b.q_params),
-        (a.q_params_solv, b.q_params_solv),
         (a.q_params_intra, b.q_params_intra),
+        (a.q_params_solv, b.q_params_solv),
         (a.lj_params, b.lj_params),
-        (a.lj_params_solv, b.lj_params_solv),
         (a.lj_params_intra, b.lj_params_intra),
+        (a.lj_params_solv, b.lj_params_solv),
     )
```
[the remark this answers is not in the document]
mcwitt: Don't immediately have any ideas for making this safer with types that don't involve unwieldy type annotations. One way to catch this would be to give each field in `ForcefieldParams` a distinct type (e.g. `HarmonicBondParams = NewType("HarmonicBondParams", NDArray)`), but this would make `combine_params` very awkward to type.

tests/test_forcefields.py
jkausrelay: Right. Not sure either. Will merge but think about refactoring this later.

mcwitt (approved):

PR #1140, opened 2023-09-06
Fix spurious truncation of HREX swap stats

mcwitt:
When there is just a pair of states (K = 2), we add an identity move to the mixture of neighbor swap moves in HREX to prevent periodicity when the swap acceptance probability is close to 1.

Originally, the identity move was added regardless of the value of K, but we changed the logic in #1128 to only add it when K = 2, because the mechanism for periodicity that it's intended to work around can't occur for K > 2. But we neglected to wrap the corresponding truncation to remove the acceptance stats for the identity move in the same conditional, resulting in truncating acceptance stats for the first neighbor swap move when K > 2.

This PR fixes the issue and adds a test.


proteneer (approved):
lgtm - thanks for the hotfix

PR #1141, opened 2023-09-07
Fix nightly test

mcwitt:
Changes the seed for a stochastic nightly test. Does not change the acceptance threshold, as it passes ~80% of seeds tried (so it seems we were just unlucky). This was missed in local development because I'm using `numpy==1.25.1`, which apparently does not produce identical pseudorandom number streams as the pinned version (`1.23.5`) for a given seed.

- [x] Verify nightly tests pass in CI


proteneer (approved):
lol - lgtm! but a little annoying that this is so seed dependent

mcwitt:
> lol - lgtm! but a little annoying that this is so seed dependent

Agreed, but relaxing the threshold (all replicas have visited all states) seems nontrivial ("most replicas visited most states"?), and it does pass for most seeds, so this change seemed appropriate for now.

PR #1142, opened 2023-09-12
Preserve velocities in HREX permutation moves

mcwitt:
Currently velocities are lost after each MD propagation step in HREX, with the subsequent MD propagation step starting with zero velocity. This is effectively freezing the system every 5 ps with the current parameters.

This PR switches from using `CoordsBox` as the replica state type to `CoordsVelBox` to avoid dropping the velocities during HREX permutation moves.

PR #1143, opened 2023-09-13
Assert ensembles compatible in run_sims_hrex

maxentile:
Follow up to https://github.com/proteneer/timemachine/pull/1142 

(In the future, if `initial_states` may contain multiple temperatures, we can generalize sample fxn (to rescale velocities from source state to target state by a factor of `sqrt(T_target / T_src)` or similar).)


timemachine/fe/free_energy.py
```diff
@@ -786,6 +786,21 @@ def run_sims_hrex(
...
+    #  consider modifying sample fxn to rescale velocities by sqrt(T_new/T_orig)
+    def _assert_velocity_distributions_compatible(state_a: InitialState, state_b: InitialState):
+        """assert masses and temperature match.
+        (so that -- up to timestep error -- (x, v, b) in state_a can be swapped with (x, v, b) from state_b)"""
+
+        intg_a = state_a.integrator
+        intg_b = state_b.integrator
+
+        assert (intg_a.masses == intg_b.masses).all()
+        assert intg_a.temperature == intg_b.temperature
```
mcwitt, line 799: question: should we also assert that pressures are equal?

mcwitt (approved):

timemachine/fe/free_energy.py
maxentile: Ahh, good call, was waffling about that.

(We seem less likely to ever change pressures, but would be good to enforce that assumption. If the pressures were unequal, then the `u_kln` matrix would just have to be computed differently (including [a `pressure * volume` term ](https://github.com/proteneer/timemachine/blob/7ac5ff6f50ef6d58e7f11b4cebe7beb2b3487730/attic/modules/md/ensembles.py#L119-L122) ).)

maxentile:
Testing note:
* `run_sims_hrex` is currently only tested in nightlies
* tested locally using `pytest tests/hrex/test_hrex_rbfe.py -k 'test_hrex_rbfe_hif2a[complex]'` (as suggested by @mcwitt)

PR #1145, opened 2023-09-15
Remove MetropolisHastingsMove

mcwitt:
This is a flawed abstraction since it assumes a symmetric proposal distribution. It seems simpler to just remove this than rename to `SymmetricMetropolisHastingsMove` or attempt to generalize the naming to apply to asymmetric cases.

### DIFF
--- a/timemachine/md/moves.py
+++ b/timemachine/md/moves.py
@@ -69,20 +69,6 @@ class MonteCarloMove(Move[_State], ABC):
         return self._n_accepted / self._n_proposed if self._n_proposed else np.nan
 
 
-class MetropolisHastingsMove(MonteCarloMove[_State], ABC):
-    @abstractmethod
-    def propose_with_log_q_diff(self, x: _State) -> Tuple[_State, float]:
-        """Return proposed state and the difference in log unnormalized probability, i.e.
-
-        log_q_diff = log(q(x_proposed)) - log(q(x))
-        """
-
-    def propose(self, x: _State) -> Tuple[_State, float]:
-        proposal, log_q_diff = self.propose_with_log_q_diff(x)
-        log_acceptance_probability = np.minimum(log_q_diff, 0.0)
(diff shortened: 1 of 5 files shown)


timemachine/md/moves.py
```diff
@@ -69,20 +69,6 @@ def acceptance_fraction(self) -> float:
...
-    def propose_with_log_q_diff(self, x: _State) -> Tuple[_State, float]:
-        """Return proposed state and the difference in log unnormalized probability, i.e.
-
-        log_q_diff = log(q(x_proposed)) - log(q(x))
-        """
-
-    def propose(self, x: _State) -> Tuple[_State, float]:
-        proposal, log_q_diff = self.propose_with_log_q_diff(x)
-        log_acceptance_probability = np.minimum(log_q_diff, 0.0)
-        return proposal, log_acceptance_probability
```
maxentile, line 83: This interface still seems like it would be fine for MH? ("Propose with log_accept_prob", where accept_prob contains contributions from both `log(q(proposal) / q(x))` and `log (proposal_prob(x | proposal) / proposal_prob(proposal | x))`)

Only misleading part is the name `log_q_diff`, which is an accurate name for symmetric moves (where `log (proposal_prob(x | proposal) / proposal_prob(proposal | x)) == 0` always), but an inaccurate name for asymmetric proposals (as in https://github.com/proteneer/timemachine/pull/1137/files#diff-fe79e3e3d73ad1887501544687228343106f245d72f970dcd6a3d1b59e9e8ebcR160-R190 , which is correctly implemented, but just goes through a misleadingly named intermediate step).

mcwitt: Hmm, wouldn't it be a bit confusing though that "propose with log acceptance probability" could return a positive result? If it's really a log probability, then couldn't we just use the `MonteCarloMove` interface?

PR #1146, opened 2023-09-18
Preserve MD context in HREX simulations

mcwitt:
Currently, we reinitialize the MD context for each state after each iteration of HREX. Thus, with the current defaults, we reinitialize the GPU potentials and context from scratch after every 5 ps of MD when running with HREX. This is already a significant overhead, but becomes extreme (~2x) when the amount of MD between HREX moves is reduced to 1 ps.

This PR eliminates a major source of overhead by preserving `Context` objects for each state across HREX iterations. Note that this does increase GPU memory usage by a factor of the number of windows, but this should still be well within reasonable bounds.

Further improvement might be achieved by avoiding copies to GPU memory when updating the context state to account for the current replica permutation. This should only require updating pointers to GPU memory, with no copying necessary.

Benchmarks (`test_benchmark_hif2a_single_topology`, solvent phase, 20 windows):

<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">

<colgroup>
<col  class="org-left" />

<col  class="org-right" />

<col  class="org-right" />

<col  class="org-left" />
</colgroup>
<thead>
<tr>
<th scope="col" class="org-left">&#xa0;</th>
<th scope="col" class="org-right">ns/day (before)</th>
<th scope="col" class="org-right">ns/day (after)</th>
<th scope="col" class="org-left">notes</th>
</tr>
</thead>

<tbody>
<tr>
<td class="org-left">no HREX</td>
<td class="org-right">370.2</td>
<td class="org-right">370.3</td>
<td class="org-left">(no difference expected)</td>
</tr>

<tr>
<td class="org-left">HREX, 5ps/move</td>
<td class="org-right">299.2</td>
<td class="org-right">361.7</td>
<td class="org-left">&#xa0;</td>
</tr>

<tr>
<td class="org-left">HREX, 1ps/move</td>
<td class="org-right">168.5</td>
<td class="org-right">307.8</td>
<td class="org-left">&#xa0;</td>
</tr>
</tbody>
</table>

### DIFF
--- a/timemachine/fe/free_energy.py
+++ b/timemachine/fe/free_energy.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass, replace
 from functools import cache
-from typing import Callable, List, Optional, Sequence, Tuple, Union, overload
+from typing import Callable, Iterable, List, Optional, Sequence, Tuple, Union, overload
 from warnings import warn
 
 import jax
@@ -32,6 +32,7 @@ from timemachine.md.barostat.utils import compute_box_center, get_bond_list, get
 from timemachine.md.hrex import HREX, HREXDiagnostics, ReplicaIdx, StateIdx, get_swap_attempts_per_iter_heuristic
 from timemachine.md.states import CoordsVelBox
 from timemachine.potentials import BoundPotential, HarmonicBond, SummedPotential
+from timemachine.potentials.potential import BoundGpuImplWrapper
 from timemachine.utils import batches, pairwise_transform_and_combine
 
 
@@ -365,6 +366,40 @@ def sample(
     """
 
     ctxt = get_context(initial_state)
+    return sample_with_context(
+        ctxt, md_params, initial_state.integrator.temperature, initial_state.ligand_idxs, max_buffer_frames
+    )
+
+
+@overload
+def sample_with_context(
+    ctxt: custom_ops.Context,
+    md_params: MDParams,
+    temperature: float,
+    ligand_idxs: NDArray[np.int32],
+    max_buffer_frames: None,
+) -> Tuple[NDArray, NDArray, NDArray]:
+    ...
+
+
+@overload
+def sample_with_context(
+    ctxt: custom_ops.Context,
+    md_params: MDParams,
+    temperature: float,
+    ligand_idxs: NDArray[np.int32],
+    max_buffer_frames: int,
+) -> Tuple[StoredArrays, NDArray, NDArray]:
+    ...
+
+
+def sample_with_context(
+    ctxt: custom_ops.Context,
+    md_params: MDParams,
+    temperature: float,
+    ligand_idxs: NDArray[np.int32],
+    max_buffer_frames: Optional[int] = None,
+) -> Tuple[Union[StoredArrays, NDArray], NDArray, NDArray]:
 
     # burn-in
     if md_params.n_eq_steps:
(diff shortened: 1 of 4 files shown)


timemachine/fe/free_energy.py
```diff
@@ -376,7 +411,7 @@ def sample(
     rng = np.random.default_rng(md_params.seed)
 
     if md_params.local_steps > 0:
-        ctxt.setup_local_md(initial_state.integrator.temperature, md_params.freeze_reference)
+        ctxt.setup_local_md(temperature, md_params.freeze_reference)
```
badisa, line 414: Think this will raise an exception on the second call. Probably worth just making this idempotent

badisa: I will put together a PR for this, to avoid adding unrelated cruft to this PR.

timemachine/fe/free_energy.py
```diff
@@ -794,26 +829,21 @@ def run_sims_hrex(
...
+    def get_summed_potential_bound_impl(bps: Iterable[BoundPotential]) -> BoundGpuImplWrapper:
+        potentials = [bp.potential for bp in bps]
+        params = [bp.params for bp in bps]
+        sp = SummedPotential(potentials, params)
+        return sp.to_gpu(np.float32).bind_params_list(params)
 
-    bps = initial_states[0].potentials  # TODO: assert initial states have compatible potentials?
-    sp = SummedPotential([bp.potential for bp in bps], [bp.params for bp in bps])
-    ubp = sp.to_gpu(np.float32)
+    bound_impls = [get_summed_potential_bound_impl(initial_state.potentials) for initial_state in initial_states]
```
badisa, line 838: Previously running larger systems (100k atoms, 48 windows) we would run out of GPU memory, and would be good to have any expectation of how large of a system with 48 windows we can fit on an A10 GPU (16GB of memory).

Note the additional changes that impact the memory since we last ran into this:
* Integrator batches up noise (10 * N doubles), we should probably change the noise to floats
* Ixn group split, now have 2-3x (solvent vs complex) the memory for hilbert sorts
* Accumulation of energies in int128

mcwitt: Ah, I had thought the main culprit in running out of GPU memory had been storage for the Hilbert grid. To be clear, have we run into this since shrinking the grid to 128^3?

badisa: I am pretty sure we had run out of GPU memory after we had changed the hilbert grid as well. We should verify again with a system of whatever size we think is a reasonable maximum, since memory usage has changed a fair bit. 

We changed the hilbert grid April 7th last year 99e8971a23f30cfa0bc6b50fbd72e7d4d308f2b9
As of 3779b1322b30c80055ab0b0b2be11eacc1209a13 on Feb 21 of this year, we had switched to only using two potentials.

proteneer: To document this in the PR directly. the current memory usage is not dominated by the hilbert curve:

3x hilbert curve grids: `d_bin_to_idx_: 3 (grids)*HILBERT_GRID_DIM^3 * 4 (bytes) = 0.024 GB`

Rather, it's the `d_ixn_atoms_` in `neighborlist.cu` when used on the `nonbonded_all_pairs` potential:

`max_ixn_count() * 4 bytes`

Where `max_ixn_count()` is:

` num_column_blocks() * num_row_blocks() * WARP_SIZE = (N//32)*(N//32)*32 = 1.25 GB`

We can probably bring this down to 1/5th - 1/10th the cost => 250MB

A typical GPU we use has 24GB of ram. Assuming a max of 48 windows, we'd still only use 12GB of ram with a 100K atom system if we make this change.

mcwitt:
Closing for now, as this approach will not scale to system sizes we want to support (>~ 100k atoms). See https://github.com/proteneer/timemachine/pull/1146#discussion_r1331664618.

PR #1147, opened 2023-09-19
Make setup_local_md idempotent

badisa:
* Would allow us to more generally share context's without having to know the prior state.
* Raise exception if setup has different parameters than previously setup
* Came up as part of HREX optimization (https://github.com/proteneer/timemachine/pull/1146#discussion_r1329427525)

### DIFF
--- a/timemachine/cpp/src/context.cu
+++ b/timemachine/cpp/src/context.cu
@@ -71,7 +71,15 @@ double Context::_get_temperature() {
 
 void Context::setup_local_md(double temperature, bool freeze_reference) {
     if (this->local_md_pots_ != nullptr) {
-        throw std::runtime_error("local md already configured");
+        if (this->local_md_pots_->temperature != temperature ||
+            this->local_md_pots_->freeze_reference != freeze_reference) {
+            throw std::runtime_error(
+                "local md configured with different parameters, current parameters: Temperature " +
+                std::to_string(this->local_md_pots_->temperature) + " Freeze Reference " +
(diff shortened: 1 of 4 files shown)


timemachine/cpp/src/context.cu
```diff
@@ -71,7 +71,14 @@ double Context::_get_temperature() {
...
     if (this->local_md_pots_ != nullptr) {
-        throw std::runtime_error("local md already configured");
+        if (this->local_md_pots_->temperature != temperature ||
+            this->local_md_pots_->freeze_reference != freeze_reference) {
+            throw std::runtime_error(
+                "local md configured with different parameters, current parameters: Temperature " +
+                std::to_string(this->local_md_pots_->temperature) + " Freeze Reference " +
+                std::to_string(this->local_md_pots_->freeze_reference));
+        }
+        return;
```
mcwitt, line 81: To be clear, would calling `setup_local_md` multiple times with different parameters lead to an inconsistent state, or is this intended to catch bugs in application code (assuming that any usage where we call this multiple times with different parameters is invalid)?

badisa: We could change out the underlying `LocalPotentials` object if there were different parameters, just risk paying the cost of setting up the potentials. This was mostly intended to catch application level bugs, as there is not currently a clear use case for different types of local MD configurations on a single context.

mcwitt (approved):

proteneer (approved):

PR #1150, opened 2023-09-20
Use MDParams for HREX params and enable HREX in run_leg

jkausrelay:
- Merge into #1151
- Use MDParams to store HREX parameters and enable HREX if hrex_n_frames_per_iter > 0
- This makes the interface consistent with local md (enabled if local_steps > 0)


mcwitt:
Curious if we have some criteria for what should live in `MDParams`? E.g. why HREX parameters, but not temperature? Motivation is to avoid ending up with a large blob of loosely-related parameters, where it becomes hard to tell what is actually being used by a given part of the code.

jkausrelay:
> Curious if we have some criteria for what should live in `MDParams`? E.g. why HREX parameters, but not temperature? Motivation is to avoid ending up with a large blob of loosely-related parameters, where it becomes hard to tell what is actually being used by a given part of the code.

We never actually change the temperature, but it could go in there too.
I think if it starts to become unwieldy we can separate out the params by simulation type:

```
@dataclass(frozen=True)
class MDParams:
    local_md_params: Optional[LocalMDParams] = None # or default
    hrex_params: Optional[HREXParams] = DEFAULT_HREX_PARAMS
    ...
```

timemachine/fe/free_energy.py
```diff
@@ -57,13 +57,21 @@ class MDParams:
     max_radius: float = 3.0  # nm
     freeze_reference: bool = True
 
+    # Number of frames to sample using MD during the initial bisection phase used to determine lambda spacing
+    hrex_n_frames_bisection: int = 100
+
+    # Number of frames to sample using MD per HREX iteration. Set to 0 to disable HREX.
+    hrex_n_frames_per_iter: int = 0
+
```
mcwitt, line 65: nit: slight preference for defining a dataclass `HREXParams` to contain these fields, and including as `Optional[HREXParams]` here. Motivation would be to avoid some confusion (e.g. is `hrex_n_frames_bisection` ignored if `hrex_n_frames_per_iter == 0`?), and to avoid needing to use `0` as a special value (could just pass `HREXParams = None`)

mcwitt: nit: change to documenting these fields as comments means they're no longer accessible in docstrings. Could consider documenting parameters of `MDParams` (or maybe just `HREXParams` if above suggestion is implemented?)

tests/test_relative_free_energy.py
```diff
@@ -162,7 +162,15 @@ def test_run_hif2a_test_system(estimate_relative_free_energy_fn):
...
 
-    md_params = MDParams(n_frames=100, n_eq_steps=1000, steps_per_frame=100, seed=2023)
+    md_params = MDParams(
+        n_frames=100,
+        n_eq_steps=1000,
+        steps_per_frame=100,
+        seed=2023,
+        hrex_n_frames_per_iter=1
+        if estimate_relative_free_energy_fn == estimate_relative_free_energy_bisection_hrex
+        else 0,
```
mcwitt, line 172: nit: is this necessary? (won't other `estimate_relative_free_energy_fn`s ignore `hrex_n_frames_per_iter`?)

timemachine/fe/rbfe.py
```diff
@@ -848,7 +857,12 @@ def run_complex(
     )
     complex_box += np.diag([0.1, 0.1, 0.1])  # remove any possible clashes, deboggle later
     complex_host_config = HostConfig(complex_sys, complex_conf, complex_box, nwa)
-    complex_res = estimate_relative_free_energy_bisection(
+    estimate_fxn = (
+        estimate_relative_free_energy_bisection_hrex
+        if md_params.hrex_n_frames_per_iter
+        else estimate_relative_free_energy_bisection
+    )
```
mcwitt, line 864: nit: would it be worth adding a generic `estimate_relative_free_energy` entry point, to avoid needing to replicate this logic in each `run_leg` function?

mcwitt:
Looks good to me aside from some nits.

Do we still want to add HREX-specific plots to `SimulationResult` in this or a separate PR?

proteneer:
+1 to adding the following plots at a minimum:

-replica population across states (either the stacked histogram, the heat map, or both)
-transition matrix

jkausrelay:
> Looks good to me aside from some nits.
> 
> Do we still want to add HREX-specific plots to `SimulationResult` in this or a separate PR?

Added the plots to SimulationResult

mcwitt (approved):

PR #1151, opened 2023-09-20
Avoid unnecessary context creation in HREX, using `BoundPotential::set_params`

mcwitt:
Alternative to https://github.com/proteneer/timemachine/pull/1146, where instead of preserving contexts for each state (which can exhaust GPU memory for large systems), we update a single context with coordinates, velocities, and parameters when iterating over states during MD updates.

#### Summary

- Reduces default `n_frames_per_iter` from 5 to 1, meaning HREX permutation moves happen after every 1 ps of MD with current defaults
- Runs MD trajectories for all states using a single context
- Adds `initial_volume_scale_factor` argument to `MonteCarloBarostat` constructor (defaulting to 0)
- Exposes all `MonteCarloBarostat` constructor args in the Python API
- Adds getter methods for integrator, potentials, and barostat to `Context`
- Uses a fixed barostat volume scale factor for HREX RBFE simulations, set to the average of the final scale factors from the bisection phase
- Removes defaults for constructor arguments from `custom_ops.MonteCarloBarostat` in favor of defining defaults in the Python API wrapper, `timemachine.lib.MonteCarloBarostat`

#### Todo

- [x] Blocked by #1156 

Benchmarks for hif2a single topology, solvent leg, 20 windows, 250 frames/window:
<table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">

<colgroup>
<col  class="org-left" />

<col  class="org-right" />

<col  class="org-right" />

<col  class="org-right" />

<col  class="org-left" />
</colgroup>
<thead>
<tr>
<th scope="col" class="org-left">&#xa0;</th>
<th scope="col" class="org-right">ns/day (master)</th>
<th scope="col" class="org-right">ns/day (<a href="https://github.com/proteneer/timemachine/pull/1146">#1146</a>)</th>
<th scope="col" class="org-right">ns/day (this PR)</th>
<th scope="col" class="org-left">notes</th>
</tr>
</thead>

<tbody>
<tr>
<td class="org-left">no HREX</td>
<td class="org-right">625.4</td>
<td class="org-right">-</td>
<td class="org-right">-</td>
<td class="org-left">no difference expected</td>
</tr>

<tr>
<td class="org-left">HREX, 5ps/move</td>
<td class="org-right">430.8</td>
<td class="org-right">578.4</td>
<td class="org-right">612.7</td>
<td class="org-left"></td>
</tr>

<tr>
<td class="org-left">HREX, 1ps/move</td>
<td class="org-right">194.5</td>
<td class="org-right">446.0</td>
<td class="org-right">498.0</td>
<td class="org-left">&#xa0;</td>
</tr>
</tbody>
</table>


timemachine/cpp/src/barostat.cu
```diff
@@ -82,7 +83,7 @@ MonteCarloBarostat<RealType>::MonteCarloBarostat(
     cudaSafeMalloc(&d_volume_scale_, 1 * sizeof(*d_volume_scale_));
     cudaSafeMalloc(&d_volume_delta_, 1 * sizeof(*d_volume_delta_));
 
-    gpuErrchk(cudaMemset(d_volume_scale_, 0, 1 * sizeof(*d_volume_scale_)));
+    gpuErrchk(cudaMemset(d_volume_scale_, initial_volume_scale_factor, 1 * sizeof(*d_volume_scale_)));
```
badisa, line 86: This needs to be a cudaMemcpy, cudaMemset is on raw bytes so this can become some random value.

https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1gf7338650f7683c51ee26aadc6973c63a

timemachine/fe/free_energy.py
```diff
@@ -865,25 +902,28 @@ def log_q_fn(replica_idx: ReplicaIdx, state_idx: StateIdx) -> float:
...
+            return xvb
 
-        ctxt = get_context(initial_state)
+        context.set_x_t(xvb.coords)
+        context.set_v_t(xvb.velocities)
+        context.set_box(xvb.box)
 
-        xs, boxes = ctxt.multiple_steps(md_params.n_eq_steps, store_x_interval=0)
-        assert len(xs) == len(boxes) == 1
+        context.get_potentials()[0].set_params(params)
```
badisa, line 913: Should we instead do `potential.unbound_potential.set_params`? Seems more explicit to access the single potential that we constructed than to access the 'first' potential on the context, or perhaps we should assert that the context only has one potential?

timemachine/fe/rbfe.py
```diff
@@ -695,31 +696,52 @@ def estimate_relative_free_energy_bisection_hrex(
...
-                trajectories_by_state,
+            replace(
+                initial_state,
+                x0=traj.frames[-1],
+                v0=traj.final_velocities,
+                box0=traj.boxes[-1],
+                barostat=replace(
+                    initial_state.barostat,
+                    adaptive_scaling_enabled=False,
+                    initial_volume_scale_factor=mean_final_barostat_volume_scale_factor,
```
badisa, line 733: nit: Looks like `mean_final_barostat_volume_scale_factor` can be None, but the Barostat doesn't allow that as an optional field. Though might be missing something, as it would seem like mypy would catch this.

badisa (changes requested):
The cudaMemset needs to be changed to cudaMemcpy, otherwise looks great!

timemachine/cpp/src/barostat.cu
badisa: Might be good to add a test for this as well

timemachine/fe/rbfe.py
```diff
@@ -695,31 +696,52 @@ def estimate_relative_free_energy_bisection_hrex(
...
 
-        # Use equilibrated samples from bisection as initial states
+        def get_mean_final_barostat_volume_scale_factor(trajectories_by_state: Iterable[Trajectory]) -> Optional[float]:
+            scale_factors = [traj.final_barostat_volume_scale_factor for traj in trajectories_by_state]
+            if any(x is not None for x in scale_factors):
+                assert all(x is not None for x in scale_factors)
+                sfs = cast(List[float], scale_factors)  # implied by assertion but required by mypy
+                return float(np.mean(sfs))
+            else:
+                return None
```
badisa, line 718: Should this None be allowed? We disable the scaling and set the value to None, does that mean we just expect to shut off the barostat in this case?

Perhaps we should at least throw a warning as this is likely a non-desirable configuration?

timemachine/fe/free_energy.py
```diff
@@ -455,7 +439,39 @@ def run_production_local_steps(n_steps: int) -> Tuple[NDArray, NDArray, NDArray]
...
+def sample(initial_state: InitialState, md_params: MDParams, max_buffer_frames: int) -> Trajectory:
+    """Generate a trajectory given an initial state and a simulation protocol
+
+    Parameters
+    ----------
+    initial_state: InitialState
+        (contains potentials, integrator, optional barostat)
+
+    md_params: MDParams
+        (specifies x0, v0, box0, number of MD steps, thinning interval, etc...)
```
maxentile, line 456: nit[docs] -- I know this wasn't added in this PR, but the descriptions are out of date and can be removed ((x0, v0, box0) are in initial_state, etc., type annotation seems sufficient)

timemachine/lib/__init__.py
```diff
@@ -1,46 +1,62 @@
...
+    adaptive_scaling_enabled: bool = True
+    initial_volume_scale_factor: float = 0.0
 
     def impl(self, bound_potentials):
         return custom_ops.MonteCarloBarostat(
-            self.N, self.pressure, self.temperature, self.group_idxs, self.interval, bound_potentials, self.seed
+            self.N,
+            self.pressure,
+            self.temperature,
+            [g.tolist() for g in self.group_idxs],
```
maxentile, line 56: [nit, type surprise] -- is the `tolist()` conversion important? (was this previously brittle w.r.t. np.int32 vs. np.int64 type, and if so, would this be a useful change in other contexts, e.g. valence idxs?)

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -1219,7 +1234,8 @@ void declare_barostat(py::module &m) {
             py::arg("frequency"),
             py::arg("bps"),
             py::arg("seed"),
-            py::arg("adaptive_scaling_enabled") = true)
+            py::arg("adaptive_scaling_enabled") = true,
+            py::arg("initial_volume_scale_factor") = 0.0)
```
maxentile, line 1238: possible doc or line comment to pre-empt worry: "0.0" is a flag meaning "use 1% of first box volume" ( https://github.com/proteneer/timemachine/pull/1152#discussion_r1333509394 )

timemachine/fe/rbfe.py
```diff
@@ -695,31 +696,52 @@ def estimate_relative_free_energy_bisection_hrex(
...
+                    adaptive_scaling_enabled=False,
+                    initial_volume_scale_factor=mean_final_barostat_volume_scale_factor,
+                )
+                if initial_state.barostat
+                else None,
             )
+            for initial_state, traj in zip(results[-1].initial_states, trajectories_by_state)
         ]
 
         pair_bar_result, trajectories_by_state, diagnostics = run_sims_hrex(
```
maxentile, line 741: [nit, future diagnostic] possibly log the accept rate of non-adaptive MC barostat by window for production sims, warn if outside range [25%, 75%].

maxentile (approved):
Looks good!

(superficial line comments, + it might be useful to log MC barostat accept rate when run with `adapt_volume_scale_factor = False`)

timemachine/cpp/src/barostat.cu
mcwitt: Good catch! Added a failing assertion in https://github.com/proteneer/timemachine/pull/1151/commits/5d161e4af77e6e6f3053af6abb43b9a25c75719a and fixed in https://github.com/proteneer/timemachine/pull/1151/commits/94225e266d3ef3fe1cca0f985431af4d2c7da1a0

badisa (approved):

timemachine/fe/free_energy.py
mcwitt: Hmm, I think this is how it was before https://github.com/proteneer/timemachine/pull/1151/commits/78b89f7bf31bc3b9b913be2e48b6d537f0049a9f? I found it more explicit to access the potential via the context to make it clear that they are linked, but can sort of see your argument too. Can add an assertion that `len(context.get_potentials()) == 1`.

timemachine/fe/rbfe.py
mcwitt: Ah, good catch. I would've expected mypy to catch this too. It seems like this was addressed in a recent version ([PR](https://github.com/python/mypy/pull/14849)).

In practice, I don't think there's an issue here, because `mean_final_barostat_volume_scale_factor` should only be `None` if none of the states have a barostat defined, but will fix the types to make this explicit.

timemachine/fe/rbfe.py
mcwitt: I think `None` is valid and expected in the case of a vacuum simulation. It would be invalid if any of the states have a barostat defined, but that should be ruled out in this branch. Will try to think of ways to make this clearer.

timemachine/lib/__init__.py
mcwitt: The type of the `group_idxs` in the C++ API is `List[List[int]]`, although the wrapper seems to accept `NDArray` as well. Previously this was not causing any static type errors because the arguments of `timemachine.lib.MonteCarloBarostat` were not annotated, but this issue surfaced when I converted it to a dataclass.

timemachine/fe/rbfe.py
mcwitt: Addressed by making `initial_volume_scale_factor` optional in `timemachine.lib.MonteCarloBarostat`. The default value of `None` initializes the C++ implementation with `0.0`.

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -1219,7 +1234,8 @@ void declare_barostat(py::module &m) {
             py::arg("frequency"),
             py::arg("bps"),
             py::arg("seed"),
-            py::arg("adaptive_scaling_enabled") = true)
```
badisa, line 1222: Is there a need to remove this default? It means any old code that constructs a barostat needs to be changed, even though the default behavior won't have changed

timemachine/fe/free_energy.py
mcwitt: Fixed in https://github.com/proteneer/timemachine/pull/1151/commits/0a10e54faa4838208ef51a6ea6fa1894d5c73645

timemachine/cpp/src/wrap_kernels.cpp
mcwitt: My idea was to have a single source of truth for defaults, to avoid specifying them here and also in `timemachine.lib.MonteCarloBarostat`, which seemed like it might be confusing and get out of sync. The reason to prefer setting them in the Python API was to allow making arguments optional, so the user can pass `None` to use the default value.

This should only require changing code that calls `custom_ops.MonteCarloBarostat` directly; I think in most cases we should be using the Python API wrapper, `timemachine.lib.MonteCarloBarostat`? (usage of the latter should not need changes)

badisa: That seems like a reasonable rationale. Would prefer in the future that these sorts of changes be made in separate PRs where that intention is communicated rather than at the tail of PR review. Was surprising to see you change this after approvals as it does change how the code base can be used.

mcwitt: Yeah, fair, sorry about that. This wasn't a change out of the blue, but was intended to address https://github.com/proteneer/timemachine/pull/1151#discussion_r1336112633.

badisa: I understand the intent of the change, however the scope of the change is beyond what the two comments implied to me.

No action needs to be taken in this case, and for the future would be good to avoid such changes at this stage of a PR or to more clearly point out this sort of API change.

mcwitt: OK, cool. I'm also happy to add the defaults back to `wrap_kernels.cpp`, and isolate the API change in a separate PR if that sounds preferable.

timemachine/fe/free_energy.py
mcwitt: Added assertion in https://github.com/proteneer/timemachine/pull/1151/commits/cae4a91c998abe2794f41698ecf698fa70a04f81

timemachine/fe/rbfe.py
mcwitt: Added assertions to clarify this in https://github.com/proteneer/timemachine/pull/1151/commits/e56be2b87703906551740bacc368be1c528debb8

tests/test_benchmark_free_energy.py
```diff
@@ -77,8 +77,7 @@ def test_benchmark_hif2a_single_topology(hif2a_single_topology_leg, enable_hrex)
             run_sims_hrex,
             initial_states,
             md_params,
-            n_frames_per_iter=5,
-            temperature=temperature,
+            n_frames_per_iter=1,
```
proteneer, line 80: why is temperature removed?

timemachine/cpp/src/wrap_kernels.cpp
proteneer: Removing defaults is probably fine for this PR - I'd actually much rather this be made explicit in this PR (since we're clearly changing default behavior).

tests/test_benchmark_free_energy.py
mcwitt: Temperature is defined at the level of initial states for HREX. (Although we currently restrict the temperature to be the same for all states: https://github.com/proteneer/timemachine/blob/5951327be3b4e71a3325d0f1188d8c26ea1757a7/timemachine/fe/free_energy.py#L817-L821)

timemachine/fe/free_energy.py
```diff
@@ -835,23 +847,48 @@ def assert_ensembles_compatible(state_a: InitialState, state_b: InitialState):
...
+            intg_impl,
+            bound_impls,
+            baro_impl,
+        )
+
+        return potential, context
 
-    def compute_log_q_matrix(xvbs: List[CoordsVelBox], params: List[NDArray]):
+    # Set up overall potential and context using the first state.
+    # NOTE: this assumes that states differ only in their parameters, but we do not check this!
```
proteneer, line 873: tbh I think this is a reasonable approach, I don't see us doing TREX/PREX where the temperature/pressure will change in the standard way since the energies are size extensive.

timemachine/cpp/src/wrap_kernels.cpp
mcwitt: Added an item to the PR description

proteneer (approved):

PR #1154, opened 2023-09-21
Re-image molecules during barostat moves.

proteneer:
This is a hacky, but extremely simple way of implementing periodic re-imaging of the coordinates. Notes:

0) Re-uses most of the existing internals of the barostat for keeping track of molecule groups and centroid computations.
1) Only works in NPT when you have a barostat turned on (and moves that are periodically rejected).
2) Is extremely fast (no loss in performance), and done fairly frequently.

(Note that this is one of those tech-debt PRs that isn't perfect, but is a strict improvement over the current state of things, and is relatively safe and easy to implement).


timemachine/cpp/src/kernels/k_barostat.cuh
```diff
@@ -41,9 +41,20 @@ void __global__ k_rescale_positions(
...
 
-        coords[atom_idx * 3 + 0] += displacement_x;
-        coords[atom_idx * 3 + 1] += displacement_y;
-        coords[atom_idx * 3 + 2] += displacement_z;
+        // centroid of the new molecule
+        centroid_x += displacement_x;
+        centroid_y += displacement_y;
+        centroid_z += displacement_z;
+
+        // compute displacement needed to shift centroid back into the scaled homebox
```
proteneer, line 49: Actually I just realized it's not safe to use scaled box here - since its only written by the 0th thread.

proteneer: (so instead we compute it from box * scale)

badisa (approved):
> Note that this is one of those tech-debt PRs that isn't perfect, but is a strict improvement over the current state of things, and is relatively safe and easy to implement

This statement is a perfect summary of this PR. It is tech debt, but a small, well isolated amount that can be easily removed when we solve this more generally.

jkausrelay:
nit: Add a test for the Barostat changes (maybe x0 outside of the box, confirm the ene is the same as x0 in the box and the output x is also in the box).

Othwerise lgtm.

proteneer:
Adding a test now to the MC barostat class.

tests/test_barostat.py
```diff
@@ -397,6 +398,93 @@ def test_barostat_varying_pressure():
     assert compute_box_volume(atm_box) > ten_atm_box_vol
 
 
+# test that barostat only proposes properly re-centered coordinates
+def test_barostat_recentering_upon_acceptance():
```
proteneer, line 402: @jkausrelay  I added this extra test that verifies the barostat only proposes molecular configurations whose centroids are recentered into the homebox. lmk if this is sufficient for approval (or if you'd like more tests).

jkausrelay (approved):

PR #1156, opened 2023-09-22
Clean up `fe.free_energy` function signatures with `Trajectory` type

mcwitt:
In `timemachine.fe.free_energy` and code that uses this module, we commonly pass around data of the type `Tuple[StoredArrays, NDArray, NDArray]`, representing the frames, boxes, and final velocities from an MD trajectory. This PR introduces a type `Trajectory` to store these data, simplifying function signatures and enabling additional validation.

In the future, this will simplify extending the data we return from MD sampling; for example, an upcoming PR will add `barostat_volume_scale_factors`.

Note: this also simplifies `timemachine.fe.free_energy.sample` to require the `max_buffer_frames` argument and always return frames as `StoredArrays`.

### DIFF
--- a/tests/test_free_energy.py
+++ b/tests/test_free_energy.py
@@ -229,9 +229,9 @@ def test_sample_max_buffer_frames_with_local_md(
     n_eq_steps = 1
 
     md_params = MDParams(n_frames, n_eq_steps, steps_per_frame, 2023, local_steps=local_steps)
-    frames, _, _ = sample(solvent_hif2a_ligand_pair_single_topology_lam0_state, md_params, max_buffer_frames)
-    assert isinstance(frames, StoredArrays)
-    assert len(frames) == n_frames
+    traj = sample(solvent_hif2a_ligand_pair_single_topology_lam0_state, md_params, max_buffer_frames)
+    assert isinstance(traj.frames, StoredArrays)
+    assert len(traj.frames) == n_frames
 
 
 @given(integers(min_value=1))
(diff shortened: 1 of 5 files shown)


tests/test_free_energy.py
```diff
@@ -260,26 +258,6 @@ def hif2a_ligand_pair_single_topology_lam0_state():
...
-    )
-
-    assert isinstance(frames_ref, np.ndarray)
-    assert isinstance(frames_test, StoredArrays)
-    assert len(frames_ref) == len(frames_test)
-
-    for frame_ref, frame_test in zip(frames_ref, frames_test):
-        np.testing.assert_array_equal(frame_ref, frame_test)
-
-
```
mcwitt, line 282: This was testing for consistency when we pass `max_buffer_frames` vs. when we don't; it is no longer needed now that `max_buffer_frames` is required.

timemachine/fe/free_energy.py
```diff
@@ -134,16 +134,47 @@ def u_kln_by_component_by_lambda(self) -> NDArray:
...
     intermediate_results: List[PairBarResult]
     hrex_diagnostics: Optional[HREXDiagnostics] = None
 
+    @property
+    def frames(self) -> List[StoredArrays]:
+        return [traj.frames for traj in self.trajectories]
+
+    @property
+    def boxes(self) -> List[NDArray]:
+        return [traj.boxes for traj in self.trajectories]
```
mcwitt, line 176: These properties maintain backwards compatibility

jkausrelay (approved):

maxentile (approved):

PR #1157, opened 2023-09-25
Misc improvements to HREX plots

mcwitt:
* Adds `annotate_threshold: int` argument, defaulting to 20; if `n_states` is larger than this, text annotations will be omitted (prevents overflowing text for large numbers of states)
* Formats annotations as percentages by default, allowing override (saves space in annotations)

### DIFF
--- a/timemachine/fe/plots.py
+++ b/timemachine/fe/plots.py
@@ -1,6 +1,6 @@
 import io
 import warnings
-from typing import Tuple
+from typing import Callable, Literal, Tuple, cast
 
 import numpy as np
 from matplotlib import pyplot as plt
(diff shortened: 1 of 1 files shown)


timemachine/fe/plots.py
```diff
@@ -335,12 +338,15 @@ def plot_hrex_transition_matrix(
     fig, ax = plt.subplots(figsize=figsize)
     p = ax.pcolormesh(states, states, transition_rate)
 
-    if annotate:
+    # Skip text annotations if "auto" and large number of states
+    annotate_ = n_states <= 20 if annotate == "auto" else annotate
```
jkausrelay, line 342: nit: Make 20 a constant to reuse below

jkausrelay (approved):

timemachine/fe/plots.py
```diff
@@ -402,11 +411,14 @@ def plot_hrex_replica_state_distribution_heatmap(
     fig, ax = plt.subplots(figsize=figsize)
     p = ax.pcolormesh(replicas, states, fraction_by_replica_by_state)
 
-    if annotate:
+    # Skip text annotations if "auto" and large number of states
+    annotate_ = n_states <= 20 if annotate == "auto" else annotate
```
maxentile, line 415: nit: "auto" behavior does not depend on figsize, and there is 2x duplication of n_states threshold between HREX heatmap plots... not for this PR, but possible future enhancement: extract a function "plot_annotated_heatmap(matrix, figsize, ...)" or similar

maxentile (approved):
Nice -- should improve figure clarity at large n_states!

timemachine/fe/plots.py
mcwitt: Changed the behavior in https://github.com/proteneer/timemachine/pull/1157/commits/bd48f8d2c13dc4fa32cb92012b14cb14cbf5f364 to allow overriding the default threshold, to avoid hardcoding implicit dependence on figsize.

PR #1139, opened 2023-09-06
Hamiltonian Replica Exchange of Water Sampling

proteneer:
This PR implements a reference calculation of water occupancies a buckyball using partial decoupling of a buckyball system under HREX. This is mainly used to ensure we can comparable statistics to the BiasedDeletion MC moves.

PR #1158, opened 2023-09-25
Hamiltonian Replica Exchange of Water Sampling

proteneer:
Summary:

A reference implementation of water sampling via replica-exchange. Unifies some shared code between the MonteCarlo based water sampler and the HREX based sampler. Asserts self-consistency reasonably well:

```
N=1: 53.5% and N=2: 46.5% (Unbiased Deletion, Targeted Insertion)
N=1: 52.2% and N=2: 47.7% (Biased Deletion, Targeted Insertion)
N=1: 51.5% and N=2: 48.4% (HREX, no MC moves at all)
```

 

Note:

Rebased version of #1139 (adapted).


proteneer:
Add @maxentile and @mcwitt for a clerical approval. This mainly cleans-up the duplicate code, re-uses the HREX code in a way that preserves velocities, and now actually computes correct statistics.

examples/water_sampling_common.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def compute_occupancy(x_t, box_t, ligand_idxs, threshold):
+    # compute the number of waters inside the buckyball ligand
+    num_ligand_atoms = len(ligand_idxs)
+    num_host_atoms = len(x_t) - num_ligand_atoms
+    host_coords = x_t[:num_host_atoms]
+    ligand_coords = x_t[num_host_atoms:]
+    ligand_centroid = np.mean(ligand_coords, axis=0)
+    count = 0
```
maxentile, line 86: nit: delete

examples/water_sampling_common.py
```diff
@@ -0,0 +1,166 @@
...
+
+    bt = BaseTopology(mol, ff)
+    afe = AbsoluteFreeEnergy(mol, bt)
+    host_config = HostConfig(solvent_sys, solvent_conf, solvent_box, num_water_atoms)
+    potentials, params, combined_masses = afe.prepare_host_edge(ff.get_params(), host_config, lamb)
+
+    ligand_idxs = np.arange(num_water_atoms, num_water_atoms + mol.GetNumAtoms())
+
+    # modify state with all ligand atoms decoupled (returned by prepare_host_edge) to
+    # state with only half of the ligand atoms decoupled. This roughly decouples a contingous
```
maxentile, line 112: nit: spelling

maxentile, line 112: nit: out of date description

examples/water_sampling_hrex.py
```diff
@@ -0,0 +1,225 @@
...
+        n_frames_per_iter=n_frames_per_iter,
+    )
+
+    plots = make_pair_bar_plots(pair_bar_result, temperature, combined_prefix)
+
+    return SimulationResult(pair_bar_result, plots, trajectories_by_state, md_params, None, diagnostics)
+
+
+def test_hrex():
+    parser = argparse.ArgumentParser(description="Test the exchange protocol in a box of water.")
```
maxentile, line 118: nit: out of date description

examples/water_sampling_hrex.py
```diff
@@ -0,0 +1,225 @@
...
+    combined_prefix = "hrex"
+
+    def make_optimized_initial_state(lamb):
+        state, _, _ = get_initial_state(water_pdb, mol, ff, seed, nb_cutoff, lamb)
+        return state
+
+    print("Starting bi-section")
+
+    md_params_bisection = replace(md_params, n_frames=n_frames_bisection)
+    # results_bisection, frames_by_state, boxes_by_state, final_velocities_by_state = run_sims_bisection(
```
maxentile, line 61: nit: delete

examples/water_sampling_hrex.py
```diff
@@ -0,0 +1,225 @@
...
+    )
+
+    def get_mean_final_barostat_volume_scale_factor(trajectories_by_state: Iterable[Trajectory]) -> Optional[float]:
+        scale_factors = [traj.final_barostat_volume_scale_factor for traj in trajectories_by_state]
+        if any(x is not None for x in scale_factors):
+            assert all(x is not None for x in scale_factors)
+            sfs = cast(List[float], scale_factors)  # implied by assertion but required by mypy
+            return float(np.mean(sfs))
+        else:
+            return None
```
maxentile, line 78: nit: duplicated with rbfe.py https://github.com/proteneer/timemachine/blob/71319ef09595b43c2c9ce5565c871bbcaa199983/timemachine/fe/rbfe.py#L738-L745

examples/water_sampling_common.py
```diff
@@ -0,0 +1,166 @@
...
+
+    host_bps.append(potentials[-1].bind(nb_params))
+
+    # (YTZ): This is disabled because the initial ligand and starting waters are pre-minimized
+    # print("Minimizing the host system...", end="", flush=True)
+    # host_conf = minimize_host_4d([mol], host_config, ff)
+    # print("Done", flush=True)
+
+    temperature = DEFAULT_TEMP
+    dt = 1.0e-3
```
maxentile, line 151: surprise: smaller timestep than default -- add note whether this is due to observed stability issues with default settings, intentionally disabling HMR for simplicity, general cautiousness, or some other reason

maxentile (approved):
>Add @maxentile and @mcwitt for a clerical approval. 

Looks good!

examples/water_sampling_common.py
```diff
@@ -0,0 +1,166 @@
...
+    # forms a hole on the "flat" side of the buckyball, 1-indexed
+    # previously failed attempts:
+    # 1) uniformly decoupling half of the buckyball
+    # 2) uniformly decoupling entire buckyball
+    # 3) uniformly shrinking vdw radii of every buckyball atom
+    # 4) uniformly shrinking vdw radii of half of the buckyball atoms
+    lambda_coupled_ligand_atoms = num_water_atoms + np.array([10, 11, 16, 17, 21, 22, 12, 14, 13, 25, 24, 23, 20]) - 1
+    fully_coupled_ligand_atoms = np.array(
+        list(set(ligand_idxs.tolist()).difference(lambda_coupled_ligand_atoms.tolist()))
+    )
```
mcwitt, line 133: nit: consider cleaning with `setdiff1d`

examples/water_sampling_common.py
```diff
@@ -0,0 +1,166 @@
...
+
+    ligand_water_params[fully_coupled_ligand_atoms, -1] = 0
+    nb_params[ligand_water_flat_idxs] = ligand_water_params.reshape(-1)
+
+    # override last potential with the updated nb_params
+    host_bps = []
+    for p, bp in zip(params[:-1], potentials[:-1]):
```
mcwitt, line 140: nit: confusing name (`bp` not a bound potential)

examples/water_sampling_common.py
```diff
@@ -0,0 +1,166 @@
...
+    group_idxs = get_group_indices(bond_list, len(final_masses))
+    barostat = MonteCarloBarostat(
+        len(final_masses), DEFAULT_PRESSURE, temperature, group_idxs, barostat_interval, seed + 1
+    )
+
+    initial_state = InitialState(
+        host_bps, integrator, barostat, final_conf, np.zeros_like(final_conf), solvent_box, lamb, ligand_idxs
+    )
+
+    num_water_mols = num_water_atoms // 3
```
mcwitt, line 164: nit: worth asserting that `num_water_atoms % 3 == 0`?

examples/water_sampling_hrex.py
```diff
@@ -0,0 +1,225 @@
...
+
+    pair_bar_result, trajectories_by_state, diagnostics = run_sims_hrex(
+        initial_states_hrex,
+        replace(md_params, n_eq_steps=0),  # using pre-equilibrated samples
+        n_frames_per_iter=n_frames_per_iter,
+    )
+
+    plots = make_pair_bar_plots(pair_bar_result, temperature, combined_prefix)
+
+    return SimulationResult(pair_bar_result, plots, trajectories_by_state, md_params, None, diagnostics)
```
mcwitt, line 114: nit: type mismatch
```suggestion
    return SimulationResult(pair_bar_result, plots, trajectories_by_state, md_params, [], diagnostics)
```

examples/water_sampling_hrex.py
```diff
@@ -0,0 +1,225 @@
...
+        print("raw occs", occs)
+        print("lambda", lamb, "occupancies", dict(zip(unique, counts)))
+
+    plt.legend()
+    plt.xlabel("frame")
+    plt.ylabel("occupancy")
+    plt.show()
+
+    plot_hrex_swap_acceptance_rates_convergence(sim_res.hrex_diagnostics.cumulative_swap_acceptance_rates)
+    # plot_hrex_swap_acceptance_rates_convergence(sim_res.hrex_diagnostics.instantaneous_swap_acceptance_rates)
```
mcwitt, line 212: nit: stale; delete?

examples/water_sampling_common.py
```diff
@@ -0,0 +1,166 @@
...
+    # (YTZ): This is disabled because the initial ligand and starting waters are pre-minimized
+    # print("Minimizing the host system...", end="", flush=True)
+    # host_conf = minimize_host_4d([mol], host_config, ff)
+    # print("Done", flush=True)
+
+    temperature = DEFAULT_TEMP
+    dt = 1.0e-3
+    barostat_interval = 25
+    integrator = LangevinIntegrator(temperature, dt, 1.0, final_masses, seed)
+    bond_list = get_bond_list(host_bps[0].potential)
```
mcwitt, line 154: nit: do we want to move away from assuming an ordering of the potentials?
```suggestion
    bond_list = get_bond_list(next(bp for bp in host_bps if isinstance(bp, HarmonicBond)).potential)
```

examples/water_sampling_hrex.py
```diff
@@ -0,0 +1,225 @@
...
+
+        plt.plot(occs, label=f"{lamb:.2f}", alpha=alpha, linewidth=lw)
+        unique, counts = np.unique(occs, return_counts=True)
+        print("raw occs", occs)
+        print("lambda", lamb, "occupancies", dict(zip(unique, counts)))
+
+    plt.legend()
+    plt.xlabel("frame")
+    plt.ylabel("occupancy")
+    plt.show()
```
mcwitt, line 209: nit: consider moving plotting code into dedicated function

mcwitt (approved):

examples/water_sampling_hrex.py
proteneer: willfix

examples/water_sampling_common.py
mcwitt: Sorry, just noticed that my suggested edit was wrong. I think it should be

```python
bond_list = get_bond_list(next(bp for bp in host_bps if isinstance(bp.potential, HarmonicBond)).potential)
```

timemachine/fe/rbfe.py
```diff
@@ -775,6 +688,16 @@ def get_mean_final_barostat_volume_scale_factor(trajectories_by_state: Iterable[
         plots = make_pair_bar_plots(pair_bar_result, temperature, combined_prefix)
 
         stored_trajectories = []
+
+        if keep_idxs is None:
+            keep_idxs = [0, -1]  # keep frames from first and last windows
+        elif keep_idxs is np.inf:
```
proteneer, line 694: @mcwitt I had to modify this logic a bit to support another indicator value for `keep_idxs` - since we don't know the # of actual initial states (as it's determined by HREX), we instead need a way of saying "keep everything". `None` is already reserved for first and last windows, so I picked `np.inf` instead. Is this too hacky?

mcwitt: Agreed that a special value for "all" is needed. How about just the string `"all"`? Can indicate this in the type as `Optional[List[int] | Literal["all"]]`.

Alternatively, maybe it makes more sense to reserve `None` for all states, and change the default value to `(0, -1)`?

proteneer: Okay I can try that - I'm hesitant about changing the default since it's API breaking. I'm not sure if upstream code depends on this in production already.

mcwitt: To be clear, I'm not suggesting changing the default behavior if `keep_idxs` is not specified -- that would still be to return frames for the first and last states. It would just be to change the meaning of `None` (which is still potentially API-breaking).

proteneer: ugh - just realized we use `keep_idxs` for all the other `run_*` functions. Not sure if we want to make this change consistently or not.

examples/water_sampling_hrex.py
proteneer: I think I've addressed this. I've refactored the original `estimate_relative_free_energy_bisection_hrex` into an `impl` method instead, so instead we just call the `impl` directly and now we D.R.Y.

timemachine/fe/rbfe.py
proteneer: (changed to "all" per suggestion)

examples/water_sampling_common.py
proteneer: re-running tests with HMR.

PR #1159, opened 2023-09-26
Make tests more robust to noise

badisa:
As a part of  https://github.com/proteneer/timemachine/pull/1155 I noticed two tests that were sensitive to the randomness of the system and this PR modifies them to be more robust to the seed.

Also fixes a bit of test code that @maxentile noticed was incorrect. 

- Increase tolerance of local MD test that fails with different seeds
- Use an obviously bad barostat scaling factor to ensure it adapts for the test
- Correctly compute distance between particles rather than the displacement in one dimension.


tests/test_barostat.py
```diff
@@ -606,9 +606,12 @@ def test_barostat_scaling_behavior():
     scaling = baro.get_volume_scale_factor()
     assert scaling > 0
 
+    # Set to an intentionally bad factor to ensure it adapts
+    bad_scaling_factor = 0.5 * compute_box_volume(host_box)
+    baro.set_volume_scale_factor(bad_scaling_factor)
     ctxt.multiple_steps(100)
     # The scaling should adapt between moves
-    assert scaling != baro.get_volume_scale_factor()
+    assert bad_scaling_factor != baro.get_volume_scale_factor()
```
maxentile, line 614: nit: could strengthen to:
(1) assert `baro.get_volume_scale_factor() == bad_scaling_factor` prior to `ctxt.multiple_steps`,
(2) assert `baro.get_volume_scale_factor() < bad_scaling_factor` (rather than just `!=`) after `ctxt.multiple_steps`

tests/test_hilbert_sort.py
```diff
@@ -24,7 +24,7 @@ def get_max_block_distances(coords: NDArray, box: NDArray, block_size: int) -> N
         start_idx = bidx * block_size
         end_idx = min((bidx + 1) * block_size, N)
         block = coords[start_idx:end_idx]
-        block_distances.append(np.max(delta_r(block[:, None], block[None, :], box=box)))
+        block_distances.append(np.max(np.linalg.norm(delta_r(block[:, None], block[None, :], box=box), axis=-1)))
```
maxentile, line 27: +1, good fix, didn't realize it was also in this file

maxentile (approved):
Nice improvements!

tests/test_barostat.py
badisa: Strengthened in https://github.com/proteneer/timemachine/pull/1159/commits/4f8b877ec86d7709e9039d5ef737b1a2b72de532

tests/test_hilbert_sort.py
mcwitt: This still looks fishy to me -- `np.linalg.norm` without an axis specified returns a scalar, so it doesn't seem like `np.max` is doing anything? I might be confused, but seems like it would be useful to annotate some of the array shapes with comments.

mcwitt: Never mind, sorry. I misread the parentheses. Looks good to me.

mcwitt (approved):

PR #1162, opened 2023-09-27
Avoid storing velocities for intermediate frames in HREX

mcwitt:
- Currently HREX unnecessarily uses O(n_iterations) memory to store velocities from intermediate frames, when only the final velocities are returned
- This PR fixes the accumulation of trajectories for each state to update velocities in place, avoiding memory usage for intermediate frames that will be discarded
- Changes the representation of `boxes` in `Trajectory` from `NDArray` to `List[NDArray]`. This allows for more efficient concatenation of intermediate trajectories and is more consistent with `frames`.
- Adds a probabilistic test asserting that memory usage is constant during HREX iterations

### DIFF
--- a/tests/hrex/test_hrex_rbfe.py
+++ b/tests/hrex/test_hrex_rbfe.py
@@ -1,12 +1,15 @@
 from dataclasses import replace
 from importlib import resources
 from typing import Optional
+from unittest.mock import patch
 
 import matplotlib.pyplot as plt
 import numpy as np
 import pytest
+from psutil import Process
+from scipy import stats
 
-from timemachine.fe.free_energy import HostConfig, HREXParams, MDParams, SimulationResult
+from timemachine.fe.free_energy import HostConfig, HREXParams, MDParams, SimulationResult, sample_with_context
 from timemachine.fe.plots import (
     plot_hrex_replica_state_distribution_convergence,
     plot_hrex_replica_state_distribution_heatmap,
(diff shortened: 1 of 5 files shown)


timemachine/fe/free_energy.py
```diff
@@ -170,28 +170,29 @@ def u_kln_by_component_by_lambda(self) -> NDArray:
...
+        self.frames.extend(other.frames)
+        self._boxes.extend(other._boxes)
+        self.final_velocities = other.final_velocities
+        self.final_barostat_volume_scale_factor = other.final_barostat_volume_scale_factor
 
-        return Trajectory(frames, boxes, final_velocities, final_barostat_volume_scale_factor)
+    # For backwards compatibility
+    @property
+    def boxes(self) -> NDArray:
+        return np.array(self._boxes)
```
mcwitt, line 195: This seemed prudent, but may not be necessary. I didn't immediately find downstream code that depends on this being an ndarray.

proteneer:
Can we add a test that `run_sims_hrex` uses constant memory?

mcwitt:
> Can we add a test that `run_sims_hrex` uses constant memory?

Added in https://github.com/proteneer/timemachine/pull/1162/commits/8a138cece416df1004b9bed474dad4d95b859963. Verified that this fails when run in master.

proteneer (approved):

tests/hrex/test_hrex_rbfe.py
```diff
@@ -57,16 +60,30 @@ def test_hrex_rbfe_hif2a(hif2a_single_topology_leg):
...
+            md_params,
+            lambda_interval=(0.0, 0.15),
+            n_windows=n_windows,
+        )
+
+    # Check that memory usage is not increasing
+    rss_traj = rss_traj[10:]  # discard initial transients
+    rss_diff_count = np.sum(np.diff(rss_traj) != 0)
+    rss_increase_count = np.sum(np.diff(rss_traj) > 0)
+    assert stats.binom.pmf(rss_increase_count, n=rss_diff_count, p=0.5) >= 0.001
```
maxentile, line 86: [nit, non-blocking] I think this detects the scenario we want, and this looks quite robust, very unlikely to lead to false positives, but I was just curious what kinds of corner cases it might miss. Two examples come to mind:

```python
corner_case_traj_false_negative = np.cumsum([[-0.1, +1]]*100)
corner_case_traj_false_positive = np.cumsum([[-0.1, +0.1]]*1000000)
```
image
`diff_count = 189, increase_count = 95, binom_pmf=0.05780851636953422`

image
`diff_count = 1999989, increase_count = 999995, binom_pmf=0.0005641909235027816`

--------

```python
from scipy import stats
import numpy as np

def assert_non_increasing(rss_traj, verbose=True):
    rss_traj = rss_traj[10:]  # discard initial transients
    rss_diff_count = np.sum(np.diff(rss_traj) != 0)
    rss_increase_count = np.sum(np.diff(rss_traj) > 0)
    binom_pmf = stats.binom.pmf(rss_increase_count, n=rss_diff_count, p=0.5)
    if verbose:
        print(f"diff_count = {rss_diff_count}, increase_count = {rss_increase_count}, binom_pmf={binom_pmf}")
    assert binom_pmf >= 0.001
```

maxentile (approved):
Good fix!

And the approach to testing that memory usage is nearly constant over time looks robust, unlikely to raise false positives.

badisa (approved):
:rocket:

PR #1163, opened 2023-09-28
Reduce max interaction count for neighborlist

badisa:
- Can reduce the buffer size by a bit less than half and still cover all possible interactions
- No impact on performance
- Memory scaling is still O(N^2), but allocate only as much as we ever need rather than a full `(N^2)/32`
- Should follow up with a further refinement of max_ixn_count for IxnGroup potentials which never change atom_idxs.
- Saves about 20MB of GPU memory for Hif2a RBFE


proteneer (approved):
LGTM

tests/test_nblist.py
```diff
@@ -362,3 +362,21 @@ def test_neighborlist_on_subset_of_system():
...
         assert_ixn_lists_are_equal(reference_complete_ixns, test_ixn_list)
+
+
+@pytest.mark.parametrize("block_size", [32])
+@pytest.mark.parametrize("tiles", [2, 10, 100])
+def test_nblist_max_interactions(block_size, tiles):
+    """Verify that if all particles in the system interact, that the neighborlist correctly assigns large enough buffers"""
+    rng = np.random.default_rng(2023)
+
+    coords = rng.random(size=(block_size * tiles, 3))
```
maxentile, line 373: nit: should this be shrunk slightly, to ensure that max(pdist(coords)) < cutoff of 1.2?

(If generated from unit cube, then max distance is like 1.7 (for opposite corners (0,0,0), (1,1,1).)

EDIT: ahh, sorry, the cutoff is defined as 10.0 a few lines later -- for ease of skimming, might name cutoff = 10.0 a few lines earlier, but not necessary -- this is correct and readable as-is

timemachine/cpp/src/neighborlist.cu
```diff
@@ -356,8 +364,19 @@ template <typename RealType> int Neighborlist<RealType>::Y() const {
 
 template <typename RealType> int Neighborlist<RealType>::num_row_blocks() const { return ceil_divide(NR_, TILE_SIZE); }
 
+// max_ixn_count determines the number of tile to atom interaction counts. For each tile that interacts with another
+// it can have TILE_SIZE interactions. Note that d_ixn_count_ is only the number of tile interactions, and differs by a factor of TILE_SIZE
 template <typename RealType> int Neighborlist<RealType>::max_ixn_count() const {
-    return num_column_blocks() * num_row_blocks() * WARP_SIZE;
+    // The maximum number of interactions is that of all tiles interacting with all other tiles. Which
+    // is the upper triangular matrix.
```
maxentile, line 371: nit, comment clarity: had trouble parsing these comments on first read. the word"interaction" is quite overloaded here (atom-atom, atom-tile, tile-tile), might be helpful to be more explicit. (e.g. saying "tile-tile" or "atom-tile" instead of just "tile")

timemachine/cpp/src/neighborlist.cu
```diff
@@ -356,8 +364,19 @@ template <typename RealType> int Neighborlist<RealType>::Y() const {
...
+    // Use the maximum NxN matrix to allocate to allow supporting any combination of atom idxs
+    const int n_blocks = ceil_divide(max_size_, TILE_SIZE);
```
maxentile, line 373: Looks good, seems conservative: don't think this would be brittle w.r.t. any input geometry, choice of w_coords, or choice of cutoff.

(In the future, hard-coding any smaller of an allocation in this way could be brittle though. If we want to reduce this further in the future, might need to consider something more dynamic, maybe adding a .resize() method or something, so we could start with a small allocation that is not wasteful for instances of typical particle density at default cutoff, but be able to replace it on the fly with a larger allocation on atypical inputs...)

maxentile (approved):
Good optimization!

timemachine/cpp/src/neighborlist.cu
```diff
@@ -356,8 +364,19 @@ template <typename RealType> int Neighborlist<RealType>::Y() const {
...
+    // Compute the size of the matrix without the diagonal
+    int max_tile_interactions = (n_blocks * n_blocks) - n_blocks;
+    // Divide by two to get the upper, then add back in the size of the diagonal
+    max_tile_interactions = ceil_divide(max_tile_interactions, 2) + n_blocks;
```
mcwitt, line 377: nit: is `ceil_divide` needed here? I think $n^2 - n = n(n-1)$ should always be even?

mcwitt: (then I think this could simplify to `max_tile_interactions = n_blocks * (n_blocks + 1) / 2`, but the way you have it commented is also nice)

mcwitt (approved):
Looks good, nice optimization!

timemachine/cpp/src/neighborlist.cu
badisa: Simplified the logic and removed the unnecessary `ceil_divide` in https://github.com/proteneer/timemachine/pull/1163/commits/f7634b031660b8376ec7e104aed3097c3328793a

timemachine/cpp/src/neighborlist.cu
badisa: One potential thing we can do to save memory is to constrain the memory usage for the IxnGroup potentials, as that is N^2 while it often doesn't need to be.

timemachine/cpp/src/neighborlist.cu
badisa: Tried to clarify in https://github.com/proteneer/timemachine/pull/1163/commits/f7634b031660b8376ec7e104aed3097c3328793a let me know if that helps

tests/test_nblist.py
badisa: Declared explicitly in https://github.com/proteneer/timemachine/pull/1163/commits/acdfdcc3732d530a5481228ddb10b9c482d5de44

timemachine/cpp/src/neighborlist.cu
maxentile: Nice -- thanks!

timemachine/cpp/src/neighborlist.cu
maxentile: That's a good point, the max number of atom-atom interactions in an IxnGroup potential, N_rows x N_cols, is loosely upper bounded by N^2, but can often be << N^2.

PR #1168, opened 2023-10-02
Bugfix/nblist rebuild logic ixn group

badisa:


maxentile (approved):
Good catch.

(Effect magnitude may have been partially limited by [this box-change check](https://github.com/proteneer/timemachine/blob/8a32004c9d0259afe596f8ef3bed8b7ba439605f/timemachine/cpp/src/kernels/k_nonbonded.cuh#L9-L28) in most simulations. (Despite bug, nblist would probably have been rebuilt every 50 steps or so, whenever a barostat move was accepted, regardless of how far any particles moved).)

tests/nonbonded/test_nonbonded_interaction_group.py
```diff
@@ -112,6 +112,57 @@ def test_nonbonded_interaction_group_correctness(
...
+    potential = NonbondedInteractionGroup(num_atoms, ligand_idxs, beta, cutoff, col_atom_idxs=col_atom_idxs)
+
+    test_impl = potential.to_gpu(precision)
+
+    # Test that if we compare the potentials then randomize the column indices that the potentials still agree.
+    for params in gen_nonbonded_params_with_4d_offsets(rng, params, cutoff):
+        GradientTest().compare_forces(conf, params, example_box, potential, test_impl, rtol=rtol, atol=atol)
+        GradientTest().assert_differentiable_interface_consistency(conf, params, example_box, test_impl)
+
+        # Randomize the column indices to trigger a nblist rebuild
```
mcwitt, line 159: nit: randomize the coordinates of the column atoms?

badisa: Fixed in [87824c0](https://github.com/proteneer/timemachine/pull/1168/commits/87824c04a3f42a0ea88cc48415856422e504f6af)

mcwitt:
Running this locally I noticed that the tests pass for `num_col_atoms=1`. Is this expected?

mcwitt (approved):

badisa:
> Running this locally I noticed that the tests pass for `num_col_atoms=1` before the fix. Is this expected? (If not, could the test be strengthened somehow?)

It is reasonable to expect that. If the column atom doesn't interact with any of the row atoms, then moving it won't change the energies/forces/etc. Could remove the randomness of this test and have a fixed row/column atoms that are known to interact rather than randomly selecting the particles as is done in this test.

tests/nonbonded/test_nonbonded_interaction_group.py
```diff
@@ -295,7 +295,7 @@ def ref_ixngroups(coords, params, box):
 
 @pytest.mark.parametrize("beta", [2.0])
 @pytest.mark.parametrize("cutoff", [1.1])
-@pytest.mark.parametrize("precision,rtol,atol", [(np.float64, 1e-8, 1e-8), (np.float32, 1e-4, 5e-4)])
```
badisa, line 247: if you run this test on its own it passes, but because the tests share the `rng` different random values change this. For getting this hotfix in, just changing.

PR #1170, opened 2023-10-04
Add example of sampling biphenyl using HREX

mcwitt:
Adds an example that exercises HREX by sampling biphenyl with a torsional barrier, in vacuum and water.

### DIFF
--- a/examples/biphenyl_torsion_sampling_hrex.py
+++ b/examples/biphenyl_torsion_sampling_hrex.py
@@ -274,8 +274,10 @@ def main():
     run_parser.add_argument(
         "--n_frames", type=int, default=4_000, help="Number of frames (picoseconds) to sample using HREX"
     )
+    run_parser.add_argument("-o", "--output", type=str, default="hrex_biphenyl_data.npz", help="Output file name")
 
     plot_parser = subparsers.add_parser("plot")
+    plot_parser.add_argument("-f", "--input", type=str, default="hrex_biphenyl_data.npz", help="Input file name")
     plot_parser.add_argument("--save", action="store_true")
 
     args = parser.parse_args()
@@ -285,7 +287,7 @@ def main():
             args.solvent or False, args.min_overlap, args.n_frames_bisection, args.n_frames
         )
         np.savez(
-            "hrex_biphenyl_data.npz",
+            args.output,
             lambdas=lambdas,
             phi_traj_by_state=phi_traj_by_state,
             phi_traj_by_state_hrex=phi_traj_by_state_hrex,
@@ -293,7 +295,7 @@ def main():
             fraction_accepted_by_pair_by_iter=diagnostics.fraction_accepted_by_pair_by_iter,
         )
     else:
-        data = np.load("hrex_biphenyl_data.npz")
+        data = np.load(args.input)
         lambdas = data["lambdas"]
         phi_traj_by_state = data["phi_traj_by_state"]
         phi_traj_by_state_hrex = data["phi_traj_by_state_hrex"]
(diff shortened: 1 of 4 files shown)


maxentile (approved):

examples/biphenyl_torsion_sampling_hrex.py
```diff
@@ -0,0 +1,331 @@
...
+    assert isinstance(ll_pot, NonbondedPairListPrecomputed)
+
+    ww_params, lw_params, ll_params = nb_pot.unflatten_params(nb_params)
+    lw_params = cast(jax.Array, lw_params)
+    ll_params = cast(jax.Array, ll_params)
+
+    lw_params = lw_params.at[decouple_atom_idxs, 3].set(lamb)
+
+    decouple_atom = np.isin(ll_pot.idxs, decouple_atom_idxs)
+    decouple_pair = np.logical_xor(decouple_atom[:, 0], decouple_atom[:, 1])
```
maxentile, line 111: Intent clarification: I think this encodes that _all pairs_ of atoms in the set `decouple_atom_idxs` are decoupled from each other at the same rate that the set `decouple_atom_idxs` are decoupled (as a unit) from the rest of the environment. Is that correct?

(If there's uncertainty here, might be useful to loosen the signature to something more explicit and flexible like `intramol_atom_pairs_to_decouple: List[Pair] = ..., atoms_to_decouple_from_env: List[Int] = ...` or similar, and possibly leave a TODO comment to revisit more flexible scheduling in the future. (I guess, this family of thermodynamic states has `n * (n-1) / 2 + n` free parameters, if we control each intramol pair's w coord independently, and each atom's env-interaction w coord independently. It is not a requirement that HREX needs to look at a 1D sequence of such states.))

(If there's further uncertainty about what other "softening" may be useful to apply to this example (e.g. optionally reducing a PeriodicTorsion force constant) leaving a note in docstring or similar will be a helpful prompt when revisiting.)

examples/biphenyl_torsion_sampling_hrex.py
```diff
@@ -0,0 +1,331 @@
...
+    ll_params.at[decouple_pair, 3].set(lamb)
```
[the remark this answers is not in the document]
mcwitt: Oof, good catch - I think this was introduced in a final round of refactoring. Will fix.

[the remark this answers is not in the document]
mcwitt: Fixed in https://github.com/proteneer/timemachine/pull/1170/commits/b53c7a2cf27674485ad65cc6de3f8335e8882a1d.

examples/biphenyl_torsion_sampling_hrex.py
mcwitt: > I think this encodes that all pairs of atoms in the set decouple_atom_idxs are decoupled from each other at the same rate that the set decouple_atom_idxs are decoupled (as a unit) from the rest of the environment. Is that correct?

Correct.

> If there's uncertainty here, might be useful to loosen the signature to something more explicit and flexible like intramol_atom_pairs_to_decouple: List[Pair] = ..., atoms_to_decouple_from_env: List[Int] = ... or similar, and

This seems cleaner and easier to understand. Will investigate.

> possibly leave a TODO comment to revisit more flexible scheduling in the future.

Will do.

mcwitt: Refined the transformation protocol to decouple only specified pairs (`[16, 17]` and `[12, 21]`) in https://github.com/proteneer/timemachine/pull/1170/commits/c86c5344d2961c12c5b20b7d8dc8f5790a1eab19

Added note and todo re revisiting more flexible decoupling schedule in https://github.com/proteneer/timemachine/pull/1170/commits/1a438cce69c583b31e47364159ae70d14463539d

proteneer:
LGTM this looks great!  I'm going to squash merge this for you.

proteneer (approved):

PR #1166, opened 2023-09-29
Switch C++/Cuda Params to float

badisa:
* Change parameters to be stored as floats in C++/Cuda, as when we run this in production we don't have float64 support in Jax which is what constructs our potentials
* Adds EnergyType for __int128 to make it easier to try int96 version at some point

# Benchmarks
Cuda Arch 8.6
Nvidia A10

About 2% faster

## Parameters as Doubles
```
dhfr-apo: N=23558 speed: 708.99ns/day dt: 2.5fs (ran 100000 steps in 30.47s)
dhfr-apo-barostat-interval-25: N=23558 speed: 637.04ns/day dt: 2.5fs (ran 100000 steps in 33.91s)
hif2a-apo: N=8805 speed: 1235.90ns/day dt: 2.5fs (ran 100000 steps in 17.48s)
hif2a-apo-barostat-interval-25: N=8805 speed: 1050.07ns/day dt: 2.5fs (ran 100000 steps in 20.57s)
hif2a-rbfe-barostat-interval-15: N=8840 speed: 784.25ns/day dt: 2.5fs (ran 100000 steps in 27.55s)
hif2a-rbfe-local: N=8840 speed: 1376.63ns/day dt: 2.5fs (ran 100000 steps in 15.69s)
solvent-apo: N=6282 speed: 1510.94ns/day dt: 2.5fs (ran 100000 steps in 14.30s)
solvent-apo-barostat-interval-25: N=6282 speed: 1347.81ns/day dt: 2.5fs (ran 100000 steps in 16.03s)
solvent-rbfe-barostat-interval-15: N=6317 speed: 1027.39ns/day dt: 2.5fs (ran 100000 steps in 21.03s)
solvent-rbfe-local: N=6317 speed: 1470.29ns/day dt: 2.5fs (ran 100000 steps in 14.70s)
```

## Parameters as Floats
```
dhfr-apo: N=23558 speed: 720.52ns/day dt: 2.5fs (ran 100000 steps in 29.98s)
dhfr-apo-barostat-interval-25: N=23558 speed: 647.59ns/day dt: 2.5fs (ran 100000 steps in 33.36s)
hif2a-apo: N=8805 speed: 1259.99ns/day dt: 2.5fs (ran 100000 steps in 17.15s)
hif2a-apo-barostat-interval-25: N=8805 speed: 1064.52ns/day dt: 2.5fs (ran 100000 steps in 20.29s)
[...]

PR #1172, opened 2023-10-05
Add C++ Typedefs for common types

badisa:
Adds types for coords, boxes, energies (@maxentile had suggested this for testing different Energy accumulators in the future) and parameters. There are several more types we might want to add (velocities, fixed du_dp, fixed du_dx, etc) but this PR already touches a good amount of code and didn't want to have it get out of hand. 

This was spun out of https://github.com/proteneer/timemachine/pull/1166 where I investigated switching our parameters to floats. 33ba60ed76f2667e6971bb171c74604a313f9256 adds a comment about the bit of code that spawned #1166

### DIFF
--- a/timemachine/cpp/src/barostat.hpp
+++ b/timemachine/cpp/src/barostat.hpp
@@ -26,10 +26,10 @@ public:
     ~MonteCarloBarostat();
 
     // inplace_move() may modify d_x and d_box
-    void inplace_move(double *d_x, double *d_box, cudaStream_t stream);
+    void inplace_move(CoordsType *d_x, CoordsType *d_box, cudaStream_t stream);
 
     // used for testing, bool return tells you if move was accepted
-    bool inplace_move_host(double *h_x, double *h_box);
+    bool inplace_move_host(double *h_x, CoordsType *h_box);
 
     void set_interval(const int interval);
 
@@ -82,8 +82,8 @@ private:
     RealType *d_length_scale_;
     double *d_volume_scale_;
 
-    double *d_x_after_;
-    double *d_box_after_;
+    CoordsType *d_x_after_;
+    CoordsType *d_box_after_;
 
     int *d_atom_idxs_;   // grouped index to atom coords
     int *d_mol_idxs_;    // grouped index to molecule index
--- a/timemachine/cpp/src/barostat.hpp
+++ b/timemachine/cpp/src/barostat.hpp
@@ -71,11 +71,11 @@ private:
     int *d_num_attempted_;
     int *d_num_accepted_;
 
-    __int128 *d_u_buffer_;
-    __int128 *d_u_after_buffer_;
+    EnergyType *d_u_buffer_;
+    EnergyType *d_u_after_buffer_;
 
-    __int128 *d_init_u_;
-    __int128 *d_final_u_;
+    EnergyType *d_init_u_;
+    EnergyType *d_final_u_;
 
     RealType *d_volume_;
     RealType *d_volume_delta_;
--- a/timemachine/cpp/src/hilbert_sort.hpp
+++ b/timemachine/cpp/src/hilbert_sort.hpp
@@ -2,6 +2,7 @@
 #include "cuda_runtime.h"
 #include "device_buffer.hpp"
 #include "math_utils.cuh"
+#include "types.hpp"
 #include <memory>
 #include <numeric>
 #include <vector>
@@ -30,12 +31,12 @@ public:
     void sort_device(
         const int N,
         const unsigned int *d_atom_idxs,
-        const double *d_coords,
-        const double *d_box,
+        const CoordsType *d_coords,
+        const CoordsType *d_box,
         unsigned int *d_output_perm,
         cudaStream_t stream);
 
-    std::vector<unsigned int> sort_host(const int N, const double *h_coords, const double *h_box);
+    std::vector<unsigned int> sort_host(const int N, const CoordsType *h_coords, const CoordsType *h_box);
 };
 
 } // namespace timemachine
--- a/timemachine/cpp/src/kernels/k_fixed_point.cuh
+++ b/timemachine/cpp/src/kernels/k_fixed_point.cuh
@@ -2,6 +2,7 @@
 
 // cuda specific version
 #include "../fixed_point.hpp"
+#include "../types.hpp"
 
 // (ytz): courtesy of @scottlegrand/NVIDIA, even faster conversion
 // This was original a hack to improve perf on Maxwell, that is now needed for Ampere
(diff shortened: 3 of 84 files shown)


timemachine/cpp/src/barostat.hpp
```diff
@@ -26,10 +26,10 @@ template <typename RealType> class MonteCarloBarostat {
     ~MonteCarloBarostat();
 
     // inplace_move() may modify d_x and d_box
-    void inplace_move(double *d_x, double *d_box, cudaStream_t stream);
+    void inplace_move(CoordsType *d_x, CoordsType *d_box, cudaStream_t stream);
 
     // used for testing, bool return tells you if move was accepted
-    bool inplace_move_host(double *h_x, double *h_box);
+    bool inplace_move_host(double *h_x, CoordsType *h_box);
```
proteneer, line 32: CoordsType h_x?

timemachine/cpp/src/types.hpp
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
```
proteneer, line 2: this should probably be namespaced to timemachine unless you truly want global visibility

proteneer:
Some questions to consider:

-Is the benefit for the abstraction worth the extra level of indirection?
-Do we want to do this at level of a typedef or a template argument? (I'm guessing template argument would be much harder to refactor)
-Are there situations in which Coords, Velocities, Boxes may use inconsistent different type? If not, maybe just name them `XVBType` or something that's more generic, seems weird to type boxes/velocities with `CoordsType`.
-Fixed precision types, I know you're deferring this for a later pass, but while `du_dx`, `du_dp` etc. are all fixed point, they may use different `FIXED_EXPONENT` values, (esp in `du_dp` where different elements may have a different `FIXED_EXPONENT`). Is it obvious that they're actually different?

proteneer:
Also looks like this isn't propagate into kernel code `which switches over to RealType` - it's not super obvious where the casts are happening (since we have to look up two types now to check for compatibility).

badisa:
> -Is the benefit for the abstraction worth the extra level of indirection?

I think so? One it makes the types more explicit, IE that the when you use coords it is all of the same type rather than the implicit assumption that they are always double. Also related to your other comment (https://github.com/proteneer/timemachine/pull/1172#issuecomment-1751017756) these sorts of templated coords/boxes stand out more obvious as incorrect (https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/kernels/k_integrator.cuh#L93 which was changed to CoordsType in this PR). 

This also allows testing things such as what the impact of storing things like the parameters, coords, etc in different precision. Not to say that we would actually switch, but I found it informative that switching from storing parameters as doubles to floats gets us a 2% speed up.  Even if we don't intend to actually make such a switch.

>-Are there situations in which Coords, Velocities, Boxes may use inconsistent different type? If not, maybe just name them `XVBType`

Having a `XVBType` would make sense to me. I could imagine velocities not needing to be doubles which is why I separated it, but at the very least naming the type to `XBType` would be good. 

> -Fixed precision types, I know you're deferring this for a later pass, but while `du_dx`, `du_dp` etc. are all fixed point, they may use different `FIXED_EXPONENT` values, (esp in `du_dp` where different elements may have a different `FIXED_EXPONENT`). Is it obvious that they're actually different?

In this case I would declare separate type defs even if the types are the same. Probably not something we will ever want to change, so maybe this is just busy work.

timemachine/cpp/src/types.hpp
```diff
@@ -0,0 +1,7 @@
...
+// CoordsType defines both the coordinates and the boxes
+typedef double CoordsType;
+typedef double ParamsType;
+// TBD: Investigate a 96bit version for energies
+typedef __int128 EnergyType;
```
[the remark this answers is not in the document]
badisa: Attempted to answer in the first part of https://github.com/proteneer/timemachine/pull/1172#issuecomment-1751042468, let me know if that makes sense

timemachine/cpp/src/hilbert_sort.hpp
```diff
@@ -30,12 +31,12 @@ class HilbertSort {
     void sort_device(
         const int N,
         const unsigned int *d_atom_idxs,
-        const double *d_coords,
-        const double *d_box,
+        const CoordsType *d_coords,
```
mcwitt, line 34: minor/style: `CoordsType` or just `Coords`?

timemachine/cpp/src/kernels/k_fixed_point.cuh
```diff
@@ -70,27 +71,27 @@ template <typename RealType> unsigned long long __device__ __forceinline__ FLOAT
 
 /* FLOAT_TO_FIXED_ENERGY converts floating point energies into fixed point. Values beyond LLONG_MAX/LLONG_MIN
 * (or non-finite) will be capped to LLONG_MAX. This is accumulated into __int128 which handles the positive/negative overflows,
-* allows to account for overflows triggered by the summation of energies.
+* allows to account for overflows triggered by the summation of energies. EnergyType is a typedef for __int128.
```
mcwitt, line 74: nit: comment at risk of becoming stale?

timemachine/cpp/src/types.hpp
badisa: Playing around with this, looks like we have a ton of things that are globally visible. Might open a separate PR to put all our code into the correct namespace to avoid wrapping a bunch of code in a namespace here.

timemachine/cpp/src/kernels/k_fixed_point.cuh
badisa: I initially referenced `EnergyType` in this comment, but seemed specific to the choice of `__int128` so I simply referenced that `EnergyType` is `__int128` so that the reader could connect it with what the code states. I am alright with just referencing `__int128`, just seemed confusing to read `(EnergyType)accumulated_energy > LLONG_MAX  - Correctly detects that energies are beyond valid range`

mcwitt:
> This also allows testing things such as what the impact of storing things like the parameters, coords, etc in different precision. Not to say that we would actually switch, but I found it informative that switching from storing parameters as doubles to floats gets us a 2% speed up. Even if we don't intend to actually make such a switch.

This motivation makes sense to me. I'm less clear on the utility of using different types for coordinates and parameters, versus a unified alias like `RealType` in the kernels. Did you have a motivation in mind for allowing xvb and parameter types to differ?

badisa:
> > This also allows testing things such as what the impact of storing things like the parameters, coords, etc in different precision. Not to say that we would actually switch, but I found it informative that switching from storing parameters as doubles to floats gets us a 2% speed up. Even if we don't intend to actually make such a switch.
> 
> This motivation makes sense to me. I'm less clear on the utility of using different types for coordinates and parameters, versus a unified alias like `RealType` in the kernels. Did you have a motivation in mind for allowing xvb and parameter types to differ?

The initial reason was https://github.com/proteneer/timemachine/pull/1166 which came out of the fact that we truncate to `float32` precision in production where jax does not have double precision enabled.

timemachine/cpp/src/types.hpp
badisa: Opened https://github.com/proteneer/timemachine/pull/1173 which moves our code to being consistently in the timemachine namespace.

Perhaps it is undesirable for our code to all be in a single namespace, though that seems to be the expectation.

timemachine/cpp/src/types.hpp
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+// CoordsType defines both the coordinates and the boxes
+typedef double CoordsType;
+typedef double ParamsType;
```
badisa, line 5: One potential thing we could do here if we only really care about using double for testing purposes (@proteneer perhaps I misunderstood the purpose beyond developer utility for this being double) we could have a compile time flag to switch this to float and get the 2% up.

timemachine/cpp/src/barostat.hpp
badisa: Good catch, fixed in https://github.com/proteneer/timemachine/pull/1172/commits/3eb4dafe59255ad9cceeddd4d88dc231ad611c85

badisa:
Closing as there is more points to discuss offline

PR #1174, opened 2023-10-11
Rotate Coordinates in C++

badisa:
- Initial implementation of rotating coordinates in C++ using quarternions
- Generating random noise as quaternions produces uniform transformations


tests/test_rotations.py
```diff
@@ -0,0 +1,110 @@
...
+
+def hamiliton_product(q1: NDArray, q2: NDArray) -> NDArray:
+    """
+    Compute the product of two quaternions, called a hamilton product.
+
+    References
+    ----------
+    https://en.wikipedia.org/wiki/Quaternion#Hamilton_product
+    """
+    w1, x1, y1, z1 = q1[0], q1[1], q1[2], q1[3]
```
badisa, line 17: I chose the ordering of `w, x, y, z` based on Wikipedia's order rather than `x, y, z, w` which is what Scipy uses. Not sure if we have a preference.

tests/test_rotations.py
```diff
@@ -0,0 +1,110 @@
...
+    w2, x2, y2, z2 = q2[0], q2[1], q2[2], q2[3]
+    output = np.empty(4)
+    output[0] = (w1 * w2) - (x1 * x2) - (y1 * y2) - (z1 * z2)
+    output[1] = (w1 * x2) + (x1 * w2) + (y1 * z2) - (z1 * y2)
+    output[2] = (w1 * y2) - (x1 * z2) + (y1 * w2) + (z1 * x2)
+    output[3] = (w1 * z2) + (x1 * y2) - (y1 * x2) + (z1 * w2)
+    return output
+
+
+def rotate_vector_by_quaternions(coords: NDArray, quaternion: NDArray) -> NDArray:
```
badisa, line 27: Not sure if we want this reference implmentation outside the test file, since you can use scipy.Rotation instead, and this isn't intended to be performant.

timemachine/cpp/src/rotations.cu
```diff
@@ -0,0 +1,37 @@
...
+#include "kernels/k_rotations.cuh"
+#include "kernels/kernel_utils.cuh"
+#include "math_utils.cuh"
+#include "rotations.hpp"
+
+namespace timemachine {
+
+// rotate_coordinates_host is used to test the rotation kernel
+template <typename RealType>
+void rotate_coordinates_host(
```
badisa, line 12: I didn't want to put kernel code into the `wrap_kernals.cpp` file, so I created this function, not sure if there is a preference of where to put this sort of code.

tests/test_rotations.py
mcwitt, line 17: nit: could this just be
```suggestion
    w1, x1, y1, z1 = q1
```
?

tests/test_rotations.py
```diff
@@ -0,0 +1,110 @@
...
+
+    References
+    ----------
+    https://math.stackexchange.com/a/535223
+    """
+    assert coords.shape == (3,)
+    assert quaternion.shape[0] == 4
+    expanded_coords = np.concatenate([np.zeros(1), coords])
+    quaternion_conjugate = quaternion.copy()
+    # Compute quaternion conjugate by taking the conjugate of all but w
```
mcwitt, line 39: nit:
```suggestion
    # Compute quaternion conjugate by negating all but w
```

tests/test_rotations.py
mcwitt, line 27: nit: should the name of this function by `rotate_vector_by_quaternion` (singular)?

tests/test_rotations.py
```diff
@@ -0,0 +1,110 @@
...
+@pytest.mark.nogpu
+@pytest.mark.parametrize("seed", list(range(10)))
+@pytest.mark.parametrize("size", [1000])
+def test_reference_rotation_by_quaternion(seed, size):
+    rng = np.random.default_rng(seed)
+
+    all_coords = rng.normal(size=(size, 3)) * 100.0
+
+    quaternions = rng.normal(size=(size, 4))
+    # Normalize the quaternions to a vector quaternion, implicitly performed by Rotation
```
mcwitt, line 64: nit: should this say
```suggestion
    # Normalize the quaternions to unit quaternions, representing a rotations
```
?

tests/test_rotations.py
```diff
@@ -0,0 +1,110 @@
...
+    quaternions = quaternions / np.expand_dims(norms, -1)
+    for coords, quaternion in zip(all_coords, quaternions):
+
+        rotation = Rotation.from_quat(convert_quaternion_for_scipy(quaternion))
+
+        ref_coords = rotation.apply(coords)
+
+        test_coords = rotate_vector_by_quaternions(coords, quaternion)
+
+        np.testing.assert_allclose(ref_coords, test_coords)
```
mcwitt, line 75: curiosity: are these actually just equal?

tests/test_rotations.py
```diff
@@ -0,0 +1,110 @@
...
+
+    quaternions = rng.normal(size=(n_rotations, 4))
+    rotate_function = custom_ops.rotate_coords_f32
+    if precision == np.float64:
+        rotate_function = custom_ops.rotate_coords_f64
+
+    rotated_coords = rotate_function(all_coords, quaternions)
+
+    norms = np.linalg.norm(quaternions, axis=-1)
+    # Normalize the quaternions to a vector quaternion after passing to C++, as should be automatically normalized
```
mcwitt, line 96: (think this should say "unit" here too?)
```suggestion
    # Normalize the quaternions to a unit quaternion after passing to C++, as should be automatically normalized
```

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -1342,6 +1347,32 @@ double py_accumulate_energy(const py::array_t<long long, py::array::c_style> &in
...
+        throw std::runtime_error("quaternions dimensions must be 2");
+    }
+    if (quaternions.shape(quaternions_ndims - 1) != 4) {
+        throw std::runtime_error("quaternions must have a shape that is 4 dimensional");
+    }
+
+    std::vector<RealType> v_quaternions(quaternions.size());
+    for (int i = 0; i < quaternions.size(); i++) {
+        v_quaternions[i] = static_cast<RealType>(quaternions.data()[i]);
+    }
```
mcwitt, line 1366: Could this use `py_array_to_vector`?

tests/test_rotations.py
mcwitt: Any reason we couldn't just use the scipy implementation as the reference and remove this entirely?

mcwitt (approved):
Looks good, just minor nits

tests/test_rotations.py
badisa: Totally could, had done this primarily to have a 1-1 implementation of the quaternion rotation rather than going through the matrix multiplication that Scipy does (https://github.com/scipy/scipy/blob/main/scipy/spatial/transform/_rotation.pyx#L1943).

timemachine/cpp/src/wrap_kernels.cpp
badisa: No, because this takes doubles and converts them to floats if `RealType=float`. Could separate this out into a different function for converting from an input type to an output type though.

tests/test_rotations.py
badisa: Yup, will change

tests/test_rotations.py
badisa: They are not, they differ by about atol/rtol of 1e-13.

tests/test_rotations.py
badisa: Changed in https://github.com/proteneer/timemachine/pull/1174/commits/ff88f8762f5e809c6df5e09b36a5ac85b268ad04

tests/test_rotations.py
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1174/commits/88e9f274e2b2a18a87869d913a4332d9f011fa53

badisa: Removed the ref implementation here:  https://github.com/proteneer/timemachine/pull/1174/commits/251899e6e8f72a1e4031d4291dad99f219526d3a

timemachine/cpp/src/wrap_kernels.cpp
badisa: Renamed to better match `py_array_to_vector` https://github.com/proteneer/timemachine/pull/1174/commits/be05c972548bbaeab7ea50bcd51ec6d01fa8c857

mcwitt: +1 for `py_array_to_vector_with_cast`. Could also see generalizing the existing `py_array_to_vector` to add a second type parameter (but would require updating existing usage and in most places the input and output types would be the same).

tests/test_rotations.py
```diff
@@ -0,0 +1,48 @@
...
+    for i, coords in enumerate(all_coords):
+        ref_rotated_coords = rotation.apply(coords)
+        for j, ref_coords in enumerate(ref_rotated_coords):
+            np.testing.assert_allclose(
+                rotated_coords[i][j],
+                ref_coords,
+                rtol=rtol,
+                atol=atol,
+                err_msg=f"Coords {i} with Rotation {j} have a mismatch",
+            )
```
proteneer, line 48: one more useful sanity test: applying the inverse of the quaternion `(x,y,z,w) -> (-x,-y,-z,w)` to `ref_rotated_coords` should get back `coords`

proteneer (approved):
lgtm!

tests/test_rotations.py
badisa: Added in https://github.com/proteneer/timemachine/pull/1174/commits/9a4bb93df3a9eae09abcceced4b36d54120dd629

PR #1175, opened 2023-10-12
Fixes and improvements for HREX replica-state convergence plot

mcwitt:
* fix missing row when `n_states` is a multiple of `ncols`
* fix incorrect axis label placement
* fix divide by zero
* change color scale to log; color pairs with zero observations white

Before:

After:

### DIFF
--- a/timemachine/fe/plots.py
+++ b/timemachine/fe/plots.py
@@ -431,20 +431,24 @@ def plot_hrex_replica_state_distribution_heatmap(
 def plot_hrex_replica_state_distribution_convergence(cumulative_replica_state_counts: NDArray, ncols: int = 4):
     """Plot distribution of states as a function of iteration for each replica."""
     n_iters, n_states, _ = cumulative_replica_state_counts.shape
-    fraction_by_replica_by_state_by_iter = (
-        cumulative_replica_state_counts / np.arange(n_iters)[:, None, None]
-    )  # (iter, state, replica) -> float
-    fraction_by_state_by_iter_by_replica = np.moveaxis(
-        fraction_by_replica_by_state_by_iter, 2, 0
-    )  # (replica, iter, state) -> float
+    iteration = np.arange(n_iters) + 1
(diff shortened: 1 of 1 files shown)


timemachine/fe/plots.py
```diff
@@ -431,20 +431,24 @@ def plot_hrex_replica_state_distribution_heatmap(
...
-    axs.shape = (ncols, nrows)
+    fig, axs = plt.subplots(ncols=ncols, nrows=nrows, figsize=(13, 10), sharex=True, sharey=True, squeeze=False)
+
+    # (replica, state) pairs with no observations will be colored white
+    cmap = plt.get_cmap("viridis")
+    cmap.set_bad("white")
 
     for replica_idx, (fraction_by_state_by_iter, ax) in enumerate(zip(fraction_by_state_by_iter_by_replica, axs.flat)):
-        p = ax.pcolormesh(np.arange(n_iters), np.arange(n_states), fraction_by_state_by_iter.T, vmin=0.0, vmax=1.0)
+        p = ax.pcolormesh(iteration, np.arange(n_states), np.log10(fraction_by_state_by_iter.T))
```
mcwitt, line 451: The log color scale seems to give slightly better visual contrast, but I think the most useful change is to use a different color for (replica, state) pairs with no observations. Could remove the log scale and just mask 0.0 values if that seems more natural.

maxentile: log seems good -- might just reverse color map for nonzero values, so that it starts from white and becomes darker?

(currently: 0->white, very small value->dark, large value->light)

maxentile (approved):

timemachine/fe/plots.py
mcwitt: Good call, reversed in https://github.com/proteneer/timemachine/pull/1175/commits/417a8a51703699376ddf4cc4effcb60487021bb3 (and updated example image)

PR #1176, opened 2023-10-12
Cuda Implementation of LogSumExp

badisa:
* Its a bit weird as I avoided computing values that probably don't make sense launching kernels for.


timemachine/cpp/src/logsumexp.cu
```diff
@@ -0,0 +1,74 @@
...
+    DeviceBuffer<RealType> d_vals(N);
+    d_vals.copy_from(h_values);
+
+    DeviceBuffer<RealType> d_out(2);
+
+    cudaStream_t stream = static_cast<cudaStream_t>(0);
+    this->sum_device(N, d_vals.data, d_out.data, stream);
+    gpuErrchk(cudaStreamSynchronize(stream));
+
+    RealType h_intermediate[2];
```
badisa, line 64: I am going back and forth on whether to run a kernel that performs the summation of the log(x) and the max value or the leave it out. Once we start using this in anger, I predict such a kernel would need to die for performance reasons (launch of a kernel time vs computation time is not great) but on the other hand you have this weird logic in the host code. There is a comment about this here (https://github.com/proteneer/timemachine/blob/f207bad143b6abd3d2264c8aca9c937ac2d5dcaa/timemachine/cpp/src/logsumexp.cu#L50)

Will think about some more.

tests/test_logsumexp.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+@pytest.mark.memcheck
+@pytest.mark.parametrize("precision,atol,rtol", [(np.float64, 1e-8, 1e-8), (np.float32, 2e-5, 1e-5)])
+@pytest.mark.parametrize("seed", list(range(2)))
+@pytest.mark.parametrize("size", [2, 100, 10000])
+def test_cuda_logsumexp(precision, atol, rtol, seed, size):
+    rng = np.random.default_rng(seed)
+
+    values = rng.normal(size=size) * 1000.0
```
proteneer, line 15: test should be modified to include both a set of very large values and a set of very negative values, i.e.:

`values = rng.normal(size=size, loc=-10000, scale=1.0)`
and
`values = rng.normal(size=size, loc=10000, scale=1.0)`

badisa: Added in https://github.com/proteneer/timemachine/pull/1176/commits/cd982aae818703cf3cea0ac33c0e4e89d6bd6e1a

timemachine/cpp/src/logsumexp.cu
badisa: Inclined to leave this out as it is a trivial kernel and likely to be done in the kernel that subtracts the logsumexp from the energies to get the log weights.

proteneer: Whatever is simplest!

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -1374,6 +1375,27 @@ void declare_fanout_summed_potential(py::module &m) {
...
 
+template <typename RealType> void declare_log_sum_exp(py::module &m, const char *typestr) {
+
+    using Class = LogSumExp<RealType>;
+    std::string pyclass_name = std::string("LogSumExp_") + typestr;
+    py::class_<Class, std::shared_ptr<Class>>(m, pyclass_name.c_str(), py::buffer_protocol(), py::dynamic_attr())
+        .def(py::init([](const int N) { return new Class(N); }), py::arg("N"))
+        .def(
+            "sum",
+            [](Class &summer,
```
proteneer, line 1386: lol "summer"

proteneer (approved):
lgtm!

PR #1177, opened 2023-10-13
Adaptive SMC

jkausrelay:
- Add an adaptive version of SMC which uses the conditional ESS to determine when to resample


tests/test_smc.py
```diff
@@ -115,8 +119,60 @@ def test_effective_sample_size():
...
+        assert pytest.approx(ess) == cess
+
+        # ESS and CESS differ when not actually resampling
+        # or conditional_multinomial_resample is used with a ess < thresh
+        indices, log_weights = identity_resample(log_weights + incremental_log_weights)
+        norm_log_weights = log_weights - logsumexp(log_weights)
+
+        ess = effective_sample_size(log_weights + incremental_log_weights)
+        cess = conditional_effective_sample_size(norm_log_weights, incremental_log_weights)
+        assert np.abs(cess - ess) > 1
```
jkausrelay, line 170: Is the sign of cess - ess consistent?

timemachine/md/smc.py
```diff
@@ -120,6 +122,158 @@ def accumulate_results(samples, indices, log_weights, incremental_log_weights):
...
+            mid_lambda += epsilon
+
+        lam_target = mid_lambda
+
+        # Stop when lam_target == 1.0
+        #   See
+        #   * discussion at https://github.com/proteneer/timemachine/pull/718#discussion_r854276326
+        #   * helper function get_endstate_samples_from_smc_result
+        if lam_target == 1.0:
+            break
```
maxentile, line 243: could we simplify using scipy.optimize.root_scalar ? (as in https://github.com/proteneer/timemachine/pull/442/files#diff-cb96fac1a84532e2c8d063fdcb9f0695dff44d884211290d23817b0d7d57651cR94 )

timemachine/md/smc.py
```diff
@@ -120,6 +122,158 @@ def accumulate_results(samples, indices, log_weights, incremental_log_weights):
...
+        ancestry_traj.append(indices)
+        log_weights_traj.append(np.array(log_weights))
+        incremental_log_weights_traj.append(np.array(incremental_log_weights))
+        lambdas_traj.append(lam_target)
+
+    lam_initial = 0.0
+    lam_target = 1.0  # adapted
+
+    # Main ASMC loop
+    while True:
```
maxentile, line 211: nit: possibly replace with a for loop up to some max iteration count?

timemachine/md/smc.py
```diff
@@ -120,6 +122,158 @@ def accumulate_results(samples, indices, log_weights, incremental_log_weights):
     return trajs_dict
 
 
+def adaptive_sequential_monte_carlo(
```
maxentile, line 125: nit: opportunity to revert duplication in the future: could express non-adaptive SMC as a special case, if this accepts a function `select_next_lam` in signature. (non-adaptive `select_next_lam` gets the next entry in a list, CESS-adaptive `select_next_lam` is implemented as below, ...)

maxentile (approved):

timemachine/md/smc.py
jkausrelay: I'm not sure how that would work. If we hit the max count what do we return? Just the partial results?

timemachine/md/smc.py
jkausrelay: Can be a deboggle case for later

timemachine/md/smc.py
maxentile: hmm, was thinking either:
* after max iteration count limit is reached, jump directly to lam=1
* raise an error

unsure if either is preferable to potentially unbounded loop

another approach to ensure the loop is bounded is to set a minimum step size epsilon, so each binary search looks at the interval `(lam_current + eps, lam_target]` rather than `(lam_current, lam_target)`

timemachine/md/smc.py
```diff
@@ -120,6 +123,153 @@ def accumulate_results(samples, indices, log_weights, incremental_log_weights):
...
+        # resample
+        indices, log_weights = resample(log_weights + incremental_log_weights)
+        norm_log_weights = log_weights - logsumexp(log_weights)
+        resampled = [sample_traj[-1][i] for i in indices]
+
+        # propagate
+        samples = propagate(resampled, lam_target)
+
+        # log
+        accumulate_results(samples, indices, log_weights, incremental_log_weights, lam_target)
```
maxentile, line 250: nit: possible future logging enhancement: forecast # steps remaining

(crudely, could guess this based on the size of the most recent step https://gist.github.com/maxentile/be328e929abf4a92bee7d26967277f54#file-cess_smc_openmm-py-L132-L133 or similar)

timemachine/md/smc.py
jkausrelay: This should be handled by xtol in the solver though. I tested this by setting cess_target=len(samples) and confirmed we don't get stuck in an infinite loop.

timemachine/md/smc.py
jkausrelay: Perhaps for the future

timemachine/md/smc.py
maxentile: >This should be handled by xtol in the solver though. I tested this by setting cess_target=len(samples) and confirmed we don't get stuck in an infinite loop.

Ahh, I missed that interaction.

(So this implicitly sets max iteration count = 1.0 / xtol, which by default is 100.)

Minor point, but controlling the solver precision and max # lambda windows separately seems preferable (binary search is cheap, esp. with cached prefactors, and a user may be tempted to make the solver very precise, e.g. setting xtol to machine epsilon or so. May want to be able to do this independent of max iteration count.

jkausrelay: Generally you can do this by setting a reasonable cess_target and set xtol to be small. It's only `max iteration count = 1.0 / xtol` when cess_target ~= n_samples.

maxentile: >It's only max iteration count = 1.0 / xtol when cess_target ~= n_samples.

Or, potentially, if something else unspecified goes wrong. (E.g. if all the MD simulations blow up, or if something else happens so that infinitesimally small changes in lambda result in very large changes in U(x, lam)...)

jkausrelay: good point.

jkausrelay: Added max iterations and raise if exceeded

maxentile (approved):

PR #1179, opened 2023-10-16
Deboggle C++

badisa:
* Removes some unused code
* Uses the templated D variable where unused
* Correctly checks for Cub Errors

### DIFF
--- a/timemachine/cpp/src/hilbert_sort.cu
+++ b/timemachine/cpp/src/hilbert_sort.cu
@@ -33,15 +33,14 @@ HilbertSort::HilbertSort(const int N)
 
     // estimate size needed to do radix sorting
     // reuse d_sort_keys_in_ rather than constructing a dummy output idxs buffer
-    cub::DeviceRadixSort::SortPairs(
+    gpuErrchk(cub::DeviceRadixSort::SortPairs(
         nullptr,
         d_sort_storage_bytes_,
         d_sort_keys_in_.data,
         d_sort_keys_out_.data,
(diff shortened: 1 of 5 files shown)


timemachine/cpp/src/hilbert_sort.cu
```diff
@@ -33,15 +33,14 @@ HilbertSort::HilbertSort(const int N)
 
     // estimate size needed to do radix sorting
     // reuse d_sort_keys_in_ rather than constructing a dummy output idxs buffer
-    cub::DeviceRadixSort::SortPairs(
+    gpuErrchk(cub::DeviceRadixSort::SortPairs(
```
proteneer, line 36: Interesting - I didn't know this returned a cudaError directly, does this kernel run synchronously?

timemachine/cpp/src/kernels/kernel_utils.cuh
```diff
@@ -13,35 +13,6 @@ template <typename RealType> struct box_cache {
     RealType inv_z;
 };
 
-#define HESS_3N3N(i, j, N, di, dj) (di * N * 3 * N + i * 3 * N + dj * N + j)
```
proteneer, line 16: oh the memories - feels good to delete these.

timemachine/cpp/src/hilbert_sort.cu
badisa: This call runs synchronously, but only to setup the size. If you provide all of the values, including a stream, it is async.

PR #1180, opened 2023-10-16
Add warning for implicit float conversion

badisa:
- Pretty easy to do and a nice to have

### DIFF
--- a/timemachine/cpp/CMakeLists.txt
+++ b/timemachine/cpp/CMakeLists.txt
@@ -19,6 +19,7 @@ string(APPEND CMAKE_CUDA_FLAGS
   " -Wextra"
   " -Wno-unused-parameter"
   " -Wreorder"
+  " -Wfloat-conversion"
 )
 message(${CMAKE_CUDA_FLAGS})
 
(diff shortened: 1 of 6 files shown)


timemachine/cpp/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ string(APPEND CMAKE_CUDA_FLAGS
   " -Wextra"
   " -Wno-unused-parameter"
   " -Wreorder"
+  " -Wfloat-conversion"
```
proteneer, line 22: What about `-Wconversion`?

proteneer: (I'll definitely take `-Wfloat-conversion` as well)

proteneer (approved):

timemachine/cpp/CMakeLists.txt
badisa: It throws a lot of warnings at the moment mostly about going from `size_t -> int`. Should work towards enabling `-Wconversion`.

PR #1182, opened 2023-10-16
Optimize Tests

badisa:
* Move HREX reproducibility on complex to nightly (solvent already exercises the host logic)
* More barostat tests to memory tests
* Move a tests to CPU only where possible
* Remove minimization where not needed

### DIFF
--- a/tests/hrex/test_hrex_rbfe.py
+++ b/tests/hrex/test_hrex_rbfe.py
@@ -24,7 +24,9 @@ from timemachine.testsystems.relative import get_hif2a_ligand_pair_single_topolo
 DEBUG = False
 
 
-@pytest.fixture(scope="module", params=[None, "complex", "solvent"])
+@pytest.fixture(
+    scope="module", params=[None, "solvent", pytest.param("complex", marks=pytest.mark.nightly(reason="slow"))]
+)
 def hif2a_single_topology_leg(request):
     host_name = request.param
     forcefield = Forcefield.load_default()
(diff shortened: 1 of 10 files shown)


badisa:
We could also remove `equilibrate_host` as it appears unused and the test (https://github.com/proteneer/timemachine/blob/master/tests/test_minimizer.py#L89) takes 20 seconds.

maxentile (approved):
Nice optimizations!

>We could also remove equilibrate_host as it appears unused and the test (https://github.com/proteneer/timemachine/blob/master/tests/test_minimizer.py#L89) takes 20 seconds.

Agnostic, leaning towards keeping the function and its test.

I don't immediately see a way to make its test much cheaper without refactoring `equilibrate_host` to expose more of its settings. (`n_steps` already specifies just 25 MD steps, but `minimize_host_4d` and `fire_minimize` are probably heavier, 2500 steps and 50 steps resp.).

badisa:
> Agnostic, leaning towards keeping the function and its test.

Sounds good.

> I don't immediately see a way to make its test much cheaper without refactoring `equilibrate_host` to expose more of its settings. (`n_steps` already specifies just 25 MD steps, but `minimize_host_4d` and `fire_minimize` are probably heavier, 2500 steps and 50 steps resp.).

I do have a change in another PR that makes `minimize_host_4d` a tiny bit faster, as we actually call `set_params` one extra time than we need.

jkausrelay (approved):

PR #1183, opened 2023-10-17
Removes float64 dtype for torsion parameterization

badisa:
* Inconsistent with the rest of the code and throws a warning when jax is not configured for float64


proteneer (approved):
lgtm and thanks for fixing the seed.

PR #1184, opened 2023-10-17
DRY conversion of py::array_t to vector

badisa:


proteneer (approved):
nice! much less error prone too.

PR #1185, opened 2023-10-17
Remove keep_idxs

proteneer:
Since we're now using disk-backed `StoredArrays`, we no longer need to manage `keep_idxs`, and certain simulation methods like HREX will always physically store `O(K_windows*N_frames)` amount of data.


badisa (approved):
Looks great, nice to clean this up!

jkausrelay (approved):

PR #1186, opened 2023-10-17
Replace execute_selective with execute with flags

badisa:
- Adds flags to `execute` and removed the `_selective` methods
  * Avoids having two separate APIs we have to test
  * Allows optimizing a single path (These get used heavily in a lot of places)
- Only compute du_dx for minimization (appears to make 3% difference for minimization tests on A4000)
- Only compute energies for final u_kln from the estimate free energy methods

## TODO
- [ ] Add tests


timemachine/cpp/src/bound_potential.cu
```diff
@@ -36,22 +36,35 @@ void BoundPotential::execute_host(
     d_x.copy_from(h_x);
     d_box.copy_from(h_box);
 
-    DeviceBuffer<unsigned long long> d_du_dx(N * D);
-    DeviceBuffer<__int128> d_u(1);
+    std::unique_ptr<DeviceBuffer<unsigned long long>> d_du_dx(nullptr);
```
proteneer, line 39: Can you explain the rationale for the `unique_ptr` here? Not sure why we need the extra level of indirection here? Performance reasons?

proteneer (approved):
lgtm save minor nit re: `unique_ptr`

timemachine/cpp/src/bound_potential.cu
badisa: The rationale here is that we don't allocate GPU memory if we don't need it. The reduction in memory allocations is where most of the performance improvement comes from.

PR #1190, opened 2023-10-18
Replace cudaMalloc with DeviceBuffer in dynamic code

badisa:
* Resolves https://github.com/proteneer/timemachine/issues/577
* Removes the last of the dynamic code that calls cudaMalloc, which avoids the explicit calls to cudaFree.
* Does *not* change classes to use `DeviceBuffer` as I find the cudaErrchk messages that point you to the `DeviceBuffer` file unhelpful compared to the actual line the in the potential that triggered a failure.


proteneer (approved):
Nice!

PR #1198, opened 2023-10-26
NonbondedMolEnergyPotential

badisa:
- [x] Unify group indices logic with the barostat
- [x] Add a plot for timings (so far seems to be 10x faster, but need to plot)
- [x] Clean up tests~, split from memcheck for the large systems~

# Benchmarks
Cuda Arch 8.6
Nvidia A4000

Reference is `BDExchangeMover.batch_U_fn`, while the single and double precision are `NonbondedMolEnergyPotential.execute`, note that `execute` allocates memory and copies data from the host to the GPU.
benchmarks_14


timemachine/cpp/src/gpu_utils.cuh
```diff
@@ -124,37 +125,39 @@ __device__ __forceinline__ void energy_warp_reduce(volatile __int128 *shared_dat
...
-__device__ __forceinline__ void block_energy_reduce(volatile __int128 *shared_data, int tid) {
+__device__ __forceinline__ void block_energy_reduce(
+    volatile __int128 *shared_data, // [THREADS_PER_BLOCK]
+    int tid) {
     static_assert(THREADS_PER_BLOCK <= 1024 && (THREADS_PER_BLOCK & (THREADS_PER_BLOCK - 1)) == 0);
 
     // For larger blocks larger than a warp
     if (THREADS_PER_BLOCK >= 1024) {
-        if (tid < 512) {
+        if (tid < 512 && tid + 512 < THREADS_PER_BLOCK) {
```
badisa, line 135: Can probably remove some of these, will revisit to make this more strict. Think this only matters where thread block is <32 and a power of two

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -430,4 +430,156 @@ void __global__ k_nonbonded_unified(
...
+        return;
+    }
+
+    __int128 energy_accumulator = 0;
+
+    const int mol_start = mol_offsets[idx];
+    const int mol_end = mol_offsets[idx + 1];
+
+    // TBD Parallelize
+    for (int i = mol_start; i < mol_end; i++) {
```
badisa, line 576: This is likely very bad in the case of a protein

timemachine/md/barostat/utils.py
```diff
@@ -43,7 +43,9 @@ def get_group_indices(bond_list: List[Tuple[int, int]], num_atoms: int) -> List[
     """Connected components of bond graph"""
 
     topology = nx.Graph(bond_list)
-    components = [np.array(list(c)) for c in nx.algorithms.connected_components(topology)]
+    components = [np.array(list(sorted(c))) for c in nx.algorithms.connected_components(topology)]
```
badisa, line 46: Ensures that the arrays are sorted, which makes things easier on the GPU.

timemachine/md/barostat/utils.py
```diff
@@ -43,7 +43,9 @@ def get_group_indices(bond_list: List[Tuple[int, int]], num_atoms: int) -> List[
...
+    for i in range(len(components)):
+        assert np.all(np.diff(components[i]) == 1)
```
badisa, line 48: Verify that all components are consecutive indices

timemachine/cpp/src/nonbonded_mol_energy.cu
```diff
@@ -0,0 +1,173 @@
...
+        throw std::runtime_error("N != N_");
+    }
+
+    if (target_mols != num_target_mols_) {
+        throw std::runtime_error("target_mols != num_target_mols_");
+    }
+
+    const int tpb = THREADS_PER_BLOCK;
+    // Brute force num_target_atoms X all atoms
+    dim3 dimGrid(BLOCK_SIZE, static_cast<int>(d_target_atom_idxs_.length), 1);
```
badisa, line 116: Having the Y part of the dim being target atoms blows up once we get above 60k atoms. Can work around this, but not sure if it is worth doing in this PR.

timemachine/cpp/src/kernels/k_nonbonded.cuh
badisa: Realized that this doesn't probably matter since we probably aren't interested in proteins as a target molecule.

proteneer:
In the limit of 20k water molecules (or 60k atoms), it's taking around1e8 nanoseconds (or 100 milliseconds) per invocation and is 100x faster than the reference implementation?

This might be okay since:

1) We do one up-front computation at the start of a sequence of MC moves. 
2) Although we re-compute the batch log weights upon acceptance, we don't have to. We just have to swap the old one with the new one upon acceptance.
3) Computation of any *particular* log weight should be extremely fast for the incremental mov (i.e. on the order of 10-20 microseconds).

TL;DR if we run say 10000 MC moves, we can amortize out the cost of the 100 milliseconds initially easily.

timemachine/cpp/src/nonbonded_mol_energy.cu
badisa: Inclined to leave this to when we do a neighborlist implementation of per mol energies, which would resolve this issue. Between now and then, don't forsee cases of a system with more than 60k atoms that need to be evaluated on a per mol basis.

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -430,4 +430,156 @@ void __global__ k_nonbonded_unified(
...
+    RealType ci_x = coords[atom_i_idx * 3 + 0];
+    RealType ci_y = coords[atom_i_idx * 3 + 1];
+    RealType ci_z = coords[atom_i_idx * 3 + 2];
+
+    int atom_j_idx = blockIdx.x * blockDim.x + threadIdx.x;
+    // Zero out the energy buffer
+    block_energy_buffer[threadIdx.x] = 0;
+    while (atom_j_idx < N) {
+        // The two atoms are in the same molecule, don't compute the energies
+        if (!(atom_j_idx >= min_atom_idx && atom_j_idx <= max_atom_idx)) {
```
proteneer, line 490: maybe simplify the predicate logic here?

Define predicates `X = atom_j_idx >= min_atom_idx`, `Y = atom_j_idx <= max_atom_idx`:

`!(X AND Y) == !X OR !Y == (atom_j_idx < min_atom_idx) OR atom_j_idx > max_atom_idx`

proteneer: Also maybe add a comment re: assumptions about continuity of atoms within a molecule.

timemachine/cpp/src/kernels/k_nonbonded.cuh
proteneer: Practically, we'd only be interested in applying something like this to waters or ions.

badisa:
> In the limit of 20k water molecules (or 60k atoms), it's taking around1e8 nanoseconds (or 100 milliseconds) per invocation and is 100x faster than the reference implementation?

Correct, it is 100x faster than the reference implementation at 20k water molecules (60k atoms).

> This might be okay since:
> 
>     1. We do one up-front computation at the start of a sequence of MC moves.
> 
>     2. Although we re-compute the batch log weights upon acceptance, we don't have to. We just have to swap the old one with the new one upon acceptance.
> 
>     3. Computation of any _particular_ log weight should be extremely fast for the incremental mov (i.e. on the order of 10-20 microseconds).
> 
> 
> TL;DR if we run say 10000 MC moves, we can amortize out the cost of the initial 100 milliseconds easily.

We should definitely do this, will be a good performance improvement regardless of implementation. And note that will be faster than 100ms when done all on the GPU as it won't have mallocs and memory copies from the host and back.

timemachine/cpp/src/nonbonded_mol_energy.cu
proteneer: Is the limitation due to max size of `blockDim.y`? Can we just do multiple round-trips for now? We can definitely have systems over 60k atoms.

proteneer (changes requested):

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -430,4 +430,156 @@ void __global__ k_nonbonded_unified(
     }
 }
 
+template <typename RealType, int THREADS>
```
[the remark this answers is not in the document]
badisa: Yes, will rename to clarify

proteneer:
> We should definitely do this, will be a good performance improvement regardless of implementation. And note that will be faster than 100ms when done all on the GPU as it won't have mallocs and memory copies from the host and back.
 
To be clear, in terms of performance, I don't think we need a neighborlist right now, and the limitation re: `blockDim.y` can probably be resolved by doing multiple rounds.

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -430,4 +430,156 @@ void __global__ k_nonbonded_unified(
...
+    const double *__restrict__ params,           // [N, PARAMS_PER_ATOM]
+    const double *__restrict__ box,              // [3, 3],
+    const RealType beta,
+    const RealType cutoff_squared,
+    __int128 *__restrict__ output_energies // [num_target_atoms, gridDim.x]
+) {
+    static_assert(THREADS <= 256 && (THREADS & (THREADS - 1)) == 0);
+    volatile __shared__ __int128 block_energy_buffer[THREADS];
+
+    int row_idx = blockIdx.y;
```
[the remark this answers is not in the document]
badisa: Correct

tests/nonbonded/test_nonbonded_mol_energy.py
```diff
@@ -0,0 +1,105 @@
...
+    def u_test(x, box, params):
+        mol_energies = mol_by_mol_pot.execute(x, params, box)
+        assert mol_energies.shape == (len(group_idxs),)
+
+        # Make sure running again gives bitwise identical results
+        comp_mol_energies = mol_by_mol_pot.execute(x, params, box)
+        np.testing.assert_array_equal(mol_energies, comp_mol_energies)
+        return mol_energies
+
+    np.testing.assert_allclose(u_test(conf, box, params), u_ref(conf, box, params), rtol=rtol, atol=atol)
```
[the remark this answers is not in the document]
proteneer: For 1) maybe you can mimic multiple multiple translation displacements like we have in the BDMover itself? (eg. try a bunch of random displacements of randomly chosen waters).

timemachine/cpp/src/nonbonded_mol_energy.cu
badisa: Correct. Yes, we can add a loop inside and use a fixed block size. Will add the tests then implement this.

tests/nonbonded/test_nonbonded_mol_energy.py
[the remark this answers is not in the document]
badisa: Added tests

1) https://github.com/proteneer/timemachine/pull/1198/commits/a8377d0cbfb35f6e694390b241c0fadfa3f1d67f
2) https://github.com/proteneer/timemachine/pull/1198/commits/44446bf3d39c9c37f8219115524f22085f480ede

timemachine/cpp/src/kernels/k_nonbonded.cuh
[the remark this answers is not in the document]
badisa: Renamed in https://github.com/proteneer/timemachine/pull/1198/commits/9287ee2d4509dc680a598c656fb181d947e30c9f

timemachine/cpp/src/kernels/k_nonbonded.cuh
badisa: Fixed up in https://github.com/proteneer/timemachine/pull/1198/commits/39a594604e770acc32223537222e24a9b0324be3 and added comment about assumption

tests/nonbonded/test_nonbonded_mol_energy.py
```diff
@@ -0,0 +1,227 @@
...
+        np.testing.assert_allclose(
+            ref_mol_energies[comparable_energies], test_mol_energies[comparable_energies], rtol=rtol, atol=atol
+        )
+        # Pull out nans, as they are effectively greater than the threshold
+        non_nan_idx = np.isfinite(test_mol_energies[large_energy_indices])
+        # Large energies are not reliable, so beyond the threshold we simply verify that both the reference and test both exceed the threshold
+        assert np.all(np.abs(test_mol_energies[large_energy_indices][non_nan_idx]) >= threshold)
+
+
+@pytest.mark.parametrize("precision,atol,rtol", [(np.float64, 1e-8, 1e-8), (np.float32, 5e-4, 3e-3)])
```
badisa, line 189: Had to tweak float32 tolerances to be pretty low, not clear to me how to rigorously approach this. Since the float64 is working across 1e-8 tolerances I felt comfortable lowering the float32 tolerances.

timemachine/cpp/src/nonbonded_mol_energy.cu
badisa: Done in https://github.com/proteneer/timemachine/pull/1198/commits/b03a0f27ae381d34b3b488a1d3ee8c9acf89a923

Seems about 10% faster as well, which is a nice benefit.

tests/nonbonded/test_nonbonded_mol_energy.py
proteneer: Hm  this is still a lot lower than we typically allow for. Typically, we compute *total* energies (as a scalar), however, here we're computing *per molecule* energies (as a vector). So the effect seems even bigger than desired. I do think we should investigate this a bit more:

Can you print out each component of the energy vector and see if the discrepancy is coming from waters that have *large* molecular energies or waters that have *small* molecular energies? If the discrepancy is coming from waters that have *small* molecular energies, this is less worrisome (since they will have negligible weights after exponentiation), but if it's coming from waters that have *large* molecular energies, then this is definitely a problem.

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -422,12 +422,165 @@ void __global__ k_nonbonded_unified(
...
+
+            atom_j_idx += gridDim.x * blockDim.x;
+            // Always zero before the end of the loop, in case one thread no longer loops
+            block_energy_buffer[threadIdx.x] = 0;
+        }
+        row_idx += gridDim.y * blockDim.y;
+    }
+}
+
+template <typename RealType, int BLOCK_SIZE>
```
proteneer, line 557: `BLOCK_SIZE` == `THREADS_PER_BLOCK`? (If so, best to be consistent in notation)

timemachine/cpp/src/nonbonded_mol_energy.cu
```diff
@@ -0,0 +1,112 @@
...
+template <typename RealType>
+NonbondedMolEnergyPotential<RealType>::NonbondedMolEnergyPotential(
+    const int N, const std::vector<std::vector<int>> target_mols, const double beta, const double cutoff)
+    : N_(N), num_target_mols_(target_mols.size()), beta_(static_cast<RealType>(beta)),
+      cutoff_squared_(static_cast<RealType>(cutoff * cutoff)) {
+    verify_group_idxs(N_, target_mols);
+
+    std::array<std::vector<int>, 3> target_flattened_groups = prepare_group_idxs_for_gpu(target_mols);
+
+    d_target_atom_idxs_.realloc(target_flattened_groups[0].size());
```
proteneer, line 22: would be nice if we could set these members to `const`, (probably time for another constructor that takes in a vector)

proteneer: (non-blocking, can save for deboggling)

timemachine/cpp/src/kernels/k_nonbonded.cuh
badisa: So `BLOCK_SIZE` is the number of threads per block that were run in `k_compute_nonbonded_target_atom_energies` but is not necessarily the number of threads per block run in  `k_compute_nonbonded_target_atom_energies`. 

Will add a comment.

timemachine/cpp/src/nonbonded_mol_energy.cu
badisa: We would need to decompose the target group indices prior to the constructor to do that. Which we could do in the wrapped kernels, though not sure how much of that sort of logic in the wrapped kernels.

timemachine/cpp/src/kernels/k_nonbonded.cuh
badisa: Renamed in https://github.com/proteneer/timemachine/pull/1198/commits/50ee91eb83f4c1138b4afd01b10e81bc4acd478d

tests/nonbonded/test_nonbonded_mol_energy.py
badisa: Need to investigate in more detail, but from what I remember the absolute errors were about 2.0 and the relative error was 0.002, so would indicate a large energy disagreeing. And we already have issues with large (magnitude) molecular energies disagreeing due to the int128 handling of overflows.

proteneer: This looks like a system straight from build_water_system, there might be a clash?

badisa: Will try minimizing and see if that helps. Though most of the time this is going to be looking at clashy systems, so not clear how that makes a difference?

proteneer: It's the expectations compared to known systems we've benchmarked and tested before (which IIRC are non-clashy systems).

1. For a non-clashy system, we do *not* expect to have large errors in 32bit (I was mainly assuming this was the case, and hence why I requested explicitly a clashy test system).
2. For a clashy system, we expect that the *clashy* atoms/molecules have a large discrepancy in 32bit.

badisa: Thanks for clarify. You were spot on, the system contained a clash and because it hadn't been minimized the difference was greater. Added 1000 steps of 1.5e-3 timestep (2.5e-3 blows up) and was able to decrease the tolerances: https://github.com/proteneer/timemachine/pull/1198/commits/c518a0c1da4dae3f5f6522a4781b8937486b579d

proteneer (approved):

PR #1199, opened 2023-10-31
Clean up code related to BDExchange

badisa:
Some various changes that I made while working on bias deletion for C++ that seemed independent enough to warrant clean up as a separate PR.


proteneer (approved):
lgtm.

PR #1201, opened 2023-11-02
Return NaN if converting invalid __int128

badisa:


badisa:
Messing around with this a bit more, there is some nuance to this. We return nan to python as we don't know if it is +inf or -inf. As it stands that is probably not the right solution for bias deletion since we really want to convert the nan to inf. And we can do that more explicitly elsewhere.

PR #1202, opened 2023-11-03
Cuda Implementation of BDExchangeMove

badisa:
- Had to fix up LogSumExp to not change values

# Benchmarks
GPU A4000
Cuda Arch 8.6

For the buckyball case in the examples with 6 waters, it is about ~2x faster~ 4-5x slower (3.49ms) on than the JAX implementation (0.8-1.0ms). This does depend on the GPU and CPU. 

# Todo
- [x] Add support for example with the C++ implementation in a vanilla box of water (https://github.com/proteneer/timemachine/pull/1202/commits/afaa92fdbbf64cc8408a60e1814bb1246f0fe421)
- [x] Figure out the API a bit more (should steps per move be a constructor? Etc) (https://github.com/proteneer/timemachine/pull/1202/commits/4f8bd0889de1f1ffdae832f0d025300e1145bb99)
- [x] Translate particles within the home box for random translations (https://github.com/proteneer/timemachine/pull/1202/commits/2cbae8cb47e51e1b13a2ba87ec937f79198c3d23)
- [x] ~Switch MC example to C++ variant once validated that 1000 moves are accepted in a similar amount of time~ Still slow to put into the examples, need the incremental change

# Follow-up

- Incremental log weights
- Targeted Insertion implementation


timemachine/cpp/src/bd_exchange_move.cu
```diff
@@ -0,0 +1,186 @@
...
+            d_coords,
+            d_num_accepted_.data);
+        gpuErrchk(cudaPeekAtLastError());
+        num_attempted_++;
+    }
+}
+
+template <typename RealType>
+std::array<std::vector<double>, 2>
+BDExchangeMove<RealType>::move_host(const int N, const int num_moves, const double *h_coords, const double *h_box) {
```
badisa, line 143: Should this match the API of the the MonteCarloMove so that it is `move(coords, box)` and instead pass in a number of proposals per call to move?

timemachine/cpp/src/bd_exchange_move.cu
```diff
@@ -0,0 +1,186 @@
...
+                d_log_weights_before_.data,
+                d_log_weights_after_.data);
+            gpuErrchk(cudaPeekAtLastError());
+        }
+        // Make a copy of the coordinates
+        gpuErrchk(cudaMemcpyAsync(
+            d_intermediate_coords_.data, d_coords, d_intermediate_coords_.size(), cudaMemcpyDeviceToDevice, stream));
+
+        // Quaternions generated from normal noise, while translations and the acceptance value are uniform
+        curandErrchk(templateCurandNormal(cr_rng_, d_quaternions_.data, d_quaternions_.length, 0.0, 1.0));
```
badisa, line 97: No batching yet, but so far this is like 0.5% of the total time. Its computing the per mol energies that are expensive.

timemachine/cpp/src/bd_exchange_move.cu
badisa: Inclined to make it match, so we can define a `MonteCarloMove` base class in C++ that has the signature `move_device(coords, box, stream)` (or another name if there is a suggestion).

proteneer: I'd

1) Keep the API of the python layer simple, for multiple moves, there are multiple variations (multiple sequential moves, multiple-try-metropolis type of moves, etc), mixing and matching different kinds of moves (eg. HREX moves). The `moves.py` file defines some of these

2) C++ is up to you! (Not sure if we want consistency with the barostat moves, HREX is implemented purely in python right now)

timemachine/cpp/src/bd_exchange_move.cu
proteneer: Yeah I don't expect this to be expensive relative to other bits of the code.

timemachine/cpp/src/bd_exchange_move.cu
badisa: > Not sure if we want consistency with the barostat moves

Think they should be consistent, that way we can have a series of MC moves in with the context. Will need to change a few things to make it all work, but as we have more MC moves would be nice to have a singular API.

badisa:
An observation having been running the 6 water buckyball example with this implementation over the last couple of days is that it seems to alternate between one water and two waters in the bucky ball but at a slower rate than what I have seen of the CPU version. Unsure if this is difference in randomness or if there is some disconnect between the implementations.

badisa:
Having run until the CPU and the GPU version got to >1000 accepted moves in a water box

CPU:
1001 / 7380000 | density 998.5856043371585 | # of waters in spherical region 0 | md step: 7370000
Acceptance Rate 0.0135%

GPU:
1003 / 6710000 | density 994.8237690435188 | # of waters in spherical region 0 | md step: 6700000
Acceptance Rate: 0.0149%

Additionally the initial benchmarking was way off thank to a background process I was running that was consuming a bunch of CPU time.
The CPU is 4-5x FASTER (0.8-1.0ms) than the GPU implementation (4ms) as done in this PR. Which further supports the need for the incremental energy computations. 

Based on the previous benchmarking in https://github.com/proteneer/timemachine/pull/1198#issue-1964407139 I would expect to see a speed up going to the incremental mode, though we may need to consider a neighborlist enabled version to get a more significant speed up.

proteneer:
```
CPU:
1001 / 7380000 | density 998.5856043371585 | # of waters in spherical region 0 | md step: 7370000
Acceptance Rate 0.0135%

GPU:
1003 / 6710000 | density 994.8237690435188 | # of waters in spherical region 0 | md step: 6700000
Acceptance Rate: 0.0149%
```

Nice! this looks reasonably comparable to me!

> The CPU is 4-5x FASTER (0.8-1.0ms) than the GPU implementation (4ms) as done in this PR. Which further supports the need for the incremental energy computations.

Not a big deal, the GPU is doing about 1000x more work than it needs to right now.

tests/test_cuda_bd_exchange_mover.py
```diff
@@ -0,0 +1,192 @@
...
+                np.exp(bdem.last_log_probability()),
+                np.exp(np.minimum(logsumexp(before_log_weights) - logsumexp(after_log_weights), 0.0)),
+                rtol=rtol,
+                atol=atol,
+            )
+        if num_moved == 0:
+            np.testing.assert_array_equal(last_conf, x_move)
+        assert num_moved <= 1, "More than one mol moved, something is wrong"
+        last_conf = x_move
+    assert accepted > 0
```
proteneer, line 81: What's the actual number of accepted moves? (The description says "Given two waters directly on top of each other in a box, the exchange mover should accept almost any move") This sounds like its testing for accepting at least one move?

tests/test_cuda_bd_exchange_mover.py
```diff
@@ -0,0 +1,192 @@
...
+                    np.exp(bdem.last_log_probability()),
+                    np.exp(np.minimum(logsumexp(before_log_weights) - logsumexp(after_log_weights), 0.0)),
+                    rtol=rtol,
+                    atol=atol,
+                )
+        if num_moved == 0:
+            np.testing.assert_array_equal(last_conf, x_move)
+        assert num_moved <= 1, "More than one mol moved, something is wrong"
+        last_conf = x_move
+    assert accepted > 0, "No moves were made, nothing was tested"
```
proteneer, line 186: what's typical accepted here?

proteneer: (if acceptance rate is someting like 0.01 then 2500*0.01 should be around 25 moves accepted on average?)

timemachine/cpp/src/bd_exchange_move.cu
```diff
@@ -0,0 +1,187 @@
...
+    mol_potential_.mol_energies_device(
+        N,
+        num_target_mols_,
+        d_coords,
+        d_params_.data,
+        d_box,
+        d_mol_energy_buffer_.data, // Don't need to zero, will be overridden
+        stream);
+
+    // Don't need to normalize to sample
```
proteneer, line 71: nice!

timemachine/cpp/src/bd_exchange_move.cu
```diff
@@ -0,0 +1,187 @@
...
+                d_log_weights_after_.data);
+            gpuErrchk(cudaPeekAtLastError());
+        }
+        // Make a copy of the coordinates
+        gpuErrchk(cudaMemcpyAsync(
+            d_intermediate_coords_.data, d_coords, d_intermediate_coords_.size(), cudaMemcpyDeviceToDevice, stream));
+
+        // Quaternions generated from normal noise, while translations and the acceptance value are uniform
+        curandErrchk(templateCurandNormal(cr_rng_, d_quaternions_.data, d_quaternions_.length, 0.0, 1.0));
+        // The last translation value is to be used to determine acceptance
```
proteneer, line 99: wait what?

timemachine/cpp/src/bd_exchange_move.cu
```diff
@@ -0,0 +1,187 @@
...
+    const int seed,
+    const int proposals_per_move)
+    : N_(N), proposals_per_move_(proposals_per_move), num_target_mols_(target_mols.size()),
+      beta_(static_cast<RealType>(1.0 / (BOLTZ * temperature))), mol_potential_(N, target_mols, nb_beta, cutoff),
+      sampler_(num_target_mols_, seed), logsumexp_(N), d_intermediate_coords_(N * 3), d_params_(params.size()),
+      d_mol_energy_buffer_(num_target_mols_), d_mol_offsets_(get_mol_offsets(target_mols).size()),
+      d_log_weights_before_(num_target_mols_), d_log_weights_after_(num_target_mols_),
+      d_log_probabilities_before_(num_target_mols_), d_log_probabilities_after_(num_target_mols_),
+      d_log_sum_exp_before_(2), d_log_sum_exp_after_(2), d_samples_(1), d_quaternions_(round_up_even(4)),
+      d_translations_(round_up_even(4)), d_num_accepted_(1), num_attempted_(0) {
```
proteneer, line 30: `round_up_even_` here looks superfluous?

timemachine/cpp/src/bd_exchange_move.cu
proteneer: edit: okay looks like the buffer is [x,y,z,w] where [x,y,z] are a random translation and w is used for acceptance

tests/test_cuda_bd_exchange_mover.py
```diff
@@ -0,0 +1,192 @@
...
+                num_moved += 1
+        if num_moved > 0 and proposals_per_move == 1:
+            accepted += 1
+            # Verify that the probabilities agree when we do accept moves
+            np.testing.assert_allclose(
+                np.exp(bdem.last_log_probability()),
+                np.exp(np.minimum(logsumexp(before_log_weights) - logsumexp(after_log_weights), 0.0)),
+                rtol=rtol,
+                atol=atol,
+            )
```
proteneer, line 76: might be useful to also add a test ensuring that home-box imaging code for translations work here: i.e. the moved water is centered into the home box?

timemachine/cpp/src/bd_exchange_move.hpp
```diff
@@ -0,0 +1,76 @@
...
+    const RealType beta_; // 1 / kT
+    NonbondedMolEnergyPotential<RealType> mol_potential_;
+    WeightedRandomSampler<RealType> sampler_;
+    LogSumExp<RealType> logsumexp_;
+    // Buffer for evaluating moves without touching the original coords
+    const DeviceBuffer<double> d_intermediate_coords_;
+    const DeviceBuffer<double> d_params_;
+    const DeviceBuffer<__int128> d_mol_energy_buffer_;
+    const DeviceBuffer<int> d_mol_offsets_;
+    const DeviceBuffer<RealType> d_log_weights_before_;
```
proteneer, line 29: why are these things const? Semantically seems weird (since the internals of them do change), i.e. if they were pointers `RealType * d_log_weights_before_`; they should not const-able since the internals change

timemachine/cpp/src/mol_utils.cpp
```diff
@@ -1,6 +1,7 @@
 #include "mol_utils.hpp"
 #include <algorithm>
 #include <set>
+#include <stdexcept>
```
proteneer, line 4: is this used?

proteneer (approved):
awesome work :) only minor comments

timemachine/cpp/src/mol_utils.cpp
badisa: This is used for the runtime error. Added since under newer versions of GCC this doesn't compile without the addition. Can remove, but seemed like a harmless addition to compile with newer GCC

timemachine/cpp/src/bd_exchange_move.hpp
badisa: You are right, it is weird. Will remove the const

timemachine/cpp/src/bd_exchange_move.cu
badisa: Mostly for pedagogical purposes as I have been bitten by that a few times.

timemachine/cpp/src/bd_exchange_move.cu
badisa: Yup, will expand the comment to be clearer

timemachine/cpp/src/kernels/k_exchange.cuh
```diff
@@ -0,0 +1,81 @@
...
+void __global__ k_attempt_exchange_move(
+    const int N,
+    const RealType *__restrict__ rand,               // [1]
+    const RealType *__restrict__ before_log_sum_exp, // [2]
+    const RealType *__restrict__ after_log_sum_exp,  // [2]
+    const double *__restrict__ moved_coords,         // [N, 3]
+    double *__restrict__ dest_coords,                // [N, 3]
+    size_t *__restrict__ num_accepted                // [1]
+) {
+    int idx = blockIdx.x * blockDim.x + threadIdx.x;
```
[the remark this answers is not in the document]
badisa: It is atom, will rename

timemachine/cpp/src/mol_utils.cpp
proteneer: No need to remove, makes sense.

tests/test_cuda_bd_exchange_mover.py
badisa: The acceptance rates I stated before (https://github.com/proteneer/timemachine/pull/1202#issuecomment-1797201324) were percentages, so it is 0.01 **%** which means the expected here would be 0.25 moves. The typical accepted here is 1/2 (float64 results in 1 move, float32 results in 2 moves)

timemachine/cpp/src/bd_exchange_move.cu
proteneer: sgtm

tests/test_cuda_bd_exchange_mover.py
proteneer: oof

timemachine/cpp/src/bd_exchange_move.cu
badisa: Expanded comment https://github.com/proteneer/timemachine/pull/1202/commits/6c1497e725916de531ccae6501a41fe9b233b768

tests/test_cuda_bd_exchange_mover.py
badisa: Added image checks https://github.com/proteneer/timemachine/pull/1202/commits/70fbc7fc996f6ebbd99e071f6b178921b39f3181

tests/test_cuda_bd_exchange_mover.py
proteneer: would this test be a bit brittle? may run 10x longer so we average 10-20 moves?

tests/test_cuda_bd_exchange_mover.py
badisa: Made it stricter to accept all moves, though that is not strictly always true. So changing the seed could break tests https://github.com/proteneer/timemachine/pull/1202/commits/038c1564c5d607e2f871c54986cb0ecb559d7154

timemachine/cpp/src/bd_exchange_move.hpp
badisa: Removed const in https://github.com/proteneer/timemachine/pull/1202/commits/07f449302f32806180e9540e0a7de8ba55d3dbff

tests/test_cuda_bd_exchange_mover.py
badisa: This test is brittle, was a trade off between test time and verifying logic. Since the acceptance rate is so low didn't want to go too longer, especially with how slow it is currently.

Tried 20k steps and only get 5 moves. Trying to find the right combination for a robust nightly test

badisa: Added a nightly test: https://github.com/proteneer/timemachine/pull/1202/commits/34d84e7a171b1920f6317bd8155742b931f50b2c

PR #1204, opened 2023-11-08
Compute BiasDeletion Log Weights Incrementally

badisa:
- Significantly faster going from 3ms per move to 0.06ms per move (~700ms per 10k moves)
- DeviceBuffers can be constructed directly from vectors now
- Moves test from nightly to regular CI

# Acceptance rate

Identical to the previous implementation (https://github.com/proteneer/timemachine/pull/1202) with an acceptance Rate of 0.0149%
```
1003 / 6710000 | density 994.8237690435188 | # of waters in spherical region 0 | md step: 6700000
```

Ran this out to 315 million steps (45696 / 315660000) and still getting a 0.014% acceptance rate in bulk water, so this appears converged.

# Todo

- [x] ~Fix bug where if you shuffle the group indices the code breaks.~ The reference doesn't support shuffling the indices (would need to shuffle the per mol energies into the right place) and as of https://github.com/proteneer/timemachine/pull/1204/commits/b9d6fd3209769ab954c0bdf20a7ad29de624482b it is validated that the indices are correctly ordered. We may want to revisit this for ion sampling, but seems worthwhile to just move ahead for now.


badisa:
Still investigating, but actually think the reference `BDExchangeMove` might be wrong in the case where you have a protein.

Here we setup the indices to be a water molecule by all coordinates: https://github.com/proteneer/timemachine/blob/master/timemachine/md/exchange/exchange_mover.py#L126-L128
And when we then construct the per water energies here: https://github.com/proteneer/timemachine/blob/master/timemachine/md/exchange/exchange_mover.py#L132-L134 we assume that the waters are first. However if you build a protein system it will be protein_coords then water coords. In the process of verifying this understanding of the code.

timemachine/cpp/src/bd_exchange_move.cu
```diff
@@ -100,28 +131,89 @@ void BDExchangeMove<RealType>::move_device(
...
             d_samples_.data,
-            d_mol_offsets_.data,
+            d_target_mol_offsets_.data,
             d_quaternions_.data,
             d_translations_.data,
             d_intermediate_coords_.data);
         gpuErrchk(cudaPeekAtLastError());
 
-        mol_potential_.mol_energies_device(
+        k_atom_by_atom_energies<<<atom_by_atom_grid, tpb, 0, stream>>>(
```
proteneer, line 181: Not urgent, but I wonder if we can combine the two `k_atom_by_atom_energies` later on (to lessen kernel overhead etc.), i.e. given `old_mol_pos`, `new_mol_pos`, and `other_coords`, (we'd only have to load in `other_coords` once as opposed to twice)

timemachine/cpp/src/kernels/k_exchange.cuh
```diff
@@ -11,6 +14,33 @@ template <typename RealType> RealType __host__ __device__ convert_nan_to_inf(con
...
+    int idx = blockIdx.x * blockDim.x + threadIdx.x;
+    while (idx < num_samples) {
+        int mol_idx = samples[idx];
+        int mol_start = mol_offsets[mol_idx];
+        int mol_end = mol_offsets[mol_idx + 1];
+        output_mol_offsets[mol_idx] = target_atoms[mol_start];
+        output_mol_offsets[mol_idx + 1] = target_atoms[mol_end - 1] + 1;
+        int num_atoms = mol_end - mol_start;
+
+        assert(num_atoms == sample_atoms);
```
proteneer, line 34: are `asserts` a no-op when compiling with `-O3`?

timemachine/cpp/src/bd_exchange_move.cu
```diff
@@ -75,7 +96,9 @@ void BDExchangeMove<RealType>::move_device(
 
     logsumexp_.sum_device(num_target_mols_, d_log_weights_before_.data, d_log_sum_exp_before_.data, stream);
 
-    const int num_samples = 1;
+    dim3 atom_by_atom_grid(ceil_divide(N, tpb), mol_size_, 1);
+
+    const int num_samples = NUM_SAMPLES;
     for (int move = 0; move < proposals_per_move_; move++) {
```
proteneer, line 102: General question - which kernel takes up the most time for this set of kernels?

timemachine/cpp/src/kernels/k_exchange.cuh
badisa: They are not, this does trigger with `-O3`

timemachine/cpp/src/bd_exchange_move.cu
badisa: We could combine these two calls into a single call

tests/nonbonded/test_atom_by_atom_energies.py
```diff
@@ -0,0 +1,48 @@
...
+    params = nb.params
+
+    beta = nb.potential.beta
+    cutoff = nb.potential.cutoff
+    func = custom_ops.atom_by_atom_energies_f32
+    if precision == np.float64:
+        func = custom_ops.atom_by_atom_energies_f64
+
+    target_atoms = np.array(target_atoms).astype(np.int32)
+    comp_atoms = np.arange(N).astype(np.int32)
```
proteneer, line 43: rename `comp` -> `complete` or `all`

proteneer: (Does this test break if `comp_atoms` is shuffled?)

timemachine/cpp/src/kernels/k_exchange.cuh
```diff
@@ -78,4 +108,83 @@ void __global__ k_store_accepted_log_probability(
...
+        int max_atom_idx = mol_atoms_idxs[mol_end - 1];
+
+        // A loop that in the case of water will be 3x3
+        for (int i = 0; i < mol_size; i++) {
+            for (int j = min_atom_idx; j <= max_atom_idx; j++) {
+                weight_accumulator += FLOAT_TO_FIXED_ENERGY<RealType>(inv_kT * per_atom_energies[i * N + j]);
+            }
+        }
+
+        weight_accumulator = Negated ? FLOAT_TO_FIXED_ENERGY<RealType>(current_log_weight) - weight_accumulator
```
proteneer, line 139: I think this is right, but just wanted to triple check that we want Negation to happen in fixed point as opposed to floating point (to be consistent with how we're doing cancellation of nans in exclusions).

timemachine/cpp/src/bd_exchange_move.cu
```diff
@@ -100,28 +131,89 @@ void BDExchangeMove<RealType>::move_device(
...
+            d_coords,
+            d_params_.data,
+            d_box,
+            nb_beta_,
+            cutoff_squared_,
+            d_sample_per_atom_energy_buffer_.data);
+        gpuErrchk(cudaPeekAtLastError());
+
+        // Subtract off the weights for the individual waters from the sampled water
+        // clobbers the sampled mol energy value, but we have to special case that anyways
```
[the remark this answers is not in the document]
badisa: Will clarify, but simply intended to mean that the sampled mol energy is in a bad state for the calls to `k_adjust_weights` as the kernel doesn't treat the sampled weight any differently (for simplicity) and that you need to call `k_set_sampled_weight`

timemachine/cpp/src/kernels/k_exchange.cuh
badisa: It does need to be consistent with how we are doing cancellation of nans, as we could have a nan due to the interaction of the water we are deleting and we would want that to `zero` out vs continue to be a nan.

tests/nonbonded/test_atom_by_atom_energies.py
badisa: This test does break if they are shuffled. We could pass the `all_atoms` into the wrapped function as well to make it work.

timemachine/cpp/src/kernels/k_exchange.cuh
```diff
@@ -78,4 +108,83 @@ void __global__ k_store_accepted_log_probability(
...
+    const int N,
+    const int mol_size,
+    const int num_samples,
+    const int *__restrict__ samples, // [num_samples]
+    const int *__restrict__ target_atoms,
+    const int *__restrict__ mol_offsets,
+    const RealType *__restrict__ per_atom_energies,
+    const RealType inv_kT, // 1 / kT
+    RealType *__restrict__ log_weights) {
+    static_assert(THREADS_PER_BLOCK <= 512 && (THREADS_PER_BLOCK & (THREADS_PER_BLOCK - 1)) == 0);
```
proteneer, line 160: What's the purpose of the second part `(THREADS_PER_BLOCK & (THREADS_PER_BLOCK - 1)) == 0`? Check that it's a power of 2?

timemachine/cpp/src/kernels/k_exchange.cuh
```diff
@@ -78,4 +108,83 @@ void __global__ k_store_accepted_log_probability(
...
+    assert(num_samples == 1);
```
proteneer, line 161: again, check to see if `assert` in `O3` has perf impact?

timemachine/cpp/src/kernels/k_exchange.cuh
```diff
@@ -78,4 +108,83 @@ void __global__ k_store_accepted_log_probability(
...
+    assert(gridDim.x == 1);
+    __shared__ __int128 accumulators[THREADS_PER_BLOCK];
+
+    int sample_idx = samples[0];
+    int min_atom_idx = target_atoms[0];
+    int max_atom_idx = target_atoms[mol_size - 1];
+
+    int atom_idx = blockIdx.x * blockDim.x + threadIdx.x;
+    // Zero all of the accumulators
+    accumulators[threadIdx.x] = 0;
```
proteneer, line 171: do you need a sync barrier here over shared mem?

badisa: Yes

timemachine/cpp/src/kernels/k_exchange.cuh
```diff
@@ -78,4 +108,83 @@ void __global__ k_store_accepted_log_probability(
     }
 }
 
+template <typename RealType, bool Negated>
+void __global__ k_adjust_weights(
```
proteneer, line 112: docstring please

timemachine/cpp/src/kernels/k_exchange.cuh
```diff
@@ -78,4 +108,83 @@ void __global__ k_store_accepted_log_probability(
...
+        RealType current_log_weight = log_weights[mol_idx];
+        __int128 weight_accumulator = 0;
+
+        int mol_start = mol_offsets[mol_idx];
+        int mol_end = mol_offsets[mol_idx + 1];
+        int min_atom_idx = mol_atoms_idxs[mol_start];
+        int max_atom_idx = mol_atoms_idxs[mol_end - 1];
+
+        // A loop that in the case of water will be 3x3
+        for (int i = 0; i < mol_size; i++) {
```
proteneer, line 133: (not important for this PR, but mol size may also be 1 later on when doing counter-ion sampling).

timemachine/cpp/src/bd_exchange_move.cu
badisa: Timings for the kernels for 2500 steps per move

The slowest is the `k_set_sampled_weight`

```
Time (%)  Total Time (ns)  Instances  Avg (ns)   Med (ns)   Min (ns)  Max (ns)  StdDev (ns)                                                  Name                                                
 --------  ---------------  ---------  ---------  ---------  --------  --------  -----------  ----------------------------------------------------------------------------------------------------
     17.5       26,416,574      2,500   10,566.6   10,640.0     8,768   425,408     10,071.8  void timemachine::k_set_sampled_weight<float, (int)512>(int, int, int, const int *, const int *, co…
     13.0       19,620,340      5,000    3,924.1    4,032.0     3,359    13,280        368.8  void timemachine::k_atom_by_atom_energies<float>(int, int, const int *, const double *, const doubl…
      7.5       11,289,498      2,500    4,515.8    4,672.0     3,839    12,576        418.8  void timemachine::k_rotate_and_translate_mols<float>(int, const double *, const double *, const int…
      7.3       10,983,198      2,500    4,393.3    4,543.0     3,807    11,968        338.8  void timemachine::k_adjust_weights<float, (bool)1>(int, int, int, const int *, const int *, const T…
      7.3       10,977,489      2,500    4,391.0    4,543.0     3,807    16,864        418.8  void timemachine::k_adjust_weights<float, (bool)0>(int, int, int, const int *, const int *, const T…
      7.2       10,863,049      5,000    2,172.6    2,208.0     1,727    12,737        280.3  void gen_sequenced<curandStateXORWOW, float, int, &curand_uniform_noargs<curandStateXORWOW>, rng_co…
      4.9        7,448,702      2,500    2,979.5    3,072.0     2,591    12,896        295.0  void cub::DeviceReduceSingleTileKernel<cub::DeviceReducePolicy<cub::KeyValuePair<int, float>, cub::…
      4.7        7,129,433      2,501    2,850.6    2,880.0     2,400    12,864        663.2  void cub::DeviceReduceSingleTileKernel<cub::DeviceReducePolicy<float, float, int, cub::Max>::Policy…
      4.6        6,886,243      2,501    2,753.4    2,816.0     2,399     6,560        201.6  void cub::DeviceReduceSingleTileKernel<cub::DeviceReducePolicy<float, float, int, cub::Sum>::Policy…
      4.1        6,240,954      2,500    2,496.4    2,560.0     2,175     5,344        184.1  timemachine::k_setup_sample_atoms(int, int, const int *, const int *, const int *, int *, int *)    
      3.9        5,834,797      2,500    2,333.9    2,400.0     1,855     7,680        266.4  void gen_sequenced<curandStateXORWOW, float2, normal_args_st, &curand_normal_scaled2<curandStateXOR…
      3.5        5,202,291      2,499    2,081.7    2,144.0     1,823    12,736        258.6  void timemachine::k_store_accepted_log_probability<float>(int, const T1 *, T1 *, const T1 *, T1 *, …
      3.5        5,202,259      2,500    2,080.9    2,112.0     1,792     8,160        233.3  void timemachine::k_setup_gumbel_max_trick<float>(int, const T1 *, T1 *)                            
      3.4        5,167,353      2,500    2,066.9    2,112.0     1,791     6,848        185.8  void timemachine::k_copy_kv_key<float>(int, const cub::KeyValuePair<int, T1> *, int *)              
      3.4        5,136,022      2,500    2,054.4    2,112.0     1,791     6,336        159.8  void timemachine::k_attempt_exchange_move<float>(int, const T1 *, const T1 *, const T1 *, const dou…
      3.4        5,056,284      2,501    2,021.7    2,049.0     1,759    11,360        278.1  void timemachine::k_exp_sub_max<float>(int, const T1 *, const T1 *, T1 *)                           
      0.5          781,280          1  781,280.0  781,280.0   781,280   781,280          0.0  void timemachine::k_compute_nonbonded_target_atom_energies<float, (int)64>(int, int, const int *, c…
      0.2          269,760          2  134,880.0  134,880.0   114,240   155,520     29,189.4  void generate_seed_pseudo<rng_config<curandStateXORWOW, (curandOrdering)101>>(unsigned long long, u…
```

timemachine/cpp/src/kernels/k_exchange.cuh
badisa: Correct, as this kernel expects everything is a power of 2

timemachine/md/exchange/exchange_mover.py
```diff
@@ -130,7 +139,10 @@ def batch_log_weights_incremental(conf, box, water_idx, new_pos, initial_weights
             # sum interaction energy of all the atoms in a water molecule
             old_water_ixn_nrgs = jnp.sum(self.beta * U_fn_unsummed(conf, box, a_idxs, b_idxs), axis=0)
             old_water_water_ixn_nrgs = jnp.sum(
-                old_water_ixn_nrgs[: (self.num_waters - 1) * 3].reshape((self.num_waters - 1), 3), axis=1
+                old_water_ixn_nrgs[self.starting_water_position :][: (self.num_waters - 1) * 3].reshape(
```
proteneer, line 142: Ah on first glance I thought this was a 2D array, but realized it's just doing:
```
old_water_ixn_nrgs[self.starting_water_position : self.starting_water_position + (self.num_waters - 1)]
```

proteneer (approved):
lgtm, mostly clerical comments

tests/nonbonded/test_atom_by_atom_energies.py
proteneer: okay, can we add a comment just saying this will break if shuffled?

timemachine/cpp/src/kernels/k_exchange.cuh
badisa: We will have to do more work to support mixed size of mols as you need to know the size of all preceding mols. The same goes for the reference, so left to only handle cases where all of the mols are the same size

timemachine/md/exchange/exchange_mover.py
badisa: Yup, but if you write it that way, JAX complains about dynamic sizes when JITed which is why it is done in two steps.

timemachine/cpp/src/kernels/k_exchange.cuh
badisa: It has at most a 100ns difference in the kernels (and I doubt that is even converged). Which seems to me negligible

proteneer: Okay, thanks for checking!

timemachine/md/exchange/exchange_mover.py
badisa: Correct, not sure why it complained. Didn't investigate further.

timemachine/cpp/src/kernels/k_exchange.cuh
badisa: Changed to use local accumulator in https://github.com/proteneer/timemachine/pull/1204/commits/b4e52490b328df745b592a2e043d194e125a5e61

timemachine/cpp/src/kernels/k_exchange.cuh
badisa: Added docstring https://github.com/proteneer/timemachine/pull/1204/commits/8e9130a1280f692b59498fd6a5c475fbcd3c5ad8

timemachine/cpp/src/bd_exchange_move.cu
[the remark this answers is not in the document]
badisa: Clarified comment in https://github.com/proteneer/timemachine/pull/1204/commits/56c55c30294dd21e284d09f6cee01130f7f88e09

tests/nonbonded/test_atom_by_atom_energies.py
badisa: Renamed comp to all in https://github.com/proteneer/timemachine/pull/1204/commits/56c55c30294dd21e284d09f6cee01130f7f88e09

Added comment stating that shuffling breaks the test: https://github.com/proteneer/timemachine/pull/1204/commits/80a1945901496c6318afc49b452bf0402aa8be3a

timemachine/cpp/src/bd_exchange_move.cu
badisa: Because it has to be computed in a single block due to being unable to do atomic accumulation on `__int128`

PR #1205, opened 2023-11-16
Cuda Implementation of Targeted Insertion

badisa:
Implementation of Targeted Insertion written in Cuda

# Benchmarks
Appears to be about ~24x faster than the JAX implementation.

## Jax Targeted Insertion
```
1 / 10000 | density 979.0931203406594 | # of waters in spherical region 1 | md step: 0 | time per mc move: 3159684.6349ns
```

## Cuda Targeted Insertion
Nvidia A4000
Cuda Arch 8.6
```
1 / 10000 | density 979.0931203406594 | # of waters in spherical region 1 | md step: 0 | time per mc move: 131927.7264ns
```

# Validation
Comparing the reference targeted insertion to the Cuda implementation using the following setup:
```
python -u examples/water_sampling_mc.py --water_pdb timemachine/datasets/water_exchange/bb_0_waters.pdb --ligand_sdf timemachine/datasets/water_exchange/bb_centered_espaloma.sdf --out_cif traj_0_waters.cif --md_steps_per_batch 10000 --mc_steps_per_batch 10000 --insertion_type targeted --use_hmr 1
```

## Acceptance Rate
~9.5e-7 acceptance in the case of the espaloma charged buckyball for both. Comparings the moves where both implementations got to `17` accepted moves.

Graph showing acceptance over number of moves, both CPU and Cuda implementation agree (within the number of steps that both have run)
image

### Reference
```
17 / 17460000 | density 984.6849481445907 | # of waters in spherical region 1 | md step: 17450000 | time per mc move: 2846452.9843ns
```

### Cuda
```
17 / 17810000 | density 983.47768990185 | # of waters in spherical region 1 | md step: 17800000 | time per mc move: 136913.8105ns
```

## Water Occupancies
The percentages of time the buckyball had either one or two waters in it.

Graph showing the number of waters in the region. Started from simulation with 0 waters, but in the first 10k moves jumps up to 1 water for both which is why the number of waters doesn't start at 0
image

### Reference
Total Moves 99770000
Num Waters in Buckyball: 1, 61.11%
Num Waters in Buckyball: 2, 38.89%

#### Starting from 0 Waters
Total Moves 509140000
Num Waters in Buckyball: 1, 59.19%
Num Waters in Buckyball: 2, 40.81%

#### Starting from 6 Waters
Total Moves 444100000
Num Waters in Buckyball: 4, 0.00%
Num Waters in Buckyball: 2, 50.71%
Num Waters in Buckyball: 1, 49.28%

image

## Average Occupancy starting from 0 or 6 waters in buckyball

There appears to be a good amount of variance, but seems similar between 0 and 6. Getting this to converge takes a long time.

image
 

## TODOS

- [x] Inherit from the BDExchangeMove class (https://github.com/proteneer/timemachine/pull/1205/commits/476b6a7ef9d7361cab52985e96b23ecd42c1c653)
- [x] Resolve translations sketchiness (change reduced tests and seems like should be able to avoid two loops) (This bit: https://github.com/proteneer/timemachine/blob/e5c2e5d073e3d26b0df3b83ef8f226206943469f/timemachine/cpp/src/kernels/k_rotations.cu#L151-L177) (https://github.com/proteneer/timemachine/pull/1205/commits/c404955a3b250e1499d377598b38db85eb6cf6c5)
- [x] Put together acceptance rates/populations for Buckyball case (CPU vs GPU) - Seem to differ but the reference cases seem to be somewhere between 40% and 50% occupancy for 1 or 2 waters in the buckyball. 
- [x] Investigate making faster by avoiding the synchronization (Need to be able to sample/logsumexp without getting the size of arrays on the host) - Added to follow ups
- [x] Validate box size before making moves (if the radius is large enough, can be very difficult/impossible to do inner -> outer moves, which triggers an assert failure) (https://github.com/proteneer/timemachine/pull/1205/commits/ef7a3158bc66ed4c6dd4fb5d7a1f7202cd562014)
- [x] Run multiple replicas of the runs to get a sense of variance

## Follow Up Work

1. Improve performance, ~reduce number of sorts~


tests/test_cuda_bd_exchange_mover.py
```diff
@@ -461,7 +470,7 @@ def hif2a_rbfe_state() -> InitialState:
 
 @pytest.mark.parametrize(
     "steps_per_move,moves",
-    [(1, 500), (5000, 5000)],
```
badisa, line 464: Find it sketchy that this changed so much. Note that for float32 it only takes ~950 moves to move once, while for float64 it takes 5900~ moves. But the average acceptance in water is 1/7000 moves, so maybe this is to be expected?? Need to think some more about

proteneer: Over how many total samples? Do you get comparable acceptance rates in biased deletion only?

badisa: There are 6000 samples (by samples I assume you mean an attempted moved) and this is biased deletion in the context of ligand + complex. Currently only know the acceptance rate of biased deletion, still running the targeted simulations.

badisa: This test started failing after https://github.com/proteneer/timemachine/pull/1205/commits/476b6a7ef9d7361cab52985e96b23ecd42c1c653 and after investigating its simply due to the change in the ordering of using the rng (previously was quaternion then translation, now translation then quaternion). This test seems very flaky (and slow), however I do think it is important to have this test logic as otherwise we don't have a test case of what is our most common case (RBFE). And now that bias deletion and targeted insertion share code, it is important to validate both of these cases.

An additional thing we could do is triple the number of moves and move it to nightly.

proteneer: IIRC the average acceptance rate was something like 1/6000 to begin (sometimes it can happen faster, sometimes slower, but on average its 1/6000), so this doesn't seem too out of place.

timemachine/cpp/src/tibd_exchange_move.cu
```diff
@@ -0,0 +1,317 @@
...
+            this->d_log_weights_before_.data,
+            this->d_log_weights_after_.size(),
+            cudaMemcpyDeviceToDevice,
+            stream));
+
+        gpuErrchk(cudaMemsetAsync(d_inner_mols_count_.data, 0, d_inner_mols_count_.size(), stream));
+        gpuErrchk(cudaMemsetAsync(d_outer_mols_count_.data, 0, d_outer_mols_count_.size(), stream));
+
+        k_split_mols_inner_outer<RealType><<<mol_blocks, tpb, 0, stream>>>(
+            this->num_target_mols_,
```
proteneer, line 125: If this is a bottleneck, we can probably cache this in an optimization pass as well (just like log weights, the assignment of water into different groups only needs to happen when a move is accepted).

timemachine/cpp/src/kernels/k_rotations.cu
```diff
@@ -112,9 +113,9 @@ void __global__ k_rotate_and_translate_mols(
         int mol_end = mol_offsets[mol_sample + 1];
         int num_atoms = mol_end - mol_start;
 
-        RealType translation_x = box_x * translations[sample_idx * 3 + 0];
-        RealType translation_y = box_y * translations[sample_idx * 3 + 1];
-        RealType translation_z = box_z * translations[sample_idx * 3 + 2];
+        RealType translation_x = SCALE ? box_x * translations[sample_idx * 3 + 0] : translations[sample_idx * 3 + 0];
```
proteneer, line 116: not sure what this SCALE is doing?

timemachine/cpp/src/kernels/k_rotations.cu
```diff
@@ -85,7 +86,7 @@ void __global__ k_rotate_coordinates(
     rotated_coords[(coord_idx * n_rotations * 3) + (rotation_idx * 3) + 2] = local_coords[3];
 }
 
-template <typename RealType>
+template <typename RealType, bool SCALE>
```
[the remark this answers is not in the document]
proteneer: (Which I guess technically is really doing `k_rotate_and_recenter`)

timemachine/cpp/src/kernels/k_rotations.cu
```diff
@@ -134,10 +135,6 @@ void __global__ k_rotate_and_translate_mols(
 
             rotate_coordinates_by_quaternion(local_coords, quat);
```
proteneer, line 136: Is this the right behavior? I'm not sure if this is centering the molecule before applying the rotation?

timemachine/cpp/src/kernels/k_rotations.cu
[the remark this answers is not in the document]
badisa: Correct, it rotates then places the centroid at `t`

timemachine/cpp/src/kernels/k_rotations.cu
badisa: Will add a docstring, but it is simply whether to use the translation as is (such as generating inner or outer translations) or to scale the translation translations by the box dimensions (uniform unit vector in the case of biased deletion).

timemachine/cpp/src/kernels/k_rotations.cu
badisa: Per offline discussion it does seem to be the right behavior, but for clarity should be made to rotate about the origin then add the translation.

timemachine/cpp/src/tibd_exchange_move.cu
badisa: Right now the bottleneck is getting the number of waters in each region as well as whether or not you are targeting the center. 

To push everything to the GPU it seems like you would need to do the following:
  * Need a new kernel for computing argmax that doesn't use cub (requires a value for N) or set a min weight on all the 'empty values'
  * Set a dummy value for the sort
  * Handle the targeting mode in all of the individual kernels

timemachine/cpp/src/kernels/k_rotations.cu
badisa: Fixed up in https://github.com/proteneer/timemachine/pull/1205/commits/c404955a3b250e1499d377598b38db85eb6cf6c5

timemachine/cpp/src/kernels/k_rotations.cu
badisa: Added docstring in https://github.com/proteneer/timemachine/commit/c404955a3b250e1499d377598b38db85eb6cf6c5

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -0,0 +1,740 @@
...
+    const int num_target_mols,
+    const int mol_size,
+    const int *__restrict__ mol_atoms_idxs,
+    const int *__restrict__ mol_offsets,
+    const double *__restrict__ per_atom_energies,
+    const double inv_kT,
+    double *__restrict__ log_weights);
+
+template <typename RealType, int THREADS_PER_BLOCK>
+void __global__ k_set_sampled_weight(
```
proteneer, line 393: discussed offline re: ways to parallelize this over multiple blocks

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -0,0 +1,740 @@
...
+    double *__restrict__ log_weights);
+
+template <typename RealType>
+void __global__ k_compute_centroid_of_atoms(
+    const int num_atoms,
+    const int *__restrict__ atom_idxs, // [num_atoms]
+    const double *__restrict__ coords, // [N, 3]
+    RealType *__restrict__ centroid    // [3]
+) {
+    __shared__ unsigned long long fixed_centroid[3];
```
proteneer, line 460: why is centroid computed in floating point? (to avoid order warp ordering dependence within a single block?)

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -0,0 +1,740 @@
...
+        centroid_x -= target_x;
+        centroid_y -= target_y;
+        centroid_z -= target_z;
+
+        centroid_x -= box_x * nearbyint(centroid_x * inv_box_x);
+        centroid_y -= box_y * nearbyint(centroid_y * inv_box_y);
+        centroid_z -= box_z * nearbyint(centroid_z * inv_box_z);
+
+        RealType dist = (centroid_x * centroid_x) + (centroid_y * centroid_y) + (centroid_z * centroid_z);
+        // This is not deterministic in the ordering....
```
proteneer, line 557: Does this mean that the generated ordering of mols differ for the inner/outer mols? Does the energy computation differ if the ordering of the mols different?

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -0,0 +1,740 @@
...
+    // TBD investigate shared memory for speed
+    RealType before_log_prob = convert_nan_to_inf<RealType>(compute_logsumexp_final<RealType>(before_log_sum_exp));
+    RealType after_log_prob = convert_nan_to_inf<RealType>(compute_logsumexp_final<RealType>(after_log_sum_exp));
+
+    RealType log_acceptance_prob = min(before_log_prob - after_log_prob + log_vol_prob, static_cast<RealType>(0.0));
+    const bool accepted = rand[0] < exp(log_acceptance_prob);
+    if (!accepted) {
+        return;
+    }
+    __syncthreads();
```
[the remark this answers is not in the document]
badisa: Yup, these kernels are only called with a single block

timemachine/cpp/src/kernels/k_exchange.cu
badisa: The ordering of the `inner_mols` and `outer_mols` is not deterministic due to the atomic adds to the counts (could do a cub IF computation like in the local MD code, but seemed more expensive). It doesn't change the energies but it does make the sampling non-deterministic so you can get different mols sampled.

timemachine/cpp/src/kernels/k_exchange.cu
badisa: Do you mean why is it not computed in floating point? Its computed in long long to allow atomic adds with deterministic results.

Though did find a bug because of this comment, should only allow a single block.

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -0,0 +1,740 @@
...
+    const RealType log_vol_prob =
+        targetting_inner == 1 ? log(inner_volume) - log(outer_vol) : log(outer_vol) - log(inner_volume);
+
+    // All kernels compute the same acceptance
+    // TBD investigate shared memory for speed
+    RealType before_log_prob = convert_nan_to_inf<RealType>(compute_logsumexp_final<RealType>(before_log_sum_exp));
+    RealType after_log_prob = convert_nan_to_inf<RealType>(compute_logsumexp_final<RealType>(after_log_sum_exp));
+
+    RealType log_acceptance_prob = min(before_log_prob - after_log_prob + log_vol_prob, static_cast<RealType>(0.0));
+    // if (atom_idx == 0) {
```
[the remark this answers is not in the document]
badisa: Nope, had thought I had removed this.

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -0,0 +1,740 @@
...
+    const RealType *__restrict__ rand,               // [1]
+    const RealType *__restrict__ before_log_sum_exp, // [2]
+    const RealType *__restrict__ after_log_sum_exp,  // [2]
+    const double *__restrict__ moved_coords,         // [N, 3]
+    double *__restrict__ dest_coords,                // [N, 3]
+    size_t *__restrict__ num_accepted                // [1]
+) {
+    int atom_idx = blockIdx.x * blockDim.x + threadIdx.x;
+
+    int targetting_inner = targeting_inner_volume[0];
```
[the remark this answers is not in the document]
badisa: Fixed the typo in https://github.com/proteneer/timemachine/pull/1205/commits/b1a29bb2bf119de56627bfa4d6df0a592262e68c

Would you prefer a different name? This was the clearest name I could think of (earlier names were confusing to me)

timemachine/cpp/src/kernels/k_exchange.cu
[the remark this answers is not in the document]
badisa: Removed as part of https://github.com/proteneer/timemachine/pull/1205/commits/b932ccb2f940813a5364ff8303d658b32d0893ae

timemachine/cpp/src/kernels/k_exchange.cu
proteneer: edit: yeah

proteneer: (explanation makes sense)

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -0,0 +1,740 @@
...
+#include "k_exchange.cuh"
+#include "k_fixed_point.cuh"
+#include "k_logsumexp.cuh"
+#include <assert.h>
+
+namespace timemachine {
+
+// When we are considering exchange we want to treat Nan probabilities as inf
+// Allows us to go from a clashy state to a non-clashy state. And no nan poisoning
+template <typename RealType> RealType __host__ __device__ convert_nan_to_inf(const RealType input) {
```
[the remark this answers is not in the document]
badisa: Added in https://github.com/proteneer/timemachine/pull/1205/commits/c9dd424efd845918a028d759c95ce1c08f978f08

timemachine/cpp/src/kernels/k_exchange.cu
proteneer: (resolved offline)

timemachine/cpp/src/kernels/k_exchange.cu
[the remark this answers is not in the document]
proteneer: (assert gridDim.x == 1) then?

timemachine/cpp/src/tibd_exchange_move.cu
```diff
@@ -0,0 +1,316 @@
...
+    curandErrchk(curandSetStream(this->cr_rng_, stream));
+
+    this->compute_initial_weights(N, d_coords, d_box, stream);
+
+    const int tpb = DEFAULT_THREADS_PER_BLOCK;
+    const int mol_blocks = ceil_divide(this->num_target_mols_, tpb);
+
+    dim3 atom_by_atom_grid(ceil_divide(N, tpb), this->mol_size_, 1);
+
+    k_compute_centroid_of_atoms<RealType>
```
proteneer, line 88: i'm guessing we're guaranteeing to never move the ligand during the sequence of moves?

badisa: Correct

timemachine/cpp/src/kernels/k_exchange.cu
badisa: Switched to partitioning in https://github.com/proteneer/timemachine/pull/1205/commits/4f2e9fc8f803866f44be56e208b400605bb52240 at @proteneer recommendation. Speeds things up by 40%

timemachine/cpp/src/tibd_exchange_move.cu
```diff
@@ -0,0 +1,295 @@
...
+        // The this->d_acceptance_ buffer contains the random value for determining where to insert and whether to accept the move
+        curandErrchk(templateCurandUniform(this->cr_rng_, this->d_acceptance_.data, this->d_acceptance_.length));
+
+        k_decide_targeted_move<<<1, 1, 0, stream>>>(
+            this->num_target_mols_, this->d_acceptance_.data, d_inner_mols_count_.data, d_targeting_inner_vol_.data);
+        gpuErrchk(cudaPeekAtLastError());
+
+        // Copy count and flag to the host, needed to know how many values to look at for
+        // sampling and logsumexp
+        gpuErrchk(cudaMemcpyAsync(
```
proteneer, line 150: note for later: optimize `logsumexp` to support segments via `SegmentMax`, so we can remove the pinned memcpys and the sync entirely.

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -0,0 +1,689 @@
...
+
+template <typename RealType>
+void __global__ k_setup_destination_weights_for_targeted(
+    const int num_target_mols,
+    const int *__restrict__ samples,                // [1]
+    const int *__restrict__ targeting_inner_volume, // [1]
+    const int *__restrict__ inner_count,            // [1]
+    const int *__restrict__ partitioned_indices,    // [inner_count]
+    const RealType *__restrict__ weights,           // [num_target_mols]
+    RealType *__restrict__ output_weights) {
```
proteneer, line 641: document size of `output_weights` as `[num_target_mols]` but maybe note that we only use up to size `count + 1`

proteneer (approved):
lgtm - just some minor documentation notes. Nice work!

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -0,0 +1,689 @@
...
+    int idx = blockIdx.x * blockDim.x + threadIdx.x;
+    const int target_inner = targeting_inner_volume[0];
+    const int local_inner_count = inner_count[0];
+    const int outer_count = num_target_mols - local_inner_count;
+    const int count = target_inner == 1 ? local_inner_count : outer_count;
+    const int offset = target_inner == 1 ? 0 : local_inner_count;
+    if (idx == 0) {
+        int sample_idx = samples[idx];
+        output_weights[count + idx] = weights[sample_idx];
```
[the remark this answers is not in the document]
badisa: Added in https://github.com/proteneer/timemachine/pull/1205/commits/6c7f11565140ec23d0bbda7731e9a63d4ba0b041

timemachine/cpp/src/kernels/k_exchange.cu
badisa: Added in https://github.com/proteneer/timemachine/pull/1205/commits/6c7f11565140ec23d0bbda7731e9a63d4ba0b041

timemachine/cpp/src/tibd_exchange_move.cu
```diff
@@ -0,0 +1,295 @@
...
+
+        // Add one to the destination count, as we just moved a mol there
+        this->logsumexp_.sum_device(dest_count + 1, d_dest_weights_.data, this->d_log_sum_exp_after_.data, stream);
+
+        k_attempt_exchange_move_targeted<RealType><<<ceil_divide(N, tpb), tpb, 0, stream>>>(
+            N,
+            d_targeting_inner_vol_.data,
+            d_box_volume_.data,
+            inner_volume_,
+            this->d_acceptance_.data + 1, // Offset to get the last value for the acceptance criteria
```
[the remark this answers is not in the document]
badisa: Renamed in https://github.com/proteneer/timemachine/pull/1205/commits/6c7f11565140ec23d0bbda7731e9a63d4ba0b041

timemachine/cpp/src/tibd_exchange_move.cu
```diff
@@ -0,0 +1,295 @@
...
+        this->sampler_.sample_device(src_count, 1, d_src_weights_.data, this->d_samples_.data, stream);
+
+        // Selected an index from the src weights, need to remap the samples idx to the mol indices
+        k_adjust_sample_idx<<<1, tpb, 0, stream>>>(
+            d_targeting_inner_vol_.data, d_inner_mols_count_.data, d_partitioned_indices_.data, this->d_samples_.data);
+        gpuErrchk(cudaPeekAtLastError());
+
+        // Don't move translations into computation of the incremental, as different translations can be used
+        // by different bias deletion movers (such as targeted insertion)
+        // Don't scale the translations as they computed to be within the region
```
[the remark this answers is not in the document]
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1205/commits/6c7f11565140ec23d0bbda7731e9a63d4ba0b041

timemachine/cpp/src/tibd_exchange_move.cu
```diff
@@ -0,0 +1,295 @@
...
+        k_separate_weights_for_targeted<RealType><<<mol_blocks, tpb, 0, stream>>>(
+            this->num_target_mols_,
+            d_targeting_inner_vol_.data,
+            d_inner_mols_count_.data,
+            d_partitioned_indices_.data,
+            this->d_log_weights_before_.data,
+            d_src_weights_.data);
+        gpuErrchk(cudaPeekAtLastError());
+
+        // targeting_inner_vol == 1 indicates that we are target the inner volume, starting from the outer mols
```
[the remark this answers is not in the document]
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1205/commits/6c7f11565140ec23d0bbda7731e9a63d4ba0b041

badisa:
Just waiting on runs replicas of the mc script.

PR #1208, opened 2023-12-04
Expand API of exchange movers to support RBFE

badisa:
* Not all movers will take params like this, such as the barostat
* Required to be able to support RBFE (https://github.com/proteneer/timemachine/blob/d8dfa73bdbf845a0612dbc23f91c0f416c0848c2/timemachine/fe/free_energy.py#L938)


timemachine/cpp/src/nonbonded_common.hpp
```diff
@@ -8,6 +8,9 @@
 
 namespace timemachine {
 
+// each atom parameterized by a 4-tuple: charge, lj sigma, lj epsilon, 4D coordinate w
+enum { PARAM_OFFSET_CHARGE = 0, PARAM_OFFSET_SIG, PARAM_OFFSET_EPS, PARAM_OFFSET_W, PARAMS_PER_ATOM };
```
badisa, line 12: Moved this enum to a hpp file so we can access the values from the wrapped kernels. Seem like these values have utility beyond just the kernels.

proteneer:
lgtm - i'm guessing the actual RBFE changes will come in a second PR?

proteneer (approved):

badisa:
> lgtm - i'm guessing the actual RBFE changes will come in a second PR? (presumably the python code will be doing an `isinstance(mover, BDExchangeMover)` check on the movers to set the parameters? since the base mover class does not define `set/get_params`)

Exactly. Want to do some more validation of the water sampling before integrating into RBFE, and these APIs will be required regardless of whether we have to change the code.

mcwitt (approved):

PR #1213, opened 2023-12-05
Makes exchange move tests less flaky

badisa:
* Moves them to nightly as well as they are slow


mcwitt (approved):
Nice! Tested this locally as well. Runtime excluding nightly tests is down from ~15 minutes to ~1 minute on my machine.

PR #1214, opened 2023-12-07
Fixes syntax typo in centroid kernel

badisa:
* Results in crashes with some combinations of nvcc (version V11.7.64, Build cuda_11.7.r11.7/compiler.31294372_0
), gcc (7.5.0), cuda arch (7.5) and nvidia driver (515.43.04). Unclear what exactly triggers this, but the code is straight up wrong.
* Not sure how this got in there, looks to have just been a typo that compiled
* This test fails https://github.com/proteneer/timemachine/blob/master/tests/test_cuda_targeted_insertion_mover.py#L47 on the tool combination stated above without the fix, though not on CI, or my personal machine.

### DIFF
--- a/tests/test_cuda_targeted_insertion_mover.py
+++ b/tests/test_cuda_targeted_insertion_mover.py
@@ -41,6 +41,7 @@ def compute_ref_log_prob(ref_exchange, water_idx, vi_mols, vj_mols, vol_i, vol_j
     return trial_coords, log_p_accept
 
 
+@pytest.mark.memcheck
 @pytest.mark.parametrize("seed", [2023, 2024])
 @pytest.mark.parametrize("radius", [0.1, 0.5, 1.2, 2.0])
 @pytest.mark.parametrize("precision", [np.float64, np.float32])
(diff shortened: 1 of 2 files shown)


mcwitt (approved):
Ah, so this was writing out of bounds when `threadIdx.x >= 3`? Fix looks correct to me.

badisa:
> Ah, so this was writing out of bounds when `threadIdx.x >= 3`? Fix looks correct to me.

Yup. It appears that under the different tool chain it would be written out of bounds, but under our CI tool chain it behaves as intended. Surprised that this ever worked, seems wrong as it was.

proteneer (commented):
The old expression should've result in an illegal memory access (since threads with `idx >=3` will be writing to `fixed_centroid`, which only has size 3) - how does cuda-memcheck not catch this?

badisa:
> The old expression should've result in an illegal memory access (since threads with `idx >=3` will be writing to `fixed_centroid`, which only has size 3, in an out of bounds manner) - how does cuda-memcheck not catch this?

No clue why cuda-memcheck isn't catching this... Probably a bug, will review the cuda release notes to see if this was broken in a release. The does does need `memcheck` as a mark, but even with that doesn't fail locally.

proteneer (approved):

PR #1215, opened 2023-12-07
Adds support for CUDA_VISIBLE_DEVICES in cuda pool client

badisa:
Previously if you were on a multi-gpu machine and set `CUDA_VISIBLE_DEVICES` it would use [0, ... max_workers-1] GPUs rather than the idxs defined in `CUDA_VISIBLE_DEVICES`. The test that passed before was simply because it was patched in the subprocess.

Verified this on a 10 GPU instance and been able to target 1 - 3 GPUs with `CUDA_VISIBLE_DEVICES` after these changes.

### DIFF
--- a/timemachine/parallel/client.py
+++ b/timemachine/parallel/client.py
@@ -229,7 +229,7 @@ class CUDAPoolClient(ProcessPoolClient):
         assert (
             len(self._gpu_list) >= self.max_workers
         ), "Fewer available GPUs than max workers expects, check CUDA_VISIBLE_DEVICES"
-        assert len(self._gpu_list) <= gpus, "More GPUs available than the machine has, check CUDA_VISIBLE_DEVICES"
+        assert len(self._gpu_list) <= gpus, "More GPUs requested than the machine has, check CUDA_VISIBLE_DEVICES"
 
 
 class BinaryFutureWrapper:
--- a/timemachine/parallel/client.py
+++ b/timemachine/parallel/client.py
@@ -199,7 +199,7 @@ class CUDAPoolClient(ProcessPoolClient):
         super().__init__(max_workers)
         visible_devices = os.environ.get("CUDA_VISIBLE_DEVICES")
         if visible_devices:
-            self._gpu_list = [i for i in map(int, visible_devices.split(","))]
+            self._gpu_list = [int(i) for i in visible_devices.split(",")]
         else:
             self._gpu_list = list(range(max_workers))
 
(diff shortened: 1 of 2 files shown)


timemachine/parallel/client.py
```diff
@@ -197,20 +197,24 @@ class CUDAPoolClient(ProcessPoolClient):
...
         super().__init__(max_workers)
+        visible_devices = os.environ.get("CUDA_VISIBLE_DEVICES")
+        if visible_devices:
+            self._gpu_list = [i for i in map(int, visible_devices.split(","))]
+        else:
+            self._gpu_list = list(range(max_workers))
 
     @staticmethod
     def wrapper(max_workers, idx, fn, *args, **kwargs):
-        # for a single worker, do not overwrite CUDA_VISIBLE_DEVICES
```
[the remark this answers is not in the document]
jkausrelay: Not sure either. Maybe this was a work around to the issue here, i.e. by setting CUDA_VISIBLE_DEVICES manually?

jkausrelay (approved):

timemachine/parallel/client.py
```diff
@@ -197,20 +197,24 @@ class CUDAPoolClient(ProcessPoolClient):
 
     def __init__(self, max_workers):
         super().__init__(max_workers)
+        visible_devices = os.environ.get("CUDA_VISIBLE_DEVICES")
+        if visible_devices:
+            self._gpu_list = [i for i in map(int, visible_devices.split(","))]
```
mcwitt, line 202: nit:
```suggestion
            self._gpu_list = [int(i) for i in visible_devices.split(",")]
```

timemachine/parallel/client.py
```diff
@@ -222,6 +226,10 @@ def verify(self):
         """
         gpus = get_gpu_count()
         assert self.max_workers <= gpus, f"More workers '{self.max_workers}' requested than GPUs '{gpus}'"
+        assert (
+            len(self._gpu_list) >= self.max_workers
+        ), "Fewer available GPUs than max workers expects, check CUDA_VISIBLE_DEVICES"
```
mcwitt, line 231: I might be confused, but wouldn't this occur when `CUDA_VISIBLE_DEVICES` lists more GPUs than the `max_workers` given to the constructor? (Doesn't seem related to available GPUs?)

timemachine/parallel/client.py
```diff
@@ -222,6 +226,10 @@ def verify(self):
...
+        assert len(self._gpu_list) <= gpus, "More GPUs available than the machine has, check CUDA_VISIBLE_DEVICES"
```
mcwitt, line 232: I'm a bit confused by this; should it instead be
```suggestion
        assert len(self._gpu_list) <= gpus, "More GPUs requested than the machine has, check CUDA_VISIBLE_DEVICES"
```
?

mcwitt (approved):

PR #1216, opened 2023-12-08
Switches default water ff to Amber14 Tip3p

badisa:
* Identical water parameters with the addition of ion parameters


badisa:
Not sure this is desirable, but does make it easier to support ions within systems and produces the same water parameters under the old and new version of tip3p.

proteneer (approved):
lgtm!

jkausrelay (approved):

PR #1217, opened 2023-12-08
Add utility for extracting replica trajectories from HREX results

mcwitt:
Adds a simple utility function for converting the output of `timemachine.md.hrex.run_hrex` into a form suitable for extracting replica trajectories.

Note: this does not yet make it easy to extract replica trajectories from (non-vacuum) RBFE simulations, since the frames from each HREX iteration are concatenated into a single long trajectory for each state in the output. Solving this cleanly in general might require modifying the RBFE output schema to additionally group frames by HREX iteration.

### DIFF
--- a/timemachine/md/hrex.py
+++ b/timemachine/md/hrex.py
@@ -168,9 +168,8 @@ def get_samples_by_iter_by_replica(
         return [i for i, _ in sorted(enumerate(p), key=lambda t: t[1])]
 
     samples_by_replica_by_iter = [
-        [samples_by_state[state_idx] for state_idx in state_idx_by_replica]
+        [samples_by_state[state_idx] for state_idx in inverse_permutation(replica_idx_by_state)]
         for samples_by_state, replica_idx_by_state in zip(samples_by_state_by_iter, replica_idx_by_state_by_iter)
-        for state_idx_by_replica in [inverse_permutation(replica_idx_by_state)]
     ]
 
     # transpose
--- a/timemachine/md/hrex.py
+++ b/timemachine/md/hrex.py
@@ -166,11 +166,8 @@ def get_samples_by_iter_by_replica(
 
     assert len(samples_by_state_by_iter) == len(replica_idx_by_state_by_iter)
 
-    def inverse_permutation(p):
-        return [i for i, _ in sorted(enumerate(p), key=lambda t: t[1])]
-
     samples_by_replica_by_iter = [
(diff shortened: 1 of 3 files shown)


proteneer:
awesome! (you definitely figured out how to do this faster than me)

badisa (approved):
Nice, this is a super handy utility

timemachine/md/hrex.py
```diff
@@ -152,6 +152,31 @@ def symmetrize(x):
...
+
+    def inverse_permutation(p):
+        return [i for i, _ in sorted(enumerate(p), key=lambda t: t[1])]
+
+    samples_by_replica_by_iter = [
+        [samples_by_state[state_idx] for state_idx in inverse_permutation(replica_idx_by_state)]
+        for samples_by_state, replica_idx_by_state in zip(samples_by_state_by_iter, replica_idx_by_state_by_iter)
+    ]
+
+    samples_by_iter_by_replica = [list(xs) for xs in zip(*samples_by_replica_by_iter)]  # transpose
```
badisa, line 175: nit: verify that the length of all lists in `samples_by_replica_by_iter` are the same length to avoid zip using the shortest length list and dropping data. Perhaps also for the samples list and the replicas list as they also get zipped

timemachine/md/hrex.py
```diff
@@ -152,6 +152,31 @@ def symmetrize(x):
...
+
+        (hrex_iter, state_idx) -> samples
+
+    to a more convenient form for analyzing the trajectories of individual replicas
+
+        (replica_idx, hrex_iter) -> samples
+    """
+
+    def inverse_permutation(p):
+        return [i for i, _ in sorted(enumerate(p), key=lambda t: t[1])]
```
maxentile, line 168: (alternatively, `np.argsort(p)`)

maxentile (approved):
Nice! This view should ease analysis of replica mixing.

timemachine/md/hrex.py
mcwitt: Good call, added assertions in https://github.com/proteneer/timemachine/pull/1217/commits/0be0d574f8233c935539c1fd6530c4c82a3caa4e and https://github.com/proteneer/timemachine/pull/1217/commits/53747236adc86a7382461bc0a91a717866741bf7

timemachine/md/hrex.py
mcwitt: Changed in https://github.com/proteneer/timemachine/pull/1217/commits/add35fb05fab9e903867420d3d956d2041328530. Not sure why I had been avoiding numpy here.

PR #1218, opened 2023-12-12
Fix Water Sampling Bug

badisa:
"Beware the undefined behavior" - @proteneer 

- When I had originally written `k_compute_nonbonded_target_atom_energies` I had somehow convinced myself that `__syncthreads()` handled exited threads. It **DOES NOT**, per https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#synchronization-functions
> __syncthreads() is allowed in conditional code but only if the conditional evaluates identically across the entire thread block, otherwise the code execution is likely to hang or produce unintended side effects.
- This issue only showed up for larger system sizes

Nightly pipeline: https://gitlab.com/relaytx/physics-platform/timemachine/-/pipelines/1106085189
Failed tests without the fix: https://github.com/proteneer/timemachine/pull/1218/commits/3ec9b712671f78709f4b8a8a3d8c374ff8034a10

- [x] Reduce memory usage by test_targeted_moves_in_bulk_water that caused failure in nightly tests

### DIFF
--- a/timemachine/cpp/src/kernels/k_nonbonded.cuh
+++ b/timemachine/cpp/src/kernels/k_nonbonded.cuh
@@ -481,13 +481,12 @@ void __global__ k_compute_nonbonded_target_atom_energies(
         RealType ci_z = coords[atom_i_idx * 3 + 2];
 
         int atom_j_idx = blockIdx.x * blockDim.x + threadIdx.x;
-        // All threads in the threadblock must loop to allow for __syncthreads() and the row accumulation.
-        while (atom_j_idx - threadIdx.x < N) {
-            // Zero out the energy buffer
-            block_energy_buffer[threadIdx.x] = 0;
+        // Zero out the energy buffer
+        block_energy_buffer[threadIdx.x] = 0;
+        while (atom_j_idx < N) {
             // The two atoms are in the same molecule, don't compute the energies
             // requires that the atom indices in each target mol is consecutive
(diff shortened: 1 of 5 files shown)


proteneer (approved):
lgtm

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -481,12 +481,13 @@ void __global__ k_compute_nonbonded_target_atom_energies(
         RealType ci_z = coords[atom_i_idx * 3 + 2];
 
         int atom_j_idx = blockIdx.x * blockDim.x + threadIdx.x;
-        // Zero out the energy buffer
-        block_energy_buffer[threadIdx.x] = 0;
-        while (atom_j_idx < N) {
+        // All threads in the threadblock must loop to allow for __syncthreads() and the row accumulation.
+        while (atom_j_idx - threadIdx.x < N) {
```
mcwitt, line 485: nit: slightly easier to comprehend for me as
```suggestion
        while (blockIdx.x * blockDim.x < N) {
```

mcwitt (approved):
good catch and fix!

timemachine/cpp/src/kernels/k_nonbonded.cuh
badisa: This would result in an infinite loop since `blockIdx.x * blockDim.x` is constant for each kernel

mcwitt: Ah, right, sorry. I missed that `atom_j_idx` is being modified in the loop.

PR #1220, opened 2023-12-14
Add set_step to ensure all MC movers treat each frame the same way

badisa:
* Not very important for the barostat, but is an issue for things like water sampling

In the future we will likely want to adopt a different API that makes it easier to be explicit about the moves we want to execute without suffering the cost of going between Cuda/Python


timemachine/fe/free_energy.py
```diff
@@ -956,6 +959,7 @@ def get_equilibrated_xvb(xvb: CoordsVelBox, params: NDArray) -> CoordsVelBox:
     fraction_accepted_by_pair_by_iter: List[List[Tuple[int, int]]] = []
 
     for iteration, n_frames_iter in enumerate(batches(md_params.n_frames, n_frames_per_iter), 1):
+        current_step = (iteration - 1) * n_frames_per_iter * md_params.steps_per_frame
```
mcwitt, line 962: nit: consider moving definition closer to where it is used below

timemachine/fe/free_energy.py
```diff
@@ -929,6 +929,9 @@ def log_q_fn(replica_idx: ReplicaIdx, state_idx: StateIdx) -> float:
     def get_equilibrated_xvb(xvb: CoordsVelBox, params: NDArray) -> CoordsVelBox:
         if md_params.n_eq_steps == 0:
             return xvb
+        # Set the movers to 0 to ensure they all equilibrate the same way
```
mcwitt, line 932: suggestion for clarity:
```suggestion
        # Ensure initial mover state is consistent across replicas
```

timemachine/fe/free_energy.py
badisa: Moved in https://github.com/proteneer/timemachine/pull/1220/commits/c49dbe14749ca5d036d478a9337eec46c1cbc415

mcwitt (approved):
As discussed offline, this seems like a reasonable compromise until we can come up with a cleaner API, via e.g. duplicating relevant context for each replica

proteneer (approved):
yeah, we need to make this better at some point. But this is a reasonable patch for now.

PR #1222, opened 2023-12-15
Avoid running kernel repeatedly to get same value

badisa:
After attempting to integrate water sampling into RBFE it became clear that water sampling is way too slow to be used at the moment. There need to be some major optimizations to the kernels that are going to be more involved. These were some minor things I noticed while trying to find low hanging fruit.

This has about a 0.5% - 1% difference depending on system size since it was called within the inner loop.


timemachine/cpp/src/tibd_exchange_move.cu
```diff
@@ -208,7 +205,7 @@ void TIBDExchangeMove<RealType>::move(
         this->sampler_.sample_device(src_count, 1, d_src_weights_.data, this->d_samples_.data, stream);
 
         // Selected an index from the src weights, need to remap the samples idx to the mol indices
-        k_adjust_sample_idx<<<1, tpb, 0, stream>>>(
+        k_adjust_sample_idx<<<1, 1, 0, stream>>>(
```
proteneer, line 208: 1 thread per block?

badisa: This kernel just modifies one value. Doesn't need to run more than one thread.

Will modify the kernel to make this clearer, as else it risk being a race condition.

badisa: Modified kernel to defend against this, though running one thread per block and only one block is still the desirable thing to do.

proteneer (approved):

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -683,10 +683,15 @@ void __global__ k_adjust_sample_idx(
     const int *__restrict__ partitioned_indices,    // [inner_count]
     int *__restrict__ sample_idxs                   // [1]
 ) {
+    int idx = blockIdx.x * blockDim.x + threadIdx.x;
+    // At the moment we only have one sample
```
mcwitt, line 687: nit: seems like this comment might make more sense (and be less likely to go stale) where the kernel is launched with one thread per block?

mcwitt: edit: sorry, I misread this initially. I think this makes sense given the `if (idx > 0)` return condition below.

timemachine/cpp/src/tibd_exchange_move.cu
mcwitt: nit/question: is there a reason to add `if (idx > 0) { return; }` to the kernel implementation in addition to this change? Seems like this could lead to confusion if we later increase the number of samples. (I didn't understand the comment about a possible race condition here)

mcwitt (approved):

timemachine/cpp/src/tibd_exchange_move.cu
badisa: The old version simply would have multiple threads read from `sample_idxs[0]` and then use that value to override the same value. This is a potential race condition if one of the threads reads the value after it has been updated and updates it again. 

As to the confusion about supporting more number of samples, I think that confusion is interspersed through the code already. Some parts of the code early on made it seem like multiple samples was an easy introduction, but as I have gone along it has become clear that it wouldn't be straightforward. At the moment it seems best to leave to resolving this when we support multiple samples.

PR #1224, opened 2023-12-18
Format colorbar labels consistently with annotations in heatmaps

mcwitt:
Minor fix to HREX heatmap plots to format colorbar tick labels consistently with the annotations that are shown on the plot when the number of states is less than 20. Currently, the former is shown as a fraction between 0 and 1, while the latter is shown as a percentage.

Before:
image
After:
image

### DIFF
--- a/timemachine/fe/plots.py
+++ b/timemachine/fe/plots.py
@@ -312,26 +312,26 @@ def plot_fwd_reverse_predictions(
 
 
 def plot_hrex_transition_matrix(
-    transition_rate: NDArray,
+    transition_probability: NDArray,
     figsize: Tuple[float, float] = (13, 10),
     annotate_threshold: int = DEFAULT_HEATMAP_ANNOTATE_THRESHOLD,
     format_annotation: Callable[[float], str] = lambda x: f"{100.0*x:.2g}",
     format_cbar_tick: Callable[[float], str] = lambda x: f"{100.0*x:.2g}%",
 ):
-    """Plot matrix of estimated transition rates for permutation moves as a heatmap."""
-    n_states, _ = transition_rate.shape
+    """Plot matrix of estimated transition probabilities for permutation moves as a heatmap."""
+    n_states, _ = transition_probability.shape
     states = np.arange(n_states)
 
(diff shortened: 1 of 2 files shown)


timemachine/fe/plots.py
```diff
@@ -315,7 +315,8 @@ def plot_hrex_transition_matrix(
     transition_rate: NDArray,
     figsize: Tuple[float, float] = (13, 10),
     annotate_threshold: int = DEFAULT_HEATMAP_ANNOTATE_THRESHOLD,
-    format_rate: Callable[[float], str] = lambda x: f"{100.0*x:.1f}",
+    format_rate: Callable[[float], str] = lambda x: f"{100.0*x:.2g}",
+    cbar_tick_label_suffix: str = "%",
```
maxentile, line 319: Unsure, but might be slightly clearer to control these independently?

(Rather than expressing cbar tick label as heatmap annotation + suffix.)

```suggestion
    format_heatmap_cell: Callable[[float], str] = lambda x: f"{100.0*x:.1f}",
    format_cbar_tick: Callable[[float], str] = lambda x: f"{100.0*x:.2g}%",
```

timemachine/fe/plots.py
```diff
@@ -315,7 +315,8 @@ def plot_hrex_transition_matrix(
     transition_rate: NDArray,
```
maxentile, line 315: (not introduced in this PR, but I just noticed a small naming imprecision here: these are transition _probabilities_, not transition _rates_, correct? (if these were rates, the diagonal would be negative))

maxentile (approved):
Nice -- this improves clarity!

timemachine/fe/plots.py
mcwitt: Nice, this seems cleaner to me too.

timemachine/fe/plots.py
mcwitt: Good catch. Renamed in https://github.com/proteneer/timemachine/pull/1224/commits/60418a4b572efeb08c5d7d46ec4e934dee5fc77e

PR #1227, opened 2024-01-02
Add atom mapping filter disallowing flipped torsions in core

mcwitt:
Adds an atom mapping filter removing mappings where a torsion is "flipped" (i.e. the sign of `torsion_volume` changes) by the mapping.

#### TODO
- [ ] validate atom mapping benchmark
- [ ] add test for filter


mcwitt:
Superseded by #1229

PR #1228, opened 2024-01-04
Bugfix for water sampling at the boundary

proteneer:
@badisa 

Here's the patch that deals with the corner cases when region i has either all the waters or exactly 1 water, the case of 0 waters is ignored.

Transition diagram:
```
0 water -100%-> 1 water -50%-> ... -50%-> N-1 water -50%-> N water
        <-50%-          <-50%- ... <-50%-           <-100%-
```

Namely, the patch addresses the fwd/rev proposal probabilities. Code below,  `n1` and `n2` are number of waters in each region, `n1+n2=N`:

``` python
if n1 == 0 and n2 == 0:
    assert 0
elif n1 > 0 and n2 == 0:
    return self.swap_vi_into_vj(inner_mols, outer_mols, x, v2_site, vol_1, vol_2)
elif n1 == 0 and n2 > 0:
    return self.swap_vi_into_vj(outer_mols, inner_mols, x, v1_site, vol_2, vol_1)
elif n1 > 0 and n2 > 0:
    if np.random.rand() < 0.5:
        return self.swap_vi_into_vj(inner_mols, outer_mols, x, v2_site, vol_1, vol_2)
    else:
        return self.swap_vi_into_vj(outer_mols, inner_mols, x, v1_site, vol_2, vol_1)
else:
    # negative numbers, dun goof'd
    assert 0
```

# Rerunning Buckyball case

In the buckyball case we run into this edge case and need to handle it correctly.

### Zero Waters

Total Moves 509140000
Num Waters in Buckyball: 1, 59.19%
Num Waters in Buckyball: 2, 40.81%

###  Six Waters

Total Moves 444100000
Num Waters in Buckyball: 4, 0.00%
Num Waters in Buckyball: 2, 50.71%
Num Waters in Buckyball: 1, 49.28%

### Zero Waters
Total Moves 100800000
Num Waters in Buckyball: 1, 57.81%
Num Waters in Buckyball: 2, 42.19%

### Six Waters
Total Moves 100000000
Num Waters in Buckyball: 1, 57.58%
Num Waters in Buckyball: 2, 42.42%

image


timemachine/md/exchange/exchange_mover.py
```diff
@@ -355,8 +355,29 @@ def swap_vi_into_vj(
...
+        # directionality is going from region i into region j
+        # take care of asymmetric boundary conditions
+        if len(vi_mols) == self.num_waters:
+            # region i has every water, so fwd probability is 100%, rev probability is 50%
+            g_fwd = 1.0
+            g_rev = 0.5
+        elif len(vi_mols) == 1:
+            # region i has only one water, so fwd probability is 50%, rev probability is 100%
+            g_fwd = 0.5
+            g_rev = 1.0
```
badisa, line 367: Is there any nuance of this 100%? For example lets say you have a buckyball as the inner volume that has a methane (no waters) inside and there is no space for the water, in that case the probability wouldn't be 100% to accept that move. Or if the radius of the inner sphere contained the carbons you could have a clash with the new water.

proteneer: The probabilities above indicate the *proposal* probability, not the *acceptance* probability. The clashiness is taken care of by pi(x), i.e. the probability assigned to the stationary distribution.

proteneer:
(updated description to be specific about the issue addressed)

timemachine/md/exchange/exchange_mover.py
```diff
@@ -355,8 +355,29 @@ def swap_vi_into_vj(
         log_weights_after_full = np.array(log_weights_after_full)
         log_weights_after = log_weights_after_full[vj_plus_one_idxs]
 
+        # directionality is going from region i into region j
+        # take care of asymmetric boundary conditions
+        if len(vi_mols) == self.num_waters:
```
maxentile, line 360: Sorry, not immediately actionable, but in a later pass I'd like to figure out how to write this in terms of a more direct comparison `len(vi_mols) == 0` or similar.

---------

Want the code to explicitly reflect that we first make a discrete random choice of whether we're doing a "delete from i" move or a "delete from j" move, where the distribution over choices is conditioned on x.

f_i(x) = is_there_at_least_one_water_in_i(x)
f_j(x) = is_there_at_least_one_water_in_j(x)

f(x) = (f_i(x), f_j(x))
move_probs(x) = f(x)  / sum(f(x))

i.e., we sample move_idx ~ Categorical(move_probs(x0)),

then we produce x_prop ~ moves[move_idx](. | x0)

... following the recipe in https://www.pnas.org/doi/full/10.1073/pnas.1106094108 (where the random choice of "protocol" can depend on x).

The current implementation is a special case where the distribution over choices is uniform over all currently feasible choices. (i.e. uniform over whichever regions contain at least one water)

Should be equivalent to the nested if statement definition, but might reduce the number of special cases that need to be listed, and might be more easily extended to the case of multiple regions, non-uniform move selection probabilities, or other directions...

maxentile (approved):
Good catch!

Not necessarily in this PR, it would be good for one of us to add a test that's sensitive enough to fail with the current code but pass with the fixed code.

(Don't know, maybe an ideal gas or LJ gas.)

timemachine/md/exchange/exchange_mover.py
proteneer: nice! I like that formulation and it makes sense. Will mark this to be cleaned-up when we move away from the special case (eg. more regions, non-uniform choices, et.c).

badisa:
Added a failing test, will follow up with a fix for the Cuda

tests/test_cuda_targeted_insertion_mover.py
```diff
@@ -36,11 +40,109 @@ def compute_ref_log_prob(ref_exchange, water_idx, vi_mols, vj_mols, vol_i, vol_j
     log_weights_after_full = np.array(log_weights_after_full)
     log_weights_after = log_weights_after_full[vj_plus_one_idxs]
 
-    log_p_accept = min(0, logsumexp(log_weights_before) - logsumexp(log_weights_after) + np.log(vol_j) - np.log(vol_i))
+    if len(vi_mols) == ref_exchange.num_waters:
```
proteneer, line 43: hm didn't realize this function was duplicated, should DRY at some point

proteneer:
@badisa  I pushed a commit that addresses the following issues:

1) Compare raw_log_prob, before taking the min (now 32bit and 64bit both fail as expected)
2) Use the PrecomputedChargeHandler
3) DRY: Refactored the code for computing the raw_log_prob a bit, and implemented a `Forcefield::load_precomputed_default(cls)` to the `Forcefield` class.

badisa:
Re-running the buckyball case with 0 and 6 starting waters to see if this significantly changes the behavior.

timemachine/md/exchange/exchange_mover.py
proteneer: > Sorry, not immediately actionable, but in a later pass I'd like to figure out how to write this in terms of a more direct comparison len(vi_mols) == 0 or similar.

Made a change in https://github.com/proteneer/timemachine/pull/1228/commits/d23078bd896db9f02a8100e4424f6d397529ef05

That should reflect this better. Also helped identify another corner case when `N_waters == 1` (which symmetrizes the probabilities again). The original intent was to deal with asymmetries in `N_waters > 1`  but `N_waters_in_vol_i`, `N_waters_in_vol_j` could be `0` or `1`

badisa:
@proteneer @maxentile Updated the description of the PR with the new occupancies of the waters in the buckyball. Seems pretty similar, though ran for 1/4 the time so definitely not converged though I don't think even the comparison values are converged.

badisa (approved):
Looks good, though I have contributed code so I am biased

PR #1230, opened 2024-01-09
Test has_chiral_atom_flips

maxentile:
`has_chiral_atom_flips` was covered in integration tests, but didn't have a unit test -- adds a small randomized test that the function is symmetric

(based on an observation by @mcwitt that checking [both ways](https://github.com/proteneer/timemachine/blob/9e1ece76b088d4b7e3cf753fbe53f4a4da2de036/timemachine/fe/chiral_utils.py#L304-L311) may be redundant and might be simplified for a 2x speedup)

### DIFF
--- a/tests/test_chiral_utils.py
+++ b/tests/test_chiral_utils.py
@@ -1,3 +1,5 @@
+from collections import Counter
+
 import numpy as np
 import pytest
 from rdkit import Chem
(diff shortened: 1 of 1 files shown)


mcwitt (approved):
Looks good to me, thanks for adding!

maxentile:
Converted to draft: Need to (1) fix a mechanical bug (used in-place shuffle instead of permutation), (2) change how the random cores are sampled to improve informativeness (with uniform random integers, all the trials are evaluating to False, which is not informative)

maxentile:
Sorry for the delay -- should be good to go now.

(Modified the distribution of random test instances to introduce explicit conflicts (since they weren't appearing often enough in uniform random cores), and added a secondary assertion that we're not just testing all Falses or all Trues.)

tests/test_chiral_utils.py
```diff
@@ -316,6 +325,61 @@ def test_chiral_conflict_mixed():
     assert len(mixed_map_flips) == 8
     assert len(mixed_map_undefineds) == 1
 
+    assert has_chiral_atom_flips(mixed_map, chiral_set_a, chiral_set_b)
+
+
+def test_has_chiral_atom_conflicts_symmetric(n_trials=100):
```
mcwitt, line 331: nit:
```suggestion
def test_has_chiral_atom_flips_symmetric(n_trials=100):
```

mcwitt (approved):
Sorry I'd missed the issues in the initial review. I've tested this locally for a few seeds, with a much larger number of trials, and checked that the random examples produced cover reasonable fractions of positive and negative cases. Looks great to me!

PR #1231, opened 2024-01-10
Add atom mapping function returning MCS diagnostics

mcwitt:
Adds `atom_mapping.get_cores_and_diagnostics` having the same signature as `get_cores` except that it additionally returns an `MCSDiagnostics` object containing diagnostic information collected during the MCS search.

Useful for benchmarking.

### DIFF
--- a/tests/test_atom_mapping.py
+++ b/tests/test_atom_mapping.py
@@ -1005,4 +1005,6 @@ def test_get_cores_and_diagnostics():
 
         assert diagnostics.core_size >= len(all_cores[0])
         assert diagnostics.num_cores >= len(all_cores)
-        assert diagnostics.total_nodes_visited >= diagnostics.core_size  # must visit at least one node per edge in core
+        assert (
+            diagnostics.total_nodes_visited >= diagnostics.core_size
+        )  # must visit at least one node per atom pair in core
--- a/tests/test_atom_mapping.py
+++ b/tests/test_atom_mapping.py
@@ -8,7 +8,7 @@ from rdkit.Chem import AllChem
 from timemachine.constants import DEFAULT_ATOM_MAPPING_KWARGS
 from timemachine.fe import atom_mapping
 from timemachine.fe.mcgregor import MaxVisitsWarning, NoMappingError
-from timemachine.fe.utils import plot_atom_mapping_grid
+from timemachine.fe.utils import plot_atom_mapping_grid, read_sdf
 
 pytestmark = [pytest.mark.nocuda]
 
@@ -376,8 +376,7 @@ def get_mol_name(mol) -> str:
 @pytest.mark.parametrize("filepath", datasets)
 @pytest.mark.nightly(reason="Slow")
 def test_all_pairs(filepath):
-    mols = Chem.SDMolSupplier(filepath, removeHs=False)
-    mols = [m for m in mols]
+    mols = read_sdf(filepath)
     for idx, mol_a in enumerate(mols):
         for mol_b in mols[idx + 1 :]:
             all_cores = atom_mapping.get_cores(
@@ -992,8 +991,7 @@ def test_min_threshold():
 
 
 def test_get_cores_and_diagnostics():
-    mols = Chem.SDMolSupplier(hif2a_set, removeHs=False)
-    mols = [m for m in mols]
+    mols = read_sdf(hif2a_set)
     n_pairs = 30
     random_pair_idxs = np.random.default_rng(2024).choice(len(mols), size=(n_pairs, 2))
 
--- a/tests/test_atom_mapping.py
+++ b/tests/test_atom_mapping.py
@@ -1005,4 +1005,4 @@ def test_get_cores_and_diagnostics():
 
         assert diagnostics.core_size >= len(all_cores[0])
         assert diagnostics.num_cores >= len(all_cores)
-        assert diagnostics.total_nodes_visited >= diagnostics.core_size
+        assert diagnostics.total_nodes_visited >= diagnostics.core_size  # must visit at least one node per edge in core
--- a/tests/test_atom_mapping.py
+++ b/tests/test_atom_mapping.py
@@ -5,6 +5,7 @@ import pytest
 from rdkit import Chem
 from rdkit.Chem import AllChem
 
+from timemachine.constants import DEFAULT_ATOM_MAPPING_KWARGS
 from timemachine.fe import atom_mapping
 from timemachine.fe.mcgregor import MaxVisitsWarning, NoMappingError
 from timemachine.fe.utils import plot_atom_mapping_grid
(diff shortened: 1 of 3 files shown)


badisa:
On a related note `nodes_visited` from `MCSResult` is reset between core size, @maxentile and I were talking about this offline today. So the `nodes_visited` could be reduced significantly without actually reducing the computation time, IE a change to the implementation results in more nodes being visited overall, but skips to a smaller core size and selects a mapping with a minimal number of visits and returns some tiny number of visited nodes.

Made a branch that reset the max visits, warrants more investigation before doing anything with it https://github.com/proteneer/timemachine/tree/task/dont-reset-visited-nodes

mcwitt:
> On a related note `nodes_visited` from `MCSResult` is reset between core size, @maxentile and I were talking about this offline today. So the `nodes_visited` could be reduced significantly without actually reducing the computation time, IE a change to the implementation results in more nodes being visited overall, but skips to a smaller core size and selects a mapping with a minimal number of visits and returns some tiny number of visited nodes.
> 
> Made a branch that reset the max visits, warrants more investigation before doing anything with it https://github.com/proteneer/timemachine/tree/task/dont-reset-visited-nodes

Good point, I hadn't realized this. In https://github.com/proteneer/timemachine/pull/1231/commits/4cde9b927cbd34fb667e204fc9224803700f1b8b I modified the code to return the total number of cores visited instead of the `MCSResult` object, since we're mainly interested in the total nodes visited as a proxy for effort, and it seemed weird to mix size-specific data in `MCSResult` with a global `nodes_visited`.

timemachine/fe/mcgregor.py
```diff
@@ -246,7 +254,7 @@ def mcs(
         core_array = np.array(sorted(core))
         all_cores.append(core_array)
 
-    return all_cores, mcs_result.all_marcs
+    return all_cores, mcs_result.all_marcs, total_nodes_visited
```
badisa, line 257: Should we also return the total number of cores created? Not sure if that would also be a useful stat

timemachine/fe/atom_mapping.py
```diff
@@ -44,6 +46,51 @@
 # - refinement of marcs matrix is done on uint8 arrays
 
 
+def get_cores_and_total_nodes_visited(
```
badisa, line 49: We should add a basic test of this method, to at least to say `total_nodes_visited > 0`

timemachine/fe/atom_mapping.py
```diff
@@ -44,6 +46,51 @@
...
+    connected_core,
+    max_cores,
+    enforce_core_core,
+    ring_matches_ring_only,
+    complete_rings,
+    enforce_chiral,
+    disallow_planar_torsion_flips,
+    min_threshold,
+) -> Tuple[List[NDArray], int]:
+    """Like :py:func:`get_cores`, but additionally returns the total number of nodes visited during the MCS search."""
```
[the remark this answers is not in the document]
mcwitt: My reasoning was that `get_cores` is part of the existing API and will likely be used more frequently, so it makes the most sense for the documentation to be there. I think it's pretty common for functions serving as the main entry point to have the bulk of the documentation, and lower-level internals to refer to this. But I don't have a strong preference either way.

timemachine/fe/mcgregor.py
mcwitt: Not sure; I guess I'd be inclined to add it as we need it, especially since we'd probably want to introduce a dataclass or something to hold the diagnostics if we return more than one.

badisa: Any reason not to add the dataclass now? Seems like having these diagnostics sooner rather than later would be handy going forward.

timemachine/fe/atom_mapping.py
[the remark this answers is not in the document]
badisa: Perhaps just clarify the comment to state that it is the implementation of `get_cores` with an additional value and to refer to the documentation of the other function.

timemachine/fe/atom_mapping.py
badisa: Thinking about it some more, if there is a case that we could test that when you change the number of valid core sizes the number changes? But might be flaky if we make further changes. Just me spit balling, most important is just to test that this is returning some sane value.

timemachine/fe/mcgregor.py
mcwitt: Added an `MCSDiagnostics` dataclass in https://github.com/proteneer/timemachine/pull/1231/commits/d576f2f2477a514d9c90202f8283baf22a51eeb5. This has the additional benefit that new fields can be added without breaking unpickling of older results (assuming new fields default to `None`).

badisa: Sweet, that will be nice. Should we add the number of visited cores now that we have the dataclass?

tests/test_atom_mapping.py
```diff
@@ -988,3 +989,22 @@ def test_min_threshold():
 
     with pytest.raises(NoMappingError, match="Unable to find mapping with at least 18 atoms"):
         atom_mapping.get_cores(mol_a, mol_b, **core_kwargs, max_visits=10000)
+
+
+def test_get_cores_and_diagnostics():
+    mols = Chem.SDMolSupplier(hif2a_set, removeHs=False)
```
badisa, line 995: nit: use `read_sdf`

badisa (approved):
This will be super useful!

timemachine/fe/atom_mapping.py
mcwitt: Added a basic test in https://github.com/proteneer/timemachine/pull/1231/commits/59e3e0267cb70b3232758605be68e0c7f89e1ddb.

> One test I thought of would be to set max_nodes=1 and assert that total_nodes_visited > 1

This makes sense to me as long as the mols are not perfectly mapped (in which case total_nodes_visited would equal 1?). Will add this case.

tests/test_atom_mapping.py
mcwitt: Fixed in https://github.com/proteneer/timemachine/pull/1231/commits/5950ca667493bba1e737e9d51396eda35d1b4224 (was just copied; fixed in both locations)

timemachine/fe/atom_mapping.py
badisa: The test you added is sufficient for me, my suggestion seems janky. But never say no to more tests lol

mcwitt: Actually a bit confused here. Setting `max_nodes=1` causes the MCS search to fail with `NoMappingError`.

timemachine/fe/mcgregor.py
mcwitt: Added additional fields in https://github.com/proteneer/timemachine/pull/1231/commits/4d6325a198af06c0c04b816b88fb67938b492c95

timemachine/fe/atom_mapping.py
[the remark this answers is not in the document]
mcwitt: Tried to clarify that they are the same implementation without getting too verbose in https://github.com/proteneer/timemachine/pull/1231/commits/d7a20250d263e64a90a7984d61d76579fc477232

timemachine/fe/atom_mapping.py
badisa: It looks like it has to traverse at least the number of atoms in mol a??? https://github.com/proteneer/timemachine/blob/master/timemachine/fe/mcgregor.py#L293-L295

So `max_nodes` has to be at least `mol_a.GetNumAtoms()`?

tests/test_atom_mapping.py
```diff
@@ -988,3 +988,21 @@ def test_min_threshold():
...
+        mol_a = mols[i_a]
+        mol_b = mols[i_b]
+        all_cores, diagnostics = atom_mapping.get_cores_and_diagnostics(mol_a, mol_b, **DEFAULT_ATOM_MAPPING_KWARGS)
+
+        assert len(all_cores) > 0
+        assert all(len(core) == len(all_cores[0]) for core in all_cores)
+
+        assert diagnostics.core_size >= len(all_cores[0])
+        assert diagnostics.num_cores >= len(all_cores)
+        assert diagnostics.total_nodes_visited >= diagnostics.core_size
```
maxentile, line 1008: nit: add a few-word comment about this comparison? wasn't immediately clear to me, but I think it makes sense (have to visit at least one search node per atom-pair added to core)

maxentile (approved):
Looks great to me!

tests/test_atom_mapping.py
mcwitt: Added in https://github.com/proteneer/timemachine/pull/1231/commits/1ab494c6d3f40c0932c9ff3802a6aec31bbf5abe

mcwitt: Whoops, your original wording was more accurate than what I added. Fixed in https://github.com/proteneer/timemachine/pull/1231/commits/6a5b88609c189705e093bd5d362a96e785fdc26b

PR #1232, opened 2024-01-10
Diagnostics for computing the effective excluded volume

proteneer:
This PR adds basic utilities for computing effective excluded volumes for a molecule and investigates how our current all-at-once-4D decoupling scheme lead to poor intermediate states (both in terms of monotonicity and in terms of sharpness. This may motivate us to move to a better, atom-by-atom, decoupling scheme to better mitigate these issues.

Some caveats:

1) Single conformation
2) Crude heuristic for estimating radii derived from a generalization of r_min
3) May not be fully indicative of replica-state problems we're seeing in practice

Screenshot from 2024-01-10 17-46-08

excluded_vol_pr


proteneer:
(closing and moving this to internal dev)

PR #1233, opened 2024-01-11
Atom mapping performance improvements

mcwitt:
Miscellaneous optimizations reduce runtime by about 40%.

Benchmarks with `run_atom_mapping.py`, single process, first 200 edges:

|                                              | approximate speedup |
|-------------------------------------------- |------------------- |
| Jit `torsion_volume`, `pyramidal_volume`     | 10%                 |
| Avoid intermediate set construction          | 5%                  |
| Avoid redundant reverse check                | 10%                 |
| Avoid conversion                             | < 5%                |
| Use boolean arrays, vectorize `refine_marcs` | 15%                 |
| Short-circuit rejection                      | < 5%                |
| Remove redundant try block                   | < 5%                |

### DIFF
--- a/tests/test_chiral_utils.py
+++ b/tests/test_chiral_utils.py
@@ -1,9 +1,7 @@
 from collections import Counter
-from typing import Sequence
 
 import numpy as np
 import pytest
-from numpy.typing import NDArray
 from rdkit import Chem
 from rdkit.Chem import AllChem
 
@@ -16,7 +14,7 @@ from timemachine.fe.chiral_utils import (
     has_chiral_atom_flips,
     setup_find_flipped_planar_torsions,
 )
-from timemachine.fe.mcgregor import UNMAPPED
+from timemachine.fe.mcgregor import core_to_perm
 from timemachine.fe.utils import get_romol_conf
 from timemachine.ff import Forcefield
 from timemachine.md.minimizer import replace_conformer_with_minimized
@@ -230,11 +228,6 @@ molblock_CN = """
   2  6  1  0
   2  7  1  0
 M  END"""
-
-
-def core_to_perm(core: NDArray, num_atoms_a: int) -> Sequence[int]:
-    a_to_b = {a: b for a, b in core}
-    return [a_to_b.get(a, UNMAPPED) for a in range(num_atoms_a)]
 
 
 def test_chiral_conflict_flip():
--- a/tests/test_chiral_utils.py
+++ b/tests/test_chiral_utils.py
@@ -399,7 +399,7 @@ def test_find_flipped_planar_torsions():
 
     find_flipped_planar_torsions = setup_find_flipped_planar_torsions(mol, mol)
 
-    assert find_flipped_planar_torsions(core_ok) == []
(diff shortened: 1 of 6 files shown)


proteneer:
whoa! nice

proteneer:
@mcwitt  should we try and get this merged?

mcwitt:
Note: force-pushed to fix an issue with some intermediate commits (introduced in an earlier rebase) to allow benchmarking commit-by-commit

tests/test_chiral_utils.py
```diff
@@ -229,6 +232,11 @@ def test_find_chiral_bonds():
 M  END"""
 
 
+def core_to_perm(core: NDArray, num_atoms_a: int) -> Sequence[int]:
+    a_to_b = {a: b for a, b in core}
+    return [a_to_b.get(a, UNMAPPED) for a in range(num_atoms_a)]
+
```
maxentile, line 238: location nit: utility fxn might be handy outside of chiral_utils_tests, maybe added to mcgregor.py or atom_mapping.py ?

timemachine/fe/mcgregor.py
```diff
@@ -69,20 +69,15 @@ def refine_marcs(g1, g2, new_v1, new_v2, marcs):
 
     if new_v2 == UNMAPPED:
         # zero out rows corresponding to the edges of new_v1
-        for e1 in g1.get_edges(new_v1):
-            # this is equivalent to new_marcs[e1] = np.zeros(marcs.shape[1])
-            new_marcs[e1] = 0
+        new_marcs[g1.get_edges_as_vector(new_v1)] = False
     else:
-        # mask out every row in marcs
```
maxentile, line 76: retain line comment?

(the np.where update on vectors of bitmasks a few lines later was a little tricky for me to skim, maybe renaming to avoid overwriting the variable `mask` might be useful. `adj1 = g1.get_edges_as_vector(new_v1); adj2 = g2.get_edges_as_vector(new_v2); mask = np.where(adj1[:, np.newaxis], adj2, ~adj2)` or similar...)

maxentile (approved):
Nice improvements!

proteneer (approved):
nice! surprised that `dtype=bool` is 15% faster than `dtype=byte`

timemachine/fe/mcgregor.py
mcwitt: Thanks for the suggestion, this looks a lot clearer. Applied in https://github.com/proteneer/timemachine/pull/1233/commits/8178dd2bf6e8e754fe5a373e3e70597872d4a9fd

tests/test_chiral_utils.py
mcwitt: Moved in https://github.com/proteneer/timemachine/pull/1233/commits/88758e57be39a473f2409be60e243b084c20503a. Also factored out the inverse function `perm_to_core` and added a test.

mcwitt:
> nice! surprised that `dtype=bool` is 15% faster than `dtype=byte`

I think the performance boost comes from enabling the use of boolean indexing to vectorize `refine_marcs` (not the switch to boolean arrays in isolation).

PR #1235, opened 2024-01-18
Add SegmentedWeightedSampler

badisa:
The previous implementation of `WeightedRandomSampler` was only capable of sampling from a single distribution and to collect multiple samples it ran the same kernels in a loop. The new `SegmentedWeightedRandomSampler` implementation adopts the segmented concept of CUB (https://nvlabs.github.io/cub/group___segmented_module.html) so the sampler can handle multiple distributions (or the same distribution repeated) to select multiple samples at once. This reduces the number of kernels and allows superior parallelism.

## Motivation

This is important for batching exchange moves where we want to generate several distributions and collect numerous samples at the same time, as kernel launch time has overhead. This PR will be followed up with a similar segmented LogSumExp implementation that will allow us to get rid of the event sync and subsequently evaluate multiple samples at each proposal. 

## Tasks
- [x] Replace usages of WeightedRandomSampler with SegmentedRandomSampler in TIBD/BD
- [x] Remove WeightedRandomSampler


mcwitt:
Request to streamline review: could you briefly explain `SegmentedWeightedSampler`, how it differs from `WeightedRandomSampler`, and the motivation for the change in the PR description?

badisa:
> Request to streamline review: could you briefly explain `SegmentedWeightedSampler`, how it differs from `WeightedRandomSampler`, and the motivation for the change in the PR description?

Updated the description, let me know if that clarifies the goals.

timemachine/cpp/src/tibd_exchange_move.cu
```diff
@@ -48,10 +48,13 @@ TIBDExchangeMove<RealType>::TIBDExchangeMove(
...
-      d_uniform_noise_buffer_(round_up_even(NOISE_PER_STEP * this->proposals_per_move_)), d_targeting_inner_vol_(1),
-      d_ligand_idxs_(ligand_idxs), d_src_weights_(this->num_target_mols_), d_dest_weights_(this->num_target_mols_),
+      d_uniform_noise_buffer_(round_up_even(NOISE_PER_STEP * this->proposals_per_move_ * this->samples_per_proposal_)),
+      d_targeting_inner_vol_(this->samples_per_proposal_), d_ligand_idxs_(ligand_idxs),
+      d_src_weights_(this->num_target_mols_), d_dest_weights_(this->num_target_mols_),
       d_inner_flags_(this->num_target_mols_), d_box_volume_(1), p_inner_count_(1), p_targeting_inner_vol_(1),
-      d_selected_translation_(3) {
+      d_selected_translation_(this->samples_per_proposal_ * 3),
+      d_sample_after_segments_(this->d_sample_segments_.length), d_weights_before_counts_(this->samples_per_proposal_),
+      d_weights_after_counts_(this->samples_per_proposal_) {
```
badisa, line 57: `d_weights_after_counts_` is currently unused, but is set up, as I got ahead of myself with these changes. I left it in as currently only takes up one float/double and seemed sensible to set up at the same time as `d_weights_before_counts_`. Happy to remove if it seems confusing

timemachine/cpp/src/segmented_weighted_random_sampler.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+public:
+    SegmentedWeightedRandomSampler(const int max_vals_per_segment, const int segments, const int seed);
+
+    ~SegmentedWeightedRandomSampler();
+
+    void sample_device(
+        const int total_values,
+        const int num_segments,
+        const int *d_segments,
```
mcwitt, line 36: naming nit: maybe `d_segment_offsets`? I was initially confused about what this represents.

timemachine/cpp/src/segmented_weighted_random_sampler.cu
```diff
@@ -0,0 +1,128 @@
...
+    const int *d_segments,               // [num_segments + 1]
+    const RealType *d_log_probabilities, // [total_values]
+    int *d_samples,                      // [num_segments]
+    cudaStream_t stream) {
+    // Perform sampling using the gumbel-max-trick like cupy
+    // https://github.com/cupy/cupy/blob/46f6ed6e78661fb1d31b41a072519f119f5a2385/cupy/random/_generator.py#L1115-L1121
+    curandErrchk(curandSetStream(cr_rng_, stream));
+    curandErrchk(templateCurandUniform(cr_rng_, d_gumbel_.data, d_gumbel_.length));
+
+    // Use the noise both as the noise and the intermediate, does change the values in d_gumbel
```
mcwitt, line 48: Without knowing the details of the trick used here, this sounds a bit suspicious. Are we sure this is valid and won't bias the sampling in some way?

timemachine/cpp/src/segmented_weighted_random_sampler.cu
```diff
@@ -0,0 +1,128 @@
...
+};
+
+template <typename RealType>
+void SegmentedWeightedRandomSampler<RealType>::sample_given_noise_device(
+    const int total_values,
+    const int num_segments,
+    const int *d_segments,               // [num_segments]
+    const RealType *d_log_probabilities, // [total_values]
+    const RealType *d_noise,             // [total_values]
+    RealType *d_intermediate,            // [total_values]
```
mcwitt, line 60: nit: is it possible to give this a more descriptive name? I'm not sure what "intermediate" refers to here.

timemachine/cpp/src/segmented_weighted_random_sampler.cu
badisa: If the sampling was biased I would expect this to show up in the tests. This is done because you need to transform the noise before doing argmax (https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/kernels/k_sampling.cuh#L20-L22) the only reason we don't want to modify it in practice is because we want to re-use the same noise repeatedly times to get the determinism between batched and serial versions.

timemachine/cpp/src/segmented_weighted_random_sampler.cu
badisa: Tried to give it a more descriptive name and added a comment in https://github.com/proteneer/timemachine/pull/1235/commits/683d8018e359f0c204abc6c400d6a6cb67fbb790

timemachine/cpp/src/segmented_weighted_random_sampler.hpp
badisa: Good call, renamed in https://github.com/proteneer/timemachine/pull/1235/commits/683d8018e359f0c204abc6c400d6a6cb67fbb790

timemachine/cpp/src/segmented_weighted_random_sampler.hpp
```diff
@@ -0,0 +1,54 @@
...
+    // Stores both the initial uniform random values and the final gumbel distribution
+    DeviceBuffer<RealType> d_gumbel_;
+    DeviceBuffer<cub::KeyValuePair<int, RealType>> d_arg_max_;
+
+    DeviceBuffer<char> d_argmax_storage_;
+
+    curandGenerator_t cr_rng_;
+
+public:
+    SegmentedWeightedRandomSampler(const int max_vals_per_segment, const int segments, const int seed);
```
proteneer, line 29: `segments->num_segments`

timemachine/cpp/src/segmented_weighted_random_sampler.cu
```diff
@@ -0,0 +1,140 @@
...
+    curandErrchk(curandCreateGenerator(&cr_rng_, CURAND_RNG_PSEUDO_DEFAULT));
+    curandErrchk(curandSetPseudoRandomGeneratorSeed(cr_rng_, seed));
+};
+
+template <typename RealType> SegmentedWeightedRandomSampler<RealType>::~SegmentedWeightedRandomSampler() {
+    curandErrchk(curandDestroyGenerator(cr_rng_));
+};
+
+template <typename RealType>
+void SegmentedWeightedRandomSampler<RealType>::sample_device(
```
proteneer, line 36: should this be a const method?

timemachine/cpp/src/segmented_weighted_random_sampler.cu
```diff
@@ -0,0 +1,140 @@
...
+    curandErrchk(templateCurandUniform(cr_rng_, d_gumbel_.data, d_gumbel_.length));
+
+    // Use the noise both as the noise and the buffer for the gumbel noise, does change the values in d_gumbel
+    this->sample_given_noise_device(
+        total_values,
+        num_segments,
+        d_segment_offsets,
+        d_log_probabilities,
+        d_gumbel_.data,
+        d_gumbel_.data,
```
proteneer, line 55: still confused about this

timemachine/cpp/src/segmented_weighted_random_sampler.cu
proteneer: (never mind this can't be const since the rng is generated here).

timemachine/cpp/src/segmented_weighted_random_sampler.cu
proteneer: @mcwitt in the old implementation, we directly overwrote the original `noise` values, but we require *all* input noise to be

1) generated upfront before a sequence of moves
2) be const (i.e. non-modifiable) 

we had a previous implementation that violated 2) (which is now addressed)

tests/test_weighted_random_sampler.py
```diff
@@ -7,55 +7,104 @@
...
+@pytest.mark.parametrize("num_samples", [8000])
 @pytest.mark.parametrize("precision", [np.float32, np.float64])
-def test_random_sampler_simple_distribution(seed, precision):
+def test_segmented_random_sampler_jagged_batches_simple_distributions(seed, num_samples, precision):
+    """Make sure that jagged arrays are handled_correctly"""
+    klass = custom_ops.SegmentedWeightedRandomSampler_f32
+    if precision == np.float64:
+        klass = custom_ops.SegmentedWeightedRandomSampler_f64
+
+    weights = []
```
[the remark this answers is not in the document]
maxentile: +1 for strengthening test with these cases.

>6. contains a single +inf (okay, since it's a huge weight)

quibble: 6 should also throw an error. (sum(weights) should be positive and finite)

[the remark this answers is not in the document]
badisa: > contains a single 0.0 (? not sure what correct behavior here is)

There is already a test for this, should never sample this value. https://github.com/proteneer/timemachine/blob/683d8018e359f0c204abc6c400d6a6cb67fbb790/tests/test_weighted_random_sampler.py#L94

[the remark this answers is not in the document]
proteneer: Ah I missed it, thanks!

re: quibble, yeah I think it's fine to throw an error if there's a `+inf` - I was thinking about computing the probability of the "reverse" of the hastings proposal factor, which can have an infinite weight (if there's a clash). But we never need to *sample* from the reverse probabilities, so it's okay to throw an error.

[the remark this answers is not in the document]
badisa: For these tests do we want asserts in the kernels or just validate these things in the host layers? Asserts will trigger the process to exit so we can't test those in the code as is, but any host layer checks won't occur in practice. Probably say best to add both, but there will be a discontinuity between the host/device code.

tests/test_weighted_random_sampler.py
```diff
@@ -7,55 +7,104 @@
...
-    sampler = klass(size, seed)
+    ref_selection = np.array([rng.choice(x, size=1, p=batch / np.sum(batch)) for batch in probs]).reshape(-1)
+    sampler = klass(size, num_samples, seed)
 
-    test_selection = sampler.sample(num_samples, probs)
+    test_selection = sampler.sample(probs.tolist())
     assert len(test_selection) == num_samples
 
     ks, pv = ks_2samp(ref_selection, test_selection)
     assert ks < 0.05, (pv, ks)
```
maxentile, line 31: non-blocking, but would be good to strengthen this test in two ways:
* increase difficulty of instances -- make weights in each of the random instances much further from uniform (i.e. sample each row of probs from dirichlet, with high concentration parameter)
* increase sensitivity of distribution comparison -- maybe by comparing to probs directly, rather than comparing to a single random sample

observation: the test still passes ~95% of the time if the instances in test_selection are shuffled:

```python
def test_segmented_random_sampler(seed, size, num_samples):

    # Get a random probability vector.
    probs = rng.exponential(scale=1 / 250, size=(num_samples, size))

    x = np.arange(size)
    ref_selection = np.array([rng.choice(x, size=1, p=batch / np.sum(batch)) for batch in probs]).reshape(-1)
    test_selection = np.array([rng.choice(x, size=1, p=batch / np.sum(batch)) for batch in probs]).reshape(-1)
    # np.random.shuffle(test_selection)  # still passes 95% of the time
    assert len(test_selection) == num_samples

    ks, pv = ks_2samp(ref_selection, test_selection)
    assert ks < 0.05, (pv, ks)

    return probs, ref_selection, test_selection, ks, pv
```

timemachine/cpp/src/tibd_exchange_move.cu
```diff
@@ -73,15 +76,27 @@ TIBDExchangeMove<RealType>::TIBDExchangeMove(
         this->num_target_mols_, d_identify_indices_.data, 0);
     gpuErrchk(cudaPeekAtLastError());
 
+    size_t flagged_bytes = 0;
     // Setup buffer for doing the flagged partition
     gpuErrchk(cub::DevicePartition::Flagged(
         nullptr,
-        temp_storage_bytes_,
+        flagged_bytes,
```
proteneer, line 83: Not relevant for this PR, but do we need to re-run flagged partition if we accept a move during a batch of moves? (we can toggle the state easily, but we still need to re-partition). If we do, is this a stable partitioning algorithm (document)?

timemachine/cpp/src/segmented_weighted_random_sampler.cu
badisa: > this function is only used for testing, in BD/TIBD we instead use sample_given_noise_device directly, which does not modify the input noise.

Currently only used in testing, but could be used in production where we don't care about being able to 'rewind' to previous noise.

tests/test_weighted_random_sampler.py
proteneer: good catch, +1 to strengthening this test

timemachine/cpp/src/tibd_exchange_move.cu
badisa: We do need to re-run the partition. This is to ensure bit wise determinism between step sizes, it is a stable partitioning algorithm. Will link to the CUB documentation.

tests/test_weighted_random_sampler.py
```diff
@@ -7,55 +7,104 @@
...
+
+    # Check the pairs of two
+    _, counts = np.unique(np.array(test_selection)[::2], return_counts=True)
+    percentages = counts / (len(weights) // 2)
+    np.testing.assert_allclose(expected_percentages[0], percentages, atol=0.05)
+
+    # Check the triplet
+    _, counts = np.unique(np.array(test_selection)[1:][::2], return_counts=True)
+    percentages = counts / (len(weights) // 2)
+    np.testing.assert_allclose(expected_percentages[1], percentages, atol=0.05)
```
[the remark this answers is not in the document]
badisa: > just ensure idempotency etc.

This is not idempotent, it generates random noise that will change the result of sampling. Will add a test to verify that given the same seed gives the same result.

tests/test_weighted_random_sampler.py
```diff
@@ -7,55 +7,104 @@
...
+    for _ in range(num_samples // 2):
+        weights.append([7.5, 2.5])
+        weights.append([3.3, 3.4, 3.3])
+    expected_percentages = [[0.75, 0.25], [0.33, 0.34, 0.33]]
+
+    largest_segment = len(max(weights, key=lambda x: len(x)))
+
+    sampler = klass(largest_segment, len(weights), seed)
+
+    test_selection = sampler.sample(weights)
```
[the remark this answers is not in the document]
badisa: > Also document shape of the return value

Documented as part of https://github.com/proteneer/timemachine/blob/683d8018e359f0c204abc6c400d6a6cb67fbb790/timemachine/cpp/src/wrap_kernels.cpp#L220-L232 how would you like it clarified?

tests/test_weighted_random_sampler.py
```diff
@@ -7,55 +7,104 @@
...
+    assert len(test_selection) == len(weights)
+
+    # Check the pairs of two
+    _, counts = np.unique(np.array(test_selection)[::2], return_counts=True)
+    percentages = counts / (len(weights) // 2)
+    np.testing.assert_allclose(expected_percentages[0], percentages, atol=0.05)
+
+    # Check the triplet
```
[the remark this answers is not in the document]
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1235/commits/476e142d4eb71007bb8102eb8f0a98ff7e10543c

timemachine/cpp/src/segmented_weighted_random_sampler.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+namespace timemachine {
+
+// A SegmentedWeightedRandomSampler for random sampling from batches of probabilities (np.random.choice(X, replace=True p=probabilities))
+// Uses a similar approach as the segmented APIs in cub
+template <typename RealType> class SegmentedWeightedRandomSampler {
+
+private:
+    const int max_vals_per_segment_; // Max number of values that can be had per segment
+    const int num_segments_;         // Number of segments to run at a time
```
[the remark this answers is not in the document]
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1235/commits/476e142d4eb71007bb8102eb8f0a98ff7e10543c

tests/test_weighted_random_sampler.py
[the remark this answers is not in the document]
badisa: Changed the total sum of the distributions to be different in https://github.com/proteneer/timemachine/pull/1235/commits/1c574fd6cf00f6868b3bb5d99d2e4ad54c641f5f

timemachine/cpp/src/tibd_exchange_move.cu
badisa: Added comment in https://github.com/proteneer/timemachine/pull/1235/commits/34147fb6a3b79629ae5a7f3354f89b5e0252ca30 

We could probably get rid of this, but would prioritize batching over making this deterministic.

proteneer (commented):
(forgot to submit these)

timemachine/cpp/src/segmented_weighted_random_sampler.cu
mcwitt: Got it, I think that makes sense. The comment initially made me think that the same random values were being used twice, but it sounds like we had just been reusing the buffer and overwriting the initial values (and since then have changed to not overwrite, so the point is moot)?

mcwitt (approved):

tests/test_weighted_random_sampler.py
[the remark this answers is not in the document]
badisa: Added in https://github.com/proteneer/timemachine/pull/1235/commits/a4ce7f96fb0dee896a9963e9ef1d2731b5758b5d

timemachine/cpp/src/tibd_exchange_move.cu
```diff
@@ -73,15 +76,27 @@ TIBDExchangeMove<RealType>::TIBDExchangeMove(
...
         d_identify_indices_.data,
         d_inner_flags_.data,
         d_partitioned_indices_.data,
         d_inner_mols_count_.data,
         this->num_target_mols_));
+
+    size_t sum_bytes = 0;
+    // Will need to compute prefix sums of the before and after weights
```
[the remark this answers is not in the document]
badisa: Cleaned up in https://github.com/proteneer/timemachine/pull/1235/commits/ea490e3170aa2c320206805877f7df1f27fa51e3

timemachine/cpp/src/tibd_exchange_move.cu
```diff
@@ -73,15 +76,27 @@ TIBDExchangeMove<RealType>::TIBDExchangeMove(
...
+    gpuErrchk(cub::DeviceScan::InclusiveSum(
+        nullptr, sum_bytes, d_weights_before_counts_.data, this->d_sample_segments_.data, this->samples_per_proposal_));
+    // Take the larger of the two to use as the temp storage data for CUB
+    temp_storage_bytes_ = max(flagged_bytes, sum_bytes);
+
+    // Zero out the sample segments, the first index will always be zero and the inclusive sum will be offset by 1
+    gpuErrchk(cudaMemset(this->d_sample_segments_.data, 0, this->d_sample_segments_.size()));
```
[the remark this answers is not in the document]
badisa: Changed in https://github.com/proteneer/timemachine/pull/1235/commits/ea490e3170aa2c320206805877f7df1f27fa51e3

tests/test_weighted_random_sampler.py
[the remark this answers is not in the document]
badisa: Added validation for 3-8 of the original comment in https://github.com/proteneer/timemachine/pull/1235/commits/b226056ebcc536eda6f0a37d03bc6ddc96984ce9

Did change some tests that had bad clashes that we wouldn't support in practice.

proteneer (approved):

tests/test_weighted_random_sampler.py
badisa: Strengthened the test here: https://github.com/proteneer/timemachine/pull/1235/commits/eefa7867b083747f7aaf76f09aed90fd700b8301

Uses the dirchlet distribution for each sample as well as compares the percentages rather than the ks test

badisa: Realized the test was way too loose and fixed it up in https://github.com/proteneer/timemachine/pull/1235/commits/1da7ca7cda67b839f19906686df8e31e12bd5320

tests/test_weighted_random_sampler.py
```diff
@@ -1,59 +1,146 @@
...
     rng = np.random.default_rng(seed)
 
     # Get a random probability vector.
-    probs = rng.exponential(scale=1 / 250, size=size)
-    prob_total = np.sum(probs)
-    probs /= prob_total
-    # Sort the probabilities to easily visualize
-    probs = np.sort(probs)[::-1]
+    alpha = rng.normal(loc=10.0, size=size)
+    probs = rng.dirichlet(alpha, size=num_samples)
```
maxentile, line 59: Based on a quick check, it looks like this will generate `num_samples` instances that are each very close to uniform.

(Sorry, I said this part of my suggestion backwards "make weights in each of the random instances much further from uniform (i.e. sample each row of probs from dirichlet, with high concentration parameter)")

To make the test more challenging, might use something like the following, where each row will be more or less "spiky." (for rows where alpha is high, the row will be closer to uniform, and for rows where alpha is low, the row will be closer to "one-hot")

```
probs = []
for _ in range(num_samples):
    # some rows will be "spiky" (small alpha), some rows will be more uniform (large alpha)
    alpha = rng.uniform(0.01, 1.0) * np.ones(size) 
    prob_vec = rng.dirichlet(alpha)

    # output will be normalized: maybe mess that up for testing
    # unnormalized_prob_vec = prob_vec * rng.uniform(0.1, 10.0)
    probs.append(prob_vec)
probs = np.array(probs)
```

I may be misunderstanding though -- should the test exercise the case where each of the segments describes a very different distribution from all the other segments?

maxentile (approved):
Didn't review the actual implementation -- only looked at a few lines of one the Python tests -- but don't have any blocking feedback.

tests/test_weighted_random_sampler.py
badisa: > should the test exercise the case where each of the segments describes a very different distribution from all the other segments

I don't think so? The distribution of log weights tends to be relatively uniform. But reading this again I should modify the test to sample the same distribution repeatedly (call sample multiple times) and compare the percentages. Else if you get spiky distributions you either miss values entirely or have large mismatches.

badisa: Modified the logic you had suggested (to avoid generating probabilities of 0.0) and now sample repeatedly. https://github.com/proteneer/timemachine/pull/1235/commits/3a8e4ddb071e294e1f046f11a815b413f93053e5

PR #1225, opened 2023-12-19
Reduce kernel calls for water sampling

badisa:
Speeds up targeted insertion by about 20%. These are the simpler changes.

## Kernel Performance
Ran following test: `pytest tests/test_cuda_targeted_insertion_mover.py -k 'test_moves_with_complex_and_ligand[2023-float32-0.0001-0.002-12500-12500-2.0]'`
Cuda Arch 8.6
GPU: A4000

### Changes

* `void timemachine::k_set_sampled_weight` previously took 20k ns per run, split into two kernels each kernel is only 4k ns each thanks to additional parallelism.
* `void timemachine::k_flag_mols_inner_outer` is now only run once, reducing 1.4% kernel run time

### Old
```
CUDA API Statistics:

 Time (%)  Total Time (ns)  Num Calls  Avg (ns)   Med (ns)   Min (ns)   Max (ns)    StdDev (ns)            Name           
 --------  ---------------  ---------  ---------  ---------  --------  -----------  -----------  -------------------------
     51.0    1,793,051,451    449,433    3,989.6    3,734.0     3,043    1,218,286      2,511.8  cudaLaunchKernel         
     22.7      798,663,183     25,013   31,929.9    1,787.0     1,467      347,431     65,191.8  cudaEventSynchronize     
      8.4      296,085,609     66,162    4,475.2    3,930.0     2,247      487,328      7,847.4  cudaMemcpyAsync          
      6.4      224,447,919      2,017  111,278.1    7,092.0       131  177,337,124  3,948,571.0  cudaMalloc               
      2.7       95,882,718     18,827    5,092.8    4,169.0     2,030      160,571      4,398.3  cudaMemsetAsync          
      2.6       92,579,194      1,175   78,790.8   48,487.0     5,311    1,333,294    119,639.1  cudaMemcpy               
      2.4       83,363,817    113,346      735.5      591.0       429       20,854        617.7  cudaEventRecord          
      2.0       70,461,715     88,420      796.9      650.0       464       19,230        667.7  cudaStreamWaitEvent      
      1.2       42,871,077      2,037   21,046.2    5,717.0       312   11,492,198    258,040.3  cudaFree                 
      0.1        5,270,871        392   13,446.1    6,890.5     1,736      420,042     30,833.5  cudaStreamSynchronize    
      0.1        4,177,442          5  835,488.4  166,974.0    97,309    3,341,631  1,406,091.4  cudaDeviceSynchronize    
[...]

PR #1239, opened 2024-01-23
Switch to usage of SegmentedSumExp

badisa:
Follow up to https://github.com/proteneer/timemachine/pull/1235 that converts the `LogSumExp` class into a `SegmentedLogSumExp`. The previous `LogSumExp` implementation only functioned on a single array of values, and the `SegmentedLogSumExp` is the last nicely separated part of the code necessary to start batching multiple proposals in each step.

This has little to no performance impact (at most 1% faster) despite removing the EventSync.


timemachine/cpp/src/segmented_logsumexp.hpp
```diff
@@ -0,0 +1,34 @@
...
+    SegmentedLogSumExp(const int max_vals_per_segment, const int num_segments);
+
+    ~SegmentedLogSumExp();
+
+    void sum_device(
+        const int total_values,
+        const int num_segments,
+        const int *d_segment_offsets,
+        const RealType *d_values,
+        RealType *d_max_sum_out,
```
proteneer, line 27: `d_max_sum_out` is written `d_max_out` in the `.cu` files.

tests/test_logsumexp.py
```diff
@@ -5,24 +5,74 @@
...
+        summer = custom_ops.SegmentedLogSumExp_f32(max_values_per_segment, num_samples)
     else:
-        summer = custom_ops.LogSumExp_f64(len(values))
+        summer = custom_ops.SegmentedLogSumExp_f64(max_values_per_segment, num_samples)
 
-    test_val = summer.sum(values)
-    assert len(test_val) == 1  # Have to return numpy array to get values back out
+    test_vals = summer.sum(values)
+    assert len(test_vals) == num_samples
+    np.testing.assert_array_equal(test_vals, summer.sum(values))
```
proteneer, line 75: one more bitwise identical test to add in between, assert that a version computing lses of each segment individually, is bitwise identical to computing multiple segments.

timemachine/cpp/src/tibd_exchange_move.cu
```diff
@@ -254,15 +239,15 @@ void TIBDExchangeMove<RealType>::move(
             this->d_samples_.data,
             stream);
 
-        gpuErrchk(cudaEventSynchronize(host_copy_event_));
-        int inner_count = p_inner_count_.data[0];
-
-        // targeting_inner_vol == 1 indicates that we are targeting the inner volume, starting from the outer mols
```
proteneer, line 260: Looks like the logic for this was removed and not replicated elsewhere on the CPU, did all this pinned memory logic get cleaned up too? i.e. is the ternary decision now made on the GPU side?

timemachine/cpp/src/segmented_logsumexp.hpp
```diff
@@ -0,0 +1,34 @@
...
+    // Buffers for finding the max/summed value
+    std::size_t temp_storage_bytes_;
+    DeviceBuffer<char> d_temp_storage_buffer_;
+
+public:
+    SegmentedLogSumExp(const int max_vals_per_segment, const int num_segments);
+
+    ~SegmentedLogSumExp();
+
+    void sum_device(
```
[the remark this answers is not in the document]
badisa: The reason for this is that to combine them would require running an additional kernel that is a trivial operation better pushed into other kernels. Will add back in a comment, was in the previous LogSumExp but removed here.

timemachine/cpp/src/kernels/k_logsumexp.cu
```diff
@@ -21,4 +21,31 @@ void __global__ k_exp_sub_max(
 template void __global__ k_exp_sub_max<float>(const int, const float *, const float *, float *);
 template void __global__ k_exp_sub_max<double>(const int, const double *, const double *, double *);
 
+template <typename RealType>
+void __global__ k_segmented_exp_sub_max(
+    const int num_segments,
```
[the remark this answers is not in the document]
badisa: It can be inferred as long as `num_segments == blockDim.y` however we are likely to eventually run into not being able to launch that many blocks (as we ran into with an early implementation of this). Don't think it is clear/safe to rely on the state of the block configuration for kernels, we have run into a bug because of this in https://github.com/proteneer/timemachine/pull/1168. 

This is why I have been adopting the while loop syntax for kernels. It allows us to be independent of the block configuration as well the kernels always processes all of the data rather than terminating early if the blocks were incorrectly configured.

timemachine/cpp/src/tibd_exchange_move.cu
badisa: Yup, all the pinned code was cleaned up. The ternary decision was always made on the GPU side, just now we don't need to pull it to the host to run logsumexp.

tests/test_logsumexp.py
```diff
@@ -5,24 +5,74 @@
...
-    reference_val = logsumexp(values)
+@pytest.mark.memcheck
+@pytest.mark.parametrize("precision,atol,rtol", [(np.float64, 1e-8, 1e-8), (np.float32, 2e-5, 1e-5)])
+@pytest.mark.parametrize("seed", [2024])
+@pytest.mark.parametrize("loc", [1000, 0, -1000])
+@pytest.mark.parametrize("num_samples", [2, 5, 10, 100])
+def test_segmented_cuda_logsumexp_ragged_arrays(precision, atol, rtol, seed, loc, num_samples):
+    rng = np.random.default_rng(seed)
+    max_values_per_segment = 100
+    values = [rng.normal(size=rng.integers(1, max_values_per_segment), loc=loc) * 1000.0 for _ in range(num_samples)]
```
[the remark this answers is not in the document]
badisa: Currently 1) and 3) pass, 2) and 4) do not. We could do the same behavior that Scipy does to handle this (https://github.com/scipy/scipy/blob/main/scipy/special/_logsumexp.py#L107-L110) in the host (would make tests pass, but wouldn't change the exchange movers), where you might have +inf/-inf (jax potentials), or could do it in the GPU but that would be an additional kernel for code that would typically have nans because of how we handle overflows (and even if you converted to +inf/-inf, they are all indistinguishable in the device code)

> if one particular segment is bad

Do you mean does having a nan/inf in one segment break the other segments?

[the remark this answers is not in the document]
proteneer: > 2) and 4) do not.

What do they currently return?

> Do you mean does having a nan/inf in one segment break the other segments?

Yes, (assuming the other segments do not have a inf/nan)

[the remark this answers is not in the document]
badisa: > What do they currently return?

They currently return Nan, which in the exchange move kernels we convert to +inf (https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/kernels/k_exchange.cu#L49)

timemachine/cpp/src/segmented_logsumexp.hpp
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1239/commits/079443277fba0a34f1fcb030d14c7f7b5951d09c

tests/test_logsumexp.py
```diff
@@ -5,24 +5,74 @@
...
     if precision == np.float32:
-        summer = custom_ops.LogSumExp_f32(len(values))
+        summer = custom_ops.SegmentedLogSumExp_f32(max_values_per_segment, num_samples)
     else:
-        summer = custom_ops.LogSumExp_f64(len(values))
+        summer = custom_ops.SegmentedLogSumExp_f64(max_values_per_segment, num_samples)
 
-    test_val = summer.sum(values)
-    assert len(test_val) == 1  # Have to return numpy array to get values back out
+    test_vals = summer.sum(values)
```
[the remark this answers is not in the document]
badisa: Added note about determinism in https://github.com/proteneer/timemachine/pull/1239/commits/c0217cb224ecf9ea3cd71c8d357afe8ca9c05283

timemachine/cpp/src/segmented_logsumexp.hpp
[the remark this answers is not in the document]
badisa: Attempted to clarify this with comments in https://github.com/proteneer/timemachine/pull/1239/commits/0743f58dfb303e8ed4b3929398594b52a7d613c8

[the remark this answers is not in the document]
proteneer: >  trivial operation better pushed into other kernels

Some considerations:

0) this is at the cost of having double buffers for `(d_lse_max_before_, d_lse_exp_sum_before_), (d_lse_max_after_, d_lse_exp_sum_after_)`, you may still end-up paying for it in terms of additional memsets etc. required. 
1) unless other downstream calls (outside of this class) depend on the individual components (max, exp sum) in a way that's outside of the logsumexp and you've benchmarked it, i'd strongly prefer to encapsulate the lse functionality inside the class as much as possible. 

wdyt?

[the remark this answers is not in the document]
badisa: > this is at the cost of having double buffers

Yes, you do pay 2x the memory usage, there are no memsets around these buffers so there is no additional computational cost.

> i'd strongly prefer to encapsulate the lse functionality inside the class as much as possible.

Think that is reasonable, though now we are going from batching the old implementation (which did this same thing, just done in one array since easy to offset without the segments, already a deboggling comment to figure out how to do that in CUB - https://github.com/proteneer/timemachine/blob/0743f58dfb303e8ed4b3929398594b52a7d613c8/timemachine/cpp/src/segmented_logsumexp.cu#L53) to modifying the implementation. Not clear of what utility this adds for our immediate goals and would rather leave that until we have sorted out batching the exchange moves more broadly.

tests/test_logsumexp.py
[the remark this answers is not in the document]
badisa: Added test (does not match up with 2. and 4.) in https://github.com/proteneer/timemachine/pull/1239/commits/17124265ffa8903feff1e5392911f6c74e94e634

tests/test_logsumexp.py
badisa: Added in https://github.com/proteneer/timemachine/pull/1239/commits/e2882a6ad5109cb36fd585117cea8c20c7a9348f

timemachine/cpp/src/segmented_logsumexp.hpp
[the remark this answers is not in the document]
badisa: Thinking about this a bit more, haven't actually benchmarked in a while (too early on the exchange move stuff to be concrete), think this would add quite a bit of overhead.

If you look at the current implementation (From PR: https://github.com/proteneer/timemachine/pull/1225) which has `k_exp_sub_max` which takes 1.7k ns per kernel call which is run twice for every TIBD proposal. Despite effectively the lowest run time possible (k_arange is 1.7k ns as well), because it is run thousands of time the kernel takes up 1.8% of the entire run time.

```
Time (%)  Total Time (ns)  Instances   Avg (ns)     Med (ns)    Min (ns)   Max (ns)   StdDev (ns)                                                  Name                                                
 --------  ---------------  ---------  -----------  -----------  ---------  ---------  -----------  ----------------------------------------------------------------------------------------------------
1.8       44,713,131     25,001      1,788.5      1,792.0      1,727      2,208         50.6  void timemachine::k_exp_sub_max<float>(int, const T1 *, const T1 *, T1 *)
```

Even assuming that the new kernel to combine them takes half that long, would be almost a percent of the entire run time.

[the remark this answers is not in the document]
badisa: Discussed offline, this was what was done previously. Need to follow up with

* Rename `SegmentedLogSumExp` -> `SegmentedSumExp` and document how to compute `logsumexp`

proteneer (approved):

timemachine/cpp/src/segmented_logsumexp.hpp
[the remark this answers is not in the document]
badisa: Renamed SegmentedLogSumExp to SegmentedSumExp to clarify that the device code does not compute LogSumExp directly, includes comment pointing to code to compute the final logsumexp value. https://github.com/proteneer/timemachine/pull/1239/commits/f512f73e4d72ca61fc4f1e433c3af0cf291345a3

tests/test_logsumexp.py
```diff
@@ -6,23 +9,115 @@
...
+        summer.logsumexp([[1.0]])
+
+    with pytest.raises(RuntimeError, match="empty array not allowed"):
+        summer.logsumexp(
+            [
+                [
+                    1.0,
+                ],
+                [],
+            ]
```
mcwitt, line 31: nit: remove trailing `,` to format more compactly (unless intentional?)

timemachine/cpp/src/segmented_sumexp.hpp
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include "device_buffer.hpp"
+
+namespace timemachine {
+
+// SegmentedSumExp is a class for computing the max and sum(exp(x - max) for x in xs) of a set of values
+// To compute the logsumexp you can use the code in kernels/k_logsumexp.cuh::compute_logsumexp_final given the
+// max and exp sum.
```
mcwitt, line 9: Should this mention `SegmentedSumExp::logsumexp_host`?

timemachine/cpp/src/kernels/k_logsumexp.cu
```diff
@@ -21,4 +21,31 @@ void __global__ k_exp_sub_max(
 template void __global__ k_exp_sub_max<float>(const int, const float *, const float *, float *);
 template void __global__ k_exp_sub_max<double>(const int, const double *, const double *, double *);
```
mcwitt, line 22: Is `k_exp_sub_max` still used anywhere?

timemachine/cpp/src/wrap_kernels.cpp
```diff
@@ -1512,25 +1512,43 @@ void declare_fanout_summed_potential(py::module &m) {
...
-
-                return py_res;
+            py::init([](const int max_vals_per_segment, const int segments) {
+                return new Class(max_vals_per_segment, segments);
+            }),
+            py::arg("max_vals_per_segment"),
+            py::arg("num_segments"))
+        .def(
+            "logsumexp",
+            [](Class &summer, const std::vector<std::vector<double>> &vals) -> std::vector<RealType> {
```
mcwitt, line 1528: nit: this results in a Python binding that accepts lists; would it be preferable to accept arrays instead (e.g. like potentials)

mcwitt (approved):

timemachine/cpp/src/kernels/k_logsumexp.cu
badisa: Nope, removed

timemachine/cpp/src/wrap_kernels.cpp
badisa: This is intentional to be able to provide ragged arrays to test the segment logic. In practice the code will typically be working on ragged arrays (as defined by segments, still a single buffer).

timemachine/cpp/src/kernels/k_logsumexp.cu
badisa: Good catch, removed in https://github.com/proteneer/timemachine/pull/1239/commits/77d48810c1b4f270e961ce2870e7093286865abd

timemachine/cpp/src/segmented_sumexp.hpp
badisa: Added comment in https://github.com/proteneer/timemachine/pull/1239/commits/d6397fbc5d7a16b0f222222f220c676e03be0b9e

tests/test_logsumexp.py
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1239/commits/43a2967746b863a8efd0b6f614ebd3f7ae514b1d

PR #1240, opened 2024-01-24
Fixes support for 0 weights in SegmentedWeightedRandomSampler

badisa:
* Thanks to @maxentile for following up on this after https://github.com/proteneer/timemachine/pull/1235 and picking up that that I had misunderstood a comment by @proteneer


timemachine/cpp/src/segmented_weighted_random_sampler.cu
```diff
@@ -99,28 +99,34 @@ void SegmentedWeightedRandomSampler<RealType>::sample_given_noise_device(
...
         if (num_vals == 0) {
             throw std::runtime_error("empty probability distribution not allowed");
         }
         offset += num_vals;
         h_segments[i + 1] = offset;
-        // Convert the probabilities into log probabilities
+        // Convert the weights into log weights
         for (unsigned long j = 0; j < num_vals; j++) {
-            RealType log_weight = log(probabilities[i][j]);
-            if (!valid_log_weight(log_weight)) {
```
proteneer, line 120: is `valid_log_weight` still used elsewhere?

proteneer:
good catch, so tl;dr:

1) weights can be positive or negative, and don't need to sum to 1
2) probabilities must be non-negative, and must sum to 1

timemachine/cpp/src/segmented_weighted_random_sampler.cu
badisa: Nope, was removed. But realized I need to add back in an assertion to the kernels `assert(isnan(weight))`

badisa:
> good catch, so tl;dr:
> 
>     1. weights can be positive or negative, and don't need to sum to 1
> 
>     2. probabilities must be non-negative, and must sum to 1

Nuance to 1), currently don't allow negative. Needs to be non-negative based on your previous comment https://github.com/proteneer/timemachine/pull/1235#discussion_r1461942127 though we could support all positive or all negative.

timemachine/cpp/src/segmented_weighted_random_sampler.cu
badisa: Added back in check for nans: https://github.com/proteneer/timemachine/pull/1240/commits/1440cc27771daaf6836a2d249a73bfbb7d4b2c77

proteneer:
thanks, addressed offline:

```
1. log weights (are equal energies/kT) can be positive or negative, and don't need to sum to 1
2. weights are exp(energies/KT) which must be non-negative, and don't need to sum to 1
3. probabilities are normalized weights, which must be non-negative, and must sum to 1
```

proteneer (approved):

tests/test_weighted_random_sampler.py
```diff
@@ -155,3 +152,25 @@ def test_segmented_random_sampler_simple_distribution(seed, num_samples, precisi
...
+
+
+@pytest.mark.parametrize("seed", [2024])
+@pytest.mark.parametrize("num_samples", [1000])
+@pytest.mark.parametrize("precision", [np.float32, np.float64])
+def test_segmented_random_sampler_zero_probability(seed, num_samples, precision):
+    """Make sure that if we zero out a probability we never sample that value"""
+
+    # Setup weights such that expected percentages are obvious
+    weights = [[0.0, 2.5] for _ in range(num_samples)]
```
mcwitt, line 164: nit: seems like elsewhere we use "weights" to mean "log probabilities"?
```suggestion
    probabilities = [[0.0, 2.5] for _ in range(num_samples)]
```

tests/test_weighted_random_sampler.py
```diff
@@ -155,3 +152,25 @@ def test_segmented_random_sampler_simple_distribution(seed, num_samples, precisi
...
+    if precision == np.float64:
+        klass = custom_ops.SegmentedWeightedRandomSampler_f64
+
+    sampler = klass(len(weights[0]), num_samples, seed)
+
+    test_selection = sampler.sample(weights)
+    assert len(test_selection) == num_samples
+
+    # All of the values will be the index that is non-zero
+    assert np.all(np.array(test_selection) == 1)
```
mcwitt, line 176: nit, optional: Is it worth strengthening this to test more than 2 choices and/or different weights in each segment?

tests/test_weighted_random_sampler.py
badisa: I think by https://github.com/proteneer/timemachine/pull/1240#issuecomment-1909145964 we are using the convention of:

* weights: unnormalized probabilities (positive sum but doesn't have to be 1), non-log space. The Exchange movers are imprecise in calling log weights 'weights', can adjust
* log-weights: log space weights
* probabilities: normalized weights that sum to 1.0

mcwitt: Ah, got it, thanks!

timemachine/cpp/src/segmented_weighted_random_sampler.cu
```diff
@@ -99,28 +99,34 @@ void SegmentedWeightedRandomSampler<RealType>::sample_given_noise_device(
...
-                throw std::runtime_error(
-                    "all values in log space must be finite and non-negative: " + std::to_string(probabilities[i][j]));
+            RealType weight = weights[i][j];
+            if (weight == -inf || weight == inf) {
```
mcwitt, line 121: nit: could be moved below the negative check and remove the `-inf` case?

mcwitt (approved):

tests/test_weighted_random_sampler.py
badisa: Looking the weights in tibd/bd need to be cleaned up in a more specific way then I had thought. Going to put that to another PR to avoid mixing up orthogonal changes.

badisa: Nevermind, misunderstood the code, renamed the variables to clarify that they are log weights in https://github.com/proteneer/timemachine/pull/1240/commits/1692502dbbcf6544b8752ec63a38e84bdeff1a9a

timemachine/cpp/src/segmented_weighted_random_sampler.cu
```diff
@@ -99,28 +99,34 @@ void SegmentedWeightedRandomSampler<RealType>::sample_given_noise_device(
 
 template <typename RealType>
 std::vector<int>
-SegmentedWeightedRandomSampler<RealType>::sample_host(const std::vector<std::vector<RealType>> &probabilities) {
-    const int num_segments = static_cast<int>(probabilities.size());
+SegmentedWeightedRandomSampler<RealType>::sample_host(const std::vector<std::vector<RealType>> &weights) {
+    const int num_segments = static_cast<int>(weights.size());
```
maxentile, line 103: should [wrap_kernels.cpp](https://github.com/proteneer/timemachine/blob/4d8787dc3529842a4bd06b81c24e9596b38d0d94/timemachine/cpp/src/wrap_kernels.cpp#L221-L227) be updated to match naming convention in https://github.com/proteneer/timemachine/pull/1240#discussion_r1466679534 ?

maxentile (approved):
Looks good!

Main thing I was looking for was for the tests to exercise a case where weight vectors contain at least one 0.

The naming changes also seem like improvements -- nicely clarifies when things are assumed normalized or not, and when things are in log space or not.

timemachine/cpp/src/segmented_weighted_random_sampler.cu
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1240/commits/5911687fc113dbbf7727aea7810c514e2b5ab162

tests/test_weighted_random_sampler.py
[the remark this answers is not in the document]
badisa: Strengthened with a modified variant of the code you suggested: https://github.com/proteneer/timemachine/pull/1240/commits/79a1424a4a0d86f21e26570dcd3420e3c3a671ef

PR #1241, opened 2024-01-26
Reduce Memory used for Coordinates in Exchange Moves

badisa:
Previously the coordinates for the new proposals were stored with all N atoms, which will cause difficulties (memory usage, duplicating coords repeatedly) when the number of proposals per step is greater than 1. To avoid this, this PR only stores the rotated/translated atom coords of each proposal. This allows us to more easily batch as well as removes the need to do a memcpy (~0.3% of the runtime) on coordinates.

This PR also cleans up the terminology and tries to document some of the changes that were made in previous PRs which weren't updated when they should have been.


timemachine/cpp/src/bd_exchange_move.hpp
```diff
@@ -23,42 +28,45 @@ template <typename RealType> class BDExchangeMove : public Mover {
...
     // Buffer for evaluating moves without touching the original coords
-    DeviceBuffer<double> d_intermediate_coords_;
-    DeviceBuffer<double> d_params_;
-    DeviceBuffer<__int128> d_mol_energy_buffer_;
-    DeviceBuffer<RealType> d_sample_per_atom_energy_buffer_; // [mol_size_ * N]
-    DeviceBuffer<int> d_atom_idxs_;
-    DeviceBuffer<int> d_mol_offsets_;
-    DeviceBuffer<RealType> d_log_weights_before_;
-    DeviceBuffer<RealType> d_log_weights_after_;
+    DeviceBuffer<double> d_intermediate_coords_;             // [batch_size_, mol_size_, 3]
```
proteneer, line 44: nice, super helpful!

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -581,14 +581,21 @@ void __global__ k_accumulate_atom_energies_to_per_mol_energies(
     per_mol_energies[idx] = energy_accumulator;
 }
 
+// k_atom_by_atom_energies is intended to be used for computing the energies of a subset of atoms
+// against all other atoms. The kernel allows changing the positions of the target atomsby passing in an
```
proteneer, line 585: atomsby -> atoms by

timemachine/cpp/src/kernels/k_rotations.cu
```diff
@@ -88,16 +88,17 @@ void __global__ k_rotate_coordinates(
...
-    double *__restrict__ coords_out) {
+    const int *__restrict__ samples,           // [batch_size]
+    const int *__restrict__ mol_offsets,       // [batch_size + 1]
+    const RealType *__restrict__ quaternions,  // [batch_size, 4]
+    const RealType *__restrict__ translations, // [batch_size, 3]
+    double *__restrict__ coords_out            // [batch_size, num_atoms, 3]
+) {
 
-    int sample_idx = blockIdx.x * blockDim.x + threadIdx.x;
+    int batch_idx = blockIdx.x * blockDim.x + threadIdx.x;
```
proteneer, line 101: `batch_idx` sounds like which particular batch we're processing when we're doing multiple batches, maybe `idx_in_batch` is more appropriate here?

timemachine/cpp/src/kernels/k_rotations.cuh
```diff
@@ -25,13 +25,13 @@ void __global__ k_rotate_coordinates(
 // at the translation as the final step, if SCALE=true then the translation that is provided will be scaled by the box vectors.
 template <typename RealType, bool SCALE>
 void __global__ k_rotate_and_translate_mols(
-    const int num_samples,
+    const int num_batches,
```
proteneer, line 28: Does this process *multiple* batches? Or do you mean `batch_size` here?

proteneer (changes requested):

timemachine/cpp/src/kernels/k_rotations.cu
badisa: Renamed to `idx_in_batch` in https://github.com/proteneer/timemachine/pull/1241/commits/3540ce7380cce947b5984fb02f7cd631a37c2306

timemachine/cpp/src/kernels/k_rotations.cuh
badisa: Renamed `batch_size` in https://github.com/proteneer/timemachine/pull/1241/commits/7fb988bf6bab6af0288abfefbac72066aa529f74

timemachine/cpp/src/kernels/k_nonbonded.cuh
badisa: Fixed typo in https://github.com/proteneer/timemachine/pull/1241/commits/aa8c0564e34e4e1f550d80e6d980d3badd456829

badisa:
Added determinism test in https://github.com/proteneer/timemachine/pull/1241/commits/8c37fd369f070e767f42d335ee5f0d627628e83d 

Matches master exactly, so this change has no impact on the bitwise determinism (as expected).

I didn't put the data into `timemachine.testsystems` like some of our other data as it doesn't seem like the type of thing we want installed with the package (which is why it doesn't use importlib). We seem to currently install test data with the timemachine package, but doesn't seem useful for ship test data with the package (unless perhaps users install with [test])`

tests/test_cuda_bd_exchange_mover.py
```diff
@@ -54,7 +54,7 @@ def test_bd_exchange_validation(precision):
         klass(N, group_idxs, params, DEFAULT_TEMP, beta, cutoff, seed, proposals_per_move, 1)
 
     # Proposals must be non-zero
-    with pytest.raises(RuntimeError, match="proposals per move must be greater than 0"):
+    with pytest.raises(RuntimeError, match="steps per move must be greater than 0"):
```
mcwitt, line 57: nit: was this change intentional? (looks like the `num_proposals_per_move` that is set to zero here?)

proteneer (approved):

tests/test_cuda_bd_exchange_mover.py
badisa: Good catch, fixed in [4de6238](https://github.com/proteneer/timemachine/pull/1241/commits/4de62383145c04e5c71e2cca1ce890b4de899881)

examples/water_sampling_mc.py
```diff
@@ -71,6 +71,9 @@ def test_exchange():
     parser.add_argument("--iterations", type=int, help="Number of iterations", default=1000000)
     parser.add_argument("--equilibration_steps", type=int, help="Number of equilibration steps", default=50000)
     parser.add_argument("--seed", default=2024, type=int, help="Random seed")
+    parser.add_argument(
+        "--save_last_frame", type=str, help="Store last frame as a npz file, used to verify bitwise determinism"
```
mcwitt, line 75: nit: could this be `action="store_true"` instead of `type=str`?

badisa: This could be, the reason it is a string is to be able to write out names to disambiguate between the targeted and untargeted modes.

mcwitt: sorry, disregard - I'd misread this as a boolean flag.

mcwitt (approved):

PR #1242, opened 2024-01-29
Avoid raising secondary exception in DeviceBuffer destructor

badisa:
* Pybind called std::terminate(https://github.com/pybind/pybind11/issues/3336#issuecomment-939527115) which causes issues

### DIFF
--- a/timemachine/cpp/src/device_buffer.cu
+++ b/timemachine/cpp/src/device_buffer.cu
@@ -3,6 +3,7 @@
 #include "gpu_utils.cuh"
 #include <cstddef>
 #include <cub/util_type.cuh>
+#include <exception>
 
 namespace timemachine {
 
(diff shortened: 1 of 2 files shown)


timemachine/cpp/src/device_buffer.cu
```diff
@@ -33,10 +34,14 @@ template <typename T> void DeviceBuffer<T>::realloc(const size_t new_length) {
...
-    // TODO: the file/line context reported by gpuErrchk on failure is
+    // If there are uncaught exceptions, attempt to clean up but don't abort the process in case
+    // the exception is an InvalidHardware exception and gpuAssert would raise an identical error causing
+    // a call to std::terminate from pybind
+    const bool abort = std::uncaught_exceptions() == 0;
+    // TODO: the file/line context reported by gpuAssert on failure is
     // not very useful when it's called from here. Is there a way to
     // report a stack trace?
-    gpuErrchk(cudaFree(data));
+    gpuAssert(cudaFree(data), __FILE__, __LINE__, abort);
```
mcwitt, line 44: Wondering if we could simplify this to just never abort from a destructor? I'm not immediately coming up with scenarios where this seems desirable.

badisa: Hmmm... not sure. 

I am inclined to make this have as small of an impact as possible, since often we do get aborts that are triggered due to invalid memory accesses. For example the following https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/context.cu#L161-L167 if we throw a runtime exception that indicates the state of the context is bad (such as https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/local_md_potentials.cu#L235) we might want to treat any cuda errors as more serious and abort, rather then silencing them and returning the exception which might be red-herring.

badisa:
Need to come up with a better way to handle this.

We currently rely on the exception raised in gpuErrchk downstream, but cause issues in destructors.

PR #1243, opened 2024-01-31
Re-enable chiral atom restr

maxentile:
Partially reverting #815, #832 .

(Focusing on chiral_atom in current PR, but chiral_bond will follow.)


tests/test_chiral_restraints.py
```diff
@@ -626,5 +638,64 @@ def assert_same_signs(a, b):
...
+        """map part of ring + other substituent"""
+        substruct = Chem.MolFromSmiles("C1CCC1O")
+        atom_map = dict(zip(mol_a.GetSubstructMatch(substruct), mol_b.GetSubstructMatch(substruct)))
+        conf_a = mol_a.GetConformer()
+        coord_map = {atom_map[idx]: conf_a.GetAtomPosition(idx) for idx in atom_map}
+        AllChem.EmbedMolecule(mol_b, randomSeed=2045, coordMap=coord_map)
+        core = get_cores(mol_a, mol_b, **DEFAULT_ATOM_MAPPING_KWARGS)[0]
+        assert len(core) == 10
+
+        return mol_a, mol_b, core
```
maxentile, line 692: should inline the resulting Mol blocks, maybe keep these as a comment

timemachine/fe/topology.py
```diff
@@ -403,30 +403,27 @@ def setup_chiral_restraints(self, restraint_k=1000.0):
...
-
-        chiral_atom_params = []
-        for idxs in chiral_atom_restr_idxs:
-            chiral_atom_params.extend(restraint_k for _ in idxs)
-
-        chiral_atom_params = np.array(chiral_atom_params)
-        chiral_atom_restr_idxs = np.array(chiral_atom_restr_idxs)
+        # chiral atoms
+        chiral_atom_restr_idxs = np.array(chiral_utils.setup_all_chiral_atom_restr_idxs(mol, conf))
+        chiral_atom_params = restraint_k * np.ones(len(chiral_atom_restr_idxs))
```
maxentile, line 411: note: size of `chiral_atom_params` was adjusted (from `len(chiral_atom_restr_idxs.flatten())` to `len(chiral_atom_restr_idxs)`) -- may need to make a similar adjustment for chiral_bonds

tests/test_chiral_restraints.py
proteneer: Good idea - I think EmbedMolecule is OS dependent (even if you're re-using the same seeds)...

tests/test_chiral_restraints.py
```diff
@@ -626,5 +636,64 @@ def assert_same_signs(a, b):
...
+
+    # check that it runs
+    mol_a, mol_b, core = generate_more_complete_mapping()
+    _ = run_vacuum(mol_a, mol_b, core, ff, None, very_short_hrex_params, n_windows=3)
+
+    mol_a, mol_b, core = generate_less_complete_mapping()
+    _ = run_vacuum(mol_a, mol_b, core, ff, None, very_short_hrex_params, n_windows=3)
+
+    # TODO: slow, assert endstate chirality is preserved
```
proteneer, line 699: Would be good to check that the assertion fails without the use of chiral restraints (and then subsequently, fixed with the use of chiral restraints).

tests/test_chiral_restraints.py
```diff
@@ -626,5 +636,64 @@ def assert_same_signs(a, b):
...
+        atom_map = dict(zip(mol_a.GetSubstructMatch(substruct), mol_b.GetSubstructMatch(substruct)))
+        conf_a = mol_a.GetConformer()
+        coord_map = {atom_map[idx]: conf_a.GetAtomPosition(idx) for idx in atom_map}
+        AllChem.EmbedMolecule(mol_b, randomSeed=2045, coordMap=coord_map)
+        core = get_cores(mol_a, mol_b, **DEFAULT_ATOM_MAPPING_KWARGS)[0]
+        assert len(core) == 10
+
+        return mol_a, mol_b, core
+
+    # check that it runs
```
proteneer, line 692: Would be it useful to generate an explicit ring-breaking atom-mapping as well (replicating the original observation)?

timemachine/fe/topology.py
proteneer: Good catch - looks like the previous implementation was just incorrect - (i.e. we had 4 times as many `k` parameters as there were actual # of restraints). I'm guessing the code still ran since we use the same `k`, and every restraint still had a `k` value they could reference.

tests/test_chiral_restraints.py
maxentile: Will make this more explicit.

(Subtlety: original observation was _not_ a ring-breaking transformation, but a partially mapped ring due to incomplete alignment, which I tried to emulate in generating these two different alignments / atom-mappings.)

```python
    def generate_more_complete_mapping():
        """map whole ring + one substituent"""
```
produces something like:

image


and 
```python
    def generate_less_complete_mapping():
        """map part of ring + other substituent"""
```
produces something like
image

proteneer: Ah got it, thank you for including the pictures, this resolves my concern. I was looking at the line `substruct = Chem.MolFromSmiles("C1CCC1O")` and thinking that it also implied the core being a complete ring somehow (since the ring atom positions of `mol_b` are then fixed using the `coordMap`) and are then fully mapped during the MCS. But I missed that `get_cores` will always end up returning a core with a partial ring due to the use of chiral filters during the MCS, even if all distance checks pass.

maxentile: Sorry, 2 independent confusing things, one possibly important, one inconsequential:

> get_cores will always end up returning a core with a partial ring due to the use of chiral filters during the MCS, even if all distance checks pass.

For future reference, this is not a correct description of the behavior of the MCS implementation.

The demapped ring atom (atom idx 1) differed by only 1.03 A between mol_a and mol_b (which is indeed below [ring_cutoff 1.2 A](https://github.com/proteneer/timemachine/blob/60b37e6fc0e541c8129575e585702d9cd1d3aa6a/timemachine/constants.py#L29)) . But increasing the distance cutoffs even further **does** cause this atom to be mapped:

image

(The partial demapping of this ring **does** appear to depend on the distance checks, and get_cores can return a complete ring match when a sufficient number of additional distance checks pass. But it may also interact with chiral filters or other filters -- I do not have a good understanding of what guarantees might hold for the overall discrete search routine including all filters.)

> was looking at the line substruct = Chem.MolFromSmiles("C1CCC1O") and thinking that it also implied the core being a complete ring somehow (since the ring atom positions of mol_b are then fixed using the coordMap) and are then fully mapped during the MCS.

(Detail that's not very pertinent: `EmbedMolecule(..., coord_map)` almost never succeeds in exactly snapping the selected substructures to requested coordinates with default settings. But toggling `useRandomCoords=True` [usually does the trick](https://greglandrum.github.io/rdkit-blog/posts/2023-02-10-more-on-constrained-embedding.html). (Note that the atoms in the matched substructure are exactly aligned in the atom map produced by `generate_more_complete_mapping()` where `useRandomCoords=True`, but not in the atom map produced by `generate_less_complete_mapping()` where `useRandomCoords=False` by default.) This was an arbitrary and implicit way to generate more or less `mol_a`-aligned conformers for `mol_b` -- removed from test script and moved outputs into tests/data/*.sdf in https://github.com/proteneer/timemachine/pull/1243/commits/79bb03df0571a42d3e292a5e098db047bc83f246 )

proteneer: (brief pedagogical note after offline discussion with @maxentile: for the last row, flipping the molecule allows for a full mapping, if every atom is allowed to map to every atom.)

tests/test_free_energy.py
```diff
@@ -280,9 +279,14 @@ def test_plot_pair_bar_plots(mock_fig, hif2a_ligand_pair_single_topology_lam0_st
     )
     make_pair_bar_plots(pair_result, DEFAULT_TEMP, "")
     assert mock_fig.call_args is not None
-    assert set(mock_fig.call_args.args[0]) == set(
-        ["HarmonicBond", "HarmonicAngleStable", "PeriodicTorsion", "NonbondedPairListPrecomputed"]
-    )
+    expcted_potentials = {
```
badisa, line 282: nit/drive-by: typo

timemachine/fe/system.py
```diff
@@ -129,29 +134,22 @@ def get_U_fn(self):
...
-            )
+            return sum(U(x, box=None) for U in U_fns)
 
         return U_fn
 
     def get_U_fns(self) -> List[BoundPotential[Potential]]:
         # For molecules too small for to have certain terms,
         # skip when no params are present
+        # Chiral bond restraints are disabled until checks are added (see GH #815)
+        potentials = [self.bond, self.angle, self.torsion, self.nonbonded, self.chiral_atom]  # , self.chiral_bond]
```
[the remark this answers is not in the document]
maxentile: Ahh, thanks for catching this -- I think that's spot-on.

(Was confused why this seemed to work in validation even though the ff test was failing -- will need to look over for other cases where we might assume specific set of potentials in a specific order...)

tests/test_chiral_restraints.py
```diff
@@ -626,5 +635,30 @@ def assert_same_signs(a, b):
...
+    mol_dict = {utils.get_mol_name(mol): mol for mol in utils.read_sdf("tests/data/1243_chiral_ring_confs.sdf")}
+    mol_a = mol_dict["A"]
+    mol_b_0 = mol_dict["B_0"]
+    mol_b_1 = mol_dict["B_1"]
+    core_0 = get_cores(mol_a, mol_b_0, **DEFAULT_ATOM_MAPPING_KWARGS)[0]
+    core_1 = get_cores(mol_a, mol_b_1, **DEFAULT_ATOM_MAPPING_KWARGS)[0]
+    assert len(core_0) == 10
+    assert len(core_1) == 15
+
+    _ = run_vacuum(mol_a, mol_b_0, core_0, ff, None, very_short_hrex_params, n_windows=3)
```
badisa, line 661: What is `run_vacuum` used here for? It looks like you don't rely on any output from this function call. Is it to verify no exception is triggered, if so could we target the code specifically that we might expect to raise the assertion rather then run vacuum HREX?

badisa: Oh, looks like this test is still incomplete

timemachine/fe/system.py
```diff
@@ -129,29 +134,22 @@ def get_U_fn(self):
...
-            # chiral_U = chiral_atom_U(x) + chiral_bond_U(x)
-            return (
-                self.bond(x, box=None)
-                + self.angle(x, box=None)
-                + self.torsion(x, box=None)
-                + self.nonbonded(x, box=None)
-                # + self.chiral_atom(x, box=None)
-                # + self.chiral_bond(x, box=None)
-            )
+            return sum(U(x, box=None) for U in U_fns)
```
badisa, line 141: Should this be `jnp.sum()`? It looks like jax handles the builtin `sum` correctly, so maybe not.

timemachine/fe/topology.py
```diff
@@ -403,30 +403,27 @@ def setup_chiral_restraints(self, restraint_k=1000.0):
...
+        assert len(chiral_atom_params) == len(chiral_atom_restr_idxs)  # TODO: can this be checked in Potential::bind ?
```
badisa, line 412: This is checked in `ChiralAtomRestraint::execute_device`, though might be a little too far from construction to be useful.

tests/test_chiral_restraints.py
maxentile: >Is it to verify no exception is triggered, if so could we target the code specifically that we might expect to raise the assertion rather then run vacuum HREX?

Correct -- all this does is run ~2 steps of MD in 3 windows, hopefully without throwing an exception. The main places I wanted to check that might raise an exception would be: (1) invoking the single topology constructor, (2) constructing chiral atom GPU potentials / MD contexts, (3) producing diagnostics that depend on list of potentials.

badisa: It would be good to add a comment with the detail you added, makes perfect sense with that context.

timemachine/fe/system.py
maxentile: You're right.

(When I wrote this, wasn't sure whether every instance of `BoundPotential` returns a jax array or if it may return a float. (Just confirmed these would never be `BoundGpuImplWrapper`, which returns float.))

proteneer (approved):
Looks great!

badisa (approved):

timemachine/fe/system.py
```diff
@@ -129,29 +135,22 @@ def get_U_fn(self):
...
-            # chiral_U = chiral_atom_U(x) + chiral_bond_U(x)
-            return (
-                self.bond(x, box=None)
-                + self.angle(x, box=None)
-                + self.torsion(x, box=None)
-                + self.nonbonded(x, box=None)
-                # + self.chiral_atom(x, box=None)
-                # + self.chiral_bond(x, box=None)
-            )
+            return jnp.sum(jnp.array([U(x, box=None) for U in U_fns]))
```
mcwitt, line 142: nit: might want to check that this isn't introducing a performance bottleneck. I think the computation on this line is dominated by `jnp.array`, which can be relatively slow (especially if not jitted). In this example, summing the python list directly is ~500x faster:
image

maxentile: Reverted in https://github.com/proteneer/timemachine/pull/1243/commits/de851032534c1a0668497e9e3195b70e97f4d118 to be on the safe side

mcwitt (approved):

maxentile:
Added a nightly test, which is passing locally.

(Will probably refine and move the test's utility functions into a more appropriate location (chiral_utils.py) in a future PR.)

PR #1245, opened 2024-02-02
Store NonbondedPairList scales as RealType

badisa:
* Had this lying around for awhile, can't remember if there is some issue with it.
* Minor doc update that I noticed as part of reviewing https://github.com/proteneer/timemachine/pull/1243

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -46,9 +46,11 @@ def fixed_overflowed(a):
     return converted_a == np.iinfo(np.int64).max
 
 
-def prepare_single_topology_initial_state(st: SingleTopology, host_config: HostConfig, lamb: float = 0.1):
+def prepare_single_topology_initial_state(st: SingleTopology, host_config: Optional[HostConfig], lamb: float = 0.1):
     temperature = DEFAULT_TEMP
-    host = rbfe.setup_optimized_host(st, host_config)
+    host = None
+    if host_config is not None:
+        host = rbfe.setup_optimized_host(st, host_config)
     initial_state = rbfe.setup_initial_states(st, host, temperature, [lamb], seed=2022)[0]
     return initial_state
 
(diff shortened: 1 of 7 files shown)


proteneer:
I think by default we prefer doubles everywhere unless there's a speed benefit to converting to float. Does it make vacuum simulations faster?

badisa:
> I think by default we prefer doubles everywhere unless there's a speed benefit to converting to RealType. 
In the case that we only read the values (and convert from double to float https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/kernels/k_nonbonded_pair_list.cuh#L99-L100) and don't write them, is that the case? We are storing 2x the number of bytes and doing a conversion that doesn't impact behavior.

> Does it make vacuum simulations faster?

Good question, realized that we don't have a vacuum simulation in our benchmarks. Will add

badisa:
@proteneer Added a vacuum benchmark in https://github.com/proteneer/timemachine/pull/1245/commits/3ba48f8308c08f1824de63740b660affd08f130b

Seems to be about 2% faster with the change, but running this on my personal laptop so may not be converged

## Before
vacuum-rbfe: N=35 speed: 12610.71ns/day dt: 2.5fs (ran 10000000 steps in 171.32s)

## After
vacuum-rbfe: N=35 speed: 12837.65ns/day dt: 2.5fs (ran 10000000 steps in 168.29s)

proteneer:
Wow! 12.8ms per day - I wonder why our vacuum simulations are so slow. Is this with or without HREX?

badisa:
> Wow! 12.8us per day - I wonder why our vacuum simulations are so slow. Is this with or without HREX? (Guessing this is single state simulations)

Without HREX. Can also look at the HREX tests, though more going on in HREX so didn't seem as clean of a benchmark.

badisa:
Seems to make no difference on Tesla cards which is target, closing. Will move over the doc changes to another PR at some point (perhaps deboggling)

PR #1252, opened 2024-02-08
Fix water sampling nightly tests.

proteneer:
This PR addresses the nightly failure on my machine locally. `N_accepts=17`


timemachine/potentials/nonbonded.py
```diff
@@ -98,7 +97,13 @@ def nonbonded_block_unsummed(xi, xj, box, params_i, params_j, beta, cutoff):
     ri = jnp.expand_dims(xi, axis=1)
     rj = jnp.expand_dims(xj, axis=0)
 
-    dij = jnp.linalg.norm(delta_r(ri, rj, box), axis=-1)
+    w_i = jnp.expand_dims(params_i[:, 3], axis=1)
+    w_j = jnp.expand_dims(params_j[:, 3], axis=0)
+
+    d2ij = delta_r(ri, rj, box)
+    d2ij = jnp.concatenate([d2ij, jnp.array(w_i - w_j).reshape(*d2ij.shape[:-1], 1)], axis=-1)
```
badisa, line 104: The lack of support for w could result in some cases failing due to lambda not being 0.0

tests/test_nonbonded.py
```diff
@@ -129,7 +129,7 @@ def test_correctness(self):
         Test against the reference jax platform for correctness.
         """
         # we can't go bigger than this due to memory limitations in the the reference platform.
-        for N in [33, 65, 231, 1050, 4080]:
+        for N in [33, 65, 231, 1050, 3080]:
```
badisa, line 132: The `jax.grad` ended up causing OOMs in the test. Best I can tell is that we were just scrapping by and with the addition of more tests it pushed it beyond the limit (https://stackoverflow.com/questions/20900380/py-test-excessive-memory-usage-with-large-number-of-tests).

tests/test_cuda_bd_exchange_mover.py
```diff
@@ -16,10 +18,77 @@
...
+        for i, mol_idxs in enumerate(ref_bdem.water_idxs_np):
+            if not np.all(x_move[mol_idxs] == last_conf[mol_idxs]):
+                num_moved += 1
+                new_pos = x_move[mol_idxs]
+                idx = i
+        if num_moved > 0:
+            accepted += 1
+        if num_moved == 1:
+            # The molecules should all be imaged in the home box
+            np.testing.assert_allclose(image_frame(mol_groups, x_move, x_box), x_move)
```
badisa, line 56: Improves the ability to test by at least verifying that when a single move was accepted within a series of proposals that the reference platform says that the move has a non-zero probability. This was how I caught the lack of W support causing issues.

tests/test_cuda_bd_exchange_mover.py
```diff
@@ -132,46 +201,17 @@ def test_pair_of_waters_in_box(moves, precision, rtol, atol, seed):
...
+        group_idxs, bdem, ref_bdem, conf, box, proposals_per_move * moves, proposals_per_move, rtol, atol
+    )
 
 
 @pytest.mark.parametrize("precision", [np.float32])
 @pytest.mark.parametrize("seed", [2023])
 def test_bias_deletion_bulk_water_with_context(precision, seed):
     ff = Forcefield.load_default()
     system, conf, box, _ = builders.build_water_system(4.0, ff.water_ff)
+    box += np.diag([0.1, 0.1, 0.1])
```
badisa, line 214: A lot of the systems weren't actually stable due to not adding the box padding. Added force assertions to ensure the system is reasonable.

tests/test_md.py
```diff
@@ -504,16 +504,12 @@ def test_get_movers():
     friction = 0.0
     seed = 2022
 
-    unbound_potentials, sys_params, masses, coords, box = get_solvent_phase_system(mol, ff, 0.0, minimize_energy=True)
+    unbound_potentials, sys_params, masses, coords, box = get_solvent_phase_system(mol, ff, 0.0, minimize_energy=False)
```
badisa, line 507: Speed up the test, it only verifies that the movers can be retrieved, no reason to minimize or execute the potentials.

tests/test_cuda_bd_exchange_mover.py
proteneer: If `num_moves > 1` then we don't run these tests?

proteneer: Hm that seems non-ideal, since a test can just by pure chance skip this block (if it happened to accept multiple moves).

tests/test_cuda_bd_exchange_mover.py
```diff
@@ -16,10 +18,77 @@
...
+                idx = i
+        if num_moved > 0:
+            accepted += 1
+        if num_moved == 1:
+            # The molecules should all be imaged in the home box
+            np.testing.assert_allclose(image_frame(mol_groups, x_move, x_box), x_move)
+            # Verify that the probabilities and per mol energies agree when we do accept moves
+            # can only be done when we only attempt a single move per step
+            before_log_weights = ref_bdem.batch_log_weights(last_conf, box)
+            after_log_weights, tested = ref_bdem.batch_log_weights_incremental(
```
proteneer, line 60: we could also use the non-incremental version to compute `after_log_weights` as well? (in the case where there are multiple moves accepted).

tests/test_cuda_bd_exchange_mover.py
```diff
@@ -16,10 +18,77 @@
...
+                last_conf, x_box, idx, new_pos, before_log_weights
+            )
+            np.testing.assert_array_equal(tested, x_move)
```
proteneer, line 63: yeah, testing the conformation will be tricky if `num_moves > 1`

tests/test_cuda_bd_exchange_mover.py
proteneer: Hm thinking about this some more - does the ordering matter here? I.e. suppose have 5 waters, and we waters 2 and 4 are moved. If we simulate moving 2 then 4 vs moving 4 then 2, shouldn't the final log weights be path independent?

badisa: The final log weights will be path independent, so as you said in https://github.com/proteneer/timemachine/pull/1252#discussion_r1497905782 you could compute the after log weights and the probability should still be > 0.0.

Not sure this is true of the targeted case where the probability is impacted by the number of waters in each region

tests/test_cuda_bd_exchange_mover.py
proteneer, line 63: (Deleted the old comment) - I'm not sure actually why `batch_log_weights_incremental` returns the a `new_conf` in addition to the weights, since it will always be simply `conf.at[water_idxs[idx]].set(new_pos)` - we can simply test that the *final* position in an arbitrary sequence of moves is the same.

tests/test_cuda_bd_exchange_mover.py
```diff
@@ -328,70 +375,24 @@ def test_moves_in_a_water_box(steps_per_move, moves, box_size, precision, rtol,
...
-    if steps_per_move == 1:
-        np.testing.assert_allclose(bdem.acceptance_fraction(), accepted / moves)
-        assert bdem.n_accepted() == accepted
-    else:
-        assert bdem.n_accepted() >= accepted
-        assert bdem.acceptance_fraction() >= accepted / moves
+    verify_bias_deletion_moves(
+        group_idxs, bdem, ref_bdem, conf, box, total_num_proposals, num_proposals_per_move, rtol, atol
+    )
+    if total_num_proposals >= 10_000:
```
badisa, line 385: Realized that this value of 1 move per 1000 proposals is way off. Using the bulk mc example, we load a water box with an empty space in the middle. And if you don't equilibrate it enough (>200k steps) you are left with a void which allows more moves to be accepted then you would expect in bulk. Looks to be closer to 1 accepted move per 7.5k proposals

tests/test_cuda_bd_exchange_mover.py
badisa: Added in https://github.com/proteneer/timemachine/pull/1252/commits/b68262103925a94a51da818f04599d01ad5a1b4b

badisa (approved):

PR #1254, opened 2024-02-09
Add support for specifying an initial mapping

proteneer:
This PR adds support for pre-specifying an initial mapping (that can be generated using another cheaper method). This *enforces* the initial mapping to be exact (unlike specifying via `priority_idxs` which allows for `UNMAPPED` and skipping the node).


badisa:
Been thinking about this PR a lot recently, seems super useful. 

One point I had brought up online was that if the initial mapping is generated with different parameters (say allowing planar torsions to be mapped) the initial mapping won't be rejected even if it would fail to be generated using the new parameters. Don't think this is an issue, but does add a bit of nuance if we start using this heavily.

timemachine/fe/atom_mapping.py
```diff
@@ -74,10 +75,14 @@ def get_cores_and_diagnostics(
...
         disallow_planar_torsion_flips=disallow_planar_torsion_flips,
         min_threshold=min_threshold,
+        initial_mapping=initial_mapping,
     )
 
     # we require that mol_a.GetNumAtoms() <= mol_b.GetNumAtoms()
     if mol_a.GetNumAtoms() > mol_b.GetNumAtoms():
+        # adjust initial_mapping if we're swapping
+        if initial_mapping is not None and len(initial_mapping) > 0:
+            core_kwargs["initial_mapping"] = np.stack([initial_mapping[:, 1], initial_mapping[:, 0]], axis=1)
```
mcwitt, line 85: nit:
```suggestion
            core_kwargs["initial_mapping"] = initial_mapping[:, ::-1]
```

tests/test_atom_mapping.py
```diff
@@ -953,3 +965,190 @@ def test_get_cores_and_diagnostics():
...
+
+            cores_hybrid, diagnostics_hybrid = atom_mapping.get_cores_and_diagnostics(
+                mol_a_with_h, mol_b_with_h, **MAPPING_KWARGS_WITH_MAPPING
+            )
+
+            tnv_hybrid = diagnostics_hybrid.total_nodes_visited
+            core_hybrid = cores_hybrid[0]
+
+            print("TNV: all_hs, no_hs, hybrid", tnv_h, tnv_no_h, tnv_no_h + tnv_hybrid)
+            print("CORE SIZE: all_hs, no_hs, hybrid", len(core_h), len(core_no_h), len(core_hybrid))
```
mcwitt, line 1103: This generates a lot of lines when running tests. Should it be commented out or behind a debug flag?

timemachine/fe/mcgregor.py
```diff
@@ -243,39 +243,48 @@ def mcs(
...
+    max_threshold = _arcs_left(base_marcs)
     total_nodes_visited = 0
     for idx in range(max_threshold):
         cur_threshold = max_threshold - idx
         if cur_threshold < min_threshold:
             raise NoMappingError(f"Unable to find mapping with at least {min_threshold} edges")
-        map_a_to_b = [UNMAPPED] * n_a
-        map_b_to_a = [UNMAPPED] * n_b
+        map_a_to_b = copy.deepcopy(base_map_a_to_b)
+        map_b_to_a = copy.deepcopy(base_map_b_to_a)
```
mcwitt, line 279: I was surprised that I could remove the `copy.deepcopy` from these lines and the tests all still passed. Does this indicate useful additional test cases, or that the copy is somehow not needed?

mcwitt: Also, I don't think `deepcopy` should be needed here, since I think these are lists of `int`s?

tests/test_atom_mapping.py
```diff
@@ -953,3 +965,190 @@ def test_get_cores_and_diagnostics():
...
+    for old, new in old_to_new_mapping.items():
+        new_to_old_mapping[new] = old
+
+    return new_to_old_mapping
+
+
+import copy
+
+
+def test_truncated_mols():
```
mcwitt, line 1046: Would be useful to have a brief description of what this is testing

tests/test_atom_mapping.py
```diff
@@ -953,3 +965,190 @@ def test_get_cores_and_diagnostics():
         assert (
             diagnostics.total_nodes_visited >= diagnostics.core_size
         )  # must visit at least one node per atom pair in core
+
```
mcwitt, line 968: A few more basic tests seem useful here (unless I'm missing something), e.g.

1. do we get the same result with `initial_mapping=None`, `initial_mapping=[]`, etc.?
2. given an arbitrary initial mapping, does the result always contain the initial mapping?
3. if we run without an initial mapping and then again passing a subset of the generated mapping as the initial mapping, is the resulting mapping equivalent to the first one we generated?
4. if the initial mapping isn't consistent with the specified filters, do we generate an error?

badisa: Added tests in https://github.com/proteneer/timemachine/pull/1254/commits/c8bf1aa22edbe57e00bb9b3a58d5ad59bc0cf137

4. has some nuance, depending on the parameter then either no core can be found and returns the original core or an exception is raised. Could throw an exception if the core returned is the identity map provided, which seems like it might be cleaner.

tests/test_atom_mapping.py
badisa: Commented out in https://github.com/proteneer/timemachine/pull/1254/commits/c8bf1aa22edbe57e00bb9b3a58d5ad59bc0cf137

tests/test_atom_mapping.py
badisa: Thinking about it some more, don't want to throw an exception if the core returned is the identity map. That would make this painful to use in practice, where the new parameters may not generate a larger core.

timemachine/fe/mcgregor.py
badisa: Per your comment about the being able to remove the copy not resulting in any copy, it is because the final map returned is the initial `base_map_a_to_b`. After we update the index, we pop it. https://github.com/proteneer/timemachine/blob/master/timemachine/fe/mcgregor.py#L421-L446

Verified this by adding in an assertion that the `map_a_to_b` at the end of recursion is identical to `base_map_a_to_b` and b_to_a. 

I would probably advocate for keeping the copy just to be super safe, but does not appear strictly neccessary

tests/test_atom_mapping.py
```diff
@@ -953,3 +971,281 @@ def test_get_cores_and_diagnostics():
...
+    assert len(cores) > 1
+    assert len(identity_core) == 1
+    # Core ordering does get shuffled, but pairs should be identical
+    assert set(tuple(pair) for pair in identity_core[0]) == set(tuple(pair) for pair in cores[0])
+
+
+def test_initial_mapping_always_a_subset_of_cores(hif2a_ligands):
+    rng = np.random.default_rng(2024)
+    mols = hif2a_ligands
+    mol_a, mol_b = rng.choice(mols, 2)
```
maxentile, line 1090: testing observation: these tests are each currently parameterized by a list of ligands `hif2a_ligands`, but run a single random pair from the list.

consider either:
* parameterizing by `(ligands, num_random_pairs=1)`
* parameterizing by `mol_pair`

tests/test_atom_mapping.py
```diff
@@ -953,3 +971,281 @@ def test_get_cores_and_diagnostics():
...
+        assert len(new_cores) == 1
+    else:
+        # If there is no connected core that can be made from the disconnected will raise an exception
+        with pytest.raises(NoMappingError):
+            atom_mapping.get_cores(mol_a, mol_b, **initial_map_kwargs)
+
+
+def test_hybrid_core_generation(hif2a_ligands):
+    """
+    Verify that generating"""
```
maxentile, line 1138: possibly remove docstring

tests/test_atom_mapping.py
```diff
@@ -953,3 +971,281 @@ def test_get_cores_and_diagnostics():
...
+            # hybrid_visits.append(tnv_no_h + tnv_hybrid)
+
+            # with_h_core_sizes.append(len(core_h))
+            # without_h_core_sizes.append(len(core_no_h))
+            # hybrid_core_sizes.append(len(core_hybrid))
+
+            assert len(core_no_h) < len(core_h), f"Mol {i} -> {j} failed to produce larger mapping by adding hydrogens"
+            assert len(cores_no_h) < len(
+                core_hybrid
+            ), f"Mol {i} -> {j} failed to produce larger mapping by running hybrid"
```
maxentile, line 1208: uncertainty: would we also expect `len(core_hybrid) == len(core_h)`?

timemachine/fe/utils.py
```diff
@@ -460,9 +460,9 @@ def sanitize_energies(full_us, lamb_idx, cutoff=10000):
     return np.where(abs_us < cutoff, full_us, np.inf)
 
 
-def read_sdf(fname: Union[str, Path]) -> List[Chem.Mol]:
+def read_sdf(fname: Union[str, Path], removeHs=False) -> List[Chem.Mol]:
     """Read list of mols from an SDF (without discarding hydrogens!)"""
-    supplier = Chem.SDMolSupplier(str(fname), removeHs=False)
+    supplier = Chem.SDMolSupplier(str(fname), removeHs=removeHs)
```
maxentile, line 465: hah, code change invalidates docstring --> remove docstring

tests/test_atom_mapping.py
```diff
@@ -953,3 +971,281 @@ def test_get_cores_and_diagnostics():
...
+
+    initial_map_kwargs = DEFAULT_ATOM_MAPPING_KWARGS.copy()
+    initial_map_kwargs["initial_mapping"] = cores_without_ring_match_ring[0]
+
+    if param_to_change == "ring_matches_ring_only":
+        # If we remap with this core that is invalid under the mapping conditions, return the original core
+        new_cores = atom_mapping.get_cores(mol_a, mol_b, **initial_map_kwargs)
+        assert len(new_cores) == 1
+    else:
+        # If there is no connected core that can be made from the disconnected will raise an exception
```
maxentile, line 1131:
```suggestion
    elif param_to_change == "connected_core":
        # If there is no connected core that can be made from the disconnected core, expect NoMappingError
```
...
```python
        # TODO: possibly test "enforce_chiral", "enforce_core_core", "disallow_planar_torsion_flips", and other filters
```

tests/test_atom_mapping.py
```diff
@@ -953,3 +971,281 @@ def test_get_cores_and_diagnostics():
...
+    assert DEFAULT_ATOM_MAPPING_KWARGS[param_to_change]
+
+    kwargs = DEFAULT_ATOM_MAPPING_KWARGS.copy()
+    kwargs[param_to_change] = False
+    cores_without_ring_match_ring = atom_mapping.get_cores(mol_a, mol_b, **kwargs)
+    # The core without the ring matches ring is larger
+    assert len(cores[0]) < len(cores_without_ring_match_ring[0])
+
+    initial_map_kwargs = DEFAULT_ATOM_MAPPING_KWARGS.copy()
+    initial_map_kwargs["initial_mapping"] = cores_without_ring_match_ring[0]
```
maxentile, line 1124: preference for less specific variable name, since the content of this variable only matches the name when `"param_to_change" == "ring_matches_ring_only"`

(out of context, a later branch may look like it only tests the "ring_matches_ring_only" path, but the implementation does test both :`cores_without_ring_match_ring` can either be a list of cores without  enforcing "connected_core" or without enforcing "ring_matches_ring_only")
```suggestion
    unfiltered_cores = atom_mapping.get_cores(mol_a, mol_b, **kwargs)
    # The core without the filter `param_to_change` is larger
    assert len(cores[0]) < len(unfiltered_cores[0])

    initial_map_kwargs = DEFAULT_ATOM_MAPPING_KWARGS.copy()
    initial_map_kwargs["initial_mapping"] = unfiltered_cores[0]
```

maxentile (approved):

tests/test_atom_mapping.py
badisa: Woops starting writing this docstring then left it lo. Will fix it up.

tests/test_atom_mapping.py
badisa: It does not appear to be the case, will take a look at why that is.

mcwitt (approved):

tests/test_atom_mapping.py
badisa: Found a pretty clear case which is that if you have several terminal atoms that are not hydrogens being transformed into hydrogens, you may build an initial map using different atoms than what the hybrid based on the hydrogen-less molecule would produce.
In the picture below the top is the full hydrogens and the bottom is the hybrid. You can see that the hybrid has the flourine mapped to another carbon which prevents the retrieval of the same size core. 
hybrid-vs-non

timemachine/fe/utils.py
badisa: Fixed docstring in https://github.com/proteneer/timemachine/pull/1254/commits/c4f70085c4cc5b664b454302fbd44df5215131b5

tests/test_atom_mapping.py
maxentile: mostly addressed in https://github.com/proteneer/timemachine/pull/1254/commits/c35b6841c28e1d2619f175e3a38521ceed5328b6 

but may still want to add a todo to expand test coverage (e.g. testing each of the other filter options)

badisa: Expanded in https://github.com/proteneer/timemachine/pull/1254/commits/813d548aaaef2bc451a4163f36a6b5446e24dd35

tests/test_atom_mapping.py
badisa: Completed writing it in https://github.com/proteneer/timemachine/pull/1254/commits/8be9c2143a0a11522de1bbafbfb1b603d8b71390

tests/test_atom_mapping.py
badisa: Parameterized the pairs in https://github.com/proteneer/timemachine/pull/1254/commits/78b1b35514448b80f7043dd2a170654b885eb0df

PR #1255, opened 2024-02-12
Improvements for chiral restraints and various bugfixes

proteneer:
This PR addresses #1243 to more robustly handle the initialization and interpolation of angle terms in dummy atoms. Three big changes:

1) If a dummy angle term is involved in a chiral restraint, it is weakened to a force constant of 10 kJ/mol at the dummy end-state
2) Chiral restraints are turned on, log linearly, from `lambda=0.0 to 0.2`
3) The angle terms are staggered to be turned on from `lambda=0.2 to 0.7` as opposed to `lambda=0.0 to lambda=0.7`
4) Stagger bond/angle/chiral_atoms/torsions to be turned on properly

Misc changes:
1) Fixed a plotting bug in `detail_overlap` where the first window was shoved to the end and mislabelled as `-1 -> 0`. (`plot_axis` was off by 1)
2) Improve histogram range and bins for the `detail_overlap`
3) Fix dummy group improper torsion bug, no idea how this code ever worked

TBD:

1) To address the [Maybe] above, investigate how strong a chiral restraint has to be to flip over a mis-initialized geometry.
2) Decide if we should leave chiral restraints that are defined purely over dummy atoms to always be turned on. Right now all dummy atom chiral restraints are turned off.

Todo:
- tests?

### DIFF
--- a/tests/test_chiral_restraints.py
+++ b/tests/test_chiral_restraints.py
@@ -403,15 +403,14 @@ $$$$""",
     U_fn = system.get_U_fn()
 
     vols_orig = []
-    frames = simulate_system(U_fn, x0, num_samples=4000)
+
+    # turn off minimization for this case to avoid inversions
+    frames = simulate_system(U_fn, x0, num_samples=4000, minimize=False)
     for f in frames:
-        print("f0", f)
         vol_list = []
         for p in normal_restr_idxs:
             vol_list.append(pyramidal_volume(*f[p]))
         vols_orig.append(vol_list)
-
-    assert 0
 
     def U_total(x):
         nrgs = [U_fn(x)]
@@ -423,7 +422,6 @@ $$$$""",
     vols_chiral = []
     frames = simulate_system(U_total, x0, num_samples=4000)
     for f in frames:
-        print("f1", f)
         vol_list = []
         for p in normal_restr_idxs:
             vol_list.append(pyramidal_volume(*f[p]))
@@ -447,7 +445,6 @@ $$$$""",
     ref_dist = np.array([x for x in vols_orig if x < 0])
     test_dist = np.array([x for x in vols_chiral if x > 0])
     # should be indistinguishable under KS-test
-    print("REF_DIST, TEST_DIST", ref_dist, test_dist)
 
     ks, pv = scipy.stats.ks_2samp(-ref_dist, test_dist)
     assert ks < 0.05 or pv > 0.10
--- a/tests/test_chiral_restraints.py
+++ b/tests/test_chiral_restraints.py
@@ -265,6 +265,9 @@ $$$$""",
     )
 
     ff = Forcefield.load_from_file("smirnoff_1_1_0_sc.py")
+    # pessimize: strengthen harmonic angle force constant to max
+    ff.ha_handle.params[:, 0] = np.max(ff.ha_handle.params[:, 0])
+
     s_top = topology.BaseTopology(mol, ff)
     x0 = utils.get_romol_conf(mol)
 
--- a/tests/test_chiral_restraints.py
+++ b/tests/test_chiral_restraints.py
@@ -398,16 +398,20 @@ $$$$""",
     assert np.all(np.asarray(U_chiral_atom_batch(x0, normal_restr_idxs, kc)) == 0)
     assert np.all(np.asarray(U_chiral_atom_batch(x0, inverted_restr_idxs, kc)) > 0)
 
+    # ugh basin-hopping is inverting the non-chiral endstates too, sigh
     system = s_top.setup_end_state()
     U_fn = system.get_U_fn()
 
     vols_orig = []
     frames = simulate_system(U_fn, x0, num_samples=4000)
     for f in frames:
+        print("f0", f)
         vol_list = []
         for p in normal_restr_idxs:
             vol_list.append(pyramidal_volume(*f[p]))
         vols_orig.append(vol_list)
+
+    assert 0
 
     def U_total(x):
         nrgs = [U_fn(x)]
@@ -419,6 +423,7 @@ $$$$""",
     vols_chiral = []
     frames = simulate_system(U_total, x0, num_samples=4000)
     for f in frames:
+        print("f1", f)
         vol_list = []
         for p in normal_restr_idxs:
             vol_list.append(pyramidal_volume(*f[p]))
@@ -442,6 +447,8 @@ $$$$""",
     ref_dist = np.array([x for x in vols_orig if x < 0])
     test_dist = np.array([x for x in vols_chiral if x > 0])
     # should be indistinguishable under KS-test
+    print("REF_DIST, TEST_DIST", ref_dist, test_dist)
+
     ks, pv = scipy.stats.ks_2samp(-ref_dist, test_dist)
     assert ks < 0.05 or pv > 0.10
 
@@ -824,6 +831,9 @@ $$$$""",
             (3, 0, 4),  # nbr_anchor - anchor - dummy
         ]
     )
+
+    # test that we weaken the angle parameters for angle idxs implied
+    # by chiral volumes that are being turned on/off between the end-states
     for idxs, params in zip(vacuum_system.angle.potential.idxs, vacuum_system.angle.params):
         k_angle = params[0]
         if tuple(idxs) in weakened_angles:
@@ -951,13 +961,13 @@ $$$$""",
     # CF3 does *not* turn off any angle idxs associated with the chiral idxs, the chiral volume
     # implied by HCF2 does turn off the angle idxs (which overlap with the the above)
     vacuum_system = st.setup_intermediate_state(0)
-    for _, params in zip(vacuum_system.angle.potential.idxs, vacuum_system.angle.params):
+    for params in vacuum_system.angle.params:
         k_angle = params[0]
         assert k_angle <= MINIMUM_CHIRAL_ANGLE_FORCE_CONSTANT
 
     # test the lambda = 1 end state with fully interacting atoms
     vacuum_system = st.setup_intermediate_state(1)
-    for _, params in zip(vacuum_system.angle.potential.idxs, vacuum_system.angle.params):
+    for params in vacuum_system.angle.params:
         k_angle = params[0]
         assert k_angle > MINIMUM_CHIRAL_ANGLE_FORCE_CONSTANT
 
--- a/tests/test_chiral_restraints.py
+++ b/tests/test_chiral_restraints.py
@@ -821,7 +821,7 @@ $$$$""",
         [
             (1, 0, 4),  # nbr_anchor - anchor - dummy
             (2, 0, 4),  # nbr_anchor - anchor - dummy
-            (3, 0, 4),  # core - anchor - dummy
+            (3, 0, 4),  # nbr_anchor - anchor - dummy
         ]
     )
     for idxs, params in zip(vacuum_system.angle.potential.idxs, vacuum_system.angle.params):
--- a/tests/test_chiral_restraints.py
+++ b/tests/test_chiral_restraints.py
@@ -907,7 +907,7 @@ $$$$""",
 
 
 def test_converting_3_dummy_atoms():
-    # convert H-H into H-CF3, Flourines are markers for dummy
+    # convert N#N into H3C-CHF3, Flourines are markers for dummy
     # test that we're weakening the angle terms correctly at the end-states
     ff = Forcefield.load_default()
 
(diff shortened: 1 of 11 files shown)


proteneer:
(requested a preliminary review before adding more extensive tests)

timemachine/fe/single_topology.py
```diff
@@ -1112,6 +1140,8 @@ def setup_intermediate_state(self, lamb) -> VacuumSystem:
             ),
         )
 
+        START_ANGLE_MIN = 0.2
```
maxentile, line 1143: Point of uncertainty: does this interact with any other terms?

(Interaction with chiral atom restraints looks correct.)

maxentile (approved):
I don't have a good handle on the problem yet, but the implementation looks like a good solution to a problem of the form "chiral atom restraints may not win if in conflict with harmonic angle terms", and an improvement relative to linear interpolation. (+ misc. other improvements that would hamper diagnosis of this issue)

(However, I cannot be confident that "bad interactions between angle terms and chiral volume terms" is an exhaustive list of all ways that SingleTopology-interpolation between arbitrary pairs of chiral molecules can go wrong.)

proteneer:
> "chiral atom restraints may not win if in conflict with harmonic angle terms"

To be clear, I'm not actually sure if our current force constants, when fully turned on, will win over the angle terms. The problem is slightly more subtle:

In `master`, for the CH3 moeity in the test case, at the lambda=0 state, the harmonic angle terms are fully turned on, and the chiral volumes are fully turned off. The initialization of the hydrogens surrounding the CH3 is arbitrary in the absence of chiral restraints, when the angle terms are fully turned on, we cannot sample between the different chiral states.

> "bad interactions between angle terms and chiral volume terms" 

I don't *think* there are bad interactions between angle terms and chiral volume terms that *are* fully turned on. But again, I need to verify that chiral volumes, parameterized at `k=1000` can always win over angle terms under typical MD settings (for vacuum), and win over intermolecular nonbonded terms (for complex/solvent).

proteneer:
Some more things I want to address:

Leave on more chiral restraint terms between dummy atoms, and dummy-core atoms that do not affect factorizability. I think it's always safe to leave a chiral restraint term turned on if it involves *exactly* a single anchoring core atom. Not sure about the other situations.

timemachine/fe/single_topology.py
```diff
@@ -1145,12 +1175,27 @@ def setup_intermediate_state(self, lamb) -> VacuumSystem:
 
         assert src_system.chiral_atom
         assert dst_system.chiral_atom
+
+        # turn on chiral atom restraints before scaling up the angle terms.
+        def rescaled_log_linear_interpolation(src_params, dst_params, lamb):
+            if lamb > START_ANGLE_MIN:
+                return dst_params
+            else:
+                return interpolate.log_linear_interpolation(src_params, dst_params, lamb / START_ANGLE_MIN, 1e-6)
```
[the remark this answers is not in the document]
mcwitt: This scheme doesn't seem symmetric under interchange of A and B, $\lambda \to 1 - \lambda$. Is this intentional? (Note that `interpolate_harmonic_force_constant` maintains symmetry by applying $\lambda \to 1 - \lambda$ in the case `dst_k == 0`).

timemachine/fe/interpolate.py
```diff
@@ -162,6 +162,7 @@ def log_linear_interpolation(src_params, dst_params, lamb, min_value):
     src_params = jnp.maximum(src_params, min_value)
     dst_params = jnp.maximum(dst_params, min_value)
 
+    # tbd: handle 0s better if lamb = 0 and src_params == 0
```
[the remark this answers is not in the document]
proteneer: if `src_params == 0 and lamb == 0`, then technically we have: `exp(log(0)) == 0`

timemachine/fe/plots.py
```diff
@@ -161,7 +166,7 @@ def plot_overlap_detail_figure(
 
         # loop over bond, angle, torsion, nonbonded terms etc.
         for u_idx in range(num_energy_components):
-            plot_axis = all_axes[lamb_idx - 1][u_idx]
+            plot_axis = all_axes[lamb_idx][u_idx]
```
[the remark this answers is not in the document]
proteneer: clarified offline, I think the last row should be the first row, the plots were just all shifted by 1 row.

timemachine/fe/interpolate.py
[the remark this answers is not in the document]
mcwitt: hmm, but isn't this handled by `src_params = jnp.maximum(src_params, min_value)` on L162?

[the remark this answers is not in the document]
proteneer: if `min_value==0` does it get handled?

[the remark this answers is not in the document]
proteneer: (sorry was playing with the setting of `min_value=0`, maybe that's not supported by this method).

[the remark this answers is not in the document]
mcwitt: > if min_value==0 does it get handled?

No, but the original intent of `min_value` was as a positive threshold. Maybe we should add an assertion that `min_value > 0`?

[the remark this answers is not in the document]
mcwitt: > (sorry was playing with the setting of `min_value=0`, maybe that's not supported by this method).

ah, I see - yeah, interpolation of functions with zero or negative values is not supported by this method. In principle it could be if we extended the functional form, e.g. to add an offset

timemachine/fe/single_topology.py
```diff
@@ -166,6 +178,19 @@ def setup_dummy_interactions(
...
+        center, i, j, k = idxs
+
+        case_1 = all(a in dummy_group for a in (center, i, j, k))
+        case_2 = (center == root_anchor_atom) and all(a in dummy_group for a in (i, j, k))
+        # TODO: maybe add more cases, if we convince ourselves they are also factorizable...
+        #    e.g. if center is root_anchor_atom, and at most one of (i, j, k) is neighbor_anchor_atom
+        #   (and maybe also if root_anchor_atom, neighbor_anchor_atom are swapped in above statement)
+        if case_1 or case_2:
+            dummy_chiral_atom_idxs.append(tuple(int(x) for x in idxs))
+            dummy_chiral_atom_params.append(params)
```
maxentile, line 193: @proteneer : I pushed a couple changes to your branch based on offline discussion.

(This now covers a couple cases that seem like safe additions, but this should be double-checked.)

tests/test_chiral_restraints.py
```diff
@@ -265,6 +265,9 @@ def test_chiral_restraints_tetrahedral():
     )
 
     ff = Forcefield.load_from_file("smirnoff_1_1_0_sc.py")
+    # pessimize: strengthen harmonic angle force constant to max
+    ff.ha_handle.params[:, 0] = np.max(ff.ha_handle.params[:, 0])
```
maxentile, line 269: >To address the [Maybe] above, investigate how strong a chiral restraint has to be to flip over a mis-initialized geometry.

This test currently addresses a similar question (but in isolation).

(Just pessimized it a little in terms of HarmonicAngle force constant, but this test does **not** yet check that chiral_atom_k is strong enough to invert in context of bulkier molecule or more complicated environment.)

timemachine/fe/single_topology.py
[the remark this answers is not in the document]
maxentile: ^ partially addressed in https://github.com/proteneer/timemachine/pull/1255/commits/1e6c4d1b0b1bf3d8cf254514c61ace3d427885dd

timemachine/fe/single_topology.py
```diff
@@ -286,17 +312,46 @@ def setup_end_state(ff, mol_a, mol_b, core, a_to_c, b_to_c):
...
+        mol_b_chiral_atom_angle_idxs.add(canonicalize_bond([i, center, j]))
+        mol_b_chiral_atom_angle_idxs.add(canonicalize_bond([i, center, k]))
+        mol_b_chiral_atom_angle_idxs.add(canonicalize_bond([j, center, k]))
+
     for anchor, (nbr, dg) in dgs.items():
         all_idxs, all_params = setup_dummy_interactions_from_ff(ff, mol_b, dg, anchor, nbr)
         all_dummy_bond_idxs.extend(all_idxs[0])
         all_dummy_angle_idxs.extend(all_idxs[1])
         all_dummy_improper_idxs.extend(all_idxs[2])
+        all_dummy_chiral_atom_idxs.extend(all_idxs[3])
```
proteneer, line 337: (Fix to only do for angles being turned off)

timemachine/fe/single_topology.py
proteneer: I did some back of envelop math and I'm fairly certain now that if exactly one atom is a root_anchor atom, and everything else is a dummy atom in the grpup, then it's also fine. (Since it will never affect the inter atomic distances of the core atoms, as whatever force you apply is only to a single core anchor atom). I'll modify the PR to address this.

timemachine/fe/single_topology.py
```diff
@@ -166,6 +178,19 @@ def setup_dummy_interactions(
         if all([a in dga for a in idxs]):
             dummy_improper_idxs.append(tuple([int(x) for x in idxs]))
             dummy_improper_params.append(params)
+    for idxs, params in zip(chiral_atom_idxs, chiral_atom_params):
+        # consider a few cases: (1) all idxs dummy, (2) a single root_anchor_atom and 3 dummies, (3) other
+        center, i, j, k = idxs
+        case_1 = all(a in dummy_group for a in (center, i, j, k))
+        three_in_dummy_group = sum(a in dummy_group for a in (center, i, j, k)) == 3
+        one_is_root_anchor_atom = sum(a == root_anchor_atom for a in (center, i, j, k)) == 1
```
proteneer, line 186: @maxentile  sanity check this?

timemachine/fe/single_topology.py
```diff
@@ -65,6 +65,9 @@ def setup_dummy_interactions_from_ff(ff, mol, dummy_group, root_anchor_atom, nbr
     bond_params, hb = top.parameterize_harmonic_bond(ff.hb_handle.params)
     angle_params, ha = top.parameterize_harmonic_angle(ff.ha_handle.params)
     improper_params, it = top.parameterize_improper_torsion(ff.it_handle.params)
+    chiral_atom_potential, chiral_bond_potential = top.setup_chiral_restraints(chiral_atom_k)
+    chiral_atom_idxs = chiral_atom_potential.potential.idxs
+    chiral_atom_params = chiral_bond_potential.params
```
maxentile, line 70: (recording from offline comments from @proteneer )

I introduced this typo in https://github.com/proteneer/timemachine/pull/1255/commits/e44f4e7788975d17e85494d3dd59c976b42c26b9 but didn't see it

```suggestion
    chiral_atom_params = chiral_atom_potential.params
```

timemachine/fe/single_topology.py
```diff
@@ -286,17 +312,49 @@ def setup_end_state(ff, mol_a, mol_b, core, a_to_c, b_to_c):
...
+            assert canonicalize_bond(angle_idxs) == angle_idxs
+            if angle_idxs in interpolating_mol_b_chiral_atom_angle_idxs:
+                modified_angle_params.append([MINIMUM_CHIRAL_ANGLE_FORCE_CONSTANT, angle_params[1]])
+            else:
+                modified_angle_params.append(angle_params)
+
+        modified_angle_params = np.array(modified_angle_params)
+        all_dummy_angle_params.extend(modified_angle_params)
         all_dummy_improper_params.extend(all_params[2])
+        all_dummy_chiral_atom_idxs.extend(all_params[3])
```
maxentile, line 357: another typo I introduced in https://github.com/proteneer/timemachine/pull/1255/commits/e44f4e7788975d17e85494d3dd59c976b42c26b9
```suggestion
        all_dummy_chiral_atom_params.extend(all_params[3])
```

timemachine/fe/single_topology.py
maxentile: fixed in https://github.com/proteneer/timemachine/pull/1255/commits/efabaa21dc081104b0bddb4cb201cb18b0a0a47b

proteneer:
Fixed a ton of random bugs
Reworked the logic for rescaling dummy group angles
Added some simple tests

timemachine/fe/single_topology.py
```diff
@@ -65,14 +65,19 @@ def setup_dummy_interactions_from_ff(ff, mol, dummy_group, root_anchor_atom, nbr
...
+    chiral_atom_params = chiral_atom_potential.params
 
     return setup_dummy_interactions(
         hb.idxs,
         bond_params,
         ha.idxs,
         angle_params,
-        improper_params,
         it.idxs,
+        improper_params,
```
proteneer, line 78: @badisa how the fuck did this ever work?

timemachine/fe/single_topology.py
maxentile: Looks safe: there's no way for a potential function of this form to influence the distribution of `root_anchor_atom`.

To test the TODO cases below, I'll do a little numerical check.

(Define 4-atom system. Fix `root_anchor_atom` at origin. Add some arbitrary (not necessarily radially symmetric) energy function in terms of position of `neighbor_anchor_atom`. Add the 4-body energy term to be tested. Assert distribution of `neighbor_anchor_atom`'s position is unaffected by changing placement of remaining two atoms.)

timemachine/fe/single_topology.py
```diff
@@ -250,6 +273,15 @@ def get_num_connected_components(num_atoms: int, bonds: Collection[Tuple[int, in
     return len(list(nx.connected_components(g)))
 
 
+MINIMUM_CHIRAL_ANGLE_FORCE_CONSTANT = 10.0
+
+
+def canonicalize_chiral_atom_idxs(idxs):
+    center, i, j, k = idxs
+    results = [(center, i, j, k), (center, j, k, i), (center, k, i, j)]
+    return sorted(results)[0]
```
maxentile, line 282: nit: can retain simpler expression, and possibly replace one more usage in this file https://github.com/proteneer/timemachine/pull/1255/files#diff-0b9bd14a098c7a129a18f14ffb94d84fc8e5c1020d1a55d0e6f53d545c1a4084L383-R456

```suggestion
    return min(results)
```

timemachine/fe/single_topology.py
```diff
@@ -330,13 +366,44 @@ def setup_end_state(ff, mol_a, mol_b, core, a_to_c, b_to_c):
...
+    all_angles_implied_by_turned_off_chiral_atom_idxs = set()
+    for dummy_center, dummy_i, dummy_j, dummy_k in chiral_atom_idxs_turned_off_at_end_state:
+        all_angles_implied_by_turned_off_chiral_atom_idxs.add(canonicalize_bond([dummy_i, dummy_center, dummy_j]))
+        all_angles_implied_by_turned_off_chiral_atom_idxs.add(canonicalize_bond([dummy_i, dummy_center, dummy_k]))
+        all_angles_implied_by_turned_off_chiral_atom_idxs.add(canonicalize_bond([dummy_j, dummy_center, dummy_k]))
+
+    all_dummy_angle_params_modified = []
+    for angle_idxs, angle_params in zip(all_dummy_angle_idxs, all_dummy_angle_params):
+        if angle_idxs in all_angles_implied_by_turned_off_chiral_atom_idxs:
+            all_dummy_angle_params_modified.append([MINIMUM_CHIRAL_ANGLE_FORCE_CONSTANT, angle_params[1]])
```
maxentile, line 401: edit: looks fine

(initially misread as taking min of a force constant and an equilibrium angle)

maxentile (approved):

timemachine/fe/single_topology.py
badisa: This 'works' because we always drop the impropers on the floor (and it seems like we still do, or we just don't exercise this in our tests). https://github.com/proteneer/timemachine/blob/master/timemachine/fe/single_topology.py#L166 The float values will never pass this check, and even with the change it seems that the output for the improper torsions is an empty array. Do we have a test case where we expect improper torsion parameters for the dummy groups?

maxentile: Thanks for shedding light on that.

>Do we have a test case where we expect improper torsion parameters for the dummy groups?

Should be able to cook one up, will look into it. I guess this should be triggered when any of these patterns are matched within a single dummy group:
https://github.com/proteneer/timemachine/blob/0b48a62def3b89712647be9e6132d434904d9c2f/timemachine/ff/params/smirnoff_2_1_0_ccc.py#L415-L421

badisa: Sorry jumped the gun saying we don't test it (machine stopped responded and had to restart).     `test_hif2a_end_state_stability` does exercise this code and constructs dummy idxs for improper torsions. We should probably run the nightly tests for this PR before merging.

maxentile: >Should be able to cook one up, will look into it.

A small test case

```python
import numpy as np

from rdkit import Chem

from timemachine.fe.single_topology import SingleTopology

# mol that contains a match for any smarts in default ImproperTorsionHandler
smi = 'C=NC'
mol_a = Chem.AddHs(Chem.MolFromSmiles(smi))
AllChem.EmbedMolecule(mol_a)
mol_b = Chem.AddHs(Chem.MolFromSmiles(smi))
AllChem.EmbedMolecule(mol_b)

# de-map atoms associated with improper torsion terms
_, it_idxs = ff.it_handle.parameterize(mol_a)
assert len(it_idxs) > 0
core_idxs = list(set(range(mol_a.GetNumAtoms())).difference(set(it_idxs.flatten())))
core = np.array(list(zip(core_idxs, core_idxs)))
assert len(core) > 0

# construct SingleTopology instance
```

timemachine/fe/free_energy.py
```diff
@@ -778,6 +778,7 @@ def compute_intermediate_result(lambdas: Sequence[float]) -> PairBarResult:
             lamb1 = lambdas[left_idx]
             lamb2 = lambdas[left_idx + 1]
             print(
+                f"Bisection iteration {iteration}: "
                 f"Current minimum BAR overlap {cost_to_overlap(max(costs)):.3g} "
```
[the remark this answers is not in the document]
proteneer: will fix inline

timemachine/fe/plots.py
```diff
@@ -132,7 +136,8 @@ def plot_overlap_detail_figure(
     num_rows = len(u_kln_by_component_by_lambda)  # L - 1 adjacent pairs
```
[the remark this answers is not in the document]
proteneer: will fix inline

timemachine/fe/single_topology.py
proteneer: good catch will fix

timemachine/fe/free_energy.py
```diff
@@ -778,7 +778,8 @@ def compute_intermediate_result(lambdas: Sequence[float]) -> PairBarResult:
             lamb1 = lambdas[left_idx]
             lamb2 = lambdas[left_idx + 1]
             print(
-                f"Current minimum BAR overlap {cost_to_overlap(max(costs)):.3g} "
+                f"Bisection iteration {iteration} (of {n_bisections}): "
+                f"Current minimum BAR overlap {cost_to_overlap(max(costs)):.3g} <= {min_overlap:.3g}"
```
badisa, line 782: Need a space in there to avoid having `0.667between`

```suggestion
                f"Current minimum BAR overlap {cost_to_overlap(max(costs)):.3g} <= {min_overlap:.3g} "
```

PR #1259, opened 2024-02-20
Remove pickling of initial states in bisection

badisa:
* Introduced in https://github.com/proteneer/timemachine/pull/1255 and guessing it was intended only for debugging


proteneer (approved):
Good catch - oopsie on my end!

PR #1260, opened 2024-02-21
Generate HREX Swap moves in a single pass

badisa:
* Seems like this makes a bit of a difference

Reference Profile
ref

With batching
profile

2% vs 5% spent in `attempt_neighbor_moves` thanks to reduced calls to numpy choice

#### Time by number of windows
image

#### Speed up by number of windows
image

#### Time by number of windows
image

#### Speed up by number of windows

image


timemachine/md/moves.py
```diff
@@ -95,6 +95,20 @@ def move(self, x: _State) -> _State:
         return x
 
 
+class BatchedMixtureOfMoves(MixtureOfMoves[_State]):
```
badisa, line 98: This seems like a very simple win, though perhaps there is some nuance I am missing. Seems to make a pretty significant improvement in vacuum, especially with larger number of windows

proteneer:
nice! add a test?

badisa:
> nice! add a test?

Added test in https://github.com/proteneer/timemachine/pull/1260/commits/48ee1154b0a1f097c257d18882b005dbab30223d

@maxentile Would you take a look at it? With @mcwitt out you would probably be the best person to make sure I haven't borked this

timemachine/md/moves.py
badisa: Any suggestions for a better name? It doesn't actually batch the mixture of moves, just generates the sequence in a batch.

timemachine/md/moves.py
```diff
@@ -95,6 +95,20 @@ def move(self, x: _State) -> _State:
...
+    """Apply a single move uniformly selected from a list"""
```
maxentile:
```suggestion
    """Repeatedly sample from a uniform mixture of moves"""
```
or similar. (To clarify that this is not "batching" over `len(mixture)` but over `num_iterations`.)

timemachine/md/moves.py
maxentile: Maybe this is cleaner as a composition of two independent concepts:
* `IteratedMove` (i.e. `for _ in range(num_iterations): x = move(x)`) 
* `MixtureMove` (i.e. `x = np.random.choice(moves)(x)`) 
?

timemachine/md/moves.py
badisa: Added in https://github.com/proteneer/timemachine/pull/1260/commits/ae4f61c02a02ee829ee11ae560e381c6f03478c7

badisa:
> Maybe this is cleaner as a composition of two independent concepts:
>    IteratedMove (i.e. for _ in range(num_iterations): x = move(x))
>   MixtureMove (i.e. x = np.random.choice(moves)(x))

Not clear on what you are suggesting exactly? That is kind of how it was before, it used the `MixtureOfMoves` move, but that results in calling `np.random.choice(48)` 48^3 times which becomes expensive in the case of vacuum sampling. The goal here is to call `np.random.choice` only a single time per swap proposal.

timemachine/md/moves.py
badisa: Not clear on what you are suggesting exactly? That is kind of how it was before (though IteratedMove was just a loop in HREX), it used the MixtureOfMoves move, but that results in calling `np.random.choice(48)` `48^3` times which becomes expensive. The goal here is to call np.random.choice only a single time per swap proposal.

maxentile: Ahh, thanks for clarifying! Sorry, I didn't catch the full PR context before commenting. This makes sense as you've written it, and the suggestion is not applicable.

The moves themselves are not batched, but the random selections of which move to apply at each iteration are batched. In the case where the cost of applying each move is comparable to a call to `np.random.choice()`, this yields a speed-up.

(Was thinking of a prior-work-inspired water-sampling context, where the moves themselves might be batched as a form of "speculative execution" -- yielding a speed-up when 1 / acceptance fraction is large relative to the parallel width (i.e. when move(x) == x a lot of the time). This form of "batching" would be independent of whether `move` is a `MixtureOfMoves` or not.)

proteneer (approved):

proteneer:
Interestingly enough, looking at the profiler, it looks like `solve_mbar_once` is the next major bottleneck? (esp. for vacuum simulations)

timemachine/md/moves.py
badisa: Added a comment to try and clarify the intention of this specialized class https://github.com/proteneer/timemachine/pull/1260/commits/5b4ca483e920634a9c84c90640108f384eb39159

maxentile (approved):
Looks good!

PR #1262, opened 2024-02-24
Minimal repro showing cyclopropane hydrogens being invertible

proteneer:
I compared the chiral volumes of cyclopropane in a typical vacuum simulation, with and without the `cos_angle=True` term. Unfortunately it looks like the default OFF parameters has them as naturally invertible. 

```
dt = 1.5e-3
with `cos_angles=True`:
pos vols 14
neg vols 1199986

with `cos_angles=False`:
pos vols 8
neg vols 1199992
```


proteneer:
(irrelevant for now)

PR #1263, opened 2024-02-26
Plot Chiral Heatmap

badisa:
* Moves the chiral heatmap code out of tests
* Reuses some of the existing single topology code (should probably move out of the cif writer at some point, left for deboggling unless we want to do it now)

# Plot examples

## Vacuum
Both from the poorly aligned cases, as in the well aligned there are no chiral restraint energies that are non-zero

### Mol A
vacuum_chiral_energies_aligned_False_a

### Mol B
vacuum_chiral_energies_aligned_False_b

### Mol A
solvent_chiral_energies_a

### Mol B
solvent_chiral_energies_b

### DIFF
--- a/tests/test_chiral_restraints.py
+++ b/tests/test_chiral_restraints.py
@@ -3,7 +3,6 @@ from dataclasses import replace
 import numpy as np
 import pytest
 import scipy
-from jax import jit
 from jax import numpy as jnp
 from rdkit import Chem
 
@@ -14,8 +13,9 @@ from timemachine.constants import (
 )
 from timemachine.fe import topology, utils
 from timemachine.fe.atom_mapping import get_cores
+from timemachine.fe.chiral_utils import make_chiral_flip_heatmaps
 from timemachine.fe.free_energy import HREXParams
-from timemachine.fe.rbfe import DEFAULT_HREX_PARAMS, run_vacuum
+from timemachine.fe.rbfe import DEFAULT_HREX_PARAMS, run_solvent, run_vacuum
 from timemachine.fe.single_topology import AtomMapMixin
 from timemachine.fe.system import simulate_system
 from timemachine.ff import Forcefield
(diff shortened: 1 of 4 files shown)


tests/test_chiral_restraints.py
```diff
@@ -687,92 +687,62 @@ def test_chiral_inversion_in_single_topology_runs(well_aligned):
...
-        mol_a_chiral_conflicts.append(np.array([U_a(x) for x in xs_a]))
-        mol_b_chiral_conflicts.append(np.array([U_b(x) for x in xs_b]))
+    vacuum_results = run_vacuum(
+        atom_map.mol_a,
+        atom_map.mol_b,
+        atom_map.core,
+        ff,
+        None,
+        DEFAULT_HREX_PARAMS,
+        min_overlap=0.667,  # No need to have a higher overlap then 0.667
```
badisa, line 706: Note this significantly speeds up this test. Previously every vacuum simulation took up 48 windows

timemachine/fe/free_energy.py
```diff
@@ -780,7 +780,7 @@ def compute_intermediate_result(lambdas: Sequence[float]) -> PairBarResult:
             if min_overlap is not None:
                 overlap_info = f"Current minimum BAR overlap {cost_to_overlap(max(costs)):.3g} <= {min_overlap:.3g} "
             else:
-                overlap_info = f"Current minimum BAR overlap {cost_to_overlap(max(costs)):.3g} (min_overlap == None)"
+                overlap_info = f"Current minimum BAR overlap {cost_to_overlap(max(costs)):.3g} (min_overlap == None) "
```
badisa, line 783: This looks weird without the space, so added it in even though this is not a common practice for us

timemachine/fe/plots.py
```diff
@@ -16,6 +16,8 @@ def plot_work(w_forward, w_reverse, axes):
     """histograms of +forward and -reverse works"""
 
     w_all = np.concatenate([+w_forward, -w_reverse])
+    # Tear out any non-finite works
+    w_all = w_all[np.isfinite(w_all)]
```
badisa, line 20: Avoids an exception that can kill simulations where states have inf works

badisa:
Running a nightly pipeline since this touches nightly tests: https://gitlab.com/relaytx/physics-platform/timemachine/-/pipelines/1191753182

timemachine/fe/chiral_utils.py
```diff
@@ -398,3 +401,72 @@ def enumerate_planar_torsions(mol):
...
+        return U_chiral_atom_batch(x_a, restr_idxs_a, chiral_k).sum()
+
+    @jit
+    def U_b(x_b):
+        return U_chiral_atom_batch(x_b, restr_idxs_b, chiral_k).sum()
+
+    return U_a, U_b
+
+
+def xs_ab_from_xs(xs: NDArray, atom_map):
```
maxentile, line 421:
```suggestion
def xs_ab_from_xs(xs: NDArray, atom_map: AtomMapMixin):
```

maxentile (approved):

timemachine/fe/chiral_utils.py
badisa: The reason this isn't imported it is that it creates a circular import. I had tried that originally and then just tried putting in the string but Mypy complained.

Will add a docstring so it is clear

badisa: Added docstring in https://github.com/proteneer/timemachine/pull/1263/commits/05cca3e74ba7f663f87a9647e4cea37d73bc740b

tests/test_chiral_restraints.py
badisa: Saves about 40 minutes of the nightly tests

proteneer (approved):

PR #1264, opened 2024-03-05
Update Mypy in CI

badisa:
* Missed in https://github.com/proteneer/timemachine/pull/1181
* Don't think this really matters, but good to have a consistent mypy across CI/local


maxentile:
>Don't think this really matters, but good to have a consistent mypy across CI/local

Unsure: should this also be updated to match https://github.com/proteneer/timemachine/blob/181221ed605dea0ed069cd811419b9039dcc9823/pyproject.toml#L30 ?

maxentile (approved):

PR #1266, opened 2024-03-05
Tweak check_chiral_validity docstring 

maxentile:
Was inconsistent with implementation, implementation was correct


proteneer (approved):
nice catch and clean-up.

PR #1268, opened 2024-03-11
Separate out Noise Buffers

badisa:
In the previous implementation some of the noise used in different places was combined into a single buffer as it is simple to offset to the correct data in the CPU. When we start batching the moves the offsetting would get much more complicated and so this PR separates all of the different types of noise into their own buffer.

Also learnt that only the Normal noise requires the buffers to be even. In a follow up PR will fix up the places we needlessly call `round_up_even`.


badisa:
Running nightly tests: https://gitlab.com/relaytx/physics-platform/timemachine/-/pipelines/1209561622

badisa:
Nightly tests failed and am fixing them locally, but just the change in the determinism has resulted in needing to take more steps.

badisa:
Follow up:
* Investigate the change in the 6.0nm water test that fails, is the accepted moves all 10k - 20k?
* Re-run the Buckyball cases to ensure occupancies stay the same.

tests/test_cuda_bd_exchange_mover.py
```diff
@@ -352,7 +352,7 @@ def test_bd_exchange_deterministic_moves(proposals_per_move, precision, seed):
         (5000, 40000, 3.0),
         (10000, 250000, 3.0),
         # The 6.0nm box triggers a failure that would occur with systems of certain sizes, may be flaky in identifying issues
-        pytest.param(1, 10000, 6.0, marks=pytest.mark.nightly(reason="slow")),
+        pytest.param(1, 20000, 6.0, marks=pytest.mark.nightly(reason="slow")),
```
badisa, line 356: Before changes
```
tests/test_cuda_bd_exchange_mover.py::test_moves_in_a_water_box[2023-float32-0.0001-0.002-1-10000-6.0]
Accepted 1 moves on step 3024
Accepted 1 moves on step 3666
Accepted 1 moves on step 3981
Accepted 1 moves on step 4869
Accepted 1 moves on step 6001
Accepted 1 moves on step 7153
Accepted 1 moves on step 7773
Accepted 1 moves on step 8244
Accepted 8 of 10000 moves
```
After separated noise
```
tests/test_cuda_bd_exchange_mover.py::test_moves_in_a_water_box[2023-float32-0.0001-0.002-1-20000-6.0]
Accepted 1 moves on step 10364
Accepted 1 moves on step 10394
Accepted 1 moves on step 17521
Accepted 1 moves on step 18071
Accepted 1 moves on step 19339
Accepted 5 of 20000 moves
```

Don't think this is a sign of bad behavior, rather not converged and the inherent randomness. Running the buckyball case as that is a more definitive analysis of this problem.

tests/test_cuda_bd_exchange_mover.py
```diff
@@ -52,6 +52,7 @@ def verify_bias_deletion_moves(
                 new_pos = x_move[mol_idxs]
                 idx = i
         if num_moved > 0:
+            print(f"Accepted {num_moved} moves on step {step}")
```
badisa, line 55: Added this as it is handy to know when moves are accepted

timemachine/cpp/src/tibd_exchange_move.cu
```diff
@@ -46,13 +43,12 @@ TIBDExchangeMove<RealType>::TIBDExchangeMove(
...
           batch_size,
-          round_up_even(TIBD_TRANSLATIONS_PER_STEP_XYZXYZ * num_proposals_per_move)),
+          TIBD_TRANSLATIONS_PER_STEP_XYZXYZ * num_proposals_per_move),
       radius_(static_cast<RealType>(radius)), inner_volume_(static_cast<RealType>((4.0 / 3.0) * M_PI * pow(radius, 3))),
       d_rand_states_(DEFAULT_THREADS_PER_BLOCK), d_inner_mols_count_(1), d_identify_indices_(this->num_target_mols_),
       d_partitioned_indices_(this->num_target_mols_), d_temp_storage_buffer_(0), d_center_(3),
-      d_uniform_noise_buffer_(round_up_even(NOISE_PER_STEP * this->steps_per_move_ * this->batch_size_)),
-      d_targeting_inner_vol_(this->batch_size_), d_ligand_idxs_(ligand_idxs),
-      d_src_log_weights_(this->num_target_mols_ * this->batch_size_),
+      d_uniform_noise_buffer_(num_proposals_per_move), d_targeting_inner_vol_(this->batch_size_),
```
proteneer, line 50: just checking, `d_uniform_noise_buffer_` is now used only for deciding the targeted move, and the `mh` noise is inherited from the base class? 

also, is `this->steps_per_move_ * this->batch_size_ == num_proposals_per_move`?

timemachine/cpp/src/tibd_exchange_move.cu
```diff
@@ -61,15 +57,15 @@ TIBDExchangeMove<RealType>::TIBDExchangeMove(
     if (radius <= 0.0) {
         throw std::runtime_error("radius must be greater than 0.0");
     }
-    if (d_uniform_noise_buffer_.length / NOISE_PER_STEP != this->d_quaternions_.length / this->QUATERNIONS_PER_STEP) {
+    if (d_uniform_noise_buffer_.length != this->d_quaternions_.length / this->QUATERNIONS_PER_STEP) {
         throw std::runtime_error("bug in the code: buffers with random values don't match in batch size");
     }
 
-    // Add 3 to the seed provided to avoid correlating with the three other RNGs
+    // Add 4 to the seed provided to avoid correlating with the four other RNGs
```
proteneer, line 64: yuck (but not sure if there's any better ways)

proteneer (approved):
lgtm - good clean-up.

proteneer (commented):
(hold off on merging until we check some nightlies)

timemachine/cpp/src/tibd_exchange_move.cu
badisa: > just checking, d_uniform_noise_buffer_ is now used only for deciding the targeted move, and the mh noise is inherited from the base class?

Exactly

> also, is this->steps_per_move_ * this->batch_size_ == num_proposals_per_move?

Yup

proteneer (approved):

badisa:
Nightlies all look good

```
Starting with Six waters - Targeted run for 10k iterations
Total Moves 100000000
Num Waters in Buckyball: 2, 49.31%
Num Waters in Buckyball: 1, 50.69%
```

Looks very similar to https://github.com/proteneer/timemachine/pull/1228, though as we discussed in that PR these are not converged. 

image

PR #1270, opened 2024-03-14
Batch BDExchangeMover

badisa:
* In testing (not looked at what is the bottleneck yet) seems to be up to 10x faster on the bucky ball case (5.5k ns per proposal)
* Opening PR to see what broke, as surely things broke


badisa:
Found a test locally that failed, closing.

PR #1272, opened 2024-03-15
Fix BDExchange Mover Bug

badisa:
* Occasionally the `k_store_accepted_log_probability` would change out the underlying before log probability due to using different random noise.
* This triggers failure in `tests/test_examples.py::test_water_sampling_mc_buckyball[untargeted]`


timemachine/cpp/src/bd_exchange_move.cu
```diff
@@ -169,7 +169,8 @@ void BDExchangeMove<RealType>::move(
             // Need the weights to sample a value and the log probs are just because they aren't expensive to copy
             k_store_accepted_log_probability<RealType><<<1, tpb, 0>>>(
                 num_target_mols_,
-                d_mh_noise_.data + (step * batch_size_),
+                // Decrement step by one to use the same noise as previously
+                d_mh_noise_.data + (step - 1 * batch_size_),
```
badisa, line 173: Need to use the noise of the previous step to ensure the right values are stored.

proteneer: erm - is this intended to be `(step - 1) * batch_size_` - the expression as written is equivalent to `(step - batch_size)`

badisa: Yes, you are right. It works because only because `batch_size_ == 1`

badisa: Fixed in https://github.com/proteneer/timemachine/pull/1272/commits/9ac9603bcce6d35d00b4565a980bf9e3ff319744

proteneer (approved):

PR #1274, opened 2024-03-15
Batch BDExchangeMove

badisa:
* Appears ~15x faster (3200ns per proposal vs 50kns per proposal) using the `water_sampling_mc.py` example script with a batch size of 250.
* The single proposal is slower thanks to the while loop 
* Some jank around the sampler kernels (not a clean abstraction, forced it to fit the use case), though would be inclined to address this later and move onto the TIBDExchangeMover batching.
* Bitwise identical to changes in https://github.com/proteneer/timemachine/pull/1272


timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -59,76 +65,47 @@ void __global__ k_setup_proposals(
...
+    const int num_atoms_in_each_mol,
+    const int *__restrict__ accepted_batched_move, // [1]
+    const int *__restrict__ mol_idx_per_batch,     // [batch_size]
+    const int *__restrict__ mol_offsets,           // [num_target_mols]
+    const double *__restrict__ moved_coords,       // [batch_size, num_atoms_in_each_mol, 3]
+    double *__restrict__ dest_coords,              // [N, 3]
+    size_t *__restrict__ num_accepted,             // [1],
+    int *__restrict__ rand_offset                  // [1]
 ) {
     // Note that this kernel does not handle multiple proposals, expects that the proposals
```
proteneer, line 79: Is this comment still relevant now that it supports batching?

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -59,76 +65,47 @@ void __global__ k_setup_proposals(
...
-    const RealType before_log_prob =
-        convert_nan_to_inf<RealType>(compute_logsumexp_final<RealType>(before_max[0], before_log_sum[0]));
-    const RealType after_log_prob =
-        convert_nan_to_inf<RealType>(compute_logsumexp_final<RealType>(after_max[0], after_log_sum[0]));
-
-    const RealType log_acceptance_prob = min(before_log_prob - after_log_prob, static_cast<RealType>(0.0));
-    const bool accepted = rand[0] < exp(log_acceptance_prob);
-    if (idx == 0 && accepted) {
+    const int batch_idx = accepted_batched_move[0];
+    // If the selected mol is not less then the total batch size, no proposal was accepted, we can exit immediately.
```
proteneer, line 85: selected mol refers to `batch_idx/accepted_batched_move[0]`?

tests/test_cuda_bd_exchange_mover.py
```diff
@@ -330,35 +325,122 @@ def test_bd_exchange_deterministic_moves(proposals_per_move, precision, seed):
...
+        batch_size=batch_size,
+    )
+
+    iterative_moved_coords = conf.copy()
+    for _ in range(iterations):
+        iterative_moved_coords, _ = bdem_a.move(iterative_moved_coords, box)
+        assert not np.all(conf == iterative_moved_coords)
+    batch_moved_coords, _ = bdem_b.move(conf, box)
+
     assert bdem_a.n_accepted() > 0
```
proteneer, line 420: document what a typical `n_accepted()` is?

timemachine/cpp/src/bd_exchange_move.cu
```diff
@@ -162,15 +165,18 @@ void BDExchangeMove<RealType>::move(
     curandErrchk(templateCurandUniform(cr_rng_translations_, d_translations_.data, d_translations_.length));
     curandErrchk(templateCurandUniform(cr_rng_samples_, d_sample_noise_.data, d_sample_noise_.length));
     curandErrchk(templateCurandUniform(cr_rng_mh_, d_mh_noise_.data, d_mh_noise_.length));
-    for (int step = 0; step < steps_per_move_; step++) {
+    size_t step = 0;
+    // For the first pass just set the value to zero on the host
+    p_noise_offset_.data[0] = 0;
```
proteneer, line 170: style-nit: should this be just `*p_noise_offset_.data = 0` since it's a pointer to a scalar?

timemachine/cpp/src/bd_exchange_move.cu
```diff
@@ -217,21 +218,39 @@ void BDExchangeMove<RealType>::move(
             d_lse_exp_sum_after_.data,
             stream);
 
-        k_attempt_exchange_move<RealType><<<1, 1, 0, stream>>>(
-            N,
-            d_mh_noise_.data + (step * batch_size_),
+        k_select_first_valid_move<RealType><<<1, min(512, batch_size_), 0, stream>>>(
```
proteneer, line 221: we should modify determinism tests above for batch_sizes: `{> 512, < 512, = 512}` to make sure this logic is correct

timemachine/cpp/src/bd_exchange_move.cu
```diff
@@ -162,15 +165,18 @@ void BDExchangeMove<RealType>::move(
     curandErrchk(templateCurandUniform(cr_rng_translations_, d_translations_.data, d_translations_.length));
     curandErrchk(templateCurandUniform(cr_rng_samples_, d_sample_noise_.data, d_sample_noise_.length));
     curandErrchk(templateCurandUniform(cr_rng_mh_, d_mh_noise_.data, d_mh_noise_.length));
-    for (int step = 0; step < steps_per_move_; step++) {
+    size_t step = 0;
+    // For the first pass just set the value to zero on the host
+    p_noise_offset_.data[0] = 0;
+    while (p_noise_offset_.data[0] < num_proposals_per_move_) {
```
proteneer, line 171: should add a brief but clear description about the overall algorithm at the start of the while loop

timemachine/cpp/src/bd_exchange_move.cu
```diff
@@ -162,15 +165,18 @@ void BDExchangeMove<RealType>::move(
     curandErrchk(templateCurandUniform(cr_rng_translations_, d_translations_.data, d_translations_.length));
     curandErrchk(templateCurandUniform(cr_rng_samples_, d_sample_noise_.data, d_sample_noise_.length));
     curandErrchk(templateCurandUniform(cr_rng_mh_, d_mh_noise_.data, d_mh_noise_.length));
-    for (int step = 0; step < steps_per_move_; step++) {
+    size_t step = 0;
```
proteneer, line 168: does `step` become redundant with `p_noise_offset_.data`?

timemachine/cpp/src/bd_exchange_move.cu
```diff
@@ -217,21 +218,39 @@ void BDExchangeMove<RealType>::move(
...
             d_lse_max_before_.data,
             d_lse_exp_sum_before_.data,
             d_lse_max_after_.data,
             d_lse_exp_sum_after_.data,
-            d_target_mol_offsets_.data,
+            d_mh_noise_.data,
+            d_selected_sample_.data);
+        gpuErrchk(cudaPeekAtLastError());
+
+        k_accepted_exchange_move<<<1, 1, 0, stream>>>(
```
proteneer, line 235: naming, why past-tense `k_accepted` vs present-tense `k_accept`? Did the attempt *already* happen by this point?

timemachine/cpp/src/bd_exchange_move.cu
```diff
@@ -217,21 +218,39 @@ void BDExchangeMove<RealType>::move(
...
+            d_noise_offset_.data);
         gpuErrchk(cudaPeekAtLastError());
-        num_attempted_++;
+        gpuErrchk(cudaMemcpyAsync(
+            p_noise_offset_.data, d_noise_offset_.data, d_noise_offset_.size(), cudaMemcpyDeviceToHost, stream));
+        step++;
+        // Synchronize to get the new offset
+        gpuErrchk(cudaStreamSynchronize(stream));
     }
+    // Number of attempts is always the batch size
```
proteneer, line 252: nit: is always incremented by the `batch_size`

proteneer (commented):
Did a first pass - will look at kernel code in more detail in the PM

timemachine/cpp/src/bd_exchange_move.cu
badisa: Yes, the acceptance is done inside of the `k_select_first_valid_move` which should probably be `k_accept_first_valid_move`

timemachine/cpp/src/kernels/k_exchange.cu
badisa: Yes, this kernel just takes the output of `k_select_first_valid_move`

timemachine/cpp/src/kernels/k_sampling.cu
```diff
@@ -38,6 +38,67 @@ template void __global__ k_setup_gumbel_max_trick<double>(
...
+
+    const int rand_offset = noise_offset[0];
+    if (rand_offset + segment_idx >= max_offset) {
+        return;
+    }
+
+    const int segment_start = segment_offsets[segment_idx];
+    const int N = segment_offsets[segment_idx + 1] - segment_start;
+    // In the case of the offset the values per segment need to match up to compute gumbel offset correctly.
+    assert(total_values % N == 0);
```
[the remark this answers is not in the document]
badisa: There is an invalid memory access happening somewhere, not clear where yet.

timemachine/cpp/src/bd_exchange_move.cu
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1274/commits/3821557f46e091272e0297d24437e906375740fe

timemachine/cpp/src/bd_exchange_move.cu
badisa: Removed in https://github.com/proteneer/timemachine/pull/1274/commits/28b3dca9f1c602bd7493ac41902009fef13cde24

timemachine/cpp/src/bd_exchange_move.cu
badisa: Switched dereference style in https://github.com/proteneer/timemachine/pull/1274/commits/9a1895a617de3ccf6adb544aa785035bf1a4568a https://github.com/proteneer/timemachine/pull/1274/commits/9a1895a617de3ccf6adb544aa785035bf1a4568a

timemachine/cpp/src/bd_exchange_move.cu
badisa: Added description here, let me know what you think: https://github.com/proteneer/timemachine/pull/1274/commits/fc62abe77506e217f422b5bf5a340e19e4c3acee

tests/test_cuda_bd_exchange_mover.py
badisa: Added and increased the acceptance requirement to half the moves: https://github.com/proteneer/timemachine/pull/1274/commits/0c1cdf823244d222f2e7fb0de475bf58ce85efab

timemachine/cpp/src/bd_exchange_move.cu
badisa: Renamed `k_select_first_valid_move` in https://github.com/proteneer/timemachine/pull/1274/commits/debf9b2c97397497ae5e9b10ab16b325326b4d1f

timemachine/cpp/src/bd_exchange_move.cu
badisa: Added test in https://github.com/proteneer/timemachine/pull/1274/commits/e8b45c58a356b3b7746445219c4ec00d841ae1c8

timemachine/cpp/src/kernels/k_exchange.cu
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1274/commits/2ec5c9a5423fb6fd79736ffb8aba7f556bbace86

timemachine/cpp/src/kernels/k_sampling.cu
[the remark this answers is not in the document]
badisa: Fixed issues in https://github.com/proteneer/timemachine/pull/1274/commits/f9c0bc54216e40885389b5bb7fb520c7c085d272

The logic for handling going beyond the offset was incorrect and could access garbage data that allowed reading beyond the buffers

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -33,15 +33,21 @@ k_copy_batch<double>(const int N, const int batch_size, const double *__restrict
 // indices for each sample. Note that the output mol offsets are constructed so that the start of the mol is
 // the starting atom idx rather than the prefix sum of mol lengths that the mol_offsets is.
 void __global__ k_setup_proposals(
-    const int batch_size,                      // Number of molecules to setup
-    const int num_atoms_in_each_mol,           // number of atoms in each sample
+    const int total_proposals,
+    const int batch_size,            // Number of molecules to setup
```
proteneer, line 37: comment here seems wrong

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -59,76 +65,47 @@ void __global__ k_setup_proposals(
...
-    const RealType before_log_prob =
-        convert_nan_to_inf<RealType>(compute_logsumexp_final<RealType>(before_max[0], before_log_sum[0]));
-    const RealType after_log_prob =
-        convert_nan_to_inf<RealType>(compute_logsumexp_final<RealType>(after_max[0], after_log_sum[0]));
-
-    const RealType log_acceptance_prob = min(before_log_prob - after_log_prob, static_cast<RealType>(0.0));
-    const bool accepted = rand[0] < exp(log_acceptance_prob);
-    if (idx == 0 && accepted) {
+    const int batch_idx = accepted_batched_move[0];
+    // If the selected batch idx is not less then the total batch size, no proposal was accepted, we can exit immediately.
```
proteneer, line 85: then -> than

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -59,76 +65,47 @@ void __global__ k_setup_proposals(
...
+    if (batch_idx >= batch_size) {
+        rand_offset[0] += batch_size;
+        return;
+    }
+    const int mol_idx = mol_idx_per_batch[batch_idx];
+    const int mol_start = mol_offsets[mol_idx];
+    // Increment offset by the index + 1, IE if the zeroth item in the batch was accepted offset by 1
+    rand_offset[0] += batch_idx + 1;
+    if (threadIdx.x == 0) {
```
proteneer, line 94: `threadIdx.x` or `idx`?

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -59,76 +65,47 @@ void __global__ k_setup_proposals(
...
-    if (idx == 0 && accepted) {
+    const int batch_idx = accepted_batched_move[0];
+    // If the selected batch idx is not less then the total batch size, no proposal was accepted, we can exit immediately.
+    if (batch_idx >= batch_size) {
+        rand_offset[0] += batch_size;
+        return;
+    }
+    const int mol_idx = mol_idx_per_batch[batch_idx];
+    const int mol_start = mol_offsets[mol_idx];
+    // Increment offset by the index + 1, IE if the zeroth item in the batch was accepted offset by 1
```
proteneer, line 92: Where's the "if statement" corresponding to the following line of code? The code looks like it increments by at least 1 regardless of the behavior

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -59,76 +65,47 @@ void __global__ k_setup_proposals(
...
-    if (accepted && idx == 0) {
-        // If accepted, move the coords of the selected mol into place
-        for (int i = 0; i < num_atoms_in_each_mol; i++) {
-            dest_coords[(mol_start + i) * 3 + 0] = moved_coords[i * 3 + 0];
-            dest_coords[(mol_start + i) * 3 + 1] = moved_coords[i * 3 + 1];
-            dest_coords[(mol_start + i) * 3 + 2] = moved_coords[i * 3 + 2];
-        }
+    // If accepted, move the coords of the selected mol into place
+    for (int i = 0; i < num_atoms_in_each_mol; i++) {
+        auto moved_x = moved_coords[num_atoms_in_each_mol * batch_idx * 3 + i * 3 + 0];
```
proteneer, line 100: auto here means double? (should be explicit)

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -837,4 +810,85 @@ void __global__ k_adjust_sample_idxs(
...
+    const int *__restrict__ samples,             // [batch_size]
+    const RealType *__restrict__ before_max,     // [1]
+    const RealType *__restrict__ before_log_sum, // [1]
+    const RealType *__restrict__ after_max,      // [batch_size]
+    const RealType *__restrict__ after_log_sum,  // [batch_size]
+    const RealType *__restrict__ rand,           // [total_proposals]
+    int *__restrict__ accepted_sample            // [1]
+) {
+    __shared__ int selected_idx;
+    assert(blockIdx.x == 0);
```
proteneer, line 832: I wonder if we can turn statements like this into compile time `static_asserts`, (blockIdx.x may not be known at compile time, but gridDim.x should be known at compile time?)

timemachine/cpp/src/bd_exchange_move.cu
```diff
@@ -66,24 +66,21 @@ BDExchangeMove<RealType>::BDExchangeMove(
...
-      d_lse_max_after_(batch_size_), d_lse_exp_sum_after_(batch_size_), d_samples_(batch_size_),
+      d_lse_max_after_(batch_size_), d_lse_exp_sum_after_(batch_size_), d_samples_(batch_size_), d_selected_sample_(1),
       d_quaternions_(round_up_even(QUATERNIONS_PER_STEP * num_proposals_per_move_)),
       d_mh_noise_(num_proposals_per_move), d_num_accepted_(1), d_target_mol_atoms_(batch_size_ * mol_size_),
       d_target_mol_offsets_(num_target_mols_ + 1),
       d_intermediate_sample_weights_(batch_size_ * num_intermediates_per_reduce_),
       d_sample_noise_(num_target_mols_ * num_proposals_per_move_),
       d_sampling_intermediate_(num_target_mols_ * batch_size_), d_translations_(translation_buffer_size),
-      d_sample_segments_offsets_(batch_size_ + 1) {
+      d_sample_segments_offsets_(batch_size_ + 1), d_noise_offset_(1), p_noise_offset_(1) {
```
proteneer, line 77: do we require that `batch_size <= num_proposals_per_move`? if not, then we should test for this

timemachine/cpp/src/kernels/k_exchange.cu
badisa: Good catc

timemachine/cpp/src/kernels/k_exchange.cu
badisa: Woops from some debugging that wasn't undone, should just copy directly without the intermediate variable

timemachine/cpp/src/bd_exchange_move.cu
badisa: We do not, but will enforce that since that is not a useful way to run this, at least it doesn't seem useful to me.

timemachine/cpp/src/kernels/k_exchange.cu
badisa: You always increment it by 1, will clarify the comment.

If you accepted on the 10th item in the batch you will want to next read at the 11th index to avoid reading in the noise from the 10th item. 
The if in the comment is a red herring, there is an if above to catch the case where nothing was accepted (which is results in the same offset as accepting the last move)
https://github.com/proteneer/timemachine/blob/feature/batch-bd-mover/timemachine/cpp/src/kernels/k_exchange.cu#L86-L93

badisa: Attempted to clarify in https://github.com/proteneer/timemachine/pull/1274/commits/b80d8d181a8da212087f4c07778dbf56a9f241d4

timemachine/cpp/src/kernels/k_exchange.cu
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1274/commits/b80d8d181a8da212087f4c07778dbf56a9f241d4

timemachine/cpp/src/kernels/k_exchange.cu
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1274/commits/b80d8d181a8da212087f4c07778dbf56a9f241d4

timemachine/cpp/src/kernels/k_exchange.cu
badisa: Removed in https://github.com/proteneer/timemachine/pull/1274/commits/21ccc9876c862db3f9704074b4f032834f9ec3fd

timemachine/cpp/src/bd_exchange_move.cu
badisa: Made that a requirement in https://github.com/proteneer/timemachine/pull/1274/commits/c9fe355d1b20b0cf19ae2775d213e0105191bde3

timemachine/cpp/src/kernels/k_exchange.cu
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1274/commits/46eafcee78c75520b6f956bfc2314309f278eb4b

timemachine/cpp/src/kernels/k_sampling.cu
```diff
@@ -38,6 +38,67 @@ template void __global__ k_setup_gumbel_max_trick<double>(
...
+
+    const int gumbel_offset = rand_offset * N + segment_start;
+    int idx = blockIdx.x * blockDim.x + threadIdx.x;
+    while (idx < N) {
+        const RealType weight = log_weights[idx];
+        assert(!isnan(weight));
+
+        // If the idx in the batch segments is beyond the max offset return a negative infinity to avoid sampling the weight
+        const RealType gumbel_rand = -log(-log(gumbel_noise[gumbel_offset + idx]));
```
proteneer, line 74: sanity checking:

`gumbel_offset + idx = rand_offset * N + segment_start + idx`

is always going to be less than `N * num_proposals_per_move`?

The while statement guarantees that `idx < N` - but less clear on the `rand_offset * N + segment_start` part

timemachine/cpp/src/kernels/k_sampling.cu
```diff
@@ -38,6 +38,67 @@ template void __global__ k_setup_gumbel_max_trick<double>(
...
+    // In the case of the offset the values per segment need to match up to compute gumbel offset correctly.
+    assert(total_values % N == 0);
+
+    const int gumbel_offset = rand_offset * N + segment_start;
+    int idx = blockIdx.x * blockDim.x + threadIdx.x;
+    while (idx < N) {
+        const RealType weight = log_weights[idx];
+        assert(!isnan(weight));
+
+        // If the idx in the batch segments is beyond the max offset return a negative infinity to avoid sampling the weight
```
proteneer, line 73: is this comment still relevant? Not clear where/how the negative infinity is returned

badisa: Nope, will remove

timemachine/cpp/src/kernels/k_sampling.cu
badisa: In the case of BiasedDeletion, the `rand_offset` represents the number the number of segments that have been computed previously. Note that this is likely to change for TIBD, but don't want to muddy the waters in this PR where the following logic is correct.

The `gumbel_noise` array is constructed with length `num_target_mols_ * num_proposals_per_move_` where `num_target_mols == N` in this kernel. 

So `rand_offset * num_target_mols (N in this kernel)` as long as the value of `rand_offset + segment_idx < num_proposals_per_move`, where `segment_idx` is the segment to evaluate, is safe and will result in at most `num_target_mols * (num_proposals_per_move - 1)`. We have a check for this here: https://github.com/proteneer/timemachine/blob/feature/batch-bd-mover/timemachine/cpp/src/kernels/k_sampling.cu#L58 where `max_offset == num_proposals_per_move`. Leaving us with the `segment_start + idx` component.

The `segment_offsets` is expected to be a a set of offsets into the arrays that are to be sampled, with `segment_offsets[0] == 0` and `segment_offsets[-1] == num_target_mols * num_segments`. In this case we take the delta from `segment_offsets[i + 1] - segment_offsets[i]` giving us the number of values in the segment. We start from `segment_offsets[i]` and iterate until we hit `segment_offsets[i+1] - 1` which at most be `num_target_mols - 1` which when added to the `rand_offset * N` will at most be `(num_target_mols * num_proposals_per_move) - 1` which is the expected final value this kernel should achieve.

timemachine/cpp/src/kernels/k_sampling.cu
badisa: Removed in https://github.com/proteneer/timemachine/pull/1274/commits/6bf2b83d9185f956eb88558cce6671b56bdb5918

timemachine/cpp/src/kernels/k_sampling.cu
badisa: Added a test with only a single mol which would catch if we were overrunning the buffer (its running under `compute-sanitizer --tool memcheck`)

proteneer (approved):

PR #1278, opened 2024-03-20
Update Targeted Bitwise determinism test

badisa:
* Angles change shifted the simulation
* ~Why did this change the targeted but not the biased deletion version??~ It did as we would expected, just missed it as I had seen `test_water_sampling_mc_bulk_water` pass which doesn't compare the final coordinates the way `test_water_sampling_mc_buckyball` does.


proteneer (approved):

proteneer:
lgtm - good catch!

PR #1282, opened 2024-03-26
Batch TIBDExchangeMover

badisa:
* Appears to about 12x faster with a batch size of 250 on the Buckyball test case (80kns vs 6.5kns), though haven't explored what the right batch size is here.
* Needs some more rigorous testing on systems like HSP90 and BRD4 before merging, worry there is something subtle in here
* There are some things that could be cleaned up later (some extra buffers lying around, the addition of the new translations buffer, etc), but want to go for a straightforward approach rather then rework it immediately.


badisa:
Determinism for batching isn't there yet.

PR #1285, opened 2024-04-05
Move switch_fn to attic, add norm'd switch_fn_2

maxentile:
Move switch_fn from nonbonded.py to attic.

Doesn't have expected boundary conditions when cutoff != 1.0, and is unused (only usages were in commented-out code when introduced in https://github.com/proteneer/timemachine/pull/205 ).

Possible normalization suggested, but untested

image


maxentile:
Will be superseded

PR #1286, opened 2024-04-08
Enable Water Sampling in Simulations

badisa:
* Disables water sampling for solvent, doesn't seem to make a difference
* Some random clean up of the example scripts that were from trying to use them to test water sampling
* Runs TIBDExchangeMove on all windows
* Defaults to 1000 MC proposals per 400 steps of MD which makes for about 10% additional runtime cost. Performance optimizations intended in a follow up. 

### Follow up questions
- How do we effectively target a region that will produce effective moves (note that probability of placing in the inner region is the furthest from the center)? Currently defaults to 1.0nm from the centroid of the ligand (want this to be ligand based, to be generic), probably in practice take the max pair wise distance of the ligands and add a padding (~0.4nm).
  * Select ligand atoms that are 'buried' in the protein to define sphere where we think it is important, unclear how to define
  *  Randomize the center of the sphere/radius (no API available short of reconstructing the entire class)

### Todo

- [x] Add a test with a 3 atom ligand (including dummy atoms)

### DIFF
--- a/tests/test_free_energy.py
+++ b/tests/test_free_energy.py
@@ -302,6 +302,8 @@ def test_get_water_sampler_params(num_windows):
     num_host_atoms = solvent_conf.shape[0]
     host_system, masses = convert_omm_system(solvent_sys)
     solvent_host = Host(host_system, masses, solvent_conf, solvent_box, num_host_atoms)
+    mol_a_only_atoms = np.array([i for i in range(st.get_num_atoms()) if st.c_flags[i] == 1])
+    mol_b_only_atoms = np.array([i for i in range(st.get_num_atoms()) if st.c_flags[i] == 2])
     for lamb in np.linspace(0.0, 1.0, num_windows, endpoint=True):
         state = setup_initial_state(st, lamb, solvent_host, DEFAULT_TEMP, 2024)
         water_sampler_nb_params = get_water_sampler_params(state)
@@ -309,6 +311,12 @@ def test_get_water_sampler_params(num_windows):
         ligand_water_params = st._get_guest_params(
             forcefield.q_handle_solv, forcefield.lj_handle_solv, lamb, nb_pot.cutoff
         )
+        if lamb == 0.0:
+            assert np.all(ligand_water_params[mol_a_only_atoms][:, 3] == 0.0)
+            assert np.all(ligand_water_params[mol_b_only_atoms][:, 3] == nb_pot.cutoff)
+        elif lamb == 1.0:
+            assert np.all(ligand_water_params[mol_a_only_atoms][:, 3] == nb_pot.cutoff)
+            assert np.all(ligand_water_params[mol_b_only_atoms][:, 3] == 0.0)
 
         np.testing.assert_array_equal(water_sampler_nb_params[num_host_atoms:], ligand_water_params)
 
--- a/tests/test_free_energy.py
+++ b/tests/test_free_energy.py
@@ -22,16 +22,18 @@ from timemachine.fe.free_energy import (
     PairBarResult,
     batches,
     estimate_free_energy_bar,
+    get_water_sampler_params,
     make_pair_bar_plots,
     run_sims_bisection,
     sample,
 )
-from timemachine.fe.rbfe import setup_initial_state, setup_initial_states, setup_optimized_host
+from timemachine.fe.rbfe import Host, setup_initial_state, setup_initial_states, setup_optimized_host
 from timemachine.fe.single_topology import SingleTopology
 from timemachine.fe.stored_arrays import StoredArrays
+from timemachine.fe.system import convert_omm_system
 from timemachine.ff import Forcefield
 from timemachine.md import builders
-from timemachine.potentials import SummedPotential
+from timemachine.potentials import Nonbonded, SummedPotential
 from timemachine.testsystems.relative import get_hif2a_ligand_pair_single_topology
 
 
@@ -287,6 +289,28 @@ def test_plot_pair_bar_plots(mock_fig, hif2a_ligand_pair_single_topology_lam0_st
         "ChiralAtomRestraint",
     }
     assert set(mock_fig.call_args.args[0]) == expected_potentials
+
+
+@pytest.mark.nocuda
+@pytest.mark.parametrize("num_windows", [2, 5])
+def test_get_water_sampler_params(num_windows):
+    mol_a, mol_b, core = get_hif2a_ligand_pair_single_topology()
+    # Use the simple charges simply because it is faster
+    forcefield = Forcefield.load_from_file("smirnoff_1_1_0_sc.py")
+    st = SingleTopology(mol_a, mol_b, core, forcefield)
+    solvent_sys, solvent_conf, solvent_box, solvent_top = builders.build_water_system(3.0, forcefield.water_ff)
+    num_host_atoms = solvent_conf.shape[0]
+    host_system, masses = convert_omm_system(solvent_sys)
+    solvent_host = Host(host_system, masses, solvent_conf, solvent_box, num_host_atoms)
+    for lamb in np.linspace(0.0, 1.0, num_windows, endpoint=True):
+        state = setup_initial_state(st, lamb, solvent_host, DEFAULT_TEMP, 2024)
+        water_sampler_nb_params = get_water_sampler_params(state)
+        nb_pot = next(p.potential for p in state.potentials if isinstance(p.potential, Nonbonded))
+        ligand_water_params = st._get_guest_params(
+            forcefield.q_handle_solv, forcefield.lj_handle_solv, lamb, nb_pot.cutoff
+        )
+
+        np.testing.assert_array_equal(water_sampler_nb_params[num_host_atoms:], ligand_water_params)
 
 
 def test_run_sims_bisection_early_stopping(hif2a_ligand_pair_single_topology_lam0_state):
--- a/timemachine/fe/free_energy.py
+++ b/timemachine/fe/free_energy.py
@@ -32,7 +32,7 @@ from timemachine.fe.stored_arrays import StoredArrays
 from timemachine.fe.utils import get_mol_masses, get_romol_conf
 from timemachine.ff import ForcefieldParams
 from timemachine.ff.handlers import openmm_deserializer
-from timemachine.lib import LangevinIntegrator, MonteCarloBarostat
+from timemachine.lib import LangevinIntegrator, MonteCarloBarostat, custom_ops
 from timemachine.lib.custom_ops import Context
 from timemachine.md.barostat.utils import compute_box_center, get_bond_list, get_group_indices
 from timemachine.md.hrex import (
@@ -44,8 +44,7 @@ from timemachine.md.hrex import (
     get_swap_attempts_per_iter_heuristic,
 )
 from timemachine.md.states import CoordsVelBox
-from timemachine.potentials import BoundPotential, HarmonicBond, SummedPotential
-from timemachine.potentials.potential import GpuImplWrapper
+from timemachine.potentials import BoundPotential, HarmonicBond, NonbondedInteractionGroup, SummedPotential
 from timemachine.utils import batches, pairwise_transform_and_combine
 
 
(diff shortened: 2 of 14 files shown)


timemachine/fe/free_energy.py
```diff
@@ -392,12 +427,64 @@ def _combine(self, ligand_values, host_values=None):
...
+        water_sampler = custom_ops.TIBDExchangeMove_f32(
+            initial_state.x0.shape[0],
+            initial_state.ligand_idxs.tolist(),
+            [water_group.tolist() for water_group in water_idxs],
+            water_params,
+            initial_state.integrator.temperature,
+            nb.beta,
+            nb.cutoff,
+            md_params.water_sampling_params.radius,
+            md_params.seed,
```
badisa, line 482: Should we add `+3` or do some sort of modification to the seed here?

timemachine/fe/free_energy.py
```diff
@@ -392,12 +427,64 @@ def _combine(self, ligand_values, host_values=None):
...
+    if md_params is not None and md_params.water_sampling_params is not None:
+        # Setup the water indices
+        hb_potential = next(p.potential for p in initial_state.potentials if isinstance(p.potential, HarmonicBond))
+        group_indices = get_group_indices(get_bond_list(hb_potential), len(initial_state.integrator.masses))
+
+        water_idxs = [group for group in group_indices if len(group) == 3]
+        # Handle the case where the ligand is a 3 atom system
+        if len(initial_state.ligand_idxs) == 3:
+            ligand_atom_set = set(initial_state.ligand_idxs)
+            water_idxs = [group for group in water_idxs if ligand_atom_set != set(group)]
```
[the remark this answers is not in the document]
badisa: Added `get_water_idxs` in https://github.com/proteneer/timemachine/pull/1286/commits/da706d324cad4e623b16bc8d28d2f161ceb7f146

maxentile (approved):

timemachine/fe/free_energy.py
```diff
@@ -951,30 +1022,35 @@ def log_q_fn(replica_idx: ReplicaIdx, state_idx: StateIdx) -> float:
...
-        box0 = boxes[0]
+            context.set_x_t(xvb.coords)
+            context.set_v_t(xvb.velocities)
+            context.set_box(xvb.box)
 
-        return CoordsVelBox(x0, v0, box0)
+            xs, boxes = context.multiple_steps(md_params.n_eq_steps, store_x_interval=0)
+            x0 = xs[0]
+            v0 = context.get_v_t()
+            box0 = boxes[0]
```
jkausrelay, line 1048: Is xs[0] == xs[-1] in this case? Just wondering why v0 is get_v_t() which would be the last frame right?

jkausrelay (approved):

timemachine/fe/free_energy.py
badisa: Yes, the length of `xs[0]` should be one, will add an assertion to be sure. Its only that we don't return the velocities from `multiple_steps`.

timemachine/fe/free_energy.py
proteneer: Probably? (Otherwise there's probably some minor correlation between the different RNGs)

timemachine/fe/free_energy.py
```diff
@@ -392,12 +428,65 @@ def _combine(self, ligand_values, host_values=None):
         return np.concatenate([host_values, ligand_values])
 
 
-def get_context(initial_state: InitialState) -> Context:
-    bound_impls = [p.to_gpu(np.float32).bound_impl for p in initial_state.potentials]
+def get_water_sampler_params(initial_state: InitialState) -> NDArray:
+    """Given an initial state, return the parameters that define the parameters of water-environment parameters
```
proteneer, line 432: Detail what environment here means, environment being both ligand + protein?

timemachine/fe/free_energy.py
```diff
@@ -392,12 +428,65 @@ def _combine(self, ligand_values, host_values=None):
...
+
+    Since we split the different components of the NB into ligand-water, ligand-protein, we want to use the ligand-water parameter
+    to ensure that the ligand component is correctly configured for the specific lambda window.
+    """
+    summed_pot = next(p.potential for p in initial_state.potentials if isinstance(p.potential, SummedPotential))
+    # TBD Figure out a better way of handling the jankyness of having to select the IxnGroup that defines the Ligand-Water
+    # Currently this just hardcodes to the first ixn group potential which is the ligand-water ixn group as returned by HostTopology.parameterize_nonbonded
```
proteneer, line 439: @jkausrelay - triple check this?

timemachine/fe/free_energy.py
```diff
@@ -392,12 +428,65 @@ def _combine(self, ligand_values, host_values=None):
         return np.concatenate([host_values, ligand_values])
 
 
-def get_context(initial_state: InitialState) -> Context:
-    bound_impls = [p.to_gpu(np.float32).bound_impl for p in initial_state.potentials]
+def get_water_sampler_params(initial_state: InitialState) -> NDArray:
```
proteneer, line 431: Is it possible to add a unit test for this? (Using initial states emitted from the SingleTopology/RBFE logic?) i.e. assert that the w-coords etc. for dummy atoms at the two end-states are being read-off correctly

timemachine/md/exchange/exchange_mover.py
```diff
@@ -13,6 +13,17 @@
 from timemachine.potentials import nonbonded
 
 
+def get_water_idxs(mol_groups: List[NDArray], ligand_idxs: Optional[NDArray] = None) -> List[NDArray]:
+    """Given a list of lists that make up the individual molecules in a system, return the subset that is only the waters.
+    Additional logic to handle the case where ligand_idxs is also of size 3.
```
proteneer, line 18: Additional logic is present or is needed (to handle cases where ligand_idxs == 3?)

timemachine/fe/free_energy.py
jkausrelay: This is correct for the current implementation (https://github.com/proteneer/timemachine/blob/05b1a7a787ed4bae0087926fdaf820f4ada648dc/timemachine/fe/topology.py#L776). @badisa  Can you add a note there to make sure to update the water sampling code if this is changed?

timemachine/fe/free_energy.py
badisa: Yes, will clarify

timemachine/fe/free_energy.py
```diff
@@ -78,6 +80,38 @@ def __post_init__(self):
...
 
 
+@dataclass(frozen=True)
+class WaterSamplingParams:
+    """
+    Parameters
+    ----------
+
+    interval:
+        How many steps of MD before running a set of water sampling moves
```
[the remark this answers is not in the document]
badisa: Before, if you set it to say 15, it will run 15 steps of the integrator before calling TIBD.

timemachine/md/exchange/exchange_mover.py
badisa: Is present, will clarify.

timemachine/fe/free_energy.py
badisa: Added comment to the topology file in https://github.com/proteneer/timemachine/pull/1286/commits/aeadaa0be1fd1430d9e5d816fb41d2d08831ee26

timemachine/md/exchange/exchange_mover.py
badisa: Clarified in https://github.com/proteneer/timemachine/pull/1286/commits/8fd9a326950e594cf1fc18aa177e1bad5729e724

timemachine/fe/free_energy.py
badisa: Removed that line, think the following sentence does a better job explaining it: https://github.com/proteneer/timemachine/pull/1286/commits/f940b6ad80ff6ccf492b05bb76c0167d791a1d09

timemachine/fe/free_energy.py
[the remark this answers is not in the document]
proteneer: Just so I understand the logic in `TIBDExchange::move()`:

```
if (N != this->N_) {
    throw std::runtime_error("N != N_");
}
this->step_++;
if (this->step_ % this->interval_ != 0) {
    return;
}
```
On every MD step we call move(), say 'interval_=1000', then 99.9% of the time this call ends up being a no-op? Doesn't this mean that we run 1000 steps of MD in **between** an actual water sampling move()?

timemachine/fe/free_energy.py
proteneer: A bit annoying/confusing since the internal Context call disallows `store_x_interval=0`:

```
std::array<std::vector<double>, 2> Context::multiple_steps(const int n_steps, int store_x_interval) {
    if (store_x_interval <= 0) {
        throw std::runtime_error("store_x_interval <= 0");
    }
```

But the internal wrapper code converts it to `n_steps`:

```        
.def(
    "multiple_steps",
    [](Context &ctxt, const int n_steps, int store_x_interval) -> py::tuple {
        // (ytz): I hate C++
        int x_interval = (store_x_interval <= 0) ? n_steps : store_x_interval;
        std::array<std::vector<double>, 2> result = ctxt.multiple_steps(n_steps, x
```

So like @badisa  - I think this is just returning the last frame.

timemachine/fe/free_energy.py
badisa: Added a test in https://github.com/proteneer/timemachine/pull/1286/commits/8ed067899ec3fcef0d619e486ad7e91a141032b2

timemachine/fe/free_energy.py
badisa: Yup and we do have this logic tested: https://github.com/proteneer/timemachine/blob/master/tests/test_md.py#L65-L70

timemachine/fe/free_energy.py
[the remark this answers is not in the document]
badisa: Oh sorry, I understand what you mean now :facepalm:. Will adjust

timemachine/fe/free_energy.py
badisa: Generates new seed on the fly in https://github.com/proteneer/timemachine/pull/1286/commits/5d10059bb7c71dd39c1e02b19cc2945c96067897

timemachine/fe/free_energy.py
```diff
@@ -78,6 +80,38 @@ def __post_init__(self):
...
+
+    n_proposals:
+        Number of proposals per make.
+
+    batch_size:
```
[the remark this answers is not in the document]
badisa: Adjusted in https://github.com/proteneer/timemachine/pull/1286/commits/a4e06e332b8c1d1d401a175a66976daa0ac8c3c7

timemachine/fe/free_energy.py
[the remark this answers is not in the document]
badisa: Fixed up in https://github.com/proteneer/timemachine/pull/1286/commits/a4e06e332b8c1d1d401a175a66976daa0ac8c3c7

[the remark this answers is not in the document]
badisa: Added assertion to be certain: https://github.com/proteneer/timemachine/pull/1286/commits/6267228611a8ff5c009c1f8663f025bca4d56840

tests/test_free_energy.py
```diff
@@ -289,6 +291,28 @@ def test_plot_pair_bar_plots(mock_fig, hif2a_ligand_pair_single_topology_lam0_st
...
+    solvent_host = Host(host_system, masses, solvent_conf, solvent_box, num_host_atoms)
+    for lamb in np.linspace(0.0, 1.0, num_windows, endpoint=True):
+        state = setup_initial_state(st, lamb, solvent_host, DEFAULT_TEMP, 2024)
+        water_sampler_nb_params = get_water_sampler_params(state)
+        nb_pot = next(p.potential for p in state.potentials if isinstance(p.potential, Nonbonded))
+        ligand_water_params = st._get_guest_params(
+            forcefield.q_handle_solv, forcefield.lj_handle_solv, lamb, nb_pot.cutoff
+        )
+
+        np.testing.assert_array_equal(water_sampler_nb_params[num_host_atoms:], ligand_water_params)
```
proteneer, line 313: can we sanity check that:

1) at lambda=0, dummy atoms of mol_b have w-coords set to cutoff?
2) at lambda=1, dummy_atoms of mol_a have w-coords set to cutoff?

you can use the look up the dummy_states from the single_topology.c_flags:

```
# 0: Core
# 1: R_A
# 2: R_B
```

badisa: Added in https://github.com/proteneer/timemachine/pull/1286/commits/c9ddadeb7d99f1546cc70c93737c18367609ebe4

proteneer (approved):

proteneer:
awesome :) we now have a legit water sampler!

PR #1287, opened 2024-04-13
Avoid repeatedly instantiating potentials

badisa:
* Re-using the preallocated potentials appears to get us back some performance

# Benchmarks
The reference refers to what is currently on master, the comparison is this branch. Generated using the test here: https://github.com/proteneer/timemachine/blob/1b8f07f102cdeab76f5b7e2126dcd8ba7bfcc49e/tests/test_benchmark_free_energy.py#L76 with the additional change to run with 2, 4, 8, 16, 32 and 48 windows as well as making sure every window always ran 100 frames. 

Run on an A10 GPU with Cuda Arch 8.6. Only sampled once, but think the differences are significant even without replicates.

Also ran a small graph (bisection + hrex + water sampling) that was about 5% faster with this change. 

## Vacuum
Doesn't make a difference on vacuum simulations, presumably the cost of constructing potentials is cheap (no neighborlist notably, tiny systems, etc)

image

## Solvent
Seems to make the biggest difference for bisection

image

### Without water sampling
image

### With water sampling
image

### DIFF
--- a/timemachine/fe/energy_decomposition.py
+++ b/timemachine/fe/energy_decomposition.py
@@ -44,7 +44,7 @@ def get_batch_u_fns(
 
     assert len(pots) == len(params)
     batch_u_fns: List[Batch_u_fn] = []
-    for p, bp in zip(params, pots):
+    for p, pot in zip(params, pots):
 
         def batch_u_fn(xs, boxes, pot_impl, pot_params):
             coords = np.array([x for x in xs])
@@ -61,7 +61,7 @@ def get_batch_u_fns(
             return us
 
         # extra functools.partial is needed to deal with closure jank
-        batch_u_fns.append(functools.partial(batch_u_fn, pot_impl=bp, pot_params=p[np.newaxis]))
+        batch_u_fns.append(functools.partial(batch_u_fn, pot_impl=pot, pot_params=p[np.newaxis]))
 
     return batch_u_fns
 
--- a/timemachine/fe/energy_decomposition.py
+++ b/timemachine/fe/energy_decomposition.py
@@ -1,12 +1,12 @@
 import functools
 from dataclasses import dataclass
-from typing import Callable, Generic, Iterable, List, Sequence, TypeVar
+from typing import Callable, Generic, List, Sequence, TypeVar
 
 import numpy as np
 from numpy.typing import NDArray
 
 from timemachine.constants import BOLTZ, DEFAULT_TEMP
-from timemachine.lib.custom_ops import BoundPotential
+from timemachine.lib.custom_ops import Potential
 
 Frames = TypeVar("Frames")
 Boxes = List[NDArray]
@@ -23,12 +23,17 @@ class EnergyDecomposedState(Generic[Frames]):
     batch_u_fns: Sequence[Batch_u_fn]  # u_fn : (frames, boxes) -> reduced_energies
 
 
-def get_batch_u_fns(bps: Iterable[BoundPotential], temperature: float = DEFAULT_TEMP) -> List[Batch_u_fn]:
+def get_batch_u_fns(
+    pots: Sequence[Potential],
+    params: Sequence[NDArray],
+    temperature: float = DEFAULT_TEMP,
+) -> List[Batch_u_fn]:
     """Get a list of functions that take in (coords, boxes), return reduced_potentials
 
     Parameters
     ----------
-    bps: list of bound potential impls
+    pots: list of potential impls
+    params: list of parameters to call potentials on
     temperature: float
 
     Returns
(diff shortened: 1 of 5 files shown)


maxentile (approved):
Changes look good to me!

Seeing similar timings locally with `pytest -k "test_benchmark_hif2a_single_topology"`

> Generated using the test here:
> ...
> with the additional change to run with 2, 4, 8, 16, 32 and 48 windows as well as making sure every window always ran 100 frames.

Tangent: The timings for each of these 3 modes at this grid of n_windows is very informative.

Would you mind adding these changes (+ either the plot or some other convenient export of the resulting timings) to test_benchmark_free_energy.py ? Will be good to track timings across this grid of conditions in the future.

A few observations about current timings. These are not necessarily surprising (most make sense in context), and are not immediately actionable, but they do deviate from some ideal:
* throughput for vacuum simulations (~6000-8000 ns/day) is only ~10x higher than for complex (~600-700 ns/day), even though the particle count is ~1000x smaller. (and only ~6x higher than for solvent (~900-1000 ns/day), even though the particle count is ~250x smaller)
* throughput decreases or stays constant with the number of windows. (in principle, should be able to achieve the same or higher throughput, since we increase the number of independent tasks to perform.)
* hrex overhead appears approximately linear in # windows. (Eyeballing: maybe a penalty of ~3-5 ns/day per window, relative to independent simulations?) Presumably, computing `u_kl` is the dominant cost, but I haven't verified.

Oddity: (probably inconsequential, just surprising)
* Vacuum sequential timings have an anomaly for small # windows ([2,4,8] windows, [~8000, ~5800, ~6200] ns/day)

badisa:
> Would you mind adding these changes (+ either the plot or some other convenient export of the resulting timings) to test_benchmark_free_energy.py ? Will be good to track timings across this grid of conditions in the future.
>

Added in https://github.com/proteneer/timemachine/pull/1287/commits/19b5c7d655e32c6ad2b3492717f0021dc9591298

badisa:
Out of curiosity I ran vacuum and solvent over night going out to 2000 frames to see if the performance differences changed. Only noticeable difference was that with solvent HREX only ~20% of ns/day was lost versus 65% lost in vacuum.  Which makes sense as solvent is more likely to be dominated by the cost of MD rather than by the cost of whatever work these methods do on their own.  

vacuum_2000_benchmarks
solvent_2000_benchmarks

timemachine/fe/energy_decomposition.py
```diff
@@ -37,17 +42,26 @@ def get_batch_u_fns(bps: Iterable[BoundPotential], temperature: float = DEFAULT_
     """
     kBT = temperature * BOLTZ
 
+    assert len(pots) == len(params)
     batch_u_fns: List[Batch_u_fn] = []
-    for bp in bps:
+    for p, bp in zip(params, pots):
```
mcwitt, line 47: nit: should we use a name other than `bp` for unbound potentials?

badisa: Good catch, will make it more explicit. Do we have a convention like we do for bound potentials?

timemachine/fe/free_energy.py
```diff
@@ -1141,19 +1137,25 @@ def format_rates(rs):
...
+    assert isinstance(potential, custom_ops.SummedPotential)
+    unbound_impls = potential.get_potentials()
+
     def make_energy_decomposed_state(
-        results: Tuple[StoredArrays, List[NDArray], InitialState]
+        results: Tuple[StateIdx, StoredArrays, List[NDArray]]
     ) -> EnergyDecomposedState[StoredArrays]:
-        frames, boxes, initial_state = results
+        state_idx, frames, boxes = results
+        initial_state = initial_states[state_idx]
```
mcwitt, line 1149: Curiosity: what was the motivation for the change to pass in a `StateIdx` and use it to look up the `initial_state`, versus directly passing in the `InitialState` as before?

timemachine/fe/free_energy.py
```diff
@@ -728,6 +726,7 @@ def run_sims_sequential(
     # u_kln matrix (2, 2, n_frames) for each pair of adjacent lambda windows and energy term
     u_kln_by_component_by_lambda = []
 
+    unbound_impls = [p.potential.to_gpu(np.float32).unbound_impl for p in initial_states[0].potentials]
```
mcwitt, line 729: Would it be worth adding a warning that this assumes all initial states have the same potentials (as [here](https://github.com/proteneer/timemachine/blob/7f1bffc7f24fd1d167b43608f1f1a0f6ae1a07fa/timemachine/fe/free_energy.py#L986))?

timemachine/fe/energy_decomposition.py
mcwitt: I'm not aware of a convention. I think I've mostly used `pot` or `potential` (or just `p` if the scope is local and no potential for confusion with parameters)

mcwitt (approved):
Looks great! Nice perf improvement!

timemachine/fe/energy_decomposition.py
badisa: Fixed up in all the places I could find: [1eff706](https://github.com/proteneer/timemachine/pull/1287/commits/1eff706be5cba8fe393e0eb2fb333d42c405a3da)

timemachine/fe/free_energy.py
badisa: Reverted to reduce diff: [da0f7f8](https://github.com/proteneer/timemachine/pull/1287/commits/da0f7f8f29d9c2576dc31bee7b49351d05e7cb28)

PR #1288, opened 2024-04-17
Highlight core bonds being formed and broken.

proteneer:
Improves visualization to support coloring bonds by core-bonds being formed/broken. This changes the default color scheme from red to green, and red is now going to be explicitly reserved for highlighting core-bonds.


maxentile (approved):
Nice! Increase informativeness of visualization, reduces need to scan for atom-color-pair correspondences visually.

badisa (approved):

PR #1292, opened 2024-04-24
Add additional dependencies for mypy in pre-commit

mcwitt:
Currently mypy, when run via pre-commit, treats imports from jax and other libraries as having type `Any`, resulting in it failing to catch some type errors. This is due to the combination of 1) pre-commit running checks in isolated environments without dependencies installed by default, and 2) the mypy pre-commit hook automatically passing `--ignore-missing-imports` in the mypy invocation. See this [issue comment](https://github.com/python/mypy/issues/13916#issue-1412194622) for more details.

This PR:
1. Adds the remaining dependencies that export types to `additional_dependencies` for mypy in the pre-commit configuration
2. Fixes any newly-exposed type errors

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,8 +27,13 @@ repos:
     rev: v1.5.1
     hooks:
     -   id: mypy
+        # include dependencies that export types (i.e. have a py.typed file in the root module) so that they can be used
+        # by mypy in pre-commit
         additional_dependencies:
+          - "jax==0.4.14"
           - "numpy==1.23.5"
+          - "scipy==1.10.1"
+          - "matplotlib==3.7.1"
 
         # Exclude custom_ops.py to work around clash with stub file when typechecking
         exclude: '^timemachine/lib/custom_ops.py$'


timemachine/potentials/nonbonded.py
```diff
@@ -129,7 +128,7 @@ def nonbonded_block_unsummed(
     lj = lennard_jones(dij, sig_ij, eps_ij)
 
     nrgs = jnp.where(dij < cutoff, es + lj, 0)
-    return nrgs
+    return cast(Array, nrgs)
```
badisa, line 131: curiosity: What type is this casting from?

badisa (approved):

timemachine/potentials/nonbonded.py
mcwitt: The return type of `jnp.where` is `Array | tuple[Array, ...]`, I think because when `where` is called with only one argument the behavior is different. In this case, we know it will return an `Array`.

timemachine/lib/fixed_point.py
```diff
@@ -1,13 +1,15 @@
...
 
 
-def fixed_to_float(v: int | jnp.uint64) -> jnp.float64:
+def fixed_to_float(v: ArrayLike) -> Array:
     """Meant to imitate the logic of timemachine/cpp/src/fixed_point.hpp::FIXED_TO_FLOAT"""
     return jnp.float64(jnp.int64(jnp.uint64(v))) / custom_ops.FIXED_EXPONENT
 
 
-def float_to_fixed(v: jnp.float32 | float) -> jnp.uint64:
+def float_to_fixed(v: ArrayLike) -> Array:
```
maxentile, line 13: These two seem more accurate but less precise

timemachine/potentials/nonbonded.py
```diff
@@ -474,7 +476,7 @@ def validate_coulomb_cutoff(cutoff=1.0, beta=2.0, threshold=1e-2):
 #   TODO: avoid repetition between this and lennard-jones
 
 
-def coulomb_prefactor_on_atom(x_i, x_others, q_others, box=None, beta=2.0, cutoff=jnp.inf) -> float:
+def coulomb_prefactor_on_atom(x_i, x_others, q_others, box=None, beta=2.0, cutoff=jnp.inf) -> Array:
```
maxentile, line 479: Agreed with the PR comment -- may be desirable later to restore an annotation that functions like this return a scalar / are non-broadcasting (i.e. they need to be transformed by vmap or similar). But the docstring + context seems sufficient.

maxentile (approved):
Good fixes!

(I agree with the notes in the PR description -- commented inline on two examples, but I don't think either are currently actionable)

PR #1293, opened 2024-04-24
Add utility method to emit RDKit mol from SingleTopology classes

proteneer:
Currently, rendering the ligand coordinates in the intermediate windows tends to result in a lot of mangled bond orders and difficult to visualize trajectories. This PR adds a utility function `SingleTopology::mol()` to emit an RDKit mol with a basic attempt at guessing whether or not a bond is present. For tracing replica trajectories in HREX, the SDF format is recommended (over PDB/mmCIF) since the bond orders are read explicitly by common viewers.


mcwitt (approved):
LGTM!

PR #1294, opened 2024-04-25
Improve performance of HREX swap moves

mcwitt:
Investigations done as part of #1287 showed that the performance of HREX falls off rapidly for larger numbers of windows. This is particularly pronounced in vacuum, where the overhead of the HREX moves is large compared to sampling and computation of the $u_{kl}$ matrix. See the right-hand panel of the following figure:

</img>

The rapid dropoff in vacuum performance is largely due to the rapid growth (as $L^3$, according to the [heuristic](https://github.com/proteneer/timemachine/blob/70706e301ab99d8f3dd513ccd0af2235f8c89e82/timemachine/md/hrex.py#L220) we currently use) of the number of swap attempts per HREX iteration with $L$ windows. E.g. with 48 windows, we do $48^3 \approx 10^5$ swap attempts per iteration. Because a batch of swap attempts is currently implemented as an unoptimized pure-Python for loop, there should be significant room for performance improvement in the implementation.

This PR adds a specialized implementation of a batch of neighbor swap moves that generates random numbers in batches and leverages JAX's jit compilation for performance. Benchmarks show that this implementation is significantly more performant (~50-100x) for large numbers of windows, although the initial compilation does introduce a small amount of overhead with a small number of windows.

### HREX MD timings
100 frames, no water sampling. As expected, performance improvement is significant for vacuum and marginal for solvent (where sampling and $u_{kl}$ computation is much more significant)

benchmarks

### Profiling permutation moves
With 4 windows, the overhead of JIT compilation is significant:

- [master, 4 windows](https://github.com/proteneer/timemachine/assets/319411/eca94d75-0bc6-427a-883c-992901cf668e)
- [branch, 4 windows](https://github.com/proteneer/timemachine/assets/319411/5946fdda-d83b-4082-93f6-ae2cea35e9df)

With 48 windows, optimized version is ~10x faster:

- [master, 48 windows](https://github.com/proteneer/timemachine/assets/319411/24a0e544-6f51-4e78-8e4b-5879ee8516bb)
- [branch, 48 windows](https://github.com/proteneer/timemachine/assets/319411/d64a94ba-a2a6-4869-afa0-47b7022ec78f)

Time to run a single permutation move (consisting of $L^3$ swap attempts), for the current version ("ref") and the version in this PR ("fast"). The time to compute the $u_{kl}$ matrix ("u_kl") is shown for reference. Note: this does not include the initial JIT compilation time for the fast version, which only runs once for a given number of windows and takes about half a second.

timings

### Other changes
- Removes `BatchedMixtureOfMoves` in favor of adding a method `move_n` to the base class. Subclasses can override the default definition of this and `sample_chain` to provide efficient implementations.

### TODO
- [x] Docstrings
- [ ] Update tests to also exercise reference version
- [ ] Test more directly comparing results of reference and fast version?

### DIFF
--- a/timemachine/md/hrex.py
+++ b/timemachine/md/hrex.py
@@ -70,7 +70,7 @@ def _run_neighbor_swaps(
         (n_pairs, 2) array representing allowed swaps
 
     log_q_kl : Array
-        (n_replicas, n_states) array with the (r, s) element giving the unnormalized probability of replica r in state s
+        (n_replicas, n_states) array with the (r, s) element giving the log unnormalized probability of replica r in state s
 
     pair_idxs : Array
         (n_swap_attempts,) array of indices of pairs for which to attempt swap moves
@@ -176,7 +176,7 @@ class HREX(Generic[Replica]):
             pairs of states between which to attempt swaps
 
         log_q : Callable[[ReplicaIdx, StateIdx], float]
-            function to compute the unnormalized probability of a given replica in a given state
+            function to compute the log unnormalized probability of a given replica in a given state
 
         n_swap_attempts : int
             number of individual swap attempts, each between a randomly-selected neighbor pair
@@ -208,7 +208,7 @@ class HREX(Generic[Replica]):
             pairs of states between which to attempt swaps
 
         log_q_kl : ArrayLike
-            (n_replicas, n_states) array with the (r, s) element giving the unnormalized probability of replica r in state s
+            (n_replicas, n_states) array with the (r, s) element giving the log unnormalized probability of replica r in state s
 
         n_swap_attempts : int
             number of individual swap attempts, each between a randomly-selected neighbor pair
--- a/timemachine/md/hrex.py
+++ b/timemachine/md/hrex.py
@@ -200,7 +200,7 @@ class HREX(Generic[Replica]):
         """Run a batch of swap attempts.
 
         See :py:meth:`attempt_neighbor_swaps` for a (typically slower) reference version.
-        Note that these methods do not generate compatible random sequences.
+        Note that these methods do not generate identical random sequences.
 
         Parameters
         ----------
--- a/timemachine/md/hrex.py
+++ b/timemachine/md/hrex.py
@@ -43,16 +43,12 @@ class NeighborSwapMove(MonteCarloMove[List[Replica]]):
 
 
 @jax.jit
-def run_neighbor_swaps(
+def _run_neighbor_swaps(
     keys: Array,
-    replica_idx_by_state: ArrayLike,
-    neighbor_pairs: ArrayLike,
-    log_q_kl: ArrayLike,
+    replica_idx_by_state: Array,
+    neighbor_pairs: Array,
+    log_q_kl: Array,
 ) -> Tuple[Array, Array, Array]:
-    replica_idx_by_state = jnp.asarray(replica_idx_by_state)
-    neighbor_pairs = jnp.asarray(neighbor_pairs)
-    log_q_kl = jnp.asarray(log_q_kl)
-
     def run_neighbor_swap(
         carry: Tuple[Array, Array, Array], key: PRNGKeyArray
     ) -> Tuple[Tuple[Array, Array, Array], None]:
@@ -142,8 +138,8 @@ class HREX(Generic[Replica]):
         key = jax.random.key(seed)
         keys = jax.random.split(key, n_swap_attempts)
 
-        replica_idx_by_state_, proposed, accepted = run_neighbor_swaps(
-            keys, jnp.asarray(self.replica_idx_by_state), jnp.asarray(neighbor_pairs), log_q_kl
+        replica_idx_by_state_, proposed, accepted = _run_neighbor_swaps(
+            keys, jnp.asarray(self.replica_idx_by_state), jnp.asarray(neighbor_pairs), jnp.asarray(log_q_kl)
         )
 
         replica_idx_by_state = replica_idx_by_state_.tolist()
--- a/timemachine/md/hrex.py
+++ b/timemachine/md/hrex.py
@@ -120,7 +120,7 @@ def _run_neighbor_swaps(
 Samples = TypeVar("Samples")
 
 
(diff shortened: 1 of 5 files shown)


maxentile:
> This PR adds a specialized implementation of a batch of neighbor swap moves that leverages JAX's jit compilation for performance. Benchmarks show that this implementation is significantly more performant (~10x) for large numbers of windows, although the initial compilation does introduce a small amount of overhead with a small number of windows.

Nice! And the attached profiling results at K=4 and K=48 are informative.

Could you please also attach to the PR description a **plot of total time to draw an approximate sample from `p(permutation | xs)` as a function of K**?

(x-axis: grid of K up to K=48, y-axis: time in seconds) 

(y-axis currently expected to be `(K^3 * avg_cost_per_swap_attempt) + (K^2 cost of computing u_kl matrix) + (any other per-HREX-iteration overheads)`, ignoring any one-time overheads such as JIT compilation... (unless JIT compiler needs to be invoked once per HREX iteration))

mcwitt:
> > This PR adds a specialized implementation of a batch of neighbor swap moves that leverages JAX's jit compilation for performance. Benchmarks show that this implementation is significantly more performant (~10x) for large numbers of windows, although the initial compilation does introduce a small amount of overhead with a small number of windows.
> 
> Nice! And the attached profiling results at K=4 and K=48 are informative.
> 
> Could you please also attach to the PR description a **plot of total time to draw an approximate sample from `p(permutation | xs)` as a function of K**?
> 
> (x-axis: grid of K up to K=48, y-axis: time in seconds)
> 
> (y-axis currently expected to be `(K^3 * avg_cost_per_swap_attempt) + (K^2 cost of computing u_kl matrix) + (any other per-HREX-iteration overheads)`, ignoring any one-time overheads such as JIT compilation... (unless JIT compiler needs to be invoked once per HREX iteration))

Good call. I added this plot to the PR description.

maxentile (approved):
Nice! Looks like this resolves the cost of performing the sequence of swap moves given `u_kl`, and the next bottleneck might be populating the `u_kl` matrix itself?

Unsure how much that can be accelerated in general, or if we might also want to mix in some other kinds of state permutation updates, e.g. the even-odd swaps used by default in https://manual.gromacs.org/current/reference-manual/algorithms/replica-exchange.html (which require `K` energy evaluations instead of `K*K` energy evaluations).

timemachine/md/hrex.py
```diff
@@ -69,16 +157,75 @@ def attempt_neighbor_swaps(
...
+
+        Parameters
+        ----------
+        neighbor_pairs : Sequence[Tuple[StateIdx, StateIdx]]
+            pairs of states between which to attempt swaps
+
+        log_q_kl : ArrayLike
+            (n_replicas, n_states) array with the (r, s) element giving the unnormalized probability of replica r in state s
+
+        n_swap_attempts : int
```
maxentile, line 208: Tangent: Not in diff, but I just noticed a small inconsistency: `n_swap_attempts` is `K^4` in docstring, `K^3` in implementation https://github.com/proteneer/timemachine/blob/aaa5868db4ef8f2c7c789aa207bb1bbaef003206/timemachine/md/hrex.py#L367-L384

timemachine/md/hrex.py
```diff
@@ -36,9 +44,89 @@ def propose(self, state: List[Replica]) -> Tuple[List[Replica], float]:
...
+        replica_idx_by_state, proposed, accepted = carry
+
+        key, subkey = jax.random.split(key)
+        pair_idx = jax.random.choice(subkey, len(neighbor_pairs))
+        s_a, s_b = neighbor_pairs[pair_idx]
+        proposed_next = proposed.at[pair_idx].add(1)
+
+        r_a = replica_idx_by_state[s_a]
+        r_b = replica_idx_by_state[s_b]
+        log_q_diff = log_q_kl[r_a, s_b] + log_q_kl[r_b, s_a] - log_q_kl[r_a, s_a] - log_q_kl[r_b, s_b]
```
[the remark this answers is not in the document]
mcwitt: Good call. Emphasized this by introducing intermediate names in https://github.com/proteneer/timemachine/pull/1294/commits/130b8a7b1e7c533fa1fdaa38a67910205d78558b.

timemachine/md/hrex.py
mcwitt: Good catch. Fixed up docstring in https://github.com/proteneer/timemachine/pull/1294/commits/8f5e4492cb91794d5fca49216597c34ba596cf96.

mcwitt:
Generating random indices and uniform noise in batches (https://github.com/proteneer/timemachine/pull/1294/commits/bb9e075985c0830b5d24e14ef76aa3219bbed355) yields another 5-10x performance improvement for permutation moves, for a total of 50-100x relative to master.

timemachine/md/hrex.py
```diff
@@ -69,16 +165,78 @@ def attempt_neighbor_swaps(
...
 
         return HREX(self.replicas, replica_idx_by_state), fraction_accepted_by_pair
 
+    def attempt_neighbor_swaps_fast(
+        self, neighbor_pairs: Sequence[Tuple[StateIdx, StateIdx]], log_q_kl: ArrayLike, n_swap_attempts: int
+    ) -> Tuple["HREX[Replica]", List[Tuple[int, int]]]:
+        """Run a batch of swap attempts.
+
+        See :py:meth:`attempt_neighbor_swaps` for a (typically slower) reference version.
+        Note that these methods do not generate compatible random sequences.
```
badisa, line 203: nit
```suggestion
        Note that these methods do not generate identical random sequences.
```

timemachine/md/moves.py
```diff
@@ -24,14 +24,26 @@ class Move(Generic[_State], ABC):
     def move(self, _: _State) -> _State:
         ...
 
+    def move_n(self, x: _State, n: int) -> _State:
```
badisa, line 27: Nice, much cleaner to have this generally be accessible by moves.

badisa (approved):
Awesome improvements, looks like it really improves vacuum HREX!

PR #1296, opened 2024-04-26
Add benchmark for water sampling

badisa:
* Want to add this prior to some other work to speed up water sampling so we can have a standard way of measuring the performance.

### DIFF
--- a/tests/test_benchmark.py
+++ b/tests/test_benchmark.py
@@ -446,7 +446,8 @@ def run_single_topology_benchmarks(
             initial_state.ligand_idxs,
         )
 
-        if host_config.num_water_atoms < initial_state.x0.shape[0] - st.get_num_atoms():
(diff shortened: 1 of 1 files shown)


tests/test_benchmark.py
```diff
@@ -377,6 +446,14 @@ def run_single_topology_benchmarks(
             initial_state.ligand_idxs,
         )
 
+        if host_config.num_water_atoms < initial_state.x0.shape[0] - st.get_num_atoms():
```
mcwitt, line 449: nit: maybe include a comment here to explain what this condition means?

mcwitt (approved):

tests/test_benchmark.py
badisa: Added comment and made the statement a bit clearer: https://github.com/proteneer/timemachine/pull/1296/commits/ceb798d4ad739590cca23d2d4142cb6347734a4d

PR #1302, opened 2024-04-30
Avoid setting global numpy PRNG seed

mcwitt:
Since #1294, the default HREX implementation no longer relies on numpy global PRNG state. This PR modifies the signature of `attempt_neighbor_swaps_fast` to accept an explicit seed rather than generating one using `numpy.random` and removes an associated warning.

Minor unrelated change: explicitly declare integer dtypes for proposed and accepted count arrays


proteneer:
awesome!

timemachine/fe/free_energy.py
```diff
@@ -1099,7 +1094,7 @@ def replica_from_samples(traj: Trajectory) -> CoordsVelBox:
         hrex, samples_by_state_iter = hrex.sample_replicas(sample_replica, replica_from_samples)
         log_q_kl = compute_log_q_matrix(hrex.replicas)
         hrex, fraction_accepted_by_pair = hrex.attempt_neighbor_swaps_fast(
-            neighbor_pairs, log_q_kl, n_swap_attempts_per_iter
+            neighbor_pairs, log_q_kl, n_swap_attempts_per_iter, md_params.seed
```
badisa, line 1097: Should this do `md_params.seed + iteration` to avoid generating the same sequence every time?

mcwitt: Yes, I think so! Good catch.

mcwitt:
Also noticed an unrelated bug (fixed in https://github.com/proteneer/timemachine/pull/1302/commits/4587c3d7932ae1075c429e15c6b088ba0ba1ed07) that I think resulted in `water_sampling_params` always being `None` in the test

timemachine/md/hrex.py
```diff
@@ -426,6 +423,9 @@ def run_hrex(
     n_samples_per_iter: int
         Number of local samples (e.g. MD frames) per HREX iteration
 
+    seed : int
```
badisa, line 426: super nit:
```suggestion
    seed: int
```

timemachine/md/hrex.py
```diff
@@ -213,13 +206,16 @@ def attempt_neighbor_swaps_fast(
         n_swap_attempts : int
             number of individual swap attempts, each between a randomly-selected neighbor pair
 
+        seed : int
```
badisa, line 209: super nit:
```suggestion
        seed: int
```

badisa (approved):
Nice, glad you cause that bug in the test.

timemachine/md/hrex.py
mcwitt: It seems we have a mixed convention. Fwiw, the numpydoc style guide includes the extra space: https://numpydoc.readthedocs.io/en/latest/format.html

Maybe I'll update so we're at least consistent within a given docstring for now?

maxentile (approved):

timemachine/md/hrex.py
mcwitt: I applied the other suggested fix. Will leave this one as-is for now for consistency.

PR #1303, opened 2024-04-30
Sample velocities when setting up initial states

badisa:
* Adds a seed to avoid using global RNG
* We have talked about this, seems like a reasonable improvement over zeros


timemachine/md/thermostat/utils.py
```diff
@@ -1,15 +1,38 @@
...
+    """
     n_particles = len(masses)
     spatial_dim = 3
 
-    v_unscaled = np.random.randn(n_particles, spatial_dim)
+    if seed is not None:
+        rng = np.random.default_rng(seed)
+        v_unscaled = rng.standard_normal(size=(n_particles, spatial_dim))
+    else:
+        v_unscaled = np.random.randn(n_particles, spatial_dim)
```
badisa, line 35: Can remove use of global RNG, wasn't sure if this is being relied on else where.

timemachine/md/thermostat/utils.py
```diff
@@ -1,15 +1,36 @@
...
+
+    Returns
+    -------
+    (N, 3) velocities array, where N is the length of masses
+    """
     n_particles = len(masses)
     spatial_dim = 3
 
-    v_unscaled = np.random.randn(n_particles, spatial_dim)
+    seed = seed if (seed is not None) else np.random.randint(100000)
```
proteneer, line 31: Should probably be explicit about the seed

badisa: Made explicit in https://github.com/proteneer/timemachine/pull/1303/commits/64cecdca7c28a9900dbd81a7c0f42b95a2ecf5f7

proteneer (approved):

PR #1307, opened 2024-05-06
Adds Nightly Memcheck job

badisa:
* Some of the memcheck tests are getting expensive and taking more time than the other jobs. Moved them to their own separate jobs where it seemed like the memcheck was not urgent
* Moved the optimized mtm test to nightly since it is relatively slow and we don't use this code much in practice
* Fixes up some warnings from pytest with the changes I had made in https://github.com/proteneer/timemachine/pull/1287/ that would have eventually turned into a pytest failure
* Reduces current CI from an hour 30 to about an hour 10. Nightlies stay around 6 hours

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -142,6 +142,24 @@ nightly-tests:
     when: on_success
     expire_in: 1 week
 
+nightly-memcheck-tests:
+  stage: test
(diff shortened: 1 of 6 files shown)


mcwitt (approved):
Nice, LGTM!

PR #1309, opened 2024-05-20
Reduce cost of u_kl computation in HREX

mcwitt:
Builds on #1308, using `Potential::execute_batch_sparse` introduced in the latter to avoid computing (very likely) unused elements of $u_{kl}$ corresponding to far-away states for a given replica. This introduces a new parameter ~`n_neighbor_states`~ `max_delta_states` controlling the number of neighbors on either side of a replica's current state for which to evaluate the potential; 4 seems to be a reasonable value empirically. With 48 windows, this would result in 1/6th of the potential evaluations we do currently, potentially yielding a significant speed boost.

With current defaults, distribution of the number of states a given replica is displaced during a single HREX permutation move (consisting of $K^3$ swap attempts). The probability mass outside the interval [-4, +4] is insignificant:

image

Note: only `n_neighbors >= 1` is needed for correctness; as long as this is satisfied, choosing a number that significantly truncates the distributions above will result in slower mixing, but still asymptotically correct sampling.

<a id="benchmarks"></a>
## Benchmarks

- system: `timemachine.testsystems.relative.get_hif2a_ligand_pair_single_topology`
- default parameters w/ HREX, except setting `n_frames=100`
- 3 replicates
- `a` is current master branch; `b` is this branch

image


timemachine/fe/free_energy.py
```diff
@@ -898,6 +904,61 @@ def compute_intermediate_result(lambdas: Sequence[float]) -> PairBarResult:
...
+
+    coords = np.array([xvb.coords for xvb in hrex.replicas])
+    boxes = np.array([xvb.box for xvb in hrex.replicas])
+
+    def compute_sparse(k: int):
+        n_states = len(hrex.replicas)
+        state_idx = np.argsort(hrex.replica_idx_by_state)
+        neighbor_state_idxs = state_idx[:, None] + np.arange(-k, k + 1)[None, :]
+        valid_idxs = np.nonzero((0 <= neighbor_state_idxs) & (neighbor_state_idxs < n_states))
+        coords_batch_idxs = valid_idxs[0].astype(np.uint32)
```
badisa, line 941: Are the coordinates sorted such that a series of parameter evaluations can be done without a neighborlist rebuild? Not clear to me from the code, but if not then might explain why there isn't a performance improvement. Or perhaps the cost of recomputing energies with different parameters is negligible compared to neighborlist rebuilds (though less confident of that)

mcwitt: Hmm, actually it turns out that the ordering happens to already be what I'd imagine is optimal for avoiding neighborlist rebuilds (i.e. outer loop over coordinates in replica order and inner loop over parameters), so I'm not sure if this is the issue after all. Will keep investigating.

mcwitt: After more careful [benchmarking](#benchmarks) with 3 replicates, the performance boost does seem mostly as expected. I think I was not careful enough with the initial benchmarks (for instance might have run other processes concurrently while the benchmarks were running).

One feature I'm not sure I understand yet is why, for solvent specifically, the ns_per_day measurement does not flatten out for larger number of windows (though it's still a significant improvement).

maxentile (approved):
Nice optimization, cleanly implemented and tested

timemachine/fe/free_energy.py
```diff
@@ -68,14 +68,20 @@ class HREXParams:
 
     n_frames_per_iter:
         Number of frames to sample using MD per HREX iteration.
+
+    n_neighbor_states:
+        Number of neighbor states on either side of a replica's current state for which to compute U(replica, state).
+        This determines the maximum distance a replica can move in a single HREX iteration.
```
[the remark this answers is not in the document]
mcwitt: Agreed - `max_delta_states` seems like an improvement to me. Maybe I'll go with that unless we have a better idea in the meantime?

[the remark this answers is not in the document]
mcwitt: Changed to `max_delta_states` in https://github.com/proteneer/timemachine/pull/1309/commits/d3ae3c578bd6fc196658a0b3e43d77020581508f

tests/test_free_energy.py
```diff
@@ -405,3 +414,50 @@ def test_estimate_free_energy_bar_with_energy_overflow():
     assert result_with_nan.dG_err == result_with_inf.dG_err
     np.testing.assert_array_equal(result_with_nan.dG_err_by_component, result_with_inf.dG_err_by_component)
     np.testing.assert_array_equal(result_with_nan.overlap, result_with_inf.overlap)
+
+
+@pytest.mark.parametrize("n_states,n_neighbor_states", [(1, 1), (1, None), (1, 2), (2, 1), (6, 3), (6, None), (30, 5)])
```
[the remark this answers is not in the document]
mcwitt: Good call; added in https://github.com/proteneer/timemachine/pull/1309/commits/d716418d876605bd7cdad257d1d8446081ece62a

tests/test_free_energy.py
```diff
@@ -405,3 +414,50 @@ def test_estimate_free_energy_bar_with_energy_overflow():
...
+
+@pytest.mark.parametrize(
+    "n_states,max_delta_states", [(1, 1), (1, None), (1, 2), (2, 1), (3, 6), (6, 3), (6, None), (30, 5)]
+)
+@pytest.mark.parametrize("seed", [2024, 2025])
+def test_compute_potential_matrix(hif2a_ligand_pair_single_topology, n_states: int, max_delta_states: int | None, seed):
+    st, _ = hif2a_ligand_pair_single_topology
+    states = [st.setup_intermediate_state(lam) for lam in np.linspace(0.0, 1.0, n_states)]
+
+    def make_summed_potential(bps: List[BoundPotential]):
```
badisa, line 427: Not for this pr: Would be good to have this utility function somewhere, its very handy.

timemachine/fe/free_energy.py
```diff
@@ -898,6 +905,61 @@ def compute_intermediate_result(lambdas: Sequence[float]) -> PairBarResult:
...
 
 
+def compute_potential_matrix(
+    potential: custom_ops.Potential,
+    hrex: HREX[CoordsVelBox],
+    params_by_state: NDArray,
+    max_delta_states: Optional[int] = None,
+) -> NDArray:
+    """Computes the (n_replicas, n_states) sparse matrix of potential energies, where a given element $(k, l)$ is
+    computed iff state $l$ is within `max_delta_states` of the current state of replica $k$, and is otherwise set to
```
badisa, line 915:
```suggestion
    computed if state $l$ is within `max_delta_states` of the current state of replica $k$, and is otherwise set to
```

badisa (approved):
Might be a good follow up to use this same approach for the final bar computation rather than computing each pair in sequence. https://github.com/proteneer/timemachine/blob/master/timemachine/fe/free_energy.py#L1157

Though since that is the full number of frames, might have some memory concerns.

timemachine/fe/free_energy.py
mcwitt: This was intended as an abbreviation for "if and only if". Removed the abbreviation in https://github.com/proteneer/timemachine/pull/1309/commits/99fc96d6c939ec669cdd37d029e074032a8b4c7f for clarity.

badisa: Ah, that would have been fine actually, just read it as a typo.

PR #1311, opened 2024-05-25
Avoid reading in an invalid sample, triggering an invalid memory access

badisa:
* Doesn't impact correctness of water sampling.
* Happens when the number of proposals left is fewer than the batch size. In this case the samples that are selected for the idx in the batch are stale and may be sampling invalid states. In the case that the final mol (by idx) was selected and it was a targeted inner move then the index of the new sample would be out of bounds triggering a failure. A very rare occurrence which is why this was not caught initially.

### DIFF
--- a/timemachine/cpp/src/kernels/k_exchange.cu
+++ b/timemachine/cpp/src/kernels/k_exchange.cu
@@ -759,19 +759,11 @@ void __global__ k_adjust_sample_idxs(
     const int local_inner_count = *inner_count;
     int idx = blockIdx.x * blockDim.x + threadIdx.x;
 
-    while (idx < batch_size) {
-        if (idx + current_offset >= total_proposals) {
-            // The value here is not important, as the result will be ignored in subsequent kernels
-            // This is done to avoid the fact that the original sampled indices may contain invalid state
-            // thanks to the way the batching performed.
-            sample_idxs[idx] = 0;
-        } else {
-            const int target_inner = targeting_inner_volume[idx];
-            const int offset = target_inner == 1 ? local_inner_count : 0;
-            const int before = sample_idxs[idx];
-            sample_idxs[idx] = partitioned_indices[before + offset];
-            idx += gridDim.x * blockDim.x;
-        }
+    while (idx < batch_size && idx + current_offset < total_proposals) {
+        const int target_inner = targeting_inner_volume[idx];
+        const int offset = target_inner == 1 ? local_inner_count : 0;
+        const int before = sample_idxs[idx];
+        sample_idxs[idx] = partitioned_indices[before + offset];
         idx += gridDim.x * blockDim.x;
     }
 }
--- a/timemachine/cpp/src/kernels/k_exchange.cu
+++ b/timemachine/cpp/src/kernels/k_exchange.cu
@@ -761,7 +761,7 @@ void __global__ k_adjust_sample_idxs(
 
     while (idx < batch_size) {
         if (idx + current_offset >= total_proposals) {
-            // The value here is not important, as the result will be ignored in subsuequent kernels
+            // The value here is not important, as the result will be ignored in subsequent kernels
             // This is done to avoid the fact that the original sampled indices may contain invalid state
             // thanks to the way the batching performed.
             sample_idxs[idx] = 0;
--- a/timemachine/cpp/src/kernels/k_exchange.cu
+++ b/timemachine/cpp/src/kernels/k_exchange.cu
@@ -752,7 +752,7 @@ void __global__ k_adjust_sample_idxs(
     const int *__restrict__ noise_offset,           // [1]
     const int *__restrict__ targeting_inner_volume, // [batch_size]
     const int *__restrict__ inner_count,            // [1]
-    const int *__restrict__ partitioned_indices,    // [K]
+    const int *__restrict__ partitioned_indices,    // [num_samples]
     int *__restrict__ sample_idxs                   // [batch_size]
 ) {
     const int current_offset = *noise_offset;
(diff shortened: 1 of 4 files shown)


timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -747,20 +747,31 @@ template void __global__ k_setup_destination_weights_for_targeted<double>(
...
     while (idx < batch_size) {
-        const int target_inner = targeting_inner_volume[idx];
-        const int offset = target_inner == 1 ? local_inner_count : 0;
-        const int before = sample_idxs[idx];
-        sample_idxs[idx] = partitioned_indices[before + offset];
+        if (idx + current_offset >= total_proposals) {
+            // The value here is not important, as the result will be ignored in subsuequent kernels
+            // This is done to avoid the fact that the original sampled indices may contain invalid state
+            // thanks to the way the batching performed.
+            sample_idxs[idx] = 0;
```
mcwitt, line 767: Curiosity: what's the effect of setting the value here, if the value is unimportant / ignored? Naively it seems like just guarding the code that is now in the "else" block would be sufficient.

mcwitt (approved):
Good catch! Had some minor comments and questions, but feel free to address these after merging.

timemachine/cpp/src/kernels/k_exchange.cu
badisa: You are right, don't actually need to change it, improved in https://github.com/proteneer/timemachine/pull/1311/commits/bbd65efa03a76cd6094d22c256bfd68b17d1c2cb

Also fixes a bug if the number of blocks was not `ceil_divide(num_samples, tpb)`

timemachine/cpp/src/kernels/k_exchange.cuh
```diff
@@ -201,10 +201,12 @@ void __global__ k_setup_destination_weights_for_targeted(
     RealType *__restrict__ output_weights);
 
 void __global__ k_adjust_sample_idxs(
```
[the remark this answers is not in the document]
badisa: Documented in https://github.com/proteneer/timemachine/pull/1311/commits/12b9cd2cf47d0a3dc7bb1b29e1357e717cab813f

timemachine/cpp/src/kernels/k_exchange.cuh
```diff
@@ -201,10 +201,12 @@ void __global__ k_setup_destination_weights_for_targeted(
...
+    const int total_proposals,
     const int batch_size,
+    const int *__restrict__ noise_offset,           // [1]
     const int *__restrict__ targeting_inner_volume, // [batch_size]
     const int *__restrict__ inner_count,            // [1]
-    const int *__restrict__ partitioned_indices,    // [inner_count]
+    const int *__restrict__ partitioned_indices,    // [K]
```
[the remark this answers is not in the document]
badisa: Adjust to be `num_samples` in https://github.com/proteneer/timemachine/pull/1311/commits/be3231c93df3d3efb02b34b6da99315235c676e3

It is not a variable, intended as a human readable value. No clue why I picked K, as that isn't helpful

proteneer (approved):

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -747,16 +747,19 @@ template void __global__ k_setup_destination_weights_for_targeted<double>(
...
+    const int *__restrict__ partitioned_indices,    // [num_samples]
     int *__restrict__ sample_idxs                   // [batch_size]
 ) {
-    const int local_inner_count = inner_count[0];
+    const int current_offset = *noise_offset;
+    const int local_inner_count = *inner_count;
     int idx = blockIdx.x * blockDim.x + threadIdx.x;
 
-    while (idx < batch_size) {
+    while (idx < batch_size && idx + current_offset < total_proposals) {
```
proteneer, line 762: trying to follow the out of bounds logic here:

`partitioned_indices is of size num_samples`
`sample_idxs` is of size `batch_size`

So while `sample_idxs[idx]` will never be out of bounds, the accessed element `before` may be out of bounds when used to index into `before+offset`?

badisa: The accessed element in `partitioned_indices[before + offset]` was what could be wrong. If `before == num_outer + 1`then it would crash. The value of `before` could be larger than expected since upstream we don't recompute the `local_inner_count` value once we have gone beyond the total number of proposals.

timemachine/cpp/src/kernels/k_exchange.cu
```diff
@@ -747,16 +747,19 @@ template void __global__ k_setup_destination_weights_for_targeted<double>(
 );
 
 void __global__ k_adjust_sample_idxs(
+    const int total_proposals,
     const int batch_size,
+    const int *__restrict__ noise_offset,           // [1]
     const int *__restrict__ targeting_inner_volume, // [batch_size]
     const int *__restrict__ inner_count,            // [1]
-    const int *__restrict__ partitioned_indices,    // [inner_count]
+    const int *__restrict__ partitioned_indices,    // [num_samples]
```
proteneer, line 755: nit: might be good to standardize on `_indices` vs `_idxs` at some point

proteneer:
did cuda-memcheck catch the illegal out of access before the patch was applied?

badisa:
> did cuda-memcheck catch the illegal out of access before the patch was applied?

Yes, the added test does fail under `compute-sanitizer` (the latest version of cuda-memcheck) without this change.

timemachine/cpp/src/kernels/k_exchange.cu
proteneer: is checking `idx + current_offset < total_proposals` equivalent to checking `before + offset < num_samples`?

proteneer: (the latter is more obvious that it prevents an out of bounds read) - less clear still on how it's guaranteed / implied that `idx + current_offset < num_samples`

badisa: No, `num_samples` is `batch_size`. `total_proposals` is equal to `proposals_per_move` in the TIBD mover and the current offset is the number proposals already made.

PR #1313, opened 2024-05-28
Adopt dummy values for segmented sumexp that compile in latest cub

badisa:
* As apart of benchmarking https://github.com/proteneer/timemachine/pull/1312 on L4s, I found that I could not compile timemachine on the latest version of cub (shipped with Cuda 12.5.0). These were the changes necessary to compile the repo.


timemachine/cpp/src/segmented_sumexp.cu
```diff
@@ -13,22 +13,21 @@ SegmentedSumExp<RealType>::SegmentedSumExp(const int max_vals_per_segment, const
     : max_vals_per_segment_(max_vals_per_segment), num_segments_(num_segments),
       d_temp_buffer_(max_vals_per_segment_ * num_segments_), temp_storage_bytes_(0), d_temp_storage_buffer_(0) {
     void *dummy_temp = nullptr;
+    int *dummy_offsets = nullptr;
```
badisa, line 16: Newer version of cub no longer excepts non-integral (ie void) types.

proteneer:
Does this work with older versions of cub still? If not, should we also update the README to require Cuda 12.5+?

badisa:
> Does this work with older versions of cub still? If not, should we also update the README to require Cuda 12.5+?

Yup, the CI is still Cuda 11. Think we should look at upgrading Cuda, but this PR doesn't require it.

proteneer (approved):

mcwitt (approved):

PR #1314, opened 2024-05-28
Adds normalized kl divergence metric

badisa:
Had previously looked at the KL Divergence for detecting bottlenecks in HREX and it seemed to grow unbounded with the number of windows, which is a problem when wanting to compare edges with different number of windows. Normalizing it appears to do the right thing

## Unnormalized
image

## Normalized
image

### DIFF
--- a/tests/hrex/test_hrex.py
+++ b/tests/hrex/test_hrex.py
@@ -149,3 +149,23 @@ def test_normalized_kl_divergence_with_bottleneck(n_windows, frames):
     hrex_matrix = simulate_bottlenecked_hrex(n_windows, frames)
     res = get_normalized_kl_divergence(hrex_matrix)
     assert res >= 0.5
+
+
+@pytest.mark.parametrize("seed", [2024])
+@pytest.mark.parametrize("iterations", [10])
+@pytest.mark.parametrize("n_windows,frames", [(5, 2000), (16, 2000), (48, 2000)])
+def test_normalized_kl_divergence_relative_values(iterations, n_windows, frames, seed):
+    """Ensure that the mean of multiple replicates of normalized KL divergence produce the right ordering.
+
+    Perfect sampling should have the lowest value, followed by slow mixing followed up by bottle necked mixing"""
+    np.random.seed(seed)
+    perfect_mixing_kl = []
+    slow_mixing_kl = []
+    bottlenecked_kl = []
+    for _ in range(iterations):
+        perfect_mixing_kl.append(get_normalized_kl_divergence(simulate_perfect_mixing_hrex(n_windows, frames)))
+        bottlenecked_kl.append(get_normalized_kl_divergence(simulate_bottlenecked_hrex(n_windows, frames)))
+        slow_mixing_kl.append(get_normalized_kl_divergence(simulate_slow_mixing_hrex(n_windows, frames)))
+
+    assert np.mean(perfect_mixing_kl) < np.mean(slow_mixing_kl)
+    assert np.mean(slow_mixing_kl) < np.mean(bottlenecked_kl)
--- a/tests/hrex/test_hrex.py
+++ b/tests/hrex/test_hrex.py
@@ -48,17 +48,16 @@ def test_get_samples_by_iter_by_replica_invalid_args():
 
 def simulate_perfect_mixing_hrex(num_states: int, num_frames: int) -> List[List[int]]:
     """assume every step of HREX perfectly mixed all replicas"""
-    rng = np.random.default_rng(num_states)
     inds = np.arange(num_states)
     traj = []
     for _ in range(num_frames):
-        rng.shuffle(inds)
+        np.random.shuffle(inds)
         traj.append(np.array(inds).tolist())
     return traj
 
 
 def simulate_slow_mixing_hrex(num_states: int, num_frames: int) -> List[List[int]]:
-    """assume every step of HREX only succeeds in making K nearest-neighbor swaps"""
+    """assume every step of HREX only succeeds in making num_states nearest-neighbor swaps"""
     traj = [np.arange(num_states).tolist()]
     for _ in range(num_frames - 1):
         current_state = np.array(traj[-1])
@@ -77,7 +76,7 @@ def simulate_slow_mixing_hrex(num_states: int, num_frames: int) -> List[List[int
 
 
 def simulate_bottlenecked_hrex(num_states: int, num_frames: int) -> List[List[int]]:
-    """simulate_slow_mixing_hrex, but there's a state near K/2 that never swaps with neighbors"""
+    """simulate_slow_mixing_hrex, but there's a state near num_states/2 that never swaps with neighbors"""
     traj = [np.arange(num_states).tolist()]
 
     bottleneck_i = int(round(num_states / 2))
@@ -103,6 +102,7 @@ def simulate_no_mixing_hrex(num_states: int, num_frames: int) -> List[List[int]]
     return traj
 
 
+@pytest.mark.parametrize("seed", [2024])
 @pytest.mark.parametrize("n_windows,frames", [(3, 2000), (16, 2000), (48, 2000)])
 @pytest.mark.parametrize(
     "simulator",
@@ -114,8 +114,9 @@ def simulate_no_mixing_hrex(num_states: int, num_frames: int) -> List[List[int]]
         simulate_no_mixing_hrex,
     ],
 )
-def test_normalized_kl_divergence(simulator, n_windows, frames):
+def test_normalized_kl_divergence(simulator, n_windows, frames, seed):
     """Verify that given any of the expected hrex simulations produces values that are greater than 0.0"""
+    np.random.seed(seed)
     hrex_matrix = simulator(n_windows, frames)
     res = get_normalized_kl_divergence(hrex_matrix)
     assert res >= 0.0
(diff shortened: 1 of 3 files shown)


tests/hrex/test_hrex_rbfe.py
```diff
@@ -114,6 +114,9 @@ def sample_and_record_rss(*args, **kwargs):
     final_replica_state_counts = result.hrex_diagnostics.cumulative_replica_state_counts[-1]
     assert np.any(np.all(final_replica_state_counts > 0, axis=0))
 
+    assert isinstance(result.hrex_diagnostics.relaxation_time, float)
```
badisa, line 117: Noticed we don't actually call this code anywhere, not sure what a good test for it would be, but thought at least verifying that it doesn't trigger an exception would be good.

timemachine/md/hrex.py
```diff
@@ -238,6 +239,26 @@ def state_replica_pairs(self) -> List[Tuple[StateIdx, Replica]]:
...
+    ----------
+    [1]: https://doi.org/10.1021/acs.jctc.0c00660, https://chemrxiv.org/engage/chemrxiv/article-details/60c74d2e702a9b007018b7ef
+    """
+    cumulative_counts = get_cumulative_replica_state_counts(replica_idx_by_state_by_iter)
+    n_iters, n_states, n_replicas = cumulative_counts.shape
+    count_by_replica_by_state = cumulative_counts[-1]
+    fraction_by_replica_by_state = count_by_replica_by_state / n_iters
+    uniform_dist = np.ones_like(fraction_by_replica_by_state) / n_states
+    # No need to add eps to the fraction_by_replica_state, as kel_entr returns a KL of 0 in that case
+    return np.sum(rel_entr(fraction_by_replica_by_state, uniform_dist)) / n_states
```
mcwitt, line 259: A few notes I had when trying to interpret this; unsure if they would help clarify the implementation:

1. It made sense to me to think of this as averaging over replicas the KL divergence between the state distribution for a given replica and the uniform distribution (rather than the KL divergence between 2-d distributions). I.e. thinking of it as the (equivalent)

    ```python
    np.mean(np.sum(rel_entr(fraction_by_replica_by_state, uniform_dist), axis=1))
    ```

3. This is also seems equivalent to

    ```python
    np.mean(entropy(fraction_by_replica_by_state, uniform_dist, axis=1))
    ```

5. Finally, I think you can avoid constructing the uniform distribution and just write this as

    ```python
    -np.mean(entropy(fraction_by_replica_by_state, axis=1)) + np.log(n_states)
    ```

    (using the identity $\sum_i p_i \log (p_i / u_i) =  \sum_i p_i \log p_i + \log N = -H_p + \log N$, where $H_p$ is the entropy of $p$)

mcwitt (commented):
Nice! I think this will be extremely useful for troubleshooting slow/bottlenecked mixing issues.

timemachine/md/hrex.py
badisa: Nice! The last variant is great, added it in https://github.com/proteneer/timemachine/pull/1314/commits/43a6b9a54f7bee3dc09477ef6a0bb5fe7edd0fd7 and added a note about the construction.

timemachine/md/hrex.py
```diff
@@ -238,6 +239,26 @@ def state_replica_pairs(self) -> List[Tuple[StateIdx, Replica]]:
         return [(StateIdx(i), self.replicas[replica_idx]) for i, replica_idx in enumerate(self.replica_idx_by_state)]
 
 
+def get_normalized_kl_divergence(replica_idx_by_state_by_iter: Sequence[Sequence[ReplicaIdx]]) -> float:
+    """Heuristic for the how uniformly windows were sampled within an HREX simulation.
+    Based on Eq. 5 from [1], but after summing up the divergences of each state, take the mean
+    so that it is possible to compare the values between simulations of different numbers of windows.
+
+    A value closer to 0.0 indicates more uniforming sampling.
```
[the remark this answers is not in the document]
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1314/commits/43a6b9a54f7bee3dc09477ef6a0bb5fe7edd0fd7

timemachine/md/hrex.py
mcwitt: One last option you might prefer:

```python
-entropy(fraction_by_replica_by_state, axis=None) + 2 * np.log(n_states)
```

(`scipy.stats.entropy` has a sneaky default of `axis=0`, which seems different from typical the numpy convention; might prefer to use the above or explicitly pass `axis=0` if it looks confusing)

tests/hrex/test_hrex.py
```diff
@@ -40,3 +44,105 @@ def test_get_samples_by_iter_by_replica_invalid_args():
         get_samples_by_iter_by_replica([[1], [2, 3]], [[ReplicaIdx(0)], [ReplicaIdx(0)]])
     with pytest.raises(AssertionError):
         get_samples_by_iter_by_replica([[1, 2], [3, 4]], [[ReplicaIdx(0)], [ReplicaIdx(0), ReplicaIdx(1)]])
+
+
+def simulate_perfect_mixing_hrex(num_states: int, num_frames: int) -> List[List[int]]:
+    """assume every step of HREX perfectly mixed all replicas"""
+    rng = np.random.default_rng(num_states)
```
maxentile, line 51: nit: unexpectedly frozen random function -- consider accepting rng or seed as argument if needed for explicitness (or depending on global numpy random state as in the next 2 fxns)

(otherwise, this fxn will produce the same sequence when called multiple times, which may cause testing bugs if we attempt to run multiple trials)

tests/hrex/test_hrex.py
```diff
@@ -40,3 +44,105 @@ def test_get_samples_by_iter_by_replica_invalid_args():
...
+    inds = np.arange(num_states)
+    traj = []
+    for _ in range(num_frames):
+        rng.shuffle(inds)
+        traj.append(np.array(inds).tolist())
+    return traj
+
+
+def simulate_slow_mixing_hrex(num_states: int, num_frames: int) -> List[List[int]]:
+    """assume every step of HREX only succeeds in making K nearest-neighbor swaps"""
```
maxentile, line 61: `K` -> `num_states`

(docstring matched signature `(K, T)`, signature now `(num_states, num_frames)`)

tests/hrex/test_hrex.py
```diff
@@ -40,3 +44,105 @@ def test_get_samples_by_iter_by_replica_invalid_args():
...
+
+            current_state[j] = x_i
+            current_state[i] = x_j
+
+        traj.append(current_state.tolist())
+    return traj
+
+
+def simulate_bottlenecked_hrex(num_states: int, num_frames: int) -> List[List[int]]:
+    """simulate_slow_mixing_hrex, but there's a state near K/2 that never swaps with neighbors"""
```
maxentile, line 80: `K` -> `num_states`

tests/hrex/test_hrex.py
```diff
@@ -40,3 +44,105 @@ def test_get_samples_by_iter_by_replica_invalid_args():
...
+    res = get_normalized_kl_divergence(hrex_matrix)
+    assert res >= 1.0
+
+
+@pytest.mark.parametrize("n_windows,frames", [(16, 2000), (48, 2000)])
+def test_normalized_kl_divergence_with_bottleneck(n_windows, frames):
+    """With a bottleneck, expect the divergence to be between 0.5 and 1.0"""
+    hrex_matrix = simulate_bottlenecked_hrex(n_windows, frames)
+    res = get_normalized_kl_divergence(hrex_matrix)
+    assert res <= 1.0 and res >= 0.5
```
maxentile, line 148: Ahh, nice -- is the threshold `res >= 0.5` since there's at least one constrained bit (left vs. right)?

(Guess: If so, then the assertion might not hold exactly for small odd n_windows, but sounds reliable when `n_windows % 2 == 0` or `n_windows` is large.)

(Don't understand the <= 1.0 assertion in this test and >= 1.0 assertion in previous test. If some of the absolute threshold values aren't super solid, might want to make some of the test assertions relative. ("Averaged over many trials with the same (n_windows, n_frames), average diagnostic values for perfect < diffusive < bottlenecked < frozen.", for a few settings of (n_windows, n_frames).)

maxentile (approved):
Looks good! Minor comments, mostly in the randomized tests

mcwitt (approved):

timemachine/md/hrex.py
mcwitt: > The `2 * np.log(n_states)` isn't immediately obvious to me, so I added `axis=0` to be explicit in [05adc51](https://github.com/proteneer/timemachine/commit/05adc5102481746ae26febacf84a4c79664471a5)

I think this is equally clear. (The extra factor of 2 comes from `N` being `n_states ** 2` in the latter case, since we're viewing the distribution as  2-d)

tests/hrex/test_hrex.py
badisa: Fixed up in https://github.com/proteneer/timemachine/pull/1314/commits/68079aab2b6b8fdcef0f050a75313708516ba6a3

tests/hrex/test_hrex.py
badisa: Fixed up in https://github.com/proteneer/timemachine/pull/1314/commits/68079aab2b6b8fdcef0f050a75313708516ba6a3

tests/hrex/test_hrex.py
badisa: Fixed up in https://github.com/proteneer/timemachine/pull/1314/commits/68079aab2b6b8fdcef0f050a75313708516ba6a3

tests/hrex/test_hrex.py
badisa: >  is the threshold res >= 0.5 since there's at least one constrained bit (left vs. right)?

Yes and as you mentioned for small windows (<=3) this fails, added a comment.

Added a relative test in https://github.com/proteneer/timemachine/pull/1314/commits/52a0a8485cc13d7932ec39ee6b54c34cfe9a5b5e

PR #1315, opened 2024-05-29
Adds verify_chiral_consistency_of_core to verify cores quickly

badisa:
* Can be slow to construct a SingleTopology object just to determine if the endstates are correct given a certain core, this PR separates out the construction of the end state potentials for the potentials necessary to check the chiral endstates (bonded, chiral atoms and not used currently but also chiral bonds).
* One of the reason this is slow in practice is AM1BCCELF10 (AM1 + Omega), this avoids the need to generate charges
* About 5x faster to call `verify_chiral_consistency_of_core` with `smirnoff_1_1_0_sc.py` as shown in https://gist.github.com/badisa/466f73d6045f16454c16239c2bb11096


badisa:
This still only makes it about as fast as atom mapping, so may not be fast enough. Potentially removing chiral bonds could make it faster, but not sure that would make much of a difference.

badisa:
25% of the cost of this is the `match_smirks` method: https://github.com/proteneer/timemachine/blob/master/timemachine/ff/handlers/utils.py#L35

proteneer (approved):
very nice! and pretty straightforward

timemachine/fe/single_topology.py
```diff
@@ -309,11 +309,15 @@ def canonicalize_chiral_atom_idxs(idxs):
     return i, jj, kk, ll
 
 
-def setup_end_state(ff, mol_a, mol_b, core, a_to_c, b_to_c):
+def setup_end_state_chiral_potentials(
```
proteneer, line 312: This function is poorly named, since it also returns the bond potential as well?

badisa: Will think of something better

badisa: Renamed to `setup_end_state_harmonic_bond_and_chiral_potentials` for lack of a better name in https://github.com/proteneer/timemachine/pull/1315/commits/b2d6b0d0354745ca4f66da28c760b287d2487c68

timemachine/fe/single_topology.py
```diff
@@ -356,49 +357,25 @@ def setup_end_state(ff, mol_a, mol_b, core, a_to_c, b_to_c):
...
         all_dummy_chiral_atom_idxs.extend(all_idxs[3])
         # append params
         all_dummy_bond_params.extend(all_params[0])
-        all_dummy_angle_params.extend(all_params[1])
-        all_dummy_improper_params.extend(all_params[2])
         all_dummy_chiral_atom_params.extend(all_params[3])
 
     # generate parameters for mol_a
     mol_a_top = topology.BaseTopology(mol_a, ff)
     mol_a_bond_params, mol_a_hb = mol_a_top.parameterize_harmonic_bond(ff.hb_handle.params)
```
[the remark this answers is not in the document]
maxentile: All that's needed from the constructed harmonic bond potentials are the bond idxs, I think (not the assigned Bond types).

A possible performance improvement, if needed, may be to get these idxs from `mol.GetBonds()` (wrapped for convenience in [`get_romol_bonds`](https://github.com/proteneer/timemachine/blob/1791077310d96a529653dc59941bab3c29939c59/timemachine/fe/utils.py#L385-L392)), rather than via [~90 specific bond smirks queries](https://github.com/proteneer/timemachine/blob/1791077310d96a529653dc59941bab3c29939c59/timemachine/ff/smirnoff_2.0.0.xml#L11-L98)).

Alternately, could make a version of the [`_sc.py` force field file](https://github.com/proteneer/timemachine/blob/1791077310d96a529653dc59941bab3c29939c59/timemachine/ff/params/smirnoff_1_1_0_sc.py#L41-L128) mentioned in [PR description](https://github.com/proteneer/timemachine/pull/1315#issue-2324253348) that only has a **single generic bond type** like `"[*:1]~[*:2]"`, so that the chiral consistency check still goes through the ff-typing code path `parameterize_harmonic_bond`, but just does fewer (unnecessary) smirks queries.

(Second option seems simpler and less error-prone.)

maxentile (approved):
Useful addition!

Possible performance improvement noted, either for this PR or a follow-up.

PR #1317, opened 2024-05-31
Speed-up connected_core MCS by 2-3x

proteneer:
This PR speeds-up the connected_core MCS by improving the data-structures and removes the use of `networkx` to perform BFS search (with a cheap early termination check). The defaults to the test_all_pairs test are also updated to reflect changes that are typical of production.

### DIFF
--- a/tests/test_atom_mapping.py
+++ b/tests/test_atom_mapping.py
@@ -379,13 +379,14 @@ def test_all_pairs(filepath):
     for idx, mol_a in enumerate(mols):
         for mol_b in mols[idx + 1 :]:
             # print("Processing", get_mol_name(mol_a), "->", get_mol_name(mol_b))
-            all_cores = atom_mapping.get_cores(
+            start_time = time.time()
+            all_cores, diagnostics = atom_mapping.get_cores_and_diagnostics(
                 mol_a,
                 mol_b,
-                ring_cutoff=0.1,
+                ring_cutoff=0.12,
                 chain_cutoff=0.2,
                 max_visits=1e7,  # 10 million max nodes to visit
-                connected_core=False,
+                connected_core=True,
                 max_cores=1000,
                 enforce_core_core=True,
                 ring_matches_ring_only=False,
@@ -393,6 +394,7 @@ def test_all_pairs(filepath):
                 disallow_planar_torsion_flips=False,
                 min_threshold=0,
             )
+            end_time = time.time()
 
             # # useful for visualization
             # for core_idx, core in enumerate(all_cores[:1]):
@@ -408,7 +410,7 @@ def test_all_pairs(filepath):
                 assert len(core) > mol_a.GetNumAtoms() // 2
 
             print(
-                f"{mol_a.GetProp('_Name')} -> {mol_b.GetProp('_Name')} has {len(all_cores)} cores of size {len(all_cores[0])}"
+                f"{mol_a.GetProp('_Name')} -> {mol_b.GetProp('_Name')} has {len(all_cores)} cores of size {len(all_cores[0])} | total nodes visited: {diagnostics.total_nodes_visited} | wall clock time: {end_time - start_time:.3f}"
             )
 
 
@@ -910,8 +912,9 @@ def test_max_cores_warning():
         min_threshold=0,
         max_visits=1e7,
     )
-    with pytest.warns(MaxVisitsWarning, match="Reached max number of visits/cores: 2 cores"):
-        atom_mapping.get_cores(mol_a, mol_b, **core_kwargs, max_cores=1)
+    with pytest.warns(MaxVisitsWarning, match="Reached max number of visits/cores: 1 cores"):
+        all_cores = atom_mapping.get_cores(mol_a, mol_b, **core_kwargs, max_cores=1)
+        assert len(all_cores) == 1
 
 
 def test_min_threshold():
--- a/timemachine/fe/mcgregor.py
+++ b/timemachine/fe/mcgregor.py
@@ -15,7 +15,14 @@ def _arcs_left(marcs):
     return min(num_row_edges, num_col_edges)
 
 
-UNMAPPED = -1
+# used in main recursion() loop
+UNMAPPED = -1  # (UNVISITED) OR (VISITED AND DEMAPPED)
+
+# used in inner loop when determining whether a mapping will necessarily result
+# in a disconnected subgraph.
+NODE_STATE_VISITED_DEMAPPED = 0  # VISITED AND DEMAPPED
+NODE_STATE_VISITED_MAPPED = 1  # VISITED AND MAPPED
+NODE_STATE_UNVISITED = 2  # UNVISITED
 
 
 def _initialize_marcs_given_predicate(g1, g2, predicate):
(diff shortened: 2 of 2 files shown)


timemachine/fe/mcgregor.py
```diff
@@ -292,31 +316,28 @@ def atom_map_pop(map_1_to_2, map_2_to_1, idx, jdx):
...
-        # see if all the mapped nodes belong to the same connected component
-        for cc in sg_ccs:
-            if set(mapped_nodes).issubset(cc):
-                return False
-        return True
     else:
         return False
 
 
+NODE_STATE_DEMAPPED = 0
```
badisa, line 336: nit: Put the constants at the top of the files

timemachine/fe/mcgregor.py
```diff
@@ -128,6 +128,30 @@ def __init__(self, n_vertices, edges):
             for edge_idx in edges:
                 self.ve_matrix[vertex_idx][edge_idx] = True
 
+    def _verify_partially_connected(self, source, node_states, n_mapped_nodes):
+        """A fast BFS node generator, adapted from _plain_bfs in networkx."""
+        unseen = [1] * self.n_vertices
```
badisa, line 133: nit: Here 1 means unseen and 0 means seen? Its a bit confusing with the constants that have 0, 1, 2 values, but have different meanings.

timemachine/fe/mcgregor.py
```diff
@@ -128,6 +128,30 @@ def __init__(self, n_vertices, edges):
             for edge_idx in edges:
                 self.ve_matrix[vertex_idx][edge_idx] = True
 
+    def _verify_partially_connected(self, source, node_states, n_mapped_nodes):
+        """A fast BFS node generator, adapted from _plain_bfs in networkx."""
```
badisa, line 132: nit: Might be good to add a permalink to the networkx code for future reference

timemachine/fe/mcgregor.py
```diff
@@ -128,6 +128,30 @@ def __init__(self, n_vertices, edges):
...
+        unseen = [1] * self.n_vertices
+        unseen[source] = 0
+        nextlevel = [source]
+        mapped_count = 1
+
+        if mapped_count == n_mapped_nodes:
+            return True
+
+        while nextlevel:
```
badisa, line 141: nit: Lots of truthy checks here, would rather be explicit but don't think there is much risk here

timemachine/fe/mcgregor.py
badisa, line 132: Would be good to explain in a bit more detail of what is being done by this function.

If I understand correctly, given a source verify that all of the other mapped nodes can be found by BFS.

timemachine/fe/mcgregor.py
```diff
@@ -347,37 +368,35 @@ def recursion(
...
     if connected_core:
         # process g1 using atom_map_1_to_2_information
-        g1_mapped_nodes = set()
-        g1_unmapped_nodes = set()
-        g1_unvisited_nodes = set()
+        g1_node_states = [NODE_STATE_DEMAPPED] * g1.n_vertices
         for a1, a2 in enumerate(atom_map_1_to_2):
             if a1 < layer:
                 # visited nodes
                 if a2 == UNMAPPED:
```
badisa, line 375: Would be good to be consistent about the wording, UNMAPPED vs DEMAPPED

timemachine/fe/mcgregor.py
badisa: DEMAPPED vs UNVISITED seems pretty subtle in their difference? It seems to be distinguished  where demapped means you can ignore the node where unvisited allows traversal?

Does this mean that if you have a unvisited, but not mapped node, that connects two sets of mapped atoms that it is considered partially connected?

timemachine/fe/mcgregor.py
```diff
@@ -128,6 +128,30 @@ def __init__(self, n_vertices, edges):
...
+
+        if mapped_count == n_mapped_nodes:
+            return True
+
+        while nextlevel:
+            thislevel = nextlevel
+            nextlevel = []
+            for v in thislevel:
+                for w in self.get_neighbors(v):
+                    if node_states[w] and unseen[w]:
```
mcwitt, line 146: nit: would prefer
```suggestion
                    if node_states[w] != NODE_STATE_DEMAPPED and unseen[w]:
```
to avoid coupling the logic with the representation of the enum

timemachine/fe/mcgregor.py
mcwitt: nit: this docstring seems like an incomplete description. I'm not sure I understand what "partially" means here?

timemachine/fe/mcgregor.py
mcwitt: nit: any reason to not use use `bool` here?

mcwitt (approved):
Nice perf improvement / simplification to avoid networkx overhead for BFS

timemachine/fe/mcgregor.py
proteneer: willfix

proteneer: If a node has been visited: it can be either `MAPPED` or explicitly `DEMAPPED`. Otherwise, it is `UNVISITED`.

> Does this mean that if you have a unvisited, but not mapped node, that connects two sets of mapped atoms that it is considered partially connected?

Yes.

timemachine/fe/mcgregor.py
proteneer: will do True/False

timemachine/fe/mcgregor.py
badisa: So the intention is to say that the core that is being evaluated *could* produce a connected core and not that it is currently a connected core?

timemachine/fe/mcgregor.py
proteneer: This is an extremely hot-loop, your suggestion was actually my old implementation (which requires an additional comparison operator) - similar to why we do `unseen[w]` vs just `not seen[w]` (as in the `_plain_bfs()` implementation)

timemachine/fe/mcgregor.py
proteneer: wow - using a bool here yielded a 50% speed-up

timemachine/fe/mcgregor.py
proteneer: correct.

timemachine/fe/mcgregor.py
proteneer: They mean different things.

UNMAPPED means either {UNVISITED or VISITED UNMAPPED}
DEMAPPED means *only* {VISITED UNMAPPED}

timemachine/fe/mcgregor.py
mcwitt: Hmm, it seems pretty surprising that the existing version should be significantly faster... In my benchmarks, the more explicit version is slightly faster

timemachine/fe/mcgregor.py
proteneer: willfix

timemachine/fe/mcgregor.py
proteneer: you're right - I just tested this again and it seems to be the same. I'll make the comparison explicit.

timemachine/fe/mcgregor.py
proteneer: I've renamed the constants to:

```
NODE_STATE_VISITED_DEMAPPED = 0 # VISITED AND DEMAPPED
NODE_STATE_VISITED_MAPPED = 1 # VISITED AND MAPPED
NODE_STATE_UNVISITED = 2 # UNVISITED
```

timemachine/fe/mcgregor.py
badisa: Would be good to add a comment, as it wasn't clear that it was distinct

tests/test_atom_mapping.py
```diff
@@ -408,8 +410,9 @@ def test_all_pairs(filepath):
                 assert len(core) > mol_a.GetNumAtoms() // 2
 
             print(
-                f"{mol_a.GetProp('_Name')} -> {mol_b.GetProp('_Name')} has {len(all_cores)} cores of size {len(all_cores[0])}"
+                f"{mol_a.GetProp('_Name')} -> {mol_b.GetProp('_Name')} has {len(all_cores)} cores of size {len(all_cores[0])} | total nodes visited: {diagnostics.total_nodes_visited} | wall clock time: {end_time - start_time:.3f}"
             )
+        return
```
badisa, line 415: Should this be removed?

badisa (approved):
:rocket:

tests/test_atom_mapping.py
proteneer: willfix

PR #1319, opened 2024-05-31
Fix nightly test

mcwitt:
Fixes a nightly test failure introduced by #1309.

* Add `hrex_params` to `params` in call to `run_sims_hrex`. As of #1309, `run_sims_hrex` asserts that `hrex_params` is not `None` because it uses the `max_delta_states` parameter.

### DIFF
--- a/tests/test_benchmark_free_energy.py
+++ b/tests/test_benchmark_free_energy.py
@@ -20,7 +20,12 @@ from timemachine.fe.free_energy import (
     run_sims_hrex,
     run_sims_sequential,
 )
-from timemachine.fe.rbfe import setup_initial_states, setup_optimized_host, setup_optimized_initial_state
+from timemachine.fe.rbfe import (
+    DEFAULT_HREX_PARAMS,
+    setup_initial_states,
+    setup_optimized_host,
+    setup_optimized_initial_state,
+)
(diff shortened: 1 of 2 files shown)


badisa (approved):
Did not realize you could call `run_sims_hrex` without providing `md_params.hrex_params`.

Once nice to have would be to move the assert to the top of `run_sim_hrex` so that we don't run any MD before hitting that assertion you added in the previous PR (https://github.com/proteneer/timemachine/blob/master/timemachine/fe/free_energy.py#L1150)

mcwitt:
> Did not realize you could call `run_sims_hrex` without providing `md_params.hrex_params`.
> 
> Once nice to have would be to move the assert to the top of `run_sim_hrex` so that we don't run any MD before hitting that assertion you added in the previous PR (https://github.com/proteneer/timemachine/blob/master/timemachine/fe/free_energy.py#L1150)

Yes, the situation is a bit convoluted as it is, with some of the parameters in `HREXParams` being used at a higher level, or extracted and passed as separate arguments. Would like to clean this up in a future PR.

Good call re moving assertion; done in https://github.com/proteneer/timemachine/pull/1319/commits/939b99ceab3c1b6210a7c531e6d1eaa4e859c624

PR #1320, opened 2024-06-04
Add convenience methods to extract ligand trajectories by replica

mcwitt:
* Adds a new subclass `HREXSimulationResult(SimulationResult)`, moves HREX-specific fields from generic `SimulationResult`
* Adds convenience methods:
  - `extract_trajectories_by_replica(atom_idxs)`
  - `extract_ligand_trajectories_by_replica()`

  to `HREXSimulationResult`


mcwitt:
This should go in after #1321

mcwitt:
(also adding @jkausrelay as reviewer, since we might want to save ligand trajectories downstream)

jkausrelay:
Does the return type of https://github.com/proteneer/timemachine/pull/1320/files#diff-49ef8469d9fb5c9d39e05a0cef32710703096c53f657e55eff49d5c3ddd47b13R459 need to be updated (estimate_relative_free_energy_bisection_or_hrex)?

mcwitt:
> Does the return type of https://github.com/proteneer/timemachine/pull/1320/files#diff-49ef8469d9fb5c9d39e05a0cef32710703096c53f657e55eff49d5c3ddd47b13R459 need to be updated (estimate_relative_free_energy_bisection_or_hrex)?

I don't think so? `HREXSimulationResult` is a subclass of `SimulationResult`, so it is valid for the return type to be the former. I don't think it's straightforward to make it more precise, since `estimate_relative_free_energy_bisection_or_hrex` returns a plain `SimulationResult` if `md_params.hrex_params is None`.

tests/test_hrex_rbfe.py
```diff
@@ -129,15 +129,43 @@ def sample_and_record_rss(*args, **kwargs):
...
+    # Check that we can extract replica trajectories
+    n_atoms = result.final_result.initial_states[0].x0.shape[0]
+    rng = np.random.default_rng(2024)
+    n_atoms_subset = rng.choice(n_atoms)
+    atom_idxs = rng.choice(n_atoms, n_atoms_subset, replace=False)
+    trajs_by_replica = result.extract_trajectories_by_replica(atom_idxs)
+    assert trajs_by_replica.shape == (n_windows, md_params.n_frames, n_atoms_subset, 3)
+
+    # Check that the frame-to-frame rmsd is lower for replica trajectories versus state trajectories
+    def pairwise_rmsd(traj):
```
maxentile, line 141: [low-stakes naming adjustment] : `[rmsd(x[t], x[t+1]) for t in range(len(traj) - 1)]` seems more like `time_lagged_rmsd` or similar than `pairwise_rmsd`.

(low-stakes: in a test, and assertion intents are documented clearly in the next few lines)

(also, 👍 , this gives a clear and clever way to assert temporal coherence)

tests/test_hrex_rbfe.py
```diff
@@ -129,15 +129,43 @@ def sample_and_record_rss(*args, **kwargs):
...
+    # (states, frames)
+    trajs_by_state = np.array(
+        [[np.array(frame)[atom_idxs] for frame in state_traj.frames] for state_traj in result.trajectories]
+    )
+
+    replica_traj_rmsds = jax.vmap(pairwise_rmsd)(trajs_by_replica)
+    state_traj_rmsds = jax.vmap(pairwise_rmsd)(trajs_by_state)
+
+    # should have rmsd(replica trajectory) < rmsd(state trajectory) for all pairs (replica, state)
+    assert np.max(replica_traj_rmsds) < np.min(state_traj_rmsds)
```
maxentile, line 154: [unsure] is there an even stronger (not-flaky) _absolute_ assertion we can make here?

(`np.max(replica_traj_rmsds) < threshold` for some informative + non-flaky choice of `threshold`)

timemachine/fe/free_energy.py
```diff
@@ -266,6 +264,45 @@ def boxes(self) -> List[NDArray]:
...
+        state_idx_by_iter_by_replica = np.argsort(replica_idx_by_iter_by_state, axis=0)
+
+        # (replicas, frames, atoms, 3)
+        trajs_by_replica = np.take_along_axis(trajs_by_state, state_idx_by_iter_by_replica[:, :, None, None], axis=0)
+
+        return trajs_by_replica
+
+    def extract_ligand_trajectories_by_replica(self):
+        """Returns an array of shape (n_replicas, n_frames, n_ligand_atoms, 3) of ligand trajectories for each replica"""
+        ligand_idxs = self.final_result.initial_states[0].ligand_idxs
```
maxentile, line 302: [paranoia] currently (and for the foreseeable future), `s.ligand_idxs` should be identical `for all s in initial_states` -- but should we assert this just in case?

maxentile (approved):

tests/test_hrex_rbfe.py
mcwitt: Also unsure. I think the current assertion should catch most issues in the implementation of `extract_ligand_trajectories_by_replica` (mixing up replicas and states, off-by-one errors of the kind fixed in #1321). It seems like an absolute assertion would depend more on the details of MD, and I'm not sure this would be the right place to put it.

timemachine/fe/free_energy.py
mcwitt: Good call. Added assertion in https://github.com/proteneer/timemachine/pull/1320/commits/892cca96bcc41fbccdeb20ab7851e86f18ae8bff

mcwitt:
> > Does the return type of https://github.com/proteneer/timemachine/pull/1320/files#diff-49ef8469d9fb5c9d39e05a0cef32710703096c53f657e55eff49d5c3ddd47b13R459 need to be updated (estimate_relative_free_energy_bisection_or_hrex)?
> 
> I don't think so? `HREXSimulationResult` is a subclass of `SimulationResult`, so it is valid for the return type to be the former. I don't think it's straightforward to make the annotation more precise, since `estimate_relative_free_energy_bisection_or_hrex` could return a plain `SimulationResult` if `md_params.hrex_params is None`.

@jkausrelay we could also annotate it as returning `SimulationResult | HREXSimulationResult` - maybe that's what you had in mind? I could see the argument for that being more readable (though I think it's equivalent from a type-checking perspective)

tests/test_hrex_rbfe.py
maxentile: Agreed -- on further thought, the relative assertion implemented seems best.

tests/test_hrex_rbfe.py
```diff
@@ -129,15 +129,43 @@ def sample_and_record_rss(*args, **kwargs):
     # Initial permutation should be the identity
     np.testing.assert_array_equal(result.hrex_diagnostics.replica_idx_by_state_by_iter[0], np.arange(n_windows))
 
+    # Check that we can extract replica trajectories
+    n_atoms = result.final_result.initial_states[0].x0.shape[0]
+    rng = np.random.default_rng(2024)
+    n_atoms_subset = rng.choice(n_atoms)
```
badisa, line 135: It looks like `rng.choice` can return 0, which I believe would be invalid here

mcwitt: Fixed in https://github.com/proteneer/timemachine/pull/1320/commits/696d57efe2376bbcefa5ba346c3af451e1432828

badisa (approved):

tests/test_relative_free_energy.py
```diff
@@ -77,6 +78,16 @@ def check_sim_result(sim_res: SimulationResult):
         assert len(sim_res.boxes[-1]) == md_params.n_frames
         assert sim_res.md_params == md_params
 
+        if isinstance(sim_res, HREXSimulationResult):
+            assert md_params.hrex_params
+            n_hrex_iters = md_params.n_frames // md_params.hrex_params.n_frames_per_iter
+
+            assert len(sim_res.hrex_diagnostics.fraction_accepted_by_pair_by_iter) == n_hrex_iters
+            assert all(len(xs) == n_windows - 1 for xs in sim_res.hrex_diagnostics.fraction_accepted_by_pair_by_iter)
```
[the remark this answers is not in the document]
mcwitt: Renamed in https://github.com/proteneer/timemachine/pull/1320/commits/e62581c90e65d9e0e476cb9d28e2e5842958f672

PR #1322, opened 2024-06-05
Clean up dhfr nonbonded tests

badisa:
* Seen by @mcwitt that these would occasionally OOM in the CI. Separated into `pytest.mark.parameterize` test both to better align with our common way of writing tests as well as make it easier to tell which set of parameters triggered the OOMs.
* This *may* fix the issue if it was accumulation of memory and slow garbage collection, but doubtful. May need to reduce the number of atoms passed in the reference platform.


tests/nonbonded/test_nonbonded.py
```diff
@@ -0,0 +1,276 @@
...
+    nonbonded_bp = next(bp for bp in host_fns if isinstance(bp.potential, potentials.Nonbonded))
+    host_conf = host_coords
+    nonbonded_bp.potential.cutoff = 1.2
+    return nonbonded_bp, host_conf, box
+
+
+@pytest.fixture(scope="function")
+def dhfr_system(_dhfr_system):
+    nonbonded_bp, host_conf, box = _dhfr_system
+    return nonbonded_bp, np.array(host_conf), np.array(box)
```
mcwitt, line 34: Should `nonbonded_bp` be copied here too? (looking at this, I'm now confused about the code in `tests/nonbonded/conftest.py`, since it seems like the function-level fixtures there aren't actually making copies of the module-level ones)

tests/nonbonded/test_nonbonded.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+@pytest.mark.parametrize("precision", [np.float64, np.float32])
+def test_nblist_hilbert(dhfr_system, precision):
+    """
+    This test makes sure that hilbert ordering has no impact on numerical results. The
+    computed forces, energies, etc. should be bitwise identical.
+    """
+
+    np.random.seed(2021)
```
mcwitt, line 44: Note that since this is now in `tests/nonbonded`, the autouse fixture `set_random_seed` applies and initially calls `np.random.seed(2022)` (I'm not sure the autouse fixture was a good idea...)

tests/nonbonded/test_nonbonded.py
mcwitt: nit: could this be combined with / reuse `_example_system` in `conftest.py`?

mcwitt (approved):
Nice! Even if this doesn't fix the OOM issue, migrating this to match our current testing conventions will be useful.

tests/nonbonded/test_nonbonded.py
badisa: Combined in https://github.com/proteneer/timemachine/pull/1322/commits/80f1e279700228948926b42735adc2a7a12fd9f8

tests/nonbonded/test_nonbonded.py
badisa: For now I just left it as is, should still be safe this way if I am not misunderstanding the code.

badisa:
@mcwitt @maxentile I noticed that after https://github.com/proteneer/timemachine/pull/1290 we had changed `test_nonbonded.py` to test `cutoff=1.2` but that `setup_dhfr` uses `cutoff=1.0` which meant that for any tests that don't override the cutoff could use the wrong cutoff.

Changed to set the cutoff to 1.2 in https://github.com/proteneer/timemachine/pull/1322/commits/f46d0e8d51582daf27bf8ddffcb1d83248ad6d11

maxentile (approved):

tests/nonbonded/conftest.py
```diff
@@ -18,19 +20,20 @@ def _example_system():
 def example_nonbonded_potential(_example_system):
     host_fns, _, _, _ = _example_system
     nonbonded_bp = next(bp for bp in host_fns if isinstance(bp.potential, potentials.Nonbonded))
-    return nonbonded_bp
+    nonbonded_bp.cutoff = 1.2  # Need to set the cutoff to 1.2 to make the switch func happy
```
badisa, line 23: Actually looks like all of the tests that rely on this fixture are overriding the cutoff to 1.1, do we want to change this?

PR #1323, opened 2024-06-05
Speed up energy decomposition

badisa:
Noticed that it could take awhile for simulations to 'finish' after HREX had finished and upon investigation it looks like we are spending a lot of time reading frames from disk. And due the way the loop was constructed for energy composition we were reading the frames into memory repeatedly.

This only makes a noticeable difference for HREX ( <=~1% for bisection/sequential), which is because there are many more disk reads due to only writing a single frame to disk in the StoredArrays per chunk, unlike sequential/bisection which writes chunks of 100 frames to disk. We probably want to resolve that in a subsequent PR.

# Benchmarks

Cuda Arch 8.6
Nvidia A10

## Vacuum

HREX now gets faster with the more windows :rofl:, shows about a 20% speed up

image

### Speed up relative to master

image

## Solvent

image

### Speed up relative to master
Up to 5%

image

## Complex

image

### Speed up relative to master
image

- [x] Verify that the performance difference between HREX and others is the chunking behavior


timemachine/fe/energy_decomposition.py
```diff
@@ -47,8 +48,12 @@ def get_batch_u_fns(
     batch_u_fns: List[Batch_u_fn] = []
     for p, pot in zip(params, pots):
 
-        def batch_u_fn(xs, boxes, pot_impl, pot_params):
-            coords = np.array([x for x in xs])
+        def batch_u_fn(xs: NDArray | StoredArrays, boxes, pot_impl, pot_params):
```
badisa, line 51: Bit janky, would rather require an NDArray always, but wanted to go for the smallest change to make sure there is a performance improvement to be had.

mcwitt:
Are the ns/day figures strongly dependent on the number of frames we run? (I think this would be the case for changes that only affect the final analysis stage?) It might be clearer to instead report the amount of time spent on the final analysis?

timemachine/fe/energy_decomposition.py
```diff
@@ -93,11 +98,13 @@ def compute_energy_decomposed_u_kln(states: List[EnergyDecomposedState]) -> np.n
         assert len(state.batch_u_fns) == n_components
 
     u_kln_by_component = np.zeros((n_components, K, K, n_frames))
-    for comp in range(n_components):
+    for l in range(K):
+        # Load the frames into memory, then evaluate all of the components
+        # Done to avoid repeatedly reading from disk
+        xs, boxes = np.array([x for x in states[l].frames]), states[l].boxes
```
mcwitt, line 104: Seems like this could get pretty large in memory for solvent and complex legs?

timemachine/fe/energy_decomposition.py
```diff
@@ -47,8 +48,12 @@ def get_batch_u_fns(
     batch_u_fns: List[Batch_u_fn] = []
     for p, pot in zip(params, pots):
 
-        def batch_u_fn(xs, boxes, pot_impl, pot_params):
-            coords = np.array([x for x in xs])
+        def batch_u_fn(xs: NDArray | StoredArrays, boxes, pot_impl, pot_params):
+            # If the coordinates are already an in-memory numpy array, don't create a copy
+            if isinstance(xs, StoredArrays):
+                coords = np.array([x for x in xs])
```
mcwitt, line 54: nit: I think
```suggestion
                coords = np.array([x for x in xs])
```
```suggestion
                coords = np.array(xs)
```

is sufficient

timemachine/fe/energy_decomposition.py
badisa: We are already loading this into memory, by doing this we load it once rather than every call to every potential

badisa: We are still incurring multiple copies of the coordinates to the GPU though. A more performance fix would be to allow running a batch of potentials on coordinates and getting out the per potential energies

timemachine/fe/energy_decomposition.py
badisa: Fixed this case and one other in https://github.com/proteneer/timemachine/pull/1323/commits/c568f9b8859bb095038cfe709d196248e038c5c5

timemachine/fe/energy_decomposition.py
mcwitt: Sounds like another argument for wrapped `DeviceBuffer` :stuck_out_tongue:

mcwitt: > We are already loading this into memory

Ah, right, didn't read carefully enough

badisa:
Updated the description to detail all of the benchmarks with the relative speed ups of this change. Most substantial for vacuum simulations, but does have a couple of percent improvement for HREX in solvent/complex.

https://github.com/proteneer/timemachine/pull/1323#issue-2336861819

timemachine/fe/energy_decomposition.py
```diff
@@ -47,8 +48,12 @@ def get_batch_u_fns(
     batch_u_fns: List[Batch_u_fn] = []
     for p, pot in zip(params, pots):
 
-        def batch_u_fn(xs, boxes, pot_impl, pot_params):
-            coords = np.array([x for x in xs])
+        def batch_u_fn(xs: NDArray | StoredArrays, boxes, pot_impl, pot_params):
+            # If the coordinates are already an in-memory numpy array, don't create a copy
+            if isinstance(xs, StoredArrays):
+                coords = np.array(xs)
```
mcwitt, line 54: nit: could this just use `np.asarray` to avoid the conditional logic here?

timemachine/fe/energy_decomposition.py
mcwitt: nit: to be consistent with `Batch_u_fn` I think `xs` should have type `Frames` here?

mcwitt (approved):
Nice! Simple code change resulting in significant efficiency improvement.

timemachine/fe/energy_decomposition.py
badisa: Done in https://github.com/proteneer/timemachine/pull/1323/commits/49d2b274ef9d06ee3abe45bdaa324905358af258

timemachine/fe/energy_decomposition.py
badisa: Made consistent in https://github.com/proteneer/timemachine/pull/1323/commits/26e81959ace2df39624ce62934c83c10817b4ace

PR #1325, opened 2024-06-07
Hardcode n_frames_per_iter to 1

mcwitt:
## Motivation

* Methods to extract trajectories by replica currently assume `n_frames_per_iter==1` and will break otherwise
* Since HREX now has minimal overhead, there's no clear use case for performing HREX moves less frequently than once per frame
* Hardcoding `n_frames_per_iter = 1` allows removing some subtle logic that exists to handle edge cases

### DIFF
--- a/timemachine/fe/free_energy.py
+++ b/timemachine/fe/free_energy.py
@@ -1184,7 +1184,10 @@ def run_sims_hrex(
         replica_idx_by_state_by_iter.append(hrex.replica_idx_by_state)
 
         hrex, fraction_accepted_by_pair = hrex.attempt_neighbor_swaps_fast(
-            neighbor_pairs, log_q_kl, n_swap_attempts_per_iter, md_params.seed + current_frame
+            neighbor_pairs,
+            log_q_kl,
+            n_swap_attempts_per_iter,
+            md_params.seed + current_frame + 1,  # NOTE: "+ 1" is for bitwise compatibility with previous version
         )
 
         if len(initial_states) == 2:
--- a/timemachine/fe/free_energy.py
+++ b/timemachine/fe/free_energy.py
@@ -59,6 +59,9 @@ class HREXParams:
     n_frames_bisection: int
         Number of frames to sample using MD during the initial bisection phase used to determine lambda spacing
 
+    n_frames_per_iter: int
+        DEPRECATED, must be set to 1. Number of frames to sample using MD per HREX iteration.
+
     max_delta_states: int or None
         If given, number of neighbor states on either side of a given replica's initial state for which to compute
         potentials. This determines the maximum number of states that a replica can move from its initial state during
@@ -66,10 +69,12 @@ class HREXParams:
     """
 
     n_frames_bisection: int = 100
+    n_frames_per_iter: int = 1
     max_delta_states: Optional[int] = 4
 
     def __post_init__(self):
         assert self.n_frames_bisection > 0
+        assert self.n_frames_per_iter == 1, "n_frames_per_iter must be 1"
         assert self.max_delta_states is None or self.max_delta_states > 0
 
 
--- a/timemachine/fe/free_energy.py
+++ b/timemachine/fe/free_energy.py
@@ -59,9 +59,6 @@ class HREXParams:
     n_frames_bisection: int
         Number of frames to sample using MD during the initial bisection phase used to determine lambda spacing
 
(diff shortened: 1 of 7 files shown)


badisa (approved):

timemachine/fe/free_energy.py
```diff
@@ -59,22 +59,17 @@ class HREXParams:
...
-        Number of frames to sample using MD per HREX iteration.
-
     max_delta_states: int or None
         If given, number of neighbor states on either side of a given replica's initial state for which to compute
         potentials. This determines the maximum number of states that a replica can move from its initial state during
         a single HREX iteration. Otherwise, compute potentials for all (replica, state) pairs.
     """
 
     n_frames_bisection: int = 100
-    n_frames_per_iter: int = 1
```
jkausrelay, line 72: Is there a backwards compatibility issue of loading old pkl'd HREXParams with this removed?

timemachine/fe/free_energy.py
```diff
@@ -1204,7 +1195,7 @@ def replica_from_samples(traj: Trajectory) -> CoordsVelBox:
 
         fraction_accepted_by_pair_by_iter.append(fraction_accepted_by_pair)
 
-        if print_diagnostics_interval and iteration % print_diagnostics_interval == 0:
+        if print_diagnostics_interval and (current_frame + 1) % print_diagnostics_interval == 0:
```
jkausrelay, line 1198: why +1 ?

timemachine/fe/free_energy.py
mcwitt: Yes, I think so, but also seemed confusing to leave it... I'm not sure what our expectations should be re backwards compatibility here; happy to defer to others.

timemachine/fe/free_energy.py
mcwitt: `current_frame` starts at 0, while `iteration` started at 1. This retains the same behavior (i.e. not printing the diagnostics after frame 0), but we could change it if preferable

timemachine/fe/free_energy.py
jkausrelay: Maybe leave it and just assert it equals one in the check below?

badisa: We should know which versions of timemachine we have per pickle, which as long as we know that we should be able to debug still

timemachine/fe/free_energy.py
jkausrelay: Should https://github.com/proteneer/timemachine/pull/1325/files/a682e19aa2943034c877764035414dd369b8a942#diff-6f4d3e6b5b0b6a38f1b9b5f8f51dcb56ec102ec03d06a7d53355bb67c4aee676R1187 be (current_frame+1) then? Not so concerned w/the diagnostics, just noticed the inconsistency between this line and the one above.

mcwitt: I don't think it matters in that case, since we just need a different seed for each frame. I was on the fence about this, as adding 1 would maintain bitwise compatibility, but would also leave a weird arbitrary artifact in the code.

jkausrelay: I see, I'd prefer maintaining bitwise compatibility where possible.

badisa: Didn't notice the seed, yeah should definitely maintain bitwise compatiblity

mcwitt: OK, will add the +1, and a note that this can be cleaned up the next time we need to break binary compatibility

timemachine/fe/free_energy.py
mcwitt: > Maybe leave it and just assert it equals one in the check below?

That works for me

timemachine/fe/free_energy.py
mcwitt: Done in https://github.com/proteneer/timemachine/pull/1325/commits/002fc06efb4e926d42feacba180553ba202d06fd

timemachine/fe/free_energy.py
mcwitt: Reverted with assertion in https://github.com/proteneer/timemachine/pull/1325/commits/a49818818b3109b4cbe4c18cce94d08b91dde5db

jkausrelay (approved):

PR #1334, opened 2024-06-21
Add method to get samples from context without storing to disk

badisa:
# Changes

* Minor changes to the StoredArrays object around memory usage
* 5% speed up of Vacuum HREX runs, 1% for solvent

# Benchmarks
Cuda arch 8.6
Nvidia A10

## Vacuum
For 48 windows of vacuum the ns/day went from 5919 -> 6318, a ~7% speed up.

### Master
vacuum_100_benchmarks_master

### Changes
vacuum_100_benchmarks

## Solvent
For 48 windows of solvent the ns/day went from 900 -> 911, a ~1% speed up.

### Master
solvent_100_benchmarks

### Changes
solvent_100_benchmarks_allmem

### DIFF
--- a/timemachine/fe/free_energy.py
+++ b/timemachine/fe/free_energy.py
@@ -547,7 +547,7 @@ def sample_with_context_iter(
     ctxt: Context, md_params: MDParams, temperature: float, ligand_idxs: NDArray, max_buffer_frames: int
 ) -> Iterator[Tuple[NDArray, NDArray, NDArray]]:
     """Sample a context using MDParams returning batches of frames up to `max_buffer_frames`. All results are returned
-    as numpy arrays that are in memory, and it is left to the user to act accordingly..
+    as numpy arrays that are in memory, and it is left to the user to act accordingly.
 
     For getting a Trajectory object that stores the frames to disk, refer to `sample_with_context`.
 
@@ -557,16 +557,17 @@ def sample_with_context_iter(
         The context to use to generate samples
 
     md_params: MDParams
-        The parameters to use to collect frames from the context
+        The parameters that define the sampling of frames from the context
 
     temperature: float
         The temperature, in kelvin, used when running Local MD moves
 
     ligand_idxs: np.ndarray
-        Array representing the indices of atoms that make up the ligand, determining the local MD region
+        Array representing the indices of atoms that make up the ligand, determines the atoms considered as the center
+        of local MD.
 
     max_buffer_frames: int
-        The most number of frames (coords and boxes) that will be kept in memory at one time
+        The most number of frames (coords and boxes) that will be kept in memory at one time.
 
     Returns
     -------
--- a/timemachine/fe/free_energy.py
+++ b/timemachine/fe/free_energy.py
@@ -1225,7 +1225,7 @@ def run_sims_hrex(
             assert md_params_replica.n_frames == 1
             # Get the next set of frames from the iterator, which will be the only value returned
             frame, box, final_velos = next(
-                sample_with_context_iter(context, md_params_replica, temperature, ligand_idxs, max_buffer_frames=1)
+                sample_with_context_iter(context, md_params_replica, temperature, ligand_idxs, batch_size=1)
             )
             assert frame.shape[0] == 1
 
--- a/timemachine/fe/free_energy.py
+++ b/timemachine/fe/free_energy.py
@@ -1224,7 +1224,7 @@ def run_sims_hrex(
             assert md_params_replica.n_frames == 1
             # Get the next set of frames from the iterator, which will be the only value returned
             frame, box, final_velos = next(
-                sample_with_context_iter(context, md_params_replica, temperature, ligand_idxs, 1)
+                sample_with_context_iter(context, md_params_replica, temperature, ligand_idxs, max_buffer_frames=1)
             )
             assert frame.shape[0] == 1
 
--- a/timemachine/fe/free_energy.py
+++ b/timemachine/fe/free_energy.py
@@ -1223,16 +1223,18 @@ def run_sims_hrex(
 
             assert md_params_replica.n_frames == 1
             # Get the next set of frames from the iterator, which will be the only value returned
-            frame, box, velos = next(sample_with_context_iter(context, md_params_replica, temperature, ligand_idxs, 1))
+            frame, box, final_velos = next(
+                sample_with_context_iter(context, md_params_replica, temperature, ligand_idxs, 1)
+            )
             assert frame.shape[0] == 1
 
             final_barostat_volume_scale_factor = barostat.get_volume_scale_factor() if barostat is not None else None
 
-            return frame, box, velos, final_barostat_volume_scale_factor
+            return frame[0], box[0], final_velos, final_barostat_volume_scale_factor
 
         def replica_from_samples(last_sample: Tuple[NDArray, NDArray, NDArray, Optional[float]]) -> CoordsVelBox:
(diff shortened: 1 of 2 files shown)


timemachine/fe/free_energy.py
```diff
@@ -1181,10 +1221,18 @@ def sample_replica(xvb: CoordsVelBox, state_idx: StateIdx) -> Trajectory:
...
+
+            final_barostat_volume_scale_factor = barostat.get_volume_scale_factor() if barostat is not None else None
+
+            return frame, box, velos, final_barostat_volume_scale_factor
 
-        def replica_from_samples(traj: Trajectory) -> CoordsVelBox:
-            return CoordsVelBox(traj.frames[-1], traj.final_velocities, traj.boxes[-1])
+        def replica_from_samples(last_sample: Tuple[NDArray, NDArray, NDArray, Optional[float]]) -> CoordsVelBox:
+            frame, box, velos, _ = last_sample
+            return CoordsVelBox(frame.squeeze(), velos.squeeze(), box.squeeze())
```
mcwitt, line 1235: nit: preference for explicit slicing instead of `squeeze` here. Is the following equivalent?
```suggestion
            return CoordsVelBox(frame[0], velos[0], box[0])
```
(might be missing something)

mcwitt: Could this actually be moved into `sample_replica`? (seems cleaner than having the latter return arrays with size 1 in the first dimension)

timemachine/fe/free_energy.py
```diff
@@ -601,22 +633,31 @@ def run_production_local_steps(n_steps: int) -> Tuple[NDArray, NDArray, NDArray]
             coords.append(x_t)
             boxes.append(box_t)
 
-        assert np.all(coords[-1][-1] == ctxt.get_x_t())
```
mcwitt, line 604: Is there a real performance impact to removing these assertions? (or are they replicated elsewhere that I'm missing?) They seem like a nice sanity check, but probably not necessary if this code is sufficiently tested.

badisa: Don't think so, but would rather not do work we don't have to. I am inclined to say it is sufficiently tested since I changed this behavior in https://github.com/proteneer/timemachine/pull/724 if we broadly feel that it isn't, I can revert these changes in this PR. Then follow up with a PR that removes them and adds more test.

timemachine/fe/free_energy.py
badisa: Moved into `sample_replica` and used explicit slicing in https://github.com/proteneer/timemachine/pull/1334/commits/f359991bb437c04636ae00d82dbad458860be13f

badisa: Moved into `sample_replica` and used explicit slicing in https://github.com/proteneer/timemachine/pull/1334/commits/f359991bb437c04636ae00d82dbad458860be13f

timemachine/fe/free_energy.py
```diff
@@ -1162,7 +1202,7 @@ def get_equilibrated_xvb(xvb: CoordsVelBox, state_idx: StateIdx) -> CoordsVelBox
 
     for current_frame in range(md_params.n_frames):
 
-        def sample_replica(xvb: CoordsVelBox, state_idx: StateIdx) -> Trajectory:
+        def sample_replica(xvb: CoordsVelBox, state_idx: StateIdx) -> Tuple[NDArray, NDArray, NDArray, Optional[float]]:
```
mcwitt, line 1205: nit: wonder if it'd be worth defining `Sample = Tuple[NDArray, NDArray, NDArray, Optional[float]]` for clarity?

timemachine/fe/free_energy.py
```diff
@@ -1181,10 +1221,20 @@ def sample_replica(xvb: CoordsVelBox, state_idx: StateIdx) -> Trajectory:
 
             md_params_replica = replace(md_params, n_frames=1, n_eq_steps=0, seed=state_idx + current_frame)
 
-            return sample_with_context(context, md_params_replica, temperature, ligand_idxs, max_buffer_frames=100)
+            assert md_params_replica.n_frames == 1
```
mcwitt, line 1224: Q: is it actually necessary to have `n_frames == 1` anymore, since we're passing `max_buffer_frames=1`?

timemachine/fe/free_energy.py
```diff
@@ -1181,10 +1221,20 @@ def sample_replica(xvb: CoordsVelBox, state_idx: StateIdx) -> Trajectory:
...
+            # Get the next set of frames from the iterator, which will be the only value returned
+            frame, box, final_velos = next(
+                sample_with_context_iter(context, md_params_replica, temperature, ligand_idxs, 1)
```
mcwitt, line 1227: nit: slight preference, for readability
```suggestion
                sample_with_context_iter(context, md_params_replica, temperature, ligand_idxs, max_buffer_frames=1)
```

timemachine/fe/free_energy.py
badisa: If `md_params.n_frames == 2` and `max_buffer_frames=1` then calling `next()` would be incorrect since it wouldn't be the last frame.

`max_buffer_frames` here doesn't really matter since `n_frames=1`

timemachine/fe/free_energy.py
badisa: Its an easy thing to do, but not clear what the value would be. Mostly because this type has changed to meet our needs over time, returning the final velocities and the state of the barostat, which feels too ill-defined to warrant a type hint devoted to do. 

Also there are already several type hints for 'Sample[s]' type aliases that are slightly different which makes me further hesitant to add another type without trying to unify this:
https://github.com/proteneer/timemachine/blob/master/timemachine/md/hrex.py#L129
https://github.com/proteneer/timemachine/blob/master/timemachine/md/smc.py#L13-L14
https://github.com/proteneer/timemachine/blob/master/timemachine/fe/reweighting.py#L16

mcwitt: Fair enough; only if you think it improves overall clarity. Re multiple instances of type aliases with the same name: I think it's perfectly OK to have local type aliases with the same name as a type alias in another module (kind of seems like a quirk of Python that type aliases have to be top-level value bindings), but we might want to start prefixing aliases intended to by local with an underscore or similar.

timemachine/fe/free_energy.py
```diff
@@ -1202,8 +1253,13 @@ def replica_from_samples(traj: Trajectory) -> CoordsVelBox:
         if len(initial_states) == 2:
             fraction_accepted_by_pair = fraction_accepted_by_pair[1:]  # remove stats for identity move
 
-        for samples, samples_iter in zip(samples_by_state, samples_by_state_iter):
-            samples.extend(samples_iter)
```
mcwitt, line 1206: nit: should we remove the `Trajectory.extend` method, since it's not used anymore?

mcwitt (approved):

timemachine/fe/free_energy.py
badisa: Seems reasonable to keep as it is a handy utility function, if you want to write some quick and dirty code where you extend a trajectory from the results of calling `sample(initial_state, md_params)`. Can pull out if we want though, as it is no longer used

mcwitt: I'm OK either way. In the future we might be able to generalize the `Trajectory` type to handle both in-memory and disk-backed arrays and go back to using `extend`.

maxentile (approved):

PR #1335, opened 2024-06-26
update to 1.20.3-1ubuntu2.1

jkausrelay:
Need to update wget, old version was pulled

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ ARG LIBXEXT_VERSION
 ARG MINICONDA_VERSION=py310_23.1.0-1
 ARG MAKE_VERSION=4.2.1-1.2
 ARG GIT_VERSION=1:2.25.1-*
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -10,7 +10,7 @@ ARG LIBXEXT_VERSION
 ARG MINICONDA_VERSION=py310_23.1.0-1
 ARG MAKE_VERSION=4.2.1-1.2
 ARG GIT_VERSION=1:2.25.1-*
-ARG WGET_VERSION=1.20.3-1ubuntu2
+ARG WGET_VERSION=1.20.3-1ubuntu2.1
```
badisa, line 13: Might be better to do
```suggestion
ARG WGET_VERSION=1.20.3-*
```

These ubuntu specific versions seem to disappear over time

badisa (approved):

PR #1336, opened 2024-06-28
Minor improvements to optimize_coordinates

badisa:
* Minor performance improvements, only look at protein/ligand (1/3 of the atoms in some systems, though don't think this makes much if any difference)
* Clean up some type hints
* Clean up docstrings
* Adds a test for the failure

### DIFF
--- a/tests/test_single_topology_confgen.py
+++ b/tests/test_single_topology_confgen.py
@@ -149,6 +149,7 @@ def test_min_cutoff_failure(pair, seed, n_windows):
     during minimization. Expectation is that all atoms of the molecules will move beyond effectively zero cutoff"""
     src, dst = pair
     box_width = 4.0
+    # The cutoff is so small that any ligand pair should trigger the exception
     min_cutoff = 1e-8
 
     ligands = "timemachine/datasets/fep_benchmark/hif2a/ligands.sdf"
--- a/tests/test_single_topology_confgen.py
+++ b/tests/test_single_topology_confgen.py
@@ -139,3 +139,41 @@ def test_confgen_spot_edges():
         mol_a = get_mol_by_name(mols, src)
         mol_b = get_mol_by_name(mols, dst)
         run_edge(mol_a, mol_b, protein_path, n_windows)
+
+
+@pytest.mark.parametrize("pair", [("35", "84")])
+@pytest.mark.parametrize("n_windows", [2])
+@pytest.mark.parametrize("seed", [2023])
+def test_min_cutoff_failure(pair, seed, n_windows):
+    """Verify that when minimizing within solvent only the ligand is considered for movement of atoms
+    during minimization. Expectation is that all atoms of the molecules will move beyond effectively zero cutoff"""
+    src, dst = pair
+    box_width = 4.0
+    min_cutoff = 1e-8
+
+    ligands = "timemachine/datasets/fep_benchmark/hif2a/ligands.sdf"
+    mols = read_sdf(ligands)
+    mol_a = get_mol_by_name(mols, src)
+    mol_b = get_mol_by_name(mols, dst)
+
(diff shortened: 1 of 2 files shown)


jkausrelay (approved):

timemachine/fe/rbfe.py
```diff
@@ -250,10 +262,11 @@ def setup_optimized_initial_state(
     temperature: float,
     seed: int,
 ) -> InitialState:
-    # Use pre-optimized initial state with the closest value of lambda as a starting point for optimization.
+    """Use pre-optimized initial state with the closest value of lambda as a starting point for optimization.
 
-    # NOTE: The current approach for generating optimized conformations in `optimize_coordinates` creates a
-    # discontinuity at lambda=0.5. Ensure that we pick a pre-optimized state on the same side of 0.5 as `lamb`:
+    NOTE: The current approach for generating optimized conformations in `optimize_coordinates` creates a
```
mcwitt, line 267: nit: this isn't really intended to be part of the docstring. It could be adapted to go in the docstring, but e.g. should probably get rid of the trailing `:`.

badisa: Good catch, moved back to being an inline comment in https://github.com/proteneer/timemachine/pull/1336/commits/197fcaa79da1e3a2112fbc5aa8f5d8b69f2d73ef and removed the trailing `:`

timemachine/fe/rbfe.py
```diff
@@ -250,10 +262,10 @@ def setup_optimized_initial_state(
     temperature: float,
     seed: int,
 ) -> InitialState:
-    # Use pre-optimized initial state with the closest value of lambda as a starting point for optimization.
+    """Use the pre-optimized initial state with the closest value of lambda as a starting point for optimization."""
```
mcwitt, line 265: nit: 
```suggestion
    """Return a minimized initial state at the specified lambda, using the pre-optimized initial state from the provided list with the closest value of lambda as a starting point for optimization."""
```

tests/test_single_topology_confgen.py
```diff
@@ -139,3 +139,41 @@ def test_confgen_spot_edges():
...
+@pytest.mark.parametrize("n_windows", [2])
+@pytest.mark.parametrize("seed", [2023])
+def test_min_cutoff_failure(pair, seed, n_windows):
+    """Verify that when minimizing within solvent only the ligand is considered for movement of atoms
+    during minimization. Expectation is that all atoms of the molecules will move beyond effectively zero cutoff"""
+    src, dst = pair
+    box_width = 4.0
+    min_cutoff = 1e-8
+
+    ligands = "timemachine/datasets/fep_benchmark/hif2a/ligands.sdf"
```
mcwitt, line 154: Assuming that this is a specific example where non-ligand atoms move more than the cutoff? (might call this out to clarify)

mcwitt: Sorry, never mind this. I see now that you're just setting the cutoff really small so that every atom is probably moving more than the cutoff, and asserting that only the ligand atoms are triggering the failure.

mcwitt (approved):

timemachine/fe/rbfe.py
badisa: Since we have tweaked this a few times, expanded the docstring to be more fleshed out in https://github.com/proteneer/timemachine/pull/1336/commits/ecf3455890d14473354a819f7441e590a0268ddd

tests/test_single_topology_confgen.py
badisa: Added comment in https://github.com/proteneer/timemachine/pull/1336/commits/5fc2486e48a1b8ab644e2cbbbc0d3202c73fbb50

PR #1338, opened 2024-07-02
Minimize InitialStates only considering fully interacting atoms for the min_cutoff threshold

badisa:
* Adds an optional field to `InitialState` to store the `AtomMapMixin` (or really the SingleTopology). Still allows unpickling older files since the field is optional.
* In the case of intermediate windows the dummy atoms can move all over the place and could see failures where min_cutoff assertion was triggered for non-interacting dummy atoms. 
* Sets min_cutoff == 0.0 for vacuum cases we had missed in the tests, to be consistent with `run_vacuum`.
* The definition of interacting atoms is entirely determined by the atom mapping

  * Had attempted to do this by looking at the parameters, but seems more error prone and was not immediately obvious how to handle NonbondedPairListPrecomputed parameters.

### DIFF
--- a/timemachine/fe/free_energy.py
+++ b/timemachine/fe/free_energy.py
@@ -24,6 +24,7 @@ from timemachine.fe.plots import (
     plot_overlap_summary_figure,
 )
 from timemachine.fe.protocol_refinement import greedy_bisection_step
+from timemachine.fe.single_topology import AtomMapMixin
 from timemachine.fe.stored_arrays import StoredArrays
 from timemachine.fe.utils import get_mol_masses, get_romol_conf
 from timemachine.ff import ForcefieldParams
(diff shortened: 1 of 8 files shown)


timemachine/fe/free_energy.py
```diff
@@ -154,11 +155,35 @@ class InitialState:
     lamb: float
     ligand_idxs: NDArray
     protein_idxs: NDArray
+    atom_map: Optional[AtomMapMixin] = None
```
jkausrelay, line 158: Add a comment that this is needs to be set for RBFE only (since we also use initial states to run MD)

timemachine/fe/free_energy.py
```diff
@@ -154,11 +155,35 @@ class InitialState:
...
+        * Raises an exception if no atom map object is provided
+
+        Return
+        ------
+        np.ndarray
+            Contains the indices of the ligand that are not considered dummy atoms at the lambda value.
+        """
+        assert self.atom_map is not None, "No atom map provided, unable to provide interacting atom indices"
+        if self.lamb == 0.0:
+            return self.ligand_idxs[self.atom_map.c_flags != 2]
```
jkausrelay, line 181: nit: It'd be nice if these constants were defined somewhere

jkausrelay (approved):
Minor comments, lgtm

timemachine/fe/free_energy.py
badisa: Comment added in https://github.com/proteneer/timemachine/pull/1338/commits/0b0a8921ad7b0d35875243facf083c8914ac78e5

timemachine/fe/free_energy.py
badisa: Added an enum that defines the types in https://github.com/proteneer/timemachine/pull/1338/commits/4891138630208cbace3fddb8b9cbe4070160972d

mcwitt:
> Had attempted to do this by looking at the parameters, but seems more error prone and was not immediately obvious how to handle NonbondedPairListPrecomputed parameters.

I would push back here and say that using the atom mapping seems more error prone. I.e. it seems like the source of truth for whether atoms are interacting should be the potential parameters? Otherwise a bug or some change in the parameterization logic might lead us to think that interacting atoms are noninteracting here.

badisa:
> > Had attempted to do this by looking at the parameters, but seems more error prone and was not immediately obvious how to handle NonbondedPairListPrecomputed parameters.
> 
> I would push back here and say that using the atom mapping seems more error prone. I.e. it seems like the source of truth for whether atoms are interacting should be the potential parameters? Otherwise a bug or some change in the parameterization logic might lead us to think that interacting atoms are noninteracting here.

For the cases of solvent and complex it is straightforward since each atom has a single parameter in the params (how I originally did it in https://github.com/proteneer/timemachine/pull/1338/commits/88a8b1821625092d52893a9d0a85e56f964222aa). In the case of vacuum where there is a NonbondedPairlistPrecomputed there are only delta_w coords which makes it difficult to reconstruct which atoms are actually dummy. Could refactor out the code in https://github.com/proteneer/timemachine/blob/master/timemachine/fe/single_topology.py#L1261-L1278 but at the moment this seems pre-mature since we always have it so that dummy atoms are non-interacting in all states but the dummy atom respective end state.

mcwitt:
Random thought: another approach to detecting noninteracting atoms might be to look at where $\partial U_{\text{nb}} / \partial x$ is zero. I'm not sure if this would be a simpler/more robust approach?

badisa:
> Random thought: another approach to detecting noninteracting atoms might be to look at where ∂Unb/∂x is zero. I'm not sure if this would be a simpler/more robust approach?

Don't know that that is always true. If you had a long chain of dummy atoms, think they could still have ∂Unb/∂x != 0.0 due to interacting with the other dummy atoms that are non-interacting. I would also like to avoid having to run potentials to determine which atoms are interacting or not.

timemachine/fe/free_energy.py
mcwitt: (independently, I've been wondering recently whether it's worth adding `AtomMapMixin` to `SimulationResult`, since I end up reconstructing the mapping in a lot investigations)

badisa: >  I've been wondering recently whether it's worth adding AtomMapMixin to SimulationResult

Part of the reason I added it to the InitialState, is that I end up having to reconstruct the AtomMapMixin a lot, but I tend to work with InitialStates rather than SimulationResults. 

Will play around with constructing the interacting atoms in the `setup_initial_state` function. If we do that do we want to break pickling by adding in a positional argument?

mcwitt:
> > > Had attempted to do this by looking at the parameters, but seems more error prone and was not immediately obvious how to handle NonbondedPairListPrecomputed parameters.
> > 
> > 
> > I would push back here and say that using the atom mapping seems more error prone. I.e. it seems like the source of truth for whether atoms are interacting should be the potential parameters? Otherwise a bug or some change in the parameterization logic might lead us to think that interacting atoms are noninteracting here.
> 
> For the cases of solvent and complex it is straightforward since each atom has a single parameter in the params (how I originally did it in [88a8b18](https://github.com/proteneer/timemachine/commit/88a8b1821625092d52893a9d0a85e56f964222aa)). In the case of vacuum where there is a NonbondedPairlistPrecomputed there are only delta_w coords which makes it difficult to reconstruct which atoms are actually dummy. Could refactor out the code in https://github.com/proteneer/timemachine/blob/master/timemachine/fe/single_topology.py#L1261-L1278 but at the moment this seems pre-mature since we always have it so that dummy atoms are non-interacting in all states but the dummy atom respective end state.

Ah OK, I think I've got it now (also after some offline discussion): I was initially tripped up by the sense in which we're using "noninteracting" here, but I think we mean "having no interaction with the core"? And as @badisa  said above, that's easy to determine for the potentials that are parameterized with a $w$ coordinate per atom, but not obvious how to tell which atom in a pair with 4-d offset $w$ is in the core if you don't have the atom mapping.

timemachine/fe/free_energy.py
badisa: Changed to interacting atoms in https://github.com/proteneer/timemachine/pull/1338/commits/0202c84561001738705484397b716d884c3a95c6

Didn't change `setup_initial_states` in `absolute_hydration.py` since it looks unused

mcwitt (approved):

PR #1339, opened 2024-07-03
Modify Minimization to avoid minimizing while mols are non-interacting

badisa:
* Completely freeze the ligands in place, so that the host coordinates aren't minimized to different poses than what the system expects downstream.
* Delete clashy waters upon system set up, which resolves issues with cagey molecules (where water could be trapped inside)
* Adds pre-equillibration to reduce 'air bubbles' in the simulations (does not modify ligand).
* Centers solvent systems around ligands when building systems to ensure consistent water box construction. IE such that the ligands don't get placed in the margin setup within the box and minimization of solvent boxes are dependent on the starting point of ligands.
* Decoupling is still required due to eps values of 0.0 in protein hydrogens, but now minimize at lambda = 0.1 and 0.0 (can be adjusted), rather than lowering down from 1.0

## Systems tested
Verified that the following public systems and the corresponding ligands minimize in the solvent/complex (including through bisection). 

Sets taken from 
* Hif2a
* PFKFB3 - Added to the tests since it has a clash that blows up
* TYK2
* CDK8 - The original set that required our previous implementation of minimization

## Todo

- [x] Add a test that if a ligand is moved into bulk around a protein that waters are deleted and that the protein is unmodified (cover the bug fixed in https://github.com/proteneer/timemachine/pull/1339/commits/330ffeea6c2dfd5caa8f4f0f2ad86c9f61d94422)
- [ ] Add test of the two particle system that fails to minimize.


timemachine/fe/rbfe.py
```diff
@@ -343,9 +344,12 @@ def _optimize_coords_along_states(initial_states: List[InitialState]) -> List[ND
...
         print(f"Optimizing initial state at λ={initial_state.lamb}")
         free_idxs = get_free_idxs(initial_state)
-        x_opt = optimize_coords_state(
-            initial_state.potentials, x_opt, initial_state.box0, free_idxs, assert_energy_decreased=idx == 0
-        )
+        try:
+            x_opt = optimize_coords_state(
+                initial_state.potentials, x_opt, initial_state.box0, free_idxs, assert_energy_decreased=idx == 0
+            )
+        except (AssertionError, minimizer.MinimizationError) as e:
```
badisa, line 351: This was just a helpful additional message to tell me which lambda window was actually blowing up.

timemachine/md/minimizer.py
```diff
@@ -140,85 +142,178 @@ def minimize_host_4d(
...
     v0 = np.zeros_like(x0)
 
-    potentials, params = parameterize_system(hgt, ff, 1.0)
+    num_host_atoms = minimized_host_coords.shape[0]
+
+    potentials, params = parameterize_system(hgt, ff, 0.0)
+    bond_pot = next(pot for pot in potentials if isinstance(pot, HarmonicBond))
+    group_idxs = get_group_indices(get_bond_list(bond_pot), x0.shape[0])
+    # Only adjust the non-ligand group idxs, we don't want to move the ligand coordinates at all
+    non_ligand_group_idxs = [group for group in group_idxs if np.all(group < num_host_atoms)]
```
badisa, line 222: Run the barostat with the ligands excluded from mols can be adjusted. Going to be less efficient, but does still reduce the box size.

tests/test_barostat.py
```diff
@@ -323,19 +323,19 @@ def test_barostat_varying_pressure():
         coords.shape[0], pressure, temperature, group_indices, barostat_interval, u_impls, seed, True, 0.0
     )
 
-    ctxt = custom_ops.Context(coords, v_0, complex_box, integrator_impl, u_impls, movers=[baro])
-    ctxt.multiple_steps(1000)
+    ctxt = custom_ops.Context(coords, v_0, box, integrator_impl, u_impls, movers=[baro])
+    ctxt.multiple_steps(3000)
```
jkausrelay, line 327: Change needed for convergence?

badisa: Change was convergence, though I ended up changing the test it looks like. Will adjust back to the way it was as we don't need to run the extra MD.

timemachine/md/minimizer.py
```diff
@@ -140,85 +142,181 @@ def minimize_host_4d(
...
     combined_masses = np.concatenate(mass_list)
     combined_coords = np.concatenate(conf_list)
 
     hgt = topology.HostGuestTopology(host_bps, top, host_config.num_water_atoms)
 
-    # this value doesn't matter since we will turn off the noise.
-    seed = 0
-
-    intg = LangevinIntegrator(0.0, 1.5e-3, 1.0, combined_masses, seed).impl()
+    intg = LangevinIntegrator(DEFAULT_TEMP, 1.5e-3, 1.0, combined_masses, seed).impl()
```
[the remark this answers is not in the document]
badisa: Presumably when you say constants, you mean define them at the top of the function? Such as `dt = 1.5e-3`?

timemachine/md/minimizer.py
badisa: Correct

timemachine/md/minimizer.py
```diff
@@ -140,85 +142,181 @@ def minimize_host_4d(
...
+        Pressure to run barostat with
+
+    temperature: float
+        Temperature to run MD with
+
+    barostat_interval: integer
+        How often to run the barostat
+
+    seed: integer
+        Seed for barostat and integrator
```
maxentile, line 167: some of these can be edited to provide more information not already present in parameter name, e.g. 
```suggestion
    pressure: float
        in bar (used by barostat)

    temperature: float
        in kelvin (used by integrator and barostat)

    barostat_interval: integer
        # of MD steps between barostat moves

    seed: integer
        integrator uses `seed`, barostat uses `seed+1`
```

timemachine/md/minimizer.py
```diff
@@ -140,85 +142,181 @@ def minimize_host_4d(
...
+
+    n_steps_per_window: integer
+        Number of steps to run the FIRE minimizer with at each window
+
+    max_lambda: float
+        The largest lambda value to run a window at. If the value is 1.0 the mols will be
+        completely decoupled.
+
+    n_windows: integer
+        The number of windows to linearly interpolate between the max_lambda value and 0.0.
```
maxentile, line 293: clarification: will this correspond to linear interpolation of `w_coord` between `0` and `max_lambda_value * cutoff`, or some pre-optimized interpolation like https://github.com/proteneer/timemachine/blob/ab3def743ad562c2d89e90b8ea54e989a6de5a2a/timemachine/fe/lambda_schedule.py#L22 ?

(linear probably suffices, but it's just good to note which one is chosen here)

timemachine/md/minimizer.py
```diff
@@ -140,85 +142,181 @@ def minimize_host_4d(
...
+
     host_bps, host_masses = openmm_deserializer.deserialize_system(host_config.omm_system, cutoff=1.2)
 
     num_host_atoms = host_config.conf.shape[0]
 
     if len(mols) == 1:
         top = topology.BaseTopology(mols[0], ff)
     elif len(mols) == 2:
-        top = topology.DualTopologyMinimization(mols[0], mols[1], ff)
+        top = topology.DualTopology(mols[0], mols[1], ff)
```
maxentile, line 194: curiosity/oddity: this PR changes the topology in the function `minimize_host_4d` from `DualTopologyMinimization` to `DualTopology`, but changes the topology in the function `make_host_du_dx_fxn` from `DualTopology` to `DualTopologyMinimization`.

(Unclear if these are neutral mutations vs. functionally important)

timemachine/md/minimizer.py
badisa: Will clarify, it is linear

timemachine/md/minimizer.py
badisa: They are pretty much interchangable, will stick with `DualTopologyMinimization` here to avoid confusion.

The other difference is that `DualTopology` ignores lambda and `DualTopologyMinimization` adjusts the w coord to cutoff at lamb = 1.0 and at lamb = 0.0 the w coord is zero.

tests/test_single_topology_confgen.py
```diff
@@ -1,15 +1,18 @@
...
 from timemachine.constants import DEFAULT_ATOM_MAPPING_KWARGS, DEFAULT_TEMP
 from timemachine.fe import atom_mapping, cif_writer, utils
-from timemachine.fe.lambda_schedule import construct_pre_optimized_relative_lambda_schedule
 from timemachine.fe.rbfe import HostConfig, setup_initial_states, setup_optimized_host
 from timemachine.fe.single_topology import AtomMapMixin, SingleTopology
 from timemachine.fe.utils import get_mol_name, read_sdf
 from timemachine.ff import Forcefield
 from timemachine.md import builders
 
+DEBUG = False
```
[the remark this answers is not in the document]
badisa: Renamed in https://github.com/proteneer/timemachine/pull/1339/commits/23dd804544745346f4843980aa83acaf317dc990

This saves a lot of time in the nightly tests (and locally)

timemachine/md/minimizer.py
```diff
@@ -140,85 +142,181 @@ def minimize_host_4d(
...
+        Wrapper class around a list of handlers
+
+    mol_coords: list of np.ndarray
+        Pre-specify a list of mol coords. Else use the mol.GetConformer(0)
+
+    n_steps_per_window: integer
+        Number of steps to run the FIRE minimizer with at each window
+
+    max_lambda: float
+        The largest lambda value to run a window at. If the value is 1.0 the mols will be
```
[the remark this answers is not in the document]
badisa: Clarified in https://github.com/proteneer/timemachine/pull/1339/commits/72480613629571dd7a5dc1edba95de5cf56b9220

timemachine/md/minimizer.py
badisa: Stick to using `DualTopologyMinimization` in https://github.com/proteneer/timemachine/pull/1339/commits/1699559316057dac6222b5686d34e6755e22c2ff

> changes the topology in the function make_host_du_dx_fxn from DualTopology to DualTopologyMinimization

Makes it possible to modify the cutoff value and use the function to run either barker/fire with this minimization approach.

timemachine/fe/rbfe.py
```diff
@@ -212,7 +212,7 @@ def setup_initial_states(
     temperature: float,
     lambda_schedule: Union[NDArray, Sequence[float]],
     seed: int,
-    min_cutoff: Optional[float] = 0.7,
+    min_cutoff: Optional[float] = None,
```
mcwitt, line 215: nit: slight preference for removing the default value here and explicitly passing `None` for solvent or `0.7` for complex.

timemachine/md/minimizer.py
badisa: Clarified in https://github.com/proteneer/timemachine/pull/1339/commits/ffdf00f744e046df632ea2b083ba6ce485231d2d

timemachine/fe/rbfe.py
mcwitt: Are we sure that removing this check in the solvent leg is what we want? I'm a bit unsure; can imagine that we'd want to know if the solvent leg is sampling a different conformational mode than the complex leg

timemachine/fe/rbfe.py
```diff
@@ -876,15 +880,19 @@ def run_solvent(
...
     if md_params is not None and md_params.water_sampling_params is not None:
         md_params = replace(md_params, water_sampling_params=None)
         warnings.warn("Solvent simulations don't benefit from water sampling, disabling")
     box_width = 4.0
-    solvent_sys, solvent_conf, solvent_box, solvent_top = builders.build_water_system(box_width, forcefield.water_ff)
+    solvent_sys, solvent_conf, solvent_box, solvent_top = builders.build_water_system(
+        box_width, forcefield.water_ff, mols=[mol_a, mol_b]
+    )
     solvent_box += np.diag([0.1, 0.1, 0.1])  # remove any possible clashes, deboggle later
     solvent_host_config = HostConfig(solvent_sys, solvent_conf, solvent_box, solvent_conf.shape[0])
```
mcwitt, line 893: nit, maybe future PR: would it be useful to move this into a reusable `get_solvent_host_config`?

tests/test_barostat.py
badisa: Fixed up in https://github.com/proteneer/timemachine/pull/1339/commits/d0b8f7c3fa1fe94fb68b4a3192830efefbc1a407

timemachine/md/builders.py
```diff
@@ -1,25 +1,90 @@
...
+def strip_units(coords) -> NDArray[np.float64]:
     return np.array(coords.value_in_unit_system(unit.md_unit_system))
 
 
-def build_protein_system(host_pdbfile: Union[app.PDBFile, str], protein_ff: str, water_ff: str):
+def remove_clashy_waters(
+    modeller: app.Modeller,
+    host_coords: NDArray[np.float64],
+    box: NDArray[np.float64],
+    water_idxs: NDArray,
```
mcwitt, line 22: nit:
```suggestion
    water_idxs: NDArray[np.int_],
```

timemachine/md/builders.py
```diff
@@ -1,25 +1,90 @@
...
+    clash_distance: float
+        Distance from a ligand atom to a water atom to consider as a clash, in nanometers
+    """
+    water_coords = host_coords[water_idxs]
+    ligand_coords = np.concatenate([get_romol_conf(mol) for mol in mols])
+    clashy_idxs = idxs_within_cutoff(water_coords, ligand_coords, box, cutoff=clash_distance)
+    if len(clashy_idxs) == 0:
+        return
+    # Offset the clashy idxs with the first atom idx, else could be pointing at non-water atoms
+    clashy_idxs += np.min(water_idxs)
```
mcwitt, line 54: Does this assume that `water_idxs` contains the indices of all of the water atoms in the solvent? If so, might note this explicitly in the docstring.

timemachine/md/builders.py
```diff
@@ -1,25 +1,90 @@
...
+    all_atoms = list(modeller.topology.atoms())
+    waters_to_delete = set()
+    for idx in clashy_idxs:
+        atom = all_atoms[idx]
+        waters_to_delete.add(atom.residue)
+        assert atom.residue.name == "HOH"
+    modeller.delete(list(waters_to_delete))
```
mcwitt, line 61: I'm slightly confused by the use of `set` here. Is the implication that multiple indices in `clashy_idxs` might refer to the same residue?

timemachine/md/minimizer.py
```diff
@@ -115,19 +113,23 @@ def shift(d, dr):
...
     mol_coords: Optional[List[NDArray]] = None,
-    windows: int = 50,
-    n_steps_per_window: int = 50,
-) -> np.ndarray:
-    """
-    Insert mols into a host system via 4D decoupling using Fire minimizer at lambda=1.0,
-    0 Kelvin Langevin integration at a sequence of lambda from 1.0 to 0.0, and Fire minimizer again at lambda=0.0
+    minimizer_steps_per_window: int = 500,
+    minimizer_windows: int = 2,
+    equilibration_steps: int = 1000,
```
mcwitt, line 123: nit: seems possibly useful to expose `max_lambda` for the minimizer here too

timemachine/md/builders.py
```diff
@@ -46,8 +111,14 @@ def build_protein_system(host_pdbfile: Union[app.PDBFile, str], protein_ff: str,
     solvated_host_coords = strip_units(modeller.positions)
 
     nha = host_coords.shape[0]
+    if mols is not None:
+        water_idxs = np.arange(nha, solvated_host_coords.shape[0])
```
mcwitt, line 115: nit: present before this PR, but seems worth replacing `nha` with a more readable name

tests/test_barostat.py
```diff
@@ -80,10 +82,14 @@ def test_barostat_with_clashes():
     mol_a, _, _ = get_hif2a_ligand_pair_single_topology()
     ff = Forcefield.load_from_file("smirnoff_1_1_0_sc.py")
 
-    # Set the lambda to 0.0 and don't minimize, resulting in clashes in the system
-    unbound_potentials, sys_params, masses, coords, box = get_solvent_phase_system(
-        mol_a, ff, lamb=0.0, minimize_energy=False
-    )
+    # Construct water box without removing the waters around the ligand to ensure clashes
+    host_system, host_coords, box, host_top = build_water_system(3.0, ff.water_ff)
```
mcwitt, line 86: nit: is it worth asserting that there actually are clashes here?

timemachine/md/minimizer.py
```diff
@@ -170,55 +196,131 @@ def minimize_host_4d(
...
 
     hgt = topology.HostGuestTopology(host_bps, top, host_config.num_water_atoms)
 
-    # this value doesn't matter since we will turn off the noise.
-    seed = 0
+    dt = 1.5e-3
+    friction = 1.0
 
-    intg = LangevinIntegrator(0.0, 1.5e-3, 1.0, combined_masses, seed).impl()
+    intg = LangevinIntegrator(DEFAULT_TEMP, dt, friction, combined_masses, seed).impl()
```
mcwitt, line 217: nit: would it be worth accepting a `temperature` argument here, versus assuming `DEFAULT_TEMP`?

mcwitt (approved):
Looks good, just non-blocking comments on my end.

timemachine/fe/rbfe.py
badisa: > can imagine that we'd want to know if the solvent leg is sampling a different conformational mode than the complex leg

I think we can want to know that the solvent leg is sampling a different mode, without failing the solvent leg for it. We are currently enforcing that the solvent leg has to have the same (or at least similar) conformational mode as complex, which I don't think is generalizable.

timemachine/md/builders.py
badisa: It does will clarify.

timemachine/md/builders.py
badisa: If you think of a water molecule that is completely overlapping with a ligand, that will be three atoms that clash which all point to the same residue. Maybe the modeller handles duplicates, but seemed cleanest to deduplicate up front

timemachine/md/minimizer.py
badisa: Lol it accepts the temperature parameter and I just don't use it. Will fix it.

timemachine/md/builders.py
mcwitt: Makes sense. Was just confused and thinking the list was at the level of individual atoms vs. water molecules.

timemachine/fe/rbfe.py
badisa: Going to leave the parameter as a keyword argument for now, but probably a good deboggling task.

timemachine/md/builders.py
badisa: Renamed in https://github.com/proteneer/timemachine/pull/1339/commits/920122027a067705dacb966f7e50808bd83b6b1c

timemachine/md/minimizer.py
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1339/commits/130b5e2c1ae9033dc9494a224b4888feced19801

timemachine/md/minimizer.py
badisa: Exposed in https://github.com/proteneer/timemachine/pull/1339/commits/52d65dc1a9fcd7eb8dc6e9a2d4a8a6b6a14db2a0

tests/test_barostat.py
badisa: Asserted in https://github.com/proteneer/timemachine/pull/1339/commits/2c68f3b702c6ef4618d26b1b04efac42cb6a5545

timemachine/md/builders.py
badisa: Clarified in https://github.com/proteneer/timemachine/pull/1339/commits/957775ecd2ab68f613f679a880bdcbea47f5bdac

timemachine/md/minimizer.py
jkausrelay: Makes sense. May be worth emphasizing in the doc str.

tests/test_barostat.py
```diff
@@ -323,19 +323,19 @@ def test_barostat_varying_pressure():
...
     ten_atm_box_vol = compute_box_volume(ten_atm_box)
     # Expect the box to shrink thanks to the barostat
-    assert compute_box_volume(complex_box) - ten_atm_box_vol > 0.4
+    assert compute_box_volume(box) - ten_atm_box_vol > 0.4
 
     # Set the pressure to 1 atm
     baro.set_pressure(DEFAULT_PRESSURE)
-    # Changing the barostat interval resets the barostat step.
-    baro.set_interval(2)
+    # Changing the step to 0
```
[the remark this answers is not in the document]
badisa: Removed in https://github.com/proteneer/timemachine/pull/1339/commits/6db2adfba9c1084d13da281f78f0526cc231a9d6

timemachine/md/minimizer.py
badisa: Emphasized in https://github.com/proteneer/timemachine/pull/1339/commits/ebec3f0a3b723a58d1b48eca0307c7761971e60b

jkausrelay (approved):

maxentile (approved):

PR #1342, opened 2024-07-15
Perf/verify_chiral_consistency_of_core

maxentile:
Implements the second option in https://github.com/proteneer/timemachine/pull/1315#discussion_r1621470240 , maybe ~10x faster on an example than using the `"smirnoff_1_1_0_sc.py"` file.

<details>

```python
from timemachine.fe.single_topology import verify_chiral_consistency_of_core
from timemachine.testsystems.relative import get_hif2a_ligand_pair_single_topology

mol_a, mol_b, core = get_hif2a_ligand_pair_single_topology()

default_ff = Forcefield.load_default()
sc_ff = Forcefield.load_from_file("smirnoff_1_1_0_sc.py")
placeholder_ff = Forcefield.load_from_file("placeholder_ff.py")

# ~500ms
_ = verify_chiral_consistency_of_core(mol_a, mol_b, core, default_ff)

# ~500ms
_ = verify_chiral_consistency_of_core(mol_a, mol_b, core, sc_ff)

# ~30ms
_ = verify_chiral_consistency_of_core(mol_a, mol_b, core, placeholder_ff)
```

</details>

Adds a `"placeholder_ff.py"` file (with a single generic type per potential) to `ff/params` in case useful in similar contexts. (wherever it might be convenient to go through a `.parameterize(mol)` code path to get `potential.idxs` but `potential.params` would be unnecessary, and you want to minimize the number of unnecessary smarts queries.)

### DIFF
--- a/tests/test_single_topology.py
+++ b/tests/test_single_topology.py
@@ -1442,7 +1442,7 @@ $$$$""",
         removeHs=False,
     )
 
-    ff = Forcefield.load_from_file("smirnoff_1_1_0_sc.py")
+    ff = Forcefield.load_from_file("placeholder_ff.py")
     core = np.array([[0, 0], [1, 1], [2, 2], [3, 3], [4, 4], [5, 5], [6, 6]])
 
     with pytest.raises(ChiralConversionError):
--- a/timemachine/ff/make_placeholder_ff.py
+++ b/timemachine/ff/make_placeholder_ff.py
@@ -1,0 +1,65 @@
+import numpy as np
+
+from timemachine.ff import Forcefield
+from timemachine.ff.handlers.bonded import (
+    HarmonicAngleHandler,
+    HarmonicBondHandler,
+    ImproperTorsionHandler,
+    ProperTorsionHandler,
+)
+from timemachine.ff.handlers.nonbonded import (
+    LennardJonesHandler,
+    LennardJonesIntraHandler,
+    LennardJonesSolventHandler,
+    SimpleChargeHandler,
+    SimpleChargeIntraHandler,
+    SimpleChargeSolventHandler,
+)
+
+# bonded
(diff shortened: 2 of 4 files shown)


mcwitt:
Nice, clever trick to greatly improve the performance of the check with a minimal change by introducing a placeholder forcefield.

Completely agreed that the minimal/safe change seems best for now, but I'm curious how much faster do you think this could be made with a more invasive change (e.g. by avoiding any SMIRKS matching and extracting the bond list directly; not parameterizing angles, torsions, chiral bonds, etc.)?

mcwitt (approved):

maxentile:
> Completely agreed that a minimal/safe change seems best for now, but I'm curious how much faster you think this could be made with a more invasive change (e.g. by avoiding any SMIRKS matching and extracting the bond list directly; not parameterizing angles, torsions, chiral bonds, etc.)?

I'm unsure -- but based on initial profiling, I'd guess maybe another ~2x from going further in that direction, since the fraction of time spent in `match_smirks` is maybe down from ~95% to ~50% after this change.

A more substantive change (e.g. re-writing the assertion in such a way that it incurs less python overhead) might be needed for further improvement.

badisa:
Should this modify `verify_chiral_consistency_of_core` to use this placeholder forcefield internally and remove the forcefield parameter?

Would avoid the user having to remember to switch out the FF

badisa (approved):

maxentile:
> Should this modify verify_chiral_consistency_of_core to use this placeholder forcefield internally and remove the forcefield parameter?
> Would avoid the user having to remember to switch out the FF

I'll leave it as-is for now to avoid changing the existing signature, but removing the argument sounds like a potentially useful simplification later (assuming we are unsuccessful in modifying the protocol to remove the need for the assertion).

PR #1343, opened 2024-07-16
[WIP] Stricter chiral check in _get_cores_impl

maxentile:
Modify `_get_cores_impl` to avoid returning any solutions that would fail an assertion error in `verify_chiral_consistency_of_core`.


maxentile:
needs further investigation

PR #1324, opened 2024-06-06
Resolve inconsistency between bar_with_bootstrapped_uncertainty and df_and_err_from_u_kln

badisa:
* `bar_with_bootstrapped_uncertainty` can return an error of zero when `df_and_err_from_u_kln` returns a large error. I don't think this is actually an issue in practice, but seems like it would be good to address in case we ever use the error for something.

PR #1344, opened 2024-07-16
Fixes bar_with_pessimistic_uncertainty convergence error

badisa:
* A bug introduced in https://github.com/proteneer/timemachine/pull/1324

### DIFF
--- a/tests/test_bar.py
+++ b/tests/test_bar.py
@@ -180,6 +180,11 @@ def test_df_from_u_kln_does_not_raise_on_incomplete_convergence():
     assert np.isfinite(df)
     assert np.isnan(ddf)  # returns NaN for uncertainty on incomplete convergence
 
+    bootstrap_df, bootstrap_ddf = bar_with_pessimistic_uncertainty(u_kln, maximum_iterations=1)
+    np.testing.assert_equal(df, bootstrap_df)
+    # With the bootstrapping, we will compute a finite error
+    assert np.isfinite(bootstrap_ddf)
+
 
 def test_pair_overlap_from_ukln():
     # identical distributions
--- a/timemachine/fe/bar.py
+++ b/timemachine/fe/bar.py
@@ -179,12 +179,8 @@ def bootstrap_bar(
     * TODO[deboggle] -- upgrade from pymbar3 to pymbar4 and remove this
     * TODO[performance] -- multiprocessing, if needed?
     """
-    u_kn, N_k = ukln_to_ukn(u_kln)
-    mbar = pymbar.MBAR(u_kn, N_k, maximum_iterations=maximum_iterations)
 
-    df, ddf = mbar.getFreeEnergyDifferences()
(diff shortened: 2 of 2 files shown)


tests/test_bar.py
```diff
@@ -180,6 +180,11 @@ def test_df_from_u_kln_does_not_raise_on_incomplete_convergence():
     assert np.isfinite(df)
     assert np.isnan(ddf)  # returns NaN for uncertainty on incomplete convergence
 
+    bootstrap_df, bootstrap_ddf = bar_with_pessimistic_uncertainty(u_kln, maximum_iterations=1)
```
badisa, line 183: This raises an exception without the fix. Didn't launch a CI job as it is an unambigious failure

timemachine/fe/bar.py
```diff
@@ -179,12 +179,8 @@ def bootstrap_bar(
     * TODO[deboggle] -- upgrade from pymbar3 to pymbar4 and remove this
     * TODO[performance] -- multiprocessing, if needed?
     """
-    u_kn, N_k = ukln_to_ukn(u_kln)
-    mbar = pymbar.MBAR(u_kn, N_k, maximum_iterations=maximum_iterations)
 
-    df, ddf = mbar.getFreeEnergyDifferences()
-    full_bar_result = df[0, 1]
-    full_bar_err = ddf[0, 1]
+    full_bar_result, full_bar_err = df_and_err_from_u_kln(u_kln, maximum_iterations=maximum_iterations)
```
badisa, line 183: df_and_err_from_u_kln handles the behavior correctly: https://github.com/proteneer/timemachine/blob/master/timemachine/fe/bar.py#L136-L139

maxentile (approved):

timemachine/fe/bar.py
```diff
@@ -193,6 +189,9 @@ def bootstrap_bar(
     seed = 2022
     rng = np.random.default_rng(seed)
 
+    u_kn, N_k = ukln_to_ukn(u_kln)
+    mbar = pymbar.MBAR(u_kn, N_k, maximum_iterations=maximum_iterations)
```
mcwitt, line 193: nit: this seems unnecessary vs. setting `initial_f_k=None` below (or maybe better, `initial_f_k=np.array([0.0, df])`)

mcwitt (approved):
Good catch. Makes sense to reuse `df_and_err_from_u_kln` since the latter has the desired workaround.

timemachine/fe/bar.py
badisa: Good call, will replace it with `initial_f_k=np.array([0.0, df])` which is the same value without constructing an extra MBAR object

badisa: Set using pre-existing value in https://github.com/proteneer/timemachine/pull/1344/commits/a7de23f8700f0af98ed72ea49ec221fdff85566f

PR #1346, opened 2024-07-18
Modified Minimization while keeping number of host atoms constant

badisa:
Most of this work was originally done in https://github.com/proteneer/timemachine/pull/1339 this branch has been rebased off of master and has added a single commit to resolve the issue of different molecules producing systems with different numbers of atoms. All of the relevant changes are in 41342275930b0095f21acba9f2bb3a47fa73d35c and https://github.com/proteneer/timemachine/pull/1346/commits/1a2610dd5d5ea6fa285ef9539e6c12b94817cd21

## Changes in this PR
* Add in the number of waters that would be deleted, only deleting waters added by `addSolvent` so that the system size is always identical.
* Compute the box dimensions after solvation due to adding waters beyond the original box size specified in the first addSolvent call. Slightly increases size of boxes returned from the builders. Should revisit the box padding in the near future.
* Include the number of waters in provided PDBs when returning the number of waters in `build_protein_system`
* Assert that the PDB has the waters at the end of the file, otherwise breaks expectations downstream.

##  Changes from https://github.com/proteneer/timemachine/pull/1339 
* Completely freeze the ligands in place, so that the host coordinates aren't minimized to different poses than what the system expects downstream.
* Delete clashy waters upon system set up, which resolves issues with cagey molecules (where water could be trapped inside)
* Adds pre-equillibration to reduce 'air bubbles' in the simulations (does not modify ligand).
* Centers solvent systems around ligands when building systems to ensure consistent water box construction. IE such that the ligands don't get placed in the margin setup within the box and minimization of solvent boxes are dependent on the starting point of ligands.
* Decoupling is still required due to eps values of 0.0 in protein hydrogens, but now minimize at lambda = 0.1 and 0.0 (can be adjusted), rather than lowering down from 1.0

## Systems tested
Verified that the following public systems and the corresponding ligands minimize in the solvent/complex (including through bisection). 

Sets taken from 
* Hif2a
* PFKFB3 - Added to the tests since it has a clash that blows up
* TYK2
* CDK8 - The original set that required our previous implementation of minimization


jkausrelay (approved):

timemachine/md/builders.py
```diff
@@ -34,45 +129,101 @@ def build_protein_system(host_pdbfile: Union[app.PDBFile, str], protein_ff: str,
...
-    box_lengths = np.amax(host_coords, axis=0) - np.amin(host_coords, axis=0)
+    water_residues_in_pdb = [residue for residue in host_pdb.topology.residues() if residue.name == WATER_RESIDUE_NAME]
+    num_host_atoms = host_coords.shape[0]
+    if len(water_residues_in_pdb) > 0:
+        host_water_atoms = len(water_residues_in_pdb) * 3
+        # Only consider non-water atoms as the host, does count excipients as the host
+        num_host_atoms = num_host_atoms - host_water_atoms
+        water_indices = np.concatenate([[a.index for a in res.atoms()] for res in water_residues_in_pdb])
+        expected_water_indices = np.arange(host_water_atoms) + num_host_atoms
+        assert np.all(expected_water_indices == water_indices), "Waters in PDB must be at the end of the file"
```
maxentile, line 140: nice addition: previously unsafe assumption.

nit: might save a few debugging seconds by printing the mismatch
```suggestion
        np.testing.assert_equal(expected_water_indices, water_indices, err_msg="Waters in PDB must be at the end of the file")
```

maxentile (approved):

tests/test_barostat.py
```diff
@@ -80,10 +82,16 @@ def test_barostat_with_clashes():
...
     ff = Forcefield.load_from_file("smirnoff_1_1_0_sc.py")
 
-    # Set the lambda to 0.0 and don't minimize, resulting in clashes in the system
-    unbound_potentials, sys_params, masses, coords, box = get_solvent_phase_system(
-        mol_a, ff, lamb=0.0, minimize_energy=False
-    )
+    # Construct water box without removing the waters around the ligand to ensure clashes
+    host_system, host_coords, box, host_top = build_water_system(3.0, ff.water_ff)
+    # Shrink the box to ensure the energies are NaN
+    box -= np.eye(3) * 0.1
```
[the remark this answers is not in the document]
badisa: With the box sizes being slightly larger now due to the additional waters on the edge of the box, the energy is now -2e8 rather than NaN. Reducing the box takes it back to Nan. The test passes either way to but just to be sure the energies are NaN I reduced the box as well.

 
Don't think it particularly matters as really the test is just to verify that if the system is undefined (NaN) that we won't make any barostat moves.

tests/common.py
```diff
@@ -61,8 +61,9 @@ def prepare_single_topology_initial_state(st: SingleTopology, host_config: Optio
     host = None
     if host_config is not None:
         host = rbfe.setup_optimized_host(st, host_config)
+    is_complex_leg = host_config is not None and len(host_config.conf) == host_config.num_water_atoms
```
mcwitt, line 64: I'm probably missing something here, but this looks to me like it should be `is_solvent_leg`? (all host atoms are waters?)

badisa: You aren't missing anything, that check is wrong. Fixed in https://github.com/proteneer/timemachine/pull/1346/commits/952dc8b069b28014c4b5d232ab8934880b21045f

tests/test_builders.py
```diff
@@ -0,0 +1,146 @@
...
+        check_force_norm(-du_dx)
+
+
+@pytest.mark.nocuda
+def test_build_protein_system_returns_correct_water_count():
+    with resources.path("timemachine.datasets.fep_benchmark.pfkfb3", "6hvi_prepared.pdb") as pdb_path:
+        protein_system, protein_coords, box, _, num_water_atoms = build_protein_system(
+            str(pdb_path), DEFAULT_PROTEIN_FF, DEFAULT_WATER_FF
+        )
+        assert protein_coords.shape[0] - num_water_atoms == 6748
```
mcwitt, line 59: nit: does 6748 need to be hardcoded here? (If so, could we add a brief comment on how to derive this value from the test data?)

badisa: It does, as previously this value was wrong. Added a comment explaining why the value is hard coded and how to determine the value: https://github.com/proteneer/timemachine/pull/1346/commits/a86d9b7b516fe99f381a1801a6e901eb4f700f5c

timemachine/md/builders.py
```diff
@@ -34,45 +129,103 @@ def build_protein_system(host_pdbfile: Union[app.PDBFile, str], protein_ff: str,
...
+        # If mols provided, center waters such that the center is the mols centroid
+        # Done to avoid placing mols at the edges and moves the water coordinates to avoid
+        # changing the mol coordinates which are finalized downstream of the builder
+        if mols is not None and len(mols) > 0:
+            mol_coords = np.concatenate([get_romol_conf(mol) for mol in mols])
+            mols_centroid = np.mean(mol_coords, axis=0)
+            box_extents = (np.max(host_coords, axis=0) - np.min(host_coords, axis=0)) * 0.5
+            box_center = np.min(host_coords, axis=0) + box_extents
+            host_coords = host_coords - box_center + mols_centroid
+        return host_coords
```
mcwitt, line 215: nit: this looks correct to me, but seems worth testing. Do you think it's worth moving it up to module scope and adding a test?

tests/test_builders.py
```diff
@@ -0,0 +1,146 @@
...
+    new_ligand_center = mol_protein_coords[num_host_atoms_with_mol + water_atom_idx]
+    for mol in [mol_a, mol_b]:
+        conf = get_romol_conf(mol)
+        centroid = np.mean(conf, axis=0)
+        conf = conf - centroid + new_ligand_center
+        set_romol_conf(mol, conf)
+    _, moved_conf, moved_box, _, num_water_after_moved = build_protein_system(
+        host_pdbfile, DEFAULT_PROTEIN_FF, DEFAULT_WATER_FF, mols=[mol_a, mol_b]
+    )
+    assert num_water_after_moved == num_water_atoms_with_mols
```
mcwitt, line 146: nit: as an alternative to the earlier suggestion to unit test `get_host_coords`, is it worth adding a couple of assertions here to check that the water box is now centered on the ligand centroid?

mcwitt (approved):

timemachine/md/builders.py
badisa: Added a test in the builders in https://github.com/proteneer/timemachine/pull/1346/commits/d603cc0c9df8ba19f4391e3daad90185fdd80c5d

Verifies that if you pass ligands it changes the centroid to match the ligands and without passing the mols the centroid doesn't match.

PR #1347, opened 2024-07-19
Add lambda schedule tuned to Bisection

badisa:
For vacuum it makes very little different to minimize more states than required (2 * (num_windows - 2)), but in solvent/complex it can get expensive. Seen in 30k atom systems that minimizing (local minimization) a new window takes 20-30 seconds which is ~15 minutes or so (for num windows = 48) spent minimizing states that aren't actually used.

| Master | PR |
|--------|--------|
| vacuum_100_benchmarks | vacuum_100_benchmarks_bisection |
| solvent_100_benchmarks | solvent_100_benchmarks_bisection |
| complex_100_benchmarks | complex_100_benchmarks_bisection |

The performance improvement with this change is about 10% (vacuum) to 20% (solvent/complex). Though it is worth noting that **the performance improvement is related to the number of windows needed and what the maximum number of windows are**. The performance gain will be O(1) and **not scale with the number of frames**, the benchmarks are done here with 100 frames to match the bisection stage of HREX/Bisection. 

Note that the bisection timings after this change are noisier since there are still new initial states that have to be minimized depending on schedule generated bisection. This should be strictly faster than the previous implementation (where num windows are even).


timemachine/fe/lambda_schedule.py
```diff
@@ -19,6 +22,38 @@ def interpolate_pre_optimized_protocol(pre_optimized_protocol, num_windows):
...
+            insert_idx = diff_idx + 1
+        else:
+            # Find the max from the upper end of the lambda schedule
+            diff_idx = (len(lambda_schedule_) - 1) - np.argmax(lambda_diffs[::-1])
+            new_lamb = (lambda_schedule_[diff_idx] + lambda_schedule_[diff_idx - 1]) / 2
+            insert_idx = diff_idx
+        lambda_schedule_.insert(insert_idx, new_lamb)
+
+    lambda_schedule = np.array(lambda_schedule_)
+    return lambda_schedule
```
mcwitt, line 54: I think this could reuse the same logic used in the bisection on the BAR overlap objective, if we use `greedy_bisection_step` with the appropriate choice of `local_cost` and `make_intermediate`. Example:

```python
import matplotlib.pyplot as plt
import numpy as np

from timemachine.fe.protocol_refinement import greedy_bisection_step


def local_cost(lam1, lam2):
    return np.abs(lam1 - lam2)


def make_intermediate(lam1, lam2):
    return (lam1 + lam2) / 2


def step(protocol):
    protocol_, _ = greedy_bisection_step(protocol, local_cost, make_intermediate)
    return protocol_


protocol = [0.0, 1.0]

for n in range(30):
    plt.cla()
    protocol = step(protocol)
    for lam in protocol:
        plt.axvline(lam)
    plt.show()
```

mcwitt: That said, would it be sufficient to just use $2^n + 1$ linearly-spaced values?

The spacings you'd get with the overlap objective for numbers of windows between these values seems sensitive to the details of the transformation, so I'm not sure they'd line up with those generated by this algorithm very well.

badisa: > That said, would it be sufficient to just use $2^n + 1$ linearly-spaced values?
Because we do an equality check, and np.linspace produces floating point articles (0.5 not in all spacings, sometimes it is ~0.500000001 which gets treated as a different state downstream), we could also change the equality to `np.allclose` and set some threshold where we don't think there is a difference in lambda schedule.

> The spacings you'd get with the overlap objective for numbers of windows between these values seems sensitive to the details of the transformation, so I'm not sure they'd line up with those generated by this algorithm very well.

Definitely, this is not going to address all cases. But based on the benchmarks I think this is a cheap improvement that for big systems is going to make a decent wall clock improvement on.

> I think this could reuse the same logic used in the bisection on the BAR overlap objective, if we use greedy_bisection_step with the appropriate choice of local_cost and make_intermediate

Will think about how to construct it using this approach. The one difference is that this will always fill in the 'smallest' gap, which will go from the start of the array and bias towards having more evenly spaced windows. Which is why the existing code.

Such as 
[0, 1.0] - Evenly split between >= 0.5 and < 0.5
[0, 0.5, 1.0] - Odd, but odd number 
[0, 0.25, 0.5, 1.0] - Off by one
[0, 0.25, 0.5, 0.75, 1.0] - Evenly split
[0, 0.125, 0.25, 0.5, 0.75, 1.0] - off by one
[0, 0.125, 0.25, 0.375, 0.5, 0.75, 1.0] - off by two

The difference will just continue growing as the number of windows. It does shrink back down eventually, but it becomes dependent on the number of windows

mcwitt:
> we could also change the equality to np.allclose

I'd be in favor of switching to `np.allclose` on lambda for determining whether two states are equivalent. This seems safe as long as the parameter interpolation is continuous, which it should be.

> The one difference is that this will always fill in the 'smallest' gap, which will go from the start of the array and bias towards having more evenly spaced windows. Which is why the existing code.

I'm not sure I understand the motivation here. If we assume that the parameter interpolation is close to optimal, then bisection will achieve equal spacing when the number of windows is $2^n + 1$ for some integer $n$. But for numbers of windows in between, the order in which we bisect the pairs is arbitrary, determined by details of the transformation and/or sampling noise. From this perspective, it seems simplest to me to choose for the size of the lambda grid the smallest $2^n + 1$ larger than the number of desired windows, or the largest $2^n + 1$ less than the number of desired windows and randomly bisect some pairs to make up the difference.

timemachine/fe/lambda_schedule.py
badisa: Switched to the approach you suggested in https://github.com/proteneer/timemachine/pull/1347#issuecomment-2240623648 as part of https://github.com/proteneer/timemachine/pull/1347/commits/a08b2466834d627b30d73dc3f1551b1ca0632dc8

badisa: The implementation will probably change the benchmarks slightly in some cases, but running the vacuum legs real quickly the original states are being picked up. May make some of the benchmarks look worse since fewer windows are generated upfront, but that is more of an issue of hiding the construction time as part of the benchmark. This way we will generate strictly fewer windows than previously.

mcwitt (approved):
LGTM. I added @maxentile for a second pair of eyes since we've previously discussed similar things.

maxentile (approved):

maxentile:
> I'm not sure I understand the motivation here. If we assume that the parameter interpolation is close to optimal, then bisection will achieve equal spacing when the number of windows is 2^n + 1 for some integer n. But for numbers of windows in between, the order in which we bisect the pairs is arbitrary, determined by details of the transformation and/or sampling noise. 

Similar reservation

timemachine/fe/lambda_schedule.py
```diff
@@ -19,6 +22,20 @@ def interpolate_pre_optimized_protocol(pre_optimized_protocol, num_windows):
     return lambda_schedule
 
 
+def bisect_lambda_schedule(
+    num_windows: int, lambda_interval: Optional[Tuple[float, float]] = None
+) -> NDArray[np.float64]:
+    """
+    Construct a lambda schedule with windows of size 2^N + 1 such that 2^N is the closest value to num_windows that is smaller.
```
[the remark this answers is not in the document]
badisa: > Would it be fair to describe this as a sort of eager pre-computation?

Exactly.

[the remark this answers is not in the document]
badisa: > Can't easily parse the first line of description
Will try to make it more interpretable. 

> the code doesn't appear to match the function name (function name: "bisect", implementation: "linspace"), but the implementation seems reasonable and useful.

I want the name to indicate that this should be used with bisection, as it is otherwise a naive lambda schedule. And how exactly the lambda schedule gets constructed should be independent (`np.linspace`, `greedy_bisection_step`) of its usage. At least that was how I thought about it.

timemachine/fe/lambda_schedule.py
```diff
@@ -19,6 +22,18 @@ def interpolate_pre_optimized_protocol(pre_optimized_protocol, num_windows):
...
 
+def bisect_lambda_schedule(num_windows: int, lambda_interval: Tuple[float, float] = (0.0, 1.0)) -> NDArray[np.float64]:
+    """
+    Construct a lambda schedule with windows of size 2^N + 1 such that 2^N is the closest value to num_windows that is smaller.
+    This means that the number of windows returned may not match num_windows. Useful in the context of `run_sims_bisection`
+    where states are created on the fly with these same lambda values. Can save the cost of minimizing more often than necessary.
+    """
+    assert num_windows >= 2
+    min_lamb, max_lamb = lambda_interval
+    schedule_windows = int(2 ** np.floor(np.log2(num_windows))) + 1
```
maxentile, line 39: Small surprise: `len(schedule_windows) == 33` when `num_windows = 48`. Is this significant? Would we want this to be the same length as `num_windows`? (i.e. return the first `num_windows` points visited by this process https://github.com/proteneer/timemachine/pull/1347#discussion_r1685069111 ?)

timemachine/fe/lambda_schedule.py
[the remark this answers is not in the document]
badisa: Renamed function to be `bisection_lambda_schedule`, which makes it sound more agnostic to the implementation: https://github.com/proteneer/timemachine/pull/1347/commits/c432836f2b3514f85dd10fbd6b3f35add44ab37f

Attempted to make the docstring clearer in https://github.com/proteneer/timemachine/pull/1347/commits/b971781149307aef05cd7f5b4e1701d44fdda551

timemachine/fe/lambda_schedule.py
badisa: I don't think this is significant, unless we think that there is a lambda step size that we don't trust local minimization to perform well under. In this case any optimization step size is 1/33 (~0.03) which seems like a pretty small lambda step. If we think there need to be a certain number of initial states for this to be trustworthy, we probably should also have some warning of running with num_windows < threshold (Does num_windows == 3 become untrustworthy?)

Will run a hif2a graph to verify performance is unchanged.

badisa:
Closing this, seems to cause issues in an edge in hif2a (complex leg is off by 3kcal from without this change) and until I understand why exactly going to close.

badisa:
After running the hif2a edge (15 -> 30) five times on https://github.com/proteneer/timemachine/commit/30e2a506f8fdda10a64af91aea40a1f7d6678774 vs 350ae7ac9f820dfaf9479a9533b5efab40fb8969 the large run to run variance was not a result of this PR, rather just got unlucky.

Still tracking down a way to identify what exactly is triggering this variance, but don't think it needs to delay this PR.

The five runs of the edge (complex)
image

PR #1348, opened 2024-07-19
Fixes conditional that triggered warning incorrectly

badisa:
* If steps_per_frame was greater than the water sampling interval would post a message, which is incorrect

### DIFF
--- a/tests/test_hrex_rbfe.py
+++ b/tests/test_hrex_rbfe.py
@@ -2,6 +2,7 @@ from dataclasses import replace
 from importlib import resources
 from typing import Optional
 from unittest.mock import patch
+from warnings import catch_warnings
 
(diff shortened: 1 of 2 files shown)


tests/test_hrex_rbfe.py
```diff
@@ -97,7 +133,7 @@ def sample_and_record_rss(*args, **kwargs):
             md_params,
             lambda_interval=(0.0, 0.15),
             n_windows=n_windows,
-            min_cutoff=0.7 if host_name is not None else None,
+            min_cutoff=0.7 if host_name == "complex" else None,
```
badisa, line 136: Should have been changed in #1346

mcwitt (approved):

PR #1349, opened 2024-07-22
Add atom mapping arguments to restrict the number of connected components in the core

mcwitt:
Removes the `connected_core` option from `get_cores` and replaces with 2 more general options:

* `max_connected_components`: set to $k$, only allow cores with at most $k$ connected components. To recover the current behavior with `connected_core=True`, set to 1. For `connected_core=False`, set to `None`.
* `min_connected_component_size`: set to $n$, only allow cores with the size of the smallest connected component being $\geq n$

For now, keeps the current default behavior by setting `max_connected_components=1` and `min_connected_component_size=1` in `DEFAULT_ATOM_MAPPING_KWARGS`.

For equivalent settings, ~performance is slightly decreased (5-10% in my testing). This could probably be optimized more if necessary, but I was starting to hit diminishing returns (and think it might be preferable to prioritize clarity of the code over few-percent speedups here).~ After optimizations in https://github.com/proteneer/timemachine/pull/1349/commits/fba342f2c25b0847433e618f9c2ea1b1c902302e, it's about 10% faster than the current implementation with default settings.

Increasing either `max_connected_components` or `min_connected_component_size` from their defaults appears to drastically reduce search space and improve performance in some examples. The intuition seems to be that allowing many smaller connected components means that we find a larger overall core (and thus fewer outer loops over decreasing core size), but are less able to prune the search tree at each iteration; the optimal values are likely somewhere in between.


timemachine/fe/mcgregor.py
```diff
@@ -135,45 +126,105 @@ def __init__(self, n_vertices, edges):
...
+        for v in mapped_nodes:
+            if v not in seen:
+                # visit component containing v
+                seen.add(v)
+                cc_size = 1
+                nextlevel = [v]
+                while nextlevel:
+                    thislevel = nextlevel
+                    nextlevel = []
+                    for v in thislevel:
```
badisa, line 181: nit: shadowing the original v value, this is safe as written but could bite someone later

timemachine/fe/mcgregor.py
```diff
@@ -367,43 +421,24 @@ def recursion(
...
-                    g1_node_states[a1] = NODE_STATE_VISITED_MAPPED
-                    g1_source = a1
-                    g1_mapped_count += 1
-            else:
-                g1_node_states[a1] = NODE_STATE_UNVISITED
-
-        if g1_source and g1.mapping_is_disconnected(g1_source, g1_node_states, g1_mapped_count):
-            return
+    if max_connected_components is not None or min_connected_component_size > 1:
+        g1_mapped_nodes = {a1 for a1, a2 in enumerate(atom_map_1_to_2[:layer]) if a2 != UNMAPPED}
```
badisa, line 425: Nice improvement to the code, but with it being so much more concise it would be useful to have a brief comment about the intention

timemachine/fe/mcgregor.py
```diff
@@ -367,43 +421,24 @@ def recursion(
...
-                if a2 != UNMAPPED:
-                    g1_node_states[a1] = NODE_STATE_VISITED_MAPPED
-                    g1_source = a1
-                    g1_mapped_count += 1
-            else:
-                g1_node_states[a1] = NODE_STATE_UNVISITED
-
-        if g1_source and g1.mapping_is_disconnected(g1_source, g1_node_states, g1_mapped_count):
-            return
+    if max_connected_components is not None or min_connected_component_size > 1:
```
badisa, line 424: It looks like if `min_connected_component_size == 0 ` then `max_connected_components` is ignored. Doesn't look like this case is tested in the code. Should it raise an exception in this case?

mcwitt: Sorry, not following... If `min_connected_component_size == 0`, doesn't the condition reduce to `max_connected_components is not None`?

badisa: Sorry, misunderstood the code.

But in the case of `max_connected_components is not None` and `min_connected_component_size = 0`, should that be allowable? If so would be good to test that case

mcwitt: I think it should be allowable -- `min_connected_component_size = 0` is just a no-op filter (same as the default value of 1, or any negative number), so I think this is already covered by `test_max_connected_components`. But this raises the point that we have a test with both filters active at the same time (`max_connected_components is not None and min_connected_component_size > 1`); will consider adding that.

badisa: If it is allowable, would be good to test for it explicitly then.

timemachine/fe/mcgregor.py
mcwitt: Renamed to avoid shadowing in https://github.com/proteneer/timemachine/pull/1349/commits/c3d4bcbacf81018878d00c54d07eaf94e5493c1f

badisa (approved):

timemachine/fe/mcgregor.py
mcwitt: Added comments in https://github.com/proteneer/timemachine/pull/1349/commits/b27be3af5e0b6c90391ccfcee77dcee4bfb0fad4; hopefully these make the logic clearer

timemachine/fe/mcgregor.py
mcwitt: Added tests for both filters active at the same time in https://github.com/proteneer/timemachine/pull/1349/commits/471ceb3eaf5ae1356b7f21f922617fd9eb77152f, and for `min_connected_component_size=0` in https://github.com/proteneer/timemachine/pull/1349/commits/c0fe6a65f8a9e1285813f1dcd6bf2567e258ba3c

timemachine/fe/mcgregor.py
```diff
@@ -135,45 +126,105 @@ def __init__(self, n_vertices, edges):
...
-        For example, let M be mapped node, D be a demapped node, and U be an unvisited node:
+    def mapping_is_disconnected(
+        self,
+        mapped_nodes: Set[int],
+        unvisited_nodes: Set[int],
+        max_connected_components: Optional[int],
+        min_connected_component_size: int,
+    ):
+        r"""Verify whether mapped nodes can still be connected at least the specified minimum number of connected
+        components. It is assumed that source is a mapped node. i.e. node_states[source] == NODE_STATE_VISITED_MAPPED.
```
proteneer, line 137: Stale documentation: I don't think we use node_states anymore.

timemachine/fe/mcgregor.py
```diff
@@ -135,45 +126,105 @@ def __init__(self, n_vertices, edges):
...
+
+        # Uncomment to assert result of fast implementation matches the reference
+        # args = (mapped_nodes, unvisited_nodes, max_connected_components, min_connected_component_size)
+        # assert self.mapping_is_disconnected_fast(*args) == self.mapping_is_disconnected_ref(*args)
+
+        return self.mapping_is_disconnected_fast(
+            mapped_nodes, unvisited_nodes, max_connected_components, min_connected_component_size
+        )
+
+    def mapping_is_disconnected_fast(
```
proteneer, line 161: how much faster is `_fast` relative to `_ref`?

proteneer: can we add some tests just to compare the correctness of the `_fast` version to `_ref` given the increased complexity in logic? - eg. on random graphs, molecules from hif2a set, etc, with arbitrary choices of mapped_nodes, unvisited nodes, max_connected_components, etc.

timemachine/fe/mcgregor.py
```diff
@@ -362,7 +416,8 @@ def recursion(
     max_cores,
     threshold,
     enforce_core_core,
-    connected_core,
+    max_connected_components: Optional[int],
```
proteneer, line 419: This is a tough parameter for me to figure out:

if `max_connected_components=0` or `None` - then we don't actually mean zero connected components, but effectively we are saying `max_connected_components=max(len(g1.nodes), len(g2.nodes))`. I'm wondering if this should default to `np.inf` to be more explicit? 

This would have the benefit of simplifying the short-circuit logic above:

```
if max_connected_components and n_ccs_with_mapped_nodes > max_connected_components:
    return True
```

timemachine/fe/mcgregor.py
```diff
@@ -135,45 +126,105 @@ def __init__(self, n_vertices, edges):
...
+
+        n_ccs_with_mapped_nodes = 0
+
+        for cc in nx.connected_components(sg):
+            if cc.intersection(mapped_nodes):
+                n_ccs_with_mapped_nodes += 1
+
+                # The number of connected components in the subgraph is a lower bound on the eventual number (since we could
+                # choose not to map nodes that are currently unvisited). Note that we do not revisit nodes that are visited
+                # but have not been mapped, so the number of connected components will not shrink.
```
proteneer, line 219: Clarification, does this sentence just mean:

Visiting the remaining nodes can only increase or maintain the number of connected components.

timemachine/fe/atom_mapping.py
```diff
@@ -137,13 +136,16 @@ def get_cores(
...
+    max_connected_components: int or None
+        Set to k to only keep the largest k connected
+        subgraphs in the mapping. The definition of connected
         here is different from McGregor. Here it means there
         is a way to reach the mapped atom without traversing
         over a non-mapped atom.
 
+    min_connected_component_size: int
+        Minimum size of a connected component to be considered a valid mapping
+
```
proteneer, line 148: Not a native english speaker, but is it obvious that `max_connected_components` is leq `<=` (as opposed to `<`) and `min_connected_component_size` is geq `>=` (as opposed to `>`)?

proteneer (changes requested):

timemachine/fe/mcgregor.py
mcwitt: > how much faster is _fast relative to _ref?

The difference is substantial -- from ~20-30% slower than the existing implementation to about 10% faster. (With default settings, running a selection of 200 mappings from public datasets.)

> can we add some tests just to compare the correctness of the _fast version to _ref given the increased complexity in logic?

Will do. My approach thus far had been to uncomment the lines in `mapping_is_disconnected` that compares the fast version with the reference for all pairs that we test, but agreed that we should have an automated test.

timemachine/fe/mcgregor.py
mcwitt: Note that `max_connected_components=0` is not allowed and will raise an `AssertionError`. My reasoning for `None` instead of `np.inf` to indicate no limit on the maximum number of connected components is just that there's no `np.inf` for integers, and `Optional[int]` seemed clearer to me than `int | float`.

timemachine/fe/mcgregor.py
mcwitt: Yes, that is a much more succinct summary. Will improve the comment.

timemachine/fe/atom_mapping.py
mcwitt: I would generally interpret max/min parameters as inclusive, but can clarify in the docstring

timemachine/fe/mcgregor.py
proteneer: > The difference is substantial -- from ~20-30% slower than the existing implementation to about 10% faster. (With default settings, running a selection of 200 mappings from public datasets.)

Thanks! FWIW when I looked at optimizing the code, roughly the causes for slowness in the networkx code was:

1. Instantiating a nx.Graph from the list of edges every time (But this can be cached).
2. Inducing a subgraph view is actually pretty fast
3. However, checking for existence of neighbors/edges in the view was very very slow. 

> Will do. My approach thus far had been to uncomment the lines in mapping_is_disconnected that compares the fast version with the reference for all pairs that we test, but agreed that we should have an automated test.

Yeah but given how critical this logic is I think we need to make this bullet proof.

timemachine/fe/mcgregor.py
proteneer: `np.iinfo(im.dtype).max`?

timemachine/fe/mcgregor.py
```diff
@@ -135,45 +126,105 @@ def __init__(self, n_vertices, edges):
...
+                        for w in self.get_neighbors(v):
+                            if w in mapped_nodes or w in unvisited_nodes:
+                                if w not in seen:
+                                    seen.add(w)
+                                    cc_size += 1
+                                    nextlevel.append(w)
+                n_ccs += 1
+                if cc_size < min_connected_component_size:
+                    return True
+                if n_ccs == max_connected_components:
```
proteneer, line 191: In `_ref` version there's a short circuit for `None` - does this still hold up for the `_fast` version?

timemachine/fe/mcgregor.py
mcwitt: Clarification Q: is the suggestion to remove `None` as an allowed value for the option `max_connected_components`, and instead pass a large integer when we don't want the filter to apply? (Note that there is some short-circuiting logic to skip the check entirely in the case when the filter is a no-op: https://github.com/proteneer/timemachine/blob/3cb2a8d63f8b4b817119a5872790a9c819ee8f14/timemachine/fe/mcgregor.py#L435; this might get a bit messier if we don't allow `None`)

timemachine/fe/mcgregor.py
proteneer: To clarify, in the event that `n_ccs=0` and `max_connected_components=None` this would lead to `0 == None`, which is `False` - I *think* this is as intended (i.e. bypassing the check completely).

timemachine/fe/mcgregor.py
proteneer: Ugh hm good point - I'm not sure the best way to handle this. Anyways I think it's probably fine as-is, some of these comments were done in a drive-by fashion to help me understand what's going on.

timemachine/fe/mcgregor.py
mcwitt: The short-circuit is [here](https://github.com/proteneer/timemachine/blob/3cb2a8d63f8b4b817119a5872790a9c819ee8f14/timemachine/fe/mcgregor.py#L435) and applies to either implementation in the case where `max_connected_components is None and min_connected_component_size <= 1`. (Even if `max_connected_components is None`, we may still have to look at the connected components to respect the minimum size filter.)

mcwitt: > To clarify, in the event that n_ccs=0 and max_connected_components=None this would lead to 0 == None, which is False - I think this is as intended (i.e. bypassing the check completely).

Yes, that is intentional, though admittedly confusing. I went back and forth on whether to explicitly handle `None` there; can put it back because I doubt there's any performance impact to being explicit.

timemachine/fe/mcgregor.py
mcwitt: Updated this docstring (with other fixes as well) in https://github.com/proteneer/timemachine/pull/1349/commits/cfd7cbf9039337c52308d049473a324fad90a9d0. Also renamed the method to reflect its generality

timemachine/fe/atom_mapping.py
mcwitt: Improved comments in https://github.com/proteneer/timemachine/pull/1349/commits/9f0bac8cc167008dffa7fdad2fc0e12909a7d4cc

timemachine/fe/mcgregor.py
mcwitt: Improved comments in https://github.com/proteneer/timemachine/commit/9f0bac8cc167008dffa7fdad2fc0e12909a7d4cc

timemachine/fe/mcgregor.py
mcwitt: Added a property test comparing the fast and reference implementations on random inputs in https://github.com/proteneer/timemachine/pull/1349/commits/c208820d09341e622fb0e5cbb249aa093e9e50a6. (This also led me to discover a bug in the reference version around networkx graph creation when there are single-node components.)

timemachine/fe/atom_mapping.py
mcwitt: Clarified in https://github.com/proteneer/timemachine/pull/1349/commits/804729bcfed73a8e05d032c661287232d6b75ec6

timemachine/fe/mcgregor.py
mcwitt: Opted to handle `None` explicitly in https://github.com/proteneer/timemachine/pull/1349/commits/38b29a4a8ee3011ee160c991e235a292445980ba

timemachine/fe/mcgregor.py
mcwitt: Will opt to keep as-is since it seems cleaner than the alternatives I've been able to think of.

proteneer (approved):

PR #1351, opened 2024-07-24
Add failing test case of minimization

badisa:
* This was seen originally in #1346 and had intended to get this in then, but didn't end up making it.
* Useful for testing future minimization schemes

### DIFF
--- a/tests/test_minimizer.py
+++ b/tests/test_minimizer.py
@@ -13,6 +13,8 @@ from timemachine.ff.handlers import openmm_deserializer
 from timemachine.md import builders, minimizer
 from timemachine.md.barostat.utils import compute_box_volume
 from timemachine.md.minimizer import equilibrate_host_barker, make_host_du_dx_fxn
+from timemachine.potentials import NonbondedPairList
+from timemachine.potentials.jax_utils import distance_on_pairs
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


tests/test_minimizer.py
```diff
@@ -197,3 +199,44 @@ def test_local_minimize_water_box():
...
+    """https://github.com/proteneer/timemachine/pull/1346 introduced a minimization scheme that minimizes
+    at lambda = 0.1 and at 0.0. The need for lambda = 0.1 was due to minimizing PFKFB3 with ligands
+    20 and 43, this test replicates the failure case with a small toy system to allow investigation in the future
+    """
+
+    # Atoms 2933 and 38815 from PFKFB3 with ligands 20, 43
+    nb_params = np.array(
+        [
+            [0.9323587, 0.13247664, 0.25629826, 0.0],
+            [0.52629348, 0.12916129, 0.26202344, 0.0],
```
badisa, line 215: I was incorrect about the eps being zero. Realized I had been looking at the padded values of the NonbondedAllpairs rather than the ligand-protein ixn group.

mcwitt: > I was incorrect about the eps being zero.

Unsure of the context?

tests/test_minimizer.py
```diff
@@ -197,3 +199,44 @@ def test_local_minimize_water_box():
...
+    beta = 2.0
+    cutoff = 1.2
+    pairlist = NonbondedPairList([0, 1], [1.0, 1.0], beta, cutoff)
+
+    coords = np.array(
+        [
+            [9.6643, 5.5546, 26.9625],
+            [9.65109, 5.51557, 26.93242],
+        ]
+    )
```
maxentile, line 227: coords can be simplified / made less pfkfb3-testsystem-specific

```python
dx = 0.04  # anywhere between 1e-20 and 0.04 seems to produce the same initial force norm

coords = np.array(
	[
		[0,  0,  0],
		[dx, dx, dx],
	]
)
```

maxentile (approved):
May want to simplify the top-level description.

(Toy system doesn't seem PFKFB3-specific: I suppose any pair of atoms initialized within a few hundreds of an Å of each other might do?)

mcwitt (approved):

tests/test_minimizer.py
badisa: Made less specific in https://github.com/proteneer/timemachine/pull/1351/commits/5c8a33798ec73a392ec9d1770fd2981c5881d2ae

PR #1355, opened 2024-07-26
Fixes Free Energy benchmarks

badisa:
* Using the bisection lambda schedule for all of the edges is going to overestimate the performance since the number of windows is less than or equal to n_windows, rather than strictly equal. Added some assertions to be sure that this doesn't happen again

### DIFF
--- a/tests/test_benchmark_free_energy.py
+++ b/tests/test_benchmark_free_energy.py
@@ -79,7 +79,6 @@ def setup_hif2a_single_topology_leg(host_name: str, n_windows: int, lambda_endpo
     single_topology = SingleTopology(mol_a, mol_b, core, forcefield)
     host = setup_optimized_host(single_topology, host_config) if host_config else None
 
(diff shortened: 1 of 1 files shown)


tests/test_benchmark_free_energy.py
```diff
@@ -80,19 +80,22 @@ def setup_hif2a_single_topology_leg(host_name: str, n_windows: int, lambda_endpo
     host = setup_optimized_host(single_topology, host_config) if host_config else None
 
     # Use the bisection schedule to require the least amount of minimization during bisection
```
mcwitt, line 82: Should this comment be removed?

mcwitt (approved):

tests/test_benchmark_free_energy.py
badisa: Removed in https://github.com/proteneer/timemachine/pull/1355/commits/76a6dedb9ed459c39ede0f1a5e0e6822f45a9478

PR #1356, opened 2024-07-28
Add option disallowing atom mappings with no chirally-valid dummy anchor assignment

proteneer:
## Background

1. Cancellation of dummy interaction contributions when computing $\Delta \Delta g$ relies on being able to factor dummy group interactions out of the partition function, which in turn requires that dummy groups be anchored to the core through at most one anchor atom. We implement this by disabling all core-dummy bonds except those involving a single core ("anchor") atom for each dummy group in the noninteracting state.
2. Numerical stability of MD requires that bond terms be turned on _before_ chiral volume terms (at least for the normalized version of the chiral volume that we currently use, which have a singularity when any of the 3 bonds shrinks to zero). Thus, when we disable a core-dummy bond in (1), we must also disable any chiral volume terms that span that bond.
3. We assert that every chiral center in the combined molecule be restrained (i.e. be associated with at least one chiral volume restraint term in the potential) at _both_ end states. This is to prevent kinetic trapping in an undesired chiral state, since our alchemical protocol turns on angle terms before chiral restraints.

Putting all this together, the present issue is that disabling bonds to maintain factorizability (1) necessitates disabling associated chiral volumes (2), which can result in a failure of the assertion (3).

## Approach

This PR seeks to work around the issue at the atom mapping level. Specifically, we add a new leaf filter (`enforce_chirally_valid_dummy_groups`) to the MCS search restricting the returned atom mappings to ones that have *at least one valid dummy group assignment* (i.e. not violating the chiral assertion (3)).

Separately, the end-state setup code in `SingleTopology` is updated to select a candidate dummy group assignment that does not violate (3), which is guaranteed to exist for atom mappings generated with the option `enforce_chirally_valid_dummy_groups=True`.

## Todo
- [x] handle X3 chiral centers
- [x] single topology test case for core-core bonds broken
- [x] consider renaming flag to something more accurate/less confusing (exposed `leaf_filter_fxn` instead)
- [x] randomize atom indices to strengthen tests


timemachine/fe/mcgregor.py
```diff
@@ -426,6 +465,7 @@ def recursion(
     enforce_core_core,
     max_connected_components: Optional[int],
     min_connected_component_size: int,
+    disallow_chiral_conversion: bool,
```
proteneer, line 468: Not sure if this is the right name here - since we also disallow sp3 chiral conversions at the atom-mapping level. And maybe this should be generalized into `leaf_filter_fxn`? i.e. a certificate we check at the leaf nodes?

timemachine/fe/single_topology.py
```diff
@@ -596,17 +608,31 @@ def find_dummy_groups_and_anchors(
     solution later on is to minimize the number of rotatable bonds?
     """
 
-    assignments = generate_anchored_dummy_group_assignments(mol_a, mol_b, core_atoms_a, core_atoms_b)
+    bond_graph_a = convert_to_nx(mol_a)
+    bond_graph_b = convert_to_nx(mol_b)
+
+    atom_map_a_to_b = {a: b for a, b in zip(core_atoms_a, core_atoms_b)}
+    core_disabled_bonds = compute_disabled_bonds_in_core(bond_graph_a, bond_graph_b, core_atoms_a, atom_map_a_to_b)
```
proteneer, line 615: Two bugs here:

1. I think this returns bonds with indices are indexed w.r.t. `mol_a`'s atoms. But `dga_disabled_bonds = compute_disabled_bonds_in_dga(bond_graph_b, core_atoms_b, dgs)` generates bonds whose indices are indexed w.r.t. `mol_b`. Forming a union over these two sets is inconsistent.

2. Should run the `is_chiral_conversion` check on just the core bonds (in the event of there being no dummy-atoms).

Here's a test case:

```
def test_chiral_core_bond_breaking_raises_error():
    r"""
    Test that we raise assertions for molecules that cannot generate valid
    dummy-group anchor assignments. In particular, we break two core-core
    bonds under an identity mapping (with no dummy atoms).
    """
    mol_a = Chem.MolFromMolBlock("""
  Mrv2311 07312412093D          

  6  7  0  0  0  0            999 V2000
    1.7504    0.2003    1.2663 N   0  0  2  0  0  0  0  0  0  0  0  0
    0.8845   -0.7367    0.7929 O   0  0  0  0  0  0  0  0  0  0  0  0
    0.0656    0.1573    0.2634 O   0  0  0  0  0  0  0  0  0  0  0  0
    0.6778    2.1410    1.4645 F   0  0  0  0  0  0  0  0  0  0  0  0
    0.9965    1.1853    0.5313 C   0  0  1  0  0  0  0  0  0  0  0  0
    2.3507    0.8788    0.2466 O   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0  0  0  0
  5  4  1  0  0  0  0
  5  1  1  0  0  0  0
  2  3  1  0  0  0  0
  6  1  1  0  0  0  0
  3  5  1  0  0  0  0
  5  6  1  0  0  0  0
M  END
$$$$""", removeHs=False)
    
    mol_b = Chem.MolFromMolBlock("""
  Mrv2311 07312412083D          

  6  5  0  0  0  0            999 V2000
    1.7504    0.2003    1.2663 N   0  0  2  0  0  0  0  0  0  0  0  0
    0.8845   -0.7367    0.7929 O   0  0  0  0  0  0  0  0  0  0  0  0
    0.0656    0.1573    0.2634 H   0  0  0  0  0  0  0  0  0  0  0  0
    0.6778    2.1410    1.4645 F   0  0  0  0  0  0  0  0  0  0  0  0
    0.9965    1.1853    0.5313 C   0  0  0  0  0  0  0  0  0  0  0  0
    2.3507    0.8788    0.2466 H   0  0  0  0  0  0  0  0  0  0  0  0
  1  2  1  0  0  0  0
  5  4  1  0  0  0  0
  5  1  1  0  0  0  0
  6  1  1  0  0  0  0
  2  3  1  0  0  0  0
M  END
$$$$""", removeHs=False)

    # need to generate SC charges on this mol - am1 fails
    ff = Forcefield.load_from_file('smirnoff_1_1_0_sc.py')
    core = np.array([[0,0],[1,1],[2,2],[3,3],[4,4],[5,5]])

    with pytest.raises(DummyGroupAssignmentError):
        SingleTopology(mol_a, mol_b, core, ff)
```

proteneer: note: I think the right thing to do here is to always return indices in terms of `mol_b` since later on we do:

```all_dummy_bond_idxs = recursive_map(all_dummy_bond_idxs, b_to_c)```

but please double check?

File: timemachine/fe/mcgregor.py [hunk 3]
```diff
@@ -411,6 +421,35 @@ def atom_map_pop(map_1_to_2, map_2_to_1, idx, jdx):
     map_2_to_1[jdx] = UNMAPPED
 
 
+def is_chiral_conversion(bond_graph: nx.Graph, disabled_bonds: Set[Tuple[int, int]]) -> bool:
```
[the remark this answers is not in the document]
proteneer: (actually not sure of the best terminology here ... probably fine as is)

timemachine/fe/mcgregor.py
```diff
@@ -411,6 +421,35 @@ def atom_map_pop(map_1_to_2, map_2_to_1, idx, jdx):
...
+    for node in bond_graph.nodes():
+        nbs = list(nx.neighbors(bond_graph, node))
+        if len(nbs) == 4:
```
proteneer, line 427: Thinking about this some more based on @maxentile 's comment that we also treat certain 3-connected tertiary nitro groups as non-invertible (where disabled_bonds_count > 0 would trigger the assertion). Should at least add a `TODO` here at the very minimum.

timemachine/fe/mcgregor.py
mcwitt: Was thinking the same thing -- that would also let us move the single-topology-specific code out of `mcgregor`

timemachine/fe/single_topology.py
proteneer: Here's one more test (fails on master, fails on the branch as of commit [1b70e84]) that will pass if you modify:

``` python
    atom_map_b_to_a = {b: a for b, a in zip(core_atoms_b, core_atoms_a)}
    core_disabled_bonds = compute_disabled_bonds_in_core(bond_graph_b, bond_graph_a, core_atoms_b, atom_map_b_to_a)
```

``` python
def test_chiral_bond_breaking_1_core_1_dummy_v2():

    mol_a = Chem.MolFromMolBlock(
        """lhs
                    3D
 Structure written by MMmdl.
 15 16  0  0  1  0            999 V2000
    2.1455    1.6402   -0.1409 O   0  0  0  0  0  0
   -0.0529   -0.7121    0.3721 C   0  0  0  0  0  0
   -1.4286   -0.4301   -0.2680 C   0  0  0  0  0  0
   -0.9221    0.8962   -0.8723 C   0  0  0  0  0  0
    1.7163    0.6542   -1.0827 C   0  0  0  0  0  0
    0.4519    0.5990   -0.2512 C   0  0  0  0  0  0
    1.0309    1.5756    0.6040 O   0  0  0  0  0  0
   -0.0750   -0.6889    1.4650 H   0  0  0  0  0  0
    0.4285   -1.6073   -0.0287 H   0  0  0  0  0  0
   -2.2238   -0.2880    0.4685 H   0  0  0  0  0  0
   -1.6971   -1.1659   -1.0311 H   0  0  0  0  0  0
   -1.4047    1.7800   -0.4452 H   0  0  0  0  0  0
   -0.9239    0.8937   -1.9651 H   0  0  0  0  0  0
    1.5863    1.0639   -2.0863 H   0  0  0  0  0  0
    2.3266   -0.2504   -1.0468 H   0  0  0  0  0  0
  6  2  1  0  0  0
  6  4  1  0  0  0
  6  5  1  0  0  0
  6  7  1  0  0  0
  2  3  1  0  0  0
  2  8  1  0  0  0
  2  9  1  0  0  0
  3  4  1  0  0  0
  3 10  1  0  0  0
  3 11  1  0  0  0
  4 12  1  0  0  0
  4 13  1  0  0  0
  5  1  1  0  0  0
  5 14  1  0  0  0
  5 15  1  0  0  0
  1  7  1  0  0  0
M  END""",
        removeHs=False,
    )

    mol_b = Chem.MolFromMolBlock(
        """rhs
                    3D
 Structure written by MMmdl.
 12 11  0  0  1  0            999 V2000
    0.4102    0.4907   -0.1997 O   0  0  0  0  0  0
   -0.0529   -0.7121    0.3721 C   0  0  0  0  0  0
   -1.4286   -0.4301   -0.2680 C   0  0  0  0  0  0
   -1.0708    0.5068   -0.6949 H   0  0  0  0  0  0
    1.5875    0.5421   -0.9739 C   0  0  0  0  0  0
    1.9148    1.2939   -0.2558 H   0  0  0  0  0  0
   -0.0750   -0.6889    1.4650 H   0  0  0  0  0  0
    0.4285   -1.6073   -0.0287 H   0  0  0  0  0  0
   -2.2238   -0.2880    0.4685 H   0  0  0  0  0  0
   -1.6971   -1.1659   -1.0311 H   0  0  0  0  0  0
    1.4575    0.9518   -1.9775 H   0  0  0  0  0  0
    2.1978   -0.3625   -0.9380 H   0  0  0  0  0  0
  1  2  1  0  0  0
  1  5  1  0  0  0
  2  3  1  0  0  0
  2  7  1  0  0  0
  2  8  1  0  0  0
  3  4  1  0  0  0
  3  9  1  0  0  0
  3 10  1  0  0  0
  5  6  1  0  0  0
  5 11  1  0  0  0
  5 12  1  0  0  0
M  END
""",
        removeHs=False,
    )

    # need to generate SC charges on this mol - am1 fails
    core = np.array(
        [[5, 0], [1, 1], [2, 2], [3, 3], [4, 4], [0, 5], [7, 6], [8, 7], [9, 8], [10, 9], [13, 10], [14, 11]]
    )

    from timemachine.fe.utils import plot_atom_mapping_grid

    print("core", core)
    res = plot_atom_mapping_grid(mol_a, mol_b, core)
    fpath = f"atom_mapping.svg"
    print("core mapping written to", fpath)

    with open(fpath, "w") as fh:
        fh.write(res)

    # should not raise an assertion
    SingleTopology(mol_a, mol_b, core, ff)
```

timemachine/fe/mcgregor.py
maxentile: > Thinking about this some more based on @maxentile 's comment that we also treat certain 3-connected tertiary nitro groups as chiral centers

[clarification] My comment Monday was that your notes covered the `X4` case but not the `X3` case. I don't think the current list of smarts patterns in `find_chiral_atoms` https://github.com/proteneer/timemachine/blob/3a174e63d32f4813f2d6cf4e9de4a677e6ba7419/timemachine/fe/chiral_utils.py#L146-L148 matches nitro groups.

[agreement] +1 on adding an inline `# TODO` for the `X3` case, if not covered in this PR. (I think @mcwitt listed this as a todo in the PR description yesterday. I do not have input on how to extend this logic to `len(nbs) == 3`.)

proteneer: Ah thanks for the clarification - I misremembered.

tests/test_atom_mapping.py
```diff
@@ -1383,3 +1385,24 @@ def test_hybrid_core_generation(hif2a_ligands):
     # plt.legend()
 
     # plt.show()
+
+
+def test_disallow_chiral_conversion():
+    mol_a = Chem.AddHs(Chem.MolFromSmiles("Cc1ccc(C)cc1"))  # phenyl linker
+    mol_b = Chem.AddHs(Chem.MolFromSmiles("CC(C1)(C2)CC12C"))  # bicyclopentane linker
+
+    AllChem.EmbedMolecule(mol_a, randomSeed=2024)
```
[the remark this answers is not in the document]
mcwitt: Replaced with mol blocks in https://github.com/proteneer/timemachine/pull/1356/commits/9495295f16352b7e5a715e1042a11db15984b5bf

timemachine/fe/single_topology.py
mcwitt: Added your test in https://github.com/proteneer/timemachine/pull/1356/commits/8b66a3524271c30d3e9d1a0d76e709dadc2c7171 and fixed in https://github.com/proteneer/timemachine/pull/1356/commits/2a8f88f22a660590582498bccd2727f1b7e98743

mcwitt: Added your first test in https://github.com/proteneer/timemachine/pull/1356/commits/8b66a3524271c30d3e9d1a0d76e709dadc2c7171 and fixed in https://github.com/proteneer/timemachine/pull/1356/commits/2a8f88f22a660590582498bccd2727f1b7e98743. Will now look into the second test.

mcwitt: Added second test case in https://github.com/proteneer/timemachine/pull/1356/commits/9dc0885bff6716c61a0332861c61dd525dfbcaf2. (It passes since https://github.com/proteneer/timemachine/pull/1356/commits/2a8f88f22a660590582498bccd2727f1b7e98743)

mcwitt: > note: I think the right thing to do here is to always return indices in terms of `mol_b` since later on we do:
> 
> `all_dummy_bond_idxs = recursive_map(all_dummy_bond_idxs, b_to_c)`
> 
> but please double check?

Correct, our convention is that indices in the dummy group assignments refer to atoms in mol_b. Tried to clarify in a few places in https://github.com/proteneer/timemachine/pull/1356/commits/c1e9d3ce91d5c465ae27b025fe5430486c8e4907

timemachine/fe/mcgregor.py
mcwitt: Added TODO comment in https://github.com/proteneer/timemachine/pull/1356/commits/9ec19c8f94a1feb838a4ae3765edd33eb6628382

timemachine/fe/mcgregor.py
mcwitt: Refactored the interface to expose a general `leaf_filter_fxn` argument to `mcs` in https://github.com/proteneer/timemachine/pull/1356/commits/214d8d03821dc319e624957c0c2d74bea3d35b4b

timemachine/fe/mcgregor.py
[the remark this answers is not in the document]
mcwitt: Still might want to revisit some of the naming here, e.g.  the flag `disallow_chiral_conversion`. This feels complicated enough that it's worth coming up with a good name if that can help make it any clearer.

mcwitt:
Converted back to draft after finding some unexpected behavior (possible bug) on an example.

timemachine/fe/atom_mapping.py
```diff
@@ -323,6 +323,41 @@ def core_bonds_broken_count(mol_a, mol_b, core):
...
+    if len(core_nodes) == bond_graph.number_of_nodes() and get_invalid_chiral_conversion(
+        bond_graph, core_disabled_bonds
+    ):
+        return True
+
+    # return False if there is _any_ chirally-valid dummy group assignment
+    for dga in generate_dummy_group_assignments(bond_graph, core_nodes):
+        dga_disabled_bonds = compute_disabled_bonds_in_dga(bond_graph, core_nodes, dga)
+        disabled_bonds = dga_disabled_bonds.union(core_disabled_bonds)
+        if not get_invalid_chiral_conversion(bond_graph, disabled_bonds):
```
badisa, line 355:
```suggestion
        if get_invalid_chiral_conversion(bond_graph, disabled_bonds) is None:
```

Would prefer to avoid truthy checks in this code.

timemachine/fe/dummy.py
```diff
@@ -37,10 +36,12 @@ def generate_dummy_group_assignments(
...
     Example
     -------
+    >>> from rdkit import Chem
+    >>> from timemachine.graph_utils import convert_to_nx
     >>> mol = Chem.MolFromSmiles("OC1COO1")
-    >>> g = convert_bond_list_to_nx(get_romol_bonds(mol))
+    >>> bond_graph = convert_to_nx(mol)
     >>> core = [1, 2]
-    >>> list(generate_dummy_group_assignments(g, core))
+    >>> list(generate_dummy_group_assignments(bond_graph, core))
```
badisa, line 44: This function has assertions in it that can trigger failures when calling `get_cores`. Need to be handled if we don't want to bail out of the MCS code.

Seeing failures for `assert len(list(nx.connected_components(bond_graph))) == 1`

timemachine/fe/atom_mapping.py
```diff
@@ -411,6 +447,27 @@ def planar_torsion_flip_filter(trial_core):
...
+                assert len(g1_core_disabled_bonds) == 0
+
+            if _graph_fails_chiral_assertion(g1, g1_mapped_nodes, g1_core_disabled_bonds):
+                return False
+
+            g2_mapped_nodes = {a2 for a2, a1 in enumerate(atom_map_2_to_1) if a1 != UNMAPPED}
+            g2_core_disabled_bonds = compute_disabled_bonds_in_core(g2, g1, g2_mapped_nodes, atom_map_2_to_1)
+
+            if enforce_core_core:
+                assert len(g2_core_disabled_bonds) == 0
```
maxentile, line 464: integration q: should this line raise an assertion or return a boolean?

timemachine/fe/dummy.py
```diff
@@ -148,43 +154,62 @@ def get_angle_anchors(bond_anchor):
...
+def compute_disabled_bonds_in_core(bond_graph_a, bond_graph_b, core_atoms_a, atom_map_a_to_b):
+    disabled_bonds = set()
+    for src_a, dst_a in bond_graph_a.edges():
+        if src_a in core_atoms_a and dst_a in core_atoms_a:
+            src_b, dst_b = atom_map_a_to_b[src_a], atom_map_a_to_b[dst_a]
+            if not bond_graph_b.has_edge(src_b, dst_b):
+                # print("missing bonds", src_a, dst_a)
+                disabled_bonds.add(canonicalize_bond((src_a, dst_a)))
+
+    return disabled_bonds
```
maxentile, line 196: [uncertainty] I'm not confident in reviewing this stretch -- e.g. I can't be very sure from reading whether there should also be a similar loop in the opposite direction, iterating over `(src_b, dst_b) in bond_graph_b.edges()` (and if `core_atoms_b, atom_map_b_to_a` should be added to the signature).

However, a 1-line docstring or line comment may prevent worry from arising about this stretch in future reads. (Also, looks closely related to `mcgregor.py::_verify_core_impl`, unsure if these can be consolidated in a future pass.)

tests/test_single_topology.py
```diff
@@ -1772,3 +1772,180 @@ def test_chiral_core_ring_opening():
...
+  5  4  1  0  0  0  0
+  5  1  1  0  0  0  0
+  6  1  1  0  0  0  0
+  2  3  1  0  0  0  0
+M  END
+$$$$""",
+        removeHs=False,
+    )
+
+    # need to generate SC charges on this mol - am1 fails
```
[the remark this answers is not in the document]
mcwitt: Good catch, I think this was accidentally copied from another test case.

timemachine/fe/dummy.py
mcwitt: This should only fail when the bond graph of one of the input mols is disconnected. A difference with this PR is now the failure happens at an earlier stage: during atom mapping instead of at `SingleTopology` setup. But I might argue this is a feature, not a bug?

timemachine/fe/single_topology.py
```diff
@@ -1042,15 +1172,45 @@ def verify_chiral_consistency_of_core(mol_a: Chem.Mol, mol_b: Chem.Mol, core: ND
...
+        ChiralConversionError
+            If chiral end states are incompatible for the given core and forcefield.
+        DummyGroupAssignmentError
+            If unable to find a chirally-consistent dummy group assignment
+    """
+    dummy_groups = find_chirally_consistent_dummy_groups(mol_a, mol_b, core)
+    if dummy_groups is None:
+        raise DummyGroupAssignmentError("Unable to find chirally consistent dummy group assignment")
+    dummy_groups_ab, dummy_groups_ba = dummy_groups
+    verify_chiral_consistency(mol_a, mol_b, core, dummy_groups_ab, dummy_groups_ba, forcefield)
```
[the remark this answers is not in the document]
proteneer: (edit - it's not, it calls into the inversion assertion checks)

timemachine/fe/single_topology.py
```diff
@@ -1061,6 +1221,23 @@ def verify_chiral_consistency_of_core(mol_a: Chem.Mol, mol_b: Chem.Mol, core: ND
...
+    core,
+    dummy_groups_ab: Dict[int, FrozenSet[int]],
+    dummy_groups_ba: Dict[int, FrozenSet[int]],
+    ff: Forcefield,
+) -> bool:
+    try:
+        verify_chiral_consistency(mol_a, mol_b, core, dummy_groups_ab, dummy_groups_ba, ff)
+        return True
+    # TODO: understand why AssertionError is raised
+    # (i.e. some dummy group assignments induce chiral inversions)
```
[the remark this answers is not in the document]
proteneer: Can you also confirm if for the cases where an inversion is happening, if it's happening over only core atoms (which should be forbidding by MCS) or if it's happening over a mix of core dummy atoms? I suspect the latter, esp if it's dummy group assignment dependent.

timemachine/fe/mcgregor.py
mcwitt: Since https://github.com/proteneer/timemachine/pull/1356/commits/546037d42cb12ff9af3303002f77d09c122c2ef9, we use the same code path for verifying chiral consistency in the MCS leaf filter and in the `SingleTopology` assertions (except for the former using a placeholder forcefield for speed, while the latter uses OFF 2.0). So I think this is no longer an issue.

timemachine/fe/atom_mapping.py
mcwitt: This code was removed when we transitioned to a new approach in https://github.com/proteneer/timemachine/pull/1356/commits/546037d42cb12ff9af3303002f77d09c122c2ef9. I don't think there are any remaining cases of truthy checks.

timemachine/fe/atom_mapping.py
```diff
@@ -323,6 +323,41 @@ def core_bonds_broken_count(mol_a, mol_b, core):
...
+
+
+def _graph_fails_chiral_assertion(
+    bond_graph: nx.Graph, core_nodes: Set[int], core_disabled_bonds: Set[Tuple[int, int]]
+) -> bool:
+    """A graph fails the chiral assertion if every choice of dummy group assignments fails.
+
+    A particular dummy group assignment fails if there exists a 4-connected atom with more than a single broken bond.
+    """
+    if len(core_nodes) == bond_graph.number_of_nodes() and get_invalid_chiral_conversion(
```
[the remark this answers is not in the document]
mcwitt: This code was removed when we transitioned to a new approach in https://github.com/proteneer/timemachine/pull/1356/commits/546037d42cb12ff9af3303002f77d09c122c2ef9. I don't think there is an equivalent in the new code.

timemachine/fe/dummy.py
mcwitt: This should be resolved by the new approach introduced in https://github.com/proteneer/timemachine/pull/1356/commits/546037d42cb12ff9af3303002f77d09c122c2ef9, where we use the same code path for both the MCS check and `SingleTopology` assertions (which does consider the full hybrid mol, i.e. dummy atoms for the A -> B and B -> A transformations simultaneously)

timemachine/fe/mcgregor.py
```diff
@@ -96,6 +96,7 @@ def __init__(self, n_vertices, edges):
         self.n_vertices = n_vertices
         self.n_edges = len(edges)
         self.edges = edges
+        self.nxg = nx.Graph(edges)
```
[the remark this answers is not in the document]
mcwitt: No, removed in https://github.com/proteneer/timemachine/pull/1356/commits/755bae826ee9db0a0f716c4776ee2b75e1530838

timemachine/fe/single_topology.py
```diff
@@ -568,12 +642,59 @@ def assert_improper_idxs_are_canonical(all_idxs):
...
+    bond_graph_a = convert_to_nx(mol_a)
+    bond_graph_b = convert_to_nx(mol_b)
+
+    ff = Forcefield.load_from_file("placeholder_ff.py")
+
+    return find_chirally_consistent_dummy_groups_impl(mol_a, mol_b, bond_graph_a, bond_graph_b, core, ff)
+
+
+def find_chirally_consistent_dummy_groups_impl(
+    mol_a, mol_b, bond_graph_a: nx.Graph, bond_graph_b: nx.Graph, core: NDArray, ff: Forcefield
```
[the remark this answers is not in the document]
mcwitt: This was intentional, for efficiency in the case where this is used in the MCS leaf filter:

https://github.com/proteneer/timemachine/blob/755bae826ee9db0a0f716c4776ee2b75e1530838/timemachine/fe/atom_mapping.py#L411-L426

There we precompute the bond graph, avoiding the need to call `convert_to_nx` at every leaf. This may have been a premature optimization, though.

timemachine/fe/atom_mapping.py
mcwitt: Shouldn't be relevant anymore since https://github.com/proteneer/timemachine/pull/1356/commits/546037d42cb12ff9af3303002f77d09c122c2ef9

timemachine/fe/single_topology.py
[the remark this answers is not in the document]
mcwitt: Looking more closely, the example I'd found was `test_ring_matches_ring_only`, which sets `enforce_chiral=False`: https://github.com/proteneer/timemachine/blob/3a174e63d32f4813f2d6cf4e9de4a677e6ba7419/tests/test_atom_mapping.py#L889

I see 2 options:
1. Make `disallow_chiral_conversion` imply `enforce_chiral` (i.e. raise an error if `disallow_chiral_conversion` but not `enforce_chiral`)
2. Separate the checks for chiral consistency (i.e. consistent chiral volume restraints) and validity (i.e. no volume sign changes), make `disallow_chiral_conversion` only check for consistency but not validity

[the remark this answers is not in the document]
mcwitt: Opted for (2) in https://github.com/proteneer/timemachine/pull/1356/commits/be96e1f64453d95b351c29d246f8ab4212de183e and https://github.com/proteneer/timemachine/pull/1356/commits/fad5e579496ff388682a9445dce5b7c9e26cf182, since it seemed cleanest to keep `disallow_chiral_conversion` and `enforce_chiral` independent.

timemachine/fe/mcgregor.py
```diff
@@ -468,6 +471,9 @@ def recursion(
     # leaf-node, every atom has been mapped
     if layer == n_a:
         if num_edges == threshold:
+            if not leaf_filter_fxn(atom_map_1_to_2):
```
badisa, line 474: Now that we have performance implications for the number of leaf nodes we visit, should we keep track of that in the MCSDiagonstics object?

timemachine/fe/single_topology.py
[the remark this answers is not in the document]
proteneer: sgtm! And this explanation makes sense.

timemachine/fe/dummy.py
proteneer: +1 - if bond_graph is disconnected this should probably barf.

proteneer:
lgtm

tests/test_single_topology.py
```diff
@@ -1772,3 +1772,179 @@ def test_chiral_core_ring_opening():
...
+  2  3  1  0  0  0  0
+M  END
+$$$$""",
+        removeHs=False,
+    )
+
+    ff = Forcefield.load_default()
+    core = np.array([[0, 0], [1, 1], [2, 2], [3, 3], [4, 4], [5, 5]])
+
+    mol_a, mol_b, core = permute_atom_indices(mol_a, mol_b, core, seed)
```
badisa, line 1850: What is the significance of this permutation?

tests/test_atom_mapping.py
```diff
@@ -542,6 +546,7 @@ def test_linker_map():
         disallow_planar_torsion_flips=False,
         min_threshold=0,
         initial_mapping=None,
+        enforce_chirally_valid_dummy_groups=False,  # TODO: understand why assertion fails when set to True
```
mcwitt, line 549: I think we could either set this to `False` or update the assertion (the assertion seems to assume a core that has no chirally-valid dummy group assignments). @proteneer does this seem right?

timemachine/fe/single_topology.py
```diff
@@ -63,6 +79,87 @@ def recursive_map(items, mapping):
...
+    #    c      c      c      d      d      d    |    d      c
+    #   . .    . .    / .    . .    . .    . .   |   . .    / \
+    #  d   d  d   d  c   d  c   d  c   c  d   d  |  c   c  c   c (only core)
+    dgc = dummy_group + list(core_atoms)
+    for idxs, params in zip(chiral_atom_idxs, chiral_atom_params):
+        center, i, j, k = idxs
+        if all([a in dgc for a in idxs]):
+            # non center dummy atom count
+            ncda_count = sum([a in dummy_group for a in (i, j, k)])
+            if ncda_count == 1 or ncda_count == 2 or ncda_count == 3:
```
badisa, line 153: Any reason not to do the following? Looks like value can only be [0, 3]
```suggestion
            if ncda_count > 0:
```

timemachine/fe/single_topology.py
```diff
@@ -1018,27 +1171,83 @@ def assert_chiral_consistency_and_validity(
...
+    dst_chiral_restr_idx_set = ChiralRestrIdxSet(dst_chiral_idxs)
 
-def verify_chiral_consistency_of_core(mol_a: Chem.Mol, mol_b: Chem.Mol, core: NDArray, forcefield):
+    assert_chiral_consistency(src_chiral_restr_idx_set, dst_chiral_restr_idx_set, src_bond_idxs, dst_bond_idxs)
+    check_chiral_validity_impl(
+        atom_map, src_chiral_restr_idx_set, dst_chiral_restr_idx_set, src_bond_idxs, dst_bond_idxs
+    )
+
+
+def verify_chiral_validity_of_core(mol_a: Chem.Mol, mol_b: Chem.Mol, core: NDArray, forcefield):
```
mcwitt, line 1221: Note breaking rename of this function (cc @badisa). This was done to standardize the definitions of "chiral consistency" and "chiral validity".

timemachine/fe/mcgregor.py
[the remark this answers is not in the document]
mcwitt: Renamed `disallow_chiral_conversion` to `enforce_chirally_valid_dummy_groups` in https://github.com/proteneer/timemachine/pull/1356/commits/e6b29ebf892b06c9d8440d1c18d5bf289322f5a7, since this seemed more accurate and less confusing.

badisa (approved):
Nice, this is going to be a real improvement.

badisa:
Would you update the description since it no longer reflects the approach initially taken?

tests/test_atom_mapping.py
proteneer: This is a test case devoid of any hydrogens, and `chiral_utils.find_chiral_atoms` fails for molecules that have implicit hydrogens. There's no sensical way to interpret chirality for implicit hydrogens.

timemachine/fe/single_topology.py
mcwitt: Not sure; this code was lifted from #1358 (and even there might have just been copy-pasted). Seems reasonable to me, but I think @proteneer was the original author (maybe there's some reason to be explicit?)

tests/test_single_topology.py
mcwitt: This just renumbers atom indices in mol_a and mol_b (and updates the core consistently) to strengthen the test (i.e. make sure the logic tested is independent of atom numbering)

tests/test_atom_mapping.py
mcwitt: Added a note (and also set `enforce_chiral=False`) in this example in https://github.com/proteneer/timemachine/pull/1356/commits/8ef161c71506397358a7d3a34bc60c69f04d7c2c

timemachine/fe/single_topology.py
```diff
@@ -568,12 +654,73 @@ def assert_improper_idxs_are_canonical(all_idxs):
...
+        # Suppress warnings from end-state setup during the search; these are only relevant for the selected candidate,
+        # and will be raised again when we construct the final SingleTopology instance
+        warnings.simplefilter("ignore", ChiralVolumeDisabledWarning)
+        warnings.simplefilter("ignore", MultipleAnchorWarning)
+
+        pairs = (
+            (dummy_groups_ab, dummy_groups_ba)
+            # NOTE: We use fair_product_2 instead of itertools.product to avoid iterating over all possible candidates
+            # for A -> B before trying the second candidate for B -> A
+            for dummy_groups_ab, dummy_groups_ba in fair_product_2(
```
maxentile, line 700: [nit/tangent/curiosity] is the expected performance with `fair_product_2` ordering >= randomized ordering > `product` ordering?

(I don't quickly see the motivation for `fair_product_2` over a random permutation, but I see the motivation over the default ordering from `product`)

maxentile (approved):
Looks great! Resolves a significant problem.

(My only line comments in this pass are tangential)

timemachine/utils.py
```diff
@@ -32,3 +32,30 @@ def pairwise_transform_and_combine(xs: Iterable[A], f: Callable[[A], B], g: Call
             yield g(prev_b, b)
 
         prev_b = b
+
+
+def fair_product_2(xs: Sequence[A], ys: Sequence[B]) -> Iterable[Tuple[A, B]]:
+    """Like py:func:`itertools.product`, except iterates over arguments in an unbiased ordering.
```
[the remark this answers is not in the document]
mcwitt: Just because it takes two arguments -- could imagine a general variadic version (like `itertools.product`) that takes an arbitrary number of arguments and is implemented in terms of the 2-argument version, but this wasn't needed here.

timemachine/fe/single_topology.py
mcwitt: At this stage I don't think there would be any benefit to using the fair product over a random permutation of the product (since `generate_dummy_group_assignments` returns an arbitrary ordering), but in principle the inputs could be ordered by some measure of optimality.

timemachine/fe/mcgregor.py
mcwitt: Added `MCSDiagnostics.total_leaves_visited` in https://github.com/proteneer/timemachine/pull/1356/commits/597181663262f5780b10b06bf5b5f76c66805db2

timemachine/utils.py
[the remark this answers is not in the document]
maxentile: Ahh, thanks! Makes sense

(had thought it might have been vestigial, if there were multiple previous versions)

timemachine/fe/single_topology.py
maxentile: Makes sense -- written in a modular way, could swap out different functions in place of `fair_product_2` if beneficial in the future

jkausrelay (approved):

timemachine/fe/single_topology.py
mcwitt: Also note that this can additionally raise a `DummyGroupAssignmentError`, if there were no cores found that are compatible with chirally-valid dummy group assignments, using the placeholder forcefield.

It can still raise `ChiralConversionError`, but this should not happen typically (only if there's a case where the placeholder forcefield and MD forcefield yield different chiral atoms).

PR #1357, opened 2024-07-31
Adds max_nodes_visited parameter

badisa:
* The default value is set incredibly high, to preserve the old behavior
* Raise NoMappingError exception rather than assertion to make the errors raised more uniform

### DIFF
--- a/timemachine/fe/atom_mapping.py
+++ b/timemachine/fe/atom_mapping.py
@@ -53,6 +53,7 @@ def get_cores_and_diagnostics(
     ring_cutoff,
     chain_cutoff,
     max_visits,
+    max_node_visits: int,
     max_connected_components: Optional[int],
     min_connected_component_size: int,
     max_cores,
@@ -68,6 +69,7 @@ def get_cores_and_diagnostics(
 
     get_cores_ = partial(
         _get_cores_impl,
+        max_node_visits=max_node_visits,
         ring_cutoff=ring_cutoff,
         chain_cutoff=chain_cutoff,
         max_visits=max_visits,
@@ -98,6 +100,7 @@ def get_cores(
     ring_cutoff,
     chain_cutoff,
     max_visits,
+    max_node_visits: int,
     max_connected_components: Optional[int],
     min_connected_component_size: int,
     max_cores,
@@ -134,7 +137,10 @@ def get_cores(
         The distance cutoff that non-ring atoms must satisfy.
 
     max_visits: int
-        Maximum number of nodes we can visit for a given threshold.
+        Maximum number of nodes we can visit for a given core size.
+
+    max_node_visits: int
+        Maximum number of nodes that can be visited overall to find a core.
 
     max_connected_components: int or None
         Set to k to only keep mappings where the number of connected components is <= k.
@@ -179,6 +185,7 @@ def get_cores(
         ring_cutoff,
         chain_cutoff,
         max_visits,
+        max_node_visits,
         max_connected_components,
         min_connected_component_size,
         max_cores,
@@ -329,6 +336,7 @@ def _get_cores_impl(
     ring_cutoff,
     chain_cutoff,
     max_visits,
+    max_node_visits: int,
     max_connected_components: Optional[int],
     min_connected_component_size: int,
     max_cores,
@@ -420,6 +428,7 @@ def _get_cores_impl(
         max_visits,
         max_cores,
         enforce_core_core,
+        max_node_visits,
         max_connected_components,
         min_connected_component_size,
         min_threshold,
--- a/timemachine/fe/mcgregor.py
+++ b/timemachine/fe/mcgregor.py
@@ -1,6 +1,5 @@
 # maximum common subgraph routines based off of the mcgregor paper
 import copy
-import time
 import warnings
 from dataclasses import dataclass
 from typing import Callable, List, Optional, Sequence, Set, Tuple
@@ -300,6 +299,7 @@ def mcs(
     max_visits,
     max_cores,
     enforce_core_core,
+    max_node_visits: int,
     max_connected_components: Optional[int],
     min_connected_component_size: int,
     min_threshold,
@@ -325,7 +325,8 @@ def mcs(
     base_layer = len(initial_mapping)
     priority_idxs = tuple(tuple(x) for x in priority_idxs)
     # Keep start time for debugging purposes below
-    start_time = time.time()  # noqa
+    # import time
+    # start_time = time.time()  # noqa
 
     mcs_result = None
 
@@ -360,10 +361,15 @@ def mcs(
 
         total_nodes_visited += mcs_result.nodes_visited
 
+        if total_nodes_visited >= max_node_visits:
+            raise NoMappingError(
+                f"Exceeded the max number of nodes ({max_node_visits}) visited, visited {total_nodes_visited}"
+            )
+
         # If timed out, either due to max_visits or max_cores, raise exception.
         if mcs_result.timed_out:
             warnings.warn(
-                f"Reached max number of visits/cores: {len(mcs_result.all_maps)} cores with {mcs_result.nodes_visited} nodes visited. "
+                f"Reached max number of visits/cores per threshold: {len(mcs_result.all_maps)} cores with {mcs_result.nodes_visited} nodes visited. "
                 "Cores may be suboptimal.",
                 MaxVisitsWarning,
             )
@@ -379,7 +385,8 @@ def mcs(
         # f"==FAILED==[NODES VISITED {mcs_result.nodes_visited} | time taken: {time.time()-start_time} | time out? {mcs_result.timed_out}]====="
         # )
 
-    assert mcs_result is not None
+    if mcs_result is None:
+        raise NoMappingError("No thresholds evaluated when generating cores")
 
     if len(mcs_result.all_maps) == 0:
         raise NoMappingError("Unable to find mapping")
--- a/tests/test_atom_mapping.py
+++ b/tests/test_atom_mapping.py
@@ -353,6 +353,7 @@ $$$$""",
         ring_cutoff=0.1,
         chain_cutoff=0.2,
         max_visits=1e7,
+        max_node_visits=1e10,
         max_connected_components=1,
         min_connected_component_size=1,
         max_cores=1e6,  # This pair has 350k cores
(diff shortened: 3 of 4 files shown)


timemachine/fe/mcgregor.py
```diff
@@ -360,10 +361,15 @@ def mcs(
 
         total_nodes_visited += mcs_result.nodes_visited
 
+        if total_nodes_visited >= max_node_visits:
```
proteneer, line 364: I think you need to do this check *within* the recursion() call to be safe - the number of nodes visited is usually exponential w.r.t. layer - so even going up one layer may result in something like 100x more nodes visited.

proteneer: (i.e. the last iteration of the inner recursion call may visit 10x the `max_node_visits`)

badisa: Reduces the visits if the per threshold quantity exceeds the total amount in https://github.com/proteneer/timemachine/pull/1357/commits/752469e52eeab9fae783911109df39c3afee70a2

proteneer:
Does it make sense to consolidate `max_visits` and `max_node_visits` into just `max_node_visits`? I don't see much of a use for a per-layer `max_visits` at this point.

badisa:
> Does it make sense to consolidate `max_visits` and `max_node_visits` into just `max_node_visits`? I don't see much of a use for a per-layer `max_visits` at this point.

In https://github.com/proteneer/timemachine/pull/1067 we introduced the warning. The intention was to return a core if we had hit the limit. I would be alright with making it so that the warning is thrown if `len(mcs_result.all_maps) > 0` and an exception otherwise. Which would avoid the additional parameter.

badisa:
@proteneer I changed the logic in https://github.com/proteneer/timemachine/pull/1357/commits/49157dc669822106daf8826ecefe779406b7ef82 to raise an exception only if no cores were found, otherwise have a warning that the core may suboptimal

timemachine/fe/mcgregor.py
```diff
@@ -358,28 +360,33 @@ def mcs(
...
+                    MaxVisitsWarning,
+                )
             # don't remove this comment and the one below, useful for debugging!
             # print(
             # f"==SUCCESS==[NODES VISITED {mcs_result.nodes_visited} | CORE_SIZE {len([x != UNMAPPED for x in mcs_result.all_maps[0]])} | NUM_CORES {len(mcs_result.all_maps)} | NUM_EDGES {mcs_result.num_edges} | time taken: {time.time()-start_time} | time out? {mcs_result.timed_out}]====="
             # )
             break
+        elif mcs_result.timed_out:
+            # If timed out, either due to max_visits or max_cores, raise exception.
+            raise NoMappingError(f"Exceeded max number of visits/cores: {mcs_result.nodes_visited} nodes visited.")
```
proteneer, line 382: I'd write this as `Exceeded max number of visits - no valid cores could be found: {mcs_result.nodes_visited} nodes visited.`

timemachine/fe/mcgregor.py
```diff
@@ -358,28 +360,33 @@ def mcs(
...
-                MaxVisitsWarning,
-            )
+        total_nodes_visited += mcs_result.nodes_visited
 
         if len(mcs_result.all_maps) > 0:
+            # If we timed out but got cores, throw a warning
+            if mcs_result.timed_out:
+                warnings.warn(
+                    f"Reached max number of visits/cores: {len(mcs_result.all_maps)} cores with {mcs_result.nodes_visited} nodes visited. "
+                    "Cores may be suboptimal.",
```
proteneer, line 372: Suboptimal implies something about the quality of the core that I don't think we can imply at this stage.

If you have at least one core that's been found, but you timed out, the core(s) you did find are probably still comparable to the hidden, remaining cores that you didn't find. Maybe to be more precise:

``` python
if mcs_result.timed_out and len(mcs_result.all_maps) < max_cores:
    warnings.warn(f"Inexhaustive search: reached max number of visits ({mcs_result.nodes_visited}) and found only {len(mcs_result.all_maps)} out of {max_cores} desired cores."
```

badisa: Adjusted in https://github.com/proteneer/timemachine/pull/1357/commits/5ae26fa556061bcf280716949f4e740783228b05 and https://github.com/proteneer/timemachine/pull/1357/commits/f6ec329025918bce2a86b5f8ae3a3c6c0fc20bac to be more precise

timemachine/fe/atom_mapping.py
```diff
@@ -115,7 +115,8 @@ def get_cores(
 
     Additional notes
     ----------------
-    1) The returned cores are sorted in increasing order based on the rmsd of the alignment.
+    1) The returned cores are jointly sorted in increasing order based on the number of core-dummy bonds broken, the
+       number of valence values changed and the rmsd of the alignment.
     2) The number of cores atoms may vary slightly, but the number of mapped edges are the same.
```
badisa, line 120: I had forgotten about this aspect of the atom mapping code, do we know how many atoms it differs by and if we should perhaps consider this in the sort (presumably fewer atoms will be selected based on the RMSD)?

badisa: Not sure this is true, I added in an assertion here that cores are all of the same size and haven't been able to trigger it. @proteneer can you comment on this?

timemachine/fe/atom_mapping.py
```diff
@@ -115,7 +115,8 @@ def get_cores(
 
     Additional notes
     ----------------
-    1) The returned cores are sorted in increasing order based on the rmsd of the alignment.
+    1) The returned cores are jointly sorted in increasing order based on the number of core-dummy bonds broken, the
+       number of valence values changed and the rmsd of the alignment.
```
maxentile, line 119: possible imprecision: "number of valence values changed " sounds like it's the number of core atoms where valence in mol_a differs from valence in mol_b `sum_i (valence_in_b(i) != valence_in_a(i))`, but instead it's something like `sum_i |valence_in_b(i) - valence_in_a(i)|`, not sure how to summarize

```suggestion
    1) The returned cores are jointly sorted in increasing order based on the number of core-dummy bonds broken,
       the sum of valence values changed, and the rmsd of the alignment.
```

timemachine/fe/mcgregor.py
```diff
@@ -358,28 +360,35 @@ def mcs(
...
+                f"Exceeded max number of visits/cores - no valid cores could be found: {mcs_result.nodes_visited} nodes visited."
+            )
         # else:
         # print(
         # f"==FAILED==[NODES VISITED {mcs_result.nodes_visited} | time taken: {time.time()-start_time} | time out? {mcs_result.timed_out}]====="
         # )
 
-    assert mcs_result is not None
+    if mcs_result is None:
+        raise NoMappingError("No thresholds evaluated when generating cores")
```
maxentile, line 391: [style] I don't know what "no thresholds evaluated" means -- "no layers completed", "timed out before completing first stage", or similar ?

maxentile (approved):

timemachine/fe/mcgregor.py
badisa: The term 'layers' in the code here means the number of layers of recursion. 

Actually means that there is no mapping, will change this to make it clearer and add a test of this logic

badisa: Fixed up in https://github.com/proteneer/timemachine/pull/1357/commits/fbe77e6b3fd5a81e9d2f9d16bba75d7acf7ee02c to be more clear that this happens when the predicate matrix is empty.

timemachine/fe/mcgregor.py
```diff
@@ -358,29 +361,33 @@ def mcs(
             filter_fxn,
         )
 
-        total_nodes_visited += mcs_result.nodes_visited
+        visits_left -= mcs_result.nodes_visited
```
mcwitt, line 364: Do we need both `total_nodes_visited` and `visits_left`? (e.g. could `visits_left` be replaced by `max_visits - total_nodes_visited`?)

timemachine/fe/mcgregor.py
```diff
@@ -325,13 +324,17 @@ def mcs(
...
     # Keep start time for debugging purposes below
-    start_time = time.time()  # noqa
-
-    mcs_result = None
+    # import time
+    # start_time = time.time()  # noqa
 
     # run in reverse by guessing max # of edges to avoid getting stuck in minima.
     max_threshold = _arcs_left(base_marcs)
+    if max_threshold < 1:
```
mcwitt, line 332: nit: not for this PR but I would prefer more descriptive names for `threshold` and `max_threshold`. Are they actually numbers of edges?

mcwitt:
> * The default value is set incredibly high, to preserve the old behavior
> * Raise NoMappingError exception rather than assertion to make the errors raised more uniform

I think this also changes the meaning of `max_visits` to accumulate over outer loop iterations (?) Should we mention that in the description?

mcwitt (approved):

timemachine/fe/mcgregor.py
badisa: Wasn't needed, was just supposed to be easier to read. I replaced it with `max_visits - total_nodes_visited` and a comment in https://github.com/proteneer/timemachine/pull/1357/commits/994c5964dd7af36501f819394ffa5f40b5ed374b

timemachine/fe/mcgregor.py
badisa: I agree. Want to leave it until another pass as I don't understand the code well enough to justify changing the naming

badisa:
> > * The default value is set incredibly high, to preserve the old behavior
> > * Raise NoMappingError exception rather than assertion to make the errors raised more uniform
> 
> I think this also changes the meaning of `max_visits` to accumulate over outer loop iterations (?) Should we mention that in the description and/or PR title?

Updated description (https://github.com/proteneer/timemachine/pull/1357#issue-2440795079) to indicate that it is the max visits across all thresholds (again think this edges, but want to get a better understanding of code before committing) and that this allows more deterministic wall clock times.

timemachine/fe/mcgregor.py
proteneer: max_threshold is the *maximum* possible number of common edges. Given a `NxM` mapping of edges in `mol_a` to edges in `mol_b`:
```
def _arcs_left(marcs):
    num_row_edges = np.sum(np.any(marcs, 1))
    num_col_edges = np.sum(np.any(marcs, 0))
    return min(num_row_edges, num_col_edges)
```

timemachine/fe/mcgregor.py
```diff
@@ -349,7 +350,8 @@ def mcs(
             base_marcs,
             mcs_result,
             priority_idxs,
-            max_visits,
+            # Decrement the max visits by the number already visited to ensure constant wall clock time
+            max_visits - total_nodes_visited,
```
proteneer, line 354: nice trick

proteneer (approved):

PR #1360, opened 2024-08-08
Refactor chiral restraint setup to avoid repeated work

mcwitt:
The chiral validity MCS filter added in #1356 is prohibitively slow in some important edge cases where the MCS search visits a large number of chirally-invalid leaf nodes before finding a valid one, because each invocation of the leaf filter is expensive.

One opportunity for improvement is to remove some of the substantial redundant work that is done when setting up chiral restraints for the hybrid mol in the leaf filter. Much of the expensive work is independent of the dummy group assignment, and can be precomputed outside of the MCS search.

This PR refactors bond and chiral restraint setup to reduce the time for a single execution of the chiral validity filter from 200-300 ms down to ~10 ms.

## Summary of changes
- Precompute quantities independent of core and dummy group assignment during MCS initialization
- Move assertion that hybrid mol is connected to `setup_end_state`
- Refactor `setup_dummy_bond_and_chiral_interactions` using vectorized numpy ops
- Remove `recursive_map`, replace with array indexing
- Vectorize bond and chiral atom ixn canonicalization
- Move sanity-check assertions behind `verify` flag (defaulting to True; disabled during MCS)

### DIFF
--- a/timemachine/fe/single_topology.py
+++ b/timemachine/fe/single_topology.py
@@ -407,13 +407,13 @@ def make_setup_end_state_harmonic_bond_and_chiral_potentials(
 
         where
 
-        core: list of 2-tuples
+        core: array of shape (core_size, 2)
             Each pair is an atom mapping from mol_a into mol_b
 
-        a_to_c: dict or array, supports []
+        a_to_c: array
             mapping from a into a common core idx
 
-        b_to_c: dict or array, supports []
+        b_to_c: array
             mapping from b into a common core idx
 
         dummy_groups: Dict[int, FrozenSet[int]]
@@ -552,13 +552,13 @@ def setup_end_state(ff, mol_a, mol_b, core, a_to_c, b_to_c, dummy_groups: Dict[i
     mol_b: Chem.Mol
         Molecule providing the dummy atoms.
 
-    core: list of 2-tuples
+    core: array of shape (core_size, 2)
         Each pair is an atom mapping from mol_a into mol_b
 
-    a_to_c: dict or array, supports []
+    a_to_c: array
         mapping from a into a common core idx
 
-    b_to_c: dict or array, supports []
+    b_to_c: array
         mapping from b into a common core idx
 
     dummy_groups: Dict[int, FrozenSet[int]]
--- a/timemachine/fe/single_topology.py
+++ b/timemachine/fe/single_topology.py
@@ -1,4 +1,5 @@
 import warnings
+from collections import defaultdict
 from collections.abc import Iterable
 from enum import IntEnum
 from functools import partial
@@ -1126,21 +1127,19 @@ class AtomMapMixin:
 _Bonded = TypeVar("_Bonded", bound=Union[ChiralAtomRestraint, HarmonicAngleStable, HarmonicBond, PeriodicTorsion])
 
 
-def get_neighbors(atom, bond_idxs) -> List[int]:
-    nbs = []
+def get_neighbors(bond_idxs: Collection[Tuple[int, int]]) -> Dict[int, List[int]]:
+    neighbors = defaultdict(list)
     for i, j in bond_idxs:
-        if i == atom:
-            nbs.append(j)
-        elif j == atom:
-            nbs.append(i)
-    return nbs
+        neighbors[i].append(j)
+        neighbors[j].append(i)
+    return neighbors
 
 
 def check_chiral_validity_src_dst(src_chiral_centers_in_mol_c, dst_chiral_restr_idx_set, src_bond_idxs):
     """Raise error unless, for every chiral center, at least 1 chiral volume is defined in both end-states."""
-
+    neighbors = get_neighbors(src_bond_idxs)
     for c in src_chiral_centers_in_mol_c:
-        nbs = get_neighbors(c, src_bond_idxs)
+        nbs = neighbors[c]
         if len(nbs) == 4:
             i, j, k, l = nbs
             # (ytz): the ordering of i,j,k,l is random if we're reading directly from the mol graph,
--- a/timemachine/fe/single_topology.py
+++ b/timemachine/fe/single_topology.py
@@ -359,7 +359,9 @@ def get_num_connected_components(num_atoms: int, bonds: Collection[Tuple[int, in
     return len(list(nx.connected_components(g)))
 
 
-def canonicalize_chiral_atom_idxs(idxs):
+def canonicalize_chiral_atom_idxs(idxs: NDArray[np.int32]) -> NDArray[np.int32]:
+    assert idxs.ndim == 2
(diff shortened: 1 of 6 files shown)


timemachine/fe/single_topology.py
```diff
@@ -382,141 +359,180 @@ def get_num_connected_components(num_atoms: int, bonds: Collection[Tuple[int, in
...
+
+        # chiral atoms need special code for canonicalization, since triple product is invariant
+        # under rotational symmetry (but not something like swap symmetry)
+        mol_c_chiral_atom_idxs = canonicalize_chiral_atom_idxs(mol_c_chiral_atom_idxs)
+
+        mol_c_chiral_bond_idxs = canonicalize_bonds(mol_a_chiral_bond_idxs)
+        mol_c_chiral_bond_signs = mol_a_chiral_bond.potential.signs
+
+        chiral_atom_potential = ChiralAtomRestraint(mol_c_chiral_atom_idxs).bind(mol_c_chiral_atom_params)
+        chiral_bond_potential = ChiralBondRestraint(mol_c_chiral_bond_idxs, mol_c_chiral_bond_signs).bind(
```
mcwitt, line 529: Note: did not end up separating chiral bond restraint setup because the performance impact was negligible compared to the chiral atom restraints

timemachine/fe/single_topology.py
```diff
@@ -382,141 +359,180 @@ def get_num_connected_components(num_atoms: int, bonds: Collection[Tuple[int, in
...
+                mol_b_bond_params,
+                mol_b_chiral_atom.potential.idxs,
+                np.asarray(mol_b_chiral_atom.params),
+                dg,
+                anchor,
+                core[:, 1],
+                verify,
+            )
+            # append idxs
+            all_dummy_bond_idxs_.extend(all_idxs[0])
```
badisa, line 458: nit: In some places you have added `np.concatenate` does replacing these extends with `np.concatenate` not improve performance?

timemachine/fe/single_topology.py
```diff
@@ -1230,63 +1254,59 @@ def verify_chiral_validity_of_core(mol_a: Chem.Mol, mol_b: Chem.Mol, core: NDArr
...
     )
 
-    src_chiral_restr_idx_set = ChiralRestrIdxSet(chiral_atom_pot_src.potential.idxs)
-    dst_chiral_restr_idx_set = ChiralRestrIdxSet(chiral_atom_pot_dst.potential.idxs)
+    def check_chiral_validity(
+        core: NDArray,
+        dummy_groups_ab: Dict[int, FrozenSet[int]],
+        dummy_groups_ba: Dict[int, FrozenSet[int]],
+    ):
+        """Verify that a core and forcefield would allow for valid chiral endstates.
```
badisa, line 1281:
```suggestion
        """Verify that a core and dummy groups would allow for valid chiral endstates.
```

badisa (approved):
Not a high confidence review, but seems correct and the performance improvement is significant.

timemachine/fe/single_topology.py
```diff
@@ -382,141 +359,180 @@ def get_num_connected_components(num_atoms: int, bonds: Collection[Tuple[int, in
...
+    ff: forcefield.Forcefield
+        Forcefield used to parameterize the molecule
 
-    b_to_c: dict or array, supports []
-        mapping from b into a common core idx
-
-    dummy_groups: Dict[int, FrozenSet[int]]
-        mapping from anchor atom to dummy group. Indices refer to atoms in mol_b.
+    verify: bool, optional
+        If True, perform additional consistency checks. Set to False when maximum performance is desired.
```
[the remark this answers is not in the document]
mcwitt: Yes, setting `verify=False` skips assertions that could in theory fail (but these mostly seemed to be sanity-check assertions intended to catch bugs in the code as opposed to unexpected inputs).

Note that `verify=False` is only used during the MCS filter and dummy group selection, and defaults to `True` for the final end-state setup. So we would still catch any assertion failures at that time.

timemachine/fe/single_topology.py
```diff
@@ -325,6 +295,13 @@ def setup_dummy_interactions(
     return bonded_idxs, bonded_params
 
 
+def canonicalize_bonds(bonds: NDArray[np.int32]) -> NDArray[np.int32]:
+    assert bonds.ndim == 2
```
proteneer, line 299: nice - good idea to batch this! (would be useful in other parts of the code later on for angles, torsions etc) too. (fine as-is for this PR).

timemachine/fe/single_topology.py
```diff
@@ -382,141 +359,180 @@ def get_num_connected_components(num_atoms: int, bonds: Collection[Tuple[int, in
...
-        The potentials involved in checking the correctness and validity of chirality at endstates.
+    Callable
+        (core, a_to_c, b_to_c, dummy_groups) -> (HarmonicBond, ChiralAtomRestraint, ChiralBondRestraint)
+
+        where
+
+        core: list of 2-tuples
+            Each pair is an atom mapping from mol_a into mol_b
 
+        a_to_c: dict or array, supports []
```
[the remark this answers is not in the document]
mcwitt: This was just copied and out of date, fixed in https://github.com/proteneer/timemachine/pull/1360/commits/27d632812aeb1620357499a369ffd0e000e14994

proteneer (approved):

timemachine/fe/single_topology.py
mcwitt: Good call; I wouldn't think concatenating arrays on the fly would be efficient because it involves unnecessary copying, but I think that appending instead of extending in the loop and calling `np.concatenate` at the end should be faster (and does appear to shave a few more percent). https://github.com/proteneer/timemachine/pull/1360/commits/e3596cd43f4a2f2670467482aa61ce5e9c371a00

timemachine/fe/single_topology.py
```diff
@@ -63,75 +63,45 @@ class DummyGroupAssignmentError(RuntimeError):
...
-    return setup_dummy_bond_and_chiral_interactions(
-        hb.idxs,
-        bond_params,
-        chiral_atom_idxs,
-        chiral_atom_params,
-        dummy_group,
-        root_anchor_atom,
-        core_atoms,
-    )
+    b0 = bonds[:, :, None] == idxs[None, None, :]  # (bonds, 2 | 3 | 4, idxs)
```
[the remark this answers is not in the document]
mcwitt: Hopefully addressed by https://github.com/proteneer/timemachine/pull/1360/commits/dbf84e98a5cdc8e3f9e56a454afa10e015d2645a and https://github.com/proteneer/timemachine/pull/1360/commits/a903eb2c91c9e09a93ae1fbec8fade3368dacd15.

maxentile (approved):

PR #1362, opened 2024-08-15
Clean up atom mapping code a bit

badisa:
* Test is no longer expected to take 30 minutes
* Don't do RMSD, since the square root is unnecessary
* No longer compute the bonds after atom mapping
* Uniquify without returning incorrect bonds
* Seems to improve wall clock time of tests by about 5%

### DIFF
--- a/tests/test_atom_mapping.py
+++ b/tests/test_atom_mapping.py
@@ -11,7 +11,7 @@ from timemachine.constants import DEFAULT_ATOM_MAPPING_KWARGS
 from timemachine.fe import atom_mapping
 from timemachine.fe.mcgregor import MaxVisitsWarning, NoMappingError
 from timemachine.fe.single_topology import DummyGroupAssignmentError, verify_chiral_validity_of_core
-from timemachine.fe.utils import get_romol_conf, plot_atom_mapping_grid, read_sdf, set_romol_conf
+from timemachine.fe.utils import get_mol_name, get_romol_conf, plot_atom_mapping_grid, read_sdf, set_romol_conf
 from timemachine.ff import Forcefield
 
 pytestmark = [pytest.mark.nocuda]
@@ -30,13 +30,12 @@ def hif2a_ligands():
     return read_sdf(hif2a_set)
 
 
-@pytest.mark.nightly(reason="Slow")
 def test_connected_core_with_large_numbers_of_cores():
     """The following tests that for two mols that have a large number of matching
     cores found prior to filtering out the molecules with disconnected cores and incomplete rings.
 
     Previously this pair could take about 45 minutes to generate a mapping, largely due to removing
-    imcomplete ring cores."""
+    incomplete ring cores."""
 
     mol_a = Chem.MolFromMolBlock(
         """CHEMBL3645089
@@ -357,7 +356,7 @@ $$$$""",
         max_visits=1e7,
         max_connected_components=1,
         min_connected_component_size=1,
-        max_cores=1e6,  # This pair has 350k cores
+        max_cores=1e6,
         enforce_core_core=True,
         ring_matches_ring_only=False,
         enforce_chiral=True,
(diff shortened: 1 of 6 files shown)


tests/test_atom_mapping.py
```diff
@@ -1032,6 +1025,7 @@ def go(k):
 def make_polyphenylene(n, dihedral_deg):
     """Make a chain of n benzene rings with each ring rotated `dihedral_deg` degrees with respect to the previous ring"""
     mol = Chem.MolFromSmiles(polyphenylene_smiles(n))
+    mol.SetProp("_Name", f"{n}_{dihedral_deg}")
```
badisa, line 1028: Added a name so that I could plot these. If you change the version of RDKit the tests that use these mols fail due to a change the output of `AllChem.EmbedMolecule`

tests/test_atom_mapping.py
```diff
@@ -368,24 +367,18 @@ def test_connected_core_with_large_numbers_of_cores():
...
 
-    # If this takes more then 30 minutes, something is off.
-    assert time.time() - start < 1800.0
-
-
-def get_mol_name(mol) -> str:
-    """Return the title for the given mol"""
-    return mol.GetProp("_Name")
+    # If this takes more then a minute, something is off.
+    assert time.time() - start < 60.0
```
badisa, line 371: This test now runs in a couple of seconds rather than 30 minutes. Reduced the threshold to ensure we don't go back to a really slow implementation.

timemachine/fe/atom_mapping.py
```diff
@@ -437,9 +417,8 @@ def leaf_filter_fxn(trial_core) -> bool:
         make_leaf_filter_fxn(),
     )
 
-    all_bond_cores = [_compute_bond_cores(mol_a, mol_b, marcs) for marcs in all_marcs]
     all_cores = remove_cores_smaller_than_largest(all_cores)
-    all_cores, _ = _deduplicate_all_cores_and_bonds(all_cores, all_bond_cores)
```
badisa, line 442: While this returned the right cores, the bonds could be meaningless because we had removed all of the smaller cores.

Changed to only de-duplicate on cores since we don't use the bonds for post-processing any longer.

timemachine/fe/atom_mapping.py
```diff
@@ -437,9 +417,8 @@ def leaf_filter_fxn(trial_core) -> bool:
         make_leaf_filter_fxn(),
     )
 
-    all_bond_cores = [_compute_bond_cores(mol_a, mol_b, marcs) for marcs in all_marcs]
```
badisa, line 440: This is where the bulk of the performance gain comes from, and we never used the bonds.

timemachine/fe/atom_mapping.py
```diff
@@ -451,8 +430,9 @@ def leaf_filter_fxn(trial_core) -> bool:
 
         # distance score
         r2_ij = np.sum(np.power(r_i - r_j, 2))
-        rmsd = np.sqrt(r2_ij / len(core))
-        dists.append(rmsd)
+        # No need to perform square root here, only care about ordering
+        mean_sq_dist = r2_ij / len(core)
```
badisa, line 434: A negligible difference, but no reason to take the square root since we only use this value to order.

timemachine/fe/utils.py
```diff
@@ -249,9 +249,8 @@ def plot_atom_mapping_grid(
         hbcs.append(bond_colors_a)
         hbcs.append(bond_colors_b)
 
-    num_mols = len(extra_mols) + 2
-
     all_mols = [mol_a_3d, mol_b_3d, *extra_mols]
+    num_mols = len(all_mols)
```
badisa, line 253: Just noticed this and decided to avoid the hard coded + 2

tests/test_atom_mapping.py
mcwitt: Worth adding `set_mol_name` to `timemachine.fe.utils` to complement `get_mol_name`?

timemachine/fe/atom_mapping.py
```diff
@@ -265,42 +265,22 @@ def induce_mol_subgraph(mol_a, core_a, bond_core_a):
     return sg_a
 
 
-def _compute_bond_cores(mol_a, mol_b, marcs):
```
mcwitt, line 268: I've noticed a couple of other functions in this module seem unused: `bfs`, `find_cycles`, `induce_mol_subgraph`, `update_bond_core`. Would it make sense to also clean these up in this PR?

timemachine/fe/atom_mapping.py
```diff
@@ -451,8 +430,9 @@ def leaf_filter_fxn(trial_core) -> bool:
 
         # distance score
         r2_ij = np.sum(np.power(r_i - r_j, 2))
-        rmsd = np.sqrt(r2_ij / len(core))
-        dists.append(rmsd)
+        # No need to perform square root here, only care about ordering
+        mean_sq_dist = r2_ij / len(core)
+        dists.append(mean_sq_dist)
```
mcwitt, line 435: nit: rename `dists` to `mean_sq_dists`?

mcwitt (approved):

timemachine/fe/atom_mapping.py
badisa: Removed in https://github.com/proteneer/timemachine/pull/1362/commits/807a91ca14b53803fb91bc0a41fe648a7fa89858

timemachine/fe/atom_mapping.py
badisa: Renamed in https://github.com/proteneer/timemachine/pull/1362/commits/4eeb264d7504193e4981f41ce4ac25634c642781

tests/test_atom_mapping.py
badisa: Added in https://github.com/proteneer/timemachine/pull/1362/commits/17f22fc9c050f52732adcf81fe00fbbfb2dbd14d

PR #1363, opened 2024-08-16
Add ring_forming_breaking_count metric

badisa:
Ring breaking/forming is often an expensive transformation, and this metrics could help us identify expensive edges upfront. Either to modify how maps are constructed (i.e. down weight edges that have ring forming/breaking) or simply as an additional indicator of the difficulty of a transformation.

Note that `core_bonds_broken_count` only considers core-core bonds broken, which is zero when `enforce_core_core` is true. 
https://github.com/proteneer/timemachine/blob/030faa23ec5e1c3e43b4f628990b002c225d2e69/timemachine/fe/atom_mapping.py#L239-L251

This function only identifies where a ring is formed or broken, (i.e. the ring is disconnected in the intermediate states) and does not consider ring insertions as forming/breaking since an insertion maintains the ring bonds through the windows.

### DIFF
--- a/tests/test_atom_mapping.py
+++ b/tests/test_atom_mapping.py
@@ -1597,3 +1597,23 @@ M  END
     )
 
     assert atom_mapping.ring_breaking_count(mol_a, mol_b, core) == (2, 0)
+
+    mol_a = Chem.AddHs(Chem.MolFromSmiles("C1CC2=C(C1)C=CC=C2"))
+    set_mol_name(mol_a, "bicycle_a")
+    mol_b = Chem.AddHs(Chem.MolFromSmiles("C1=CC2=C(C=C1)C=CC=C2"))
+    set_mol_name(mol_b, "bicycle_b")
+
+    AllChem.EmbedMolecule(mol_a, randomSeed=2024)
+    AllChem.EmbedMolecule(mol_b, randomSeed=2024)
+
+    core = np.array(
+        [
+            [5, 1],
+            [6, 0],
+            [7, 5],
+            [8, 4],
+        ]
+    )
+    AllChem.AlignMol(mol_a, mol_b, atomMap=[(6, 0), (7, 5)])
+
+    assert atom_mapping.ring_breaking_count(mol_a, mol_b, core) == (1, 1)
--- a/tests/test_atom_mapping.py
+++ b/tests/test_atom_mapping.py
@@ -1589,6 +1589,11 @@ def test_ring_breaking_count(hif2a_ligands):
     assert atom_mapping.ring_breaking_count(mol_a, mol_b, core) == (0, 0)
     assert atom_mapping.ring_breaking_count(mol_a, mol_b, core) == ref_ring_breaking_count(mol_a, mol_b, core)
 
+    cores = atom_mapping.get_cores(mol_a, mol_b, **{**DEFAULT_ATOM_MAPPING_KWARGS, "max_connected_components": 2})
+    core = cores[0]
+    assert atom_mapping.ring_breaking_count(mol_a, mol_b, core) == (1, 1)
+    assert atom_mapping.ring_breaking_count(mol_a, mol_b, core) == ref_ring_breaking_count(mol_a, mol_b, core)
+
     mol_a = Chem.MolFromMolBlock(
         """bridgehead system
   MJ240300
--- a/tests/test_atom_mapping.py
+++ b/tests/test_atom_mapping.py
@@ -1561,3 +1561,39 @@ def test_ring_forming_breaking_count(hif2a_ligands):
     cores = atom_mapping.get_cores(mol_a, mol_b, **DEFAULT_ATOM_MAPPING_KWARGS)
     core = cores[0]
     assert atom_mapping.ring_forming_breaking_count(mol_a, mol_b, core) == 2
+
+    mol_a = Chem.MolFromMolBlock(
+        """bridgehead system
+  MJ240300
+
+  5  6  0  0  0  0  0  0  0  0999 V2000
+    1.2868    0.3797   -0.0018 C   0  0  0  0  0  0  0  0  0  0  0  0
+    1.5802   -0.3964   -0.0438 C   0  0  0  0  0  0  0  0  0  0  0  0
+    0.7505   -0.3790    0.0024 C   0  0  0  0  0  0  0  0  0  0  0  0
+    0.7069    0.2170   -0.5741 C   0  0  0  0  0  0  0  0  0  0  0  0
+    0.7711    0.1787    0.6173 C   0  0  0  0  0  0  0  0  0  0  0  0
+  1  2  1  0  0  0  0
+  2  3  1  0  0  0  0
+  3  4  1  0  0  0  0
+  4  1  1  0  0  0  0
+  3  5  1  0  0  0  0
+  5  1  1  0  0  0  0
+M  END
+"""
+    )
+    mol_a = Chem.AddHs(mol_a)
+    mol_b = Chem.AddHs(Chem.MolFromSmiles("CCC"))
+    set_mol_name(mol_b, "chain")
+    AllChem.EmbedMolecule(mol_a, randomSeed=2024)
+    AllChem.EmbedMolecule(mol_b, randomSeed=2024)
+    AllChem.AlignMol(mol_b, mol_a, atomMap=[(0, 0), (1, 1), (2, 2)])
+
+    core = np.array(
+        [
+            [0, 0],
+            [1, 1],
+            [2, 2],
+        ]
+    )
+
+    assert atom_mapping.ring_forming_breaking_count(mol_a, mol_b, core) == 2
--- a/tests/test_atom_mapping.py
+++ b/tests/test_atom_mapping.py
@@ -1520,3 +1520,44 @@ M  END""",
     assert diagnostics_1.num_cores < diagnostics_1.total_leaves_visited
     for core in cores_1:
         verify_chiral_validity_of_core(mol_a, mol_b, core, ff)
+
+
+def test_ring_forming_breaking_count(hif2a_ligands):
+    mol_a = Chem.AddHs(Chem.MolFromSmiles("C1=CC=CC=C1"))  # benzene
+    set_mol_name(mol_a, "benzene")
+    mol_b = Chem.AddHs(Chem.MolFromSmiles("CC1=CC=CC=C1"))  # benzene with a methyl
+    set_mol_name(mol_b, "methyalated")
+    AllChem.EmbedMolecule(mol_a, randomSeed=2024)
+    AllChem.EmbedMolecule(mol_b, randomSeed=2024)
+    AllChem.AlignMol(mol_b, mol_a, atomMap=[(0, 0), (4, 3)])
+
+    # Construct core over an incomplete ring, resulting in a ring forming in both endstates
+    core = np.array(
+        [
+            [3, 3],
+            [4, 4],
+            [5, 5],
+        ]
+    )
+    assert atom_mapping.ring_forming_breaking_count(mol_a, mol_b, core) == 2
+    assert atom_mapping.ring_forming_breaking_count(mol_b, mol_a, core[:, [1, 0]]) == 2
+
+    mols_by_name = {get_mol_name(m): m for m in hif2a_ligands}
+    # This transformation will always form a ring
+    mol_a = mols_by_name["15"]
+    mol_b = mols_by_name["30"]
+    cores = atom_mapping.get_cores(mol_a, mol_b, **DEFAULT_ATOM_MAPPING_KWARGS)
+    core = cores[0]
+
+    assert atom_mapping.ring_forming_breaking_count(mol_a, mol_b, core) == 1
+    assert atom_mapping.ring_forming_breaking_count(mol_b, mol_a, core[:, [1, 0]]) == 1
+
+    # Will construct two pair of 3 ring systems of which only a single ring can be mapped
+    # with one atom in the second ring mapped. This means that the second ring is formed/broken (in both endstates)
+    # while the third ring is simply an insertion, which this function doesn't consider.
+    mol_a = make_polyphenylene(3, 0.0)
+    mol_b = make_polyphenylene(3, 90.0)
+
+    cores = atom_mapping.get_cores(mol_a, mol_b, **DEFAULT_ATOM_MAPPING_KWARGS)
+    core = cores[0]
+    assert atom_mapping.ring_forming_breaking_count(mol_a, mol_b, core) == 2
--- a/tests/test_atom_mapping.py
+++ b/tests/test_atom_mapping.py
@@ -1,7 +1,9 @@
 import copy
 import time
 from functools import partial
+from typing import Tuple
 
+import networkx as nx
 import numpy as np
 import pytest
 from rdkit import Chem
@@ -10,7 +12,7 @@ from rdkit.Chem import AllChem
 from timemachine.constants import DEFAULT_ATOM_MAPPING_KWARGS
 from timemachine.fe import atom_mapping
 from timemachine.fe.mcgregor import MaxVisitsWarning, NoMappingError
-from timemachine.fe.single_topology import DummyGroupAssignmentError, verify_chiral_validity_of_core
+from timemachine.fe.single_topology import DummyGroupAssignmentError, SingleTopology, verify_chiral_validity_of_core
 from timemachine.fe.utils import (
     get_mol_name,
     get_romol_conf,
@@ -20,6 +22,7 @@ from timemachine.fe.utils import (
     set_romol_conf,
 )
 from timemachine.ff import Forcefield
+from timemachine.graph_utils import convert_to_nx
 
 pytestmark = [pytest.mark.nocuda]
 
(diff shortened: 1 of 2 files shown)


timemachine/fe/atom_mapping.py
```diff
@@ -236,6 +243,78 @@ def _deduplicate_all_cores(all_cores):
...
+def ring_forming_breaking_count(mol_a, mol_b, core: NDArray) -> int:
+    """Counts the number of rings broken and formed given a core. Does not count ring insertion/deletions. Considers
+    the number number of rings broken/formed in both endstates, reversing the mol order and the core order produces
+    the same value.
+
+    The algorithm is to add all ring atoms/bonds to a graph, and count the cycles that contain both core and mol a or
+    mol b atoms.
+    """
+    g = nx.Graph()
+    core_a_to_b = {a: b for a, b in core}
```
mcwitt, line 255: nit: it seems like we only look at the keys of this dict. Simpler to define e.g. `core_atoms_a` as a set?

timemachine/fe/atom_mapping.py
```diff
@@ -236,6 +243,78 @@ def _deduplicate_all_cores(all_cores):
...
+                g.add_node(idx, atom_type=AtomMapFlags.MOL_A)
+    for atom in mol_b.GetAtoms():
+        idx = atom.GetIdx()
+        if atom.IsInRing():
+            if idx in core_b_to_a:
+                g.add_node(idx + mol_a.GetNumAtoms(), atom_type=AtomMapFlags.CORE)
+            else:
+                g.add_node(idx + mol_a.GetNumAtoms(), atom_type=AtomMapFlags.MOL_B)
+
+    if len(g.nodes) == 0:
```
mcwitt, line 272: nit:
```suggestion
    # if there are no ring atoms in the core, return 0
    if len(g.nodes) == 0:
```

timemachine/fe/atom_mapping.py
```diff
@@ -236,6 +243,78 @@ def _deduplicate_all_cores(all_cores):
...
+        if atom.IsInRing():
+            if idx in core_a_to_b:
+                g.add_node(idx, atom_type=AtomMapFlags.CORE)
+            else:
+                g.add_node(idx, atom_type=AtomMapFlags.MOL_A)
+    for atom in mol_b.GetAtoms():
+        idx = atom.GetIdx()
+        if atom.IsInRing():
+            if idx in core_b_to_a:
+                g.add_node(idx + mol_a.GetNumAtoms(), atom_type=AtomMapFlags.CORE)
```
mcwitt, line 268: I'm a bit confused here. Are the nodes in the graph intended to be labeled according to the combined mol atom index? If so, should this use `atom_map.b_to_c`?

tests/test_atom_mapping.py
```diff
@@ -1520,3 +1520,44 @@ def test_enforce_chirally_valid_dummy_groups(seed):
...
+
+    # Construct core over an incomplete ring, resulting in a ring forming in both endstates
+    core = np.array(
+        [
+            [3, 3],
+            [4, 4],
+            [5, 5],
+        ]
+    )
+    assert atom_mapping.ring_forming_breaking_count(mol_a, mol_b, core) == 2
```
mcwitt, line 1542: Question: is it possible to determine the number of rings formed and broken separately? (e.g. could this return a tuple `(# rings formed, # rings broken)`)

timemachine/fe/atom_mapping.py
badisa: Converted to sets in https://github.com/proteneer/timemachine/pull/1363/commits/dc38731f87e365add2183c2c2b91fffd914bb896

timemachine/fe/atom_mapping.py
```diff
@@ -236,6 +243,79 @@ def _deduplicate_all_cores(all_cores):
...
+    for i, comp in enumerate(nx.connected_components(g)):
+        subgraph = g.subgraph(comp)
+        if all([data["atom_type"] == AtomMapFlags.CORE for _, data in subgraph.nodes(data=True)]):
+            continue
+
+        for atom_type in [AtomMapFlags.MOL_A, AtomMapFlags.MOL_B]:
+            ring_atoms = subgraph.subgraph(
+                [n for n, data in subgraph.nodes(data=True) if data["atom_type"] != atom_type]
+            )
+            core_atoms = [data["atom_type"] == AtomMapFlags.CORE for _, data in ring_atoms.nodes(data=True)]
```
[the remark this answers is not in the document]
badisa: Renamed in https://github.com/proteneer/timemachine/pull/1363/commits/f8defe648e3ea89213803d0c66a7725f1a3648a1

timemachine/fe/atom_mapping.py
```diff
@@ -236,6 +243,79 @@ def _deduplicate_all_cores(all_cores):
...
+                continue
+            # If the ring atoms are all core atoms, no ring break has occurred
+            if all(ring_atom_is_core):
+                continue
+            try:
+                nx.find_cycle(ring_atoms)
+                core_breaks += 1
+            except nx.exception.NetworkXNoCycle:
+                pass
+    return core_breaks
```
mcwitt, line 316: nit: name seems misleading, IIUC: is this counting the number of breaks and formations, and not just in the core?

tests/test_atom_mapping.py
mcwitt: Mostly just wondering, don't have a concrete use case yet. It does seem nice to provide the additional information if it's an easy change (could imagine this being useful for "directional" estimates like SMC at some point).

timemachine/fe/atom_mapping.py
badisa: Renamed in https://github.com/proteneer/timemachine/pull/1363/commits/8728b64ecc6fb30ff546bd1209016e31afd0091d

timemachine/fe/atom_mapping.py
badisa: Per offline discussion moved to using alchemical indices: https://github.com/proteneer/timemachine/pull/1363/commits/8728b64ecc6fb30ff546bd1209016e31afd0091d

badisa:
Came upon an edge case I am not certain on how to handle, closing for now.

badisa:
Figured it out and simplified some of the code

tests/test_atom_mapping.py
badisa: Added in https://github.com/proteneer/timemachine/pull/1363/commits/46e46f7f65c77ed68833693709e73b922c0756d8

mcwitt:
In trying to understand the definition of ring breaking in this PR, I looked for differences relative to this very simple (but inefficient) implementation:

```python
def ring_breaking_count(mol_a, mol_b, core, ff=Forcefield.load_default()) -> Tuple[int, int]:
    g_a = convert_to_nx(st.mol(0.0))
    g_b = convert_to_nx(st.mol(1.0))
    return ring_breaking_count_impl(g_a, g_b)


def ring_breaking_count_impl(g_a: nx.Graph, g_b: nx.Graph) -> Tuple[int, int]:
    cycles_a = {frozenset(cycle) for cycle in nx.cycle_basis(g_a)}
    cycles_b = {frozenset(cycle) for cycle in nx.cycle_basis(g_b)}
    return len(cycles_a - cycles_b), len(cycles_b - cycles_a)
```

One difference is in the way the following case is classified:
image

The version in this PR yields (1, 1), while the code above gives (0, 0). Would we really want to consider this as ring formation, though? Intuitively, it doesn't seem much different in difficulty than if the single carbon in the second ring had been removed from the map.

badisa:
> ```python
> convert_to_nx
> ```

Looks like I misunderstood when we do or don't break bonds. I had thought we would break a bond here. But I guess in this case we leave all the bonds attached since a single core. Should be an easy fix as it simply means ensuring that any cycle contains at least 1 core-core bond. 

Will add the reference implementation and compare against it in the tests

badisa:
> In trying to understand the definition of ring breaking in this PR, I looked for differences relative to this very simple (but relatively inefficient) implementation:
> 
> ```python
> def ring_breaking_count(st: SingleTopology) -> Tuple[int, int]:
>     def get_cycle_basis(mol):
>         g = convert_to_nx(mol)
>         return {frozenset(cycle) for cycle in nx.minimum_cycle_basis(g)}
> 
>     cycles_0 = get_cycles(st.mol(0.0))
>     cycles_1 = get_cycles(st.mol(1.0))
> 
>     return len(cycles_0 - cycles_1), len(cycles_1 - cycles_0)
> ```
> 
> One difference is in the way the following case is classified: image
> 
> The version in this PR yields (1, 1), while the code above gives (0, 0). But I'm not sure why we would consider this transformation to be ring breaking?

Great catch, fixed up in https://github.com/proteneer/timemachine/pull/1363/commits/2a3cdea5fa06bbfae2cb63d29178dda907d49ea3

badisa:
Woops, didn't add the right condition. Need to have a core-core bond in the cycle, as otherwise linker changes would count as a ring forming/closing

badisa:
> Woops, didn't add the right condition. Need to have a core-core bond in the cycle, as otherwise linker changes would count as a ring forming/closing

Take that back, realized for linker cases it is a ring breaking transformation.

mcwitt:
Thinking about this more, my preferred eventual approach would be to make this a method of `SingleTopology`, reusing the same logic that is used to parameterize end-state harmonic bond potentials for the combined mol. Otherwise, it seems like we need duplication of the logic that determines which bonds are present at each end state (e.g. which bonds are broken to maintain factorizability; speaking of which, does this need to be handled here?)

badisa:
> Thinking about this more, my preferred eventual approach would be to make this a method of `SingleTopology`, reusing the same logic that is used to parameterize end-state harmonic bond potentials for the combined mol. Otherwise, it seems like we need to duplicate and keep in sync the logic that determines which bonds are present at each end state (e.g. which bonds are broken to maintain factorizability; speaking of which, does this need to be handled here?)
> 
> Here's a possible implementation along those lines:
> 
> ```python
> def ring_breaking_count(self: "SingleTopology") -> Tuple[int, int]:
>     def get_cycles(bonds):
>         g = nx.Graph()
>         g.add_edges_from(bonds.tolist())
>         return {frozenset(cycle) for cycle in nx.minimum_cycle_basis(g)}
> 
>     cycles_src = get_cycles(self.src_system.bond.potential.idxs)
>     cycles_dst = get_cycles(self.dst_system.bond.potential.idxs)
> 
>     return len(cycles_src - cycles_dst), len(cycles_dst - cycles_src)
> ```

Will play around with this, but even computing the dummy groups right now is significantly slower than this metric. To the point that I wouldn't want to use the number of dummy groups as a metric where the number of transformations is large, even though it is a useful metric.

badisa:
> ```python
>  def get_cycles(bonds):
>         g = nx.Graph()
>         g.add_edges_from(bonds.tolist())
>         return {frozenset(cycle) for cycle in nx.minimum_cycle_basis(g)}
> 
>     cycles_src = get_cycles(self.src_system.bond.potential.idxs)
>     cycles_dst = get_cycles(self.dst_system.bond.potential.idxs)
> 
>     return len(cycles_src - cycles_dst), len(cycles_dst - cycles_src)
> ```

Trying something as you described off of `AtomMapMixin`
```
    def ring_breaking_count(self):
        find_chirally_valid_dummy_groups = make_find_chirally_valid_dummy_groups(self.mol_a, self.mol_b)
        dummy_groups = find_chirally_valid_dummy_groups(self.core)
        if dummy_groups is None:
            raise DummyGroupAssignmentError("Unable to find chirally-valid dummy group assignment")
        dummy_groups_ab, dummy_groups_ba = dummy_groups
        ff = Forcefield.load_from_file("placeholder_ff.py")
        setup_end_state_harmonic_bond_and_chiral_potentials_ab = make_setup_end_state_harmonic_bond_and_chiral_potentials(
            self.mol_a, self.mol_b, ff, verify=False
        )
        setup_end_state_harmonic_bond_and_chiral_potentials_ba = make_setup_end_state_harmonic_bond_and_chiral_potentials(
            self.mol_b, self.mol_a, ff, verify=False
        )

        bond_pot_src, _, _ = setup_end_state_harmonic_bond_and_chiral_potentials_ab(
            self.core, self.a_to_c, self.b_to_c, dummy_groups_ab
        )
        bond_pot_dst, _, _ = setup_end_state_harmonic_bond_and_chiral_potentials_ba(
            self.core[:, ::-1], self.b_to_c, self.a_to_c, dummy_groups_ba
        )

        def get_cycles(bonds):
            g = nx.Graph()
            g.add_edges_from(bonds.tolist())
            return {frozenset(cycle) for cycle in nx.minimum_cycle_basis(g)}

        cycles_src = get_cycles(bond_pot_src.potential.idxs)
        cycles_dst = get_cycles(bond_pot_dst.potential.idxs)

        return len(cycles_src - cycles_dst), len(cycles_dst - cycles_src)
```

It appears to be about 30x (guessing this is worse in some cases based on trying this on a large set of ~3000 edges) slower testing off of `get_hif2a_ligand_pair_single_topology()`.
```
In [9]: %timeit mixin.ring_breaking_count()
34.1 ms ± 226 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [11]: %timeit ring_breaking_count(mol_a, mol_b, core)
1.31 ms ± 2.93 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

Now there is obviously more work being done than needs to be (chiral and angles), though it seems useful to be able to recapitulate how exactly we expect ring breaking to occur (and that may not be this PR) without having to go through code that is doing significantly more than just determining which bonds break/form.

mcwitt:
One remaining question I have is whether the choice of dummy group anchor can affect the number of rings broken. If not (and so far I haven't been able to think of a counterexample), the implementation in this PR seems consistent to me.

badisa:
> One remaining question I have is whether the choice of dummy group anchor can affect the number of rings broken. If not (and so far I haven't been able to think of a counterexample), the implementation in this PR seems consistent to me.

That is a very good question... I can compare the code I showed above to the implementation in this PR against the 3000 edges I have to see. I would hope not, but since the groups are picked only based on being chirally correct I could imagine that it can be different, but no case comes to mind.

mcwitt:
> > ```python
> >  def get_cycles(bonds):
> >         g = nx.Graph()
> >         g.add_edges_from(bonds.tolist())
> >         return {frozenset(cycle) for cycle in nx.minimum_cycle_basis(g)}
> > 
> >     cycles_src = get_cycles(self.src_system.bond.potential.idxs)
> >     cycles_dst = get_cycles(self.dst_system.bond.potential.idxs)
> > 
> >     return len(cycles_src - cycles_dst), len(cycles_dst - cycles_src)
> > ```
> 
> Trying something as you described off of `AtomMapMixin`
> 
> ```
>     def ring_breaking_count(self):
>         find_chirally_valid_dummy_groups = make_find_chirally_valid_dummy_groups(self.mol_a, self.mol_b)
>         dummy_groups = find_chirally_valid_dummy_groups(self.core)
>         if dummy_groups is None:
>             raise DummyGroupAssignmentError("Unable to find chirally-valid dummy group assignment")
>         dummy_groups_ab, dummy_groups_ba = dummy_groups
>         ff = Forcefield.load_from_file("placeholder_ff.py")
>         setup_end_state_harmonic_bond_and_chiral_potentials_ab = make_setup_end_state_harmonic_bond_and_chiral_potentials(
>             self.mol_a, self.mol_b, ff, verify=False
>         )
>         setup_end_state_harmonic_bond_and_chiral_potentials_ba = make_setup_end_state_harmonic_bond_and_chiral_potentials(
>             self.mol_b, self.mol_a, ff, verify=False
>         )
> 
>         bond_pot_src, _, _ = setup_end_state_harmonic_bond_and_chiral_potentials_ab(
>             self.core, self.a_to_c, self.b_to_c, dummy_groups_ab
>         )
>         bond_pot_dst, _, _ = setup_end_state_harmonic_bond_and_chiral_potentials_ba(
>             self.core[:, ::-1], self.b_to_c, self.a_to_c, dummy_groups_ba
>         )
> 
>         def get_cycles(bonds):
>             g = nx.Graph()
>             g.add_edges_from(bonds.tolist())
>             return {frozenset(cycle) for cycle in nx.minimum_cycle_basis(g)}
> 
>         cycles_src = get_cycles(bond_pot_src.potential.idxs)
>         cycles_dst = get_cycles(bond_pot_dst.potential.idxs)
> 
>         return len(cycles_src - cycles_dst), len(cycles_dst - cycles_src)
> ```
> 
> It appears to be about 30x (guessing this is worse in some cases based on trying this on a large set of ~3000 edges) slower testing off of `get_hif2a_ligand_pair_single_topology()`.
> 
> ```
> In [9]: %timeit mixin.ring_breaking_count()
> 34.1 ms ± 226 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
> 
> In [11]: %timeit ring_breaking_count(mol_a, mol_b, core)
> 1.31 ms ± 2.93 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
> ```
> 
> Now there is obviously more work being done than needs to be (chiral and angles), though it seems useful to be able to recapitulate how exactly we expect ring breaking to occur (and that may not be this PR) without having to go through code that is doing significantly more than just determining which bonds break/form.

This might be further optimized by avoiding `find_chirally_valid_dummy_groups` and just choosing an arbitrary dummy group assignment with `next(generate_dummy_group_assignments(...))`, under the assumption that the choice of dummy groups doesn't affect the result (which I think is implicitly assumed in this PR anyway?).

mcwitt (approved):

badisa:
> under the assumption that the choice of dummy groups doesn't affect the result (which I think is implicitly assumed in this PR anyway?).

This turned out to be wrong. Found that this can be off by one in either in one direction or both directions based on the dummy anchor. Indicates that we might be able to improve things by choosing better anchors (at least in the case I have looked closely at), but not sure having two core breaks is all that different to one.

Going to close this for now as this is really only a heuristic rather than a measure of ring breaking. Might still be useful, but probably doesn't need to live in the repo.

PR #1364, opened 2024-08-19
Remove guard threshold from local minimize

badisa:
* Make it impossible to minimize in some cases

---------

### DIFF
--- a/tests/test_minimizer.py
+++ b/tests/test_minimizer.py
@@ -7,7 +7,8 @@ from rdkit import Chem
 from rdkit.Chem import AllChem
 
 from timemachine.fe.free_energy import HostConfig
-from timemachine.fe.utils import get_mol_name, read_sdf
+from timemachine.fe.model_utils import get_vacuum_val_and_grad_fn
+from timemachine.fe.utils import get_mol_name, get_romol_conf, read_sdf
 from timemachine.ff import Forcefield
 from timemachine.ff.handlers import openmm_deserializer
 from timemachine.md import builders, minimizer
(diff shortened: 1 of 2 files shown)


timemachine/md/minimizer.py
```diff
@@ -488,18 +488,10 @@ def local_minimize(
...
-    # deal with overflow, empirically obtained by testing on some real systems.
-    guard_threshold = 5e4
-
     def val_and_grad_fn_local(x_local):
         x_prime = x0.copy()
         x_prime[local_idxs] = x_local
         u_full, grad_full = val_and_grad_fn(x_prime)
-        # avoid being trapped when overflows spuriously appear as large negative numbers
-        # remove after resolution of https://github.com/proteneer/timemachine/issues/481
-        if not np.isfinite(u_full) or u_0 - u_full > guard_threshold:
```
mcwitt, line 500: Is there still a circumstance where we'd want to convert NaNs to inf here?

proteneer: I don't remember if the fixed point code now can still emit nans (since iirc every fixed point energy now has a representable float32 representation).

badisa: It can still return Nan. https://github.com/proteneer/timemachine/blob/master/tests/test_energy_overflows.py#L23 but has to be a pretty egregious clash. 

I think it could help to convert the NaN to an inf, will add back the `if not np.isfinite(u_full)`

mcwitt (approved):

timemachine/md/minimizer.py
badisa: Added back in the `np.isfinite(u_full)` in https://github.com/proteneer/timemachine/pull/1364/commits/d9940a78c33ecaad68e0a9122b2e4a00d273b4c1

Though do we want this to catch `-np.inf` in the case of using the jax platform?

mcwitt: Maybe `np.isnan`, specifically to handle the case where the GPU platform encounters a fixed-point overflow and returns NaN?

badisa: Would we want to set the forces to Nan when the energies are +/- inf?

tests/test_minimizer.py
```diff
@@ -201,6 +202,84 @@ def test_local_minimize_water_box():
     np.testing.assert_array_equal(g_init, g_init_test)
 
 
+def test_local_minimize_strained_ligand():
```
badisa, line 205: Fails without the removal of the guard threshold

timemachine/md/minimizer.py
mcwitt: > Would we want to set the forces to Nan when the energies are +/- inf?

I might be missing something, but I don't see why this is necessary.

badisa: Added `isnan` check in https://github.com/proteneer/timemachine/pull/1364/commits/2fad3ce6e8425c84f28d42bb41ba41090ab4a98e

> I might be missing something, but I don't see why this is necessary.

Just don't know BFGS well enough to know what the behavior is when the energy is non-finite while the gradients are finite.

maxentile (approved):

PR #1366, opened 2024-08-23
Assert that states differ only in their parameters

mcwitt:
Improves the safety of an optimization we currently use in free energy computations where we only instantiate potentials for one end state and assume the other states differ only in their parameters.

### DIFF
--- a/tests/test_free_energy.py
+++ b/tests/test_free_energy.py
@@ -1,3 +1,5 @@
+from copy import deepcopy
+from dataclasses import replace
 from functools import partial
 from importlib import resources
 from typing import List, Optional
@@ -23,6 +25,7 @@ from timemachine.fe.free_energy import (
     MinOverlapWarning,
     PairBarResult,
     Trajectory,
+    assert_potentials_compatible,
     batches,
     compute_potential_matrix,
     estimate_free_energy_bar,
@@ -41,6 +44,7 @@ from timemachine.md import builders
 from timemachine.md.hrex import HREX, HREXDiagnostics
 from timemachine.md.states import CoordsVelBox
 from timemachine.potentials import BoundPotential, Nonbonded, SummedPotential
+from timemachine.potentials.potentials import HarmonicBond, NonbondedPairListPrecomputed
 from timemachine.testsystems.relative import get_hif2a_ligand_pair_single_topology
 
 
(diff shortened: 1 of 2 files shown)


timemachine/fe/free_energy.py
```diff
@@ -803,6 +803,29 @@ def make_pair_bar_plots(res: PairBarResult, temperature: float, prefix: str) ->
     return PairBarPlots(dG_errs_png, overlap_summary_png, overlap_detail_png)
 
 
+def assert_potentials_compatible(bps1: Sequence[BoundPotential], bps2: Sequence[BoundPotential]):
+    """Asserts that two sequences of bound potentials are equivalent except for their parameters"""
+
+    def field_equal(x1, x2):
+        assert type(x1) is type(x2)
```
badisa, line 810: Could you add string messages to these assertions? Be helpful to know what the error is immediately without having to find the traceback (which we don't always have)

badisa: Would also strengthen tests to verify that the error you expect is being exercised

badisa (approved):

proteneer (approved):
nice safety check!

tests/test_free_energy.py
```diff
@@ -553,3 +557,46 @@ def test_trajectories_by_replica_to_by_state(seed, n_states, n_iters, n_atoms):
     traj_by_replica = sim_res.extract_trajectories_by_replica(atom_idxs)
     traj_by_state = trajectories_by_replica_to_by_state(traj_by_replica, replica_idx_by_state_by_iter)
     np.testing.assert_array_equal(traj_by_state, frames[:, :, atom_idxs])
+
+
+def test_assert_potentials_compatible(hif2a_ligand_pair_single_topology):
```
badisa, line 562: nit: Think this can be put to `nogpu` marker.

mcwitt: Added in https://github.com/proteneer/timemachine/pull/1366/commits/4d1d26f3e81391bd08dd6f02b675acae4fed5cf7

timemachine/fe/free_energy.py
mcwitt: Improved assertion failure messages and strengthened tests in https://github.com/proteneer/timemachine/pull/1366/commits/ba6fb80cb2c9db0b658520627724777d7efc4e05

PR #1368, opened 2024-08-27
Adds test that verifies Neighborlist Tile Density

badisa:
* A handy thing to have when we want to improve performance of NBList since we talk about tile density a lot.


tests/test_nblist.py
```diff
@@ -376,7 +380,109 @@ def test_nblist_max_interactions(block_size, tiles):
...
+        nblist = custom_ops.Neighborlist_f64(initial_state.x0.shape[0])
+    occupancies = []
+    occupancy = compute_block_occupancy(nblist, initial_state.x0, initial_state.box0, cutoff)
+    occupancies.append(occupancy)
+    for i, (frame, box) in enumerate(zip(traj.frames, traj.boxes)):
+        occupancy = compute_block_occupancy(nblist, frame, box, cutoff)
+        occupancies.append(occupancy)
+    total_steps = md_params.steps_per_frame * md_params.n_frames + md_params.n_eq_steps
+    print(f"Starting occupancy {occupancies[0]}")
+    print(f"Final occupancy after {total_steps:} {occupancies[-1]}")
```
mcwitt, line 480: Can we assert anything meaningful here?

tests/test_nblist.py
```diff
@@ -376,7 +380,109 @@ def test_nblist_max_interactions(block_size, tiles):
...
+    host = setup_optimized_host(st, host_config)
+
+    lambda_grid = np.array([0.0])
+
+    initial_state = setup_initial_states(st, host, DEFAULT_TEMP, lambda_grid, seed=2024, min_cutoff=None)[0]
+
+    return st, host, host_name, initial_state
+
+
+def compute_block_occupancy(
```
mcwitt, line 424: nit: `compute_mean_block_occupancy`?

badisa: Fixed in https://github.com/proteneer/timemachine/pull/1368/commits/c72b0aba081df393cfd72b25dc10831ec153abe3

tests/test_nblist.py
badisa: Added two sanity checks, first that the occupancy improves as the barostat runs and that the occupancies are greater than 40%.

https://github.com/proteneer/timemachine/pull/1368/commits/4f6af9dc56144ad0c577461f231b974ce24abde4

mcwitt (approved):

tests/test_nblist.py
proteneer: can we rename this to nblist_tile_density to be more precise instead of occupancy?

tests/test_nblist.py
```diff
@@ -376,7 +380,114 @@ def test_nblist_max_interactions(block_size, tiles):
...
+def compute_mean_block_occupancy(
+    nblist, frame: NDArray, box: NDArray, cutoff: float, column_block: int = 32, row_block: int = 32
+) -> float:
+    assert column_block == 32
+    assert column_block == row_block
+    N = frame.shape[0]
+    assert nblist.get_num_row_idxs() == N
+    ixn_list = nblist.get_nblist(frame, box, cutoff)
+    block_occupancies = []
+    max_ixn_per_block = column_block * row_block * ((N + row_block - 1) // row_block)
```
proteneer, line 433: why the extra  `((N + row_block - 1) // row_block)`?

tests/test_nblist.py
```diff
@@ -376,7 +380,114 @@ def test_nblist_max_interactions(block_size, tiles):
...
+    md_params = MDParams(n_eq_steps=0, steps_per_frame=100, n_frames=frames, seed=2024)
+
+    traj = sample(initial_state, md_params, max_buffer_frames=md_params.n_frames)
+
+    if precision == np.float32:
+        nblist = custom_ops.Neighborlist_f32(initial_state.x0.shape[0])
+    else:
+        nblist = custom_ops.Neighborlist_f64(initial_state.x0.shape[0])
+    occupancies = []
+    occupancy = compute_mean_block_occupancy(nblist, initial_state.x0, initial_state.box0, cutoff)
```
proteneer, line 473: is `initial_state.x0` equilibrated? (equilibrated tile densities will be different from non-equilibrated tile densities)

tests/test_nblist.py
badisa: Because the structure of `ixn_list` from `nblist.get_nblist` returns the full rectangular block of row atoms by all column atoms. So it is the number of ixns in a single tile times the number of tiles.

tests/test_nblist.py
proteneer: also, how much does the starting tile density differ from the final frame's tile density?

tests/test_nblist.py
badisa: It will have a pre-equilibrated (1000 steps with Barostat, from `pre_equilibrate_host`) but is not equilibrated. If you run this out longer you can see a significant shift in the nblist density.

tests/test_nblist.py
```diff
@@ -376,7 +380,114 @@ def test_nblist_max_interactions(block_size, tiles):
...
+    host = setup_optimized_host(st, host_config)
+
+    lambda_grid = np.array([0.0])
+
+    initial_state = setup_initial_states(st, host, DEFAULT_TEMP, lambda_grid, seed=2024, min_cutoff=None)[0]
+
+    return st, host, host_name, initial_state
+
+
+def compute_mean_block_occupancy(
```
proteneer: Does this compute the `block` or the `tile` density? a block usually means a "single group of atoms", where as a tile means, "two groups of atoms forming a tile of interactions"

tests/test_nblist.py
```diff
@@ -376,7 +380,148 @@ def test_nblist_max_interactions(block_size, tiles):
...
+            # Find the interactions that would make up a tile, if the tile is empty
+            # don't compute density.
+            tile_column_ixns = column_ixn_atoms[
+                (column_ixn_atoms >= column_block_offset) & (column_ixn_atoms < (column_block_offset + column_block))
+            ]
+            column_block_offset += column_block
+            if tile_column_ixns.size == 0:
+                continue
+            col_coords = np.expand_dims(frame[tile_column_ixns], axis=1)
+            deltas = image_coords(row_coords - col_coords, box_diag)
```
maxentile, line 458: identical, but may be clearer to use `delta_r` https://github.com/proteneer/timemachine/blob/ea7af897d3656d1a386c0d0d1d73b1adcdfce832/timemachine/potentials/jax_utils.py#L37

tests/test_nblist.py
```diff
@@ -376,7 +380,148 @@ def test_nblist_max_interactions(block_size, tiles):
...
+            dij = np.linalg.norm(deltas, axis=-1)
+            ixns = np.sum(dij < cutoff)
+            tile_densities.append(ixns / ixns_per_tile)
+        row_block_offset += row_block
+    return float(np.mean([tile_densities]))
```
maxentile, line 463: small observation: since `ixns_per_tile` is not constant, `mean(tile_densities)` will differ slightly from `sum(ixns) / sum(ixns_per_tile)`. The computed quantity does seem like a more performance-relevant number than `sum(ixns) / sum(ixns_per_tile)`.

Maybe a few other performance-relevant numbers to monitor might include:
* how many warp-synchronous loop iterations will be executed, vs. the bare minimum number required. The idealized minimum number might be something like `ceil(sum(ixns) / 32)`, but the number executed might be something like the number of rows (or is it columns?) that have at least 1 `dij < cutoff`, summed over all tiles.
* how many 32x32 tiles are produced, vs. the bare minimum number of 32x32 tiles
* fraction of atoms per tile row/column that have interactions (there was a comment in [[Eastman, Pande, 2009]](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2841709/) about having two code paths depending on whether 12 or fewer of the 32 atoms in a block have interactions)

tests/test_nblist.py
```diff
@@ -376,7 +380,148 @@ def test_nblist_max_interactions(block_size, tiles):
...
+        nblist = custom_ops.Neighborlist_f32(coords.shape[0])
+    else:
+        nblist = custom_ops.Neighborlist_f64(coords.shape[0])
+    unsorted_density = compute_mean_tile_density(nblist, coords, box, cutoff)
+    print(f"DHFR NBList Occupancy with Cutoff {cutoff}, no sort: {unsorted_density * 100.0:.3g}%")
+    perm = hilbert_sort(coords, box)
+    coords = coords[perm]
+    density = compute_mean_tile_density(nblist, coords, box, cutoff)
+    print(f"DHFR NBList Occupancy with Cutoff {cutoff}: {density * 100.0:.3g}%")
+    assert density > 0.10
```
badisa, line 482: This is 15% for a cutoff of 0.9, @proteneer does that seem right? Offline you had said you expect 13%, but after fixing up the incorrect computation of the tile density didn't see 13%.

tests/test_nblist.py
badisa: Removed `image_coords` and used `delta_r` in https://github.com/proteneer/timemachine/pull/1368/commits/e5857d644f1f110dd44bacea33b54fbeecf11f1c

tests/test_nblist.py
badisa: Its the tile density, renamed to be more precise. Also fixed the incorrect implementation.

tests/test_nblist.py
badisa: The tile density does not differ by more than about 1%.

tests/test_nblist.py
mcwitt: Would it be informative to plot the full distribution of tile densities? If so, could this function return the list of densities? (the mean could still be computed where we print the summary)

mcwitt (approved):

proteneer (approved):

PR #1371, opened 2024-09-03
Removes Neighborlist sync from Nonbonded potentials

badisa:
* Need to remove the Ligand-Complex IxnGroup before merging this as it will determine the performance impact. Will open after #1372 to ensure that this makes a difference 
* Re-orders potentials, so that the Nonbonded potentials are upfront and we can hide the faster kernels in the 'shadow' of the neighborlist/nonbonded kernels.

# Benchmarks

A10
Cuda Arch 8.6

Lose some performance on the non-RBFE simulations (no ixn-group, pairlist) due to the overhead of running no-op kernels but make up for it in the RBFE simulations and local MD. About 5% for RBFE and 8% for local MD

## Master
```
dhfr-apo: N=23558 speed: 698.28ns/day dt: 2.5fs (ran 100000 steps in 30.94s)
dhfr-apo-barostat-interval-25: N=23558 speed: 630.02ns/day dt: 2.5fs (ran 100000 steps in 34.29s)

hif2a-apo: N=8805 speed: 1213.16ns/day dt: 2.5fs (ran 100000 steps in 17.81s)
hif2a-apo-barostat-interval-25: N=8805 speed: 1036.86ns/day dt: 2.5fs (ran 100000 steps in 20.84s)
hif2a-rbfe-barostat-interval-25: N=8840 speed: 868.29ns/day dt: 2.5fs (ran 100000 steps in 24.88s)
hif2a-rbfe-local: N=8840 speed: 1280.69ns/day dt: 2.5fs (ran 100000 steps in 16.87s)
hif2a-rbfe-barostat-interval-25-water-sampling-interval-400: N=8840 speed: 802.98ns/day dt: 2.5fs (ran 100000 steps in 26.90s)

solvent-apo: N=6282 speed: 1690.23ns/day dt: 2.5fs (ran 100000 steps in 12.78s)
solvent-apo-barostat-interval-25: N=6282 speed: 1347.52ns/day dt: 2.5fs (ran 100000 steps in 16.03s)
solvent-rbfe-barostat-interval-25: N=6317 speed: 1086.41ns/day dt: 2.5fs (ran 100000 steps in 19.89s)
solvent-rbfe-local: N=6317 speed: 1377.21ns/day dt: 2.5fs (ran 100000 steps in 15.69s)
```

## PR
```
dhfr-apo: N=23558 speed: 640.36ns/day dt: 2.5fs (ran 100000 steps in 33.73s)
[...]

PR #1374, opened 2024-09-10
Removes code that assumes ordering of potentials

badisa:
* Came out of https://github.com/proteneer/timemachine/pull/1371
* Good to move away from assuming fixed orders of potentials anyways


mcwitt (commented):
Nice, I think this will make the code more readable and future code changes safer.

Observation and optional suggestion: the pattern

```python
pot = next(bp.potential for bp in bps if isinstance(bp.potential, PotentialType)
```

and similar variations are common enough that I wonder if we should wrap them into utility functions? This would also have the benefit of raising a more interpretable error than `StopIteration` if the requested potential is not present.

Maybe a couple of functions with the following signatures would cover all of the use cases?

```
(List[Potential], PotentialType) -> Potential
(List[BoundPotential], PotentialType) -> BoundPotential
```

jkausrelay (approved):
Nice simplification, lgtm!

timemachine/potentials/potential.py
```diff
@@ -77,3 +77,39 @@ def get_custom_ops_class_name_suffix(precision: Precision):
...
+    Raises
+    ------
+        ValueError:
+            Unable to find potential with the expected type
+    """
+    result: Optional[BoundPotential[_P]] = None
+    for bp in bps:
+        if isinstance(bp.potential, pot_type):
+            result = bp
+            break
```
mcwitt, line 94: nit: I think this is equivalent to
```suggestion
    result = next((bp for bp in bps if isinstance(bp.potential, pot_type)), None)
```
in case this seems cleaner to you. (I also think it's fine as is.)

badisa: Find the version you suggested harder to read, as that None is not obvious as a default value. But I am far less experienced with the functional parts of python

mcwitt: Seems worth adding a basic test asserting the expected behavior in the simple case as well as when the requested potential type is absent or appears multiple times? (Might have missed this, but would have expected to find it in `tests/test_potential.py` if it did exist)

badisa: Added tests in https://github.com/proteneer/timemachine/pull/1374/commits/afaa0bdb9480045cf3682adb4a5103a8b51f213b

mcwitt (approved):

PR #1375, opened 2024-09-10
Add plotting to benchmarks

badisa:
* Difficult at times to tell if things are converged and the plots can help determine if the benchmarks are well converged

# Examples

Poorly converged
hif2a-apo

Well converged
hif2a-apo_converged

### DIFF
--- a/tests/test_benchmark.py
+++ b/tests/test_benchmark.py
@@ -604,7 +604,7 @@ def test_nonbonded_interaction_group_potential(hi2fa_test_frames):
     bps, frames, boxes, ligand_idxs = hi2fa_test_frames
     nonbonded_potential, nonbonded_params = get_nonbonded_pot_params(bps)
 
-    config = BenchmarkConfig(num_batches=2, steps_per_batch=0, verbose=False, generate_plots=False)
+    config = BenchmarkConfig(num_batches=2, steps_per_batch=1, verbose=False, generate_plots=False)
 
     num_param_batches = 5
     beta = 1 / (constants.BOLTZ * constants.DEFAULT_TEMP)
@@ -635,7 +635,7 @@ def test_nonbonded_interaction_group_potential(hi2fa_test_frames):
 def test_hif2a_potentials(hi2fa_test_frames):
     bps, frames, boxes, _ = hi2fa_test_frames
 
-    config = BenchmarkConfig(num_batches=2, steps_per_batch=0, verbose=False, generate_plots=False)
+    config = BenchmarkConfig(num_batches=2, steps_per_batch=1, verbose=False, generate_plots=False)
 
     num_param_batches = 5
 
--- a/tests/test_benchmark.py
+++ b/tests/test_benchmark.py
@@ -8,6 +8,7 @@ from dataclasses import dataclass
 from importlib import resources
 from typing import List, Optional
 
+import matplotlib.pyplot as plt
 import numpy as np
 import pytest
 from common import prepare_single_topology_initial_state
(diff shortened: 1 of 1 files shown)


tests/test_benchmark.py
```diff
@@ -41,6 +42,18 @@ class BenchmarkConfig:
...
+    generate_plots: bool
+
+
+def plot_batch_times(batch_times: List[float], label: str):
+    plt.title(label)
+    plt.plot(batch_times)
+    plt.axhline(np.mean(batch_times), linestyle="--", c="gray", label="Mean")
+    plt.legend()
+    plt.xlabel("Batch")
+    plt.ylabel("Batch times (s)")
```
maxentile, line 54: images attached to pr description indicate we might want to add
```suggestion
    plt.ylabel("Batch times (s)")
    plt.tight_layout()
```

maxentile (approved):

tests/test_benchmark.py
```diff
@@ -41,6 +42,18 @@ class BenchmarkConfig:
     num_batches: int
     steps_per_batch: int
     verbose: bool
+    generate_plots: bool
+
+
+def plot_batch_times(batch_times: List[float], label: str):
```
maxentile, line 48: uncertainty: more informative to plot batch times, or steps/second ?

badisa: Probably most informative to plot ns per day as that is the final value reported and this will give us a sense of how much that number of interest jumps around.

badisa: Plotting the ns per day for each batch timing: https://github.com/proteneer/timemachine/pull/1375/commits/b24d627a28b49a9d943d91e6d830e3817e7692c3

maxentile: Nice. Steps per second and maybe box volume could also be relevant.

(I assume the benchmark issue is still that the box volume is decreasing -- aka the number of particle pairs within cutoff is increasing -- relative to initialization. Maybe it's some other issue though.)

tests/test_benchmark.py
```diff
@@ -41,6 +42,22 @@ class BenchmarkConfig:
     num_batches: int
     steps_per_batch: int
     verbose: bool
+    generate_plots: bool
+
+
+def plot_batch_times(steps_per_batch: int, dt: float, batch_times: List[float], label: str):
```
mcwitt, line 48: nit: would be useful to indicate the expected units of `dt` and `batch_times` here (getting a bit confused trying to follow the conversion below).

mcwitt (approved):

tests/test_benchmark.py
badisa: The issue is just that I am benchmarking on a machine with other work going on. Added this code to be sure that when I do this 'for real' that the values are well converged.

tests/test_benchmark.py
badisa: Added in https://github.com/proteneer/timemachine/pull/1375/commits/aac5753f56de435af88715ad77c6af07542100e5

tests/test_benchmark.py
badisa: Don't think at the moment the steps per second or box volume are particularly informative. The steps per second is just the same value scaled differently and at least how I use this script is to maximize ns per day so seems best to plot that value. As for box volume it doesn't seem to matter based on what I have seen running this code on a machine without any GUI also based on work in https://github.com/proteneer/timemachine/pull/1368 the tile density doesn't change very much even after running the barostat (which surprised me)

maxentile: For some of the benchmark conditions (e.g. the ones where the box is initialized at an atypical volume), box volume appears highly correlated with speed

image

but for others (e.g. in conditions omitting a barostat), less so:

image

(Note: bad benchmarking practice -- I did not control for possible variation in clock rate over time here -- nvidia docs suggest to do so: https://developer.nvidia.com/blog/advanced-api-performance-setstablepowerstate/ )

-------------

Personally, I have found the inclusion of box volume to provide informative context on timings here, since one of the first things I might want to branch on is something like, "are the timings explainable by some factor that influences overall particle density". (Also, any big fluctuations or transients are useful to know about when assessing small measured differences in ns/day.)

If timing drifts like the one shown in the PR description persist after controlling for possible variation in clock speed (and ruling out variations in overall particle density), that could be suggestive of something else equilibrating slowly within the benchmark simulation boxes, in a way that the hilbert sort/etc. performance assumptions are sensitive to...

badisa: Added plotting of box volumes in https://github.com/proteneer/timemachine/pull/1375/commits/64dfe6d53f3eaa06f83da45e3ca0640fbb8b16ae

Also added a CLI parameter to increase the number of batches run upfront to be able to do some burn in: https://github.com/proteneer/timemachine/pull/1375/commits/a14cda6386b78a26a562316f32af8f829d03dc3a

PR #1380, opened 2024-09-12
(Do not merge) Reference Smoothcore Implementation.

proteneer:
This PR designs and implements a new softcore potential, the `SmoothcoreNonbondedInteractionGroup`, that attempts to overcome two major failings of our existing 4D decoupling regime:

1) The net-charge and polarization can vary drastically along the lambda schedule. In fact even computing an effective net charge can be quite difficult due to the non-linear response of the 4D coupling parameter. This can result in rapid oscillations that require a concomitant re-orientation of the environment (eg. water inversions). 
2) Second-order phase transition effects with large dummy groups when they're lifted "all-at-once" in 4D. Note that a sequential approach can mitigate some of this, but may still suffer from the large oscillations in net charge.

Instead, `SmoothcoreNonbondedInteractionGroup` takes in a list of `anchoring_idxs` of size `num_atoms`, specifying each atom's "anchoring" point. Each atom still retains the `w` component in the parameters, whose values are now strictly in the closed interval `[0, 1]` (as opposed to the old `[0, cutoff]`) and monotone along lambda. Here, `w=0` still denotes "fully-interacting"/"fully-extended", and `w=1` denotes "fully-non-interacting"/"fully-absorbed". Using the `anchor_idx` information in conjunction with `w`, a vector is drawn and used for scaling purposes to determine the virtual site location:

``` python
anchor_confs = old_conf[anchor_idxs]
rescale_factors = 1 - w_coords
new_conf = anchor_confs + rescale_factors * (old_conf - anchor_confs)
```

For example:

image
above: `O` is the anchoring atom for the dummy hydrogen. A scaled vector `r_OH` is used to position the location of the hydrogen.

This approach also makes it much easier to balance the net-charge, the all the charge sites are at full strength (but may be smushed/averaged out). At the end-states, `lambda=0`, and `lambda=1`, the resulting energies should be identical to the current SingleTopology end-states. A test is added to verify this behavior. 

Todos
------
- [x] charge interpolation maintains net charge
- [x] lennard jones interpolation retains a minimum sigma/eps pair at the end-state
- [x] python reference implementation
- [x] c++ implementation
- [x] self consistency with single topology code at end-states
- [x] GradientTest passes
- [ ] Fix `N_ != NR_ + NC_` in interaction group code.
- [x] Add logic for automatically setting up the anchor_idxs
- [ ] update single topology code to allow switching for the smoothcore and the 4d-decoupling method
- [ ] optimize w_interpolation schedule
- [ ] try multiple systems
- [ ] fix water sampling (currently only valid for solvent/vacuum leg where WS is turned off).


maxentile (approved):
Nice! Useful generalization and extension of https://github.com/proteneer/timemachine/pull/415 

Makes sense, but a couple aspects are outside my current intuition:
* [stability worry] Do any forces become unstable as the virtual sites become very close to the anchor site?
    * This appears fine, but some randomized checks (e.g. running MD on random RHFE/RBFE edges with random values of w_i) would provide additional peace of mind before proceeding to the GPU implementation.
* [possible efficiency enhancements] When the group's w_i's approach 1, the collapsed anchor group will look like a single particle that might make stronger-than usual interactions with environment.
    * (consider if the group's |\sum_i q_i| is large..., or, even if |\sum_i q_i|== 0, but there's a large number of atoms in the group, then the collapsed site will appear sort of like a single LJ particle with a large epsilon parameter / a very deep well)
    * In both cases, I suspect it may help with efficiency if an additional parameter is exposed, which scales the interaction energy by some factor <= 1, `factor * U_interaction_group(x)`... (Or, this could be emulated by users of this function, by scaling down the LJ epsilons by some factor like `mean(epsilons) / sum(epsilons)`, etc.)

In the description of https://github.com/proteneer/timemachine/pull/415 , you mentioned this implementation is related to a suggestion from Gilson in the context of docking -- if you happen to have a specific work in mind, capturing that reference in the docstring or PR description would be helpful.

proteneer:
- Improved API a bit so that each row atom takes in an anchoring atom (will expand later on to groups)
- Added tests for interpolation of charges and lennard jones parameters
- Added test for solvated hif2a system with standard pair of hif2a mols

proteneer:
Self-consistency w.r.t. Single Topology code at the end-states achieved on a real system (from hif2a) - GPU implementation tomorrow!

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -54,6 +54,38 @@ void __global__ k_check_rebuild_coords_and_box_gather(
...
+    RealType anchor_z = coords[anchor_idx * COORDS_DIM + 2];
+    RealType rescale =
+        static_cast<RealType>(1.0) - static_cast<RealType>(params[atom_idx * PARAMS_DIM + 3]); // old w_coordinates
+
+    // Coords have 3 dimensions, params have 4
+#pragma unroll COORDS_DIM
+    for (int i = 0; i < COORDS_DIM; i++) {
+        RealType atom_pos = coords[atom_idx * COORDS_DIM + i];
+        RealType anchor_pos = coords[anchor_idx * COORDS_DIM + i];
+        coords_out[atom_idx * COORDS_DIM + i] = anchor_pos + rescale * (atom_pos - anchor_pos);
```
mcwitt, line 85: nit: could this be simplified to

```suggestion
        coords_out[atom_idx * COORDS_DIM + i] = atom_pos + params[atom_idx * PARAMS_DIM + 3] * (anchor_pos - atom_pos);
```

(removing `rescale` definition)?

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -54,6 +54,38 @@ void __global__ k_check_rebuild_coords_and_box_gather(
     }
 }
 
+template <typename RealType, int COORDS_DIM, int PARAMS_DIM>
+void __global__ k_rescale_coords(
+    const int NUM_ROW_ATOMS,
+    const unsigned int *__restrict__ row_atom_idxs,
+    const unsigned int *__restrict__ row_anchor_idxs,
```
mcwitt, line 61: nit: could this just be `NUM_ATOMS`, `atom_idxs`, `anchor_idxs`? (seems orthogonal to concept of "row" atoms)

timemachine/cpp/src/smoothcore_nonbonded_interaction_group.cu
```diff
@@ -0,0 +1,391 @@
...
+        if (unique_row_idxs.find(col_atom_idx) != unique_row_idxs.end()) {
+            throw std::runtime_error("row and col indices must be disjoint");
+        }
+    }
+}
+
+template class SmoothcoreNonbondedInteractionGroup<double>;
+template class SmoothcoreNonbondedInteractionGroup<float>;
+
+} // namespace timemachine
```
mcwitt, line 391: Request: briefly describe what changed here relative to `nonbonded_interaction_group.cu`? Ideally, could create one commit that copies from `nonbonded_interaction_group.cu` and a second commit with the changes.

timemachine/cpp/src/smoothcore_nonbonded_interaction_group.hpp
```diff
@@ -0,0 +1,103 @@
+#pragma once
```
mcwitt, line 1: Would it be desirable to also have a smoothcore implementation of `NonbondedPairListPrecomputed` for intramolecular nonbonded interactions?

timemachine/potentials/nonbonded.py
```diff
@@ -155,6 +159,56 @@ def nonbonded_block(xi, xj, box, params_i, params_j, beta, cutoff):
     return jnp.sum(nonbonded_block_unsummed(xi, xj, box, params_i, params_j, beta, cutoff))
 
 
+import itertools
+
+
+def nonbonded_block_rescaled(xi, xj, box, params_i, params_j, beta, cutoff, anchors_i, groups_i):
```
mcwitt, line 165: nit: is this currently a different interface than is exposed by the GPU potential? (don't immediately see the concept of "groups" there)

timemachine/potentials/nonbonded.py
```diff
@@ -155,6 +159,56 @@ def nonbonded_block(xi, xj, box, params_i, params_j, beta, cutoff):
...
+    params_i = jnp.array(params_i)
+
+    assert len(anchors_i) == len(groups_i)
+
+    # transform xi
+    xi_a = xi[np.repeat(anchors_i, [len(x) for x in groups_i])]
+
+    group_idxs = jnp.array(list(itertools.chain(*groups_i)))
+
+    excluded_idxs = list(set(range(len(xi))).difference(group_idxs.tolist()))
```
mcwitt, line 195: nit: can be optimized to avoid Python loops (OK to leave for later)

timemachine/cpp/src/smoothcore_nonbonded_interaction_group.cu
proteneer: Best way I can think of is to do a manual diff

timemachine/cpp/src/kernels/k_nonbonded.cuh
proteneer: (stale)

timemachine/cpp/src/kernels/k_nonbonded.cuh
proteneer: rescale is removed by the compiler (would prefer to make the rescaling explicit).

timemachine/cpp/src/smoothcore_nonbonded_interaction_group.hpp
proteneer: yes - but this is tricky. We'd need to do this *only* for all interactions outside of a dummy group, this can be all done in python though.

timemachine/potentials/nonbonded.py
proteneer: stale

timemachine/potentials/nonbonded.py
proteneer: stale - added a standard API and wrapped inside `potentials.py`

proteneer:
> * [stability worry] Do any forces become unstable as the virtual sites become very close to the anchor site?

Checking the gradients for some of these computations - I don't think so. The exact expression for the gradient computation would only ever *scale down* the forces, not *scale up*. 

>   * (consider if the group's |\sum_i q_i| is large..., or, even if |\sum_i q_i|== 0, but there's a large number of atoms in the group, then the collapsed site will appear sort of like a single LJ particle with a large epsilon parameter / a very deep well)

Correct, there's two to-be-tuned parameters describing the strength near the `(0, 1)` lambda boundary.
```
MIN_SMOOTHCORE_SIG_HALF = 0.05
MIN_SMOOTHCORE_EPS_SQRT = 0.10
```
>   * In both cases, I suspect it may help with efficiency if an additional parameter is exposed, which scales the interaction energy by some factor <= 1, `factor * U_interaction_group(x)`... (Or, this could be emulated by users of this function, by scaling down the LJ epsilons by some factor like `mean(epsilons) / sum(epsilons)`, etc.)

As noted, same effect can be achieved by scaling down eps/qs via parameter interpolation

> In the description of #415 , you mentioned this implementation is related to a suggestion from Gilson in the context of docking -- if you happen to have a specific work in mind, capturing that reference in the docstring or PR description would be helpful.

I need to dig this up - having a tough time finding it right now.

proteneer:
Updated description.

proteneer:
Rebased onto master.

PR #1383, opened 2024-09-17
[WIP] Add InteractionGroupTraj

maxentile:
support `Us(params) = [U_ig(x; params) for x in traj]`


tests/test_interaction_group_traj.py
```diff
@@ -0,0 +1,40 @@
+import numpy as np
+from jax import grad, jit
+from jax.scipy.special import logsumexp
+
+from timemachine.fe.interaction_group_traj import InteractionGroupTraj, nb_pair_fxn
+
+
+def test_ig_traj():
```
badisa, line 8: nit: Would be good to mark this as `pytest.mark.nocuda` and to make the name of the test more explicit

timemachine/fe/interaction_group_traj.py
```diff
@@ -0,0 +1,113 @@
...
+
+    return vmap(within_cutoff)(x_env)
+
+
+class InteractionGroupTraj:
+    def __init__(self, pair_fxn: PairFxn, xs: Array, box_diags: Array, ligand_idxs: Array, env_idxs: Array, cutoff=1.2):
+        r"""support [U_ig(x; params) for x in traj]
+
+        where U_ig = \sum_i \sum_j pair_fxn(||x_j - x_j||; params_i, params_j)
+
```
jkausrelay, line 61: Add docs with the expected array shapes of the inputs

timemachine/fe/interaction_group_traj.py
```diff
@@ -0,0 +1,113 @@
...
+        self.n_frames = len(xs)
+        self.pair_fxn = pair_fxn
+        self.ligand_idxs = ligand_idxs
+        self.all_env_idxs = env_idxs
+        num_lig, num_env = len(ligand_idxs), len(env_idxs)
+
+        # apply pair_fxn(x_i, x_j, param_i, param_j, box)
+        axes_a = (0, None, 0, None, None)
+        axes_b = (None, 0, None, 0, None)
+        self.all_pairs_fxn = vmap(vmap(pair_fxn, axes_a), axes_b)
```
jkausrelay, line 89: Can you modify compute_Us to take in the pair_fxn and set up all_pairs_fxn there? This would allow for more flexibility (i.e. if we want to compute q or LJ on the same traj).

timemachine/fe/interaction_group_traj.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # TODO[symmetry]:
+        #   maybe generalize so we select a subset of ligand atoms too...
+        #   (currently: assume we always look at all ligand atoms)
+        self.xs_lig = xs[:, ligand_idxs]
+        _xs_env = xs[:, env_idxs]  # will select subsets...
+
+        print(f"precomputing neighborlist on ({num_lig}, {num_env}) interaction group, at cutoff={cutoff}")
```
jkausrelay, line 97: Add a verbose flag to optionally turn off the debug prints

timemachine/fe/interaction_group_traj.py
```diff
@@ -0,0 +1,113 @@
...
+        return jnp.where(d2_ij <= cutoff**2, d2_ij, jnp.inf)
+
+    def within_cutoff(point):
+        return jnp.any(d2_others(point, x_lig) < cutoff**2)
+
+    return vmap(within_cutoff)(x_env)
+
+
+class InteractionGroupTraj:
+    def __init__(self, pair_fxn: PairFxn, xs: Array, box_diags: Array, ligand_idxs: Array, env_idxs: Array, cutoff=1.2):
```
jkausrelay, line 57: I think it'd also be useful to have a serialize (to dict) and deserialize classmethod. This will let us store the state in an npz (instead of needing to pickle the object).

timemachine/fe/interaction_group_traj.py
```diff
@@ -0,0 +1,187 @@
...
+            """
+            print(msg)
+
+        idxs_within_env_block = np.argsort(mask, axis=1)[:, -padded_num_env_atoms:]
+        self.selected_env_idxs = jnp.array(all_env_idxs[idxs_within_env_block], dtype=jnp.uint32)
+
+        self.xs_env = np.array([_x_env[idxs] for (_x_env, idxs) in zip(_xs_env, idxs_within_env_block)])
+        self.box_diags = box_diags
+
+    def to_npz(self, fname):
```
jkausrelay, line 120: to/from a dictionary would be more flexible

maxentile:
rebasing on https://github.com/proteneer/timemachine/pull/1386

timemachine/fe/interaction_group_traj.py
maxentile: Added in https://github.com/proteneer/timemachine/pull/1383/commits/e5233faab2fb9144808ec0b0f871caca695d96bd

timemachine/fe/interaction_group_traj.py
maxentile: Good idea : and it also simplifies serialization (since now only arrays and floats (not jax functions) appear as instance attributes). 

Modified from `_ = traj.compute_Us(params)` to `compute_Us = traj.make_U_fxn(pair_fxn); _ = compute_Us(params)` in https://github.com/proteneer/timemachine/pull/1383/commits/8f5c064b93b7100dc0a82b071cdcf53a884de70a

timemachine/fe/interaction_group_traj.py
```diff
@@ -0,0 +1,187 @@
...
+    @classmethod
+    def from_npz(cls, fname):
+        npz_archive = np.load(fname, allow_pickle=False)
+
+        traj = cls.__new__(cls)
+        traj.xs_lig = npz_archive["xs_lig"]
+        traj.xs_env = npz_archive["xs_env"]
+        traj.box_diags = npz_archive["box_diags"]
+        traj.cutoff = npz_archive["cutoff"]
+        traj.selected_env_idxs = jnp.array(npz_archive["selected_env_idxs"])  # TODO: gross
```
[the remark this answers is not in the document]
maxentile: Yes, but I pushed the np -> jnp conversion of this array into the relevant function, for simpicity, in https://github.com/proteneer/timemachine/pull/1383/commits/1549e91777e418df6cd9da5e5b499b003e15ec14

timemachine/fe/interaction_group_traj.py
maxentile: Good idea : added in https://github.com/proteneer/timemachine/pull/1383/commits/c19938c3db299a4b3feee7e67131f45cf51660fa

(to_npz / from_npz become short functions in terms of to_dict / from_dict)

timemachine/fe/interaction_group_traj.py
```diff
@@ -0,0 +1,193 @@
...
+
+        Returns
+        -------
+        vector-valued function of nb params
+        """
+
+        # apply pair_fxn(x_i, x_j, param_i, param_j, box)
+        axes_a = (0, None, 0, None, None)
+        axes_b = (None, 0, None, 0, None)
+        all_pairs_fxn = vmap(vmap(pair_fxn, axes_a), axes_b)
```
jkausrelay, line 167: Is vmap here particularly expensive? i.e. why not just do this in compute_Us(pair_fxn, nb_params)?

tests/test_interaction_group_traj.py
```diff
@@ -0,0 +1,56 @@
...
+
+    xs = np.random.rand(n_frames, n_env + n_lig, 3) * box_size
+
+    env_idxs = np.arange(n_env)
+    lig_idxs = np.arange(n_lig) + n_env
+    xs[:, lig_idxs] /= box_size
+
+    box_diags = np.ones((n_frames, 3)) * box_size
+
+    traj = InteractionGroupTraj(xs, box_diags, lig_idxs, env_idxs)
```
jkausrelay, line 31: Add a test to validate against a reference implementation: (nonbonded.nonbonded_interaction_groups)

timemachine/fe/interaction_group_traj.py
```diff
@@ -0,0 +1,193 @@
...
+        box_diags : positive float array of box diagonals [T, 3]
+        ligand_idxs: int array (subset of arange(N))
+        env_idxs : int array (subset of arange(N))
+        cutoff : float
+        verbose: bool
+
+        Notes
+        -----
+        * Constructor precomputes neighborlist using a naive jax implementation, which is expensive, but done once
+            TODO: extract from GPU neighborlist
```
jkausrelay, line 76: May not want this since we can run this on cpu only instances.

timemachine/fe/interaction_group_traj.py
```diff
@@ -0,0 +1,193 @@
...
+        """
+        self.cutoff = cutoff
+        self.n_frames = len(xs)
+        self.ligand_idxs = ligand_idxs
+        all_env_idxs = env_idxs
+        num_lig, num_env = len(ligand_idxs), len(env_idxs)
+
+        # TODO[symmetry]:
+        #   maybe generalize so we select a subset of ligand atoms too...
+        #   (currently: assume we always look at all ligand atoms)
```
jkausrelay, line 87: I'm not really sure what the application would be of looking a subset of ligand atoms.

timemachine/fe/interaction_group_traj.py
```diff
@@ -0,0 +1,193 @@
...
+            xs_env=np.array(self.xs_env),
+            box_diags=np.array(self.box_diags),
+            cutoff=self.cutoff,
+            selected_env_idxs=np.array(self.selected_env_idxs),
+            ligand_idxs=np.array(self.ligand_idxs),
+        )
+
+    @classmethod
+    def from_dict(cls, dict_of_np_arrays):
+        archive = dict_of_np_arrays
```
jkausrelay, line 132: nit: why not just call the arg archive?

timemachine/fe/interaction_group_traj.py
maxentile: Not expensive, just style preference (define once, invoke >once).

Since this definition doesn't depend on `params`, this seems like a natural place for this definition, and it keeps a few lines of clutter out of the body of `compute_Us(params)`.

timemachine/fe/interaction_group_traj.py
maxentile: Good point, but I will leave the reminder here for now.

(For example, it should be more efficient for the neighborlist idxs to be exported right after the trajectory is simulated (on the same GPU instance that performs the simulation), even if subsequent analysis steps can be performed at a tolerable speed using Jax's CPU backend. Also, the performance of the all-pairs Jax implementation probably becomes intolerable at a much lower max num_env_atoms than the neighborlist.)

timemachine/fe/interaction_group_traj.py
maxentile: I think it makes sense to be asymmetric as implemented (all of ligand, subset of env), but it could foreseeably also be useful to be symmetric (subset of group A, subset of group B)

timemachine/fe/interaction_group_traj.py
maxentile: Wanted a descriptive hint in signature but didn't want to repeat a clunky name num_fields times in the subsequent lines.

Can rename since the hint doesn't actually add a lot

tests/test_interaction_group_traj.py
maxentile: Added in https://github.com/proteneer/timemachine/pull/1383/commits/1797834e22798f55183bb7fcdee293555a0bf14b

jkausrelay (approved):

PR #1385, opened 2024-09-19
[wip] Offline overlap-based protocol rebalancing

maxentile:
Currently TM has a few methods for adapting the number and spacing of lambda windows along a fixed path, depending on the setting:
* greedy bisection : starts with no samples between 0<lam<1, finds a sequence of lams such that overlap(lams[i], lams[i+1]) > threshold. ( https://github.com/proteneer/timemachine/pull/959 )
* adaptive smc : starts with no samples at lam>0, finds a sequence of lams such that ~ chi2_div(lams[i], lams[i+1]) ~= threshold ( https://github.com/proteneer/timemachine/pull/1177 )
* offline "rebalancing" : starts with samples that allow reweighting to any lam between 0 and 1, finds a sequence of lams such that heuristic_dist(lams[i], lams[i+1]) ~= threshold ( https://github.com/proteneer/timemachine/pull/538 )

This PR will extend the offline "rebalancing" functionality, with the intent of compressing the output of greedy bisection. (Motivation: bisection guarantees that the estimated overlap(lams[i], lams[i+1]) > threshold, but does not minimize len(lams).)

PR #1387, opened 2024-09-20
Protocol overlap utils

maxentile:
Splitting out utilities from https://github.com/proteneer/timemachine/pull/1385 , omitting the optimization method, but retaining the utilities for producing an approximate function `overlap(lam_a, lam_b) -> scalar` from a `bisection_result`

(see commit https://github.com/proteneer/timemachine/commit/c7faca718fa1cd939b01b36dc264800a2b47c421 for detailed omission)


timemachine/optimize/protocol.py
```diff
@@ -208,6 +219,56 @@ def max_work_stddev_distance(prev_lam, next_lam, max_step=0.25):
...
+    log_mean_f = logsumexp(log_f_n - jnp.log(sum(mask)))
+    overlap = jnp.exp(log_mean_f)
+    return 2 * overlap  # so that it goes to 1.0 when u_n_A == u_n_B
+
+
+def make_approx_overlap_distance_fxn(lambdas, u_kn, f_k, N_k):
+    """make a distance function d(a,b) = 1 - overlap(a,b)
+
+    where overlap(a,b) uses fast approximations: reweighting, based on linear interpolation of energies"""
+    u_lam = linear_u_kn_interpolant(lambdas, np.nan_to_num(u_kn, nan=np.inf))
```
maxentile, line 259: On the fence: might want to make this optimization optional

timemachine/fe/free_energy.py
```diff
@@ -267,6 +274,38 @@ def frames(self) -> List[StoredArrays]:
     def boxes(self) -> List[NDArray]:
         return [np.array(traj.boxes) for traj in self.trajectories]
 
+    def compute_u_kn(self) -> Tuple[NDArray, NDArray]:
```
maxentile, line 277: On the fence : might want to merge this function with https://github.com/proteneer/timemachine/blob/60d823020dfb0c7f0bbe4b25a4b9b8a62c9115cd/timemachine/fe/free_energy.py#L1065-L1073 , but may need to generalize signature. (in other function, we have a length-K container of permuted (x,box) states in `HREX[CoordsVelBox]`, here we have a length-K container of (xs, boxes) trajectories, ...)

timemachine/fe/free_energy.py
```diff
@@ -163,6 +163,13 @@ def __post_init__(self):
         assert self.ligand_idxs.dtype == np.int32 or self.ligand_idxs.dtype == np.int64
         assert self.protein_idxs.dtype == np.int32 or self.protein_idxs.dtype == np.int64
 
+    def to_bound_impl(self, precision=np.float32):
+        bps = self.potentials
+        summed_pot = SummedPotential([bp.potential for bp in bps], [bp.params for bp in bps])
+        params = np.concatenate([bp.params.reshape(-1) for bp in bps])
+        impl = summed_pot.to_gpu(precision).bind(params).bound_impl
+        return impl
```
mcwitt, line 171: nit: can be simplified using `make_summed_potential` from `timemachine.potentials`:

```suggestion
        return make_summed_potential(self.potentials).bound_impl
```

maxentile: Ahh, overlooked that -- simplified in https://github.com/proteneer/timemachine/pull/1387/commits/3ef36610d6cb8b51edb3a53eb46769435c8961a2

tests/test_protocol_optimization.py
```diff
@@ -121,3 +128,109 @@ def simulate_protocol(lambdas_k, n_samples_per_window=100, seed=None):
...
+    print("initial protocol (from bisection): ")
+    _ = summarize_protocol(lambdas, overlap_dist)
+    target_overlap_levels = [0.05, 0.1, 0.2, 0.4, 0.666]
+    protocol_lengths = []
+    for target_overlap in target_overlap_levels:
+        target_dist = 1 - target_overlap
+        print(f"optimized with target dist = {target_dist} (target overlap = {target_overlap}):")
+        greedy_prot = greedily_optimize_protocol(overlap_dist, target_dist)
+        greedy_nbr_dist = summarize_protocol(greedy_prot, overlap_dist)
+        protocol_length = len(greedy_prot)
```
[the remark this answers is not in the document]
maxentile: Nice -- numbers look consistent with https://github.com/proteneer/timemachine/pull/1385#discussion_r1768799074

tests/test_protocol_optimization.py
```diff
@@ -121,3 +128,109 @@ def simulate_protocol(lambdas_k, n_samples_per_window=100, seed=None):
...
+        # assert x < y < z implies d(x, y) <= d(x, z)
+        eps = 1e-3
+        np.testing.assert_array_less(distances_to_sorted_larger_lams[:-1], distances_to_sorted_larger_lams[1:] + eps)
+
+        # assert symmetric
+        _distances_flipped = np.array([overlap_dist(lam_j, lam_i) for lam_j in lams_above])
+        np.testing.assert_allclose(distances_to_sorted_larger_lams, _distances_flipped)
+
+
+# @pytest.mark.nightly(reason="Slow")  # TODO: restore pytest.mark.nightly before merging
```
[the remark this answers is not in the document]
maxentile: Might be able to reuse a bisection result / fixture from another test

tests/test_protocol_optimization.py
```diff
@@ -121,3 +128,109 @@ def simulate_protocol(lambdas_k, n_samples_per_window=100, seed=None):
...
+def test_greedy_overlap_on_st_vacuum():
+    # get bisection result
+    mol_a, mol_b, core = get_hif2a_ligand_pair_single_topology()
+    ff = Forcefield.load_default()
+    bisection_result = estimate_relative_free_energy_bisection(
+        mol_a, mol_b, core, ff, None, n_windows=48, min_overlap=2.0 / 3
```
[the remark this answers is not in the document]
maxentile: Thanks -- can you clarify if you are requesting a change?

timemachine/fe/free_energy.py
```diff
@@ -268,6 +278,38 @@ def frames(self) -> List[StoredArrays]:
     def boxes(self) -> List[NDArray]:
         return [np.array(traj.boxes) for traj in self.trajectories]
 
+    def compute_u_kn(self) -> Tuple[NDArray, NDArray]:
+        """get MBAR input matrices u_kn and N_k"""
+
+        # TODO: there's probably a much faster idiom for this -- shouldn't take a whole minute...
```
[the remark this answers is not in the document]
maxentile: Didn't think to suspect that `len` of these objects was slow, but good to confirm. Comment was indeed about the function overall, which is inefficient and can take up to a minute.

tests/test_protocol_optimization.py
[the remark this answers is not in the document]
proteneer: no need to change - just a nit.

timemachine/optimize/protocol.py
proteneer: What are the situations in which this would break down?

proteneer: Reading the docstrings:

```
Given a matrix u_kn[k, n] = u(xs[n], lambdas[k]) produce linear interpolated estimates of u(xs[n], lam) at arbitrary new values lam
```

Our 4D nonbonded softcore implementation and log-linear bond interpolation schemes would probably result in very non-linear energies as a function of lambda.

tests/test_protocol_optimization.py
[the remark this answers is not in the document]
maxentile: Ahh, sorry, had misread the suggestion

(Initially thought the suggestion was `//` -> `/` which didn't seem applicable, then saw that the suggested edit was `2.0` -> `2`. Applied in https://github.com/proteneer/timemachine/pull/1387/commits/4716d0f2bab6751ef0393d140dfa2cb7aaf6eb83 )

timemachine/optimize/protocol.py
proteneer: > This assumption can be reasonable -- even though u is a nonlinear function of lam -- if lambdas[k] is dense enough.

Is there a rule of thumb/ballpark estimate on what "dense enough" and "reasonable" is? I don't know if the initial bisection protocol makes guarantees about amount of error introduced by a linear approximation of the energy. (I might have missed these values originally in #538). 

Can we quantify this error in some way? Some ideas that come to mind are:

For each [solvent, vacuum] using `st`:
1) [unit test] Assert on the maximum energy deviation relative to the reference energies over multiple trajectories (may be too stringent).
2) [unit test] Assert on the `reweighted_pair_overlap` using the linear approximation energy vs the reference energy
3) [integration test] See if we end up with approximately the same # of windows after greedily optimizing the protocol. i.e. do we know that the reduction in # of windows isn't a result of the linear approximation?

tests/test_free_energy.py
```diff
@@ -672,3 +672,24 @@ def modify_nb_cutoff(nb):
...
+    solvent_host_config = HostConfig(solvent_sys, solvent_conf, solvent_box, solvent_conf.shape[0], solvent_top)
+    solvent_host = setup_optimized_host(st, solvent_host_config)
+    initial_state = setup_initial_states(st, solvent_host, DEFAULT_TEMP, [0.5], 2024)[0]
+
+    # minimal test
+    bound_impl = initial_state.to_bound_impl()
+    x, box = initial_state.x0, initial_state.box0
+    du_dx, u = bound_impl.execute(x, box, compute_du_dx=True, compute_u=True)
+    assert np.isfinite(u)
+    assert np.isfinite(du_dx).all()
```
mcwitt, line 695: nit: with `to_bound_impl` implemented using `make_summed_potential`, this feels a bit redundant. Noticing that `make_summed_potential` is untested; maybe the ideal thing to do would be to add a lightweight test for this in `test_potentials.py` (which should also be easier to make a bit stronger and actually verify that the result is the sum of the component terms)

timemachine/optimize/protocol.py
maxentile: These are good test ideas. This is an uncontrolled approximation, and I only spot-checked approximation quality in a cursory way.

The spot-checks I've done so far are most similar to your suggestion (2) -- compare reweighting estimates based on approximate energies against reweighting estimates based on reference energies.

Here's an example of a comparison from internal documentation, using the `get_hif2a_ligand_pair_single_topology` testsystem* (*modified: 10 atoms truncated from the atom mapping).

image

x-axis is `lam`, y-axis is estimated `overlap(lam, 0.25)`

blue dots are from reweighting using linear interpolation of `u_kn`, red dots are from reweighting using the potential energy function.

Will expand this sort of comparison into a test...

> I don't know if the initial bisection protocol makes guarantees about amount of error introduced by a linear approximation of the energy.

I don't think it does.

proteneer: > Here's an example of a comparison from internal documentation, using the `get_hif2a_ligand_pair_single_topology` testsystem* (*modified: 10 atoms truncated from the atom mapping).
> 
> image
> 
> x-axis is `lam`, y-axis is estimated `overlap(lam, 0.25)`
> 
> blue dots are from reweighting using linear interpolation of `u_kn`, red dots are from reweighting using the potential energy function.

Nice! This plot looks re-assuring to me. Clarification: was this done in vacuum? If the plots look similar in solvent/complex it would significantly boost my confidence in this approximation (rough thinking is that the # of interactions and sensitivity of energies to lambda is greater when nonbonded interaction groups are present). 

> Will expand this sort of comparison into a test...

Awesome!

maxentile: >Clarification: was this done in vacuum? If the plots look similar in solvent/complex it would significantly boost my confidence in this approximation (rough thinking is that the # of interactions and sensitivity of energies to lambda is greater when nonbonded interaction groups are present).

This one's vacuum : will expand.

timemachine/fe/free_energy.py
```diff
@@ -268,6 +278,38 @@ def frames(self) -> List[StoredArrays]:
...
+        N_k = [len(traj.frames) for traj in self.trajectories]
+        N, K = sum(N_k), len(N_k)
+        assert len(self.final_result.initial_states) == K
+
+        kBTs = [BOLTZ * state.integrator.temperature for state in self.final_result.initial_states]
+
+        # TODO: validate assumption that all states are in compatible thermodynamic states
```
mcwitt, line 291: Could this just reuse https://github.com/proteneer/timemachine/blob/80e4aae47169688200a802612c48cc54a0ee1010/timemachine/fe/free_energy.py#L1164?
(and also `assert_potentials_compatible`?)

timemachine/fe/free_energy.py
```diff
@@ -268,6 +278,38 @@ def frames(self) -> List[StoredArrays]:
...
+        # pressures = [state.barostat.pressure for state in self.final_result.initial_states]
+
+        u_kn = np.zeros((K, N))
+        for k, initial_state in enumerate(self.final_result.initial_states):
+            kBT = kBTs[k]
+
+            bound_impl = initial_state.to_bound_impl()
```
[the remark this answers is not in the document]
mcwitt: A third option could be `execute_batch` (as noted in a TODO comment below). I think this would be my preferred approach.

timemachine/fe/free_energy.py
```diff
@@ -268,6 +278,41 @@ def frames(self) -> List[StoredArrays]:
...
+            assert_potentials_compatible(s_0.potentials, s.potentials)
+
+        u_kn = np.zeros((K, N))
+        for k, initial_state in enumerate(self.final_result.initial_states):
+            kBT = kBTs[k]
+
+            bound_impl = initial_state.to_bound_impl()
+
+            def u(x, box):
+                # TODO: bound_impl.execute_batch
```
mcwitt, line 304: Could this actually use the unbound `execute_batch`, and batch over both coordinates and parameters? (OK with postponing optimizations like this especially if they require significant refactoring)

timemachine/fe/free_energy.py
mcwitt: hmm, it seems like the common element is the batch evaluation of potentials over (x, box) pairs? Do you imagine `compute_u_kn` could make use of the sparse evaluation strategy that's implemented by `compute_potential_matrix`? Otherwise, it doesn't seem like much is shared (and extra logic that isn't shared, e.g. the translation between replica and state indices in `compute_potential_matrix`)

timemachine/optimize/protocol.py
```diff
@@ -208,6 +219,56 @@ def max_work_stddev_distance(prev_lam, next_lam, max_step=0.25):
     return max_work_stddev_distance
 
 
+# (2) pair overlap
+def reweighted_pair_overlap(u_n_A, u_n_B, u_n_ref):
+    """given arrays [u(x) for x in xs] (for u in {u_A, u_B, u_ref}),
+    estimate overlap(A, B) by reweighting from ref->A and ref->B"""
```
mcwitt, line 225: Could we add a mathematical definition of the pair overlap to the docstring here, or point to where it's defined? I often am confused about the precise definition, whether or not there's a factor of 2, etc.

timemachine/optimize/protocol.py
maxentile: Running a larger spot-check in vacuum with a large number of random `(lam_a, lam_b)` pairs, and comparing the estimated overlap using reference potential energy function vs. linearly interpolated potential energy fxn produces roughly the following comparison:

image

close agreement most of the time (MAE ~= 0.00841), but occasional large disagreement (max disagreement found so far ~= 0.256). also, occasionally out of expected bounds.

maxentile: Sorry for missing detail.

> Are (lam_a, lam_b) always adjacent pairs from the bisected schedule?

No: linear interpolation is exact cases where `lam_a, lam_b` are both in `set(lambdas)`, and there are only a few dozen of those discrete adjacent pairs to check. 

In the above plot, 1000 random `(lam_a, lam_b)` pairs were sampled according to: `lam_a = 0.8 * rand(); lam_b = lam_a + 0.2 * rand()`. Will be curious to figure out what's going on in the cases with the largest estimation error.

proteneer: Thanks for the additional clarification.

Perhaps intentional but the scheme you proposed uniformly samples `lam_a in [0, 0.8]`, but does *not* uniformly sample `lam_b` (ie. there will be very few values of `lam_b > 0.95` tested relative to `lam_a < 0.05`). i.e. this ultimately may not stress test the higher values of lambda as much.

image

Some additional empirical observations of the error plot:

1. In the cases where error is significant (>0.1), the estimated overlap based on interpolated energies seems to consistently greater than the reference energies. If it were the other around that's less of a concern given the use case: i.e. (underestimating the overlap is a more defensive outcome than overestimating the overlap if the ultimate goal is to rebalance the number of windows).

2. There appears to be two additional off-diagonal bands in the upper right quadrant of the plot separated by 0.1 apart. 

3. As you noted, not sure how to interpret out of bounds where `estimated_pair_overlap > 1`.

maxentile: > Perhaps intentional but the scheme you proposed uniformly samples lam_a in [0, 0.8], but does not uniformly sample lam_b

That's true: could adjust the sampling distribution of this spot-check.

(In the spot-check, I sampled a uniform random interval width `width ~ Unif(0, 0.2)`, and a uniform random left bound `lam_a ~ Unif(0, 0.8)`. Could instead do something like sample a uniform `midpoint ~ Unif(width/2, 1 - width/2)`, `lam_a = midpoint - width/2, lam_b = midpoint + width/2`, to be more symmetric.)

maxentile: Investigation is still pending. One additional note: even if the approximation quality is poor for some query points `(lam_a, lam_b)`, the overall context of the optimizer could still be minimally impacted. (i.e. the sequence of decisions made by the optimizer might be about the same with or without this approximation).

Note that the optimizer interacts with this approximation **only by checking the sign** of `overlap(lam_a, lam_b) - target_val`, where `target_val` is maybe 0.1 or 0.25 or 0.5 or 0.666. So, even if this approximation introduces a large absolute error on some query points (e.g. `approx_overlap(a, b) = 1.2` when `overlap(a, b) = 0.95`), the error might be inconsequential to the optimizer.

For illustration, here is the sequence of estimates produced on the sequence of points queried by the optimizer, with a target overlap of 0.25 between adjacent states:

image

Here is a plot of the agreement between the approximation and the reference on this sequence of points:

image

The approximation gets the sign right on ~90% of the points queried in this particular sequence. Although there are queried points where the approximation is very wrong in terms of absolute error, the approximation gets the sign right on those points (i.e. the large errors do not lead to a different optimizer outcome). In the cases where it gets the sign wrong, the absolute approximation error seems small (say, < 0.015).

Still want to figure out what's causing the estimate to sometimes exceed 1 though...

... Anyway, if the approach looks unrecoverable after further investigation, a different approach that might make sense is to try to approximate the whole function `overlap(a,b) -> R`, rather than trying to approximate the function `u(lam) -> R^N` that appears as an intermediate step of the reweighting approximation.

maxentile: > Will be curious to figure out what's going on in the cases with the largest estimation error.

Looking at this again: appears to be due to `f64` -> `f32` precision loss in the step that linearly interpolates `u_kn` (which starts in f64)

> linear interpolation is exact in cases where lam_a, lam_b are both in set(lambdas)

This statement was not true, due to precision loss

timemachine/optimize/protocol.py
maxentile: Have added a literature reference here, and posted additional notes in an internal system.

Specifically about the factor of 2, I have been inconsistent about writing `p_a * p_b / (p_a + p_b)` (and then multiplying by an ad-hoc-looking `2`) vs. `p_a * p_b / (0.5 * (p_a + p_b))` (denominator is an equally weighted mixture of states `a` and `b`).

timemachine/fe/free_energy.py
maxentile: Applied in https://github.com/proteneer/timemachine/pull/1387/commits/978b1da1f6974263f0cfc8ce8b5a06a9293e1a1c  (using `unbound_impl.execute_batch_sparse`), did not require significant refactoring

timemachine/fe/free_energy.py
maxentile: Good call -- applied in https://github.com/proteneer/timemachine/pull/1387/commits/cfbfead6616231a4e1ad636b5f66b60a4e641653

timemachine/optimize/protocol.py
```diff
@@ -47,16 +51,22 @@
...
 from jax import vmap
-from jax.scipy.special import logsumexp
+from jax.scipy.special import logsumexp as _logsumexp
 from jax.typing import ArrayLike
 from scipy.optimize import bisect
 
+from timemachine.fe.reweighting import interpret_as_mixture_potential
+
+# performance bottleneck in several functions below
+logsumexp = jit(_logsumexp, static_argnames=["axis"])
```
mcwitt, line 61: nit: would it be preferable to instead add a `@jax.jit` decorator to `reweighted_pair_overlap`? (This looked like the only usage in this file, but might have missed something.)

timemachine/optimize/protocol.py
```diff
@@ -208,7 +223,142 @@ def max_work_stddev_distance(prev_lam, next_lam, max_step=0.25):
...
+    log_p_ref_n = log_q_ref - logsumexp(log_q_ref - log_N)
+
+    # normalize A and B wrt ref
+    log_p_A = log_q_A - logsumexp(log_q_A - log_p_ref_n - log_N)
+    log_p_B = log_q_B - logsumexp(log_q_B - log_p_ref_n - log_N)
+
+    log_prod_AB_n = log_p_A + log_p_B
+    # assert (log_prod_AB_n < jnp.inf).all()
+
+    log_p_mix_n = logsumexp(jnp.array([log_p_A, log_p_B]), axis=0)
```
mcwitt, line 252: nit/unsure: is it conventional to define this as

```suggestion
    log_p_mix_n = logsumexp(jnp.array([log_p_A, log_p_B]), axis=0) - jnp.log(2)
```

and omit the final factor of 2 on L262?

timemachine/optimize/protocol.py
maxentile: This would be more natural and provide additional opportunities for optimization, but I had hit a snag when I first attempted this. (I did not immediately see why (and haven't investigated), but jit compilation appeared to hang / take an excessive amount of time when `jit` is invoked on the function `reweighted_pair_overlap`. )

And a couple other functions in this file, such as `log_weights_from_mixture` and `reweighted_stddev`, also use this function.

timemachine/optimize/protocol.py
```diff
@@ -208,7 +223,142 @@ def max_work_stddev_distance(prev_lam, next_lam, max_step=0.25):
...
+    -------
+    overlap_fxn: (lam_a, lam_b) -> overlap_estimate
+    """
+
+    def overlap_fxn(lam_a, lam_b):
+        estimate = reweighted_pair_overlap(u_lam(lam_a), u_lam(lam_b), src_u_n)
+        clamped = jnp.clip(estimate, 0.0, 1.0)  # TODO: revert this line if estimates > 1 are avoided
+        return clamped
+
+    return overlap_fxn
```
mcwitt, line 293: Nice! I like how this is defined so that the user can provide $u(\lambda$ (e.g. if the linear approximation is insufficient?), but using a common code path.

tests/test_protocol_optimization.py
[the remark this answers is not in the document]
mcwitt: (bump) - should we uncomment the nightly mark before merging?

timemachine/fe/free_energy.py
```diff
@@ -1120,6 +1136,91 @@ def compute_dense():
...
+def make_u_kl_fxn(trajs, initial_states):
+    """fxn(k, l) = "trajs[k] evaluated in ensembles[l]"
+
+    usage note: be careful of axis-ordering convention, see: https://github.com/proteneer/timemachine/issues/1100
+    """
+
+    summed_potentials = [make_summed_potential(s.potentials) for s in initial_states]
+
+    # validate assumption that initial states all have compatible potentials / ensembles
+    kBTs = [BOLTZ * state.integrator.temperature for state in initial_states]
```
mcwitt, line 1148: nit: could we just define a single `kBT = s_0.integrator.temperature` (below the assertions that immediately follow)?

mcwitt: (same for `summed_potentials`)

timemachine/optimize/protocol.py
maxentile: Exactly -- if the user is concerned about the quality of the default recommended approximation (linearly interpolate axis 0 of  pre-computed `u_kn` given `lambda_k`), the user can supply a different implementation.

I shied away from supplying these implementations in the current PR, but the case of single topology intermediates in vacuum, for example, might look like:

```python
def make_reference_u_lam_vacuum(st: SingleTopology, trajs: List[Trajectory]):
	xs = np.vstack([traj.frames for traj in trajs])
	kBT = BOLTZ * temperature

	def ref_u_lam(lam):
		U_lam_fxn = st.setup_intermediate_state(lam).get_U_fn()
		return vmap(U_lam_fxn)(xs) / kBT
	
	return ref_u_lam
```

Perhaps there is an appropriate place to define such functions for all common cases (`SingleTopology` in {vacuum, solvent, complex}, absolute in solvent, ...), but initially seems best to leave that as a detail that can be optionally supplied by a user.

timemachine/fe/free_energy.py
```diff
@@ -1120,6 +1136,91 @@ def compute_dense():
...
+
+    def batch_U_fxn(xs, ps, bs, x_idxs, p_idxs):
+        Us = sp_gpu.unbound_impl.execute_batch_sparse(xs, ps, bs, x_idxs, p_idxs, False, False, True)[2]
+        return np.nan_to_num(Us, nan=+np.inf)
+
+    def u_kl(k, l):
+        coords = trajs[k].frames
+        boxes = trajs[k].boxes
+
+        params = np.array([summed_potentials[l].params])
```
mcwitt, line 1165: nit: `np.asarray` to avoid copy

timemachine/fe/free_energy.py
```diff
@@ -1120,6 +1136,91 @@ def compute_dense():
     return U_kl
 
 
+def make_u_kl_fxn(trajs, initial_states):
+    """fxn(k, l) = "trajs[k] evaluated in ensembles[l]"
```
mcwitt, line 1140: If I'm interpreting this correctly, the returned function has signature `(sample state index, evaluation state index) -> energies, where energies is an array of the energy at each frame of the trajectory? Might clarify this with a hint in the docstring or naming change. (Was initially confused that the return type was not a be a scalar, given the name of the function)

timemachine/fe/free_energy.py
```diff
@@ -1120,6 +1136,91 @@ def compute_dense():
...
+        # also, assert barostat and integrator are self-consistent
+        assert intg_a.temperature == baro_a.temperature
+
+    else:
+        # assert (A, B) are compatible NVT ensembles
+        assert (state_a.box0 == state_b.box0).all()
+
+
+def compute_u_kn(trajs, initial_states) -> Tuple[NDArray, NDArray]:
+    """makes K^2 calls to execute_batch_selective"""
```
mcwitt, line 1208: nit: should this be
```suggestion
    """makes K^2 calls to execute_batch_sparse"""
```
?

(where each call to `execute_batch_sparse` is doing `trajs.shape[1]` calls to `execute_batch_selective`?)

timemachine/fe/free_energy.py
maxentile: Correct -- this computes a `len(trajs[k])` slice of the `u_kln` matrix (`u_kln.shape == (K, K, max(N_k))`). Intent is to avoid possible memory issues if the number of trajectories is large. (Each invocation of the resulting function processes only a single trajectory.)

timemachine/fe/free_energy.py
```diff
@@ -1120,6 +1136,91 @@ def compute_dense():
...
+    def u_kl(k, l):
+        coords = trajs[k].frames
+        boxes = trajs[k].boxes
+
+        params = np.array([summed_potentials[l].params])
+
+        coords_batch_idxs = np.arange(len(coords)).astype(np.uint32)
+        params_batch_idxs = np.zeros_like(coords_batch_idxs).astype(np.uint32)
+
+        Us = batch_U_fxn(coords, params, boxes, coords_batch_idxs, params_batch_idxs)
```
mcwitt, line 1170: nit: this doesn't seem like it's using the full generality of `execute_batch_sparse` -- would `execute_batch` suffice in the definition of `batch_U_fxn` for clarity?

mcwitt (approved):
Looks great; I'm excited to start using this! Only minor comments.

timemachine/optimize/protocol.py
mcwitt: Thanks. I didn't see this before I posted https://github.com/proteneer/timemachine/pull/1387#discussion_r1809225572; feel free to ignore that.

timemachine/fe/free_energy.py
maxentile: Weakly held style preference to leave this as-is unless there's a performance concern.

(also, if we later relax the definition of `assert_ensembles_compatible(a, b)` to allow  `a` and `b` to have different temperatures, this won't need to be updated)

timemachine/fe/free_energy.py
maxentile: Will leave this as-is for simplicity (don't have to think about possibility of accidental in-place mutation of any `summed_potential`), unless the copy is costly

timemachine/fe/free_energy.py
maxentile: Good catch -- fixed in https://github.com/proteneer/timemachine/pull/1387/commits/3523176d4efa5ed88ca1691160a437182bd28941

timemachine/fe/free_energy.py
mcwitt: Makes sense; I'd misunderstood the intent as ensuring that `params` is a numpy array

timemachine/fe/free_energy.py
maxentile: Either `execute_batch` or `execute_batch_sparse` I think would suffice, but the `execute_batch_sparse` interface is slightly more explicit

timemachine/fe/free_energy.py
maxentile: That is the main intent, and another intent is to make it have shape `(1, len(summed_potentials[l].params))`

timemachine/fe/free_energy.py
maxentile: >was just initially confused by the implied generality

The implementation (`u_kn = u_k(x_n) / kBTs[k]`) is I think compatible with `len(set(kBTs)) > 1`, just the assertion `assert_ensembles_compatible` currently enforces `len(set(kBTs)) == 1`.

(For readability, we could add an assertion `assert len(set(kBTs)) == 1` or a comment to that effect, but I'd be inclined to leave the logic as-is.)

mcwitt: (I guess instantiating the identical summed potentials does duplicate the system parameters, which could be a nontrivial chunk of memory for very large systems, but still probably not a concern)

timemachine/fe/free_energy.py
mcwitt: OK, this is only a minor stylistic preference (i.e. using the least powerful abstraction available), but can see either argument, so not a blocker.

timemachine/fe/free_energy.py
maxentile: Ahh, that's a good point about the needless storage of the otherwise identical summed potentials, when we could just store `sp.params for sp in summed_potentials` -- didn't check how much additional space is occupied by a summed potential instance besides its parameter array, but have modified this to store an array of params rather than a list of summed_potentials in https://github.com/proteneer/timemachine/pull/1387/commits/4f5c0fc9e3a3f1591077dfcca6d59bdfd9bd07e4

maxentile: > didn't check how much additional space is occupied by a summed potential instance besides its parameter array

Quick check: `len(pickle.dumps(sp)) / len(pickle.dumps(sp.params))` is maybe around 2-3x on a couple vacuum and solvent examples

timemachine/optimize/protocol.py
maxentile: I think the suggestion is cleaner -- applied in https://github.com/proteneer/timemachine/pull/1387/commits/1fb110a82c33093dd1f529660ef01d426c7fc416

proteneer (approved):

PR #1388, opened 2024-09-20
Update jax to 0.4.28 to fix memory corruption issue

jkausrelay:
See https://github.com/jax-ml/jax/releases/tag/jaxlib-v0.4.28


Dockerfile
```diff
@@ -40,7 +40,8 @@ ARG ENV_NAME=timemachine
 RUN . /opt/conda/etc/profile.d/conda.sh && \
     conda env create -n "${ENV_NAME}" --force -f environment.yml && \
     conda clean -a && \
-    conda activate ${ENV_NAME}
+    conda activate ${ENV_NAME} && \
+    conda list -n ${ENV_NAME}
```
badisa, line 44: Is this necessary?

tests/test_protocol_optimization.py
```diff
@@ -113,8 +115,9 @@ def poorly_spaced_path(lam):
 def simulate_protocol(lambdas_k, n_samples_per_window=100, seed=None):
     """Generate samples from each lambda window, plug into MBAR"""
     O_k, K_k = poorly_spaced_path(lambdas_k)
-    testsystem = HarmonicOscillatorsTestCase(O_k, K_k)
-    N_k = [n_samples_per_window] * len(O_k)
-    xs, u_kn, N_k, s_n = testsystem.sample(N_k, seed=seed)
+    with patch("numpy.int", int):
```
badisa, line 118: Not for this PR: We should revisit upgrading pymbar, do we know if the previous issue we had with the newest versions were fixed?

badisa (approved):

Dockerfile
jkausrelay: removed, was just to debug versions

tests/test_protocol_optimization.py
jkausrelay: I'm not sure

PR #1391, opened 2024-09-23
Clean up HREX, BDExchangeMover and TIBDExchangeMover

badisa:
* Fixes a potential bug in the BDExchangeMover that doesn't pass the stream to a kernel. Not seen in practice because we use the default stream as well as the fact that we synchronize the stream immediately before calling the kernel in question. Good to fix regardless
* Fixes an HREX test that was no longer patching the correct function as of #1334, added assertions to avoid missing again
* Removes duplicate equilibration logic in `run_sims_hrex` to avoid code becoming stale. 
* Avoids setting the same seed for TIBDExchangeMover in `get_context` across initial states. Use the integrator seed, which is distinct to each InitialState.

### DIFF
--- a/timemachine/fe/free_energy.py
+++ b/timemachine/fe/free_energy.py
@@ -538,8 +538,8 @@ def get_context(initial_state: InitialState, md_params: Optional[MDParams] = Non
 
         water_params = get_water_sampler_params(initial_state)
 
-        # Generate a new random seed based on the md params seed
-        rng = np.random.default_rng(md_params.seed)
+        # Generate a new random seed based on the integrator seed, MDParams seed is constant across states
+        rng = np.random.default_rng(initial_state.integrator.seed)
         water_sampler_seed = rng.integers(np.iinfo(np.int32).max)
 
         water_sampler = custom_ops.TIBDExchangeMove_f32(
--- a/timemachine/fe/free_energy.py
+++ b/timemachine/fe/free_energy.py
@@ -1221,54 +1221,13 @@ def run_sims_hrex(
         # Add an identity move to the mixture to ensure aperiodicity
         neighbor_pairs = [(StateIdx(0), StateIdx(0))] + neighbor_pairs
 
-    # Setup the barostat to run more often for equilibration
-    equil_barostat_interval = 15
     barostat = context.get_barostat()
-    if barostat is not None:
-        barostat.set_interval(equil_barostat_interval)
 
-    def get_equilibrated_xvb(xvb: CoordsVelBox, state_idx: StateIdx) -> CoordsVelBox:
-        # Set up parameters of the water sampler movers
-        if md_params.water_sampling_params is not None:
-            for mover in context.get_movers():
-                if isinstance(mover, WATER_SAMPLER_MOVERS):
-                    assert water_params_by_state is not None
-                    mover.set_params(water_params_by_state[state_idx])
-        if md_params.n_eq_steps > 0:
-            # Set the movers to 0 to ensure they all equilibrate the same way
-            # Ensure initial mover state is consistent across replicas
-            for mover in context.get_movers():
-                mover.set_step(0)
-
-            params = params_by_state[state_idx]
-            bound_potentials[0].set_params(params)
-
-            context.set_x_t(xvb.coords)
-            context.set_v_t(xvb.velocities)
-            context.set_box(xvb.box)
-
-            xs, boxes = context.multiple_steps(md_params.n_eq_steps, store_x_interval=0)
-            assert len(xs) == 1
-            x0 = xs[0]
-            v0 = context.get_v_t()
-            box0 = boxes[0]
-            xvb = CoordsVelBox(x0, v0, box0)
-        return xvb
-
-    initial_replicas = [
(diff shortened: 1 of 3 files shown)


timemachine/fe/free_energy.py
```diff
@@ -1221,55 +1221,14 @@ def get_flattened_params(initial_state: InitialState) -> NDArray:
...
-            box0 = boxes[0]
-            xvb = CoordsVelBox(x0, v0, box0)
-        return xvb
-
-    initial_replicas = [
-        get_equilibrated_xvb(CoordsVelBox(s.x0, s.v0, s.box0), StateIdx(i)) for i, s in enumerate(initial_states)
-    ]
-
-    hrex = HREX.from_replicas(initial_replicas)
+    hrex = HREX.from_replicas([CoordsVelBox(s.x0, s.v0, s.box0) for s in initial_states])
```
badisa, line 1226: Could revert this as it does corrupt the timings estimate, but since we typically call into `estimate_relative_free_energy_bisection_hrex_impl` which does all the equilibration in the bisection stage it doesn't impact our hot path.

Mostly wanted to avoid having two duplicate implementations of equilibration.

badisa: > I'm actually not sure why we added equilibration here to begin with. (Do you recall?) 

I am not sure either. Though I vaguely remember `sample` requiring `n_eq_steps > 0` at one point, so maybe to get around that?

>  the other function with a similar interface accepting a list of initial states, run_sims_sequential, also does not do equilibration

To be precise `run_sims_sequential` does do equilibration, if you pass `md_params.n_eq_steps > 0` by calling `sample` (which eventually calls into https://github.com/proteneer/timemachine/blob/master/timemachine/fe/free_energy.py#L601-L613) but does not separate it out.

mcwitt: > To be precise run_sims_sequential does do equilibration

Got it, I was confused before. And I was forgetting that the initial states returned from the bisection phase are not equilibrated (only pre-equilibrated, I think?). So equilibration does seem necessary here. The question is whether to have this duplicated code, or do the equilibration as part of the first frame? Is there a third option where we pull out a generic function like `get_equilibrated_xvb` and use it in both places?

timemachine/fe/free_energy.py
```diff
@@ -1298,7 +1257,13 @@ def sample_replica(xvb: CoordsVelBox, state_idx: StateIdx) -> Tuple[NDArray, NDA
                 # Set the step so that all windows have the movers be called the same number of times.
                 mover.set_step(current_step)
 
-            md_params_replica = replace(md_params, n_frames=1, n_eq_steps=0, seed=state_idx + current_frame)
+            md_params_replica = replace(
+                md_params,
+                n_frames=1,
+                # Run equilibration as part of the first frame
+                n_eq_steps=md_params.n_eq_steps if current_frame == 0 else 0,
```
mcwitt, line 1264: This does seem a bit confusing and tech-debt-y. Is there a way to avoid duplicated code while also avoiding the need to hack the MDParams like this?

timemachine/fe/free_energy.py
badisa: > And I was forgetting that the initial states returned from the bisection phase are not equilibrated (only pre-equilibrated, I think?)

They are equilibrated as bisection calls `sample` which will perform the equilibration before collecting the samples for bisection, we also update the initial states according: https://github.com/proteneer/timemachine/blob/master/timemachine/fe/rbfe.py#L666-L714. We have incorporated our equilibration behavior into the family of `sample` functions, so think that is right.

mcwitt: Ah, OK, I forgot that we do update the bisection initial states with the coordinates from the final frame. So maybe the reason for equilibration here was just for consistency with the other `run_sims_*` functions, even though additional MD equilibration seems unnecessary in our current RBFE implementation.

timemachine/fe/free_energy.py
badisa: I find this to be less tech-debt than where we currently split out equilibration where it could deviate from our reference version of equilibration (at least I interpret the code to be in `sample` to be the definitive version). Could make this an if/else to make it more readable, but nothing else immediately jumps out.

Guess you could also run `sample_context_with_iter` on all of the frames upfront with `n_frames == 0`, but notably different to what this is doing.

mcwitt: Agreed that duplication is the least desirable option. Would it make sense to instead pull the reference equilibration code out of `sample_with_iter` into a separate function, and call this function both from `sample_with_iter` and `run_sims_hrex`?

badisa: We could do that, though I am not clear what this is resolving beyond readability.

We are already modifying the MDParams on each state, dependent on the current frame. And if we end up parallelizing this function we will have to parallelize two call sites rather than one, which seems like a more serious bit of tech debt. Though maybe should be less focused on the latter for this PR, but these changes were spawned by my parallelization attempts.

mcwitt: No reason other than readability/subjective preference for "decoupling". Not blocking, feel free to disregard

mcwitt (approved):

PR #1397, opened 2024-09-30
Ensure permutation equivariance of `oe_assign_charges`

maxentile:
Intent: ensure `oe_assign_charges(mol)[perm] == oe_assign_charges(mol[perm])` for any renumbering of atoms.

(related to https://github.com/openforcefield/openff-toolkit/issues/983 )

---------------

Initial approach: Apply `CanonicalRankAtoms` to rdkit mol before converting to oemol, calling into omega, .... This does not resolve the issue. (This confirms sensitivity to other factors, such as bond ordering.)

Revised approach: Toggle omega setting `SetCanonOrder` from `False` to `True`. (And recover initial atom ordering using `{Set/Get}MapIdx` in calling context.)


maxentile:
Reordering just the atoms doesn't achieve the initial goal.

@badisa suggested to toggle `SetCanonOrder` https://docs.eyesopen.com/toolkits/python/omegatk/OEConfGenClasses/OEMolBuilderOptions.html#OEConfGen::OEMolBuilderOptions::SetCanonOrder (which also reorders bond records, and reorders atoms within each bond record) -- currently set to False in: https://github.com/proteneer/timemachine/blob/80e4aae47169688200a802612c48cc54a0ee1010/timemachine/ff/handlers/nonbonded.py#L70-L71

badisa:
> Reordering just the atoms doesn't achieve the initial goal.
> 
> @badisa suggested to toggle `SetCanonOrder` https://docs.eyesopen.com/toolkits/python/omegatk/OEConfGenClasses/OEMolBuilderOptions.html#OEConfGen::OEMolBuilderOptions::SetCanonOrder (which also reorders bond records, and reorders atoms within each bond record) -- currently set to False in:
> 
> https://github.com/proteneer/timemachine/blob/80e4aae47169688200a802612c48cc54a0ee1010/timemachine/ff/handlers/nonbonded.py#L70-L71

To clarify, I was not suggesting to toggle `SetCanonOrder` was pointing that out that `SetCanonOrder` also reorders bonds which this PR does not. This was because we saw that `SetCanonOrder(True)` resolved an issue we saw. I do not know if the bond ordering also impacts the charges, if it does we may want to also canonicalize the bond ordering.

tests/test_minimizer.py
```diff
@@ -107,7 +107,10 @@ def test_pre_equilibrate_host_pfkfb3(host_name, mol_pair):
         host_config = HostConfig(complex_system, complex_coords, complex_box, num_water_atoms, complex_top)
     x_host, x_box = minimizer.pre_equilibrate_host(mols, host_config, ff)
     assert x_host.shape == host_config.conf.shape
-    assert compute_box_volume(x_box) < compute_box_volume(host_config.box)
+    box_vol_before = compute_box_volume(host_config.box)
```
badisa, line 110: Surprised by this being sensitive to the ligand charges... thanks for cleaning this up.

timemachine/ff/handlers/nonbonded.py
```diff
@@ -83,6 +87,12 @@ def oe_assign_charges(mol, charge_model=AM1BCCELF10):
     """assign partial charges, then premultiply by sqrt(ONE_4PI_EPS0)
     as an optimization"""
 
+    oemol = convert_to_oe(mol)
+
+    # recover the original atom order in case modified by SetCanonOrder(True)
```
badisa, line 92:
```suggestion
    # recover the original atom order in case ordering modified when generating conformers due to SetCanonOrder(True)
```

tests/test_minimizer.py
maxentile: This assertion was False by a very small margin `assert 89.02487850982907 < 88.91724368879996` at https://github.com/proteneer/timemachine/pull/1397/commits/895748d4907f2f536e44657d39ea10edfaa7215c .

(I didn't double-check if the change is still necessary with the slightly different charges at https://github.com/proteneer/timemachine/pull/1397/commits/828a5bdf46e019968a56077eb9efd62f3c65a2da , but would leave the extra wiggle room just in case of future changes.)

tests/test_handlers.py
```diff
@@ -722,45 +723,30 @@ def test_compute_or_load_am1_charges():
...
+
+@pytest.mark.parametrize("mol_idx", [0, 1, 2, 3, 4, 5])
+def test_partial_charge_equivariance_on_freesolv(mol_idx):
+    ff = Forcefield.load_default()
 
-    return dict(mol=test_mol, precomputed_charges=precomputed_charges)
+    seed = 2024
+    rng = np.random.default_rng(seed)
 
+    from timemachine.datasets import fetch_freesolv
```
badisa, line 744: Move to the top of the file.  Unless we have a circular import here??

badisa (approved):
Some minor things, otherwise great to have ordering dependence gone from charge generation!

tests/test_minimizer.py
badisa: No need to check, fundamentally there is no guarantee that the box volume is smaller. As long as the box volume is about the same it makes sense to me.

tests/test_handlers.py
maxentile: Addressed in https://github.com/proteneer/timemachine/pull/1397/commits/be9dda53d41ca9956f0882b106c37aea7be503ba

timemachine/ff/handlers/nonbonded.py
maxentile: Clarified in https://github.com/proteneer/timemachine/pull/1397/commits/88f8ff0e3dae7d61f4cd67286ae12b572d6b1d37

jkausrelay (approved):

PR #1398, opened 2024-10-01
Remove application code from rbfe module

mcwitt:
Moves application code that is fairly specific to a use case (running a list of edges asynchronously) out of the `fe.rbfe` module to the example where it is used.

### DIFF
--- a/examples/rbfe_edge_list.py
+++ b/examples/rbfe_edge_list.py
@@ -1,15 +1,211 @@
 import argparse
 import csv
+import pickle
+import traceback
 from dataclasses import replace
-from typing import List
+from typing import Any, Dict, List, NamedTuple, Optional, Sequence
 
+import numpy as np
 from openmm import app
+from rdkit import Chem
 
-from timemachine.constants import KCAL_TO_KJ
-from timemachine.fe import rbfe
-from timemachine.fe.free_energy import WaterSamplingParams
-from timemachine.fe.utils import read_sdf
+from timemachine.constants import DEFAULT_ATOM_MAPPING_KWARGS, KCAL_TO_KJ
+from timemachine.fe import atom_mapping, rbfe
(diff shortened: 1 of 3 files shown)


examples/rbfe_edge_list.py
```diff
@@ -1,15 +1,211 @@
 import argparse
 import csv
+import pickle
```
badisa, line 3: Not necessarily for this PR: Do we need this example anymore since it relies on the results csv?

badisa (approved):

examples/rbfe_edge_list.py
mcwitt: For now I think it's useful to keep around as it's used in our only (I think?) test of process-level reproducibility: https://github.com/proteneer/timemachine/blob/dbb45e6f10646884b3cbacc6127a18adcff538b5/tests/test_examples.py#L262

badisa: We do have tests that verify reproducibility of our solvent RBFE methods, should probably expand at some point to vacuum/complex: https://github.com/proteneer/timemachine/blob/master/tests/test_relative_free_energy.py#L226

As for process-level reproducibility, since we don't use this script in practice I don't know how useful it is to verify that this is reproducible. Low priority, would just free up CI time

mcwitt: I think verifying process-level reproducibility here is actually useful generally (not just for this script), e.g. to catch a case where we forget to explicitly seed a module-level PRNG (iirc this test previously caught a bug like this that was missed by the other tests that assert reproducibility within a single process)

badisa: I agree that is useful, but don't think this is the right place for that. We should have that for the script we use in practice.

Or we should have a more generic/usable example for this. Just don't want to maintain an entire example just to test a behavior that could be different than what happens in practice.

PR #1399, opened 2024-10-02
Fix 1-4 scaling factors for both ligand and protein.

proteneer:
This PR addresses a long standing gotcha with our 1-4 scaling factors for electrostatics for both proteins and ligands. This PR implements them correctly (net effect is strengthening the 1-4 electrostatic interactions from 0.5 to 0.833).


timemachine/ff/handlers/openmm_deserializer.py
```diff
@@ -111,7 +125,7 @@ def deserialize_nonbonded_force(force, N):
     beta = 2.0  # erfc correction
 
     # use the same scale factors for electrostatics and lj
```
maxentile, line 127:
```suggestion
```

tests/test_topology.py
```diff
@@ -12,9 +12,47 @@
...
+    for (i, j), qljw in zip(nb.idxs, nb_params):
+        kvs[(i, j)] = qljw
+
+    qs = ff.q_handle.parameterize(mol)
+    ljs = ff.lj_handle.parameterize(mol)
+
+    sigmas = ljs[:, 0]
+    epsilons = ljs[:, 1]
+
+    # 27-28 correspond to an H-O-C-H torsion, we expect:
```
jkausrelay, line 43: nit: label the atom idxs in H-O-C-H

timemachine/ff/handlers/openmm_deserializer.py
```diff
@@ -85,9 +91,17 @@ def deserialize_nonbonded_force(force, N):
...
             lj_scale_factor = 1 - new_eps / expected_eps
 
-        scale_factors_.append(lj_scale_factor)
+        if expected_q == 0:
+            if new_q == 0:
+                q_scale_factor = 1
+            else:
+                raise RuntimeError("Divide by zero in charge calculation")
+        else:
+            q_scale_factor = 1 - new_q / expected_q
```
maxentile, line 100: Don't 100% understand why a divide-by-zero is possible here. (But this works as-is: returns a solution to "(1 - scale_factor) * expected_q = new_q".)

maxentile (approved):
Presumably the bitwise tests will need to be updated, but otherwise looks good!

jkausrelay (approved):

tests/test_topology.py
proteneer: willfix

timemachine/ff/handlers/openmm_deserializer.py
proteneer: Roughly the statement is saying: if the *desired* charge product is non-zero, but we're *starting* with a zero charge product, there's no number that we can multiply a zero starting charge with to arrive at a non-zero desired charge. The naming is confusing - I will rename it.

timemachine/ff/handlers/openmm_deserializer.py
proteneer: willfix

PR #1402, opened 2024-10-09
Fixes Support for Disabling Optimizations

badisa:
This PR resolves the inability to build and test the repo without `-O3`. Due to having multiple definitions of `allocate` in the `timemachine` namespace, the `allocate` defined in `DeviceBuffer` was being used to allocate pinned memory when `-O3` was not included, but worked as expected with `-O3`.

The other changes in the PR:
* Use `static` and unnamed namespaces to isolate functions and structs inside of impl files. Should be part of our best practice going forward.
* If env var `DEBUG_BUILD` is set, disable `-O3` and add debugging information. Useful to catch issues like this in the future. 
* Adds `-O3` to gcc. Looks like this makes some tests 15% faster :exploding_head: 
* Prints both the GCC/NVCC compiler flags, just to be consistent and to make sure things are work as expected.

### DIFF
--- a/timemachine/cpp/CMakeLists.txt
+++ b/timemachine/cpp/CMakeLists.txt
@@ -9,14 +9,15 @@ project(timemachine LANGUAGES CXX CUDA)
 find_package(PythonInterp 3.10 REQUIRED)
 find_package(PythonLibs 3.10 REQUIRED)
 
-string(APPEND CMAKE_CXX_FLAGS "-Werror -Wall -Wextra")
+set(COMMON_COMPILER_FLAGS, "-Werror -Wall -Wextra -O3")
 
-string(APPEND CMAKE_CUDA_FLAGS "-O3 -lineinfo")
+string(APPEND CMAKE_CXX_FLAGS "${COMMON_COMPILER_FLAGS}")
+string(APPEND CMAKE_CUDA_FLAGS "${COMMON_COMPILER_FLAGS}")
+
 string(APPEND CMAKE_CUDA_FLAGS
-  " -Werror all-warnings"
+  " -lineinfo"  # Disabling lineinfo reduces .so file by ~40MB
   " -Wall"
   " -Wno-sign-compare"
-  " -Wextra"
   " -Wno-unused-parameter"
   " -Wreorder"
   " -Wfloat-conversion"
--- a/timemachine/cpp/CMakeLists.txt
+++ b/timemachine/cpp/CMakeLists.txt
@@ -9,18 +9,20 @@ project(timemachine LANGUAGES CXX CUDA)
 find_package(PythonInterp 3.10 REQUIRED)
 find_package(PythonLibs 3.10 REQUIRED)
 
-set(COMMON_COMPILER_FLAGS, "-Werror -Wall -Wextra -O3")
+set(COMMON_COMPILER_FLAGS "-Wall -Wextra")
 
-string(APPEND CMAKE_CXX_FLAGS "${COMMON_COMPILER_FLAGS}")
+string(APPEND CMAKE_CXX_FLAGS "${COMMON_COMPILER_FLAGS} -Werror")
 string(APPEND CMAKE_CUDA_FLAGS "${COMMON_COMPILER_FLAGS}")
 
 string(APPEND CMAKE_CUDA_FLAGS
   " -lineinfo"  # Disabling lineinfo reduces .so file by ~40MB
-  " -Wall"
+  " -Werror all-warnings"
   " -Wno-sign-compare"
   " -Wno-unused-parameter"
   " -Wreorder"
   " -Wfloat-conversion"
+  " -Wno-deprecated-declarations"
+  " -restrict"  # Warn if device pointers are not restricted
 )
 message(${CMAKE_CUDA_FLAGS})
 
--- a/timemachine/cpp/CMakeLists.txt
+++ b/timemachine/cpp/CMakeLists.txt
@@ -9,7 +9,15 @@ project(timemachine LANGUAGES CXX CUDA)
 find_package(PythonInterp 3.10 REQUIRED)
 find_package(PythonLibs 3.10 REQUIRED)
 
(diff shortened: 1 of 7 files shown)


proteneer:
Wait what?

> Due to having multiple definitions of allocate in the timemachine namespace, the allocate defined in DeviceBuffer was being used to allocate pinned memory when -O3 was not included, but worked as expected with -O3.

Currently on master, both `device_buffer.cu` and `pinned_host_buffer.cu` defines an `allocate` function under `namespace timemachine` but only locally in the `.cu` files (i.e. they're never exposed to the headers). But you're saying if we compile with `-O3` everything works as expected, (i.e. `device_buffer.cu` only ever calls into `cudaSafeMalloc`, and `pinned_host_buffer.cu` only ever calls into `cudaHostMalloc`.) However, if we compile with any other flag (eg. `-O2`) both `device_buffer.cu` and `pinned_host_buffer.cu` calls into the  `cudaSafeMalloc` version?

Can you just preface the `allocate` functions within each `.cu` file with `static`?

timemachine/cpp/src/device_buffer.cu
```diff
@@ -6,20 +6,24 @@
 
 namespace timemachine {
 
-template <typename T> T *allocate(const std::size_t length) {
+namespace {
+
+template <typename T> T *allocate_gpu_memory(const std::size_t length) {
```
proteneer, line 11: try:

`template <typename T> static T *allocate(const std::size_t length)`

proteneer: (`static` limits accessibility to the translation unit/file.)

proteneer:
> Use unnamed namespaces, should be our best practice going forward. To quote the Zen of Python 'Namespaces are one honking great idea -- let's do more of those!'

I *strongly* prefer `static` over unnamed namespaces for its simplicity See: https://stackoverflow.com/questions/154469/why-should-you-prefer-unnamed-namespaces-over-static-functions

timemachine/cpp/src/device_buffer.cu
badisa: Whoops statics works as well, had borked the syntax yesterday though. Will change

badisa: Referred to static in https://github.com/proteneer/timemachine/pull/1402/commits/8f2da9315309689fafe79942e6aa43affdd9af2f

Though kept one unnamed namespace for a struct definition that is local to the file.

timemachine/cpp/src/pinned_host_buffer.cu
```diff
@@ -3,7 +3,7 @@
 
 namespace timemachine {
 
-template <typename T> T *allocate(const std::size_t length) {
+template <typename T> static T *allocate_pinned_host_memory(const std::size_t length) {
```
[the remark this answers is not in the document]
badisa: With `static` the idiosyncrasy disappears as expected. I verified this on master before committing it.

proteneer (approved):
nice!

timemachine/cpp/CMakeLists.txt
```diff
@@ -9,19 +9,31 @@ project(timemachine LANGUAGES CXX CUDA)
...
+  string(APPEND COMMON_COMPILER_FLAGS " -g")
+endif()
+
+
+string(APPEND CMAKE_CXX_FLAGS "${COMMON_COMPILER_FLAGS} -Werror")
+string(APPEND CMAKE_CUDA_FLAGS "${COMMON_COMPILER_FLAGS}")
 
-string(APPEND CMAKE_CUDA_FLAGS "-O3 -lineinfo")
 string(APPEND CMAKE_CUDA_FLAGS
+  " -lineinfo"  # Disabling lineinfo reduces .so file by ~40MB
```
mcwitt, line 26: Clarification: this change results in generating line info, not disabling it, correct? (Comment seems a bit ambiguous). Should line info be enabled only for debug builds?

timemachine/cpp/src/local_md_potentials.cu
```diff
@@ -13,13 +13,17 @@
...
+namespace {
+
 // Struct representing the CUB < operation
 struct LessThan {
     int compare;
     CUB_RUNTIME_FUNCTION __device__ __forceinline__ explicit LessThan(int compare) : compare(compare) {}
     CUB_RUNTIME_FUNCTION __device__ __forceinline__ bool operator()(const int &a) const { return (a < compare); }
 };
 
+} // namespace
```
mcwitt, line 25: Unsure/question: is it possible use `static` here to stick with the convention of static over unnamed namespaces?

timemachine/cpp/src/pinned_host_buffer.cu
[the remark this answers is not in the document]
mcwitt: Not a C++ expert, but I'm surprised that the compiler/linker didn't generate an error before when there was a collision / arbitrary selection of `allocate` implementations. Is there a flag we could enable to turn this situation into an error in the future?

mcwitt (approved):

timemachine/cpp/CMakeLists.txt
badisa: We have always generated line info, which is useful in practice. But didn't realize it made the .so so much larger. Just added a comment regarding the fact that it is taking up a significant amount of space by enabling it.

timemachine/cpp/src/local_md_potentials.cu
badisa: No, can't define static structs/classes covered in https://stackoverflow.com/questions/4422507/how-are-unnamed-namespaces-superior-to-the-static-keyword

timemachine/cpp/src/pinned_host_buffer.cu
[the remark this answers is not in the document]
badisa: I was also surprised by this and tried to find a compiler/linker flag to trigger this. Was not able to find one.

PR #1405, opened 2024-10-15
Adds support for different minimization schemes

badisa:
* Starting point to explore different Minimization approaches, adds support for Scipy and FIRE. Allows users to adopt different parameters without having to adjust the details of minimization.
* Does not address minimization issues with clashy molecules, will be done in a follow up PR.

### DIFF
--- a/tests/test_minimizer.py
+++ b/tests/test_minimizer.py
@@ -179,7 +179,15 @@ def test_equilibrate_host_barker():
         print(f"\tdone in {(t1 - t0):.3f} s")
 
 
-def test_local_minimize_water_box():
+@pytest.mark.parametrize(
+    "minimizer_options",
+    [
+        minimizer.FireMinimizationOptions(100),
+        minimizer.ScipyMinimizationOptions("BFGS", {}),
+        minimizer.ScipyMinimizationOptions("L-BFGS-B", {}),
+    ],
+)
+def test_local_minimize_water_box(minimizer_options):
     """
     Test that we can locally relax a box of water by selecting some random indices.
     """
@@ -197,7 +205,7 @@ def test_local_minimize_water_box():
 
     u_init, g_init = val_and_grad_fn(x0)
 
-    x_opt = minimizer.local_minimize(x0, box0, val_and_grad_fn, free_idxs)
+    x_opt = minimizer.local_minimize(x0, box0, val_and_grad_fn, free_idxs, minimizer_options)
 
     np.testing.assert_array_equal(x0[frozen_idxs], x_opt[frozen_idxs])
     assert np.linalg.norm(x0[free_idxs] - x_opt[free_idxs]) > 0.01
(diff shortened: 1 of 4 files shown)


tests/test_minimizer.py
```diff
@@ -209,9 +217,56 @@ def test_local_minimize_water_box():
...
+    bounds = list(zip(lower_bounds, upper_bounds))
+
+    minimizer_config = minimizer.ScipyMinimizationConfig("L-BFGS-B", bounds=bounds)
+
+    u_init, g_init = val_and_grad_fn(x0)
+
+    x_opt = minimizer.local_minimize(x0, box0, val_and_grad_fn, free_idxs, minimizer_config)
+
+    np.testing.assert_array_equal(x0[frozen_idxs], x_opt[frozen_idxs])
+    assert 0.01 < np.linalg.norm(x0[free_idxs] - x_opt[free_idxs]) <= np.linalg.norm([allowed_diff * 2] * 3)
```
mcwitt, line 250: nit: worth strengthening to assert that `x_opt` is within the bounds?

timemachine/md/minimizer.py
```diff
@@ -509,20 +581,31 @@ def local_minimize(
...
     Optimized set of coordinates (N,3)
 
+    Raises
+    ------
+    AssertionError
+    MinimizationError
+
     """
 
+    if type(minimizer_config) not in (FireMinimizationConfig, ScipyMinimizationConfig):
```
mcwitt, line 596: nit: I think `isinstance` is generally preferred to `type(...) in ...`

```suggestion
    if not isinstance(minimizer_config, MinimizationConfig):
```

mcwitt (approved):
:rocket:

timemachine/md/minimizer.py
badisa: Don't know what I was thinking...

tests/test_minimizer.py
badisa: All of the `x_opt` values are identical besides `free_idxs`, as tested above. Wouldn't strengthen it to do `x0 - x_opt` if I am understanding what you are suggesting

badisa: Ah makes sense, will do.

badisa: Strengthened in https://github.com/proteneer/timemachine/pull/1405/commits/5e128eacf3cbf43324c59b488a8733a4030a1058

timemachine/md/minimizer.py
```diff
@@ -535,40 +618,28 @@ def local_minimize(
...
         x_prime[free_idxs] = x_local
-        u_full, grad_full = minimizer_val_and_grad(x_prime)
+        U_full, grad_full = minimizer_val_and_grad(x_prime)
         # The GPU Potentials can return NaN if value would have overflowed in uint64
-        if np.isnan(u_full):
-            u_full = np.inf
+        # FIRE only looks at the gradients and can recover from NaN energies
+        if method != "FIRE" and np.isnan(U_full):
+            U_full = np.inf
             grad_full = np.nan * grad_full
```
maxentile, line 626: Docstring wording: I don't think it can "recover from" NaN energies, it just doesn't look at energies (which is a benefit in cases where energy is nan, but forces are still accurate)

maxentile (approved):
Nice -- will ease testing of different optimizers

timemachine/md/minimizer.py
badisa: Clarified in https://github.com/proteneer/timemachine/pull/1405/commits/b2d9e5136c786f19191e8cb79504fdfcea6e9287

PR #1406, opened 2024-10-19
Quick fix for chiral volume conversions.

proteneer:
This PR is an attempt at a quick fix for:

1) Disabling the chiral validity check
2) Explicitly identifying bonds/angles involved in chiral volumes that are being toggled (either from off->on, or on->off) and using a different alchemical schedule for them. 

Pending a lot more systematic validation.

### DIFF
--- a/tests/test_chiral_restraints.py
+++ b/tests/test_chiral_restraints.py
@@ -854,3 +854,125 @@ def test_chiral_inversion_in_single_topology_solvent():
     # with open("solvent_chiral_energies_b.png", "wb") as ofs:
     #     data_b = plot_as_png_fxn(plot_chiral_restraint_energies, heatmap_b)
     #     ofs.write(data_b)
+
+
+import matplotlib.pyplot as plt
+
+from timemachine.fe.utils import plot_atom_mapping_grid
+
+
+def plot_and_save(f, fname, *args, **kwargs):
+    """
+    Given a function which generates a plot, return the plot as png bytes.
+    """
+    plt.clf()
+    f(*args, **kwargs)
+    with open(fname, "wb") as fh:
+        plt.savefig(fh, format="png", bbox_inches="tight")
+
+
+from timemachine.fe.plots import (
+    plot_hrex_replica_state_distribution,
+    plot_hrex_replica_state_distribution_heatmap,
+    plot_hrex_swap_acceptance_rates_convergence,
+    plot_hrex_transition_matrix,
+)
+
+
+def test_ring_breaking_chiral_restraints_failure():
+    mol_a = Chem.MolFromMolBlock(
+        """
+  Mrv2311 10092413403D
+
+  6  6  0  0  0  0            999 V2000
+    0.1292    1.5540   -0.4103 O   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.8029    0.8102    0.1698 O   0  0  0  0  0  0  0  0  0  0  0  0
+    0.0572    0.0397    0.8217 O   0  0  0  0  0  0  0  0  0  0  0  0
+    1.1063    0.7870    0.2530 C   0  0  2  0  0  0  0  0  0  0  0  0
+    2.1161    1.7939    1.5497 F   0  0  0  0  0  0  0  0  0  0  0  0
+    1.8910    0.0536   -0.6026 Cl  0  0  0  0  0  0  0  0  0  0  0  0
+  1  2  1  0  0  0  0
+  2  3  1  0  0  0  0
+  3  4  1  0  0  0  0
+  1  4  1  0  0  0  0
+  4  5  1  0  0  0  0
+  4  6  1  0  0  0  0
+M  END
+$$$$""",
+        removeHs=False,
+    )
+
+    # keeps 2 chiral restraints
+    perm = [0, 1, 2, 3, 4, 5]
+
+    # keeps all chiral restraints
+    # perm = [3, 4, 5, 0, 2, 1]
+
+    perm_kv = {}
+    for new_idx, old_idx in enumerate(perm):
+        perm_kv[old_idx] = new_idx
+
+    mol_a = Chem.RenumberAtoms(mol_a, perm)
+    mol_a.SetProp("_Name", "mol")
+    mol_b = Chem.Mol(mol_a)
+
+    core = np.array([[1, 1], [2, 2], [3, 3], [4, 5]], dtype=np.int32)
+    core_perm = []
+    for i, j in core:
+        core_perm.append([perm_kv[i], perm_kv[j]])
+    core = np.array(core_perm, dtype=np.int32)
+
+    res = plot_atom_mapping_grid(mol_a, mol_b, core)
+    fpath = "atom_mapping.svg"
+    print("core mapping written to", fpath)
+    with open(fpath, "w") as fh:
+        fh.write(res)
+
+    short_hrex_params = replace(DEFAULT_HREX_PARAMS, n_frames=1000, n_eq_steps=10000, steps_per_frame=400)
+    ff = Forcefield.load_default()
+
+    vacuum_res = run_vacuum(mol_a, mol_b, core, ff, None, short_hrex_params, n_windows=48, min_overlap=0.666)
+
+    with open("vacuum_res.pkl", "wb") as fh:
+        import pickle
+
+        pickle.dump((vacuum_res, mol_a, mol_b, core), fh)
+
+    with open("vacuum_overlap.png", "wb") as fh:
+        fh.write(vacuum_res.plots.overlap_detail_png)
+
+    plot_and_save(
+        plot_hrex_swap_acceptance_rates_convergence,
+        "vac_plot_hrex_swap_acceptance_rates_convergence.png",
+        vacuum_res.hrex_diagnostics.cumulative_swap_acceptance_rates,
+    )
+    plot_and_save(
+        plot_hrex_transition_matrix,
+        "vac_plot_hrex_transition_matrix.png",
+        vacuum_res.hrex_diagnostics.transition_matrix,
+    )
+
+    plot_and_save(
+        plot_hrex_replica_state_distribution,
+        "vac_plot_hrex_replica_state_distribution.png",
+        vacuum_res.hrex_diagnostics.cumulative_replica_state_counts,
+    )
+    plot_and_save(
+        plot_hrex_replica_state_distribution_heatmap,
+        "vac_plot_hrex_replica_state_distribution_heatmap.png",
+        vacuum_res.hrex_diagnostics.cumulative_replica_state_counts,
+    )
+
+    from timemachine.fe.plots import plot_as_png_fxn, plot_chiral_restraint_energies
+
+    atom_map_mixin = AtomMapMixin(mol_a, mol_b, core)
+
+    heatmap_a, heatmap_b = make_chiral_flip_heatmaps(vacuum_res, atom_map_mixin)
+
+    with open("solvent_chiral_energies_a.png", "wb") as ofs:
+        data_a = plot_as_png_fxn(plot_chiral_restraint_energies, heatmap_a)
+        ofs.write(data_a)
+
+    with open("solvent_chiral_energies_b.png", "wb") as ofs:
+        data_b = plot_as_png_fxn(plot_chiral_restraint_energies, heatmap_b)
+        ofs.write(data_b)
--- a/tests/test_single_topology.py
+++ b/tests/test_single_topology.py
@@ -38,6 +38,7 @@ from timemachine.fe.single_topology import (
     canonicalize_chiral_atom_idxs,
     canonicalize_improper_idxs,
     cyclic_difference,
+    find_induced_bonds_angles_from_chiral_idxs,
     handle_ring_opening_closing,
     interpolate_harmonic_bond_params,
     interpolate_harmonic_force_constant,
@@ -2005,3 +2006,66 @@ M  END
 
     # should not raise an assertion
     verify_chiral_validity_of_core(mol_a, mol_b, core, ff)
+
+
+def test_find_converting_chiral_volumes():
+    mol_a = Chem.MolFromMolBlock(
+        """
+  Mrv2311 10092413403D
+
+  6  6  0  0  0  0            999 V2000
+    0.1292    1.5540   -0.4103 O   0  0  0  0  0  0  0  0  0  0  0  0
+   -0.8029    0.8102    0.1698 O   0  0  0  0  0  0  0  0  0  0  0  0
+    0.0572    0.0397    0.8217 O   0  0  0  0  0  0  0  0  0  0  0  0
+    1.1063    0.7870    0.2530 C   0  0  2  0  0  0  0  0  0  0  0  0
+    2.1161    1.7939    1.5497 F   0  0  0  0  0  0  0  0  0  0  0  0
+    1.8910    0.0536   -0.6026 Cl  0  0  0  0  0  0  0  0  0  0  0  0
+  1  2  1  0  0  0  0
+  2  3  1  0  0  0  0
+  3  4  1  0  0  0  0
+  1  4  1  0  0  0  0
+  4  5  1  0  0  0  0
+  4  6  1  0  0  0  0
+M  END
+$$$$""",
+        removeHs=False,
+    )
+
+    for atom in mol_a.GetAtoms():
+        atom.SetProp("atomLabel", str(atom.GetIdx()))
+    # keeps 2 chiral restraints, 6 are converting
+    perm = [0, 1, 2, 3, 4, 5]
+    # keeps all chiral restraints
+    # perm = [3, 4, 5, 0, 2, 1]
+    perm_kv = {}
+    for new_idx, old_idx in enumerate(perm):
+        perm_kv[old_idx] = new_idx
+    mol_a = Chem.RenumberAtoms(mol_a, perm)
+    mol_a.SetProp("_Name", "mol")
+    mol_b = Chem.Mol(mol_a)
+    core = np.array([[1, 1], [2, 2], [3, 3], [4, 5]], dtype=np.int32)
+    core_perm = []
+    for i, j in core:
+        core_perm.append([perm_kv[i], perm_kv[j]])
+    core = np.array(core_perm, dtype=np.int32)
+
+    # res = plot_atom_mapping_grid(mol_a, mol_b, core)
+    # fpath = "atom_mapping.svg"
+    # print("core mapping written to", fpath)
+    # with open(fpath, "w") as fh:
+    #     fh.write(res)
+
+    ff = Forcefield.load_default()
+
+    st = SingleTopology(mol_a, mol_b, core, ff)
+
+    affected_chiral_atom_idxs = st.find_converting_chiral_volumes()
+    assert len(affected_chiral_atom_idxs) == 6
+
+    bond_idxs, angle_idxs = find_induced_bonds_angles_from_chiral_idxs(affected_chiral_atom_idxs)
+
+    # keep bonds the same, only stagger the angles/chiral terms differently
+
+    # print(bond_idxs, angle_idxs)
+
+    st.setup_intermediate_state(0.5)
--- a/timemachine/fe/single_topology.py
+++ b/timemachine/fe/single_topology.py
@@ -1066,6 +1066,23 @@ class AtomMapFlags(IntEnum):
     MOL_B = 2
 
 
+def find_induced_bonds_angles_from_chiral_idxs(chiral_idxs):
+    bonds = set()
+    angles = set()
+
+    for c, j, k, l in chiral_idxs:
+        bonds.add(canonicalize_bond([c, j]))
+        bonds.add(canonicalize_bond([c, k]))
+        bonds.add(canonicalize_bond([c, l]))
+
+    for c, j, k, l in chiral_idxs:
+        angles.add(canonicalize_bond([j, c, k]))
+        angles.add(canonicalize_bond([j, c, l]))
+        angles.add(canonicalize_bond([k, c, l]))
+
+    return bonds, angles
+
+
 class AtomMapMixin:
     """
     A Mixin class containing the atom_mapping information. This Mixin sets up the following
@@ -1241,9 +1258,9 @@ def assert_chiral_consistency_and_validity(
     dst_chiral_restr_idx_set = ChiralRestrIdxSet(dst_chiral_idxs)
 
     assert_chiral_consistency(src_chiral_restr_idx_set, dst_chiral_restr_idx_set, src_bond_idxs, dst_bond_idxs)
-    check_chiral_validity_impl(
-        atom_map, src_chiral_restr_idx_set, dst_chiral_restr_idx_set, src_bond_idxs, dst_bond_idxs
-    )
+    # check_chiral_validity_impl(
+    #     atom_map, src_chiral_restr_idx_set, dst_chiral_restr_idx_set, src_bond_idxs, dst_bond_idxs
+    # )
 
 
 def verify_chiral_validity_of_core(mol_a: Chem.Mol, mol_b: Chem.Mol, core: NDArray, forcefield):
@@ -1614,6 +1631,54 @@ class SingleTopology(AtomMapMixin):
 
         return ChiralBondRestraint(chiral_bond_idxs, np.array(chiral_bond_signs)).bind(jnp.array(chiral_bond_params))
 
+    def find_converting_chiral_volumes(self):
+        # this can't call into setup_intermedaite_state since it'd end up recursive
+        src_chiral_idxs = set(tuple(x) for x in self.src_system.chiral_atom.potential.idxs)
+        dst_chiral_idxs = set(tuple(x) for x in self.dst_system.chiral_atom.potential.idxs)
+        return dst_chiral_idxs.symmetric_difference(src_chiral_idxs)
+
+    def split_chiral_bonds_and_angles(self, bond_potential, angle_potential):
+        """
+        Given an angle potential, split the angle idxs into those involved in a convert chiral set and those not involved
+        """
+        bad_bond_idxs, bad_angle_idxs = find_induced_bonds_angles_from_chiral_idxs(
+            self.find_converting_chiral_volumes()
+        )
+
+        chiral_bond_idxs = []
+        chiral_bond_params = []
+        achiral_bond_idxs = []
+        achiral_bond_params = []
+
+        for idxs, params in zip(bond_potential.potential.idxs, bond_potential.params):
+            if tuple(idxs) in bad_bond_idxs:
+                chiral_bond_idxs.append(tuple(idxs))
+                chiral_bond_params.append(params)
+            else:
+                achiral_bond_idxs.append(tuple(idxs))
+                achiral_bond_params.append(params)
+
+        chiral_bond = HarmonicBond(np.array(chiral_bond_idxs, dtype=np.int32)).bind(chiral_bond_params)
+        achiral_bond = HarmonicBond(np.array(achiral_bond_idxs, dtype=np.int32)).bind(achiral_bond_params)
+
+        chiral_angle_idxs = []
+        chiral_angle_params = []
+        achiral_angle_idxs = []
+        achiral_angle_params = []
+
+        for idxs, params in zip(angle_potential.potential.idxs, angle_potential.params):
+            if tuple(idxs) in bad_angle_idxs:
+                chiral_angle_idxs.append(tuple(idxs))
+                chiral_angle_params.append(params)
+            else:
+                achiral_angle_idxs.append(tuple(idxs))
+                achiral_angle_params.append(params)
+
+        chiral_angle = HarmonicAngleStable(np.array(chiral_angle_idxs, dtype=np.int32)).bind(chiral_angle_params)
+        achiral_angle = HarmonicAngleStable(np.array(achiral_angle_idxs, dtype=np.int32)).bind(achiral_angle_params)
+
+        return chiral_bond, achiral_bond, chiral_angle, achiral_angle
+
     def setup_intermediate_state(self, lamb) -> VacuumSystem:
         r"""
         Set up intermediate states at some value of the alchemical parameter :math:`\lambda`.
@@ -1640,15 +1705,45 @@ class SingleTopology(AtomMapMixin):
         src_system = self.src_system
         dst_system = self.dst_system
 
-        # stagger the lambda schedule
+        # split schedule depending on if we're breaking/forming chiral volumes
         bonds_min, bonds_max = [0.0, 0.7]
         angles_min, angles_max = [0.0, 0.7]
         torsions_min, torsions_max = [0.7, 1.0]
-        chiral_atoms_min, chiral_atoms_max = [0.7, 1.0]
 
-        bond = self._setup_intermediate_bonded_term(
-            src_system.bond,
-            dst_system.bond,
+        # bonds involved in chiral volumes need to be turned on before volumes
+        # angles involved in chiral volumes need to be turned on after volumes
+        chiral_bonds_min, chiral_bonds_max = [0.0, 0.5]
+        chiral_atoms_min, chiral_atoms_max = [0.5, 0.6]
+        chiral_angles_min, chiral_angles_max = [0.6, 0.7]
+
+        (
+            src_affected_bond,
+            src_unaffected_bond,
+            src_affected_angle,
+            src_unaffected_angle,
+        ) = self.split_chiral_bonds_and_angles(src_system.bond, src_system.angle)
+        (
+            dst_affected_bond,
+            dst_unaffected_bond,
+            dst_affected_angle,
+            dst_unaffected_angle,
+        ) = self.split_chiral_bonds_and_angles(dst_system.bond, dst_system.angle)
+
+        bond_affected = self._setup_intermediate_bonded_term(
+            src_affected_bond,
+            dst_affected_bond,
+            lamb,
+            interpolate.align_harmonic_bond_idxs_and_params,
+            partial(
+                interpolate_harmonic_bond_params,
+                k_min=0.1,  # ~ BOLTZ * (300 K) / (5 nm)^2
+                lambda_min=chiral_bonds_min,
+                lambda_max=chiral_bonds_max,
+            ),
+        )
+        bond_unaffected = self._setup_intermediate_bonded_term(
+            src_unaffected_bond,
+            dst_unaffected_bond,
             lamb,
             interpolate.align_harmonic_bond_idxs_and_params,
             partial(
@@ -1658,10 +1753,25 @@ class SingleTopology(AtomMapMixin):
                 lambda_max=bonds_max,
             ),
         )
+        bond = HarmonicBond(np.concatenate([bond_affected.potential.idxs, bond_unaffected.potential.idxs])).bind(
+            np.concatenate([bond_affected.params, bond_unaffected.params])
+        )
 
-        angle = self._setup_intermediate_bonded_term(
-            src_system.angle,
-            dst_system.angle,
+        angle_affected = self._setup_intermediate_bonded_term(
+            src_affected_angle,
+            dst_affected_angle,
+            lamb,
+            interpolate.align_harmonic_angle_idxs_and_params,
+            partial(
+                interpolate_harmonic_angle_params,
+                k_min=0.05,  # ~ BOLTZ * (300 K) / (2 * pi)^2
+                lambda_min=chiral_angles_min,
+                lambda_max=chiral_angles_max,
+            ),
+        )
+        angle_unaffected = self._setup_intermediate_bonded_term(
+            src_unaffected_angle,
+            dst_unaffected_angle,
             lamb,
             interpolate.align_harmonic_angle_idxs_and_params,
             partial(
@@ -1671,6 +1781,9 @@ class SingleTopology(AtomMapMixin):
                 lambda_max=angles_max,
             ),
         )
+        angle = HarmonicAngleStable(
+            np.concatenate([angle_affected.potential.idxs, angle_unaffected.potential.idxs])
+        ).bind(np.concatenate([angle_affected.params, angle_unaffected.params]))
 
         assert src_system.torsion
         assert dst_system.torsion


timemachine/fe/single_topology.py
```diff
@@ -1028,17 +1015,18 @@ def interpolate_periodic_torsion_params(src_params, dst_params, lamb, lambda_min
     """
 
     src_k, src_phase, src_period = src_params
-    dst_k, dst_phase, _ = dst_params
+    dst_k, dst_phase, dst_period = dst_params
 
-    k = handle_ring_opening_closing(interpolate.linear_interpolation, src_k, dst_k, lamb, lambda_min, lambda_max)
+    # assert src_period == dst_period
```
proteneer, line 1020: note to self, remove

timemachine/fe/single_topology.py
```diff
@@ -1615,6 +1517,326 @@ def _setup_intermediate_chiral_bond_term(
 
         return ChiralBondRestraint(chiral_bond_idxs, np.array(chiral_bond_signs)).bind(jnp.array(chiral_bond_params))
 
+    def _expand_chiral_volumes_into_angles(self, chiral_idxs):
```
proteneer, line 1520: Start of the main logic part of the PR

timemachine/fe/single_topology.py
```diff
@@ -1638,84 +1860,35 @@ def setup_intermediate_state(self, lamb) -> VacuumSystem:
...
-                k_min=0.025,
-                lambda_min=chiral_atoms_min,
-                lambda_max=chiral_atoms_max,
-            ),
-        )
+        bond = self.align_and_interpolate_bonds(lamb)
+        angle = self.align_and_interpolate_angles(lamb)
+        chiral_atom = self.align_and_interpolate_chiral_atoms(lamb)
+        torsion = self.align_and_interpolate_torsions(lamb)
+        nonbonded = self.align_and_interpolate_intramolecular_nonbonded(lamb)
```
proteneer, line 1887: End of the main logic of the PR

proteneer:
Updated description and opened for comments.

timemachine/fe/single_topology.py
```diff
@@ -796,102 +826,50 @@ def find_dummy_groups_and_anchors(
...
 
     Returns
     -------
     float
         interpolated force constant
     """
-
     return jnp.where(
-        lamb == 0.0,
+        lamb <= lambda_min,
```
mcwitt, line 857: The `where` branches here seem redundant with the logic in `intepolate.pad`, unless I'm missing something?

timemachine/fe/single_topology.py
```diff
@@ -980,21 +968,20 @@ def interpolate_harmonic_angle_params(src_params, dst_params, lamb, k_min, lambd
...
+    k = padded_interpolation(log_linear_fn, src_k, dst_k, lamb, lambda_min, lambda_max)
 
-    phase = interpolate.linear_interpolation(
-        src_phase,
-        src_phase + cyclic_difference(src_phase, dst_phase, period=2 * np.pi),
-        lamb,
-    )
+    src_x = src_phase
+    dst_x = src_phase + cyclic_difference(src_phase, dst_phase, period=2 * np.pi)
+    x = padded_interpolation(interpolate.linear_interpolation, src_x, dst_x, lamb, lambda_min, lambda_max)
```
mcwitt, line 976: nit: seems more descriptive to keep this named `phase`, unless there's a convention I'm unaware of

timemachine/fe/single_topology.py
mcwitt, line 1520: nit: 
```suggestion
    def _expand_chiral_volumes_into_angles(self, chiral_atom_idxs):
```

timemachine/fe/single_topology.py
```diff
@@ -1615,6 +1517,326 @@ def _setup_intermediate_chiral_bond_term(
...
+        core_atoms = self.get_core_atoms()
+        return all([x in core_atoms for x in idxs])
+
+    def _idxs_belong_to_dummy_a(self, idxs):
+        dummy_atoms = self.get_dummy_atoms_a()
+        return any([x in dummy_atoms for x in idxs])
+
+    def _idxs_belong_to_dummy_b(self, idxs):
+        dummy_atoms = self.get_dummy_atoms_b()
+        return any([x in dummy_atoms for x in idxs])
```
mcwitt, line 1538: nit: seems confusing that `_idxs_belong_to_core` uses `all`, while `_idxs_belong_to_dummy_a` and `_idxs_belong_to_dummy_b` use `any`. Maybe renaming would clarify?

mcwitt (commented):
Just a couple quick comments on the first skim. Tomorrow I'll dig into the single topology logic changes.

timemachine/fe/free_energy.py
```diff
@@ -1407,17 +1407,17 @@ def format_rate(r):
                 return f"{r * 100.0:5.1f}%"
 
             def format_rates(rs):
-                return " | ".join(format_rate(r) for r in rs)
+                return " |".join(format_rate(r) for r in rs)
```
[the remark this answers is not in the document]
proteneer: Minor but intentional: there's a spacing issue in the current HREX print statements where there's an extra space to the right of the "|" character, eg: ` 26.3% |  28.%` because the formatting is for 3 digits to the left of `.` - I was trying to make it more compact

timemachine/fe/free_energy.py
```diff
@@ -1407,17 +1407,17 @@ def format_rate(r):
...
                 f"{estimated_wall_time_remaining:.1f} s remaining at "
                 f"{wall_time_per_frame_average:.2f} s/frame "
                 f"({wall_time_per_frame_current:.2f} s/frame since last message)"
             )
-            print("HREX acceptance rates, current :", format_rates(instantaneous_swap_acceptance_rates))
-            print("HREX acceptance rates, average :", format_rates(average_swap_acceptance_rates))
-            print("HREX replica permutation       :", hrex.replica_idx_by_state)
+            print("HREX acceptance rates, current:", format_rates(instantaneous_swap_acceptance_rates))
+            print("HREX acceptance rates, average:", format_rates(average_swap_acceptance_rates))
+            print("HREX replica permutation      :", hrex.replica_idx_by_state)
```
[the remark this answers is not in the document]
proteneer: Minor but intentional: just wanted to squeeze as much as possible into a single line without line breaking in my terminal.

timemachine/fe/single_topology.py
proteneer: Ah, I think you're right - I removed some symmetrization code in an earlier version that made it obsolete.

timemachine/fe/single_topology.py
```diff
@@ -796,102 +826,50 @@ def find_dummy_groups_and_anchors(
...
-            ),
+            interpolate.pad(interpolate_fn, src_k, dst_k, lamb, lambda_min, lambda_max),
         ),
     )
 
 
+# TBD: generalize to non-log-linear interpolation
+# def interpolate_harmonic_force_constant(src_k, dst_k, lamb, k_min, lambda_min, lambda_max):
+#     log_linear_fn = partial(interpolate.log_linear_interpolation, min_value=k_min)
+#     return symmetrically_padded_interpolation(log_linear_fn, src_k, dst_k, lamb, lambda_min, lambda_max)
```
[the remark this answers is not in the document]
proteneer: yes, will remove

timemachine/fe/single_topology.py
proteneer: willfix

timemachine/fe/single_topology.py
proteneer: explicitly renamed to:

```
    def all_idxs_belong_to_core(self, idxs):
    def any_idxs_belong_to_dummy_a(self, idxs):
    def any_idxs_belong_to_dummy_b(self, idxs):
```

timemachine/fe/single_topology.py
proteneer: fixed by deletion

timemachine/fe/single_topology.py
```diff
@@ -1615,6 +1471,315 @@ def _setup_intermediate_chiral_bond_term(
...
+                chiral_atom_idxs.append(idxs)
+                chiral_atom_params.append(params)
+
+        chiral_atom_idxs = np.array(chiral_atom_idxs).reshape(-1, 4)
+        chiral_atom_params = np.array(chiral_atom_params).reshape(-1)
+
+        src_chiral_atom = ChiralAtomRestraint(chiral_atom_idxs).bind(chiral_atom_params)
+        dst_chiral_atom = ChiralAtomRestraint(chiral_atom_idxs).bind(chiral_atom_params)
+
+        return src_chiral_atom, dst_chiral_atom
```
maxentile, line 1550: unclear intent: `find_core_chiral_atoms_unaffected` returns two identical objects (and the three functions `find_core_chiral_atoms_{unaffected/turning_on/turning_off}` appear not to be referenced currently)

maxentile (commented):
I think this addresses an important problem, but in the current review pass I've mostly gotten tripped up on superficial aspects of the diff, such as unused(?) code.

timemachine/fe/chiral_utils.py
```diff
@@ -145,7 +145,7 @@ def find_chiral_atoms(mol):
     chiral_patterns = [
         "[X4:1]",  # any tetrahedral atom
         "[#16X3,#15X3:1]",  # trivalent sulfur, phosphorous are assumed to be non-invertible
-        # "[#7X3:1](~[R])(~[R])~[R]",  # nitrogen directly bonded to three ring atoms  # TODO: handle pyramidal nitrogen
+        "[#7X3:1](~[#9])(~[#9])~[#9]",  # nitrogen directly bonded to three ring atoms  # TODO: handle pyramidal nitrogen
```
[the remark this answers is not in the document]
proteneer: reverted

timemachine/fe/single_topology.py
proteneer: sorry stale code from a bad merge earlier - will remove the dead code

timemachine/fe/interpolate.py
```diff
@@ -168,14 +168,14 @@ def log_linear_interpolation(src_params, dst_params, lamb, min_value):
 
 def pad(f, src_params, dst_params, lamb, lambda_min, lambda_max):
     """
-    Use the specified interpolation function in the interval [lambda_min, lambda_max], otherwise pin to the end-state
+    Use the specified interpolation function in the interval (lambda_min, lambda_max), otherwise pin to the end-state
```
proteneer, line 171: @mcwitt - does this seem reasonable? not sure what the original intent here was with the closed interval, open seems more appropriate (eg. interpolating over lambda_min=0, lambda_max=1 seems safer if we guarantee that the end-states are exactly as src_params, and dst_params, respectively).

timemachine/fe/single_topology.py
```diff
@@ -1725,7 +1841,7 @@ def setup_intermediate_state(self, lamb) -> VacuumSystem:
 
         return VacuumSystem(bond, angle, torsion, nonbonded, chiral_atom, chiral_bond)
 
-    def mol(self, lamb, min_bond_k=100.0) -> Chem.Mol:
+    def mol(self, lamb, min_bond_k=30.0) -> Chem.Mol:
```
proteneer, line 1786: This was modified to make the test_single_topology_mol_utils.py tests pass, the initial `min_bond_k` was too high, and was set fairly arbitrarily.

timemachine/fe/single_topology.py
```diff
@@ -1615,6 +1471,257 @@ def _setup_intermediate_chiral_bond_term(
...
+
+    def any_idxs_belong_to_dummy_a(self, idxs):
+        dummy_atoms = self.get_dummy_atoms_a()
+        return any([x in dummy_atoms for x in idxs])
+
+    def any_idxs_belong_to_dummy_b(self, idxs):
+        dummy_atoms = self.get_dummy_atoms_b()
+        return any([x in dummy_atoms for x in idxs])
+
+    def _find_core_angles_turning_on(self):
```
proteneer, line 1494: sorry I think this is stale too - will remove.

timemachine/fe/single_topology.py
```diff
@@ -1205,27 +1168,7 @@ def assert_chiral_consistency(
...
-    src_chiral_centers_in_mol_c = [atom_map.a_to_c[x] for x in chiral_centers_in_mol_a]
-    dst_chiral_centers_in_mol_c = [atom_map.b_to_c[x] for x in chiral_centers_in_mol_b]
-
-    check_chiral_validity_src_dst(src_chiral_centers_in_mol_c, dst_chiral_restr_idx_set, src_bond_idxs)
-    check_chiral_validity_src_dst(dst_chiral_centers_in_mol_c, src_chiral_restr_idx_set, dst_bond_idxs)
-
-
-def assert_chiral_consistency_and_validity(
-    atom_map: AtomMapMixin,
+def assert_chiral_consistency(
```
maxentile, line 1171: * remove references to the distinction between `consistency` and `validity` from docstring,
* possibly inline the definition of `_assert_chiral_consistency` here, rather than keeping both functions `assert_chiral_consistency` and `_assert_chiral_consistency` with slightly different signatures? (but if it makes something convenient somewhere, keeping both signatures is fine)

timemachine/fe/single_topology.py
```diff
@@ -1615,6 +1471,219 @@ def _setup_intermediate_chiral_bond_term(
...
+            if self._angle_idxs_belong_to_chiral_volume_turning_on(idxs):
+                min_max = DUMMY_B_CHIRAL_ANGLE_CONVERTING_ON_MIN_MAX
+            elif self._angle_idxs_belong_to_chiral_volume_turning_off(idxs):
+                assert 0
+            else:
+                min_max = DUMMY_B_ANGLE_MIN_MAX
+        else:
+            assert 0
+
+        k_min = 0.05
```
maxentile, line 1587:
```suggestion
```
(already defined a few lines earlier in this function)

tests/test_interpolate.py
```diff
@@ -351,3 +352,623 @@ def test_duplicate_idxs_period_pairs():
...
+    atol_a0 = 1e-4
+    for lamb in np.linspace(0.05, 0.95, 12):
+        itm = st.setup_intermediate_state(lamb)
+        itm_bond_k = itm.bond.params[0][0]
+        itm_bond_b0 = itm.bond.params[0][1]
+        itm_angle_k = itm.angle.params[0][0]
+        itm_angle_a0 = itm.angle.params[0][1]
+        itm_chiral_k = itm.chiral_atom.params[0]
+
+        # bonds are lambda independent
```
mcwitt, line 554: Should this say "bonds are interpolated over the full range of lambda"? (They don't seem lambda-independent, comparing with assertions immediately below)

tests/test_interpolate.py
```diff
@@ -351,3 +352,623 @@ def test_duplicate_idxs_period_pairs():
     assert duplicate_torsion_idxs not in counts_kv_dst
 
     st.setup_intermediate_state(0.5)
+
+
+def test_padded_interpolation():
+    # verify that symmetrically padded interpolation has the correct symmetry, staggering, truncating,
+    # and interpolation behavior.
```
mcwitt, line 359: nit: currently this only seems to test that we correctly return the end-state values outside of the interpolation range?

tests/test_interpolate.py
```diff
@@ -351,3 +352,623 @@ def test_duplicate_idxs_period_pairs():
...
+    #  5                 *    xx
+    #  4                *       xxxxxx
+    #  3              **              xxx
+    #  2             *                   xxxxxxxx
+    #  1 [0] ********
+    #  0     |------|------|------|------|------|
+    #        0     0.2    0.4    0.6    0.8    1.0
+
+    def interpolate_fn(*_):
+        # epsilon used to mimic effect of k_min creating an artficat in the log_linear functions
```
mcwitt, line 377: Stale comment (no epsilon)

tests/test_interpolate.py
```diff
@@ -351,3 +352,623 @@ def test_duplicate_idxs_period_pairs():
...
+    st = SingleTopology(mol_a, mol_b, core, ff)
+
+    lhs = st.setup_intermediate_state(0.0)
+    rhs = st.setup_intermediate_state(1.0)
+
+    # check that chiral volume is undefined in both end-states
+    assert len(lhs.chiral_atom.potential.idxs) == 0
+    assert len(rhs.chiral_atom.potential.idxs) == 0
+
+    lhs_bond_k = lhs.bond.params[0][0]
```
mcwitt, line 477: Could this (and several of the tests below) be strengthened to check all interactions (right now seems to only look at the first)? I think you can just vectorize all of the assertions pretty trivially unless I'm missing something. E.g.
```suggestion
    lhs_bond_k = lhs.bond.params[:, 0]
```

tests/test_interpolate.py
```diff
@@ -351,3 +352,623 @@ def test_duplicate_idxs_period_pairs():
...
+    atol_b0 = 1e-5
+    atol_a0 = 1e-4
+    for lamb in np.linspace(0.05, 0.95, 12):
+        itm = st.setup_intermediate_state(lamb)
+        itm_bond_k = itm.bond.params[0][0]
+        itm_bond_b0 = itm.bond.params[0][1]
+        itm_angle_k = itm.angle.params[0][0]
+        itm_angle_a0 = itm.angle.params[0][1]
+
+        assert abs(itm_bond_k - lhs_bond_k) > atol_k
```
mcwitt, line 498: nit: consider `pytest.approx` (or `np.testing.assert_allclose`, if vectorizing over all ixns) for readability and more useful failure diffs

tests/test_chiral_restraints.py
```diff
@@ -698,6 +698,46 @@ def make_chiral_flip_pair(well_aligned=True):
         return AtomMapMixin(mol_a, mol_b_0, core_0)
 
 
+def make_chiral_sp2_to_sp3():
```
mcwitt, line 701: This seems unused?

timemachine/fe/single_topology.py
```diff
@@ -1241,77 +1184,7 @@ def assert_chiral_consistency_and_validity(
...
-        check_chiral_validity_impl(
-            atom_map,
-            src_chiral_restr_idx_set,
-            dst_chiral_restr_idx_set,
-            bond_pot_src.potential.idxs,
-            bond_pot_dst.potential.idxs,
-        )
-
-    return check_chiral_validity
+    _assert_chiral_consistency(src_chiral_restr_idx_set, dst_chiral_restr_idx_set, src_bond_idxs, dst_bond_idxs)
```
jkausrelay, line 1187: nit: Is there a reason not to just merge this with the above?

tests/test_interpolate.py
```diff
@@ -351,3 +352,623 @@ def test_duplicate_idxs_period_pairs():
...
+
+
+@pytest.mark.parametrize(
+    "mol_a, mol_b, core",
+    [
+        (get_non_inv_nitrogen(), get_aldehyde(), np.array([[0, 0], [1, 1], [2, 2], [3, 3]], dtype=np.int32)),
+        (get_non_inv_nitrogen(), get_inv_nitrogen(), np.array([[0, 0], [1, 1], [2, 2], [3, 3]], dtype=np.int32)),
+    ],
+)
+def test_core_chiral_interpolation_sp2_to_invertible_sp3_rev(mol_a, mol_b, core):
```
mcwitt, line 586: nit: this seems almost identical to the forward version except for the ordering of the input mols? Can they be combined (maybe just by appending parameters to the first)?

timemachine/fe/single_topology.py
```diff
@@ -1615,6 +1471,219 @@ def _setup_intermediate_chiral_bond_term(
...
+        dummy_atoms = self.get_dummy_atoms_a()
+        return any([x in dummy_atoms for x in idxs])
+
+    def any_idxs_belong_to_dummy_b(self, idxs):
+        dummy_atoms = self.get_dummy_atoms_b()
+        return any([x in dummy_atoms for x in idxs])
+
+    def _chiral_volume_is_turning_on(self, idxs):
+        src_chiral_idxs = set(tuple(x) for x in self.src_system.chiral_atom.potential.idxs)
+        dst_chiral_idxs = set(tuple(x) for x in self.dst_system.chiral_atom.potential.idxs)
```
jkausrelay, line 1488: nit: add as property since this is repeated a few times

timemachine/fe/single_topology.py
```diff
@@ -1615,6 +1471,219 @@ def _setup_intermediate_chiral_bond_term(
...
+            if self._bond_idxs_belong_to_chiral_volume_turning_on(idxs):
+                min_max = DUMMY_B_CHIRAL_BOND_CONVERTING_ON_MIN_MAX
+            elif self._bond_idxs_belong_to_chiral_volume_turning_off(idxs):
+                assert 0
+            else:
+                min_max = DUMMY_B_BOND_MIN_MAX
+        else:
+            assert 0
+
+        k_min = 0.025
```
jkausrelay, line 1558: Add a comment (why 0.025?)

timemachine/fe/interpolate.py
mcwitt: I think it's reasonable; there was no particular reason for the choice of closed interval initially (but it didn't matter then because `interpolate.pad` was never used directly to do parameter interpolation; it was wrapped with `handle_ring_opening_closing`, which special-cased lambda=0 and 1 to return the end-state values).

timemachine/fe/plots.py
```diff
@@ -454,3 +454,208 @@ def plot_as_png_fxn(f, *args, **kwargs) -> bytes:
...
+    axs[row, 1].set_xlabel("lambda window")
+
+    # axs[row, 0].legend(bbox_to_anchor=(1, 0.5))
+    # axs[row, 1].legend(bbox_to_anchor=(1, 0.5))
+
+
+def plot_interpolation_schedule(st, filter_fn, fig_title):
+    fig, axs = plt.subplots(5, 2, figsize=(9, 12))
+    # plot the force constant and equilibrium bond lengths along lambda
+    n_windows = 21
```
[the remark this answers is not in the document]
mcwitt: If this is slow, also note that `setup_intermediate_state` can be called inside of a function that takes a lambda and returns parameter arrays, and that function should be jax-transformable (`vmap`, `jit`, etc.). This can make a big difference in my experience.

timemachine/fe/plots.py
```diff
@@ -454,3 +454,208 @@ def plot_as_png_fxn(f, *args, **kwargs) -> bytes:
...
+    axs[row, 1].set_ylabel("torsion degree")
+
+    axs[row, 0].set_xlabel("lambda window")
+    axs[row, 1].set_xlabel("lambda window")
+
+    # axs[row, 0].legend(bbox_to_anchor=(1, 0.5))
+    # axs[row, 1].legend(bbox_to_anchor=(1, 0.5))
+
+
+def plot_interpolation_schedule(st, filter_fn, fig_title):
```
mcwitt, line 635: Nice! This utility seems very useful.

timemachine/fe/single_topology.py
```diff
@@ -1123,6 +1065,27 @@ def __init__(self, mol_a, mol_b, core):
         self.c_to_a = {v: k for k, v in enumerate(self.a_to_c)}
         self.c_to_b = {v: k for k, v in enumerate(self.b_to_c)}
 
+    def get_dummy_atoms_a(self):
+        dummies = set()
+        for idx, flag in enumerate(self.c_flags):
+            if flag == AtomMapFlags.MOL_A:
+                dummies.add(idx)
+        return dummies
```
mcwitt, line 1073: nit/style preference: these can be one-liners with set comprehensions
```suggestion
        return {idx for idx, flag in enumerate(self.c_flags) if flag == AtomMapFlags.MOL_A}
```

tests/test_interpolate.py
```diff
@@ -351,3 +352,623 @@ def test_duplicate_idxs_period_pairs():
...
+        )
+        _verify_identical_end_states(
+            ChiralAtomTag.N0_H1_H3_D4, itm.chiral_atom.params, lhs.chiral_atom.params, rhs.chiral_atom.params
+        )
+        _verify_identical_end_states(
+            ChiralAtomTag.N0_H2_H3_D4, itm.chiral_atom.params, lhs.chiral_atom.params, rhs.chiral_atom.params
+        )
+
+
+def test_dummy_achiral_conversion():
```
jkausrelay, line 973: todo?

timemachine/fe/single_topology.py
proteneer: slightly different, there, `k_min=0.025`, but that's really too weak to be called anything resembling a bond.

timemachine/fe/plots.py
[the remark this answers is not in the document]
proteneer: willfix to higher resolution and use jit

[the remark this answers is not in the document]
proteneer: (running into some issues with jit right now, namely the returned VacuumSystems are not jit-compatible, so I will need to figure out how to use Matt's approach later)

timemachine/fe/plots.py
```diff
@@ -454,3 +454,208 @@ def plot_as_png_fxn(f, *args, **kwargs) -> bytes:
...
+    # plot the force constant and equilibrium bond lengths along lambda
+    n_windows = 21
+    xs = np.linspace(0, 1.0, n_windows)
+    systems = []
+    for lamb in xs:
+        systems.append(st.setup_intermediate_state(lamb))
+    _plot_bond_interpolation(st, xs, systems, filter_fn, axs, row=0)
+    _plot_chiral_atom_interpolation(xs, systems, filter_fn, axs, row=1)
+    _plot_angle_interpolation(st, xs, systems, filter_fn, axs, row=2)
+    _plot_torsion_interpolation(xs, systems, filter_fn, axs, row=3)
```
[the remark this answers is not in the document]
proteneer: will fix - the greatest tragedy is that lambda is a reserved keyword in python

timemachine/fe/single_topology.py
```diff
@@ -41,6 +42,45 @@
 OpenMMTopology = Any
 
 
+# Master Schedule
+# tbd: separate torsions into proper and improper later
+def _flip_min_max(min_max):
+    lamb_min, lamb_max = min_max
+    return 1 - lamb_max, 1 - lamb_min
```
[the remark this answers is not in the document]
proteneer: will add inline

timemachine/fe/single_topology.py
```diff
@@ -41,6 +42,45 @@
...
+DUMMY_B_CHIRAL_ANGLE_CONVERTING_ON_MIN_MAX = [0.5, 0.7]  # angles are all turned off
+
+# chiral and converting dummy A groups are turning off
+DUMMY_A_CHIRAL_BOND_CONVERTING_OFF_MIN_MAX = _flip_min_max(DUMMY_B_CHIRAL_BOND_CONVERTING_ON_MIN_MAX)
+DUMMY_A_CHIRAL_ATOM_CONVERTING_OFF_MIN_MAX = _flip_min_max(DUMMY_B_CHIRAL_ATOM_CONVERTING_ON_MIN_MAX)
+DUMMY_A_CHIRAL_ANGLE_CONVERTING_OFF_MIN_MAX = _flip_min_max(DUMMY_B_CHIRAL_ANGLE_CONVERTING_ON_MIN_MAX)
+
+# torsions are the same throughout
+DUMMY_B_TORSION_MIN_MAX = [0.7, 1.0]  # chiral and achiral both use the same min/max for torsions
+DUMMY_A_TORSION_MIN_MAX = _flip_min_max(DUMMY_B_TORSION_MIN_MAX)
```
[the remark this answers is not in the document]
proteneer: Great point - hm let me think about this a bit. I debated something like this but I didn't want to give a false impression that the boundaries are "firm", i.e. there's probably some lee-way in how much they can overlap. (eg. there's already overlap between
`DUMMY_B_CHIRAL_BOND_CONVERTING_ON_MIN_MAX = [0.0, 0.7]`
and
`DUMMY_B_CHIRAL_ATOM_CONVERTING_ON_MIN_MAX = [0.3, 0.5]`

Even the left-bound of torsion checkpoint `DUMMY_B_TORSION_MIN_MAX = [0.7, 1.0]` can probably overlap with the right-bound of the angle checkpoint:

`DUMMY_B_CHIRAL_ANGLE_CONVERTING_ON_MIN_MAX = [0.5, 0.7]`

The key distinction here is that an angle term doesn't have to *fully* convert to its end-state before turning on the *torsion* term, it just has to have a reasonable force constant.

timemachine/fe/single_topology.py
```diff
@@ -1030,15 +974,13 @@ def interpolate_periodic_torsion_params(src_params, dst_params, lamb, lambda_min
...
-        src_phase,
-        src_phase + cyclic_difference(src_phase, dst_phase, period=2 * np.pi),
-        lamb,
-    )
+    src_x = src_phase
+    dst_x = src_phase + cyclic_difference(src_phase, dst_phase, period=2 * np.pi)
+    x = pad(interpolate.linear_interpolation, src_x, dst_x, lamb, lambda_min, lambda_max)
 
-    return jnp.array([k, phase, src_period])
+    return jnp.array([k, x, src_period])
```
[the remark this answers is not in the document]
proteneer: thanks - this was unintentional, and fixed

timemachine/fe/single_topology.py
```diff
@@ -1123,6 +1065,27 @@ def __init__(self, mol_a, mol_b, core):
...
+            if flag == AtomMapFlags.MOL_B:
+                dummies.add(idx)
+        return dummies
+
+    def get_core_atoms(self):
+        dummies = set()
+        for idx, flag in enumerate(self.c_flags):
+            if flag == AtomMapFlags.CORE:
+                dummies.add(idx)
+        return dummies
```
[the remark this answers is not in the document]
proteneer: mostly for performance reasons, (checking for something is in a set is faster than checking if it's a list). But I haven't profiled it extensively, and `setup_intermediate_state(x)` is pretty fast (~1 sec) for me. So I haven't bothered to optimize this.

timemachine/fe/single_topology.py
proteneer: ah yeah, good points. Willl clean-up.

timemachine/fe/single_topology.py
proteneer: willfix

timemachine/fe/single_topology.py
```diff
@@ -1615,6 +1517,326 @@ def _setup_intermediate_chiral_bond_term(
...
+            else:
+                min_max = DUMMY_B_BOND_MIN_MAX
+        else:
+            assert 0
+
+        k_min = 0.025
+        return interpolate_harmonic_bond_params(src_params, dst_params, lamb, k_min, *min_max)
+
+    def _interpolate_angle(self, idxs, src_params, dst_params, lamb):
+        k_min = 0.05
```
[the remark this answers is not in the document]
proteneer: willfix

tests/test_interpolate.py
proteneer: gah, good catch, another stale comment.

tests/test_interpolate.py
proteneer: willfix

timemachine/fe/single_topology.py
proteneer: willfix

timemachine/fe/single_topology.py
proteneer: (Direct copy-pasta from before) - I don't remember why actually, I think @mcwitt  did the initial tuning for k_min.

timemachine/fe/single_topology.py
proteneer: willfix

timemachine/fe/single_topology.py
proteneer: fixed

timemachine/fe/single_topology.py
mcwitt: I'm not sure where that `k_min` came from. It looks like it was added in https://github.com/mcwitt/timemachine/commit/b03f1bcee97e972e0fac4714337482aef4b0bf62.

tests/test_interpolate.py
proteneer: willfix

tests/test_chiral_restraints.py
proteneer: willfix

timemachine/fe/plots.py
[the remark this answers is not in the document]
proteneer: (discussed offline with @mcwitt  and will defer jit to a later pass, for now, setting n_windows=48 to improve resolution in the near term).

timemachine/fe/interpolate.py
proteneer: thanks! resolving this

timemachine/fe/chiral_utils.py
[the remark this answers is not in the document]
proteneer: re-reverted - I had to add back in the NF3 pattern for some of the tests (namely, `test_interpolate.py::test_core_chiral_interpolation_sp2_to_invertible_sp3_fwd`).

I originally added this so I can easily create a test case for a non-invertible 3-connected pyramidal group (without resorting to 4-connected groups). This pattern is exact, i.e. it will *only* match the NF3 molecule exactly (it's a highly volatile compound with no practical purposes).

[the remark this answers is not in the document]
proteneer: tagging @maxentile for visibility. Added a comment specifying the intended usage.

tests/test_single_topology.py
```diff
@@ -673,41 +663,41 @@ def test_setup_intermediate_state_not_unreasonably_slow(arbitrary_transformation
...
-
-    def setup_intermediate_bonded_term_ref(src_bond, dst_bond, lamb, align_fn, interpolate_fn):
-        bond_idxs_and_params = align_fn(
-            src_bond.potential.idxs,
-            src_bond.params,
-            dst_bond.potential.idxs,
-            dst_bond.params,
-        )
+# @pytest.mark.nocuda
+# def test_setup_intermediate_bonded_term(arbitrary_transformation):
```
proteneer, line 667: note to self: figure out how/if to re-add these some of these tests - though doesn't seem that useful to begin. It was testing an old ref implementation that is no longer relevant.

timemachine/fe/single_topology.py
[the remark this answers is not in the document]
maxentile: > The key distinction here is that an angle term doesn't have to fully convert to its end-state before turning on the torsion term, it just has to have a reasonable force constant.

That's interesting.

Mostly, I'm unsure if there are 5 independent barrier-location parameters in this schedule that all happen to == `0.7` at the moment (but which _could_ all vary independently in the future), or if there are fewer independent parameters. Also, maybe some of the `0.7` barriers need to be coordinated with the `0.3` barriers...

If the ~10 literals (excluding `0.0, 1.0`) that appear in the schedule are all independent already, or if the amount of "slack" in each barrier is pretty large, maybe no change to code or documentation is warranted...

---------------

[tangent] Out of scope for current PR, but do you think we can make sharper assertions about _individual terms_ (e.g. "force constant on bond (5,6) must be greater than some threshold before the force constant on torsion (1,4,5,6) is > some threshold") rather than whole groups of terms ("activate all dummy bonds before any dummy torsions")?

If we can develop a reasonably complete list of such local assertion / constraint expressions (maybe the list will have a lot of special cases for ring-opening...), maybe we can then generate a fine-grained (i.e. per-term) force-constant-interpolation schedule for any given transformation, satisfying the constraints. @mcwitt and I discussed something along these lines a while ago -- maybe this could be useful to flesh out in a later PR.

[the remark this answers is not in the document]
proteneer: > If the ~10 literals (excluding 0.0, 1.0) that appear in the schedule are all independent already, or if the amount of "slack" in each barrier is pretty large, maybe no change to code or documentation is warranted...

I do think documentation is warranted re: slack at the boundaries. I will add it. 

> [tangent] Out of scope for current PR, but do you think we can make sharper assertions about individual terms (e.g. "force constant on bond (5,6) must be greater than some threshold before the force constant on torsion (1,4,5,6) is > some threshold") rather than whole groups of terms ("activate all dummy bonds before any dummy torsions")?

I think that's reasonable, but it will require a lot of careful testing on small systems. Roughly, there's two classes of assertions:

1) Assertions for numerical stability (eg. angle terms going to 180 degrees) can lead to large forces in torsions, etc. This class of assertions would be dependent on the time-step, whether or not we use HMR, etc. other numerical details.
2) Assertions for "trapped" sampling (eg. chiral restraints), these are dependent on length of simulation, number of expected flips in a typical 2ns sampling run, whether or not we use HREX/REST, etc.

I don't know really know how to address them systematically right now.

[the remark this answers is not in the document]
proteneer: For the case of chiral volumes, I think there's a simple assertion we can make. I can set that up as an example.

timemachine/fe/single_topology.py
```diff
@@ -1638,94 +1690,47 @@ def setup_intermediate_state(self, lamb) -> VacuumSystem:
...
         chiral_bond = self._setup_intermediate_chiral_bond_term(
             src_system.chiral_bond,
             dst_system.chiral_bond,
             lamb,
             interpolate.linear_interpolation,
         )
 
-        return VacuumSystem(bond, angle, torsion, nonbonded, chiral_atom, chiral_bond)
+        vs = VacuumSystem(bond, angle, torsion, nonbonded, chiral_atom, chiral_bond)
+        assert_bonds_defined_for_chiral_volumes(vs, DEFAULT_BOND_IS_PRESENT_K)
```
proteneer, line 1730: @maxentile I added a function to verify that bonds are present for chiral volumes in the `setup_intermediate_state()` call - however, I'm running into a problem that this is incompatible with `jax.jit`

@mcwitt - any ideas? maybe checkify?

``` python
    def assert_bonds_defined_for_chiral_volumes(vs: VacuumSystem, bond_k_min: float):
        """
        Assert that bonds defined for every chiral volume is present and has a force constant greater than bond_k_min
        """
        bonds_present = set()
        for idxs, (bond_k,_) in zip(vs.bond.potential.idxs, vs.bond.params):
>           if bond_k > bond_k_min:
E           jax.errors.TracerBoolConversionError: Attempted boolean conversion of traced array with shape bool[]..
E           The error occurred while tracing the function U at /home/yzhao/Code/timemachine/tests/test_single_topology.py:641 for jit. This concrete value was not available in Python because it depends on the value of the argument lam.
E           See https://jax.readthedocs.io/en/latest/errors.html#jax.errors.TracerBoolConversionError
```

mcwitt: I think we have a pattern in a few other places where we add a `verify` argument (defaulting to true), and guard the assertions so they are only executed if `verify == True`. Then I think `jit(partial(f, verify=False))` or something like `jit(f, static_argnames=("verify",))` (and then passing `verify=False`) can be supported?

proteneer:
Introduced a new constant `DEFAULT_BOND_IS_PRESENT_K = 30.0` that is now referenced by both the plotting code, as well as assertion code indicating whether or not a bond is considered to be "present".

timemachine/fe/single_topology.py
mcwitt: Looking again, I only see one example of this in the current code (and the flag is called `runtime_validate`, not `verify`): https://github.com/proteneer/timemachine/blob/be34fbf458b54de5454564b67aeb1d70450a1b3f/timemachine/potentials/nonbonded.py#L228

I think `checkify` would work too, but the downside is I think it complicates usage without jax?

mcwitt: I'd also be OK with breaking jax transformability in the short term; I don't think this is used in production, and `jit` probably isn't the right long-term solution for SingleTopology slowness

timemachine/fe/single_topology.py
proteneer: I don't remember either

tests/test_interpolate.py
proteneer: let me clean this up a bit

tests/test_single_topology.py
proteneer: (yeah doesn't seem to make much sense to keep a version of this - will have to rewrite one from scratch).

timemachine/fe/single_topology.py
```diff
@@ -504,13 +559,22 @@ def concatenate(arrays, empty_shape, empty_dtype):
         # either 0 or 180, since the normalized chiral volume is still smooth wrt perturbations
         all_proper_dummy_chiral_atom_idxs_ = []
         all_proper_dummy_chiral_atom_params_ = []
+
+        print("all_dummy_chiral_atom_idxs", all_dummy_chiral_atom_idxs)
```
mcwitt, line 563: Remove debug print?

timemachine/fe/single_topology.py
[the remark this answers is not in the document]
proteneer: Added an assertion `assert_bonds_defined_for_chiral_volumes` to check for bonds.

We can add some more assertions later, eg:

1) checking that "naked pyramids" without chiral volumes defined are invertible (by asserting that angles are close to 120, low force constants, ring membership, presence of impropers, etc.)
2) checking that impropers are not conflicting with highly pyramidal structures (right now proper + impropers are collected into a single type for interpolation purposes, so it's hard to separate).

tests/test_interpolate.py
proteneer: re-used the `_assert_bonded_term` for this.

tests/test_interpolate.py
proteneer: obsolete, re-using `_assert_bonded_term` for this now.

tests/test_interpolate.py
proteneer: combined fwd/rev versions into a function, parameterized by "direction"

timemachine/fe/single_topology.py
proteneer: I'm going to punt on this for now, and just do the verification *outside* of the calling function.

timemachine/fe/single_topology.py
proteneer: removed

timemachine/fe/single_topology.py
```diff
@@ -544,13 +559,21 @@ def concatenate(arrays, empty_shape, empty_dtype):
...
+            missing_bonds = []
+            for x in [i, j, k]:
+                if (c, x) not in mol_c_bond_idxs_set and (x, c) not in mol_c_bond_idxs_set:
+                    missing_bonds.append((c, x))
+
+            if len(missing_bonds) == 0:
                 all_proper_dummy_chiral_atom_idxs_.append((c, i, j, k))
                 all_proper_dummy_chiral_atom_params_.append(p)
             elif verify:
+                # print("VERIFY", [(c, x) in mol_c_bond_idxs_set or (x, c) in mol_c_bond_idxs_set for x in [i, j, k]])
```
jkausrelay, line 573: remove debug?

jkausrelay (approved):

timemachine/fe/single_topology.py
proteneer: thanks - this snuck by. resolved - I modified the code so that the warnings are more informative (showing which bonds are broken in a chiral volume).

timemachine/fe/single_topology.py
mcwitt: (note: for harmonic bonds we use `k_min = 0.1`, bug agreed that this is probably too small to be useful for visualization
https://github.com/proteneer/timemachine/blob/be34fbf458b54de5454564b67aeb1d70450a1b3f/timemachine/fe/single_topology.py#L1657)

timemachine/fe/single_topology.py
proteneer: ugh good call - I actually screwed up the k_min here, should be 0.1 instead. I think net effect is that we were ramping up the bonds slower than it should've been doing (k_min=0.025 vs what should be k_min=0.1).

https://github.com/proteneer/timemachine/blob/master/timemachine/fe/single_topology.py#L1657

timemachine/fe/chiral_utils.py
```diff
@@ -145,7 +145,8 @@ def find_chiral_atoms(mol):
     chiral_patterns = [
         "[X4:1]",  # any tetrahedral atom
         "[#16X3,#15X3:1]",  # trivalent sulfur, phosphorous are assumed to be non-invertible
-        # "[#7X3:1](~[R])(~[R])~[R]",  # nitrogen directly bonded to three ring atoms  # TODO: handle pyramidal nitrogen
+        "[#7X3:1](~F)(~F)~F",  # highly explosive, non-invertible nitrogen used for testing
```
maxentile, line 148:
```suggestion
        "[#7X3:1](~F)(~F)~F",  # for software testing
```

(tangent: comment maybe referring to a different nitrogen trihalide? https://en.wikipedia.org/wiki/Nitrogen_trifluoride )

proteneer:
Rebased onto master

timemachine/fe/rbfe.py
```diff
@@ -94,6 +95,7 @@ def setup_in_env(
     """Prepare potentials, concatenate environment and ligand coords, apply HMR, and construct barostat"""
     barostat_interval = 25
     system = st.combine_with_host(host.system, lamb, host.num_water_atoms, st.ff, host.omm_topology)
+    assert_bonds_defined_for_chiral_volumes(system)
```
maxentile, line 98: Nice.

Perhaps in a later stage we could expand this into a function `validate_hybrid_params(system)` or something that includes this assertion (`HarmonicBond(i, j).force_constant > DEFAULT_BOND_IS_PRESENT_K` whenever a chiral volume includes center atom `i` and non-center atom `j`) in addition to other assertions, such as the ones you suggested in https://github.com/proteneer/timemachine/pull/1406/files#r1831720012 .

maxentile (approved):

mcwitt (approved):

PR #1407, opened 2024-10-21
Accept conformer id in utilities

mcwitt:
Small quality of life improvement to accept the target conformer ID as an optional argument in several utility functions.

Also adds `read_sdf_mols_by_name`, similar to `read_sdf` except returning a dict mapping name to mol.

### DIFF
--- a/timemachine/fe/utils.py
+++ b/timemachine/fe/utils.py
@@ -391,20 +391,20 @@ def get_romol_bonds(mol):
     return bond_list
 
 
-def get_romol_conf(mol) -> NDArray:
-    """Coordinates of mol's 0th conformer, in nanometers"""
-    conformer = mol.GetConformer(0)
+def get_romol_conf(mol, conf_id: int = 0) -> NDArray:
+    """Coordinates of the specified conformer, in nanometers"""
+    conformer = mol.GetConformer(conf_id)
     guest_conf = np.array(conformer.GetPositions(), dtype=np.float64)
-    return guest_conf / 10  # from angstroms to nm
+    return guest_conf / 10.0  # from angstroms to nm
 
 
-def set_romol_conf(mol, new_coords: NDArray):
+def set_romol_conf(mol, new_coords: NDArray, conf_id: int = 0):
     """Sets coordinates of mol's 0th conformer. Expects coords in nanometers and converts to angstrom"""
     assert new_coords.shape[0] == mol.GetNumAtoms()
     # convert from nm to angstroms
-    angstrom_coords = new_coords * 10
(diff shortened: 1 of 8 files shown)


timemachine/fe/utils.py
```diff
@@ -485,6 +485,12 @@ def read_sdf(fname: Union[str, Path], removeHs: bool = False) -> List[Chem.Mol]:
     return mols
 
 
+def read_sdf_mols_by_name(fname: str | Path, removeHs: bool = False) -> Dict[str, Chem.Mol]:
```
badisa, line 488: Would be good to add some usage in the code base or having a test of it.

timemachine/fe/utils.py
```diff
@@ -391,20 +391,20 @@ def get_romol_bonds(mol):
...
+    """Coordinates of the specified conformer, in nanometers"""
+    conformer = mol.GetConformer(conf_id)
     guest_conf = np.array(conformer.GetPositions(), dtype=np.float64)
-    return guest_conf / 10  # from angstroms to nm
+    return guest_conf / 10.0  # from angstroms to nm
 
 
-def set_romol_conf(mol, new_coords: NDArray):
+def set_romol_conf(mol, new_coords: NDArray, conf_id: int = 0):
     """Sets coordinates of mol's 0th conformer. Expects coords in nanometers and converts to angstrom"""
```
badisa, line 402: Update the docstring

```suggestion
    """Sets coordinates of the specified conformer. Expects coords in nanometers and converts to angstrom"""
```

badisa (approved):

timemachine/fe/utils.py
badisa, line 488: nit: dict is preferred over Dict (deprecated in 3.9) per https://docs.python.org/3/library/typing.html#typing.Dict

```suggestion
def read_sdf_mols_by_name(fname: str | Path, removeHs: bool = False) -> dict[str, Chem.Mol]:
```

maxentile (approved):

timemachine/fe/utils.py
mcwitt: Good catch on the deprecation. I might prefer to leave this as-is for now though for consistency, in favor of fixing all of the deprecations in a separate PR (e.g. it looks like `List`, `Tuple`, and `Set` are also deprecated)

timemachine/fe/utils.py
mcwitt: Fixed in https://github.com/proteneer/timemachine/pull/1407/commits/aa7abb7eb54b8a37f60ff6a6e7b506dd0505a64a

PR #1409, opened 2024-10-23
Remove OpenMM import

badisa:


jkausrelay (approved):

timemachine/fe/single_topology.py
```diff
@@ -39,6 +38,8 @@
 )
 from timemachine.utils import fair_product_2
 
+OpenMMTopology = TypeVar("OpenMMTopology")
```
mcwitt, line 41: nit: I think
```suggestion
OpenMMTopology = Any
```
might be more appropriate here. (I get a linter error from pyright in the usages of this below, since `TypeVar`s are typically used more than once in a function signature.

badisa: Switched to `Any` in https://github.com/proteneer/timemachine/pull/1409/commits/6544f611364cef8588302ff720d536d5bb86d945

mcwitt: > Should we switch to pyright? If we think pyright is more trustworthy than mypy think we should enforce it at the repo level.

I think I'd be in favor of this, though not certain what is considered best practice currently. In general I find that pyright's suggestions make more sense to me, and it's quite a bit faster.

PR #1410, opened 2024-10-23
store bond elements in EnvironmentBCCHandler

jkausrelay:
- Store Bond elements in EnvironmentBCCHandler to make it easier to see which parameters correspond to which elements

### DIFF
--- a/timemachine/ff/handlers/nonbonded.py
+++ b/timemachine/ff/handlers/nonbonded.py
@@ -606,6 +606,7 @@ class EnvironmentBCCHandler(SerializableMixIn):
         bond_idxs = []
         param_idxs = []
         signs = []
+        bond_elements = []
 
         # find typing information for each bond in the topology
         for src_atom, dst_atom in topology.bonds():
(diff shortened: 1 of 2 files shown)


timemachine/ff/handlers/nonbonded.py
```diff
@@ -619,6 +620,7 @@ def __init__(self, patterns, params, protein_ff_name, water_ff_name, topology):
                     # Skip waters
                     continue
                 bond_idxs.append((src_atom.index, dst_atom.index))
+                bond_elements.append((src_atom.element.symbol, dst_atom.element.symbol))
```
badisa, line 623: Should we use `element.number` instead? Would be more consistent with how we do other element checks (https://github.com/proteneer/timemachine/blob/master/timemachine/ff/handlers/nonbonded.py#L725-L731) and would probably also enable some more flexibility with using numpy arrays downstream

maxentile (approved):

badisa (approved):

PR #1413, opened 2024-10-25
Add resolve_intramol_clashes

maxentile:
Add a utility function that energy-minimizes a `mol`'s 0'th conformer, while otherwise remaining as close as possible to the input conformer.

For example, if `mol` has a pair of atoms placed on top of each other:

image

(resulting in |force(x)| of +inf)

`resolve_intramol_clashes(mol, ff)` will produce:

image

This is implemented by introducing a softened potential function `U(x, lam)`, with `lam` increasing the distances used in `NonbondedPairListPrecomputed`. At `lam=1` all intramolecular pairs `ij` have their effective distance `d_ij` increased to `>= 0.75 * sig_ij`, and at `lam=0` we recover the original `U(x)`.

A restrained minimization is performed at `lam=1`, then another restrained minimization is performed at `lam=0`.

### DIFF
--- a/tests/test_minimizer.py
+++ b/tests/test_minimizer.py
@@ -399,7 +399,7 @@ def test_minimizer_failure_toy_system():
 
 
 def test_resolve_intramol_clashes():
-    """start from a conformer where |force| is +inf"""
+    """start from a conformer where |force(x.astype(f32))| is +inf"""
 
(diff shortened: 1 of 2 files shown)


maxentile:
Forgot this was still open -- might revisit if needed

PR #1393, opened 2024-09-25
Perform restrained minimization on InitialStates

badisa:
* Address ligand poses moving too far from the original poses.

PR #1414, opened 2024-10-31
Fixes incorrect print in minimizer

badisa:
* Was printing out the wrong atom's forces in the logging for the atom with the largest force norm. Introduced in https://github.com/proteneer/timemachine/pull/1393
* Correctly assigns float64 as the return value, avoiding warnings if 64bit is not enabled in JAX. Introduced in https://github.com/proteneer/timemachine/pull/1405


timemachine/md/minimizer.py
```diff
@@ -525,7 +525,7 @@ def scipy_minimize(
     def val_and_grad_fn_bfgs(x_flattened):
         x = x_flattened.reshape(final_shape)
         u, grad_full = val_and_grad_fn(x)
-        return u, grad_full.reshape(-1).astype(np.float64)
+        return u, np.asarray(grad_full.reshape(-1)).astype(np.float64)
```
badisa, line 528: Without `jax.config.update("jax_enable_x64", True)` this would trigger tons of errors. Also has an impact on the results of BFGS, my current hypothesis is that passing float32 to BFGS (https://github.com/scipy/scipy/blob/main/scipy/optimize/_optimize.py#L1314) is losing enough precision to produce worse results.

badisa: Also without this change leads to failures if using `L-BFGS-B` when not enabling 64 bit.

maxentile (approved):

proteneer (approved):

PR #1415, opened 2024-10-31
Best-first MCS search

mcwitt:
Modifies the search strategy in the maximum common substructure (MCS) search (used during atom mapping) to greatly reduce the number of nodes that must be searched in certain problem instances.

## Background

The current approach is based on [McGregor 1982](https://doi.org/10.1002/spe.4380120103), and proceeds by depth-first search over all possible mappings `A -> Optional[B]`.

The tree is pruned according to an upper bound on the number of edges that are required to be preserved by the mapping (in the sense that if a1 maps to b1 and a2 maps to b2, then `adj(a1, a2) == adj(b1, b2)`) as follows:

1. Initially, we search for a mapping that preserves the maximum possible number of edges, i.e. set `min_num_edges = min(edges in A, edges in B)`. During DFS, we prune branches whose `num_edges_upper_bound` is less than `min_num_edges`.
2. If no mapping is found, we decrement `min_num_edges` by one and restart DFS.
3. Repeat (2) until a mapping is found.

This strategy, while efficient in cases where the number of edges preserved by an optimal mapping is close `min(edges in A, edges in B)`, requires searching a large number of nodes when this is not the case.

## Changes

This PR changes the search strategy as follows:

1. Removes the outer loop over decreasing `min_num_edges`. By itself, this change is catastrophically inefficient, since we no longer have any mechanism for pruning the search tree.
2. Keeps track of the leaf node with the largest number of preserved edges during the search (as in McGregor 1982); this is used to prune branches of the tree according to their upper bound. This recovers most of the performance lost in (1), but is still 2-3x slower compared with the existing approach according to small-molecule benchmarks. Note that to guarantee optimality of the solution (in the sense of number of edges preserved), we must perform an exhaustive search over all leaf nodes; only then can we determine the optimal subset.
3. Modifies the search ordering from depth-first to a [best-first](https://en.wikipedia.org/wiki/Best-first_search) variant. We use as a heuristic the same upper bound on the number of edges preserved by the mapping as above. Note that this ordering guarantees that the first node returned is optimal; this could be exploited further (not in this PR) by incorporating more of the post-search ranking heuristics into the search heuristic, allowing termination the search when the first solution is found and potentially further performance gains.

## Validation

* Added a golden test to ensure that atom mappings produced by the new version are identical to those produced by the existing version (https://github.com/proteneer/timemachine/pull/1415/commits/67e363bd1efc507c2558508ffb5a4efec606c61b)
* Verified that ~core sizes are unchanged~ mappings are identical on large internal repository of examples

## Public FEP Benchmarks

a = master
b = this branch

For each instance, mapping generated by B verified to be identical to that generated by A.

|       |            a |            b |
|:------|-------------:|-------------:|
| count | 1301         | 1301         |
| mean  |    0.132628  |    0.116224  |
| std   |    0.284466  |    0.127008  |
| min   |    0.0223501 |    0.0273046 |
| 25%   |    0.0658681 |    0.0691361 |
| 50%   |    0.0887287 |    0.0941987 |
| 75%   |    0.11967   |    0.119321  |
| max   |    5.44522   |    1.95834   |


mcwitt:
> Also, `atom_mapping.py` is not currently in the diff, but I think there may be points in this documentation that should be updated:

Updated in https://github.com/proteneer/timemachine/pull/1415/commits/11610994e827997ab78cc778d596a3fc4b19987b and https://github.com/proteneer/timemachine/pull/1415/commits/65fc2aad961a3f03645779921ce65cf2a7eff014

timemachine/fe/tree_search.py
```diff
@@ -0,0 +1,57 @@
...
+    while queue:
+        node = heapq.heappop(queue)
+        children = expand(node)
+        yield node
+        for child in children:
+            heapq.heappush(queue, child)
+
+
+def best_first_stateful(
+    expand: Callable[[Node, State], tuple[Sequence[Node], State]], root: Node, initial_state: State
```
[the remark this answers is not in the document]
mcwitt: Changed in https://github.com/proteneer/timemachine/pull/1415/commits/6975b25886163e37f03d552a640d3ab7d2fbb48b

timemachine/fe/tree_search.py
```diff
@@ -0,0 +1,57 @@
...
+) -> Iterator[Node]:
+    """Stateful variant of :py:func:`best_first`.
+
+    Compared with the latter, this accepts an augmented `expand` function that updates the global search state in
+    addition to producing the children of a given node, and an initial state.
```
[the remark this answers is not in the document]
mcwitt: Was on the fence about this too; the `best_first` variant was intended for pedagogical purposes in case it helped to understand the more complicated type of `best_first`. But I agree that this could be equally well addressed with a comment in the docs. Removed in https://github.com/proteneer/timemachine/pull/1415/commits/6975b25886163e37f03d552a640d3ab7d2fbb48b

timemachine/fe/mcgregor.py
```diff
@@ -244,13 +167,191 @@ def to_networkx(self) -> nx.Graph:
...
-    for neighbors in priority_list:
-        cur_layer_size *= len(neighbors)
-        layer_sizes.append(cur_layer_size)
-    return sum(layer_sizes)
+@dataclass(frozen=True)
+class Marcs:
+    marcs: NDArray[np.bool_]
+
+    @cached_property
+    def num_edges_upper_bound(self) -> int:
```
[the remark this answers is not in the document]
mcwitt: Added a note in https://github.com/proteneer/timemachine/pull/1415/commits/db8391e13a4e014d481a75eea5725e750cddbb89. Personally I find "arcs left" to be confusing since it somehow suggests to me that it does not include the edges that are already in correspondence at a given point in the search, only the ones "left" to possibly map. (I think it's just using "left" in a different sense that I first assumed.)

timemachine/fe/mcgregor.py
```diff
@@ -303,242 +396,161 @@ def mcs(
...
+            f"Exceeded max number of visits/cores - no valid cores could be found: {mcs_result.nodes_visited} nodes visited."
+        )
+    # else:
+    # print(
+    # f"==FAILED==[NODES VISITED {mcs_result.nodes_visited} | time taken: {time.time()-start_time} | time out? {mcs_result.timed_out}]====="
+    # )
+
     if len(mcs_result.all_maps) == 0:
-        raise NoMappingError("Unable to find mapping")
+        raise NoMappingError(f"Unable to find mapping with at least {min_num_edges} edges")
```
proteneer, line 467: Hm - this looks like it removes the "fast termination" feature we had before when we specified a minimum value for the required # of edges. (i.e. before we can quickly exit in the outer layer, now we only exit *after* the search is complete).

```
if cur_threshold < min_threshold:
    raise NoMappingError(f"Unable to find mapping with at least {min_threshold} edges")
```

IIRC @badisa used this feature for graph generation at some point - is it still being used?

timemachine/fe/mcgregor.py
```diff
@@ -303,242 +396,161 @@ def mcs(
...
+                }
+                if g2.mapping_incompatible_with_cc_constraints(
+                    g2_mapped_nodes, g2_unvisited_nodes, max_connected_components, min_connected_component_size
+                ):
+                    return False
+
+        return True
+
+    def expand(node: Node, best_num_edges: int) -> Tuple[List[Node], int]:
+        if node.marcs.num_edges_upper_bound < best_num_edges:
```
proteneer, line 525: does it make sense to also check `node.marcs.num_edges_upper_bound < min_num_edges`? i.e. return if the predicate is `True`

timemachine/fe/mcgregor.py
```diff
@@ -244,13 +167,183 @@ def to_networkx(self) -> nx.Graph:
...
+        return Node(
+            self.atom_map.add(self.layer, new_v2),
+            self.marcs.refine(g1, g2, self.layer, new_v2),
+            self.layer + 1,
+        )
+
+    def skip(self, g1: Graph, g2: Graph) -> "Node":
+        return Node(
+            self.atom_map,
+            self.marcs.refine(g1, g2, self.layer, UNMAPPED),
```
proteneer, line 283: IIRC there was one optimization I'd forgotten to incorporate earlier on but might be easy to do now:

We can tell when a node in `g1` is `UNMAPPED` because we've skipped past the current `layer`. However, we don't do any marking of nodes in `g2` to the `UNMAPPED` state - but this can be easily determined by checking the "missing" nodes from the union of the priority idxs, as in we can set:

`[self.marcs.refine(g1, g2, UNMAPPED, v2) for v2 in nodes_missing_from_priority_idxs]`

timemachine/fe/mcgregor.py
```diff
@@ -244,13 +167,183 @@ def to_networkx(self) -> nx.Graph:
...
+        )
+
+    @cached_property
+    def is_leaf(self):
+        return self.layer == len(self.atom_map.a_to_b)
+
+    @cached_property
+    def priority(self):
+        """Compute the priority of this node. By convention, lowest numerical value is highest priority"""
+        return (-self.marcs.num_edges_upper_bound, -self.layer)
```
proteneer, line 294: Interesting - I wonder if we can tie-break the equal priority further, eg:

```
(-self.marcs.num_edges_upper_bound, -self.marcs.num_of_ones, +self.marcs.num_connected_components, -self.layer)
```

proteneer:
this PR is :fire::fire::fire::fire::fire:

timemachine/fe/mcgregor.py
mcwitt: If I understand correctly, the early termination behavior is roughly the same (note that `min_threshold` was renamed `min_num_edges`). Because `best_num_edges` is initialized to `min_num_edges` (https://github.com/mcwitt/timemachine/blob/3137fad307d23ba70f83b265f17bcfbd711b85b8/timemachine/fe/mcgregor.py#L454), the search is pruned in the same way it would be during the initial iteration of DFS in the existing code, so I don't think the changes here result in any more nodes being searched in this case.

timemachine/fe/mcgregor.py
mcwitt: Since `best_num_edges` is initialized to `min_num_edges`, `min_num_edges <= best_num_edges` is an invariant, so I think this would be redundant.

timemachine/fe/mcgregor.py
mcwitt: > the search is pruned in the same way it would be during the initial iteration of DFS in the existing code

Hmm, thinking about it more, that's not quite right. It seems like the behavior in this PR is to search _fewer_ nodes than the current version in the case where `min_num_edges` rules out any possible mappings. This is because in this PR we always prune according to `min_num_edges`, versus the current behavior that iterates down from `min(num_edges_a, num_edges_b)` to `min_num_edges`. Does that make sense?

proteneer: ah, yes, I think that makes sense. Thank you for the clarification.

proteneer (approved):
lgtm!

timemachine/fe/mcgregor.py
mcwitt: Interesting idea to incorporate the number of connected components. I want to play around with this, but will leave for a future PR as I think this will require additional bookkeeping to be fast enough.

timemachine/fe/mcgregor.py
mcwitt: I played around with this and did see a reduction in nodes visited for some examples, but the effect wasn't dramatic enough to offset the cost of my (naive, unoptimized) implementation. I want to explore this more, but will defer to a future PR.

Added TODO comments in https://github.com/proteneer/timemachine/pull/1415/commits/e409bf2a6179eb22930e5f1ca1bcbf82bc072a6f.

PR #1419, opened 2024-11-08
Fix nightly tests.

proteneer:
Fixes zero length array shapes and types in one of the end-state symmetry nightly tests.
Update water sampling example files since buckyball energies were affected by the improper torsion change.


tests/test_single_topology.py
```diff
@@ -1708,14 +1708,15 @@ def test_hif2a_end_state_symmetry_nightly_test():
...
                     np.array(
                         [
                             [x, y]
                             for x, y in zip(amm_fwd.b_to_c, amm_rev.a_to_c)
                             if x not in core[:, 0] and y not in core[:, 1]
-                        ]
-                    ),
+                        ],
+                        dtype=np.int32,
+                    ).reshape(-1, 2),
```
maxentile, line 1719: Note that a similar `fused_map_from(amm_fwd, amm_rev)` snippet appears in another unit test: 

https://github.com/proteneer/timemachine/blob/70bec1d7db4d0ccc0ab7651f81b5d3990e4446d4/tests/test_single_topology.py#L1741-L1748

This zero-length corner case isn't encountered in that unit test, but it might still be good to extract a single copy of this utility later.

maxentile (approved):

tests/test_single_topology.py
mcwitt, line 1719: nit/non-blocking: I think this could be simplified by moving the array construction to the outside:

```python
            core_map = [(x, y) for x, y in zip(amm_fwd.a_to_c, amm_rev.b_to_c)]
            dummy_map = [
                (x, y) for x, y in zip(amm_fwd.b_to_c, amm_rev.a_to_c) if x not in core[:, 0] and y not in core[:, 1]
            ]
            fused_map = np.array(core_map + dummy_map).reshape(-1, 2)
```

mcwitt (approved):

proteneer:
Going to merge this as-is - currently over whelmed with zero sized lists in other areas of the code. I'll come back to this later when I have more bandwidth since comments indicate non-blocking.

PR #1420, opened 2024-11-08
Code Clean-Up and split proper + improper terms.

proteneer:
This PR cleans up a lot of the cruft in the current code as I was trying to split the torsions into proper + improper idxs. 

1) Splits `improper` and `proper` torsions from a single `torsion` array - this will allow us to fine-tune the alignment and the interpolation logic in a separate PR. This *can* change expected results for single topology simulations: since we no longer allow an proper torsion to be accidentally aligned and interpolated into an improper torsion - they are now strictly separated. For all other types of simulations, (eg. vanilla MD, AHFE, water sampling etc.) results should be bitwise identical.
2) Replace all instances of `cutoff=1.2` with `constants.DEFAULT_NONBONDED_CUTOFF`, (tbd: change erfc's `alpha` to `DEFAULT_ALPHA` as well)
3) Removes the cruft of `convert_omm_system` by re-using the `VacuumSystem` container more systematically throughout the code - allowing us to directly grab the correct bonded terms unambiguously. 

``` python
# before
system, conf, box, top = builders.build_water_system(4.0, ff.water_ff)
bps, _ = openmm_deserializer.deserialize_system(system, cutoff=DEFAULT_NONBONDED_CUTOFF)
nb = get_bound_potential_by_type(bps, Nonbonded)
bond_pot = get_bound_potential_by_type(bps, HarmonicBond).potential

all_group_idxs = get_group_indices(get_bond_list(bond_pot), conf.shape[0])
```

``` python
# after
system, conf, box, _ = builders.build_water_system(4.0, ff.water_ff)
named_system, _ = openmm_deserializer.deserialize_system(system, cutoff=DEFAULT_NONBONDED_CUTOFF)
nb = named_system.nonbonded
bond = named_system.bond
all_group_idxs = get_group_indices(get_bond_list(bond.potential), conf.shape[0])
```
4) `VacuumSystem` no longer has any `Optional` fields. This allows removal of assertion logic in many locations (eg. `assert system.proper` in setup code).
5) `Potential` classes that have an `idxs` field (eg. `HarmonicBond`, `NonbondedPairlistPrecompute`, etc.) now automatically reshapes zero-length idxs (both lists and np arrays) to the correct `dim` via a `__post_init__()` hook. For example, if an empty list is passed in to `HarmonicBond([])`, `__post_init__()` will reshape idxs to `(-1,2) -> (0,2)`. In addition, the `dtype` will be fixed to an `int32`. 
6) `BoundPotential` classes now automatically reshapes zero-length params (`np` or `jnp` arrays are okay, but `list`s are disallowed) to the appropriate shape, dimension, and `dtype` (always set to `np.float64`).


timemachine/potentials/potential.py
```diff
@@ -48,6 +49,16 @@ def __call__(self, conf: Conf, box: Optional[Box]) -> float | Array:
...
 
+    def __post_init__(self):
+        # if isinstance(self.params, list) and len(self.params) == 0:
+        #     # untraced
+        #     self.params = np.array(self.params, dtype=np.float64)
+        # else:
+        #     print(self.params)
+        #     self.params = self.params.astype(np.float64)
+    
+        self.params = jnp.array(self.params, dtype=np.float64).reshape(self.potential.params_shape())
```
proteneer, line 60: @mcwitt I tried to post init all the params from bound potentials so that they're well-formed, well-shaped jax potentials, but this ended-up breaking all of the interpolation code

proteneer: eg: pytest -xsv tests/test_single_topology_combined.py::test_nonbonded_host_params_independent_of_lambda

results in:

```
tests/test_single_topology_combined.py:232:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/test_single_topology_combined.py:230: in get_nonbonded_host_params
    return st.combine_with_host(host_sys, lamb, num_water_atoms, st.ff, omm_topology).nonbonded_host.params
timemachine/fe/single_topology.py:1968: in combine_with_host
    guest_system = self.setup_intermediate_state(lamb=lamb)
timemachine/fe/single_topology.py:1722: in setup_intermediate_state
    proper = self.align_and_interpolate_propers(lamb)
timemachine/fe/single_topology.py:1659: in align_and_interpolate_propers
    return self._align_and_interpolate_bonded_term(
timemachine/fe/single_topology.py:1625: in _align_and_interpolate_bonded_term
    bonded_params.append(interpolate_fn(idxs, src_params, dst_params, lamb))
timemachine/fe/single_topology.py:1595: in _interpolate_torsion
    if self.all_idxs_belong_to_core(idxs):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <timemachine.fe.single_topology.SingleTopology object at 0x7fb02f8db460>, idxs = Traced<ShapedArray(float64[])>with<DynamicJaxprTrace(level=1/0)>

    def all_idxs_belong_to_core(self, idxs):
        core_atoms = self.get_core_atoms()
        print(core_atoms)
        print(idxs)
>       return all([x in core_atoms for x in idxs])
E       TypeError: iteration over a 0-d array
```

tests/test_bonded.py
```diff
@@ -389,7 +389,7 @@ def test_chiral_atom_restraint(self, n_particles=64, n_restraints=35, dim=3):
         for precision, rtol in relative_tolerance_at_precision.items():
             self.compare_forces(x, params, box, potential, potential.to_gpu(precision), rtol)
 
-            with self.assertRaises(Exception):
+            with self.assertRaises(ValueError):
```
proteneer, line 392: (ValueErrors are raised by the wrapper code).

tests/test_free_energy.py
```diff
@@ -486,6 +485,7 @@ def result_with_overlap(overlap):
         assert len(results) == 1 + 2
 
 
+@pytest.mark.skip(reason="BUGGY")
```
proteneer, line 488: (Todo: this fails locally for me, but passes on CI. The correct exception is raised but is not caught by the assertion catcher)

tests/test_relative_free_energy.py
```diff
@@ -99,8 +99,8 @@ def check_pair_bar_result(res: PairBarResult):
             assert res.overlap_by_component_by_lambda.shape[0] == n_pairs
             assert res.overlap_by_component_by_lambda.shape[1] == res.dG_err_by_component_by_lambda.shape[1]
             for overlaps in [res.overlaps, res.overlap_by_component_by_lambda]:
-                assert np.all(0.0 < np.asarray(overlaps))
-                assert np.all(np.asarray(overlaps) < 1.0)
+                assert np.all(0.0 <= np.asarray(overlaps))
+                assert np.all(np.asarray(overlaps) <= (1.0 + 1e-5))  # need to add a small epsilon for loss of precision
```
proteneer, line 103: This fails on master for me locally as well - not sure why, but guessing loss of precision.

tests/test_single_topology_combined.py
```diff
@@ -5,9 +5,12 @@
 import pytest
 
 from timemachine import potentials
+
+# from timemachine.fe.system import convert_omm_system
```
proteneer, line 9: todo: remove

timemachine/fe/plots.py
```diff
@@ -590,12 +590,12 @@ def _plot_chiral_atom_interpolation(xs, systems, filter_fn, axs, row):
     # axs[row, 0].legend()
 
 
-def _plot_torsion_interpolation(xs, systems, filter_fn, axs, row):
+def _plot_proper_interpolation(xs, systems, filter_fn, axs, row):
```
proteneer, line 593: (dedup and unify with other bonded plotting code later)

timemachine/potentials/potentials.py
```diff
@@ -49,46 +73,76 @@ def __call__(self, conf: Conf, params: Params, box: Optional[Box]) -> float | Ar
...
+    def idxs_shape(self):
+        return (-1, 2)
+
+    def params_shape(self):
+        return (-1, 3)
+
 
 @dataclass
-class PeriodicTorsion(Potential):
+class PeriodicTorsion(ShapedIdxsPotential):
```
badisa, line 134: Should we add a Proper/Improper set of classes so we can identify these in a list of Potentials? Else you need the VacuumSystem to tell which is which (or munging parameters)?

timemachine/potentials/potentials.py
```diff
@@ -13,29 +13,53 @@
 from .types import Box, Conf, Params
 
 
+# classes that inherit from this will sanitize idxs to the expected shape and type
+class ShapedIdxsPotential(Potential):
+    def __post_init__(self):
+        self.idxs = np.array(self.idxs, dtype=np.int32).reshape(self.idxs_shape())
```
badisa, line 19: nit:

```suggestion
        self.idxs = np.asarray(self.idxs, dtype=np.int32).reshape(self.idxs_shape())
```

timemachine/potentials/potential.py
```diff
@@ -48,6 +48,9 @@ def __call__(self, conf: Conf, box: Optional[Box]) -> float | Array:
     def to_gpu(self, precision: Precision) -> "BoundGpuImplWrapper":
         return self.potential.to_gpu(precision).bind(np.asarray(self.params))
 
+    def __post_init__(self):
+        self.params = self.params.astype(np.float64).reshape(self.potential.params_shape())
```
badisa, line 52: What is the intention of casting the parameters to float64? Do we want to silently convert parameters to float64?

timemachine/potentials/potentials.py
```diff
@@ -227,6 +302,12 @@ def __call__(self, conf: Conf, params: Params, box: Optional[Box]) -> float | Ar
         )
         return jnp.sum(vdW) + jnp.sum(electrostatics)
 
+    def idxs_shape(self):
+        return (-1, 2)
+
+    def params_shape(self):
+        return (-1, 4)  # q s e w
```
badisa, line 309: nit:
```suggestion
        return (-1, 4)  # q sig eps w
```

tests/test_bonded.py
badisa: Where is this raised exactly, don't see any ValueError being raised explicitly?

timemachine/fe/single_topology.py
```diff
@@ -690,37 +690,22 @@ def setup_end_state(ff, mol_a, mol_b, core, a_to_c, b_to_c, dummy_groups: Dict[i
     mol_c_angle_idxs = np.concatenate([mol_a_angle_idxs, all_dummy_angle_idxs])
     mol_c_angle_params = np.concatenate([mol_a_angle_params, all_dummy_angle_params])
 
-    mol_c_proper_idxs = mol_a_proper_idxs
+    mol_c_proper_idxs = np.array([canonicalize_bond(idxs) for idxs in mol_a_proper_idxs])
```
mcwitt, line 693: nit: could this use the vectorized `canonicalize_bonds` for efficiency?

timemachine/potentials/potentials.py
proteneer: I thought about this but decided against it. I view the classes in this file to be "functional forms" (and shouldn't be carrying semantics specific to the system of interest).

timemachine/potentials/potentials.py
proteneer: willfix

timemachine/potentials/potential.py
proteneer: I think

1) all of our parameters on GPU implementation are 64 bit - regardless of `_f32` or `_f64`
2) up-casting should always be safe (but it can make things slower)

mcwitt, line 52: Have we verified that this doesn't break jax traceability of code that goes through `BoundPotential`?

I think my personal preference would be to assert on the shape here rather than automagically reshape (since I think it's useful to enforce shape discipline at the call sites), but could be persuaded otherwise.

timemachine/potentials/potentials.py
badisa: The specifics of the system seem valuable, whether or not it should be defined in this file. Mostly thinking down the line where we want to work with a list of potentials, where we don't have the `VacuumSystem` object (such as pickling InitialState).

Not a strong preference for this, just thinking of the future.

timemachine/potentials/potential.py
badisa: We are already up-casting when we go to the GPU implementation. But we lose the original dtype that the potential was constructed with, which seems undesirable behavior. 

Also this potentially throws a bunch of warnings in jax if `self.params` is a jax array and 64 bit is not enabled.

proteneer: iirc - it's the other way around. Assertions on param shapes can trigger jax traceability exceptions (since they're not traceable).

proteneer: >  Have we verified that this doesn't break jax traceability of code that goes through BoundPotential?

(PS this shouldn't break jax traceability - if `self.params` is a `jax.TracedArray`, the methods `astype` and `reshape` are defined methods on the `jax.TracedArray` object).

badisa:
Could you post the benchmarks on master and with this change? Would be good to verify the impact is minimal.

proteneer:
Closing this - will revisit at a later point in time. I cherry picked just the proper + improper torsion split into a separate PR.

tests/test_bonded.py
proteneer: For posterity: this was raised in the `__post_init__()` hooks

PR #1421, opened 2024-11-12
Enables minimizing system with only parts of system restrained

badisa:
* Adds support for specifying a subset of the local idxs as being restrained in `minimizer.local_minimize`.
* Passes the ligand interacting atoms as the restrained atoms rather than restraining all of the free indices. Originally had tried restraining heavy atoms in https://github.com/proteneer/timemachine/pull/1421/commits/02786256cf88be678e8142b68319352da1ca012f but code is a bit janky and didn't resolve any additional failures. Where hydrogen atoms clash and trigger failures, line search causes the ligand to blow up even if the hydrogens are unrestrained. @maxentile is working on a minimization approach that will resolve this in the future.

### DIFF
--- a/timemachine/md/minimizer.py
+++ b/timemachine/md/minimizer.py
@@ -550,6 +550,7 @@ def local_minimize(
     verbose: bool = True,
     assert_energy_decreased: bool = True,
     restraint_k: Optional[float] = None,
+    restrained_idxs: Optional[NDArray] = None,
 ):
     """
     Minimize a local region given selected idxs.
@@ -577,10 +578,15 @@ def local_minimize(
     assert_energy_decreased: bool
         Throw an assertion if the energy does not decrease
 
-    restraint_k float, optional
+    restraint_k: float, optional
         Restraint k to wrap val_and_grad_fn in a positional harmonic restraint to the input positions of the local_idxs.
         If None, minimize with no positional restraint. Refer to `timemachine.potentials.bonded.harmonic_positional_restraint`
         for implementation.
+
+    restrained_idxs: np.ndarray, optional
+        A subset of idxs to restrain, must be a subset of local_idxs. If not provided and
+        restraint_k is not None, all local_idxs are restrained.
+
 
     Returns
     -------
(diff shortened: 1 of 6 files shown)


maxentile (approved):

timemachine/md/minimizer.py
```diff
@@ -578,11 +581,16 @@ def local_minimize(
     assert_energy_decreased: bool
         Throw an assertion if the energy does not decrease
 
-    restraint_k float, optional
+    restraint_k: float, optional
```
proteneer, line 584: Not sure how to think about `None`support here - is it equivalent to 0?

timemachine/md/minimizer.py
```diff
@@ -578,11 +581,16 @@ def local_minimize(
...
         Restraint k to wrap val_and_grad_fn in a positional harmonic restraint to the input positions of the local_idxs.
         If None, minimize with no positional restraint. Refer to `timemachine.potentials.bonded.harmonic_positional_restraint`
         for implementation.
 
+    restrained_idxs: np.ndarray, optional
+        A subset of idxs to restrain, must be a subset of local_idxs. If not provided and
```
proteneer, line 590: I'd simplify this (and the double negatives) 

If None then will be set to `local_idxs`.

timemachine/md/minimizer.py
```diff
@@ -551,6 +553,7 @@ def local_minimize(
     verbose: bool = True,
     assert_energy_decreased: bool = True,
     restraint_k: Optional[float] = None,
```
proteneer, line 555: Preference is to avoid support for `None` - should be equivalent to `restraint_k: Optional[float] = 0.0`

proteneer (approved):
minor comment re: simplifying `restraint_k` + comments but otherwise lgtm.

timemachine/md/minimizer.py
badisa: Defaults to 0.0 and remove optional in https://github.com/proteneer/timemachine/pull/1421/commits/20c2ee2b5363a5fe893bce3da4f1db3436a41a77

timemachine/md/minimizer.py
badisa: Fixed in https://github.com/proteneer/timemachine/pull/1421/commits/20c2ee2b5363a5fe893bce3da4f1db3436a41a77

PR #1422, opened 2024-11-12
Split of propers and impropers from torsions, add ST symmetry test.

proteneer:
This PR:

1. Splits improper and proper torsions from a single torsion array - this will allow us to fine-tune the alignment and the interpolation logic in a separate PR. This can change expected results for single topology simulations: since we no longer allow an proper torsion to be accidentally aligned and interpolated into an improper torsion - they are now strictly separated. For all other types of simulations, (eg. vanilla MD, AHFE, water sampling etc.) results should be bitwise identical.
2. Removes `Optional[]` tags from the `[Vacuum/HostGuest]System` classes, enforcing a stronger requirement upfront and removing the need to check for `None` downstream.   
3. Adds a strong symmetry check for consistency of energies, forces, and the *definition* of the potentials themselves. Roughly, we test that `ST(mol_a, mol_b, lamb) == ST(mol_b, mol_a, 1-lamb)`.

### DIFF
--- a/tests/test_interpolate.py
+++ b/tests/test_interpolate.py
@@ -384,8 +384,6 @@ $$$$""",
             if key not in counts_kv_src:
                 counts_kv_src[key] = 0
             counts_kv_src[key] += 1  # store period
-        else:
-            assert 0
 
     assert counts_kv_src[duplicate_proper_idxs] == 2
 
@@ -396,8 +394,6 @@ $$$$""",
             if key not in counts_kv_dst:
                 counts_kv_dst[key] = 0
             counts_kv_dst[key] += 1  # store period
-        else:
-            assert 0
 
     assert duplicate_proper_idxs not in counts_kv_dst
 
--- a/tests/test_interpolate.py
+++ b/tests/test_interpolate.py
@@ -123,7 +123,40 @@ def test_align_proper():
     assert test_set == ref_set
 
 
-# TBD: implement a test_align_improper
+def test_align_improper():
+    """
+    Test that we can align idxs and parameters correctly for improper torsions.
+
+    Currently, improper torsions all have period == 2 and phase == pi
+    """
+    a, b, c, d, e, f, g, h, i, j, k, l, m, n = np.random.rand(14)
+
+    src_idxs = [(0, 1, 2, 3), (2, 1, 3, 0), (3, 1, 0, 2)]
+    src_params = [(a, b, 2), (c, d, 1), (e, f, 3)]
+    dst_idxs = [(2, 1, 3, 0), (3, 1, 0, 2), (0, 1, 2, 3)]
+    dst_params = [(i, b, 2), (j, k, 1), (l, f, 3)]
+
+    test_set = interpolate.align_improper_idxs_and_params(src_idxs, src_params, dst_idxs, dst_params)
+    ref_set = {
+        ((0, 1, 2, 3), (a, b, 2), (l, f, 3)),
+        ((2, 1, 3, 0), (c, d, 1), (i, b, 2)),
+        ((3, 1, 0, 2), (e, f, 3), (j, k, 1)),
+    }
+
+    assert test_set == ref_set
+
+    src_idxs = [(0, 1, 2, 3)]
+    src_params = [(a, b, 2)]
+    dst_idxs = [(0, 1, 3, 2)]
+    dst_params = [(i, c, 2)]
+
+    test_set = interpolate.align_improper_idxs_and_params(src_idxs, src_params, dst_idxs, dst_params)
+    ref_set = {
+        ((0, 1, 2, 3), (a, b, 2), (0, b, 2)),
+        ((0, 1, 3, 2), (0, c, 2), (i, c, 2)),
+    }
+
+    assert test_set == ref_set
 
 
 def test_align_nonbonded():
--- a/tests/test_interpolate.py
+++ b/tests/test_interpolate.py
@@ -90,10 +90,10 @@ def test_align_harmonic_angle():
     assert test_set == ref_set
 
 
-def test_align_torsion():
+def test_align_proper():
     """
-    Test that we can align idxs and parameters correctly for periodic torsions.
-    Periodic torsions differ from bonds and angles in that their uniqueness is
+    Test that we can align idxs and parameters correctly for proper torsions.
+    Proper torsions differ from bonds and angles in that their uniqueness is
     also determined by the "period", which is currently encoded as one of the parameters.
 
     We expect that decoupled terms have their force constants turned set to zero,
@@ -109,7 +109,7 @@ def test_align_torsion():
     dst_idxs = [(2, 3, 9, 4), (2, 3, 9, 4), (0, 1, 4, 2), (3, 0, 2, 6)]
     dst_params = [(i, b, 2), (j, k, 1), (l, f, 3), (m, n, 4)]
 
-    test_set = interpolate.align_torsion_idxs_and_params(src_idxs, src_params, dst_idxs, dst_params)
+    test_set = interpolate.align_proper_idxs_and_params(src_idxs, src_params, dst_idxs, dst_params)
 
     ref_set = {
         ((2, 3, 9, 4), (a, b, 2), (i, b, 2)),
@@ -121,6 +121,9 @@ def test_align_torsion():
     }
 
     assert test_set == ref_set
+
+
+# TBD: implement a test_align_improper
 
 
 def test_align_nonbonded():
@@ -339,27 +342,27 @@ $$$$""",
     ff = Forcefield.load_default()
     st = SingleTopology(mol_a, mol_b, core, ff)
 
-    duplicate_torsion_idxs = (0, 1, 3, 4)
+    duplicate_proper_idxs = (0, 1, 3, 4)
 
     counts_kv_src = dict()
-    for idxs, p in zip(st.src_system.torsion.potential.idxs, st.src_system.torsion.params):
+    for idxs, p in zip(st.src_system.proper.potential.idxs, st.src_system.proper.params):
         key = tuple(idxs)
         if p[2] == 2.0:
             if key not in counts_kv_src:
                 counts_kv_src[key] = 0
             counts_kv_src[key] += 1  # store period
 
-    assert counts_kv_src[duplicate_torsion_idxs] == 2
+    assert counts_kv_src[duplicate_proper_idxs] == 2
 
     counts_kv_dst = dict()
-    for idxs, p in zip(st.dst_system.torsion.potential.idxs, st.dst_system.torsion.params):
+    for idxs, p in zip(st.dst_system.proper.potential.idxs, st.dst_system.proper.params):
         key = tuple(idxs)
         if p[2] == 2.0:
             if key not in counts_kv_dst:
                 counts_kv_dst[key] = 0
             counts_kv_dst[key] += 1  # store period
 
-    assert duplicate_torsion_idxs not in counts_kv_dst
+    assert duplicate_proper_idxs not in counts_kv_dst
 
     st.setup_intermediate_state(0.5)
 
--- a/tests/test_interpolate.py
+++ b/tests/test_interpolate.py
@@ -384,6 +384,8 @@ $$$$""",
             if key not in counts_kv_src:
                 counts_kv_src[key] = 0
             counts_kv_src[key] += 1  # store period
+        else:
+            assert 0
 
     assert counts_kv_src[duplicate_proper_idxs] == 2
 
@@ -394,6 +396,8 @@ $$$$""",
             if key not in counts_kv_dst:
                 counts_kv_dst[key] = 0
             counts_kv_dst[key] += 1  # store period
+        else:
+            assert 0
 
     assert duplicate_proper_idxs not in counts_kv_dst
 
--- a/tests/test_single_topology.py
+++ b/tests/test_single_topology.py
@@ -1649,7 +1649,11 @@ def assert_symmetric_interpolation(mol_a, mol_b, core):
     conf_b = get_romol_conf(mol_b)
     test_conf = st_fwd.combine_confs(conf_a, conf_b, 0)
 
-    for lamb in np.linspace(0, 1, 12):
+    seed = 2024
+    np.random.seed(seed)
+    lambda_schedule = np.concatenate([np.linspace(0, 1, 12), np.random.rand(10)])
+
+    for lamb in lambda_schedule:
         sys_fwd = st_fwd.setup_intermediate_state(lamb)
         sys_rev = st_rev.setup_intermediate_state(1 - lamb)
 
@@ -1667,7 +1671,7 @@ def assert_symmetric_interpolation(mol_a, mol_b, core):
             sys_rev.proper,
             test_conf,
             fused_map,
-            canon_fn=lambda idxs, params: tuple([*canonicalize_bond(idxs), f"{params[1]:.5f}", int(params[2])]),
+            canon_fn=lambda idxs, params: tuple([*canonicalize_bond(idxs), f"{params[1]:.5f}", int(round(params[2]))]),
         )
 
         _assert_u_and_grad_consistent(
--- a/tests/test_single_topology.py
+++ b/tests/test_single_topology.py
@@ -1577,16 +1577,17 @@ def get_vacuum_system_and_conf(mol_a, mol_b, core, lamb):
 
 
 def _assert_consistent_hamiltonian_term_impl(fwd_bonded_term, rev_bonded_term, rev_kv, canon_fn):
-    fwd_canonical_map = dict()
+    canonical_map = dict()
     for fwd_idxs, fwd_params in zip(fwd_bonded_term.potential.idxs, fwd_bonded_term.params):
         fwd_key = canon_fn(fwd_idxs, fwd_params)
-        fwd_canonical_map[fwd_key] = fwd_params
+        canonical_map[fwd_key] = [fwd_params, None]
 
-    # tbd: if fwd has more terms than rev then this logic will miss the inconsistency, this checks
-    # only for subset equivalence
     for rev_idxs, rev_params in zip(rev_bonded_term.potential.idxs, rev_bonded_term.params):
         rev_key = canon_fn([rev_kv[x] for x in rev_idxs], rev_params)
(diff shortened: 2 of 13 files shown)


timemachine/fe/single_topology.py
```diff
@@ -1604,13 +1594,56 @@ def _align_and_interpolate_bonded_term(self, lamb, src_potential, dst_potential,
             dst_potential.potential.idxs,
             dst_potential.params,
         )
+
+        # (ytz): sigh, this is the garbage code that we need to write if we don't do re-shaping
+        # at the potential level, sigh. If we have zero length arrays, for both src_potential and dst_potential,
+        # then we lose information about the correct shapes.
```
mcwitt, line 1600: nit: I'm a bit confused by this comment: I don't think we necessarily lose information about the correct shapes for zero-length arrays (e.g. an empty array of torsion indices should have shape `(0, 4)`)? Is the issue that there are still places in the code where we construct parameter arrays from lists without specifying the desired shape? Would this be addressed by adding an assertion to `BoundPotential.__post_init__`?

timemachine/fe/system.py
```diff
@@ -86,30 +86,20 @@ def simulate_system(U_fn, x0, num_samples=20000, steps_per_batch=500, num_worker
...
+    assert isinstance(bps[4].potential, potentials.Nonbonded)  # cursed: NB AllPairs in VacuumSystem
+
+    chiral_atom = ChiralAtomRestraint(np.array([[]], dtype=np.int32).reshape(-1, 4)).bind(
+        np.array([], dtype=np.float64).reshape(-1)
+    )
+    idxs = np.array([[]], dtype=np.int32).reshape(-1, 4)
+    signs = np.array([[]], dtype=np.int32).reshape(-1)
+    chiral_bond = ChiralBondRestraint(idxs, signs).bind(np.array([], dtype=np.float64).reshape(-1))
+
+    return VacuumSystem(bps[0], bps[1], bps[2], bps[3], bps[4], chiral_atom, chiral_bond)
```
mcwitt, line 102: It seems like this was previously agnostic to the ordering of input potentials, but now requires a specific ordering?

timemachine/fe/system.py
```diff
@@ -130,16 +120,16 @@ class VacuumSystem(Generic[_Nonbonded, _HarmonicAngle]):
     # utility system container
     bond: BoundPotential[HarmonicBond]
     angle: BoundPotential[_HarmonicAngle]
-    torsion: Optional[BoundPotential[PeriodicTorsion]]
+    proper: BoundPotential[PeriodicTorsion]
+    improper: BoundPotential[PeriodicTorsion]
     nonbonded: BoundPotential[_Nonbonded]
-    chiral_atom: Optional[BoundPotential[ChiralAtomRestraint]]
-    chiral_bond: Optional[BoundPotential[ChiralBondRestraint]]
```
mcwitt, line 136: Nice, making these required will let us remove a lot of assertions in the code. Not recalling the original motivation for making them optional.

mcwitt (approved):

timemachine/fe/single_topology.py
proteneer: Sorry this was probably an overreaction on my part. To be more precise: we do have *shape* information defined on :

`src_potential.potential.idxs, src_potential.params, dst_potential.potential.idxs, dst_potential.params`

However, if they're of length zero, the following loop:

``` python
bonded_idxs = []
bonded_params = []
for idxs, src_params, dst_params in set_of_tuples:
    bonded_idxs.append(idxs)
    bonded_params.append(interpolate_fn(idxs, src_params, dst_params, lamb))
```

Will result in length 0 lists for `bonded_idxs`, and `bonded_params`

Then I need to re-determine the `ndim` of the original inputs (or write a conditionals based on type of input), and if it's 1 or 2, check `shape[-1]`, re-shape the `bond_idxs` and `bond_params`:

``` python
if src_params_ndim == 1:
    # chiral atom
    bonded_params = jnp.array(bonded_params, dtype=src_params_dtype)
    pass
elif src_params_ndim == 2:
    # everything else
    src_params_last_shape = src_potential.params.shape[1]
    bonded_params = jnp.array(bonded_params, dtype=src_params_dtype).reshape(-1, src_params_last_shape)
else:
    assert 0
```

Instead, if each `Potential` class automatically knows how to re-shape on its own, then I can just pass in the `bond_idxs=[]` etc. and rely on the class's `__post_init__()` to do the re-shaping.

proteneer: I think the worst is that this pattern is repeated in *many* parts of the code, any time there's a risk of a zero length array/list this type of pattern pops-up.

timemachine/fe/system.py
proteneer: Yes but we've only ever used this for things out of the deserializer (host_bps). I tried to work around it using the original code, but the problem is that now we can possibly emit TWO `PeriodicTorsions` for proper and improper, and there's no sane way to determine which is which. The other PR explicitly named them into "proper" and "improper" via a `NamedSystem/VacuumSystem` but that requires a *much* larger refactor.

timemachine/fe/topology.py
```diff
@@ -476,13 +458,13 @@ def setup_end_state(self):
...
-
-        torsion_params = np.concatenate([mol_proper_params, mol_improper_params])
-        torsion_idxs = np.concatenate([mol_pt.idxs, mol_it.idxs])
-        torsion_potential = potentials.PeriodicTorsion(torsion_idxs).bind(torsion_params)
+        proper_potential = mol_pt.bind(mol_proper_params)
+        improper_potential = mol_it.bind(mol_improper_params)
         nonbonded_potential = mol_nbpl.bind(mol_nbpl_params)
 
-        system = VacuumSystem(bond_potential, angle_potential, torsion_potential, nonbonded_potential, None, None)
+        system = VacuumSystem(
```
proteneer, line 465: @mcwitt - any idea why `mypy` permits this to be to `None`, `None` (since we removed the `Optional` qualifier)

tests/test_single_topology.py
```diff
@@ -450,7 +455,8 @@ def get_max_distance(x0):
             # assert that the idxs are canonicalized.
             assert bond_idxs_are_canonical(system.bond.potential.idxs)
             assert bond_idxs_are_canonical(system.angle.potential.idxs)
-            assert bond_idxs_are_canonical(system.torsion.potential.idxs)
+            assert bond_idxs_are_canonical(system.proper.potential.idxs)
+            assert_improper_idxs_are_canonical(system.improper.potential.idxs)
```
maxentile, line 459: nit[reduced mental overhead + name-mangling]: could have each of these lines be `assert <potential_type>_idxs_are_canonical(system.<potential_type>.potential.idxs)` e.g. by defining aliases `nonbonded_idxs_are_canonical proper_idxs_are_canonical = angle_idxs_are_canonical = chiral_bond_idxs_are_canonical = bond_idxs_are_canonical ...` and tweaking `assert_improper_idxs_are_canonical` to `improper_idxs_are_canonical` . Then, a reader might safely skim over these repetitive lines, and just check whether the definitions are reasonable. 

Tangent: The general logic of the new function `assert_improper_idxs_are_canonical` (assert `idxs == canonicalize(idxs)` seems clearer than the general logic of `bond_idxs_are_canonical` (check first_idx < last_idx).

timemachine/fe/topology.py
mcwitt: I think this is because we've configured mypy to ignore skip checking untyped functions/methods: https://github.com/proteneer/timemachine/blob/1e1ce2f6b663e1129b4e67471f005cea060e3573/pyproject.toml#L20

If you annotate `setup_end_state` as returning `VacuumSystem`, I think mypy will complain.

tests/test_single_topology.py
```diff
@@ -1691,42 +1614,85 @@ def _assert_u_and_grad_consistent(u_fwd, u_rev, x_fwd, x_rev, fused_map):
...
+    # map atoms in the combined mol_ab to the atoms in the combined mol_ba
+    fused_map = _get_fused_map(mol_a, mol_b, core)
+
+    st_fwd = SingleTopology(mol_a, mol_b, core, ff)
+    st_rev = SingleTopology(mol_b, mol_a, core[:, ::-1], ff)
+    conf_a = get_romol_conf(mol_a)
+    conf_b = get_romol_conf(mol_b)
+    test_conf = st_fwd.combine_confs(conf_a, conf_b, 0)
+
+    for lamb in np.linspace(0, 1, 12):
```
maxentile, line 1652: [nit] possibility to strengthen: add special values + randomize? (e.g. `set_prng_seed; lams = [0,0.3,0.5,0.7,1] + list(np.random.rand(0,1,10))` or similar)

(Not sure if there are any informative special-case intermediate lams to included, guessed the stage boundaries from `set(DUMMY_B_CHIRAL_ATOM_CONVERTING_ON_MIN_MAX + DUMMY_B_CHIRAL_ANGLE_CONVERTING_ON_MIN_MAX + ...)` may be useful https://github.com/proteneer/timemachine/blob/ebdb3e4212ae97df4b418c470c726bd3f4f2e6b7/timemachine/fe/single_topology.py#L67-L96 )

tests/test_single_topology.py
```diff
@@ -1691,42 +1614,85 @@ def _assert_u_and_grad_consistent(u_fwd, u_rev, x_fwd, x_rev, fused_map):
...
+            sys_fwd.angle, sys_rev.angle, test_conf, fused_map, canon_fn=lambda idxs, _: tuple(canonicalize_bond(idxs))
+        )
+
+        # for propers, we format the phase as a 5 decimal string to guard against loss of precision
+        _assert_u_and_grad_consistent(
+            sys_fwd.proper,
+            sys_rev.proper,
+            test_conf,
+            fused_map,
+            canon_fn=lambda idxs, params: tuple([*canonicalize_bond(idxs), f"{params[1]:.5f}", int(params[2])]),
```
maxentile, line 1670: surprised to see this non-numeric , might use `np.round` or similar

tests/test_single_topology.py
```diff
@@ -1737,18 +1703,22 @@ def test_hif2a_end_state_symmetry_unit_test():
...
-    sys_rev, conf_rev = get_vacuum_system_and_conf(mol_b, mol_a, core[:, ::-1], 1.0)
 
-    _assert_u_and_grad_consistent(sys_fwd.bond, sys_rev.bond, conf_fwd, conf_rev, fused_map)
+@pytest.mark.nocuda
+@pytest.mark.nightly(reason="slow")
+def test_hif2a_end_state_symmetry_nightly_test():
+    """
+    Test that end-states are symmetric for a large number of random pairs
+    """
+    seed = 2029
```
maxentile, line 1715: future-oriented

timemachine/fe/plots.py
```diff
@@ -590,43 +590,43 @@ def _plot_chiral_atom_interpolation(xs, systems, filter_fn, axs, row):
...
+        proper_ks.append(proper_params[keep_idxs, 0])
+        proper_bs.append(proper_params[keep_idxs, 1])
 
-    torsion_idxs = torsion_idxs[keep_idxs]
-    torsion_ks = np.array(torsion_ks).T  # torsions x K
-    torsion_bs = np.array(torsion_bs).T * (180 / np.pi)  # torsions x K *
-    num_torsions = torsion_ks.shape[0]
+    proper_idxs = proper_idxs[keep_idxs]
+    proper_ks = np.array(proper_ks).T  # propers x K
+    proper_bs = np.array(proper_bs).T * (180 / np.pi)  # propers x K *
```
maxentile, line 609: nit[line comment] ` * ` -> `in degrees` or similar

(and again below)

timemachine/fe/topology.py
proteneer: Ah, thanks. Not sure how I feel here.. since this function sets up a non-chiral state to begin with, so `None` seems less technically correct but simpler.

proteneer: In the spirit of correctness, and since we're now "bare" in terms of downstream assertions, I'll be verbose.

proteneer: Added return types and removed the bare `None, None`.

maxentile (approved):
Good changes throughout -- makes testing stronger, and sets the stage for additional improvements.

Also, I had been wondering what happens if proper torsion params in mol_a get aligned onto improper torsion params in mol_b , but never looked into it -- it'll be interesting to see if enforcing this separation leads to ST efficiency changes (improvements?), too.

I didn't detect that anything functional was missing or incorrect -- my line comments are non-blocking (stylistic or superficial).

timemachine/fe/single_topology.py
mcwitt: > Instead, if each Potential class automatically knows how to re-shape on its own, then I can just pass in the HarmonicBond(idxs=np.arrray([])) # wrong shape and dtype etc. and rely on the class's __post_init__() to do the re-shaping.

I'm probably misunderstanding something, but I don't see why the conditional branches on `ndim` would be necessary in practice? Wouldn't you typically do something like

``` python
idxs = np.array(possibly_empty_list_of_pairs, dtype=np.int32).reshape(-1, 2)
HarmonicBond(idxs)
```
?

I'm not completely against handling the conversion in `__post_init__`, but it does seem like there's a potential for surprising behavior if the input array is not empty. Maybe a compromise where we convert empty input arrays to the required shape and dtype, but assert that non-empty arrays already have the correct shape and dtype, would be palatable to everyone?

proteneer: > I'm probably misunderstanding something, but I don't see why the conditional branches on ndim would be necessary in practice? Wouldn't you typically do something like

`idxs` are fine, as they always have `ndims=2`, but `params` are variadiac.

eg: HarmonicBonds:
``` python
params_for_bonds = np.array(possibly_empty_list_of_pairs, dtype=np.float64).reshape(-1, 2)
HarmonicBond(...).bind(params_for_bonds)
```

but for ChiralAtomRestraints and ChiralBdonRestraints, `ndims=1`:
``` python
params_for_chiral_atoms = np.array(possibly_empty_list_of_pairs, dtype=np.float64).reshape(-1)
ChiralAtomRestraints(...).bind(params_for_chiral_atoms )
```

tests/test_interpolate.py
```diff
@@ -339,27 +375,27 @@ def test_duplicate_idxs_period_pairs():
...
     st = SingleTopology(mol_a, mol_b, core, ff)
 
-    duplicate_torsion_idxs = (0, 1, 3, 4)
+    duplicate_proper_idxs = (0, 1, 3, 4)
 
     counts_kv_src = dict()
-    for idxs, p in zip(st.src_system.torsion.potential.idxs, st.src_system.torsion.params):
+    for idxs, p in zip(st.src_system.proper.potential.idxs, st.src_system.proper.params):
         key = tuple(idxs)
         if p[2] == 2.0:
```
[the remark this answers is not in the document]
proteneer: Guessing you mean this comment to apply to improper torsions, since proper torsions already can have different periods:

```
  'ProperTorsion': { 'patterns': [ ['[*:1]-[#6X4:2]-[#6X4:3]-[*:4]', [[0.4834078610240992, 0.0, 3.0]]],
                                   ['[#6X4:1]-[#6X4:2]-[#6X4:3]-[#6X4:4]', [[0.3859268900186583, 0.0, 3.0], [1.3221171897434616, 3.141592653589793, 2.0], [1.358878183988969, 3.141592653589793, 1.0]]],
                                   ['[#1:1]-[#6X4:2]-[#6X4:3]-[#1:4]', [[0.7999501384731329, 0.0, 3.0]]],
```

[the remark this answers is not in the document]
proteneer: (oh sorry, I think you meant for this *molecule* in particular)... will fix.

tests/test_single_topology.py
proteneer: I need to do this more systematically than just this file re: canonicalizing bonds - there's at least 3 different repetitive definitions of `canonicalize_bond` in `dummy.py, single_topology.py, and ff/handlers/utils.py` - okay if I defer this to a separate deboggling task? Really trying to minimize clerical changes for this PR to avoid a blow-up of diffs like #1420

tests/test_single_topology.py
proteneer: good call - I think `int()` is effectively a `floor()` - will fix.

tests/test_single_topology.py
proteneer: Good idea.

timemachine/fe/single_topology.py
proteneer: (let's defer this to a future design session) - I'm okay with things as-is for now.

tests/test_single_topology.py
proteneer: had to use int(round(params[2])) to go from int(round(jax.ArrayImpl) -> jax.ArrayImpl) -> int

tests/test_interpolate.py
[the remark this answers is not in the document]
proteneer: hm I tried the above but I'm seeing p[2]'s with values  of 4.0 as well. Not sure what's going on here, will revisit later.

tests/test_single_topology.py
proteneer: addressed in https://github.com/proteneer/timemachine/pull/1422/commits/5806877e21289caa91bff126db54c501f395af3e

tests/test_interpolate.py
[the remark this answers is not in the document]
maxentile: Ahh, sorry, mistaken comment

>Guessing you mean this comment to apply to improper torsions, since proper torsions already can have different periods:

Yes, I had misread this as applying to improper torsions (where [all periodicities are currently 2.0](https://github.com/proteneer/timemachine/blob/205b1f04ef7ce77614dcb5a00154fc6da80898cd/timemachine/ff/params/smirnoff_2_0_0_ccc.py#L411-L417)), but these are proper torsions.

>(oh sorry, I think you meant for this molecule in particular)

Sorry for ambiguity -- wasn't thinking of a particular test molecule

> Not sure what's going on here, will revisit later.

Non-issue -- mistaken comment

tests/test_single_topology.py
maxentile: Ahh, sorry meant about the non-numeric type (suggesting `f"{params[1]:.5f}"` -> `np.round(params[1], 5)` or similar). This doesn't matter functionally (only used for equivalence-testing), converting this parameter to a non-numeric type is fine.

PR #1423, opened 2024-11-13
Stagger interpolation for core torsions that are turning on or off.

proteneer:
This PR addresses issues where a core torsion is turning from off to on with a dependency on a non-linear angle terms. For example, morphing a linear H-C#C-H into a H-O-O-H requires us to first partially convert the angles *before* we turn on the associated torsion. This PR also adds a system constraint to ensure that torsions (both propers and impropers) are never defined over linear angles. 

Note that dummy torsions that are being turned on/off are already interpolated at the 0.7 to 1.0 range.

### DIFF
--- a/timemachine/fe/single_topology.py
+++ b/timemachine/fe/single_topology.py
@@ -1164,16 +1164,24 @@ def assert_torsions_defined_over_non_linear_angles(vacuum_system):
     for (i, j, k, l), (proper_k, _, _) in zip(vacuum_system.proper.potential.idxs, vacuum_system.proper.params):
         if proper_k > 0:
             if canonicalize_bond((i, j, k)) in linear_angles:
-                raise TorsionsDefinedOverLinearAngleException(f"angle {(i,j,k)} is linear in torsion {(i,j,k,l)}")
+                raise TorsionsDefinedOverLinearAngleException(
+                    f"angle {(i,j,k)} is linear in proper torsion {(i,j,k,l)}"
+                )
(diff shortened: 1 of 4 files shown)


timemachine/fe/single_topology.py
```diff
@@ -1135,6 +1148,34 @@ def assert_bonds_defined_for_chiral_volumes(vacuum_system, bond_k_min=DEFAULT_BO
...
+        angle_k, angle_a0 = angle_params[0], angle_params[1]
+
+        if angle_k > 0:
+            if abs(angle_a0 - np.pi) < 0.174533:  # 10 degrees, arbitrary but conservative threshold
+                linear_angles.add(tuple(idxs))
+
+    for (i, j, k, l), (proper_k, _, _) in zip(vacuum_system.proper.potential.idxs, vacuum_system.proper.params):
+        if proper_k > 0:
+            if canonicalize_bond((i, j, k)) in linear_angles:
+                raise TorsionsDefinedOverLinearAngleException(f"angle {(i,j,k)} is linear in torsion {(i,j,k,l)}")
```
badisa, line 1167: nit: Would be nice to specify whether the torsion is proper/improper in the warnings to make debugging in the future.

badisa (approved):
Nice, super simple fix

timemachine/fe/single_topology.py
proteneer: good call - will fix.

mcwitt (approved):

PR #1424, opened 2024-11-14
Add option to enable REST in single-topology transformations with HREX

mcwitt:
Adds an option to enable REST-like enhancement of intermediate states in single topology transformations via scaling of certain interactions:

* ligand proper torsions (i, j, k, l) where (j, k) is a rotatable bond
* ligand proper torsions (i, j, k, l) where (j, k) is an aliphatic ring bond
* ligand-ligand coulomb and LJ
* ligand-environment coulomb and LJ

The effective temperature of these interactions in intermediate states is controlled by a configurable symmetric interpolation function (linear, quadratic, and exponential functional forms are predefined):


timemachine/fe/rest/single_topology.py
```diff
@@ -0,0 +1,143 @@
...
+        omm_topology: app.topology.Topology,
+    ) -> HostGuestSystem:
+        ref_state = super().combine_with_host(host_system, lamb, num_water_atoms, ff, omm_topology)
+        num_atoms_host = host_system.nonbonded.potential.num_atoms
+        energy_scale = self.get_energy_scale_factor(lamb)
+
+        nonbonded_host_guest_ixn = replace(
+            ref_state.nonbonded_host_guest_ixn,
+            params=jnp.asarray(ref_state.nonbonded_host_guest_ixn.params)
+            .at[num_atoms_host:, 0]
```
jkausrelay, line 137: Use https://github.com/proteneer/timemachine/blob/master/timemachine/constants.py#L56 instead of numbers here

timemachine/fe/rest/single_topology.py
```diff
@@ -0,0 +1,143 @@
...
+        assert ref_state.proper
+        proper = replace(
+            ref_state.proper,
+            params=jnp.asarray(ref_state.proper.params).at[self.target_proper_idxs, 0].mul(energy_scale),
+        )
+
+        nonbonded = replace(
+            ref_state.nonbonded,
+            params=jnp.asarray(ref_state.nonbonded.params)
+            .at[:, 0]
```
jkausrelay, line 114: Use https://github.com/proteneer/timemachine/blob/master/timemachine/constants.py#L56 instead of numbers here

timemachine/fe/rest/single_topology.py
```diff
@@ -0,0 +1,143 @@
...
+    def __init__(
+        self,
+        mol_a,
+        mol_b,
+        core,
+        forcefield,
+        max_temperature_scale: float,
+        temperature_scale_interpolation_fxn: InterpolationFxnName = "exponential",
+    ):
+        super().__init__(mol_a, mol_b, core, forcefield)
```
jkausrelay, line 47: Add docs for max_temperature_scale, temperature_scale_interpolation_fxn

timemachine/fe/rest/queries.py
```diff
@@ -0,0 +1,29 @@
...
+    ----------
+    mol: ROMol
+        Input molecule
+
+    Returns
+    -------
+    set of CanonicalBond
+        Set of bonds identified as rotatable
+    """
+    pattern = Chem.MolFromSmarts("[!$(*#*)&!D1]-&!@[!$(*#*)&!D1]")
```
jkausrelay, line 23: OpenEye can find rotatable bonds (https://docs.eyesopen.com/toolkits/python/oechemtk/OEChemClasses/OEBondBase.html#OEChem::OEBondBase::IsRotor), may be worth investigating later

timemachine/fe/rest/bond.py
```diff
@@ -0,0 +1,80 @@
...
+    @classmethod
+    def from_idxs(cls, i: int, j: int, k: int):
+        return cls(i, j, k, _unsafe=True) if i < k else cls(k, j, i, _unsafe=True)
+
+    def translate(self, a_to_b: Mapping[int, int] | Sequence[int] | NDArray[np.int32]) -> "CanonicalAngle":
+        return CanonicalAngle.from_idxs(a_to_b[self.i], a_to_b[self.j], a_to_b[self.k])
+
+
+@dataclass(frozen=True)
+class CanonicalTorsion:
```
jkausrelay, line 50: Add a comment this is for Proper torsions only

timemachine/fe/rest/single_topology.py
mcwitt: Nice, I was unaware of this definition. Changed in https://github.com/proteneer/timemachine/pull/1424/commits/81809eafc89ed864eddb7c27cd4282ecbbc44926

timemachine/fe/rest/single_topology.py
mcwitt: Documented in https://github.com/proteneer/timemachine/pull/1424/commits/54953cdcb6c8a62077ef1a0853190715703eee31

timemachine/fe/rest/bond.py
mcwitt: Renamed to emphasize that this is intended for propers in https://github.com/proteneer/timemachine/pull/1424/commits/3290ffa526e0f007f5731323d8556756089215bb

(for impropers, we might want a similar definition that remembers which index corresponds to the central atom)

timemachine/fe/rest/bond.py
```diff
@@ -0,0 +1,100 @@
+from abc import ABC, abstractmethod
```
mcwitt, line 1: This module might eventually be generally applicable, but is limited to the `rest` submodule for now to avoid too large a diff. The idea is to introduce dataclasses representing canonical ixns, which can then be used as argument types to ensure on the type level that indices are canonicalized where expected. Similar vectorized version could be introduced to handle contexts where we have 2d ndarrays of indices.

jkausrelay (approved):

timemachine/fe/rest/single_topology.py
```diff
@@ -0,0 +1,182 @@
...
+        num_atoms_host = host_system.nonbonded.potential.num_atoms
+        energy_scale = self.get_energy_scale_factor(lamb)
+
+        nonbonded_host_guest_ixn = replace(
+            ref_state.nonbonded_host_guest_ixn,
+            params=jnp.asarray(ref_state.nonbonded_host_guest_ixn.params)
+            .at[num_atoms_host:, NBParamIdx.Q_IDX]
+            .mul(energy_scale)  # scale ligand charges
+            .at[num_atoms_host:, NBParamIdx.LJ_EPS_IDX]
+            .mul(energy_scale),  # scale ligand epsilons
```
maxentile, line 179: [style] instead of inlining the index expression `num_atoms_host:`, consider defining an index array `ligand_idxs` or similar and referencing it here. 

[just to confirm] To scale this interaction group potential by scaling the q and eps parameters, would all of the following three implementations be functionally equivalent? (1) scale the ligand's `q, eps` parameters by `energy_scale`, (2) scale the environment's `q, eps` parameters by `energy_scale`, (3) scale all particles' `q, eps` parameters by `sqrt(energy_scale)`?

maxentile (approved):
Looks great! A few non-blocking (mostly stylistic) comments inline

timemachine/fe/rest/single_topology.py
maxentile: (If so, implementation option (3) seems potentially safer -- wouldn't be sensitive to possible errors in computing ligand idxs vs. host idxs.)

timemachine/fe/rest/interpolation.py
```diff
@@ -0,0 +1,106 @@
...
+class Symmetric(Generic[F]):
+    f: F
+
+    @property
+    def src(self):
+        return self.f.src
+
+    @property
+    def dst(self):
+        return self.f.src
```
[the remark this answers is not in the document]
mcwitt: This looks like a typo but I think it's correct as-is. `Symmetric` generates a symmetric interpolation function from an arbitrary `f` by flipping and concatenating, so `Symmetric(f).src == Symmetric(f).dst == f.src`.

timemachine/fe/rest/queries.py
```diff
@@ -0,0 +1,29 @@
+from rdkit import Chem
+from rdkit.Chem import rdMolDescriptors
+
+from .bond import CanonicalBond, mkbond
+
+
+def get_rotatable_bonds(mol: Chem.rdchem.Mol) -> set[CanonicalBond]:
```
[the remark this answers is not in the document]
mcwitt: Yes, this is a bit ad-hoc; this definition is actually mostly duplicated from https://github.com/proteneer/timemachine/blob/f4721d90f2475c77a0654a47fbc36283a26df1e7/timemachine/md/enhanced.py#L35, mainly because importing the latter definition resulted in a circular import, but also because I imagined that the two definitions might diverge in the future.

I agree that moving this to a common utility module seems like a sensible way to organize in the future, but am unsure of the right place (e.g. seems a bit weird to import definition from the `fe` submodule into `md`, but maybe this is fine). Thoughts?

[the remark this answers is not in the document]
maxentile: [organization] Makes sense -- inclined to keep the new file as-is! There are at least 2 other definitions that will probably go here (`get_aliphatic_ring_bonds`, maybe `get_angles_to_soften`), and these might diverge from related non-REST uses.

[flexibility]

>I'd think I'd lean toward waiting to add the strict pattern (or some variant) until we've demonstrated a case where the current definition is too permissive. Thoughts?

Makes sense -- and I don't have a clear adjustment to suggest now. (If there were some known classes of possibly-rotatable bonds that we're missing, or some classes of shouldn't-soften torsions that we're currently hitting, maybe we would want to add toggles for these. I don't have a clear idea of such classes currently.) If this becomes clearer in the future, we can just change the definition of `identify_rotatable_bonds`.

timemachine/fe/rest/single_topology.py
```diff
@@ -0,0 +1,182 @@
...
+    core: np.array (C, 2)
+        Atom mapping from mol_a to mol_b
+
+    forcefield: ff.Forcefield
+        Forcefield to use for parameterization
+
+    max_temperature_scale: float
+        Maximum temperature scale factor
+
+    temperature_scale_interpolation_fxn: str
```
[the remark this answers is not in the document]
mcwitt: Good call. I removed the `_fxn` suffix from the argument name in https://github.com/proteneer/timemachine/pull/1424/commits/6ee28f96ccab25ef4d3d299acd39b7f72e3ff188.

I opted not to accept a callable at this level to simplify the interface exposed by `SingleTopologyREST`, but could consider this in the future.

timemachine/fe/rest/interpolation.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def __call__(self, x: ArrayLike) -> NDArray:
+        x = np.asarray(x)
+        return np.where(
+            x < 0.5,
+            self.f(2.0 * x),
+            self.f(2.0 * (1.0 - x)),
+        )
```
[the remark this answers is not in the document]
mcwitt: Good call; the definitions in `fe.rest.interpolation` all implicitly assume a domain of [0, 1], but this was not validated. Added a check in https://github.com/proteneer/timemachine/pull/1424/commits/4ab60753bdd46e91456b03ca339c90692fdfb421.

timemachine/fe/rest/single_topology.py
```diff
@@ -0,0 +1,182 @@
...
+            for bond in [mkbond(j, k)]
+            if bond in self.rotatable_bonds or bond in self.aliphatic_ring_bonds
+        ]
+
+    @cached_property
+    def target_propers(self) -> NDArray[np.int32]:
+        return self.propers[self.target_proper_idxs, :]
+
+    def get_energy_scale_factor(self, lamb: float) -> float:
+        temperature_factor = self._temperature_scale_interpolation_fxn(lamb).item()
```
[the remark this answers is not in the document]
mcwitt: The reason for this is that I wanted to keep the interpolation function vectorizable (so it takes `ArrayLike` and returns `NDArray`, useful e.g. for plotting). The `.item()` amounts to an assertion that the result is a scalar, and unpacks it into a float. Unsure if there's a neater way to accomplish this; do you have an idea?

timemachine/fe/rest/interpolation.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def plot_interpolation_fxn(f: InterpolationFxn):
+    x = np.linspace(0.0, 1.0, 100)
```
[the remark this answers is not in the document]
mcwitt: I think it's correct as-is: the domain here is always [0, 1] (`f.src` and `f.dst` are elements of the range)

timemachine/fe/rest/interpolation.py
```diff
@@ -0,0 +1,106 @@
...
+
+class InterpolationFxn(Protocol):
+    @property
+    def src(self) -> ArrayLike:
+        ...
+
+    @property
+    def dst(self) -> ArrayLike:
+        ...
+
```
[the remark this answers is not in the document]
mcwitt: Note that `src` and `dst` determine the range of interpolation; the domain is already hard-coded to [0, 1].

timemachine/fe/rest/single_topology.py
```diff
@@ -0,0 +1,182 @@
...
+            return [
+                mkbond(
+                    mol.GetBondWithIdx(bond_idx).GetBeginAtomIdx(),
+                    mol.GetBondWithIdx(bond_idx).GetEndAtomIdx(),
+                )
+                for ring_bond_idxs in mol.GetRingInfo().BondRings()
+                for is_aromatic in [all(mol.GetBondWithIdx(bond_idx).GetIsAromatic() for bond_idx in ring_bond_idxs)]
+                if not is_aromatic
+                for bond_idx in ring_bond_idxs
+            ]
```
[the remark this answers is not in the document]
mcwitt: Ah, good catch. That's definitely the consistent place for it. Moved in https://github.com/proteneer/timemachine/pull/1424/commits/03f6fed1b0ed366ba825559ea52542c48bf9b708.

timemachine/fe/rest/single_topology.py
mcwitt: > (If so, implementation option (3) seems potentially safer -- wouldn't be sensitive to possible errors in computing ligand idxs vs. host idxs.)

Ah, that's a good point. I had gone with (1) just to minimize the number of parameters changed, but I think your argument for (3) is stronger. Will fix.

mcwitt: Changed to (3) in https://github.com/proteneer/timemachine/pull/1424/commits/4e07f8f261bc83784bfda3390b28c17f4623c79c.

timemachine/fe/rest/interpolation.py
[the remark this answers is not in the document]
mcwitt: Unfortunately the assertions broke compositionality in `Symmetric`, because `np.where` evaluates both branches of the piecewise function. Thinking about this more, it seems like the fundamental issue was that previously it was unclear that the definitions in `fe.rest.interpolate` were specialized to a domain of [0, 1]. I've tried to clarify this with module and class docstrings in https://github.com/proteneer/timemachine/pull/1424/commits/8b9332bbbe6014d38a472b719921ccea7aabf6ec.

timemachine/fe/rest/queries.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    Identify rotatable bonds in a molecule.
+
+    NOTE: This is an extremely crude and inaccurate method. This misses simple cases like benzoic acids, amides, etc.
```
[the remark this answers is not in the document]
mcwitt: Thanks for spot-checking. I'm honestly still a little unsure of the inherent subtleties with this definition, but it does appear that the caveat about benzoic acids was incorrect. For now I updated the docstring to reference the pattern in the rdkit code (https://github.com/proteneer/timemachine/pull/1424/commits/25a078dac3feff1188f8ca5333c00e6324fac796); maybe we can update the docstring with caveats as we discover them?

PR #1425, opened 2024-11-14
Move HostConfig into submodule to prevent circular imports

mcwitt:
Cherry-picked from #1424 to simplify review of the latter.

Currently `timemachine.fe.free_energy.HostConfig` is imported by many submodules, e.g. https://github.com/proteneer/timemachine/blob/374a2de64596135e82b1cea30004f2749aa6c2d2/timemachine/md/enhanced.py#L21

This means that importing (even transitively) these submodules from `free_energy` will result in a circular import. Since `free_energy` is currently a higher-level submodule containing application logic, this is difficult to avoid (e.g., I ran into this in #1424 with the following dependency cycle: `fe.free_energy > fe.rest > md.enhanced > fe.free_energy`). Moving `HostConfig` into a separate submodule is a relatively minimal change to work around this issue.

This PR:

* moves `timemachine.fe.free_energy.HostConfig` to `timemachine.fe.host_config.HostConfig`
* converts  `HostConfig` to a (frozen) dataclass

### DIFF
--- a/examples/biphenyl_torsion_sampling_hrex.py
+++ b/examples/biphenyl_torsion_sampling_hrex.py
@@ -19,7 +19,8 @@ from numpy.typing import NDArray
 
 from timemachine.constants import DEFAULT_TEMP
 from timemachine.fe import model_utils
-from timemachine.fe.free_energy import HostConfig, HREXParams, InitialState, MDParams, run_sims_bisection, run_sims_hrex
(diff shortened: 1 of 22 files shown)


mcwitt:
Closing for now; this was incomplete (e.g. `md.enhanced` imports `AbsoluteFreeEnergy` from `fe.free_energy`), and I found a simpler workaround in #1424.

PR #1426, opened 2024-11-15
Optimize Lambda Schedule Targeting Uniform Overlap Between Windows 

badisa:


timemachine/fe/rbfe.py
```diff
@@ -743,27 +835,48 @@ def get_mean_final_barostat_volume_scale_factor(trajectories_by_state: Iterable[
...
+            traj = trajectories_by_state[state_idx]
+            if np.isclose(nearest_state.lamb, lamb):
+                state = nearest_state
+            else:
+                # If the lambda value is different, reconstruct the initial state to the correct parameters
+                state = make_initial_state_fn(lamb, False)
+
+                # Verify that the forces of the nearest lambda value's frames are stable, since frames were not generated
+                # with the same parameters
+                du_dx, _ = state.to_bound_impl().execute(traj.frames[-1], traj.boxes[-1], compute_u=False)
```
badisa, line 850: Had previously been concerned about taking the frame from the nearest initial state. In pratice (tested on hif2a) this seems to not be an issue.

Had talked to @maxentile offline about doing a probabilistic approach, but ended up being too slow to be practice.

timemachine/optimize/protocol.py
```diff
@@ -358,34 +358,36 @@ def greedily_optimize_protocol(
...
+    protocol_interval: tuple[float, float] = (0.0, 1.0),
 ) -> Array:
     """Optimize a lambda protocol from "left to right"
 
     Sequentially pick next_lam so that
     distance_fxn(prev_lam, next_lam) ~= target_distance
     """
-    protocol = [0.0]
+    start_lamb, end_lamb = protocol_interval
+    protocol = [start_lamb]
```
maxentile, line 369: Nice -- and now it doesn't even require start_lamb < end_lamb

timemachine/fe/free_energy.py
```diff
@@ -77,16 +77,22 @@ class HREXParams:
         If given, number of neighbor states on either side of a given replica's initial state for which to compute
         potentials. This determines the maximum number of states that a replica can move from its initial state during
         a single HREX iteration. Otherwise, compute potentials for all (replica, state) pairs.
+
+    optimize_target_overlap: float or None
+        If given, optimize the lambda schedule out of the initial bisection phase to target a specific minimum overlap
+        between all adjacent windows. Must be in the interval (0.0, 1.0] if provided.
```
maxentile, line 83: Huh, come to think of it, `optimize_target_overlap = 1.0` would probably lead to an infinite loop?

... Ahh, no, since `greedily_optimize_protocol` defaults to `max_iterations=1000`.

Maybe
```suggestion
        between all adjacent windows. Must be in the interval (0.0, 1.0) if provided.
```
or
```suggestion
        between all adjacent windows. Must be in the interval [0.0, 1.0) if provided.
```

timemachine/fe/rbfe.py
```diff
@@ -274,9 +277,57 @@ def setup_initial_states(
...
+        overlap_dist, target_dist, bisection_xtol=xtol, protocol_interval=(lambda_min, lambda_max)
+    )
+    new_schedule = np.asarray(greedy_prot)
+    if len(greedy_prot) > len(initial_lambs):
+        warnings.warn("Optimized schedule has more windows than initial schedule, falling back to initial schedule")
+        new_schedule = initial_lambs
+    else:
+        print(
+            f"Optimized schedule has {len(new_schedule)} windows compared to {len(initial_lambs)} windows initially, target overlap {target_overlap}"
+        )
```
maxentile, line 311: Maybe just warn, rather than override, in case optimized schedule has more windows than initial schedule? There may be cases where this is intentional. (e.g. if `target_overlap > bisection_min_overlap`)

Something like:
```suggestion
    if len(greedy_prot) > len(initial_lambs):
        warnings.warn("Optimized schedule has more windows than initial schedule, falling back to initial schedule")    
    
    print(
        f"Optimized schedule has {len(new_schedule)} windows compared to {len(initial_lambs)} windows initially, target overlap {target_overlap}"
    )
```
(may have messed up whitespace in GH comment editor)

maxentile:
PR naming note: This doesn't quite target "uniform overlap" between all adjacent pairs of windows `[(lambdas[i], lambdas[i+1]) for i in range(K-1)]`. It targets a fixed value of overlap for the first `K-2` pairs `[(lambdas[i], lambdas[i+1]) for i in range(K-2)]`. The very last pair `(lambdas[-2], lambdas[-1])` might have a much larger overlap.

(In these cases, uniform overlap might have been achieved if we had known to target a slightly higher `overlap(i, i+1) for all i`. One possible brute-force approach (https://github.com/proteneer/timemachine/pull/1385) would be to call the current method several times with different trial values of `target_overlap`, trying to make `target_overlap` as large as possible without increasing `len(lambdas)`.)

image
(x-axis: lambda `idx`, y-axis: `overlap(lambdas[idx], lambdas[idx+1])`, dashed grey line: target overlap, blue curve: current approach, orange curve: brute-force (but maybe more uniform) approach)

timemachine/fe/rbfe.py
badisa: > There may be cases where this is intentional. (e.g. if target_overlap > bisection_min_overlap)

I am not necessarily opposed to just warning and returning the new protocol (though do we want the number of windows to be greater than 48?), but the code prevents the scenario you describe in: https://github.com/proteneer/timemachine/blob/6c5358077b0c3cadff718892eb5868b542e6dbd0/timemachine/fe/rbfe.py#L800-L802 

I enforce that the bisection `overlap >= target_overlap` based on previous discussions, though would be opening to removing this restriction.

timemachine/fe/rbfe.py
```diff
@@ -274,9 +277,57 @@ def setup_initial_states(
...
+
+    return [setup_initial_state_fn(lamb) for lamb in new_schedule]
+
+
+def get_nearest_state_idx(lamb: float, initial_states: Sequence[InitialState]) -> int:
+    """
+    Return the index of the initial state with the closest lambda value. When determining the nearest initial state, will only consider states on the same side of
```
mcwitt, line 318: nit: formatting: wrap docstring to consistent number of columns

timemachine/fe/rbfe.py
```diff
@@ -298,27 +349,24 @@ def setup_optimized_initial_state(
     InitialState
         Optimized at the specified lambda value
     """
-
-    # NOTE: The current approach for generating optimized conformations in `optimize_coordinates` creates a
-    # discontinuity at lambda=0.5. Ensure that we pick a pre-optimized state on the same side of 0.5 as `lamb`
-    states_subset = [s for s in optimized_initial_states if (s.lamb <= 0.5) == (lamb <= 0.5)]
-    nearest_optimized = min(states_subset, key=lambda s: abs(lamb - s.lamb))
+    nearest_optimized_idx = get_nearest_state_idx(lamb, optimized_initial_states)
```
mcwitt, line 352: nit: any reason for this not to return the `InitialState` itself rather than its index?

timemachine/fe/rbfe.py
```diff
@@ -702,18 +751,63 @@ def estimate_relative_free_energy_bisection_hrex_impl(
     lambda_max: float,
     md_params: MDParams,
     n_windows: int,
-    make_optimized_initial_state_fn: Callable[[float], InitialState],
+    make_initial_state_fn: Callable[[float, bool], InitialState],
```
mcwitt, line 754: nit/non-blocking: idea for an alternative in case adding an argument to the callback seems unwieldy: could this accept a separate `optimize_initial_state: Callable[[InitialState], InitialState]` argument, which is only called during bisection but not refinement?

mcwitt (approved):
Looks good, I'm excited to see this in action!

timemachine/fe/rbfe.py
badisa: Because I need to retrieve the associated trajectory to update the state with the coords, box and velocities.

timemachine/fe/rbfe.py
badisa: Hmmm that seems unwieldy to me as well especially since one function would be optional depending on the setup of MDParams.

What about having a `make_initial_state_fn` and a `optimize_initial_state_fn`? That way we don't have two functions to produce the same type of object with the nuance of being optimized? Also the `optimize_initial_state_fn` could be the identity function if we wanted.

timemachine/fe/rbfe.py
mcwitt: I might be missing something, but I don't see any other usage of `nearest_optimized_idx` in the diff?

badisa: It is used here https://github.com/proteneer/timemachine/blob/af49751774bb234b8f5a16d9d0fa24bb44aa55a4/timemachine/fe/rbfe.py#L840

timemachine/fe/rbfe.py
mcwitt: > What about having a make_initial_state_fn and a optimize_initial_state_fn? That way we don't have two functions to produce the same type of object with the nuance of being optimized? Also the optimize_initial_state_fn could be the identity function if we wanted.

I think this is what I intended to suggest - i.e. accept two function arguments, one with type `float -> InitialState` and another with type `InitialState -> InitialState` (that could be the identity).

badisa: Woops, that is what you suggested and I just misread it.

PR #1430, opened 2024-11-22
Revert chiral validity removal. Now fall-back if no valid groups are found.

proteneer:
This PR reverts a subtle change that can impact the selection of dummy groups (and their respective dummy anchors).


proteneer:
stale

PR #1434, opened 2024-12-04
Use base docker file and copy curand

badisa:
* Only need curand and cudart, can avoid hundreds of MBs of data by copying curand from the devel image
* Reduces the container size from 5.93GB to 3.26GB.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -113,13 +113,18 @@ COPY . /code/timemachine/
 WORKDIR /code/timemachine/
 RUN pip install --no-cache-dir -e . && rm -rf ./build
 
-# Container with only cuda runtime, half the size of the timemachine_cuda_dev container
-FROM nvidia/cuda:12.4.1-runtime-ubuntu20.04 as timemachine
+# Container with only cuda base, half the size of the timemachine_cuda_dev container
+# Need to copy curand/cudart as these are dependecies of the Timemachine GPU code
+FROM nvidia/cuda:12.4.1-base-ubuntu20.04 as timemachine
 ARG LIBXRENDER_VERSION
 ARG LIBXEXT_VERSION
 RUN (apt-get update || true) && apt-get install --no-install-recommends -y libxrender1=${LIBXRENDER_VERSION} libxext-dev=${LIBXEXT_VERSION} \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
+# Copy all of the include directory, more headers than neccessary but headers are small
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -113,17 +113,24 @@ COPY . /code/timemachine/
...
-FROM nvidia/cuda:12.4.1-runtime-ubuntu20.04 as timemachine
+# Container with only cuda base, half the size of the timemachine_cuda_dev container
+# Need to copy curand/cudart as these are dependecies of the Timemachine GPU code
+FROM docker.io/nvidia/cuda:12.4.1-base-ubuntu20.04 as timemachine
 ARG LIBXRENDER_VERSION
 ARG LIBXEXT_VERSION
 RUN (apt-get update || true) && apt-get install --no-install-recommends -y libxrender1=${LIBXRENDER_VERSION} libxext-dev=${LIBXEXT_VERSION} \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
-COPY --from=timemachine_cuda_dev /opt/ /opt/
```
badisa, line 123: This was previously copying `/opt/nvidia` which is about 1.5Gb for nvidia-compute which we never use.

Dockerfile
```diff
@@ -113,17 +113,24 @@ COPY . /code/timemachine/
...
+COPY --from=timemachine_cuda_dev /usr/local/cuda/targets/x86_64-linux/include/ /usr/local/cuda/targets/x86_64-linux/include/
+COPY --from=timemachine_cuda_dev /usr/local/cuda/targets/x86_64-linux/lib/libcurand* /usr/local/cuda/targets/x86_64-linux/lib/
+COPY --from=timemachine_cuda_dev /usr/local/cuda/lib64/libcurand* /usr/local/cuda/lib64/
+COPY --from=timemachine_cuda_dev /opt/conda/ /opt/conda/
+COPY --from=timemachine_cuda_dev /opt/openmm_install/ /opt/openmm_install/
 COPY --from=timemachine_cuda_dev /code/ /code/
 COPY --from=timemachine_cuda_dev /root/.bashrc /root/.bashrc
 RUN ln -s /opt/conda/etc/profile.d/conda.sh /etc/profile.d/conda.sh
 ARG ENV_NAME=timemachine
+ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/cuda/lib64/
```
badisa, line 134: Need to specify the linker path, else the copied .so files aren't picked up.

PR #1435, opened 2024-12-10
Reduce Final Docker Image Size

badisa:
* Only need curand and cudart, can avoid hundreds of MBs of data by copying curand from the devel image
* Reduced container size from 5.87GB to 3.18GB, by removing /opt/nvidia and only keeping cudart/curand.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -121,8 +121,7 @@ ARG LIBXEXT_VERSION
 RUN (apt-get update || true) && apt-get install --no-install-recommends -y libxrender1=${LIBXRENDER_VERSION} libxext-dev=${LIBXEXT_VERSION} \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
-# Copy all of the include directory, more headers than neccessary but headers are small
-COPY --from=timemachine_cuda_dev /usr/local/cuda/targets/x86_64-linux/include/ /usr/local/cuda/targets/x86_64-linux/include/
+
 # Copy curand libraries from image, only require cudart and curand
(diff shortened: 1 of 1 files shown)


jkausrelay (approved):

Dockerfile
```diff
@@ -113,17 +113,26 @@ COPY . /code/timemachine/
...
+# Need to copy curand/cudart as these are dependecies of the Timemachine GPU code
+FROM docker.io/nvidia/cuda:12.4.1-base-ubuntu20.04 as timemachine
 ARG LIBXRENDER_VERSION
 ARG LIBXEXT_VERSION
 RUN (apt-get update || true) && apt-get install --no-install-recommends -y libxrender1=${LIBXRENDER_VERSION} libxext-dev=${LIBXEXT_VERSION} \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
-COPY --from=timemachine_cuda_dev /opt/ /opt/
+# Copy all of the include directory, more headers than neccessary but headers are small
+COPY --from=timemachine_cuda_dev /usr/local/cuda/targets/x86_64-linux/include/ /usr/local/cuda/targets/x86_64-linux/include/
```
mcwitt, line 125: Dumb question, do we actually need to copy any headers at this point (since I don't think we're going to build anything with this image)?

badisa: Nope, will remove.

mcwitt (approved):

Dockerfile
badisa: Removed in https://github.com/proteneer/timemachine/pull/1435/commits/e691853a45d19b2c589f02002951a821efb6ff16

PR #1442, opened 2024-12-25
Enforce Existing C++ Conventions in Clang Format

badisa:
Modifies the clang-format configuration to enforce existing conventions:
* Enforces alignment of comments for function parameters
* Enforces style for pointers and references (`int *foo` rather than `int* foo`).
* Enforces that namespaces have a comment at the closing brace which specifies the namespace's name.


mcwitt (approved):

proteneer (approved):
This looks great!

PR #1445, opened 2025-01-02
Clean up SingleTopology code

badisa:
Over the holiday, I was read through the SingleTopology code to familiarize myself with it. These are some minor changes I made to clean up the code as I went.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -6,3 +6,8 @@ BinPackArguments: false
 BinPackParameters: false
 AccessModifierOffset: -4
 ReflowComments: false
+AlignTrailingComments:
+  Kind: Always
+  OverEmptyLines: 2
+FixNamespaceComments: true
+PointerAlignment: Right
--- a/timemachine/cpp/src/barostat.hpp
+++ b/timemachine/cpp/src/barostat.hpp
@@ -76,9 +76,9 @@ private:
     double *d_x_after_;
     double *d_box_after_;
 
-    int *d_atom_idxs_;   // grouped index to atom coords
-    int *d_mol_idxs_;    // grouped index to molecule index
-    int *d_mol_offsets_; // Offset of molecules to determine size of mols
+    int *d_atom_idxs_;                // grouped index to atom coords
+    int *d_mol_idxs_;                 // grouped index to molecule index
+    int *d_mol_offsets_;              // Offset of molecules to determine size of mols
 
     unsigned long long *d_centroids_; // Accumulate centroids in fixed point to ensure deterministic behavior
 
--- a/timemachine/cpp/src/bd_exchange_move.cu
+++ b/timemachine/cpp/src/bd_exchange_move.cu
@@ -440,8 +440,8 @@ void BDExchangeMove<RealType>::compute_incremental_log_weights_device(
 template <typename RealType>
 std::vector<std::vector<RealType>> BDExchangeMove<RealType>::compute_incremental_log_weights_host(
     const int N,
-    const double *h_coords, // [N, 3]
-    const double *h_box,    // [3, 3]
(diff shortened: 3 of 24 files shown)


timemachine/fe/rest/single_topology.py
```diff
@@ -42,37 +43,38 @@ class SingleTopologyREST(SingleTopology):
...
+
+        forcefield: ff.Forcefield
+            Forcefield to use for parameterization
+
+        max_temperature_scale: float
+            Maximum temperature scale factor
+
+        temperature_scale_interpolation: str
+            Interpolation function to use for temperature scaling. One of "linear", "quadratic", or "exponential"
+        """
```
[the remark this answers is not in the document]
mcwitt: One wrinkle I see with the PEP style is that it requires an exception for dataclasses that do not have an explicit `__init__` method (the PEP predates dataclasses). Aside from this minor gripe, I'm fine with either style (but would prefer to be consistent).

timemachine/fe/topology.py
```diff
@@ -318,14 +318,14 @@ def parameterize_nonbonded_pairlist(
         inclusion_idxs, rescale_mask = [], []
         for i in range(self.mol.GetNumAtoms()):
             for j in range(i + 1, self.mol.GetNumAtoms()):
-                scale_factor = exclusions_kv.get((i, j), (0.0, 0.0))  # how much to remove
-                rescale_factor = 1 - np.array(scale_factor)  # how much to keep
+                scale_factor = exclusions_kv.get((i, j), np.zeros(2))  # how much to remove
```
proteneer, line 321: Possible to specify a `dtype=np.float64` here? (To avoid guessing if it's `int32` vs `float64` vs `float32`)

proteneer (approved):

timemachine/fe/dummy.py
```diff
@@ -101,12 +89,12 @@ def get_bond_anchors(dummy_group):
...
 
 def generate_anchored_dummy_group_assignments(
-    dummy_groups: Dict[int, FrozenSet[int]],
+    dummy_groups: dict[int, frozenset[int]],
     bond_graph_a: nx.Graph,
     bond_graph_b: nx.Graph,
-    core_atoms_a: Sequence[int],
-    core_atoms_b: Sequence[int],
-) -> Iterator[Dict[int, Tuple[Optional[int], FrozenSet[int]]]]:
+    core_atoms_a,
```
mcwitt, line 95: nit: why remove the annotations here?

timemachine/fe/rest/single_topology.py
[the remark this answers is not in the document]
mcwitt: > For the dataclasses, I would argue that they are 'self-documenting' since it is intended only as a data container.

I think I disagree with this in general, since it still seems useful in many cases to document the data fields (even if there is no "behavior" to document like in a full-fledged class).

(I'm fine with just adding a docstring at the class level when documenting dataclass fields is desired.)

timemachine/fe/single_topology.py
```diff
@@ -1274,23 +1301,23 @@ def combine_masses(self, use_hmr=False):
...
                 mass = max(mass_a, mass_b)
-            elif indicator == 1:
+            elif flag == AtomMapFlags.MOL_A:
                 mass = mol_a_masses[self.c_to_a[c_idx]]
-            elif indicator == 2:
+            elif flag == AtomMapFlags.MOL_B:
                 mass = mol_b_masses[self.c_to_b[c_idx]]
             else:
-                assert 0
+                assert 0, f"Unknown atom flag: {flag}"
```
mcwitt, line 1314: Nice!

mcwitt (approved):

PR #1447, opened 2025-01-03
Move pre-commit to CI requirements

badisa:
* Avoid hard coding a version, which was inconsistent with setup.py, in the dockerfile
* Fixes a typo in the Dockerfile


mcwitt (approved):
Good catch :+1:

PR #1453, opened 2025-01-21
Remove legacy code related to chiral conversion

mcwitt:
Completes the removal of code that was added in #1356 and #1360  to work around our initial lack of support for chiral volume conversions (since added in #1406).

Highlights:
- Completely removes the atom mapping leaf filter `enforce_chirally_valid_dummy_groups`. (The infrastructure supporting arbitrary leaf filters is left in place for now.)
- Simplifies dummy group selection to choose an arbitrary assignment independently for A and B (since we no longer need to jointly select A and B to avoid chiral volume conversions)
- Re-unifies end state setup code (since we no longer have a use case for parameterizing bonds and chiral volume restraints separately)


proteneer (approved):
Oh this one feels good to remove!

PR #1454, opened 2025-01-21
Bump to Python 3.12

badisa:
* Had to bump Scipy/Numpy/RDKit to upgrade to Python 3.12. Avoided going all the way to Numpy 2, as it has implications for our wrapping code.
* Updates Black/flake8. Upgrading Python created a conflict between the two, regarding whitespace and newlines. In the end, had to skip certain checks in flake8 to make them work together.
* Sadly no noticeable speed up. May want to investigate [specialist](https://github.com/brandtbucher/specialist), in case we can take advantage of the new [specialization Interpreter](https://peps.python.org/pep-0659/).
* Specifies Numpy/Scipy in the `environment.yml` to avoid having to uninstall newer versions in a later step. Slightly reduces docker container size (~50mb).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,6 @@ ARG LIBXRENDER_VERSION
 ARG LIBXEXT_VERSION
 
 # Copied out of anaconda's dockerfile
-ARG MINICONDA_VERSION=py310_23.1.0-1
 ARG MAKE_VERSION=4.2.1-1.2
 ARG GIT_VERSION=1:2.25.1-*
 ARG WGET_VERSION=1.20.3-*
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -1,9 +1,9 @@
...
-scipy==1.10.1
+numpy==1.26.4
+scipy==1.15.1
 pymbar==3.1.0
 networkx==2.8.8
 matplotlib==3.7.1
-rdkit==2022.3.5
+rdkit==2024.9.4
 --extra-index-url https://pypi.anaconda.org/OpenEye/simple
 openeye-toolkits==2020.2.0
```
jkausrelay, line 8: Should we update openeye at some point?

.pre-commit-config.yaml
```diff
@@ -13,12 +13,12 @@ repos:
...
 
 -   repo: https://github.com/psf/black
-    rev: 23.9.1
+    rev: 24.10.0
     hooks:
     -   id: black
 
 -   repo: https://github.com/pycqa/flake8
-    rev: 6.1.0
+    rev: 7.1.0
```
mcwitt, line 21: Why not [`7.1.1`](https://pypi.org/project/flake8/7.1.1/)?

mcwitt (approved):

mcwitt:
> Avoided going all the way to Numpy 2, as it has implications for our wrapping code.

Curious, do you foresee a specific issue here? Anecdotally,  I'm not running into any issues building the extension or running tests that exercise it (with `numpy==2.2.2`).

Assuming no major issues, I think upgrading numpy would be nice as several major libraries (e.g. pandas) are now requiring `numpy>=2`.

PR #1458, opened 2025-01-24
Speed up u kln construction for Pairbar

badisa:
* Speeds up Sequential and HREX (by about 2%), not bisection. Bisection requires more energy comparisons, due to the dynamic nature of the schedule.
* Only load each set of frames into memory once
* Evaluate all parameter sets of interest once, reducing the number of neighborlist constructions

# Benchmarks
Run on Nvidia A10g
Cuda Arch 8.6

## Complex
image
image

## Solvent
image
image

## Master
<details><summary>Benchmark Plots</summary>
<p>

vacuum_1000_benchmarks_master
solvent_1000_benchmarks_master
complex_1000_benchmarks_master
</p>
</details> 

## PR
<details><summary>Benchmark Plots</summary>
<p>

vacuum_1000_benchmarks
solvent_1000_benchmarks
complex_1000_benchmarks
</p>
</details>


tests/test_hrex_rbfe.py
```diff
@@ -225,6 +227,30 @@ def time_lagged_rmsd(traj):
...
+        batch_u_fns = get_batch_u_fns(unbound_impls, [p.params for p in state.potentials], temperature)
+        energy_decomp_states.append(EnergyDecomposedState(traj.frames, traj.boxes, batch_u_fns))
+
+    ref_bar_results = []
+    for state_a, state_b in zip(energy_decomp_states, energy_decomp_states[1:]):
+        u_kln_by_component = compute_energy_decomposed_u_kln([state_a, state_b])
+        ref_bar_results.append(estimate_free_energy_bar(u_kln_by_component, temperature))
+    for ref_res, comp_res in zip(ref_bar_results, comp_bar_results):
+        assert ref_res.overlap == comp_res.overlap
+        assert np.all(ref_res.dG_err_by_component == comp_res.dG_err_by_component)
```
mcwitt, line 249: nit: slight preference for `np.testing.assert_array_equal` for more useful failure messages

timemachine/fe/free_energy.py
```diff
@@ -1259,6 +1243,67 @@ def compute_u_kn(trajs, initial_states) -> Tuple[NDArray, NDArray]:
...
+    """
+
+    assert len(initial_states) > 0
+    assert len(initial_states) == len(samples_by_state)
+    if unbound_impls is None:
+        unbound_impls = [pot.potential.to_gpu(np.float32).unbound_impl for pot in initial_states[0].potentials]
+    assert len(unbound_impls) == len(initial_states[0].potentials)
+    kBT = temperature * BOLTZ
+    # Construct an empty array
+    state_to_params = np.zeros((len(initial_states), len(initial_states), len(unbound_impls)), dtype=object)
```
mcwitt, line 1270: I find the naming of this confusing. Is it correct that this is an array of shape (states, states, potentials) whose values are energies?

timemachine/fe/free_energy.py
```diff
@@ -1259,6 +1243,67 @@ def compute_u_kn(trajs, initial_states) -> Tuple[NDArray, NDArray]:
...
+
+        state_idxs = []
+        if i > 0:
+            state_idxs.append(i - 1)
+        state_idxs.append(i)
+        if i < len(initial_states) - 1:
+            state_idxs.append(i + 1)
+        for j, pot in enumerate(state.potentials):
+            params = np.array([initial_states[idx].potentials[j].params for idx in state_idxs])
+            _, _, Us = unbound_impls[j].execute_batch(
```
mcwitt, line 1283: Would it be possible to use `execute_batch_sparse` here instead (similar to the usage in [`compute_potential_matrix`](https://github.com/proteneer/timemachine/blob/caaaad2a43756435b18e7303f2621ccb7220d4db/timemachine/fe/free_energy.py#L1116)) to compute the full `U_kln` matrix with a single custom_ops call per potential?

mcwitt: Never mind - I wasn't quite understanding what this was doing. I don't see an obvious way of doing this.

mcwitt (approved):

timemachine/fe/free_energy.py
```diff
@@ -1459,30 +1504,10 @@ def format_rates(rs):
...
+    neighbor_ulkns = generate_pair_bar_ulkns(initial_states, samples_by_state, temperature, unbound_impls=unbound_impls)
 
-    bar_results = list(
-        pairwise_transform_and_combine(
-            results_by_state,
-            make_energy_decomposed_state,
-            lambda s1, s2: estimate_free_energy_bar(compute_energy_decomposed_u_kln([s1, s2]), temperature),
-        )
-    )
+    pair_bar_results = [estimate_free_energy_bar(u_kln, temperature) for u_kln in neighbor_ulkns]
```
maxentile, line 1509: [local naming self-consistency] There might be an error in one or the other name? "`... for u_kln in neighbor_ulkns ...`"

timemachine/fe/free_energy.py
```diff
@@ -1259,6 +1243,67 @@ def compute_u_kn(trajs, initial_states) -> Tuple[NDArray, NDArray]:
...
+) -> NDArray:
+    """Generate pair bair u_klns.
+    This is a specialized variant of generating u_klns, only loading each set of frames into memory once.
+    Each set of frames is loaded once then all of the parameters of interest are run in a batch.
+    This improves throughput for potentials that use Neighborlists, as there are at most len(frames) neighborlist
+    rebuilds, rather than 3 * len(frames).
+
+    Returns
+    -------
+        u_klns: np.array[len(initial_states) - 1, len(unbound_impls), 2, 2, n_frames]
```
maxentile, line 1260: [performance uncertainty] I wonder if there's a meaningful performance difference between computing `u(x)` vs. `[u_i(x) for u_i in components]`. The estimate depends on the summed value `u(x)`, and some of the diagnostics depend on the decomposition into ~8 components. If there's a meaningful performance difference, and if we decide the energy-decomposed diagnostic is optional, we might adjust this to compute `u(x)` using `SummedPotential` or similar whenever the diagnostic is not needed.

-----

[naming observation] `u_klns` or `ulkns` were maybe a little cryptic to start with (describing arrays where `array[k][l][n] = u_k(trajs[l][n])` https://pymbar.readthedocs.io/en/latest/mbar.html ), but the name is less accurate when used to describe the return type of these functions, where `array[k][p][i][j][n]` is the `p`'th component of the energy of `trajs[k + j][n]` evaluated in state `k + i`. The previous name `u_kln_by_component_by_lambda` was a little more verbose but maybe more accurate.

timemachine/fe/free_energy.py
```diff
@@ -1259,6 +1243,67 @@ def compute_u_kn(trajs, initial_states) -> Tuple[NDArray, NDArray]:
     return u_kn, np.array(N_k)
 
 
+def generate_pair_bar_ulkns(
+    initial_states: Sequence[InitialState],
+    samples_by_state: Sequence[Trajectory],
+    temperature: float,
+    unbound_impls: Sequence[custom_ops.Potential] | None,
```
maxentile, line 1250: surprise `| None` -- what corner case is this for?

maxentile (approved):
Performance, correctness look good-- and it's well-tested.

(Review contains non-blocking observations.)

timemachine/fe/free_energy.py
maxentile: Ahh, makes sense -- thanks!

timemachine/fe/free_energy.py
maxentile: Ahh, I just meant that `*_lkns` sounds like it's a list of elements (each `*_lkn`) not transposed elements (each `*_kln`). It's just naming though -- the logic looked fine.

PR #1462, opened 2025-01-25
Upgrade to Numpy 2

badisa:
* Upgrades Pybind11 to latest
* Upgrades Numpy to 2.2.2


mcwitt:
Added fixes for the forcefield serialization issues in https://github.com/proteneer/timemachine/pull/1462/commits/561e7560e9582f5e1f5d41ba3d0e52fe9cb30d4b and https://github.com/proteneer/timemachine/pull/1462/commits/0295c800959ccebfe24ed7054b0aa98c79ac2db7. This seems to be related to the change in e.g. `np.float64.__repr__`:

`numpy==1.26.4`:
```
repr(np.float64(0.0))
'0.0'
```

`numpy==2.2.2`:
```
>>> repr(np.float64(0.0))
'np.float64(0.0)'
```

Added fixes for new mypy complaints in https://github.com/proteneer/timemachine/pull/1462/commits/bfcb54f21c81243bb442b07d84f7f9b6a7dfcec2. The major theme here seems to be that `ndarray.tolist` is improperly inferred to return `list[str]` (pyright correctly infers this, so maybe this issue will go away if/when we upgrade mypy, but didn't want to open that can of worms in this PR).

tests/nonbonded/test_nonbonded_interaction_group.py
```diff
@@ -61,7 +61,7 @@ def test_nonbonded_interaction_group_zero_interactions(rng: np.random.Generator)
 
     potential = NonbondedInteractionGroup(num_atoms, ligand_idxs, beta, cutoff)
 
-    du_dx, du_dp, u = potential.to_gpu(np.float64).unbound_impl.execute(conf, params, box)
+    du_dx, du_dp, u = potential.to_gpu(np.float64).unbound_impl.execute(conf, params, box.astype(np.float64))
```
proteneer, line 64: what `dtype` does `box = 10.0 * np.eye(3)` produce? (is this a numpy2 change? or a mypy linting thing?)

proteneer: (by default I was under the impression that np.eye(3) is already a `np.float64` in numpy1:

```
>>> import numpy as np
>>> np.eye(3).dtype
dtype('float64')
```

proteneer (approved):

tests/nonbonded/test_nonbonded_interaction_group.py
mcwitt: Yes, this is to deal with a mypy / numpy typing peculiarity. When a float64 array is multiplied by a Python float, mypy infers the more general dtype `numpy.floating[Any]]` for the result, leading to an error later on when we expect `numpy.float64`:

`example.py`:

```python
from typing import reveal_type

import numpy as np

reveal_type(np.eye(3))
reveal_type(10.0 * np.eye(3))
```

```
$ mypy example.py
tmp.py:5: note: Revealed type is "numpy.ndarray[builtins.tuple[builtins.int, ...], numpy.dtype[numpy.float64]]"
tmp.py:6: note: Revealed type is "numpy.ndarray[builtins.tuple[builtins.int, ...], numpy.dtype[numpy.floating[Any]]]"
Success: no issues found in 1 source file
```

mcwitt (approved):

maxentile (approved):

PR #1467, opened 2025-01-30
Rename DualTopologyMinimization -> DualTopology

proteneer:
This PR:

-Deprecates the old `DualTopology` class and removes its single usage from the deprecated `md/align.py` code.
-Renames the `DualTopologyMinimization` -> `DualTopology`.

Semantically, the only key difference was that for nonbonded **intermolecular** interactions, `DualTopology` ignored the `lamb` argument when setting the `w-coords`, where as `DualTopologyMinimization` obeyed it. There's no good use-case for the former.


timemachine/fe/topology.py
```diff
@@ -526,7 +526,6 @@ def parameterize_nonbonded(
         lamb: float,
```
jkausrelay, line 526: If lamb is not used, remove the arg?

proteneer: willfix

PR #1469, opened 2025-01-30
Assert energies are valid before HREX swaps

badisa:
* Being defensive here

### DIFF
--- a/timemachine/ff/handlers/nonbonded.py
+++ b/timemachine/ff/handlers/nonbonded.py
@@ -569,7 +569,7 @@ class NNHandler(SerializableMixIn):
 
     @staticmethod
     def static_parameterize(flat_params, encoded_unflatten_str, mol):
-        am1_charges = compute_or_load_oe_charges(mol, mode=AM1BCC)
+        am1_charges = compute_or_load_oe_charges(mol, mode=AM1BCCELF10)
         bond_idxs, deltas = NNHandler.get_bond_idxs_and_charge_deltas(flat_params, encoded_unflatten_str, mol)
         final_charges = apply_bond_charge_corrections(am1_charges, bond_idxs, jnp.array(deltas), runtime_validate=False)
         return final_charges


timemachine/fe/free_energy.py
```diff
@@ -1165,6 +1165,17 @@ def compute_dense():
...
+    diag_idxs = np.diag_indices_from(U_kl)
+    assert np.all(np.isfinite(U_kl[diag_idxs])), "A state is no longer valid"
+    assert np.all(np.abs(U_kl[diag_idxs]) < 1e9), "Energies larger in magnitude than tolerated"
+    if np.any(np.isnan(U_kl)):
+        warn(
+            "Encountered NaNs in U_kl matrix. Replacing each instance with inf prior to HREX swaps",
+            IndeterminateEnergyWarning,
+        )
+        U_kl = np.where(np.isnan(U_kl), np.inf, U_kl)
+
```
mcwitt, line 1178: nit: would prefer to put this code in a separate function (`sanitize_potential_matrix_for_hrex` or similar?) and update the usage in `run_sims_hrex` to something like

```python
U_kl_raw = compute_potential_matrix(...)
U_kl = sanitize_potential_matrix_for_hrex(U_kl_raw, abs_energy_threshold=1e9)
```

to avoid entangling `compute_potential_matrix` with HREX.

mcwitt: (The assertion that the energies have not blown up seems useful to have independently of HREX.)

mcwitt (approved):

timemachine/fe/free_energy.py
```diff
@@ -1165,6 +1165,17 @@ def compute_dense():
 
     U_kl = compute_sparse(max_delta_states) if max_delta_states is not None else compute_dense()
 
+    # Verify that all of the diagonal energies are finite, else a state has blown up
+    diag_idxs = np.diag_indices_from(U_kl)
+    assert np.all(np.isfinite(U_kl[diag_idxs])), "A state is no longer valid"
```
mcwitt, line 1170: super nit: any reason not to just do
```suggestion
    assert np.all(np.isfinite(np.diagonal(U_kl))), "Got non-finite energy while evaluating the energy of replica"
```
?

Could also make assertion message more descriptive.

timemachine/fe/free_energy.py
```diff
@@ -1168,10 +1168,13 @@ def compute_dense():
...
-def verify_and_sanitize_potential_matrix(U_kl: NDArray, abs_energy_threshold: float = 1e9) -> NDArray:
+def verify_and_sanitize_potential_matrix(
+    U_kl: NDArray, hrex: HREX[CoordsVelBox], abs_energy_threshold: float = 1e9
+) -> NDArray:
     """Ensure energies in the diagonal are finite and below some threshold and sanitizes NaNs to infs."""
-    # Verify that the energies along the diagonal are finite, else a replica is no longer valid
-    replica_energies = np.diagonal(U_kl)
+    # Verify that the energies that the energies of the replica in the same state are finite, else a replica is no longer valid
+    state_idx = np.argsort(hrex.replica_idx_by_state)
+    replica_energies = np.diagonal(U_kl[:, state_idx])
```
mcwitt, line 1177: nit: is this equivalent? (I might be confused)

```suggestion
    replica_energies = np.diagonal(U_kl[hrex.replica_idx_by_state, :])
```

mcwitt: nit: could this function just take `replica_idx_by_state` (I don't think it needs the whole `HREX` object?)

mcwitt (commented):
Good catch, I'd missed the bug on my first pass.

proteneer:
> Good catch, I'd missed the bug on my first pass.

Is there an obvious consequence of the bug? (i.e. does the presence of NaNs ever lead to accidental swaps of two neighboring states? eg. if

a) fwd_dU is nan, rev_dU is finite, or if
b) both fwd_dU, and rev_dU are nans

proteneer (approved):

proteneer:
> > Good catch, I'd missed the bug on my first pass.
> 
> Is there an obvious consequence of the bug? (i.e. does the presence of NaNs ever lead to accidental swaps of two neighboring states? eg. if
> 
> a) fwd_dU is nan, rev_dU is finite, or if b) both fwd_dU, and rev_dU are nans

edit: clarified offline with @badisa - not an issue

PR #1471, opened 2025-01-31
Refactor VacuumSystem and HostGuestSystem

proteneer:
We currently have two polymorphic quirks we have in our poorly named `VacuumSystem` class:

1) The `VacuumSystem` dataclass is currently used by both *host* system and *guest* systems.
2)  `VacuumSystem` has a polymorphic attribute `nonbonded`, that can be used to represent both a nonbonded pairlist, as well as the nonbonded all pairs potential.

This PR:

1) Renames `VacuumSystem` -> `GuestSystem`, and its `nonbonded` attribute to `nonbonded_pair_list`
2) Introduces `HostSystem`, a container:
  - that is devoid of `chiral_atom` and `chiral_bond` attributes since the upstream deserializers do not generate them
  - that adds the `nonbonded_all_pairs` attribute.
3) Keeps the `HostGuestSystem` class, which has three nonbonded attributes: `nonbonded_pair_list`, `nonbonded_all_pairs`, `nonbonded_ixn_group`
4) Introduces an `AbstractSystem` class, defining the methods `get_U_fn()` and `get_U_fns()`


tests/test_single_topology.py
```diff
@@ -739,15 +739,15 @@ def setup_intermediate_nonbonded_term_ref(src_nonbonded, dst_nonbonded, lamb, al
 
     for lamb in np.linspace(0.0, 1.0, 10):
         nonbonded_ref = setup_intermediate_nonbonded_term_ref(
-            st.src_system.nonbonded,
-            st.dst_system.nonbonded,
+            st.src_system.nonbonded_pair_list,
```
proteneer, line 742: I'm hoping this change makes it easier for the reader to be able to quickly infer if potential type for `src_system.nonbonded` is one of {pair_list, all_pairs, ixn_group}

timemachine/fe/rest/single_topology.py
```diff
@@ -117,7 +117,7 @@ def get_energy_scale_factor(self, lamb: float) -> float:
         temperature_factor = self._temperature_scale_interpolation_fxn(lamb).item()
         return 1.0 / temperature_factor
 
-    def setup_intermediate_state(self, lamb: float) -> VacuumSystem[NonbondedPairListPrecomputed, HarmonicAngleStable]:
```
proteneer, line 120: I guess this is one of the areas where we "lose" the additional "specificity" of the type, since now we return something more generic/encompassing.

proteneer: stale comment since we brought back specific systems

tests/test_single_topology.py
```diff
@@ -758,7 +758,7 @@ def setup_intermediate_nonbonded_term_ref(src_nonbonded, dst_nonbonded, lamb, al
 
 
 @pytest.mark.nocuda
-def test_combine_with_host():
+def test_combine_achiral_ligand_with_host():
```
proteneer, line 761: Note to reader: this test was split into two because the original `HostGuestSystem` and `VacuumSystem` classes had different implementations of `get_U_fns()`, one pruned all the potentials with 0-length params, and the other kept it. So I added a test to catch this going forward. i.e. we should be deleting potentials with 0-length parameters in the future.

timemachine/fe/system.py
```diff
@@ -137,41 +126,94 @@ def U_fn(x):
...
+            self.angle,
+            self.proper,
+            self.improper,
+            self.nonbonded_all_pairs,
+        ]
+        # (TODO): len(p.params) > 0 is dangerous if we later on have potentials
+        # that do *not* have "free" parameters defined but should still be left on.
+        # (eg. if the force constants for something like virtual sites are defined by
+        # properties/attributes on the class, rather than attributes)
+        return [p for p in potentials if len(p.params) > 0]
```
mcwitt, line 155: nit: is there a way to avoid repeating this code and comment? One idea I had is to change this to `_get_U_fns`, and add an implementation of `get_U_fns` to the base class that calls this and then does the filtering.

timemachine/fe/single_topology.py
```diff
@@ -1040,7 +1040,7 @@ def assert_bonds_defined_for_chiral_volumes(
                 raise MissingBondsInChiralVolumeException(f"bond {(c, k)} missing from Chiral Volume {(c, i, j, k)}")
 
 
-def assert_torsions_defined_over_non_linear_angles(system: VacuumSystem | HostGuestSystem):
+def assert_torsions_defined_over_non_linear_angles(system: GuestSystem | HostGuestSystem | HostSystem):
```
mcwitt, line 1043: nit: it feels more precise for this to take `angle`, `proper`, and `improper` as arguments (then it could work with future system containers with these potentials too)

timemachine/fe/system.py
```diff
@@ -137,41 +126,94 @@ def U_fn(x):
 
         return U_fn
 
+    @abstractmethod
+    def get_U_fns(self):
+        raise NotImplementedError()
+
+
+@dataclass  # mcwitt: Generic can be removed in python 3.12
+class HostSystem(Generic[_HarmonicAngle], AbstractSystem):
```
mcwitt, line 135: nit/future: I wonder if we can just get rid of the current `HarmonicAngle` and rename `HarmonicAngleStable` (since IIRC the former is a special case of the latter with `eps=0`). Then we could remove this polymorphism.

mcwitt (approved):
LGTM; only non-blocking comments

timemachine/fe/system.py
proteneer: yes, and lets defer it to a future case (since it will enlarge this PR considerably)

timemachine/fe/system.py
proteneer: I can clean this up a bit.

timemachine/fe/single_topology.py
proteneer: yes - but lets defer it, I have to update quite a number of tests to do this properly.

timemachine/fe/system.py
proteneer: I moved the `get_U_fns()` itself into the base abstract class, and added a bit of introspection magic to automagically extract the fields:

``` python

@dataclass
class AbstractSystem(ABC):
    def get_U_fn(self):
        """
        Return a jax function that evaluates the potential energy of a set of coordinates.
        """
        U_fns = self.get_U_fns()

        def U_fn(x):
            return sum(U(x, box=None) for U in U_fns)

        return U_fn

    def get_U_fns(self) -> List[BoundPotential]:
        """
        Return a list of bound potential"""
        potentials: List[BoundPotential] = []
        for f in fields(self):
            v = getattr(self, f.name)
            potentials.append(v)

        # (TODO): len(p.params) > 0 is dangerous if we later on have potentials
        # that do *not* have "free" parameters defined but should still be left on.
        # (eg. if the force constants for something like virtual sites are defined by
        # properties/attributes on the class, rather than attributes)
        return [p for p in potentials if len(p.params) > 0]
```

PR #1472, opened 2025-01-31
Update linters

mcwitt:
- clang-format: 16.0.6 -> 19.1.7
- mypy: 1.5.1 -> [1.14.1](https://mypy.readthedocs.io/en/stable/changelog.html#mypy-1-14)
- pre-commit: 3.4.1 -> 4.1.0
- pre-commit-hooks: 4.5.0 -> 5.0.0


timemachine/potentials/potential.py
```diff
@@ -13,7 +13,7 @@
 
 Precision = Any
 
-_P = TypeVar("_P", bound="Potential")
+_P = TypeVar("_P", bound="Potential", covariant=True)
```
mcwitt, line 16: This allows mypy to treat e.g. `BoundPotential[HarmonicBond]` as an instance of `BoundPotential[Potential]`, allowing more types to be inferred without explicit annotation.

timemachine/cpp/src/gpu_utils.cuh
```diff
@@ -24,7 +24,9 @@ curandStatus_t templateCurandUniform(curandGenerator_t generator, float *outputP
 curandStatus_t templateCurandUniform(curandGenerator_t generator, double *outputPtr, size_t n);
 
 #define gpuErrchk(ans)                                                                                                 \
-    { gpuAssert((ans), __FILE__, __LINE__); }
+    {                                                                                                                  \
```
proteneer, line 27: this looks kind of weird - esp with the end of line line-breaks (`\`) - maybe `#defines` aren't well handled?

proteneer (approved):

timemachine/cpp/src/gpu_utils.cuh
mcwitt: Will look into it, there might be an option controlling the behavior here.

proteneer: not a big deal tbh

mcwitt: I'm not seeing an obvious way to avoid this (though might be missing something, `clang-format` has a [bajillion options](https://clang.llvm.org/docs/ClangFormatStyleOptions.html)). Will leave as-is if not a big deal.

PR #1473, opened 2025-01-31
Restore dropped pycodestyle, Pyflakes rules in ruff

mcwitt:
In the transition to ruff (#1459), we unintentionally dropped the Pyflakes linting rules (e.g. unused imports) that were previously handled by `flake8`. This PR adds them back in by selecting the `F` rules in the ruff configuration ([reference](https://docs.astral.sh/ruff/linter/#rule-selection)).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,11 +12,13 @@ line-length = 120
 
 [tool.ruff.lint]
 select = [
+   "E", # pycodestyle
    "F", # Pyflakes
-   "I", # sort imports
+   "I", # isort
 ]
 ignore = [
   "E402", # Module level import not at top of file
+  "E501", # Line too long
   "E731", # Do not assign a lambda expression, use a def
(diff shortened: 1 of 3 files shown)


proteneer:
good catch - I was wondering why I can suddenly get away with unused imports...

proteneer:
Also: do we need to retroactively apply this to master?

mcwitt:
> Also: do we need to retroactively apply this to existing files that may have slipped through on master?

This shouldn't be necessary, since we run pre-commit with `--all-files` in CI: https://github.com/proteneer/timemachine/blob/b3308905dc226cea3961e2b6f3b5980678b13cae/Makefile#L35

mcwitt:
I did a little more looking into this and realized that we also need to explicitly enable the `E` (pycodestyle) rules for parity with the checks we had before. Added this in https://github.com/proteneer/timemachine/pull/1473/commits/e2c166d04a3a7b452fa20d89901a07cdfd0ea593.

PR #1474, opened 2025-02-04
Switch to CUB for Energy Accumulation

badisa:
* Avoid managing our own implementation of parallel sum, instead rely on CUB. I believe at the time of implementing our own version, I had convinced myself that CUB didn't support int128, which is not true.
* Does **not** resolve issue seen while running timemachine under MPS
* Appears that to be slightly slower than the old implementation, but within 1%. The reduced code maintenance seems worth it..


mcwitt:
Sanity check: have we verified that we get bitwise-identical results compared with master?

timemachine/cpp/src/fanout_summed_potential.cu
```diff
@@ -1,14 +1,22 @@
...
-    : potentials_(potentials), parallel_(parallel), d_u_buffer_(potentials_.size()) {};
+    : potentials_(potentials), parallel_(parallel), d_u_buffer_(potentials_.size()), sum_storage_bytes_(0) {
+
+    gpuErrchk(
+        cub::DeviceReduce::Sum(nullptr, sum_storage_bytes_, d_u_buffer_.data, d_u_buffer_.data, potentials_.size()));
+
+    gpuErrchk(cudaMalloc(&d_sum_temp_storage_, sum_storage_bytes_));
+};
+
+FanoutSummedPotential::~FanoutSummedPotential() { gpuErrchk(cudaFree(d_sum_temp_storage_)); };
```
mcwitt, line 19: can `DeviceBuffer` be used for cases like this where the lifetime of the buffer is the same as the lifetime of the object?

mcwitt (approved):
Nice! Looks like a pretty straightforward improvement.

One comment: it would be nice to use the RAII pattern (i.e. `DeviceBuffer`) for `d_sum_temp_storage_` (and other cases where buffer and object lifetime are always identical). Though maybe the reason not to was for consistency, since we have many other device buffers that are still manually allocated and freed?

mcwitt:
> require me to track down where failures are coming from manually

Ah, right, I'd forgotten about that (severe) limitation in the current implementation of `DeviceBuffer`. 

For some future PR, I wonder if there might be a way to overcome this, e.g. if there were a way to attach a backtrace to a custom exception class (perhaps using some more modern C++ features?).

In any case, agreed that it doesn't seem worth it for now.

timemachine/cpp/src/centroid_restraint.hpp
```diff
@@ -22,6 +22,9 @@ template <typename RealType> class CentroidRestraint : public Potential {
     double kb_;
     double b0_;
 
+    size_t sum_storage_bytes_;
```
proteneer, line 25: do you still need the internal `__int128 *d_u_buffer_;`?

timemachine/cpp/src/centroid_restraint.cu
```diff
@@ -65,12 +61,9 @@ void CentroidRestraint<RealType>::execute_device(
             kb_,
             b0_,
             d_du_dx,
-            d_u == nullptr ? nullptr : d_u_buffer_);
+            d_u // Can write directly to the energy buffer for this potential.
```
proteneer, line 64: I think this is pretty dangerous - the input API assumes that `__int128 *d_u` can set to some initial value `x`, and this function can increment it by `y` resulting in `x+y` (like the forces). However, the actual internal kernel call directly *sets* `d_u` via:
``` cpp
//  k_centroid_restraint.cuh
if (t_idx == 0 && d_u) {
    RealType nrg = kb * (dij - b0) * (dij - b0);
    d_u[t_idx] = FLOAT_TO_FIXED_ENERGY<RealType>(nrg);
}
```
i.e. it blows away the old value  of `x` - we should talk about the semantics of `execute_device` more carefully tmrw.

timemachine/cpp/src/kernels/k_nonbonded.cuh
```diff
@@ -532,19 +538,18 @@ void __global__ k_compute_nonbonded_target_atom_energies(
...
                     // Store the atom by atom energy
-                    block_energy_buffer[threadIdx.x] = FLOAT_TO_FIXED_ENERGY<RealType>(u);
+                    energy_accumulator = FLOAT_TO_FIXED_ENERGY<RealType>(u);
                 }
             }
-            // Sync to ensure the shared buffers are populated
-            __syncthreads();
 
-            block_energy_reduce<THREADS_PER_BLOCK>(block_energy_buffer, threadIdx.x);
+            // Sum's return value is only valid in thread 0
```
proteneer, line 545: interesting switch of the location of the `__syncthreads()` - from before the reduce to after the reduce - intentional?

proteneer: i.e. guessing that BlockReduce::Sum() implicitly calls a syncthreads

proteneer (approved):
nice changes!

PR #1478, opened 2025-02-05
Fix nightly test failures

proteneer:


proteneer:
@jkausrelay - this is a hot fix. I'm waiting on one last test to finish in the nightlies then it should be good to go.

jkausrelay (approved):

PR #1480, opened 2025-02-05
Avoids warning if Bisection reaches min overlap on the final iteration

badisa:
Saw this recently with the following message

> Reached n_bisections=46 iterations without achieving min_overlap=0.667. The minimum BAR overlap was 0.678

### DIFF
--- a/examples/water_sampling_common.py
+++ b/examples/water_sampling_common.py
@@ -6,6 +6,7 @@ from openmm import app
 from timemachine.constants import AVOGADRO, DEFAULT_PRESSURE, DEFAULT_TEMP, NBParamIdx
 from timemachine.fe.free_energy import AbsoluteFreeEnergy, HostConfig, InitialState
 from timemachine.fe.model_utils import apply_hmr
+from timemachine.fe.system import HostSystem
 from timemachine.fe.topology import BaseTopology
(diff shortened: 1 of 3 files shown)


timemachine/fe/free_energy.py
```diff
@@ -1103,7 +1103,7 @@ def compute_intermediate_result(lambdas: Sequence[float]) -> PairBarResult:
         result = compute_intermediate_result(lambdas)
         results.append(result)
     else:
-        if min_overlap is not None:
+        if min_overlap is not None and np.min(result.overlaps) < min_overlap:
```
mcwitt, line 1106: nit: use native Python `min`, since `result.overlaps` is a list
```suggestion
        if min_overlap is not None and min(result.overlaps) < min_overlap:
```

mcwitt (approved):

timemachine/fe/free_energy.py
mcwitt: Hmm, on second thought, maybe it is better to use the numpy version here because the behavior with NaNs is more consistent:

```
>>> min([1.0, np.nan])
1.0
>>> np.min([1.0, np.nan])
np.float64(nan)
```

Non-blocking, OK either way.

PR #1481, opened 2025-02-05
Avoid races in neighborlist

badisa:
* Removes shared memory for `sync_start`, instead using warp syncs. Probably all around better
* Adds `__syncthreads();` everywhere to avoid the race checker complaining. Think this is over kill, but want to sanity check things

### DIFF
--- a/timemachine/cpp/src/kernels/k_neighborlist.cuh
+++ b/timemachine/cpp/src/kernels/k_neighborlist.cuh
@@ -133,7 +133,7 @@ void __global__ k_compact_trim_atoms(
     const int warpMask = (1 << indexInWarp) - 1;
     const int row_block_idx = blockIdx.x;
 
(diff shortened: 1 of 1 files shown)


proteneer:
Does this clear up the error messages from `--tool=racecheck` in the compute-sanitizer?

proteneer:
(and to be clear, the output of the neighborlist `interactingTiles` and `interactingAtoms` are still non-deterministic, but this addresses the *intra*-warp race conditions)

proteneer:
Thanks, do these two lines trigger anything in racecheck?:

https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/kernels/k_nonbonded.cuh#L27
https://github.com/proteneer/timemachine/blob/master/timemachine/cpp/src/kernels/k_nonbonded.cuh#L53

PR #1101, opened 2023-07-26
Refactor Barostat

badisa:
# What Changed
- Refactors the barostat as I didn't know what I was doing when I wrote this originally
- Removes some unnecessary memsets introduced in #1090 that didn't need to be there
- Minor optimizations to the barostat that make no noticeable difference to the wall clock time

# Benchmarks
- Nvidia A10
- Cuda Arch 86

## Benchmarking
Using the `python tests/test_benchmark.py` we see minor differences. Multiple runs showed the differences to be noise.

### Current
```
dhfr-apo-barostat-interval-25: N=23558 speed: 622.98ns/day dt: 2.5fs (ran 100000 steps in 34.68s)
hif2a-apo-barostat-interval-25: N=8805 speed: 1002.75ns/day dt: 2.5fs (ran 100000 steps in 21.54s)
solvent-apo-barostat-interval-25: N=6282 speed: 1278.64ns/day dt: 2.5fs (ran 100000 steps in 16.90s)
```
### Barostat changes
```
dhfr-apo-barostat-interval-25: N=23558 speed: 621.48ns/day dt: 2.5fs (ran 100000 steps in 34.76s)
hif2a-apo-barostat-interval-25: N=8805 speed: 1006.45ns/day dt: 2.5fs (ran 100000 steps in 21.46s)
solvent-apo-barostat-interval-25: N=6282 speed: 1266.70ns/day dt: 2.5fs (ran 100000 steps in 17.06s)
```

## On a per kernel basis
- Nvidia A4000
- Cuda Arch 86

Looking at the nsys stats, there is slight improvement in some kernels. The most significant difference in total time is the `gen_sequenced` only being called once every 1000 moves. Computing energies is slow enough that this is negligible.

### Current

```
CUDA Kernel Statistics:

 Time (%)  Total Time (ns)  Instances  Avg (ns)   Med (ns)   Min (ns)  Max (ns)  StdDev (ns)                                                  Name                                                
 --------  ---------------  ---------  ---------  ---------  --------  --------  -----------  ----------------------------------------------------------------------------------------------------
      0.1        1,196,193         80   14,952.4   15,552.0    11,744    16,831      1,533.4  timemachine::k_decide_move(int, int, double, double, const double *, double *, double *, const __in…
      0.1          822,368         80   10,279.6   10,096.0     9,664    16,992        821.0  timemachine::rescale_positions(int, double *, const double *, const double *, double *, const int *…
[...]

PR #1484, opened 2025-02-06
Fixes Race Condition in MonteCarloBarostat

badisa:
Resolves a race condition that could lead to the coordinates set by the Barostat being inconsistent. This was due to some thread blocks seeing the move as accepted, while others saw it as rejected. Not a reliable test, but large systems fail more often due to the difficultly of scheduling all of the blocks at the same time. 

Bug introduced in https://github.com/proteneer/timemachine/pull/1101


tests/test_barostat.py
```diff
@@ -294,6 +294,72 @@ def test_barostat_is_deterministic():
     assert compute_box_volume(atm_box) == compute_box_volume(ctxt.get_box())
 
 
+def test_barostat_is_deterministic_on_large_systems():
+    """Verify that the barostat results in the same box size shift after a fixed number of steps
+    This is important to debugging as well as providing the ability to replicate
+    simulations.
+
+    This test is not reliable. It tests a race condition in which
```
proteneer, line 302: unfinished sentence

timemachine/cpp/src/kernels/k_barostat.cuh
```diff
@@ -92,23 +92,25 @@ template <typename RealType>
...
     }
     const RealType volume = d_box[0 * 3 + 0] * d_box[1 * 3 + 1] * d_box[2 * 3 + 2];
-    if (d_volume_scale[0] == 0.0 && adaptive) {
-        d_volume_scale[0] = 0.01 * volume;
-    }
     const RealType delta_volume = d_volume_scale[0] * 2 * (rand[0] - 0.5);
     const RealType new_volume = volume + delta_volume;
-    d_volume_delta[0] = delta_volume;
-    d_length_scale[0] = cbrt(new_volume / volume);
+    if (adaptive && *d_volume_scale == 0.0) {
```
proteneer, line 108: is this reordering intended?


if:
1) `adaptive == True`
2) `d_volume_scale[0] == 0.0`
3) `volume == 1.0`

before:

`const RealType delta_volume = 0.01  * 2 * (rand[0] - 0.5);` # this is not zero

after:

`const RealType delta_volume = 0  * 2 * (rand[0] - 0.5);` # this is zero

proteneer: off topic, but does it make sense for pre-commit to run a `make clean && make`? (or at least a `make`)

timemachine/cpp/src/kernels/k_barostat.cuh
```diff
@@ -121,20 +123,23 @@ void __global__ k_decide_move(
...
+    const double *__restrict__ d_box_output,     // [3*3]
+    double *__restrict__ d_x,                    // [N*3]
+    const double *__restrict__ d_x_output,       // [N*3]
+    int *__restrict__ num_accepted,              // [1]
+    int *__restrict__ num_attempted              // [1]
+) {
     int idx = blockIdx.x * blockDim.x + threadIdx.x;
 
-    const RealType volume = d_box[0 * 3 + 0] * d_box[1 * 3 + 1] * d_box[2 * 3 + 2];
+    // Don't compute volume from the box. It leads to a race condition since `d_box` is updated in this kernel
```
proteneer, line 141: I'd also be in favor of renaming `d_x_output` and `d_box_output` to  `d_x_proposed` and `d_box_proposed`.

This actually tripped me up on first read since usually things that are `_output` are *not* `const`-qualified, here, it's actually an `_input` to the function. I had to read it a second time to catch it.

tests/test_barostat.py
```diff
@@ -232,8 +232,16 @@ def test_barostat_partial_group_idxs():
     ctxt.multiple_steps(barostat_interval * 100)
 
 
-@pytest.mark.memcheck
-def test_barostat_is_deterministic():
+@pytest.mark.parametrize(
+    "box_width, iterations",
+    [
+        pytest.param(3.0, 20, marks=pytest.mark.memcheck),
+        # fyork: This test only fails 50-50 times. It tests a race condition in which the coordinates could be corrupted.
```
mcwitt, line 239: Could we make the system bigger or increase the number of iterations to increase the failure rate (assuming this would make a significant difference)? (this only takes ~seconds on my machine)

mcwitt: Also, is it possible to trigger the error without running MD in between the barostat moves?

timemachine/cpp/src/kernels/k_barostat.cuh
mcwitt: Posting offline resolution of pre-commit issue for reference: we currently do not check that the C++ code still compiles with just `pre-commit run` (unless `wrap_kernels.cpp` was modified in the diff). (`pre-commit run --all-files`, which is run by CI, should catch this).

tests/test_barostat.py
mcwitt: > the comment refers to the settings below the comment

oops, sorry, I had misread this and run the wrong set of parameters. Makes sense that we probably don't want to go bigger for a unit test.

mcwitt: > Possibly? Though it only triggers on accepted moves, so would have to construct a system that reliably accepts moves.

One idea (not necessary for this PR): scale the coordinates by (e.g. by 0.99) between each barostat move?

mcwitt: > > Possibly? Though it only triggers on accepted moves, so would have to construct a system that reliably accepts moves.
> 
> One idea (not necessary for this PR): scale the coordinates by (e.g. by 0.99, tune to get desired acceptance rate) between each barostat move?

Disregard since it sounds like the symptom is a blowup in MD; would need to figure another way to detect the invalid behavior, which is probably out of scope for this PR.

proteneer (approved):

mcwitt (approved):

PR #1485, opened 2025-02-06
Remove stubgen workaround

mcwitt:
Trying something similar to #1479. We need to update `custom_ops.pyi`, but this seems OK with current versions (mypy checks pass).


mcwitt:
OK, I see why this is still broken without the workaround. The generated `pyi` file has invalid annotations like `numpy.ndarray[numpy.float64]` (the dtype is supposed to be the second type parameter, see e.g. https://numpy.org/devdocs/reference/typing.html#numpy.typing.NDArray)

Issue #1100, opened 2023-07-26
Inconsistent conventions for axis ordering of `u_kln` matrices


mcwitt:
Sometimes our usage aligns with the pymbar convention where the first two dimensions correspond to sampling and evaluation state respectively, e.g.: https://github.com/proteneer/timemachine/blob/c4369bdd3accbb88d48cabd24c9dc209d19d94f6/timemachine/maps/estimators.py#L54-

and sometimes we use a convention where the meaning of `k` and `l` is swapped, e.g.:
https://github.com/proteneer/timemachine/blob/c4369bdd3accbb88d48cabd24c9dc209d19d94f6/timemachine/fe/energy_decomposition.py#L82-L88

Given that inconsistencies can easily lead to bugs, it would be useful to align on a single convention.

PR #1486, opened 2025-02-07
switch to use PyMBAR axis convention for u_kln

jkausrelay:
- Update the u_kln to use the PyMBAR convention (k is the state index, l is the parameter index)


tests/test_free_energy.py
```diff
@@ -565,7 +565,7 @@ def test_estimate_free_energy_bar_with_energy_overflow():
     assert np.isfinite(result_with_nan.dG_err)
 
     u_kln_with_inf = np.array(u_kln)
-    u_kln_with_inf[0, 1, 10] = np.inf
+    u_kln_with_inf[1, 0, 10] = np.inf
```
mcwitt, line 568: curiosity: does this change (and the one on L553) make any difference?

mcwitt (approved):
Nice! All of the changes look correct to me.

tests/test_free_energy.py
jkausrelay: No it doesn't. I just wanted to be consistent with what the previous code was doing.

maxentile (approved):
Switches to the convention 

```python
 u_kn[k][n] = u_k(x_n)
 u_kln[k][l][n] = u_l(trajs[k][n])
```

resolving https://github.com/proteneer/timemachine/issues/1100 .

------------

This increases consistency with a convention in MBAR.

A less ambiguous format to consider in the future might be something like `(u_kn, s_n)`, where 
* `u_kn = u_k(x_n)`, 
* `s_n` is state of origin for sample `x_n`.

PR #1493, opened 2025-02-10
Adds Example for Running RBFE pair

badisa:
* Removes `rbfe_edge_list.py` as it is not used in practice and differs from our best practices
* Adds `run_rbfe_legs.py` to simulates legs for a pair of molecules. Intended as a summary of our best practices. The initial pass covers the most obvious parameters to adjust, though over time we may want to add more.

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -4,7 +4,6 @@ import os
 import unittest
 from collections.abc import Iterator
 from dataclasses import dataclass, replace
-from importlib import resources
 from tempfile import TemporaryDirectory
 from typing import Optional
 
@@ -24,6 +23,7 @@ from timemachine.md.builders import build_protein_system
 from timemachine.potentials import Nonbonded
 from timemachine.potentials.potential import GpuImplWrapper
 from timemachine.potentials.types import PotentialFxn
+from timemachine.utils import path_to_internal_file
 
 HILBERT_GRID_DIM = 128
 
@@ -446,7 +446,7 @@ def check_split_ixns(
 ):
     ffs = load_split_forcefields()
 
-    with resources.path("timemachine.testsystems.data", "hif2a_nowater_min.pdb") as path_to_pdb:
+    with path_to_internal_file("timemachine.testsystems.data", "hif2a_nowater_min.pdb") as path_to_pdb:
         host_config = build_protein_system(str(path_to_pdb), ffs.ref.protein_ff, ffs.ref.water_ff)
         host_config.box += np.diag([0.1, 0.1, 0.1])
 
--- a/tests/nonbonded/test_nonbonded_mol_energy.py
+++ b/tests/nonbonded/test_nonbonded_mol_energy.py
@@ -1,5 +1,3 @@
-from importlib import resources
-
 import numpy as np
 import pytest
 from common import assert_energy_arrays_match
@@ -13,6 +11,7 @@ from timemachine.lib.fixed_point import fixed_to_float
 from timemachine.md import builders, minimizer
 from timemachine.md.barostat.utils import get_bond_list, get_group_indices
 from timemachine.md.exchange.exchange_mover import BDExchangeMove, randomly_rotate_and_translate
+from timemachine.utils import path_to_internal_file
 
 
 @pytest.mark.memcheck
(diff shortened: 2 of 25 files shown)


examples/run_rbfe_legs.py
```diff
@@ -0,0 +1,254 @@
...
+    leg_name: str,
+    mol_a: Chem.Mol,
+    mol_b: Chem.Mol,
+    core: NDArray,
+    ff: Forcefield,
+    pdb_path: str | None,
+    md_params: MDParams,
+    n_windows: int,
+    min_overlap: float,
+):
```
jkausrelay, line 48: nit: Add docs

jkausrelay (approved):

tests/test_examples.py
```diff
@@ -361,3 +214,141 @@ def test_water_sampling_mc_buckyball(batch_size, insertion_type):
...
+
+                for ref_state, comp_state in zip(
+                    ref_res.final_result.initial_states, comp_res.final_result.initial_states
+                ):
+                    np.testing.assert_array_equal(ref_state.x0, comp_state.x0)
+                    np.testing.assert_array_equal(ref_state.v0, comp_state.v0)
+                    np.testing.assert_array_equal(ref_state.box0, comp_state.box0)
+                    np.testing.assert_array_equal(ref_state.ligand_idxs, comp_state.ligand_idxs)
+                    np.testing.assert_array_equal(ref_state.protein_idxs, comp_state.protein_idxs)
+                    assert len(ref_state.potentials) == len(comp_state.potentials)
```
mcwitt, line 344: nit: might be overkill, but could you use [`assert_deep_eq`](https://github.com/proteneer/timemachine/blob/63e601360c1c05005c801690d197d6b922f46284/timemachine/fe/free_energy.py#L858) here?

examples/run_rbfe_legs.py
```diff
@@ -0,0 +1,291 @@
...
+)
+from timemachine.fe.utils import get_mol_name, plot_atom_mapping_grid, read_sdf_mols_by_name
+from timemachine.ff import Forcefield
+from timemachine.md.exchange.utils import get_radius_of_mol_pair
+from timemachine.parallel.client import CUDAPoolClient, FileClient
+from timemachine.parallel.utils import get_gpu_count
+
+
+def run_leg(
+    file_client: FileClient,
```
mcwitt, line 38: nits:

1. could the type here be `AbstractFileClient` (do we need to commit to a specific implementation?)
2. could this default to `FileClient()` (i.e., write files locally if left unspecified)

examples/run_rbfe_legs.py
```diff
@@ -0,0 +1,291 @@
...
+    leg_name: str,
+    mol_a: Chem.Mol,
+    mol_b: Chem.Mol,
+    core: NDArray,
```
mcwitt, line 42: tiny nit / style opinion: the ordering `mol_a`, `mol_b`, `core`, `leg_name` seems more natural to me (i.e. `(*edge, leg_name)`)

examples/run_rbfe_legs.py
```diff
@@ -0,0 +1,291 @@
...
+from timemachine.fe.utils import get_mol_name, plot_atom_mapping_grid, read_sdf_mols_by_name
+from timemachine.ff import Forcefield
+from timemachine.md.exchange.utils import get_radius_of_mol_pair
+from timemachine.parallel.client import CUDAPoolClient, FileClient
+from timemachine.parallel.utils import get_gpu_count
+
+
+def run_leg(
+    file_client: FileClient,
+    leg_name: str,
```
mcwitt: nit:
```suggestion
    leg_name: Literal["complex", "solvent", "vacuum"],
```

examples/run_rbfe_legs.py
mcwitt: > It could, but not clear what the value of that would be.

Was just thinking that `run_leg` might be a useful utility outside of the usage in the `__main__` block, and in that case it seems nice to not have to remember the `FileClient` API to run a simple job. But this seems OK given that `run_leg` is not intended to be used elsewhere.

mcwitt (approved):

PR #1495, opened 2025-02-12
Make sure overlap can't be negative

jkausrelay:
- Update to make sure the overlap can't be negative due to rounding
- Revert to old method of computing the overlap which seems to be more stable

### DIFF
--- a/tests/test_bar.py
+++ b/tests/test_bar.py
@@ -195,7 +195,9 @@ def test_pair_overlap_from_ukln():
 
     # non-overlapping
     u_kln, _ = make_gaussian_ukln_example((0, 0.01), (1, 0.01))
+    assert pair_overlap_from_ukln(u_kln) >= 0.0
     assert pair_overlap_from_ukln(u_kln) < 1e-10
+    print(pair_overlap_from_ukln(u_kln))
 
     # overlapping
     u_kln, _ = make_gaussian_ukln_example((0, 0.1), (0.5, 0.2))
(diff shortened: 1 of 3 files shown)


timemachine/fe/bar.py
```diff
@@ -322,7 +322,7 @@ def pair_overlap_from_ukln(u_kln: NDArray) -> float:
 
     """
     u_kn, N_k = ukln_to_ukn(u_kln)
-    return pymbar.mbar.MBAR(u_kn, N_k).compute_overlap()["scalar"]  # type: ignore
+    return max(0, 2 * pymbar.MBAR(u_kn, N_k).compute_overlap()["matrix"][0, 1])  # type: ignore
```
mcwitt, line 325: Noting that this returns 0 when `compute_overlap()` returns NaN. Might want to note this in the docstring or use `np.maximum` instead?

maxentile, line 325: Might:
(1) clamp output to `0 <= output <= 1`, (since, on bad inputs, entries of the matrix can also exceed 1.0)
(2) display a warning and/or throw an error if `overlap > (1 + epsilon)` or `overlap < (0 - epsilon)`, (this probably only happens when the estimate is also unreliable)

tests/test_bar.py
```diff
@@ -195,7 +195,9 @@ def test_pair_overlap_from_ukln():
 
     # non-overlapping
     u_kln, _ = make_gaussian_ukln_example((0, 0.01), (1, 0.01))
+    assert pair_overlap_from_ukln(u_kln) >= 0.0
     assert pair_overlap_from_ukln(u_kln) < 1e-10
+    print(pair_overlap_from_ukln(u_kln))
```
mcwitt, line 200: Intentional to leave `print` in?

mcwitt (approved):

maxentile (approved):

timemachine/fe/bar.py
jkausrelay: - clamped to `0 <= output <= 1`

PR #1497, opened 2025-02-12
Specify L-BFGS-B in MLE

maxentile:
Speed up by ~10-100x (depending on num nodes)

### DIFF
--- a/timemachine/fe/mle.py
+++ b/timemachine/fe/mle.py
@@ -115,7 +115,7 @@ def infer_node_vals(edge_idxs, edge_diffs, edge_stddevs, ref_node_idxs=tuple(), 
 
     K = np.max(edge_idxs) + 1
     x0 = np.zeros(K)  # maybe initialize smarter, e.g. using random spanning tree?
-    result = minimize(loss, x0, jac=True, tol=0).x
+    result = minimize(loss, x0, jac=True, tol=0, method="L-BFGS-B").x
 
     centered_node_vals = result - result[0]
 


timemachine/fe/mle.py
```diff
@@ -115,7 +115,7 @@ def loss(x):
 
     K = np.max(edge_idxs) + 1
     x0 = np.zeros(K)  # maybe initialize smarter, e.g. using random spanning tree?
-    result = minimize(loss, x0, jac=True, tol=0).x
+    result = minimize(loss, x0, jac=True, tol=0, method="L-BFGS-B").x
```
maxentile, line 118: If not specified, this defaults to method `"BFGS"`, whose per-iteration cost scales quadratically with `len(x0)` -- eventually dominating the cost of computing the value and gradient of `loss` for large num nodes.

Specifying `"L-BFGS-B"` (with default history size `maxcor` of 10) reduces the per-iteration cost to something closer to linear in `len(x0)`, without greatly increasing the number of required optimizer iterations, resulting in a meaningful speed-up overall for num nodes > some small threshold.

-----

Note: There is also a `jax.scipy.optimize.minimize` implementation of L-BFGS-B, but it's currently marked experimental

https://github.com/jax-ml/jax/blob/f7e2901e8b3a7ef2ebc68110cc0d1b48ed9ab6ca/jax/_src/scipy/optimize/minimize.py#L121

(and wasn't faster than the trusty implementation wrapped in `scipy` on random MLE instances)

-----

Note: For the Gaussian likelihood model, there is also a linear algebra solution https://github.com/forcefield/DiffNet/blob/63da35032dbe9e9b10783499fb4fe067fe1a45b4/diffnet.py#L475-L544  (as noted in earlier reviews https://github.com/proteneer/timemachine/pull/399 , https://github.com/proteneer/timemachine/pull/932 ), which is still faster than the PR implementation for instances with num nodes < some threshold ~= 800-1000.

PR #1498, opened 2025-02-12
Remove trivial assertions

mcwitt:
We currently assert that Python weakrefs to C++ objects are cleaned up after `gc.collect()`, but in general they can be cleaned up any time after all Python references go out of scope.


mcwitt:
Cherry-picked into https://github.com/proteneer/timemachine/pull/1488

PR #1500, opened 2025-02-13
Removes warnings about deprecated usage of importlib resources

badisa:
* Define a single function so we can just change it in one place going forward


mcwitt (approved):
Nice! LGTM.

PR #1503, opened 2025-02-15
Release GIL on Context calls, Parallelize HREX with threads

badisa:
* After the fixes #1464 and #1484, I suspect the issues we had seen previously with threading were actually exposing those bugs.
* Should be able to get further performance by leveraging threads, especially for vacuum simulations (though batching would be the most efficient method).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ NOGPU_MARKER := nogpu
 MEMCHECK_MARKER := memcheck
 NIGHTLY_MARKER := nightly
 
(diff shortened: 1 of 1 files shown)


mcwitt:
Drive-by comment/question: assuming the overhead of constructing a `ThreadPool` is negligible compared to sampling, could the logical changes be isolated inside a function (e.g. called `sample_parallel`) that takes and returns a list of replicas? This seems like it might better avoid obscuring the core HREX logic.

PR #1506, opened 2025-02-18
Verify Bitwise Reproducibility of RBFE Endstate Trajectories

badisa:
* Adds a `pytest-archive` directory to CI to get out artifacts, in case we can't emulate the CI (which I can't seem to do for RBFE, while I can for the Buckyball test case)
* Adds hashes of the endstates of RBFE runs so we can identify when simulations changes. Decided on hashes since the file sizes are large. This does **not** detect changes to the dGs.

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -254,17 +254,17 @@ def test_run_rbfe_legs(
     # Hashes are of results.npz, lambda0_traj.npz and lambda1_traj.npz respectively.
     leg_results_hashes = {
         "vacuum": (
-            "438baa7efaf19e1c8e2986f573e20593c48317650c48de129262a12d149d2d17",
+            "e71c2c30a18ea154f4eb6c6fb544aeadb0e9dda91c592115cd3eb6f40947df42",
             "c74ba1b503532af7fe83be82090cca82c33f5fa72744e74613dcaad9d978283a",
             "59ca76d1c5d43925ae6b163667c4b3e60f6c765ce10ad85d4f1d14c7d7854933",
         ),
         "solvent": (
-            "3a7edd9d3aed109c7478baa4fac16fd65959a4cbf878307e91cedb697435584d",
+            "e0ea86b104aeee1f1940e954b85df4f28150cc9561a3b932468d339479c3cf75",
             "4bc121e901f0719ae6fc01b84443ecf48738d28b581d5fc881a486e361406e7a",
             "53b75a556b5f63c90af108144bac5bace681bed581f1332b4e5fa3da8ac659a6",
         ),
         "complex": (
-            "c5e91675d7b396fa19429a20b70ec69a90b8f8ded4399bc5b15c6ae62a8579fa",
+            "59d62aa748f483aba93e83645c2e3512dd14ad9e9a8333fcce7053f4a9779892",
             "3e2fbf623b2c89eb4833a7812df00b3108cc977f97fef7ac68bdec40ef1b5fb2",
             "4ca200408ac0525bf89d992df0e75caeb2265aaaef577942d12cb456c1078cd2",
         ),


tests/test_examples.py
```diff
@@ -236,118 +248,142 @@ def test_run_rbfe_legs(
...
+                traj_data = np.load(str(leg_dir / f"lambda{lamb:d}_traj.npz"))
+                assert len(traj_data["coords"]) == n_frames
+                assert len(traj_data["boxes"]) == n_frames
+
+        def verify_endstate_hashes(output_dir: Path):
+            leg_dir = output_dir / leg
+            endstate_0_hash = hash_file(leg_dir / "lambda0_traj.npz")
+            endstate_1_hash = hash_file(leg_dir / "lambda1_traj.npz")
+            assert endstate_0_hash == endstate_hashes[leg][0] and endstate_1_hash == endstate_hashes[leg][1], (
+                f"{endstate_0_hash} != {endstate_hashes[leg][0]} and/or {endstate_1_hash} != {endstate_hashes[leg][1]}"
```
mcwitt, line 331: nit: probably don't need to repeat the hashes in the assert message, since pytest should print the diff anyway?

tests/test_examples.py
```diff
@@ -236,118 +248,142 @@ def test_run_rbfe_legs(
...
-                sdf_path=hif2a_dir / "ligands.sdf",
-                pdb_path=hif2a_dir / "5tbm_prepared.pdb",
-                seed=seed,
-                legs=leg,
-                n_eq_steps=n_eq_steps,
-                n_frames=n_frames,
-                n_windows=n_windows,
-                # Use simple charges to avoid os-dependent charge differences
-                forcefield="smirnoff_1_1_0_sc.py",
+    # Can generate hashes from CI artifacts
```
mcwitt, line 251: Might be worth documenting the process for updating one of these hashes. E.g. when we intentionally break bitwise reproducibility, can we just copy the new hash printed by the failed CI job here?

tests/common.py
```diff
@@ -26,6 +26,8 @@
 from timemachine.utils import path_to_internal_file
 
 HILBERT_GRID_DIM = 128
+# Directory to write files to that will be stored as artifacts in CI.
+ARTIFACT_DIR_NAME = "pytest-artifacts"
```
mcwitt, line 30: Is there any benefit to this being an env var? (e.g. so it could be read by both the CI configuration and the pytest process?)

tests/test_examples.py
```diff
@@ -236,118 +248,142 @@ def test_run_rbfe_legs(
...
+            sdf_path=hif2a_dir / "ligands.sdf",
+            pdb_path=hif2a_dir / "5tbm_prepared.pdb",
+            seed=seed,
+            legs=leg,
+            n_eq_steps=n_eq_steps,
+            n_frames=n_frames,
+            n_windows=n_windows,
+            # Use simple charges to avoid os-dependent charge differences
+            forcefield="smirnoff_1_1_0_sc.py",
+            output_dir=f"{ARTIFACT_DIR_NAME}/rbfe_{mol_a}_{mol_b}_{leg}_{seed}",
```
mcwitt, line 279: Does each CI run get a fresh artifact directory? (not sure how this works on the gitlab side...)

If not, should we add an ID or timestamp to the path to avoid races with concurrent CI jobs?

proteneer:
1) Can you add a comment in the code to document the right way to update the artifacts?
2) Should the water sampling tests also be updated to use the pytest-artifact mechanism?

tests/test_examples.py
```diff
@@ -236,118 +248,142 @@ def test_run_rbfe_legs(
...
+            assert isinstance(results["overlaps"], np.ndarray)
+            assert all(isinstance(overlap, float) for overlap in results["overlaps"])
+
+            for lamb in [0, 1]:
+                traj_data = np.load(str(leg_dir / f"lambda{lamb:d}_traj.npz"))
+                assert len(traj_data["coords"]) == n_frames
+                assert len(traj_data["boxes"]) == n_frames
+
+        def verify_endstate_hashes(output_dir: Path):
+            leg_dir = output_dir / leg
```
proteneer, line 327: any reason not to just recursively hash everything in the subdirectory? (vs. the current behavior of only hashing the trajectory npz file)

proteneer:
I think we should add some more documentation about the intended behavior/consistency:

Do we:
0) expect inputs to the trajectories to be bitwise identical? (assumed yes) 
1) expect trajectories to be bitwise identical? (assumed yes)
2) expect estimated `dGs` and `dG_errs` to be bitwise identical? 
3) expect analysis files to be bitwise identical?

tests/test_examples.py
mcwitt: Ah, I see, makes sense to avoid the need to run the test twice to update both hashes.

I suppose comparing tuples in the assertion would also print both hashes as desired?

```python
assert (endstate_0_hash, endstate_1_hash) == (endstate_hashes[leg][0], endstate_hashes[leg][1])
```

proteneer (approved):

mcwitt (approved):

PR #1508, opened 2025-02-20
Add invert_permutation utility

mcwitt:
Adds a minor utility `invert_permutation`, used to invert a permutation of the integers 1..len(p). The main purpose is to clarify the intent of the code at usage sites.


mcwitt:
We probably need both numpy and JAX versions (e.g. the version in this PR is incompatible with `jax.vmap`). Closing for now, since this will require more thought.

PR #1510, opened 2025-02-24
Add support for None bond anchors

proteneer:
This PR adds support for handling dummy groups that are completely unanchored (i.e. devoid of bond anchors). The previous behavior was to omit the unanchored dummy groups entirely from the output of  `generate_dummy_group_assignments`.

### DIFF
--- a/tests/test_dummy.py
+++ b/tests/test_dummy.py
@@ -1,7 +1,11 @@
 import pytest
 from rdkit import Chem
 
-from timemachine.fe.dummy import generate_dummy_group_assignments
+from timemachine.fe.dummy import (
+    ZeroBondAnchorWarning,
+    generate_anchored_dummy_group_assignments,
+    generate_dummy_group_assignments,
+)
 from timemachine.graph_utils import convert_to_nx
 
 # These tests check the various utilities used to turn off interactions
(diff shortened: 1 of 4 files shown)


timemachine/fe/single_topology.py
```diff
@@ -664,9 +664,8 @@ def concatenate(arrays, empty_shape, empty_dtype):
     )
 
     num_atoms = mol_a.GetNumAtoms() + mol_b.GetNumAtoms() - len(core)
-    assert get_num_connected_components(num_atoms, bond_potential.potential.idxs) == 1, (
-        "hybrid molecule has multiple connected components"
-    )
+    if get_num_connected_components(num_atoms, bond_potential.potential.idxs) > 1:
+        warnings.warn("Hybrid molecule has multiple connected components")
```
maxentile, line 668: Is there some weaker property than `num_connected_components == 1` we should still assert here?

I don't know what the correct behavior is, but maybe something like "for every `connected_component` in the input topology, assert that `num_connected_components(attach_dumy_atoms(connected_component)) == 1`"

proteneer: I thought about this a bit... and I'm still not sure how to implement something like `attach_dumy_atoms` at the "generic" level if we need to support AHFE. I think having something like this being re-implemented later on specifically for RBFE/RHFE code-paths would make a lot of sense

tests/test_dummy.py
```diff
@@ -92,4 +96,61 @@ def test_generate_dummy_group_assignments():
...
+    assert anchored_dummy_group_assignments[0] == {1: (2, frozenset({0}))}
+
+    # reverse direction
+    anchored_dummy_group_assignments = generate_anchored_dummy_group_assignments(expected_dga, g_b, g_a, core_b, core_a)
+
+    anchored_dummy_group_assignments = list(anchored_dummy_group_assignments)
+    assert len(anchored_dummy_group_assignments) == 1
+    assert anchored_dummy_group_assignments[0] == {1: (2, frozenset({0}))}
+
+    # Test that providing an empty core results in None values for both the bond anchor
```
mcwitt, line 148: Core is not empty here?

mcwitt:
> This PR adds support for handling dummy groups that are completely unanchored (i.e. devoid of bond anchors).

Is the motiv

timemachine/fe/dummy.py
```diff
@@ -138,11 +148,16 @@ def generate_anchored_dummy_group_assignments(
...
     """
 
+    assert len(core_atoms_a) == len(core_atoms_b)
+
     core_bonds_c = get_core_bonds(bond_graph_a.edges(), bond_graph_b.edges(), core_atoms_a, core_atoms_b)
     c_to_b = {c: b for c, b in enumerate(core_atoms_b)}
     core_bonds_b = frozenset(translate_bonds(core_bonds_c, c_to_b))
 
     def get_angle_anchors(bond_anchor):
+        if bond_anchor is None:
```
mcwitt, line 158: Should the type of the `dummy_groups` argument to this function be updated to `dict[Optional[int], frozenset[int]]`?

Will the restriction of this representation to only one unanchored dummy group be an issue in practice?

timemachine/fe/dummy.py
```diff
@@ -7,14 +7,18 @@
...
+    pass
+
+
+class ZeroBondAnchorWarning(UserWarning):
     pass
 
 
 def generate_dummy_group_assignments(
     bond_graph: nx.Graph, core_atoms: Collection[int]
 ) -> Iterator[dict[int, frozenset[int]]]:
```
mcwitt, line 20: Should this be updated to 
```suggestion
) -> Iterator[dict[Optional[int], frozenset[int]]]:
```
?

Is there an issue if there are multiple unanchored dummy groups? The current behavior seems to be to keep only the last one(?)

tests/test_dummy.py
proteneer: fixed - meant to say:


```
# Test that providing an empty core and a None bond anchor results in a None angle anchor
anchored_dummy_group_assignments = generate_anchored_dummy_group_assignments(
    {None: {0, 1, 2, 3, 4}}, g_a, g_b, [], []
)
```

timemachine/fe/dummy.py
proteneer: > Will the restriction of this representation to only one unanchored dummy group be an issue in practice?

1) Multiple disconnected dummy groups (and their atoms) will all be assigned to the "None" dummy group. 

``` python
def test_multiple_none_dummy_groups():
    # Test that if we break a core-core bond, we only have one valid
    # choice of the angle anchor
    #
    # mol_a: H-O-H.H-O-H
    # mol_b: H-O-H.H-O-H
    g_a = convert_to_nx(Chem.AddHs(Chem.MolFromSmiles("O.O")))
    core_a = []

    dgas = list(generate_dummy_group_assignments(g_a, core_a))
    expected_dga = {None: {0,1,2,3,4,5}}

    assert equivalent_assignment(dgas, [expected_dga])
```

2) As a result, their dummy angle anchors will also be assigned to the "None" angle anchor:

``` python
def test_multiple_none_dummy_groups():
    # Test that if we break a core-core bond, we only have one valid
    # choice of the angle anchor
    #
    # mol_a: H-O-H.H-O-H
    # mol_b: H-O-H.H-O-H
    g_a = convert_to_nx(Chem.AddHs(Chem.MolFromSmiles("O.O")))
    g_b = convert_to_nx(Chem.AddHs(Chem.MolFromSmiles("O.O")))
    core_a = []
    core_b = []

    # Test that providing an empty core and a None bond anchor results in a None angle anchor
    anchored_dummy_group_assignments = generate_anchored_dummy_group_assignments(
        {None: {0, 1, 2, 3, 4, 5}}, g_a, g_b, core_a, core_b
    )

    anchored_dummy_group_assignments = list(anchored_dummy_group_assignments)

    assert anchored_dummy_group_assignments == [{None: (None, frozenset({0, 1, 2, 3, 4, 5}))}]
```

timemachine/fe/dummy.py
proteneer: (See above comment for intended behavior)

timemachine/fe/dummy.py
proteneer: Added a test in https://github.com/proteneer/timemachine/pull/1510/commits/ade39b727757556b2f6c3ddddd6ac9ed47d873de

mcwitt: > Multiple disconnected dummy groups (and their atoms) will all be assigned to the "None" dummy group.

Got it. I wanted to be sure that the apparent behavior of unioning the disconnected dummy groups into a single group anchored by `None` was intentional (versus using a different representation that preserves information about them being disconnected; I don't think I'm totally clear on the use case for this, so this may not be a useful comment)

timemachine/fe/dummy.py
mcwitt: I think the return type should still be updated, though? (To reflect that the anchor can be `None`; I'm not sure why mypy doesn't complain about this)

timemachine/fe/dummy.py
mcwitt: > Should the type of the dummy_groups argument to this function be updated to dict[Optional[int], frozenset[int]]?

IIUC, both the type of the `dummy_groups` argument and the return type of `generate_anchored_dummy_group_assignments` should be updated to indicate that the bond anchor might be `None`.

timemachine/fe/dummy.py
```diff
@@ -64,18 +68,26 @@ def generate_dummy_group_assignments(
        latter might be desirable for efficiency but is more complicated to implement).
     """
     assert len(set(core_atoms)) == len(core_atoms)
-    assert len(list(nx.connected_components(bond_graph))) == 1
+
+    if assert_single_connected_component:
+        assert len(list(nx.connected_components(bond_graph))) == 1
```
mcwitt, line 73: nit/optional: noting that this change to an optional assertion reduces the power of static checking to catch errors somewhat, because now the bond anchor is optional in the return type even if we pass `assert_single_connected_component=True` (and thus downstream code not intended to handle empty cores or disconnected graphs must assert `bond_anchor is not None`).

One way around this could be to instead define a variant `generate_dummy_group_assignments_connected` that calls this function with `assert_single_connected_component=True` and returns the old type (i.e. asserting in the types that the dummy anchor is not null). Another option could be to use `typing.overload`.

(Non-blocking, OK to clean this up later.)

proteneer: This is temporary - it will be none optional (i.e. the assertion will default to False after the refactoring is done).

timemachine/fe/dummy.py
proteneer: Updated type signatures in https://github.com/proteneer/timemachine/pull/1510/commits/727b1516bd6e25c249bdd2859ebaee884f637aff

timemachine/fe/dummy.py
proteneer: Yep, updated in https://github.com/proteneer/timemachine/pull/1510/commits/727b1516bd6e25c249bdd2859ebaee884f637aff

timemachine/fe/dummy.py
mcwitt: (Looks like the type of `find_dummy_groups_and_anchors` also needs to be updated)

mcwitt: Might be because the pre-commit git hook only runs on files in the diff by default, while in CI we run with the `--all-files` flag

mcwitt (approved):

PR #1511, opened 2025-02-24
Clean up test data

jkausrelay:
- Move all test data to testsystems/data
- Also move the code from timemachine/datasets to testsystems


jkausrelay:
Close in favor of #1518

PR #1512, opened 2025-02-24
[WIP] Modify protocol optimization for REST compatibility

maxentile:
REST (https://github.com/proteneer/timemachine/pull/1424) assumes inclusion of an easy-to-sample state @ lam=0.5, but the protocol optimization stages of bisection ( https://github.com/proteneer/timemachine/pull/959 ) and rebalancing ( https://github.com/proteneer/timemachine/pull/1387, https://github.com/proteneer/timemachine/pull/1426 ) can possibly bypass lam=0.5.

This PR introduces some logic to ensure that a state @ lam=0.5 is included.

* Bisection: Instead of initializing with the list `[0, 1]`, initialize with `[0, 0.5, 1]`. (With a slight complication to handle `(lambda_min, lambda_max)` not necessarily `(0, 1)`, etc.)
* Rebalancing: Instead of optimizing once over the interval `(0, 1)`, optimize twice over the intervals `(0, 0.5)`, `(0.5, 1)`. (Again handling general `(lambda_min, lambda_max)` .)


maxentile:
- [ ] This doesn't consistently handle the `n_windows` argument yet (`run_vacuum(..., n_windows=2)` can produce a result with 3 windows, for example)
- [ ] this might need some adjustment to keep track of whether the scaling factor in RESTParams is > 1.0 or not. (Currently, this PR introduces a lam=0.5 state even if REST scaling factor == 1.0.)

mcwitt:
> This doesn't consistently handle the n_windows argument yet (run_vacuum(..., n_windows=2) can produce a result with 3 windows, for example)

I think I'd be in favor of requiring `n_windows >= 3` assuming it significantly simplifies the implementation.

> this might need some adjustment to keep track of whether the scaling factor in RESTParams is > 1.0 or not. (Currently, this PR introduces a lam=0.5 state even if REST scaling factor == 1.0.)

This doesn't seem like a major issue in practice to me. It seems consistent with requiring `n_windows >= 3` to also guarantee 3 or more windows when `n_windows` is unspecified.

PR #1517, opened 2025-02-26
Support all python 3.12 versions

badisa:
* Oversight in https://github.com/proteneer/timemachine/pull/1454 and realized couldn't install timemachine with Python 3.12.9
* Require >=3.12 python for building C++

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -16,6 +16,7 @@
 Dockerfile
 .dockerignore
 .gitlab-ci.yml
+pytest-artifacts/
 
 # Git Files
 **/.git*
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ dist/
 *.egg-info/
 .coverage*
 .hypothesis/
+pytest-artifacts/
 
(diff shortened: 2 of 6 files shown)


timemachine/cpp/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 project(timemachine LANGUAGES CXX CUDA)
 
-find_package(Python 3.10...3.12.8 COMPONENTS Interpreter Development REQUIRED)
+find_package(Python 3.10...<3.13 COMPONENTS Interpreter Development REQUIRED)
```
mcwitt, line 8: Should this be consistent with https://github.com/proteneer/timemachine/blob/94ee2190911803c42954e0c1277854d129a265b7/setup.py#L94?

I'd be in favor of requiring >=3.12 consistently with what's currently in `setup.py` so that we can adopt some of the newer syntax features.

mcwitt (approved):

PR #1522, opened 2025-03-13
Fix Water Sampling When REST Is Enabled

badisa:
The water sampler needs to produce the same nonbonded interaction energies as used in MD to ensure correct sampling. Because of the Ligand-Env split with the `NonbondedInteractionGroup`, we have to merge the ligand parameters of the `NonbondedInteractionGroup` and the host environment of the `NonbondedAllPair` to get the correct parameters. The `NonbondedAllPair` contains all of the host parameters, but does **not** contain ligand parameters.

Prior to REST we would merge the protein parameters from the NonbondedAllPair potential into the parameters of `NonbondedInteractionGroup`. When REST was introduced we chose to modify both the ligand and environment parameters, which resulted in the `TIBDExchangeMover` softening all of the waters (but not protein) which produced an inconsistency between the water sampling and MD. This PR resolves this inconsistency and adds assertions to avoid the issue in the future.

## Changes
* Take all of the environment parameters from `NonbondedAllPair` and only use `NonbondedInteractionGroup` for the ligand parameters
* REST no longer modifies the environment parameters, only modifying the ligand parameters to provide a canonical source of parameters. There is some debate about this, and in the future we may want to adopt a different approach in the future.

### DIFF
--- a/timemachine/fe/free_energy.py
+++ b/timemachine/fe/free_energy.py
@@ -580,7 +580,7 @@ def get_water_sampler_params(initial_state: InitialState) -> NDArray:
     changes are expected to be in P0 to ensure correct water sampling.
     """
     nb_ixn_pot = get_bound_potential_by_type(initial_state.potentials, NonbondedInteractionGroup)
-    water_params = np.array(nb_ixn_pot.params)
+    water_sampler_params = np.array(nb_ixn_pot.params)
 
     # If the system contains a host, use the parameters of the Nonbonded potential for the water sampler the
     # NonbondedInteractionGroup host parameters may be modified for protein-ligand charge fitting
@@ -588,12 +588,12 @@ def get_water_sampler_params(initial_state: InitialState) -> NDArray:
         host_idxs = np.delete(np.arange(initial_state.x0.shape[0]), initial_state.ligand_idxs)
         water_idxs = np.delete(host_idxs, initial_state.protein_idxs)
         nb_all_pairs_params = get_bound_potential_by_type(initial_state.potentials, Nonbonded).params
-        assert (nb_all_pairs_params[water_idxs] == water_params[water_idxs]).all()
+        assert (nb_all_pairs_params[water_idxs] == water_sampler_params[water_idxs]).all()
         host_params = nb_all_pairs_params[host_idxs]
-        water_params[host_idxs] = host_params
+        water_sampler_params[host_idxs] = host_params
 
-    assert water_params.shape[1] == 4
-    return water_params
+    assert water_sampler_params.shape[1] == 4
+    return water_sampler_params
 
 
 def get_context(initial_state: InitialState, md_params: Optional[MDParams] = None) -> Context:


timemachine/fe/rest/single_topology.py
```diff
@@ -147,24 +146,24 @@ def combine_with_host(
...
             params=jnp.asarray(ref_state.nonbonded_ixn_group.params)
-            .at[:, NBParamIdx.Q_IDX]
-            .mul(sqrt_energy_scale)  # scale ligand charges
-            .at[:, NBParamIdx.LJ_EPS_IDX]
-            .mul(sqrt_energy_scale),  # scale ligand epsilons
+            .at[num_host_atoms:, NBParamIdx.Q_IDX]
+            .mul(energy_scale)  # scale ligand charges
+            .at[num_host_atoms:, NBParamIdx.LJ_EPS_IDX]
+            .mul(energy_scale),  # scale ligand epsilons
         )
```
maxentile, line 167: Mistaken comment: although these 3 approaches define equivalent interaction-group potentials, they are not functionally equivalent in the context of how the water sampler infers parameters from the interaction-group potentials.

And for the record, I think @mcwitt had initially used implementation option (1), prior to my review comment here: https://github.com/proteneer/timemachine/pull/1424#discussion_r1887738947

timemachine/fe/rest/single_topology.py
```diff
@@ -147,24 +146,26 @@ def combine_with_host(
...
+        energy_scale = self.get_energy_scale_factor(lamb)
 
         nonbonded_host_guest_ixn = replace(
             ref_state.nonbonded_ixn_group,
             params=jnp.asarray(ref_state.nonbonded_ixn_group.params)
-            .at[:, NBParamIdx.Q_IDX]
-            .mul(sqrt_energy_scale)  # scale ligand charges
-            .at[:, NBParamIdx.LJ_EPS_IDX]
-            .mul(sqrt_energy_scale),  # scale ligand epsilons
+            .at[num_host_atoms:, NBParamIdx.Q_IDX]
```
mcwitt, line 165: nit: might help with readability to define and reuse an index expression: `ligand_idxs = np.index_exp[num_host_atoms:]`

timemachine/fe/free_energy.py
```diff
@@ -554,12 +554,12 @@ def get_water_sampler_params(initial_state: InitialState) -> NDArray:
...
     water_params = np.array(nb_ixn_pot.params)
 
-    # If the protein is present, use the original protein parameters for the water sampler
-    if len(initial_state.protein_idxs):
-        prot_params = get_bound_potential_by_type(initial_state.potentials, Nonbonded).params[
-            initial_state.protein_idxs
-        ]
-        water_params[initial_state.protein_idxs] = prot_params
+    # If the system contains a host, use the parameters of the Nonbonded potential for the water sampler
+    # the NonbondedInteractionGroup parameters may be modified for enhanced sampling purposes
```
mcwitt, line 558: Might have misunderstood, but I thought we were planning to assert that water parameters are identical in the all pairs and interaction group potentials, so this seems redundant?

tests/test_free_energy.py
```diff
@@ -430,16 +432,24 @@ def test_get_water_sampler_params(num_windows):
...
+        nb_pot = bound_nb_pot.potential
+        system = st.combine_with_host(
+            solvent_host.system, lamb, solvent_host.num_water_atoms, st.ff, solvent_host.omm_topology
+        )
+        ligand_water_params = system.nonbonded_ixn_group.params[host_config.conf.shape[0] :]
         if lamb == 0.0:
-            assert np.all(ligand_water_params[mol_a_only_atoms][:, 3] == 0.0)
-            assert np.all(ligand_water_params[mol_b_only_atoms][:, 3] == nb_pot.cutoff)
+            assert np.all(ligand_water_params[mol_a_only_atoms][:, NBParamIdx.W_IDX] == 0.0)
+            print
```
proteneer, line 443: stray print

timemachine/fe/free_energy.py
```diff
@@ -550,15 +550,45 @@ def _combine(self, ligand_values: NDArray, host_values: Optional[NDArray] = None
...
+    Let W0, P0, L0 be the parameters from NonbondedAllPairs, W1, P1, L1 from the NonbondedInteractionGroup
+
+    We expect the parameters used for the water sampler to meet the following requirements:
+
+    1) Waters Parameters: assert that W0 == W1, we use NonbondedAllPairs but could use either
+    2) Protein Parameters: read protein parameters (P0) from NonbondedAllPairs
+    3) Ligand Parameters: read ligand parameters (L1) from NonbondedInteractionGroup
+
+    Important to note that P1 can change in Protein-Ligand charge fitting, which is okay because it is only expected
+    to impact Protein-Ligand interactions and not Protein-Water.
```
proteneer, line 579: Also note that P0 can possibly change from REST on side-chains?

timemachine/fe/free_energy.py
```diff
@@ -550,16 +550,47 @@ def _combine(self, ligand_values: NDArray, host_values: Optional[NDArray] = None
...
+    1) Waters Parameters: assert that W0 == W1, we use NonbondedAllPairs but could use either
+    2) Protein Parameters: read protein parameters (P0) from NonbondedAllPairs
+    3) Ligand Parameters: read ligand parameters (L1) from NonbondedInteractionGroup
+
+    Important to note that P1 can change in Protein-Ligand charge fitting, which is okay because it is only expected
+    to impact Protein-Ligand interactions and not Protein-Water. If protein side chains are to be enhanced, the parameter
+    changes are expected to be in P0 to ensure correct water sampling.
     """
     nb_ixn_pot = get_bound_potential_by_type(initial_state.potentials, NonbondedInteractionGroup)
     water_params = np.array(nb_ixn_pot.params)
```
mcwitt, line 583: nit: naming seems confusing since these are not all water parameters; should this be `nb_ixn_group_params` or similar?

timemachine/fe/free_energy.py
```diff
@@ -1243,6 +1274,12 @@ def assert_ensembles_compatible(state_a: InitialState, state_b: InitialState):
         # also, assert barostat and integrator are self-consistent
         assert intg_a.temperature == baro_a.temperature
 
+        # The protein and water parameters should match exactly for the water sampling parameters
```
mcwitt, line 1277: Is this necessary for correctness, or is the assertion that the water parameters are consistent in the all pairs and interaction group potentials sufficient?

proteneer (approved):

maxentile (approved):

PR #1523, opened 2025-03-14
Speed-up setup_intermediate_state by 30x

proteneer:
This PR speeds-up `setup_intermediate_state()` considerably when called multiple times, bringing 128 calls from 21s -> 0.6s.

Main benefits come from

1) [5x speed-up] - switching from `jax.numpy` to `numpy`
2) [2x speed-up] - pre-align parameters in the constructor.
3) [1.5x speed-up] - caching `get_dummy_atoms_a`, `get_dummy_atoms_b`, `get_core_atoms` 
4) [1.5x speed-up] - caching `src_chiral_idxs` and `dst_chiral_idxs`

Main downside, `setup_intermediate_state` is no longer `jittable` nor `Traceable`. I think our previous motivation for this was mainly for speed anyways. 

`cProfile` now shows:

```
      640    0.017    0.000    0.341    0.001 /home/yzhao/Code/timemachine/timemachine/fe/single_topology.py:1587(_align_and_interpolate_bonded_term)
    53760    0.102    0.000    0.147    0.000 /home/yzhao/Code/timemachine/timemachine/fe/interpolate.py:174(pad)
    14848    0.008    0.000    0.142    0.000 /home/yzhao/Code/timemachine/timemachine/fe/single_topology.py:1570(_interpolate_torsion)
      128    0.000    0.000    0.127    0.001 /home/yzhao/Code/timemachine/timemachine/fe/single_topology.py:1651(align_and_interpolate_angles)
    14848    0.013    0.000    0.125    0.000 /home/yzhao/Code/timemachine/timemachine/fe/single_topology.py:853(interpolate_periodic_torsion_params)
     7424    0.007    0.000    0.119    0.000 /home/yzhao/Code/timemachine/timemachine/fe/single_topology.py:1510(_interpolate_angle)
      128    0.000    0.000    0.109    0.001 /home/yzhao/Code/timemachine/timemachine/fe/single_topology.py:1670(align_and_interpolate_propers)
     7424    0.014    0.000    0.092    0.000 /home/yzhao/Code/timemachine/timemachine/fe/single_topology.py:805(interpolate_harmonic_angle_params)
    22272    0.046    0.000    0.068    0.000 /home/yzhao/Code/timemachine/timemachine/fe/single_topology.py:787(cyclic_difference)
      128    0.000    0.000    0.052    0.000 /home/yzhao/Code/timemachine/timemachine/fe/single_topology.py:1660(align_and_interpolate_bonds)
[...]

PR #1524, opened 2025-03-17
Refine selection of atoms in REST region using simple heuristic

mcwitt:
The initial implementation of REST in single topology transformations in #1424 applies temperature scaling to all ligand-environment nonbonded interactions, and all ligand-ligand proper torsions that are rotatable or in aliphatic rings. I.e., the "REST region" (here, the set of ligand atoms whose interactions may potentially be scaled), is implicitly _all_ of the combined ligand atoms.

This PR refines the REST region using a simple heuristic: a combined ligand atom in included iff it is involved in a bond, angle, or improper torsion interaction whose parameters differ between the end states of the transformation. The logic for scaling the relevant interactions is retained from the previous version, except that we now exclude interactions from scaling that do not involve an atom in the REST region.

This is intentionally a very minimal heuristic. Later on, we might extend this for example to include in the REST region ligand atoms that are not directly mutated in the transformation, but are nearby (e.g. by graph distance) to a mutation.

This PR also adds a utility to visualize the REST region using `py3Dmol` (similar to the existing `view_atom_mapping_3d`). Example from the hif2a public benchmark:

TODO:
- [x] add tests exercising heuristic
- [x] rebase once #1523 is merged

### DIFF
--- a/tests/rest/test_single_topology_rest.py
+++ b/tests/rest/test_single_topology_rest.py
@@ -123,11 +123,19 @@ def test_single_topology_rest_vacuum(mol_pair, temperature_scale_interpolation_f
             )
             return proper(ligand_conf, None)
 
-        U_proper_subset = compute_proper_energy(state, st_rest.target_proper_idxs)
-        U_proper_subset_ref = compute_proper_energy(state_ref, st_rest.target_proper_idxs)
-        np.testing.assert_allclose(U_proper_subset, energy_scale * U_proper_subset_ref)
+        # check that propers in the REST region are scaled appropriately
+        rest_proper_idxs = st_rest.target_proper_idxs
+        U_proper_rest = compute_proper_energy(state, rest_proper_idxs)
+        U_proper_rest_ref = compute_proper_energy(state_ref, rest_proper_idxs)
+        np.testing.assert_allclose(U_proper_rest, energy_scale * U_proper_rest_ref)
 
-        np.testing.assert_allclose(U_nonbonded, energy_scale * U_nonbonded_ref)
+        # check that propers outside of the REST region are not scaled
+        num_propers = len(st_rest.propers)
+        rest_complement_proper_idxs = set(range(num_propers)) - set(rest_proper_idxs)
+        rest_complement_proper_idxs = list(rest_complement_proper_idxs)
+        U_proper_complement = compute_proper_energy(state, rest_complement_proper_idxs)
+        U_proper_complement_ref = compute_proper_energy(state_ref, rest_complement_proper_idxs)
+        np.testing.assert_allclose(U_proper_complement, U_proper_complement_ref)
 
 
 @cache
@@ -165,11 +173,17 @@ def test_single_topology_rest_solvent(mol_pair, temperature_scale_interpolation_
         )
         return nonbonded_ixn_group(conf, host_config.box)
 
+    # check that interactions involving atoms in the REST region are scaled appropriately
     U = compute_host_guest_ixn_energy(st_rest, st_rest.rest_region_atom_idxs)
     U_ref = compute_host_guest_ixn_energy(st, st_rest.rest_region_atom_idxs)
-
     energy_scale = st_rest.get_energy_scale_factor(lamb)
     np.testing.assert_allclose(U, energy_scale * U_ref, rtol=1e-5)
+
+    # check that interactions involving atoms outside of the REST region are not scaled
+    rest_complement_atom_idxs = set(range(st_rest.get_num_atoms())) - st_rest.rest_region_atom_idxs
+    U_complement = compute_host_guest_ixn_energy(st_rest, rest_complement_atom_idxs)
+    U_complement_ref = compute_host_guest_ixn_energy(st, rest_complement_atom_idxs)
+    np.testing.assert_allclose(U_complement, U_complement_ref, rtol=1e-5)
 
 
 def get_mol(smiles: str):
--- a/tests/rest/test_single_topology_rest.py
+++ b/tests/rest/test_single_topology_rest.py
@@ -1,3 +1,4 @@
+from collections.abc import Sequence
 from dataclasses import replace
 from functools import cache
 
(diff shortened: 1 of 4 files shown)


mcwitt:
Closing temporarily. The REST region inference needs more thought; the current approach incorrectly includes all ligand atoms in the REST region if there is a mutation of any ligand atom due to the ligand-ligand pair interaction parameters changing.

timemachine/fe/rest/single_topology.py
```diff
@@ -79,37 +80,71 @@ def __init__(
             max_temperature_scale, temperature_scale_interpolation
         )
 
+    @cached_property
+    def rest_region_atom_idxs(self) -> set[int]:
+        """Returns the set of indices of atoms in the combined ligand that are in the REST region."""
+        return {
+            idx
+            for diffs in [
+                self.aligned_bond_tuples,
```
proteneer, line 89: What if we diff'd on just the idxs?

mcwitt:
> Closing temporarily. The REST region inference needs more thought; the current approach incorrectly includes all ligand atoms in the REST region if there is a mutation of any ligand atom due to the ligand-ligand pair interaction parameters changing.

Addressed this by moving to a heuristic only considering differences in bond, angle, and improper torsion valence parameters (dummy atoms are always included in the REST region). This heuristic seems to produce reasonable-looking REST regions in a few examples, but needs further validation.

timemachine/fe/rest/single_topology.py
```diff
@@ -79,37 +81,89 @@ def __init__(
...
     def propers(self) -> list[CanonicalProper]:
+        """Returns a list of proper torsions in the combined ligand."""
         # TODO: refactor SingleTopology to compute src and dst alignment at initialization
         return [mkproper(*idxs) for idxs in super().setup_intermediate_state(0.0).proper.potential.idxs]
 
     @cached_property
-    def target_proper_idxs(self) -> list[int]:
-        return [
-            idx
+    def candidate_propers(self) -> set[tuple[int, CanonicalProper]]:
```
mcwitt, line 143: Changed the naming slightly so that `candidate_propers` refer to interactions that could potentially be softened if they involve an atom in the REST region, while `target_propers` refer to `candidate_propers` that will be softened because they involve an atom in the REST region. The previous behavior was effectively set `target_propers = candidate_propers`.

tests/rest/test_single_topology_rest.py
```diff
@@ -71,9 +78,11 @@ def test_single_topology_rest_vacuum(mol_pair, temperature_scale_interpolation_f
     st = get_single_topology(mol_a, mol_b, core)
     st_rest = get_single_topology_rest(mol_a, mol_b, core, 2.0, temperature_scale_interpolation_fxn)
 
+    assert 0 < len(st_rest.rest_region_atom_idxs) < st_rest.get_num_atoms()
```
proteneer, line 81: does this actually work? i.e. does it evaluate to:

(0 < len(st_rest.rest_region_atom_idxs) == {True,False}) < st_rest.get_atoms()?

tests/rest/test_single_topology_rest.py
```diff
@@ -99,20 +108,22 @@ def test_single_topology_rest_vacuum(mol_pair, temperature_scale_interpolation_f
...
 
         if has_rotatable_bonds or has_aliphatic_rings:
-            assert 0 < len(st_rest.target_propers)
+            assert 0 < len(st_rest.candidate_propers)
 
             if energy_scale < 1.0:
                 assert U_proper < U_proper_ref
 
-        def get_proper_subset_energy(state: GuestSystem, ixn_idxs):
+        def compute_proper_energy(state: GuestSystem, ixn_idxs):
```
proteneer, line 116: typing on `ixn_idxs: list/NDArray[int]`?

tests/rest/test_single_topology_rest.py
mcwitt: Admittedly seems a bit cursed but it does work as you say: https://docs.python.org/3/reference/expressions.html#comparisons

> Comparisons can be chained arbitrarily, e.g., x < y <= z is equivalent to x < y and y <= z, except that y is evaluated only once

tests/rest/test_single_topology_rest.py
mcwitt: Added in https://github.com/proteneer/timemachine/pull/1524/commits/97814ada93445aceb8d4fef4b51aa5f6951e5274

tests/rest/test_single_topology_rest.py
```diff
@@ -143,12 +154,18 @@ def test_single_topology_rest_solvent(mol_pair, temperature_scale_interpolation_
...
+        nonbonded_ixn_group = replace(
+            hgs.nonbonded_ixn_group,
+            potential=replace(hgs.nonbonded_ixn_group.potential, row_atom_idxs=ligand_idxs_),
+        )
+        return nonbonded_ixn_group(conf, host_config.box)
 
-    U = get_nonbonded_host_guest_ixn_energy(st_rest)
-    U_ref = get_nonbonded_host_guest_ixn_energy(st)
+    U = compute_host_guest_ixn_energy(st_rest, st_rest.rest_region_atom_idxs)
+    U_ref = compute_host_guest_ixn_energy(st, st_rest.rest_region_atom_idxs)
```
proteneer, line 169: is it also worth sanity checking that ligand atoms outside of the rest region are *not* scaled by energy_scale? (in case of accidental overflows)

maxentile (approved):

tests/rest/test_single_topology_rest.py
mcwitt: Good call. Added in https://github.com/proteneer/timemachine/pull/1524/commits/39f85080b9818994be43dba89c97b8f24bbd85f3

tests/rest/test_single_topology_rest.py
```diff
@@ -71,9 +79,11 @@ def test_single_topology_rest_vacuum(mol_pair, temperature_scale_interpolation_f
     st = get_single_topology(mol_a, mol_b, core)
     st_rest = get_single_topology_rest(mol_a, mol_b, core, 2.0, temperature_scale_interpolation_fxn)
 
+    assert 0 < len(st_rest.rest_region_atom_idxs) < st_rest.get_num_atoms()
```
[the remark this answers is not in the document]
mcwitt: I was on the fence about this. Only the weaker form you suggest is generally true, but the stronger version is true for the hif2a edges under test. I think you're right that it makes more sense to weaken to the generally-true property to avoid confusion.

proteneer (approved):
awesome!

PR #1525, opened 2025-03-18
Speed-up setup_intermediate_state while maintaining trace-ability.

proteneer:
This PR speeds-up `setup_intermediate_state` by about 30-50x while still maintaining our calling to call `jax.jit`, `jax.vmap`, and `jax.grad` on functions that invoke `setup_intermediate_state()`

[2x speed-up] - pre-align idxs and parameters in the constructor.
[2x speed-up] - pre-compute min/max bounds used by interpolation code in the constructor.
[2x speed-up] - vectorize `interpolate_*_params` with `jax.vmap` and `jax.jit`
[1.5x speed-up] - caching `get_dummy_atoms_a, get_dummy_atoms_b, get_core_atoms`.
[1.5x speed-up] - caching `src_chiral_idxs` and `dst_chiral_idxs`.

timings for 128 calls to `setup_intermediate_state()` on the hif2a ligand pair:

master:

jit time: 0.63s
run time: 20.91s

this pr:

jit time: 0.41s
run time: 0.56s

``` python
import time

def test_setup_intermediate_state():
    mol_a, mol_b, core = get_hif2a_ligand_pair_single_topology()
    st = single_topology.SingleTopology(mol_a, mol_b, core, ff)

    import cProfile
    import pstats

    # call once to warm-up jit
    start_time = time.time()
    st.setup_intermediate_state(0.5)
    print("jit time", time.time() - start_time)
    profiler = cProfile.Profile()
    profiler.enable()
    start_time = time.time()
    for lam in np.linspace(0, 1, 128):
        st.setup_intermediate_state(lam)
    print("run time", time.time() - start_time)
    profiler.disable()
    pstats.Stats(profiler).sort_stats(pstats.SortKey.CUMULATIVE).print_stats(100)


if __name__ == "__main__":
    test_setup_intermediate_state()
```

### DIFF
--- a/timemachine/fe/interpolate.py
+++ b/timemachine/fe/interpolate.py
@@ -14,6 +14,11 @@ Params = Any
 Key = Any
 
 
+# used to convert arrays to a hashable type for use as dict keys and in sets
+def to_hashable(x):
+    return tuple(to_hashable(e) for e in x) if isinstance(x, Iterable) else x
+
+
 def align_idxs_and_params(
     src_idxs,
     src_params,
@@ -78,10 +83,6 @@ def align_idxs_and_params(
     for all_idxs in [src_idxs, dst_idxs]:
         for idxs in all_idxs:
             validate_idxs(idxs)
-
-    # used to convert arrays to a hashable type for use as dict keys and in sets
-    def to_hashable(x):
-        return tuple(to_hashable(e) for e in x) if isinstance(x, Iterable) else x
 
     def make_kv(all_idxs, all_params):
         kvs = [(to_hashable(key(idxs, params)), params) for idxs, params in zip(all_idxs, all_params)]
--- a/timemachine/fe/single_topology.py
+++ b/timemachine/fe/single_topology.py
@@ -1,7 +1,8 @@
 import warnings
 from collections.abc import Collection, Sequence
+from dataclasses import dataclass
 from enum import IntEnum
-from functools import partial
+from functools import cache, cached_property, partial
 from typing import Any, Optional
 
 import jax
@@ -62,6 +63,8 @@ def _flip_min_max(min_max):
 # however, this led to roughly a 10-15% increase in the # of windows. So instead, we can switch to a softer scheme by setting
 # the bond interpolation bounds to [0.0, 0.7]. This way, when lambda=0.3, the start of the chiral volume interpolation, the bonds
 # are still present with a force constant of ~30kJ/mol (still strong enough to keep the chiral volumes numerically stable)
+
+DEFAULT_MIN_MAX = [0.0, 1.0]
 
 # core-bonds are never turned off if enforce_core_core=True
 CORE_BOND_MIN_MAX = [0.0, 1.0]
@@ -769,7 +772,7 @@ def interpolate_harmonic_bond_params(src_params, dst_params, lamb, k_min, lambda
     k = pad(log_linear_fn, src_k, dst_k, lamb, lambda_min, lambda_max)
     x = pad(interpolate.linear_interpolation, src_x, dst_x, lamb, lambda_min, lambda_max)
 
-    return jnp.array([k, x])
+    return [k, x]
 
 
 def interpolate_chiral_volume_params(src_params, dst_params, lamb, k_min, lambda_min, lambda_max):
@@ -779,7 +782,7 @@ def interpolate_chiral_volume_params(src_params, dst_params, lamb, k_min, lambda
     log_linear_fn = partial(interpolate.log_linear_interpolation, min_value=k_min)
     k = pad(log_linear_fn, src_k, dst_k, lamb, lambda_min, lambda_max)
 
-    return jnp.array(k)
+    return [k]
 
 
 def cyclic_difference(a, b, period):
@@ -791,7 +794,6 @@ def cyclic_difference(a, b, period):
 
     with minimum abs(x).
     """
-
     d = jnp.fmod(b - a, period)
 
     def f(d):
@@ -845,7 +847,7 @@ def interpolate_harmonic_angle_params(src_params, dst_params, lamb, k_min, lambd
     # the end states with neighboring intermediates.
     eps = jnp.where((lamb == 0.0) | (lamb == 1.0), 0.0, 1e-3)
 
-    return jnp.array([k, phase, eps])
+    return [k, phase, eps]
 
 
 def interpolate_periodic_torsion_params(src_params, dst_params, lamb, lambda_min, lambda_max):
@@ -887,7 +889,7 @@ def interpolate_periodic_torsion_params(src_params, dst_params, lamb, lambda_min
     dst_phase = src_phase + cyclic_difference(src_phase, dst_phase, period=2 * np.pi)
     phase = pad(interpolate.linear_interpolation, src_phase, dst_phase, lamb, lambda_min, lambda_max)
 
-    return jnp.array([k, phase, src_period])
+    return [k, phase, src_period]
 
 
 def interpolate_w_coord(w0: float | jax.Array, w1: float | jax.Array, lamb: float):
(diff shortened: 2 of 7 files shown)


timemachine/fe/interpolate.py
```diff
@@ -14,6 +14,10 @@ class DuplicateAlignmentKeysError(RuntimeError):
 Key = Any
 
 
+def to_hashable(x):
```
proteneer, line 17: this was moved out so it shows up in profiling.

timemachine/fe/single_topology.py
```diff
@@ -910,12 +911,69 @@ def interpolate_w_coord(w0: float | jax.Array, w1: float | jax.Array, lamb: floa
     )
 
 
+batch_interpolate_harmonic_bond_params = jax.jit(
```
proteneer, line 914: debating about putting these functions and the AlignedPotential classes into interpolate.py

tests/test_plotting.py
```diff
@@ -102,3 +102,32 @@ def test_plot_interpolation_schedule():
     plot_core_interpolation_schedule(st)
     plot_dummy_a_interpolation_schedule(st)
     plot_dummy_b_interpolation_schedule(st)
+
+
+import time
```
proteneer, line 107: (will remove this blob right before merging)

timemachine/fe/single_topology.py
```diff
@@ -910,12 +912,68 @@ def interpolate_w_coord(w0: float | jax.Array, w1: float | jax.Array, lamb: floa
...
+
+
+@jax.jit
+def batch_interpolate_nonbonded_params(
+    cutoff,
+    src_params,
+    dst_params,
+    lamb: float,
+):
+    src_qlj, src_w = src_params[:, : NBParamIdx.W_IDX], src_params[:, NBParamIdx.W_IDX]
```
jkausrelay, line 936: add docs

timemachine/fe/single_topology.py
jkausrelay: I think it'd help w/organization considering the size of single_topology

timemachine/fe/interpolate.py
mcwitt: Aside: I think we might be able to just delete this in a future PR; IIRC it exists to deal with some inconsistencies in the types and number of dimensions of the index tuples for different potentials.

timemachine/fe/single_topology.py
```diff
@@ -972,12 +1027,15 @@ def __init__(self, mol_a, mol_b, core: NDArray):
         self.c_to_a = {int(v): k for k, v in enumerate(self.a_to_c)}
         self.c_to_b = {int(v): k for k, v in enumerate(self.b_to_c)}
 
+    @cache
```
mcwitt, line 1030: nit: it might be worth making `AtomMapMixin` immutable (e.g. dataclass with `frozen=True`); otherwise, the cached values could become inconsistent with the instance attributes.

timemachine/fe/single_topology.py
```diff
@@ -1088,6 +1146,71 @@ def assert_chiral_consistency(src_chiral_idxs: NDArray, dst_chiral_idxs: NDArray
...
+        raise NotImplementedError()
+
+
+class AlignedBond(AlignedPotential):
+    def interpolate(self, lamb):
+        k_min = 0.1
+        params = batch_interpolate_harmonic_bond_params(
+            self.src_params, self.dst_params, lamb, k_min, self.mins, self.maxes
+        )
+        params = jnp.array(params, dtype=np.float64).T
```
mcwitt, line 1167: nit: is this an unnecessary copy? (`params` is already a jax array if I'm reading correctly)

timemachine/fe/single_topology.py
```diff
@@ -1133,6 +1255,117 @@ def __init__(self, mol_a: Chem.Mol, mol_b: Chem.Mol, core: NDArray, forcefield:
...
+    def _align_bonds(self):
+        idxs, src_params, dst_params, mins, maxes = self._align_bonded_term(
+            interpolate.align_harmonic_bond_idxs_and_params,
+            self._assign_bond_idxs_min_max,
+            self.src_system.bond,
+            self.dst_system.bond,
+        )
+        idxs = idxs.reshape(-1, 2)
+        src_params = src_params.reshape(-1, 2)
+        dst_params = dst_params.reshape(-1, 2)
```
mcwitt, line 1289: Are these reshapes necessary? (Looks to me like the arrays returned by `_align_bonded_term` already have the correct shapes, but probably missing something.

timemachine/fe/single_topology.py
```diff
@@ -1270,116 +1503,16 @@ def _setup_end_state_dst(self):
...
-        return self._setup_intermediate_nonbonded_term(
-            self.src_system.nonbonded_pair_list,
-            self.dst_system.nonbonded_pair_list,
-            lamb,
-            interpolate.align_nonbonded_idxs_and_params,
-            interpolate.linear_interpolation,
-        )
-
     @property
+    @cache
```
mcwitt, line 1507: nit: use `functools.cached_property`

timemachine/fe/single_topology.py
proteneer: Yes, this was needed to fix a couple of tests when propers/chiral volumes were missing.  `_align_bonded_term ` calls:

```
  aligned_src_params = jnp.array([x[1] for x in aligned_tuples], dtype=np.float64)
  aligned_dst_params = jnp.array([x[2] for x in aligned_tuples], dtype=np.float64)
```

if `len(aligned_tuples) == 0` then this can return a `jnp.array` with the wrong dimensions.

mcwitt: Ah, got it. Empty arrays strike again.

timemachine/fe/interpolate.py
```diff
@@ -52,8 +56,8 @@ def align_idxs_and_params(
 
     Returns
     -------
-    set
-        set of tuples (idxs, src_params, dst_params)
+    3-tuple
```
proteneer, line 59: (nit: fix this, accidental change).

timemachine/fe/single_topology.py
proteneer: `params` is return from `vmap` around `interpolate_harmonic_bond_params`, which returns `lists`

timemachine/fe/interpolate.py
```diff
@@ -80,9 +84,6 @@ def align_idxs_and_params(
             validate_idxs(idxs)
 
     # used to convert arrays to a hashable type for use as dict keys and in sets
```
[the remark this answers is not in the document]
proteneer: willfix

timemachine/fe/single_topology.py
proteneer: added a note in docstring of `SingleTopology` for now:

"""
        Note: this class is assumed to be frozen and immutable, and changing attributes like the core
        and/or the forcefield will lead to undefined behavior.
"""

proteneer:
I also verified on 25 edges, 12 windows each, that the interpolated parameters on master and this branch are within machine precision of each other:

``` python

# compare.py
from jax import config
config.update("jax_enable_x64", True)

from timemachine.utils import path_to_internal_file
import numpy as np
from timemachine.fe import atom_mapping
from timemachine.fe.single_topology import SingleTopology
from timemachine.constants import DEFAULT_ATOM_MAPPING_KWARGS
from timemachine.fe.utils import get_mol_name
import pickle
from rdkit import Chem

def read_sdf(fname, removeHs: bool = False) -> list[Chem.Mol]:
    supplier = Chem.SDMolSupplier(str(fname), removeHs=removeHs)
    mols = [mol for mol in supplier]
    return mols

def _get_core_by_mcs(mol_a, mol_b):
    all_cores = atom_mapping.get_cores(
        mol_a,
        mol_b,
        **DEFAULT_ATOM_MAPPING_KWARGS,
    )

    core = all_cores[0]
    return core

def test_hif2a_end_state_stability(num_pairs_to_setup=25):
    """
    Pick some random pairs from the hif2a set and ensure that they're numerically stable at the
    end-states under a distance based atom-mapping protocol. For a subset of them, we will also run
    simulations.
    """

    seed = 2024

    with path_to_internal_file("timemachine.testsystems.data", "ligands_40.sdf") as path_to_ligand:
        mols = read_sdf(path_to_ligand)

    pairs = [(mol_a, mol_b) for mol_a in mols for mol_b in mols]

    np.random.seed(seed)
    np.random.shuffle(pairs)
    res = []
    for pair_idx, (mol_a, mol_b) in enumerate(pairs[:num_pairs_to_setup]):
        print("Checking", get_mol_name(mol_a), "->", get_mol_name(mol_b))
        core = _get_core_by_mcs(mol_a, mol_b)
        for lamb in np.linspace(0, 1.0, 24):
            gs = st.setup_intermediate_state(lamb)
            res.append((
                gs.bond.params,
                gs.angle.params,
                gs.proper.params,
                gs.improper.params,
                gs.chiral_atom.params,
                gs.nonbonded_pair_list.params))

    with open("res_branch.npz", "wb") as fh:
        pickle.dump(res, fh)

if __name__ == "__main__":
    test_hif2a_end_state_stability(25)
```

``` python
# diff.py
import pickle
import numpy as np

with open("res_branch.npz", "rb") as fh:
    ref_objs = pickle.load(fh)

with open("res_master.npz", "rb") as fh:
    test_objs = pickle.load(fh)

for a,b in zip(ref_objs, test_objs):
    for idx, (aa, bb) in enumerate(zip(a, b)):
        np.testing.assert_allclose(aa, bb)
```

mcwitt (approved):

proteneer:
Thanks @badisa  for helping me track down an annoying subtley:

While this PR emits the exact same parameters for potential objects in 32-bit, it fails to do preserve it in 64-bit. This unfortunately leads to different bitwise results. 

We can try some tricks to improve the behavior of this (eg. by round tripping 64 bit -> 32 bit -> 64bit), so very small tolerances in 64bit are still permitted.

timemachine/fe/single_topology.py
proteneer: I need to think about how to organize this a little better - deferring it

PR #1526, opened 2025-03-18
Tracks water sampling proposals in run_sims_hrex

badisa:
* Enables tracking of acceptances of water moves in HREX
* Adds plotting code to plot acceptances by state

## Example Water Sampling plot
Note that the total number of proposals in each state is in the title, as the ratio of acceptances to proposals tends to be ~ 1 in a million proposals. The example plot is taken from the tests where the acceptance rate is much higher. 
example_ws

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ memcheck_tests:
 
 .PHONY: unit_tests
 unit_tests:
-	pytest -m 'not $(NOCUDA_MARKER) and not $(NOGPU_MARKER) and not $(MEMCHECK_MARKER) and not $(NIGHTLY_MARKER)' $(PYTEST_CI_ARGS)
+	pytest -x -m 'not $(NOCUDA_MARKER) and not $(NOGPU_MARKER) and not $(MEMCHECK_MARKER) and not $(NIGHTLY_MARKER)' $(PYTEST_CI_ARGS)
 
 .PHONY: nightly_tests
 nightly_tests:
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,7 +1,7 @@
 [pytest]
 xfail_strict = true
 verbosity_assertions = 2
-addopts = -svx --hypothesis-profile=no-deadline
+addopts = -sv --hypothesis-profile=no-deadline
 testpaths =
     tests/
 markers =
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -254,19 +254,19 @@ def test_run_rbfe_legs(
     # Hashes are of results.npz, lambda0_traj.npz and lambda1_traj.npz respectively.
     leg_results_hashes = {
         "vacuum": (
-            "e71c2c30a18ea154f4eb6c6fb544aeadb0e9dda91c592115cd3eb6f40947df42",
+            "3b2fe61d8df0d80edcf482d9dfe8c38aabf79b1108b17a9042321f7d8d380ee4",
             "c74ba1b503532af7fe83be82090cca82c33f5fa72744e74613dcaad9d978283a",
-            "59ca76d1c5d43925ae6b163667c4b3e60f6c765ce10ad85d4f1d14c7d7854933",
+            "69bcc498fba25febbf1a9c0bd2bda6ebd8855ef382d731b17307d4b7a4ae2540",
         ),
         "solvent": (
-            "b4af69c91a1cfdbdf911c27f929de56f2436b20d5c6d0b3bc3f59021065df2a2",
+            "d6464cc055e486acf6d6d1311aadc90d0ba27118d818be3750d569ace27f39bf",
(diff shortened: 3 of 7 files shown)


timemachine/fe/free_energy.py
```diff
@@ -1461,6 +1463,7 @@ def get_flattened_params(initial_state: InitialState) -> NDArray:
     last_update_time = begin_loop_time
 
     for current_frame in range(md_params.n_frames):
+        per_state_water_sampling = [(0, 0) for _ in range(len(initial_states))]
```
mcwitt, line 1466: nit: could this name be more descriptive? I don't have great ideas; e.g. `water_sampling_acceptance_proposal_counts_by_state` seems a bit long but maybe still preferable?

timemachine/fe/free_energy.py
```diff
@@ -1449,6 +1450,7 @@ def get_flattened_params(initial_state: InitialState) -> NDArray:
 
     samples_by_state: list[Trajectory] = [Trajectory.empty() for _ in initial_states]
     replica_idx_by_state_by_iter: list[list[ReplicaIdx]] = []
+    water_sampler_proposals: list[list[tuple[int, int]]] = []
```
mcwitt, line 1453: nit/naming: following the pattern of the other fields, can we indicate the dimensions in the name? (e.g. maybe `water_sampler_proposals_by_state_by_iter`, assuming that is correct?

timemachine/fe/free_energy.py
```diff
@@ -1494,6 +1501,13 @@ def sample_replica(xvb: CoordsVelBox, state_idx: StateIdx) -> tuple[NDArray, NDA
             )
             assert frame.shape[0] == 1
 
+            if md_params.water_sampling_params is not None:
+                for mover in context.get_movers():
+                    if isinstance(mover, WATER_SAMPLER_MOVERS):
```
mcwitt, line 1506: nit: would it be simpler to retain a handle to the water sampler mover above to avoid the need to search for it again here?

timemachine/fe/free_energy.py
```diff
@@ -1472,10 +1475,14 @@ def sample_replica(xvb: CoordsVelBox, state_idx: StateIdx) -> tuple[NDArray, NDA
...
             current_step = current_frame * md_params.steps_per_frame
             # Setup the MC movers of the Context
+            starting_water_acceptances = 0
+            starting_water_proposals = 0
             for mover in context.get_movers():
                 if md_params.water_sampling_params is not None and isinstance(mover, WATER_SAMPLER_MOVERS):
                     assert water_params_by_state is not None
                     mover.set_params(water_params_by_state[state_idx])
+                    starting_water_proposals = mover.n_proposed()
+                    starting_water_acceptances = mover.n_accepted()
```
mcwitt, line 1485: Does this logic assume that there is only one water sampler in `context.get_movers()`? If so, should we assert this here?

timemachine/md/exchange/exchange_mover.py
```diff
@@ -51,6 +53,28 @@ def translate_coordinates(coords, new_loc):
     return centered_coords + new_loc
 
 
+@dataclass(frozen=True)
+class WaterSamplingDiagnostics:
+    proposals_by_state_by_iter: NDArray
+    replica_idx_by_state_by_iter: list[list[ReplicaIdx]] | None = None
```
mcwitt, line 59: It seems redundant to include this here, since it's already in `HREXDiagnostics`. I think it'd make more sense to me for `proposals_by_replica_by_iter` to be a method accepting a `replica_idx_by_state_by_iter` argument, or to rely on a general utility to invert permutations, since we now have a few instances of functions from `x_by_state -> x_by_replica`.

timemachine/fe/plots.py
```diff
@@ -711,3 +711,14 @@ def plot_dummy_a_interpolation_schedule(st, n_windows=48):
 
 def plot_dummy_b_interpolation_schedule(st, n_windows=48):
     plot_interpolation_schedule(st, st.any_idxs_belong_to_dummy_b, "Dummy Group B Interpolation Schedule", n_windows)
+
+
+def plot_water_proposals_by_state(lambdas: Sequence[float], proposals_by_state: NDArray[np.int32], prefix: str = ""):
+    assert proposals_by_state.shape == (len(lambdas), 2)
+    # Determine total number of proposals
+    proposals_per_state = proposals_by_state[0][1]
```
mcwitt, line 719: Should we assert that the number of proposals per state is constant?

timemachine/md/exchange/exchange_mover.py
mcwitt: > Could you write a general utility for the permutations? Would be very handy.

I had something like this started in https://github.com/proteneer/timemachine/pull/1508, but ran into some minor annoyances making it compatible with both JAX and numpy arrays. Will plan to revisit.

mcwitt (approved):

PR #1529, opened 2025-03-19
Adds support for neutralizing systems and adding ions

badisa:
* Useful for exploring whether or not neutralizing systems or adding ions makes a difference for RBFE.

### DIFF
--- a/tests/rest/test_single_topology_rest.py
+++ b/tests/rest/test_single_topology_rest.py
@@ -6,6 +6,7 @@ import jax
 import matplotlib.pyplot as plt
 import numpy as np
 import pytest
+from numpy.typing import NDArray
 from rdkit import Chem
 from rdkit.Chem import AllChem, rdMolDescriptors
 
@@ -13,6 +14,7 @@ from timemachine.constants import DEFAULT_ATOM_MAPPING_KWARGS
 from timemachine.fe import atom_mapping
 from timemachine.fe.free_energy import HostConfig
 from timemachine.fe.rbfe import Host, setup_optimized_host
+from timemachine.fe.rest.bond import mkproper
 from timemachine.fe.rest.interpolation import (
     Exponential,
     InterpolationFxnName,
@@ -86,7 +88,8 @@ def test_single_topology_rest_vacuum(mol_pair, temperature_scale_interpolation_f
 
     state = st_rest.setup_intermediate_state(lamb)
     state_ref = st.setup_intermediate_state(lamb)
-    assert len(st_rest.candidate_propers) < len(state_ref.proper.potential.idxs)
+    assert set(st_rest.propers) == set(mkproper(*idxs) for idxs in state_ref.proper.potential.idxs)
+    assert set(st_rest.candidate_propers.values()) < set(st_rest.propers)
 
     ligand_conf = st.combine_confs(get_romol_conf(mol_a), get_romol_conf(mol_b))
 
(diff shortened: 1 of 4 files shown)


maxentile (approved):

tests/test_builders.py
```diff
@@ -90,6 +92,140 @@ def test_build_protein_system_returns_correct_water_count():
...
+    negative_mol = ligand_from_smiles("[N+](=O)([O-])[O-]")
+    neutral_mol = ligand_from_smiles("c1ccccc1")
+
+    box_size = 2.0
+
+    host_config_no_ions = build_water_system(box_size, DEFAULT_WATER_FF, ionic_concentration=0.0)
+    # Host system will have zero net charge if no ionic concentration and not neutralized
+    assert np.sum(host_config_no_ions.host_system.nonbonded_all_pairs.params[:, NBParamIdx.Q_IDX]) == 0.0
+
+    # Can't mix ligands of different charges when neutralizing the system
```
proteneer, line 109: Is there a hard reason for this? (Eg. adding a -1 mol and +1 mol should end up in not needing any neutralization)

tests/test_builders.py
```diff
@@ -90,6 +92,140 @@ def test_build_protein_system_returns_correct_water_count():
...
+            assert num_ions % 2 == abs(expected_charge)
+        else:
+            # Should have the number of ions extra to account for the charge of the ligand
+            assert num_ions == abs(expected_charge)
+
+
+@pytest.mark.nocuda
+@pytest.mark.parametrize("ionic_concentration", [0.0, 0.15])
+@pytest.mark.parametrize("neutralize", [False, True])
+def test_protein_system_ion_concentration_and_neutralization(ionic_concentration, neutralize):
```
proteneer, line 156: There's a lot of duplicate code with the above test - can DRY a bit?

timemachine/md/builders.py
```diff
@@ -77,24 +92,83 @@ def get_waters_to_delete():
...
+    box: NDArray[np.float64],
+    ff: app.ForceField,
+    water_ff: str,
+    mols: list[Chem.Mol] | None,
+    ionic_concentration: float = 0.0,
+    neutralize: bool = False,
+):
+    """Utility function to handle the neutralization of adding mols"""
+    dummy_chain_id = "DUMMY"
+    if neutralize and mols is not None:
```
proteneer, line 129: What happens if `mols=[]` empty list gets passed in? I think `charges[0]` would be invalid?

tests/test_builders.py
proteneer: ah got it, so typical use case is `len(mols)=2` that correspond to the end-state molecules. I think later on, after we refactor the single topology code, we'll be able to seamlessly morph counter-ions into waters or other counter-ions to neutralize the system.

timemachine/md/builders.py
```diff
@@ -77,24 +92,83 @@ def get_waters_to_delete():
...
+def solvate_modeller(
+    modeller: app.Modeller,
+    box: NDArray[np.float64],
+    ff: app.ForceField,
+    water_ff: str,
+    mols: list[Chem.Mol] | None,
+    ionic_concentration: float = 0.0,
+    neutralize: bool = False,
+):
+    """Utility function to handle the neutralization of adding mols"""
```
proteneer, line 127: Maybe also note that this is a non-idempotent method that will directly modify the modeler object? (Or maybe it should deep copy and return a clone?)

timemachine/md/builders.py
```diff
@@ -77,24 +92,83 @@ def get_waters_to_delete():
...
+            topology = app.Topology()
+            dummy_chain = topology.addChain(dummy_chain_id)
+            for _ in range(abs(charge)):
+                if charge < 0:
+                    res = topology.addResidue(CHLORINE_ION_RESIDUE, dummy_chain)
+                    topology.addAtom(CHLORINE_ION_RESIDUE, app.Element.getBySymbol("Cl"), res)
+                else:
+                    res = topology.addResidue(SODIUM_ION_RESIDUE, dummy_chain)
+                    topology.addAtom(SODIUM_ION_RESIDUE, app.Element.getBySymbol("Na"), res)
+            coords = np.zeros((topology.getNumAtoms(), 3)) * unit.angstroms
```
proteneer, line 147: Do we ever use the coords from the modeller directly for anything? (Maybe set to `np.nan` to be even more defensive?).

proteneer: Hm right, but as-is it sets them all to `(0,0,0)`?

proteneer: Hm... weird that it's even attempting to cast a `float` to an `int`. Anyways - I guess adding it to the origin is probably (?) okay - was just a bit surprised that `addSolvent` doesn't check for clashes or anything on the existing atoms. Will `addSolvent` add atoms/molecules that avoids the "origin" (0,0,0) in this case?

proteneer (approved):

PR #1531, opened 2025-03-20
Separate out Local MD Params from MDParams

badisa:
* Be consistent, also lets us add some more local MD parameters

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -11,11 +11,14 @@ import jax
 import numpy as np
 from hilbertcurve.hilbertcurve import HilbertCurve
 from numpy.typing import NDArray
+from rdkit import Chem
+from rdkit.Chem import AllChem
 
 from timemachine.constants import DEFAULT_TEMP, ONE_4PI_EPS0
 from timemachine.fe import rbfe
 from timemachine.fe.free_energy import HostConfig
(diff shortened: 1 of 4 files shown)


timemachine/fe/free_energy.py
```diff
@@ -139,30 +139,40 @@ def __post_init__(self):
...
-    k: float = 1_000.0  # kJ/mol/nm^4
-    min_radius: float = 1.0  # nm
-    max_radius: float = 3.0  # nm
-    freeze_reference: bool = True
 
+    # Set to LocalMDParams to enable local MD
+    local_md_params: LocalMDParams | None = None
     # Set to HREXParams or None to disable HREX
-    hrex_params: Optional[HREXParams] = None
+    hrex_params: HREXParams | None = None
```
mcwitt, line 166: nit: should we adopt a convention for this? Personally I'm OK with replacing `Optional[A]` with `A | None`, but would prefer we do this consistently if at all.

mcwitt (approved):

timemachine/fe/free_energy.py
mcwitt: OK. Mostly just trying to understand what's essential to the PR vs. flipping conventions.

PR #1535, opened 2025-03-24
Fix REST heuristics

mcwitt:
Fixes 2 separate bugs in the REST heuristics introduced in #1524.

1. Fixes the implementation of `target_torsions`. Previously, the filter predicate was trivially true due to a subtle issue with the implementation using `dataclasses.astuple`.
2. Refines REST softening of ligand-ligand nonbonded interactions to only include pairs with at least one atom in the REST region. (Previously all pairs were softened.)

TODO
- [x] Update result hashes in nightly tests

### DIFF
--- a/tests/rest/test_single_topology_rest.py
+++ b/tests/rest/test_single_topology_rest.py
@@ -114,8 +114,8 @@ def test_single_topology_rest_vacuum(mol_pair, temperature_scale_interpolation_f
         assert energy_scale < 1.0
 
         # check that a proper subset of ligand-ligand nonbonded interactions are scaled
-        assert U_nonbonded != U_nonbonded_ref
-        assert U_nonbonded != energy_scale * U_nonbonded_ref
+        assert not np.isclose(U_nonbonded, U_nonbonded_ref)
+        assert not np.isclose(U_nonbonded, energy_scale * U_nonbonded_ref)
 
         if has_rotatable_bonds or has_aliphatic_rings:
             assert 0 < len(st_rest.candidate_propers)
--- a/tests/rest/test_single_topology_rest.py
+++ b/tests/rest/test_single_topology_rest.py
@@ -119,6 +119,7 @@ def test_single_topology_rest_vacuum(mol_pair, temperature_scale_interpolation_f
 
         if has_rotatable_bonds or has_aliphatic_rings:
             assert 0 < len(st_rest.candidate_propers)
+            assert not np.isclose(U_proper, U_proper_ref)
             assert U_proper < U_proper_ref
 
         def compute_proper_energy(state: GuestSystem, ixn_idxs: Sequence[int]):
--- a/tests/rest/test_single_topology_rest.py
+++ b/tests/rest/test_single_topology_rest.py
@@ -6,6 +6,7 @@ import jax
 import matplotlib.pyplot as plt
 import numpy as np
 import pytest
+from numpy.typing import NDArray
 from rdkit import Chem
 from rdkit.Chem import AllChem, rdMolDescriptors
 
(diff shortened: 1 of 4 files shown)


timemachine/fe/rest/single_topology.py
```diff
@@ -148,7 +148,7 @@ def target_propers(self) -> dict[int, CanonicalProper]:
         return {
             idx: proper
             for (idx, proper) in self.candidate_propers.items()
-            if any(idx in self.rest_region_atom_idxs for idx in astuple(proper))
+            if any(idx in self.rest_region_atom_idxs for idx in proper.idxs)
```
proteneer, line 151: was this a bug fix or a nit?

proteneer: edit: missed the description, this was a bug fix.

tests/rest/test_single_topology_rest.py
```diff
@@ -111,11 +113,13 @@ def test_single_topology_rest_vacuum(mol_pair, temperature_scale_interpolation_f
...
+        # check that a proper subset of ligand-ligand nonbonded interactions are scaled
+        assert not np.isclose(U_nonbonded, U_nonbonded_ref)
+        assert not np.isclose(U_nonbonded, energy_scale * U_nonbonded_ref)
+
         if has_rotatable_bonds or has_aliphatic_rings:
             assert 0 < len(st_rest.candidate_propers)
-
-            if energy_scale < 1.0:
-                assert U_proper < U_proper_ref
+            assert U_proper < U_proper_ref
```
proteneer, line 122: Is the energy strictly lower in the REST case? eg. if we have a torsion potential that was strictly between `[-10, 0]`, removing it would increase the energy instead?

timemachine/fe/rest/single_topology.py
mcwitt: This is the bug fix. `astuple` includes the `_unsafe` field of `CanonicalProper`:

```
>>> astuple(CanonicalProper.from_idxs(1, 2, 3, 4))
(1, 2, 3, 4, True)
```

leading to unintended results in the filter because e.g.

```
>>> True in {1}
True
```

tests/rest/test_single_topology_rest.py
proteneer, line 122: Should also sanity check that this isn't just testing some small floating point round-off, eg. `0.99999<1.00000`

mcwitt: Good point; this assumes that k is positive, which may not be strictly true in our parameterization? (It does seem to be for the examples tested here)

mcwitt: Added assertion in https://github.com/proteneer/timemachine/pull/1535/commits/ef9abac03369cc1270759ad4590f0e209050b493

timemachine/fe/rest/single_topology.py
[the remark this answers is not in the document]
mcwitt: Described in resolved convo above: https://github.com/proteneer/timemachine/pull/1535#discussion_r2010895262

tests/rest/test_single_topology_rest.py
```diff
@@ -111,11 +113,14 @@ def test_single_topology_rest_vacuum(mol_pair, temperature_scale_interpolation_f
     else:
         assert energy_scale < 1.0
 
+        # check that a proper subset of ligand-ligand nonbonded interactions are scaled
+        assert not np.isclose(U_nonbonded, U_nonbonded_ref)
```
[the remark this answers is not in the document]
mcwitt: > The tests below assert that the target propers and the candidate propers are different, but don't verify that rest region propers are adjusted while propers out of the rest region are not.

These assertions verify that the energy of the propers in the REST region is scaled and that of the propers not in the REST region is not scaled: https://github.com/proteneer/timemachine/blob/51248370bdd32ec207bb706fd5c0bfb6568a3e2f/tests/rest/test_single_topology_rest.py#L129-L141

proteneer (approved):
nice!

tests/rest/test_single_topology_rest.py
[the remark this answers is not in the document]
mcwitt: Added an assertion in https://github.com/proteneer/timemachine/pull/1535/commits/fe8a3a5cffa929b7fa464cc7f7fc0dd4c15a60c6 that only ligand-ligand ixns with >=1 atom in the REST region are scaled, using an alternate implementation to find the pair indices.

PR #1538, opened 2025-03-26
Aligned NonbondedInteractionGroup and NonbondedAllPairs potentials

proteneer:
This PR adapts the same alignment logic we use for bonded terms to the nonbonded terms, and re-organized the aligned potentials into `timemachine/fe/aligned_potential.py`.


tests/common.py
```diff
@@ -596,3 +598,62 @@ def ligand_from_smiles(smiles: str, seed: int = 2024) -> Chem.Mol:
     AllChem.EmbedMolecule(mol, randomSeed=seed)
     set_mol_name(mol, smiles)
     return mol
+
+
+def get_guest_params(mol_a, mol_b, atom_map_mixin, q_handle, lj_handle, lamb: float, cutoff: float) -> jax.Array:
```
proteneer, line 603: This is not a new function, it is moved from the SingleTopology class but now serves as a reference path for testing purposes.

tests/test_aligned_potential.py
```diff
@@ -0,0 +1,204 @@
...
+from timemachine.fe.single_topology import AtomMapMixin, SingleTopology
+from timemachine.fe.topology import get_ligand_ixn_pots_params
+from timemachine.fe.utils import get_romol_conf, read_sdf_mols_by_name
+from timemachine.ff import Forcefield
+from timemachine.md.builders import build_protein_system
+from timemachine.potentials import BoundPotential, Nonbonded, NonbondedInteractionGroup
+from timemachine.utils import path_to_internal_file
+
+
+def _parameterize_host_nonbonded_reference(
```
proteneer, line 21: Again, not a new function, moved from `SingleTopology`

tests/test_aligned_potential.py
```diff
@@ -0,0 +1,204 @@
...
+        scale_factors,
+        beta,
+        cutoff,
+        atom_idxs=np.arange(num_host_atoms, dtype=np.int32),
+    )
+
+    return combined_nonbonded.bind(hg_nb_params)
+
+
+def _parameterize_host_guest_nonbonded_ixn_reference(
```
proteneer, line 49: Again, not a new function, moved from SingleTopology

timemachine/fe/single_topology.py
```diff
@@ -1835,144 +1750,27 @@ def mol(self, lamb: float, min_bond_k: float = DEFAULT_BOND_IS_PRESENT_K) -> Che
...
+        return self._align_nonbonded_interaction_group(lhs_nb_ixn_group, rhs_nb_ixn_group, camm)
 
     def combine_with_host(
         self,
         host_system: HostSystem,
         lamb: float,
-        num_water_atoms: int,
-        ff: Forcefield,
-        omm_topology: OpenMMTopology,
+        num_water_atoms: int,  # deprecated/unused
```
proteneer, line 1771: This will be removed in a future PR.

timemachine/fe/single_topology.py
```diff
@@ -1835,144 +1750,27 @@ def mol(self, lamb: float, min_bond_k: float = DEFAULT_BOND_IS_PRESENT_K) -> Che
...
-        )
+    def _wrap_align_host_nonbonded(self, host_nonbonded: BoundPotential[Nonbonded]) -> AlignedNonbondedAllPairs:
+        lhs_nb_ap = build_end_state_reference_all_pairs(self.mol_a.GetNumAtoms(), host_nonbonded)
+        rhs_nb_ap = build_end_state_reference_all_pairs(self.mol_b.GetNumAtoms(), host_nonbonded)
+        camm = build_combined_atom_map_mixin(host_nonbonded.potential.num_atoms, self.mol_a, self.mol_b, self.core)
+        return self._align_nonbonded_all_pairs(lhs_nb_ap, rhs_nb_ap, camm)
 
-        bound_ixn_pot = ixn_pot.bind(ixn_params)
-        return bound_ixn_pot
+    def _wrap_aligned_host_guest_nonbonded_ixn_group(
```
proteneer, line 1759: This is temporary scaffolding code. Will be removed once `HostGuestSystems` can be directly passed into `SingleTopology`.

timemachine/fe/single_topology.py
```diff
@@ -1393,6 +1130,184 @@ def _align_nonbonded_pair_list(self):
...
+        for (i, j), sf in zip(src_exclusion_idxs, src_exclusion_scale_factors):
+            # mypy complaint: error: Value of type variable "SupportsRichComparisonT" of "sorted" cannot be "ndarray[tuple[int, ...], dtype[signedinteger[_32Bit]]]"  [type-var]
+            key = tuple(sorted([combined_atom_map_mixin.a_to_c[i], combined_atom_map_mixin.a_to_c[j]]))  # type: ignore
+            exclusions_kv[key] = sf
+
+        for (i, j), sf in zip(dst_exclusion_idxs, dst_exclusion_scale_factors):
+            # mypy complaint: error: Value of type variable "SupportsRichComparisonT" of "sorted" cannot be "ndarray[tuple[int, ...], dtype[Any]]"  [type-var]
+            key = tuple(sorted([combined_atom_map_mixin.b_to_c[i], combined_atom_map_mixin.b_to_c[j]]))  # type: ignore
+            if key in exclusions_kv:
+                np.testing.assert_array_equal(exclusions_kv[key], sf)
```
proteneer, line 1269: This is a hard requirement - we require that any duplicate exclusions in `src` and `dst` share the same scaling factors (since we're not allowed interpolate over them currently with our HREX `.set_params()` requirement).

timemachine/fe/single_topology.py
```diff
@@ -1393,6 +1130,184 @@ def _align_nonbonded_pair_list(self):
...
+            dst_nb_ixn_group.potential.col_atom_idxs,
+            dst_nb_ixn_group.params,
+            combined_atom_map_mixin.a_to_c,
+            combined_atom_map_mixin.b_to_c,
+            src_nb_ixn_group.potential.cutoff,
+        )
+
+        assert len(set(col_idxs).intersection(set(row_idxs))) == 0
+
+        # (ytz): does this have to be jnp?
```
proteneer, line 1196: To answer this question:

`test_single_topology_combined.py::test_nonbonded_host_params_independent_of_lambda()` fails otherwise.

timemachine/fe/single_topology.py
```diff
@@ -1158,81 +904,83 @@ def assert_torsions_defined_over_non_linear_angles(system: GuestSystem | HostGue
...
-            self.src_params, self.dst_params, lamb, self.mins, self.maxes
-        )
-        params = jnp.array(params).T
-        return PeriodicTorsion(self.idxs).bind(params)
+def build_end_state_reference_ixn_group(
+    mol: Chem.Mol, nonbonded_all_pairs: BoundPotential[Nonbonded], ff
+) -> BoundPotential[NonbondedInteractionGroup]:
+    guest_q = ff.q_handle.parameterize(mol)
+    guest_lj = ff.lj_handle.parameterize(mol)
+    guest_w = jnp.zeros_like(guest_q)
```
proteneer, line 951: A bit annoying, this is required for 
`test_single_topology_combined.py::test_nonbonded_host_params_independent_of_lambda()` to be traceable through `jax.jit`

From what I can tell, the only reason `guest_w` needs to be `jnp` since `guest_q` is emitted because our ff handlers emit `jnp`, even though it could technically be a numpy array.

proteneer:
Updated description and marked as ready for review.

proteneer:
Need to fix a few things with the interaction groups at the end-states - will re-request a review in an hour or so.

timemachine/fe/single_topology.py
proteneer: (comment is wrong - these still need to be here for now since the custom interaction group needs them.)

timemachine/fe/single_topology.py
```diff
@@ -944,31 +946,53 @@ def build_end_state_reference_all_pairs(
...
-    col_atom_idxs = np.arange(num_host_atoms)
 
-    host_cutoff = nonbonded_all_pairs.potential.cutoff
-    host_beta = nonbonded_all_pairs.potential.beta
-    num_atoms = num_host_atoms + mol.GetNumAtoms()
-    ixn_group = NonbondedInteractionGroup(num_atoms, row_atom_idxs, host_beta, host_cutoff, col_atom_idxs)
-    return ixn_group.bind(combined_params)
+    # (ytz): Typing errors
+    # timemachine/fe/single_topology.py:956: error: Item "NNHandler" of "SimpleChargeHandler | AM1BCCHandler | AM1CCCHandler | AM1BCCCCCHandler | PrecomputedChargeHandler | NNHandler | None" has no attribute "parameterize"  [union-attr]
+    # timemachine/fe/single_topology.py:956: error: Item "None" of "SimpleChargeHandler | AM1BCCHandler | AM1CCCHandler | AM1BCCCCCHandler | PrecomputedChargeHandler | NNHandler | None" has no attribute "parameterize"  [union-attr]
```
proteneer, line 959: ("None" can be dealt with by checking under an `if`)

proteneer:
Fixed issues with the NB Ixn Group, re-requested reviews.

timemachine/fe/single_topology.py
```diff
@@ -1158,81 +906,105 @@ def assert_torsions_defined_over_non_linear_angles(system: GuestSystem | HostGue
...
-        params = jnp.array(params).T
+    return AtomMapMixin(num_atoms_a, num_atoms_b, new_core)
 
-        return HarmonicBond(self.idxs).bind(params)
 
+def build_end_state_reference_all_pairs(
+    num_guest_atoms: int, host_nonbonded_all_pairs: BoundPotential[Nonbonded]
+) -> BoundPotential[Nonbonded]:
+    host_params = host_nonbonded_all_pairs.params
+    nb_params = np.concatenate([host_params, np.zeros((num_guest_atoms, host_params.shape[1]))])
```
jkausrelay, line 931: (keep the comment about why we have 0 params here): 

\# Note: The choice of zeros here is arbitrary. It doesn't affect the
\# potentials or grads, but any function like the seed could depend on these values.

proteneer: comment added in https://github.com/proteneer/timemachine/pull/1538/commits/097ef98248ad6fe6998af8d81514f91519709b36

timemachine/fe/single_topology.py
proteneer: Resolved.

tests/test_aligned_potential.py
```diff
@@ -0,0 +1,204 @@
...
+        np.testing.assert_array_equal(ref_nb_ap.potential.scale_factors, test_nb_ap.potential.scale_factors)
+
+        # a bit redundant, since if all the parameters are identical between the two BPs, all downstream calculations
+        # should also be identical.
+        # reference version
+        ref_u = ref_nb_ap(coords, box0)
+        test_u = test_nb_ap(coords, box0)
+        np.testing.assert_equal(ref_u, test_u)
+
+        # gpu version
```
jkausrelay, line 201: todo?

timemachine/fe/aligned_potential.py
```diff
@@ -0,0 +1,400 @@
...
+        qse_idxs = [NBParamIdx.Q_IDX, NBParamIdx.LJ_SIG_IDX, NBParamIdx.LJ_EPS_IDX]
+        src_qse = self.src_params[:, qse_idxs]
+        dst_qse = self.dst_params[:, qse_idxs]
+        qse = interpolate.linear_interpolation(src_qse, dst_qse, lamb)
+        src_w = self.src_params[:, NBParamIdx.W_IDX]
+        dst_w = self.dst_params[:, NBParamIdx.W_IDX]
+        w = interpolate_w_coord(src_w, dst_w, lamb).reshape(-1, 1)
+        # (ytz): deal with spurious loss of precision when we interpolate nearly identical values:
+        # eg. 0.3*x + 0.7*x ~= x
+        params = jnp.hstack([qse, w]).astype(np.float32).astype(np.float64)
```
jkausrelay, line 372: Should we just interpolate in 64 bit to prevent loss of precision?

tests/test_aligned_potential.py
proteneer: Fixed

timemachine/fe/aligned_potential.py
proteneer: We already interpolate in 64bit, and tagging @badisa for visibility here.

This is unfortunately really tricky. There's a bunch of things happening here:

1) Original diagnose was from `FAILED tests/test_examples.py::test_run_rbfe_legs[2025-15-30-complex-5-50-1000]`, where in `assert_ensembles_compatible` we test that water sampler parameters, which are read directly from the ixn group, are identical: `np.testing.assert_array_equal(water_sampler_params_a[non_ligand_idxs], water_sampler_params_b[non_ligand_idxs])`
2) On `master`, the assertion passes, because we *never* interpolate water parameters on the ixn group potential. They're instead built once, and the ligand/protein parameters are then `spliced` in. 
3) On this branch, we instead generate all the parameters (water, ligand, and protein) up front, and then interpolate *all* terms, even if the end-states are identical. Consequently we end-up doing a bunch of `0.5x + 0.5x` kind of operations on the water parameters.

This *hack* above, mitigates the loss of precision by *losing even more precision* through truncation down to 32-bit, i.e. small differences like `4.10000003, 4.10000002` get truncated down to `4.1` in 32-bit, then get up-casted identically to something like `4.10000001`.

proteneer: Another alternative is to modify `interpolate.[log]_linear_interpolation` from:

``` python
def linear_interpolation(src_params, dst_params, lamb):
    """
    Linearly interpolate between src and dst params
    """
    return (1 - lamb) * src_params + lamb * dst_params
```

to a variant where directly return src_params/dst_params if they're identical:

``` python
def linear_interpolation(src_params, dst_params, lamb):
    """
    Linearly interpolate between src and dst params
    """
    return np.where(src_params == dst_params, src_params, (1 - lamb) * src_params + lamb * dst_params)
```

@mcwitt - what do you think?

proteneer: Okay, playing around with this a bit. I was able to recover exactly equality by:

1) Modifying the reference code in `tests/common.py::get_guest_params()` to use `interpolate.linear_interpolation` instead
2) Updating `interpolate.linear_interpolation` itself to avoid interpolating if `src_params == dst_params`

proteneer: (See [get_guest_params](https://github.com/proteneer/timemachine/pull/1538/commits/d9d9f2a010dbf8225f1ef7396c7308008d30ac9c))

tests/common.py
mcwitt: Rather than duplicate here, could this be defined in the `single_topology` module and serve as the implementation of `SingleTopology._get_guest_params`?

tests/test_aligned_potential.py
mcwitt: (same comment as https://github.com/proteneer/timemachine/pull/1538/files#r2017057695)

tests/common.py
proteneer: This function isn't used anymore by anything inside of the SingleTopology class anymore.

timemachine/fe/aligned_potential.py
```diff
@@ -0,0 +1,398 @@
...
+
+    return pair_params
+
+
+@dataclass(kw_only=True)
+class AlignedPotential:
+    src_params: Union[NDArray[np.float64], jax.Array]
+    dst_params: Union[NDArray[np.float64], jax.Array]
+    mins: Optional[NDArray[np.float64]]
+    maxes: Optional[NDArray[np.float64]]
```
mcwitt, line 282: nit: I find the naming a bit unclear; better as `lambda_min`, `lambda_max`?

tests/common.py
proteneer: `SingleTopology._get_guest_params()` has been fully removed.

timemachine/fe/aligned_potential.py
mcwitt: nit: worth adding a `__post_init__` assertion that `np.all(0.0 < mins)`, `np.all(mins < maxes)`, `np.all(maxes < 1.0)`, etc.?

timemachine/fe/single_topology.py
jkausrelay: Can you add this as a comment in the code?

timemachine/fe/aligned_potential.py
```diff
@@ -0,0 +1,398 @@
...
+
+    def interpolate(self, lamb):
```
mcwitt, line 284: nit: consider making `AlignedPotential` subclass `ABC`, decorate this with `@abstractmethod` (this will make `AlignedPotential()` a runtime error)

timemachine/fe/aligned_potential.py
```diff
@@ -0,0 +1,398 @@
...
+
+
+@dataclass(kw_only=True)
+class AlignedNonbondedPairlist(AlignedPotential):
+    idxs: NDArray[np.int32]
+    cutoff: float
+    beta: float
+
+    def interpolate(self, lamb):
+        # (ytz): batch_interpolate_nonbonded_pair_list_params currently fails to respect the self.mins and self.maxes
```
mcwitt, line 347: Worth asserting that `(self.mins is None and self.maxes is None) or np.all((self.mins == 0.0) & (self.maxes == 1.0))` to avoid surprises if this is unsupported?

timemachine/fe/aligned_potential.py
```diff
@@ -0,0 +1,398 @@
...
+
+@dataclass(kw_only=True)
+class AlignedNonbondedInteractionGroup(AlignedPotential):
+    row_atom_idxs: NDArray[np.int32]
+    col_atom_idxs: NDArray[np.int32]
+    cutoff: float
+    beta: float
+
+    def interpolate(self, lamb):
+        # (ytz): self.mins and self.max is currently ignored.
```
mcwitt, line 362: (same comment as https://github.com/proteneer/timemachine/pull/1538/files#r2017079070)

timemachine/fe/interpolate.py
```diff
@@ -150,7 +150,7 @@ def linear_interpolation(src_params, dst_params, lamb):
     """
     Linearly interpolate between src and dst params
     """
-    return (1 - lamb) * src_params + lamb * dst_params
+    return jnp.where(src_params == dst_params, src_params, (1 - lamb) * src_params + lamb * dst_params)
```
mcwitt, line 153: Might add a comment on why this is necessary?

timemachine/fe/aligned_potential.py
proteneer: A bit torn on preciseness here:

```
        if self.lambda_min is not None and self.lambda_max is not None:
            assert np.all(0.0 <= self.lambda_min)
            assert np.all(self.lambda_min <= self.lambda_max) # not sure if this should be < or <=
            assert np.all(self.lambda_max <= 1.0) 
        else:
            assert self.lambda_min is None and self.lambda_max is None
```

timemachine/fe/rest/single_topology.py
```diff
@@ -89,7 +91,7 @@ def rest_region_atom_idxs(self) -> set[int]:
         tends to result in larger REST regions than seem desirable.
         """
 
-        aligned_potentials: list[AlignedPotential] = [
+        aligned_potentials: list[Union[AlignedBond, AlignedAngle, AlignedTorsion]] = [
```
mcwitt, line 94: nit: consider
```suggestion
        aligned_potentials: list[AlignedBond | AlignedAngle | AlignedTorsion] = [
```

timemachine/fe/aligned_potential.py
proteneer: I can try it out, never played around with ABC + dataclass combos before.

timemachine/fe/single_topology.py
proteneer: comment inlined.

timemachine/fe/rest/single_topology.py
mcwitt, line 94: nit/later: might be able to clean this up by introducing an abstract `AlignedBondedPotential` that adds `idxs` to the fields

timemachine/fe/single_topology.py
proteneer: addressed in https://github.com/proteneer/timemachine/pull/1538/commits/1f5f30c9e58a747173bba292635d8b4e040dca91

timemachine/fe/rest/single_topology.py
proteneer: A bit torn - I thought about this as well, but was debating between the cost of having yet another class in the inheritance hierarchy.

proteneer (commented):
Comments

timemachine/fe/interpolate.py
proteneer: added in https://github.com/proteneer/timemachine/pull/1538/commits/f357272df94849dfbb05722c41c23af0452ec81d

timemachine/fe/single_topology.py
```diff
@@ -1158,81 +906,105 @@ def assert_torsions_defined_over_non_linear_angles(system: GuestSystem | HostGue
...
-        k_min = 0.1
-        params = batch_interpolate_harmonic_bond_params(
-            self.src_params, self.dst_params, lamb, k_min, self.mins, self.maxes
-        )
-        params = jnp.array(params).T
+    return AtomMapMixin(num_atoms_a, num_atoms_b, new_core)
 
-        return HarmonicBond(self.idxs).bind(params)
 
+def build_end_state_reference_all_pairs(
```
mcwitt, line 927: Might add a short docstring? (What does "reference" refer to here?)

timemachine/fe/aligned_potential.py
mcwitt: > A bit torn on preciseness (of None values, and < vs <=) here:

(edited my comment above; agreed that we need to allow [0, 1] inclusive)

I guess we might allow `min == max` too, in the case where we want an instantaneous change in the parameter?

timemachine/fe/single_topology.py
proteneer: Renamed `build_end_state_reference_all_pairs` -> `build_end_state_nonbonded_all_pairs` and `build_end_state_reference_ixn_group` -> `build_end_state_nonbonded_ixn_group`. Will add a docstring.

My intention is for this to ultimately go into the builders class with the next refactor.

timemachine/fe/aligned_potential.py
```diff
@@ -0,0 +1,409 @@
...
+        # check that
+        if self.lambda_min is not None and self.lambda_max is not None:
+            assert np.all(0.0 <= self.lambda_min)
+            assert np.all(self.lambda_min <= self.lambda_max)
+            assert np.all(self.lambda_max <= 1.0)
+        else:
+            assert self.lambda_min is None and self.lambda_max is None
+
+
+@dataclass(kw_only=True)
```
mcwitt, line 300: curiosity: is this just to force explicitness, or is there another reason this is necessary?

timemachine/fe/single_topology.py
mcwitt: > This is temporary scaffolding code.

Worth adding a comment to the code so we don't lose track of this?

timemachine/fe/single_topology.py
mcwitt: > we require that any duplicate exclusions in src and dst share the same scaling factors

question/unsure: is this needed given that we already have a check ensuring that the (unbound) potentials for all states are identical?

https://github.com/proteneer/timemachine/blob/71a20bd61400f72d12ae09ace74b9c1939667056/timemachine/fe/free_energy.py#L1430-L1433

timemachine/fe/aligned_potential.py
proteneer: Mostly to avoid ambiguity in the ordering of `args` in the constructor if the dataclass is inherited. (eg. if base class attributes like `src_params`, `dst_params`, `lambda_min`, `lambda_max`, should come before or after the subclass attributes like `idxs`, etc.)

timemachine/fe/aligned_potential.py
proteneer: Okay, I think we're aligned then (i.e. python code above meets both our expectations).

timemachine/fe/single_topology.py
proteneer: willfix

timemachine/fe/aligned_potential.py
proteneer: fixed

timemachine/fe/aligned_potential.py
proteneer: looks like they're compatible and working as intended. resolved

timemachine/fe/single_topology.py
```diff
@@ -1387,12 +1201,193 @@ def _align_nonbonded_pair_list(self):
...
             dst_params=dst_params,
-            mins=mins,
-            maxes=maxes,
+            lambda_min=mins,
+            lambda_max=maxes,
             cutoff=src_cutoff,
             beta=src_beta,
         )
 
+    def _process_ixn_group_idxs_and_params(
```
mcwitt, line 1210: nit: more descriptive name and/or docstring would be useful

timemachine/fe/single_topology.py
mcwitt: > I don't think it hurts to be defensive and try and catch this kind of behavior a little bit early

Agreed; mostly just wanted to check my understanding

mcwitt (approved):

timemachine/fe/single_topology.py
proteneer: Updated docstring, added type annotation, and renamed some variables to be more descriptive.

PR #1549, opened 2025-05-27
Improved nonbonded interpolation.

proteneer:
This PR improves the nonbonded interpolation schedule for replacement types of transformations. Some key changes are:

1) For insertions, the dummy group starts off with a scaled down epsilon and is fully decharged. It's `w`-coordinate is scaled from `cutoff` to `0` between `lambda=0` and `lambda=1/3`. From `lambda=1/3` to `lambda=2/3`, the epsilons and charges are linearly scaled up to the fully interacting values. 
2) At `lambda=0.5`, both R-groups are interacting with the environment.
3) The net charge is preserved throughout the lambda schedule. i.e:
``` python
np.testing.assert_array_almost_equal(np.sum(guest_charges), np.sum(guest_a_q), decimal=3)
np.testing.assert_array_almost_equal(np.sum(guest_charges), np.sum(guest_b_q), decimal=3)
```
4) New plotting code is added, splitting nonbonded and bonded interpolation schemes into two separate diagrams.
5) For a couple of identity transformations, roughly `10-20%` fewer lambda windows are observed.

### DIFF
--- a/tests/test_single_topology.py
+++ b/tests/test_single_topology.py
@@ -659,56 +659,56 @@ def test_setup_intermediate_state_not_unreasonably_slow(arbitrary_transformation
     assert elapsed_time / n_states <= 1.0
 
 
-@pytest.mark.nocuda
-def test_setup_intermediate_nonbonded_term(arbitrary_transformation):
-    """Tests that the current vectorized implementation _setup_intermediate_nonbonded_term is consistent with the
-    previous implementation"""
-    st, _ = arbitrary_transformation
+# @pytest.mark.nocuda
+# def test_setup_intermediate_nonbonded_term(arbitrary_transformation):
+#     """Tests that the current vectorized implementation _setup_intermediate_nonbonded_term is consistent with the
+#     previous implementation"""
+#     st, _ = arbitrary_transformation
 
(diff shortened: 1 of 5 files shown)


tests/test_single_topology.py
```diff
@@ -659,56 +659,56 @@ def test_setup_intermediate_state_not_unreasonably_slow(arbitrary_transformation
...
-            st.dst_system.nonbonded_pair_list,
-            lamb,
-            align_nonbonded_idxs_and_params,
-            linear_interpolation,
-        )
-        nonbonded_test = st.aligned_nonbonded_pair_list.interpolate(lamb)
-
-        np.testing.assert_array_equal(nonbonded_ref.potential.idxs, nonbonded_test.potential.idxs)
-        np.testing.assert_array_almost_equal(nonbonded_ref.params, nonbonded_test.params)
+# @pytest.mark.nocuda
```
proteneer, line 662: This is going to get nuked - too hard to maintain, and we have tests for end-state correctness anyways.

mcwitt: Makes sense to me. IIRC, this was mainly to convince ourselves that the existing code was equivalent to what it replaced in the last rewrite. (But now the existing code may as well be the source of truth.)

PR #1556, opened 2025-06-14
Keep Residue Ids when writing CIF files

tkramer-motion:
Keep residue ids from the topology when writing cif files using the CIFWriter. This allows users who use truncated proteins in FEP to keep residue ids consistent when looking at trajectories or representative structures from the simulation.

### DIFF
--- a/timemachine/fe/cif_writer.py
+++ b/timemachine/fe/cif_writer.py
@@ -90,10 +90,8 @@ class CIFWriter:
                     new_residue = combined_topology.addResidue(
                         name=old_residue.name,
                         chain=chain_obj,
-                        id=old_residue.id + (old_residue.insertionCode if old_residue.insertionCode else "")
-                        if old_residue.name != "HOH"
-                        else str(max([*used_residue_ids, 0]) + 1),
-                        insertionCode=old_residue.insertionCode
+                        id=old_residue.id if old_residue.name != "HOH" else str(max([*used_residue_ids, 0]) + 1),
+                        insertionCode=old_residue.insertionCode,
                     )
 
                     try:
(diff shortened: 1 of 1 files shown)


timemachine/fe/cif_writer.py
```diff
@@ -85,7 +87,19 @@ def __init__(self, objs, out_filepath):
                 old_atom_obj_kv = {}
                 for old_residue in old_topology.residues():
                     chain_obj = old_chain_obj_kv[old_residue.chain]
-                    new_residue = combined_topology.addResidue(name=old_residue.name, chain=chain_obj)
+                    new_residue = combined_topology.addResidue(
+                        name=old_residue.name,
+                        chain=chain_obj,
+                        id=(old_residue.id + old_residue.insertionCode if old_residue.insertionCode else "")
```
proteneer: Couple of questions:

1) Is `old_residue.id` always an integer?
2) The current logic sets `new_residue.id = ""` if `old_residue.insertionCode` evaluates to `False`, even if `old_residue.id` is set. Did you mean `(old_residue.id + old_residue.insertionCode if old_residue.insertionCode else old_residue.id)`?
3) Why not just set the `new_residue.insertionCode` directly based off of `old_residue.insertionCode`? (after new_residue has been instantiated, since it's ic is not part of the addResidue API). I think ICs have to be distinct from IDs:

*per https://bioinformatics.stackexchange.com/questions/11587/what-is-the-aim-of-insertion-codes-in-the-pdb-file-format
*openmm IC insertion for pdbx https://github.com/openmm/openmm/blob/master/wrappers/python/openmm/app/pdbxfile.py#L479C35-L480C1

tkramer-motion: 1. yes
2. good catch. parens was in the wrong spot.
3. done

proteneer (approved):

tkramer-motion:
So we need to add the insertion code to the residue id because of what looks like a bug in openmm to me here - https://github.com/openmm/openmm/blob/1ce5d91d9dedfdc273066fafa1a618bf05c25b85/wrappers/python/openmm/app/pdbxfile.py#L128 should be 

```
atomKey = ((row[resNumCol], row[chainIdCol], row[resInsertionCol], row[atomNameCol]))
```

proteneer:
Yuck - okay, let's file an issue with OpenMM? Meanwhile, lgtm.

proteneer (approved):

